PR #15, opened 2017-06-27
EVM.configure now takes a list of starting blocks and their VMs

gsalgado:
@pipermerriam Is this is what you had in mind? If so, should I delete the code that deals with range validation as well?


evm/vm/evm.py
```diff
@@ -5,10 +5,7 @@
 import rlp
 
 from evm.exceptions import (
-    ValidationError,
-)
-from evm.validation import (
-    validate_vm_block_ranges,
+    EVMNotFound, ValidationError,
```
pipermerriam: *nitpick*

Prefer these be split to one per line.

evm/vm/evm.py
```diff
@@ -52,27 +45,15 @@ def __init__(self, header):
         self.header = header
 
     @classmethod
-    def configure(cls, name=None, vm_block_ranges=None, db=None):
-        if vm_block_ranges is None:
+    def configure(cls, name=None, vm_configuration=None, db=None):
```
pipermerriam: oooh, +1 to the name change.  I like `vm_configuration`.

evm/vm/evm.py
```diff
@@ -125,10 +106,10 @@ def get_vm_class_for_block_number(cls, block_number):
         """
         Return the vm class for the given block number.
         """
-        range = find_range(tuple(cls.vms_by_range.keys()), block_number)
-        base_vm_class = cls.vms_by_range[range]
-        vm_class = base_vm_class.configure(db=cls.db)
-        return vm_class
+        for n in reversed(cls.vms_by_range.keys()):
```
pipermerriam: Can you change this to use a binary search.  The `bisect` module from the standard library will do this for you.

https://docs.python.org/3.0/library/bisect.html

Assuming the ranges are sorted which can be done as a normalization step during configuration, this should be trivial to implement.

evm/vm/evm.py
```diff
@@ -125,10 +106,10 @@ def get_vm_class_for_block_number(cls, block_number):
         """
         Return the vm class for the given block number.
         """
-        range = find_range(tuple(cls.vms_by_range.keys()), block_number)
-        base_vm_class = cls.vms_by_range[range]
-        vm_class = base_vm_class.configure(db=cls.db)
```
pipermerriam: I believe that this line still needs to be preserved.  The `db` property is set on the class and there are some `classmethod`s on the `VM` class that either currently require or are likely to require db access.

This manual API of passing the DB down through the object classes isn't ideal as it's easy to forget to do so if you have better ideas I'm all ears.

evm/vm/flavors/mainnet/__init__.py
```diff
@@ -28,5 +28,8 @@
 
 MainnetEVM = EVM.configure(
     'MainnetEVM',
-    vm_block_ranges=MAINNET_BLOCK_RANGES,
+    vm_configuration=(
+        (0, FrontierVM),
```
pipermerriam: *maybe* go ahead and set `constants.GENESIS_BLOCK_NUMBER = 0` and then use that value here.....   or is that silly?

evm/vm/evm.py
gsalgado: Done

evm/vm/flavors/mainnet/__init__.py
gsalgado: No, that sounds good to me

evm/vm/evm.py
gsalgado: I've just moved that to the `get_vm()` method as that way `get_vm_class_for_block_number()` returns the value passed in `vm_configuration` without modifying it, which makes testing easier.

But I just noticed that although `get_vm_class_for_block_number()` is a classmethod, it is only called from `get_vm()`, which is a regular method. Do you expect there to be other places that will call `get_vm_class_for_block_number()` using a EVM class or should it be a regular method as well?

evm/vm/evm.py
gsalgado: Maybe I'm mistaken but I think a binary search here wouldn't provide any significant improvements given that most of the time this will be called for the latest block number, in which case we'd almost invariably pick the last item from vms_by_range, no?  But even if that's not the case, I also wonder if vms_by_range will ever have enough items to warrant a binary search?

pipermerriam: +1 to your suggestion, optimizing for the common case.  I like it.  Disregard my request.

gsalgado:
I've made the changes you suggested, except for https://github.com/pipermerriam/py-evm/pull/15#discussion_r124305265, and pushed them in a separate commit to make it easier to review

evm/vm/evm.py
pipermerriam: You may have to look in the diff for https://github.com/pipermerriam/py-evm/pull/12 to see the how `get_vm_class_for_block_number()` elsewhere as I believe those uses are only present in that branch.

Short answer is that the `db` property needs to be set as a class level property since it will be used in both normal methods and class methods.

gsalgado: Even in #12 I only see `get_vm_class_for_block_number()` used in regular instance methods, and I'm still not sure I understand in which situations you expect to have an EVM class being passed around -- ISTM that upon starting the client we'd always get one of the pre-configured EVM classes (e.g. MainnetEVM), create an instance of it (with a db) and then pass that around.  If that's the case we should be able to have just EVM.configure() as a @classmethod. Or am I missing something?

gsalgado: My main gripe with this approach is that it feels unnatural to have classes configured with a DB, and also the fact that sometimes we use the configure() method and sometimes we use __init__() to set class/instance attributes. Maybe we need a clear distinction between class/instance attributes and standardise on using configure() for class attributes and __init__() for instance attributes. Something like https://github.com/pipermerriam/py-evm/pull/15/commits/a1752ccd2f89a686d8a63b1e65e77a327e7c2e1d

pipermerriam: I'm 👍  with a1752cc .  I've futzed with this API enough that I've lost track of some of the "why" certain things are the way they are.  Instance level attribute makes sense and is easier to reason about.  Lets see if we run into any situations where this doesn't work.

gsalgado: The only annoyance I encountered is that before this change, the `VM.get_block_class()` method would return a class with a `db` property set, and that was handy because we could just pass that to `rlp.decode()` to get a valid block, but now with this change we need to [pass the db as an extra keyword argument to rlp.decode()](https://github.com/gsalgado/py-evm/blob/issue-11/evm/vm/flavors/frontier/blocks.py#L144-L147)

gsalgado: Apart from that it seems to [work in all situations ;-)](https://travis-ci.org/pipermerriam/py-evm/builds/247939268). Should I squash those commits or would you rather keep them separate?

pipermerriam: I pushed one more commit to this branch so if you squash, make sure you grab that one.  Assuming CI passes, this all seems good to merge and I'll do so when it's green.  Thanks for doing this.

PR #21, opened 2017-06-29
Blockchain fixture tests

gsalgado:
Creating a new PR as I couldn't figure out how to push to https://github.com/pipermerriam/py-evm/tree/piper/blockchain-fixture-tests


evm/vm/base.py
```diff
@@ -122,6 +133,49 @@ def get_block_reward(self, block_number):
     def get_nephew_reward(self, block_number):
         return NEPHEW_REWARD
 
+    def import_block(self, block):
+        parent_header = self.evm.get_block_header_by_hash(block.header.parent_hash)
+        init_header = self.create_header_from_parent(parent_header)
+        evm = type(self.evm)(self.db, init_header)
+        vm = type(self)(evm, self.db)
```
pipermerriam: Just a comment as I'm not sure how to fix it but:

This is indicative that I've gotten an abstraction wrong.  The shared header object across the `evm`, `vm`, and `block` is making for awkward code like this.  It's not clear to me how to fix it but I think it's something that *needs* to eventually be sorted out.  If you come up with any ideas I'm all ears.

Here are some extra thoughts that may or may not help solve this.

* The `EVM` class uses the header to know the *current* block number and thus, which `VM` class needs to be used.
* The `VM` class uses the header to initialize its `self.block` instance.
* When mining a new block, the `EVM` class uses the header to initialized the header for the next block.
* The `EVM.header` property gets mutated in the methods `EVM.apply_transaction()`, `EVM.import_block()`, `EVM.mine_block()`, and `EVM.configure_header()`.  Mutation is inherently problematic in this situation because both the `EVM` and `VM` classes use the header as an anchor for the current chain state and since it gets mutated, it opens up ways for us to have two components with different opinions on what the current chain state should be.
* `EVM.from_genesis()` has awkward code at the end which first initializes the `EVM` instance with the genesis header and then replaces it with a new header initialized from the genesis header.

I suspect we're not looking at one architecture problem but 2 or 3 that are combining into something more hairy.

gsalgado: Yeah, that will have to be sorted out, ideally sooner rather than later, but right now I'm not sure how we should go about it. I hope the right abstractions will become more clear as I get more familiar with this part of the code.

Also, another abstraction that is not ideal is for block validation, which is currently done exclusively at the block level (in `Block.validate()`), but AIUI part of the block validation involves looking up a block's ancestor chain (up to the max uncle reference depth) as well as the uncles included in recent blocks.

pipermerriam: I think that `Block.validate()` can be fixed by setting up an api on the `VM`.  Something like `VM.validate_block(block)` which merely delegates to call both `block.validate()` as well as performing all performing all of the validation that occurs above the block.

evm/consensus/pow.py
```diff
@@ -0,0 +1,36 @@
...
+                len(nonce), nonce))
+
+    cache = mkcache_bytes(block_number)
+    mining_output = hashimoto_light(
+        block_number, cache, mining_hash, big_endian_to_int(nonce))
+    if mining_output[b'mix digest'] != mix_hash:
+        raise ValidationError("mix hash mistmatch; {0} != {1}".format(
+            encode_hex(mining_output[b'mix digest']), encode_hex(mix_hash)))
+    result = big_endian_to_int(mining_output[b'result'])
+    if not (result <= 2**256 // difficulty):
```
pipermerriam: You can use `evm.validation.validate_lte` for this.

evm/consensus/pow.py
```diff
@@ -0,0 +1,36 @@
...
+)
+from evm.utils.numeric import (
+    big_endian_to_int,
+)
+
+
+def check_pow(block_number, mining_hash, mix_hash, nonce, difficulty):
+    if len(mix_hash) != 32:
+        raise ValidationError(
+            "mix_hash should have length 32, got {0}: {1}".format(
```
pipermerriam: You can use `evm.validation.validate_length` for this.  The error message won't be quite as specific but I think that's ok for now.

evm/vm/base.py
```diff
@@ -188,6 +272,8 @@ def get_block_class(self):
 
     def get_block_by_hash(self, block_hash):
         block_header = self.evm.get_block_header_by_hash(block_hash)
+        if block_header is None:
+            return None
```
pipermerriam: Trying to understand the `None` return here.  My instinct says this should throw an exception when it can't find the requested block.

evm/vm/flavors/frontier/blocks.py
```diff
@@ -177,15 +240,20 @@ def add_transaction(self, transaction, computation):
...
+            gas_refund = computation.get_gas_refund()
+            tx_gas_used = (
+                transaction.gas - gas_remaining
+            ) - min(
+                gas_refund,
+                (transaction.gas - gas_remaining) // 2,
             )
-            gas_used = self.get_cumulative_gas_used() + base_gas_used - gas_refunded
+            logger = logging.getLogger('evm.foo')
+            logger.debug('REFUND: %s | tx_gas_used %s', gas_refund, tx_gas_used)
```
pipermerriam: These two logger lines are leftover debug code that can be removed.

evm/vm/base.py
gsalgado: I chose to return None when the block is not found because I assumed it'd be a common idiom to use this method to check if a block has been received by the node yet, and that a block not being found is not necessarily an exceptional occurrence. Happy to change it if you disagree

pipermerriam: My *aversion* to this mechanism is it creates the following pattern

```python
block = evm.get_block_by_hash(....)
if block is None:
    # handle it.
...
```

The thing I dislike about this pattern is that it requires knowledge that the return value *may* be `None` and *if* the user doesn't know this, then the error may arise in a very different part of the application.

If instead it raises an exception, then the error is guaranteed to be raised at the point that `get_block_by_hash` was called, and you can still have quite clean handling with `try/except`.

Thoughts?

gsalgado: Yeah, that's a good point, I'll change it to raise BlockNotFound

PR #26, opened 2017-07-13
Uncle ancestry and PoW validation

gsalgado:


evm/utils/db.py
```diff
@@ -3,6 +3,8 @@ def make_block_hash_to_number_lookup_key(block_hash):
     return hash_to_number_key
 
 
+# XXX: Isn't this going to be a problem when we have multiple chains, since
+# then there can be multiple blocks with a given number?
```
pipermerriam: Yes, this whole scheme is going to need to be reworked to deal with competing chain histories.

evm/vm/evm.py
```diff
@@ -246,18 +252,12 @@ def import_block(self, block):
...
-        # Now it actually makes sense as the instantiation of the EVM for the
-        # parent's block is right above (as opposed to inside the VM's
-        # import_block method), so it becomes clear that we import the block
-        # using that EVM for the parent and once everything is validated
-        # and saved in the DB we advance this EVM to the recently-imported block.
         imported_block = parent_evm.get_vm().import_block(block)
+        self.validate_block(imported_block)
 
         persist_block_to_db(self.db, imported_block)
+        # FIXME: We must only do this when the imported block has higher
```
pipermerriam: Just to standardize comments, lets use `# TODO:` for easy grepping of the codebase of *things* that need to be addressed.

evm/vm/evm.py
```diff
@@ -267,12 +267,58 @@ def mine_block(self, **mine_params):
...
+    def validate_seal(self, header):
+        check_pow(
+            header.block_number, header.mining_hash,
+            header.mix_hash, header.nonce, header.difficulty)
+
+    def get_ancestors(self, limit):
+        blocks = []
+        lower_limit = max(self.header.block_number - limit, 0)
+        for n in range(lower_limit, self.header.block_number):
+            blocks.insert(0, self.get_block_by_number(n))
```
pipermerriam: It is my understanding that this is one of the slowest operations you can do in python, inserting elements into position `0` of a list.  I'd suggest one of the following.

1. Reverse the range so that you can use `list.append(...)`
2. Use a `collections.deque` which has fast inserts to the head of lists.

gsalgado: Fair enough, changed to reverse the range

pipermerriam:
What's the state of this?  Ready for review/merge?  Work in progress?

gsalgado:
I've made the changes you suggested, and I think this implements all the uncle validity checks

pipermerriam:
+1 merged and good to go.

Issue #24, opened 2017-07-03
Pluggable ECC backends - Part 1


pipermerriam:
https://www.pivotaltracker.com/n/projects/2066033

### What is wrong?

The library current uses a local python implementation of the necessary ECC functions for signature creation and verification.  The library `coincurve` provides faster implementations of these functions but at the cost of requiring a system dependency.

* Some users will value being able to do a vanilla install without futzing with system dependencies
* Some users will value speed and be willing to install the necessary system dependencies to get it.

### How can it be fixed?

Implement a pluggable backend system for all of the ECC functionality.   This should start with a default backend which only delegates the current ECC functionality.

#### Implementation Details

* Since we don't have a formal *configuration* spec for Py-EVM yet, lets use environment variables.  for this one, lets use `PYEVM_ECC_BACKEND`
* Copy the `import_string` function from populus: https://github.com/pipermerriam/populus/blob/master/populus/utils/module_loading.py
* Implement a `get_ecc_backend_class` function which reads the `PYEVM_ECC_BACKEND` value from `os.environ` and imports the class.
* Implement a `get_ecc_backend` function which returns the instantiated ECC backend class.
* Implement a `BaseECCBackend` class which implements the base needed for ECC operations.
* Implement a `PurePythonECCBackend` class which uses the functions found at `evm.utils.ecdsa`.

DavidKnott:
@pipermerriam This looks complicated but I think I'll try to take a crack at it

PR #27, opened 2017-07-13
Create pure python ecc backend

DavidKnott:
I began to implement pluggable ECC backends in response to issue #24 
@pipermerriam Here's what I've done so far

#### Cute Animal Picture
image


pipermerriam:
Some architecture suggestions.

* Move `evm/utils/ecc_backend/ecc_backend.py` to `evm/ecc/__init__.py` so that we can do `from evm.ecc import ....`.  This import path seems more intuitive.
* Within `evm/ecc/__init__.py` write the following two simple functions.

```python
def get_ecc_backend_class(import_path=None):
    if import_path is None:
        import_path = os.environ.get('EVM_ECC_BACKEND_CLASS', 'evm.the_default_import_path')

def get_ecc_backend():
    backend_class = get_ecc_backend_class()
    return backend_class()
```

Then anywhere that the ecc backend is used, you can simply import `get_ecc_backend` and then call it within the function to get the backend class.

DavidKnott:
@pipermerriam  I made the changes you requested

PR #28, opened 2017-07-14
Avoid unnecessarily regenerating the ethash cache in check_pow()

gsalgado:
Before this change we'd re-generate it every time check_pow() was
called, but now we keep a cache of caches and only generated a new one
when needed.

Run time for the fast blockchain tests:

Before: ====================== 246 passed, 2 deselected in 400.65 seconds
 After: ====================== 246 passed, 2 deselected in 50.84 seconds

### DIFF
--- a/evm/consensus/pow.py
+++ b/evm/consensus/pow.py
@@ -1,7 +1,16 @@
-from pyethash import mkcache_bytes, hashimoto_light
+from collections import OrderedDict
+
+from pyethash import (
+    EPOCH_LENGTH,
+    hashimoto_light,
+    mkcache_bytes,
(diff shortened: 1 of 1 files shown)


evm/consensus/pow.py
```diff
@@ -15,11 +24,31 @@
 )
 
 
+cache_seeds = ['\x00' * 32]
+cache_by_seed = OrderedDict()
+cache_by_seed.max_items = 10
```
pipermerriam: If you want to add `pylru` to the dependencies to get access to a legit LRU cache I've got no problems with that.  Your call.

PR #30, opened 2017-07-19
Handle chain splits

gsalgado:


evm/vm/base.py
```diff
@@ -252,15 +252,17 @@ def get_block_class(self):
...
         """
         ancestor_depth = self.block.number - block_number
-        if 1 <= ancestor_depth <= 256:
-            return lookup_block_hash(self.db, block_number)
-        else:
+        if ancestor_depth > 256 or ancestor_depth < 1:
             return b''
+        h = get_block_header_by_hash(self.db, self.block.header.parent_hash)
+        while h.block_number != block_number:
+            h = get_block_header_by_hash(self.db, h.parent_hash)
```
pipermerriam: Some functional cleverness that avoids this loop mutation, though questionable if it's a good idea.  Requires the [`toolz`](http://toolz.readthedocs.io/) library.

```python
from toolz.functoolz import pipe

header = pipe(
    self.block.header.parent_hash,
    *itertools.repeat(
        lambda h: get_block_header_by_hash(self.db, h.parent_hash), 
        ancestor_depth,
    ),
)
return header
```

Mostly just wanted to see what it looks like but lately I've been loving the `toolz` library.  Lots of really nice utilities in there that have helped with avoiding mutation in my code.

pipermerriam:
Will give this a look tomorrow.

evm/vm/base.py
pipermerriam: can make the above even less readable by replacing the lambda

```python
import operator
from toolz.functoolz import pipe, compose, partial

header = pipe(
    self.block.header,
    *itertools.repeat(
        compose(
            partial(get_block_header_by_hash, self.db),
            operator.attrgetter('parent_hash'),
        ),
        ancestor_depth,
    ),
)
return header
```

evm/utils/blocks.py
```diff
@@ -26,12 +32,27 @@ def persist_block_to_db(db, block):
         rlp.encode(block.hash, sedes=rlp.sedes.binary),
     )
 
+
+def get_score(db, block_hash):
+    if block_hash == GENESIS_PARENT_HASH:
+        return GENESIS_DIFFICULTY
```
pipermerriam: This seems like a potential "future" problem as all chains will not start with the same genesis difficulty.  Is this something that can be EVM aware so that it can pull the genesis block and pull the genesis difficulty from there?

evm/vm/evm.py
```diff
@@ -243,30 +236,42 @@ def import_block(self, block):
         except BlockNotFound:
             raise ValidationError("Parent ({0}) of block {1} not found".format(
                 block.header.parent_hash, block.header.hash))
+
         init_header = self.create_header_from_parent(parent_header)
         parent_evm = type(self)(self.db, init_header)
         imported_block = parent_evm.get_vm().import_block(block)
```
pipermerriam: These two lines make me think this should be a method.  Something like `EVM.get_ancestor_evm(....)` or `EVM.get_parent_evm(...)`.  Hides some of the grossness of creating a temporary header and then using `type(self)` to derive the class.  Maybe it can be a `classmethod` to have the class directly available.

I'm not even remotely sure this is a good idea.  Use your best judgement.

evm/vm/evm.py
```diff
@@ -243,30 +236,42 @@ def import_block(self, block):
...
+        # It feels wrong to call validate_block() on self here, but we do that
+        # because we want to look up the recent uncles starting from the
+        # currenct canonical chain head.
```
pipermerriam: This comment doesn't make sense to me.  It's not clear to me what is *wrong* about this.

evm/vm/evm.py
```diff
@@ -243,30 +236,42 @@ def import_block(self, block):
...
 
-    def mine_block(self, **mine_params):
-        """
-        Mine the current block, applying
-        """
-        block = self.get_vm().mine_block(**mine_params)
-        self.validate_block(block)
-        persist_block_to_db(self.db, block)
-
+    def maybe_add_to_canonical_chain(self, block):
```
pipermerriam: Something doesn't sit well with me regarding the *maybe* aspect of this method.  I'm thinking it may be better to split this into two pieces.

* a function that returns whether the block should be the new canonical head of the chain.
* a function that sets the block as the new canonical head of the chain (regardless of the score).

Then we perform the check, and if the check confirms it should be the new head block, we then set it as the head block.  Thoughts?

evm/vm/evm.py
```diff
@@ -243,30 +236,42 @@ def import_block(self, block):
...
+        current_head = self.get_block_by_hash(self.header.parent_hash)
+        if get_score(self.db, block.hash) <= get_score(self.db, current_head.hash):
+            return
+
+        new_chain = []
+        b = block
+        while b.number >= GENESIS_BLOCK_NUMBER:
+            new_chain.append(b)
```
pipermerriam: Not a deal breaker but can you explore if there is a good way to remove this mutation?  I've had good luck maintaining readability in cases like this by extracting this stuff into a stand-alone function which uses `yield` statements in place of `.append(..)` calls.   Then you can decorate that function with [`eth_utils.to_tuple`](https://github.com/pipermerriam/ethereum-utils#to_tuplecallable--callable---tuple).

```python
from eth_utils import to_tuple

@to_tuple
def assemble_new_chain(...):
    while b.number > GENESIS_BLOCK_NUMBER:
        yield b
        ... # other code
```

evm/vm/evm.py
gsalgado: Yeah, that is a nice improvement, but it can't be a classmethod since we need access to instance variables (e.g. self.db). Also, the lines above 239 probably belong in the new method as well since they look up the current block's parent.

evm/vm/evm.py
gsalgado: What I meant is that it feels wrong because we use parent_evm to import the block but self to validate it. Should I just reword the comment to make that clear or do you think it's not necessary at all?

evm/vm/evm.py
gsalgado: I think that makes sense

evm/vm/evm.py
gsalgado: Oh, interesting, I'd seen this pattern of combining generators with the @to_tuple decorator and I wasn't sure what was the advantage, but now I understand you do it to avoid the mutation and make it more readable? Do you just find it's more readable than appending to a list and returning it at the end or is there anything else that makes you prefer this?

evm/utils/blocks.py
gsalgado: Hmm, yeah, it could be but then we'd probably need to extend the persist_block_to_db() function to take the block's total score as well as it doesn't have access to the evm. Would it be ok if I leave it as a TODO for now?

gsalgado:
I think I've addressed all your suggestions, so it's ready for another review when you have a chance

Issue #25, opened 2017-07-03
Pluggable ECC backends - Part 2: coincurve ECC backend.


pipermerriam:
### What is wrong?

Many users will be willing to install the necessary system dependencies to gain the speed provided by the `coincurve` library.

### How can it be fixed.

Once #24 is done, implement a coincurve based ECC backend which uses the coincurve library for faster ECC operations.

PR #31, opened 2017-07-20
Create and test coincurve ecc backend

DavidKnott:
This PR is in response to #25 
@pipermerriam Right now I have `coincurve` as an extra requirement like you suggested.  I'm also testing `coincurve` by comparing it with the pure python implementation.

#### Cute Animal Picture

image

### DIFF
--- a/evm/ecc/backends/coincurve.py
+++ b/evm/ecc/backends/coincurve.py
@@ -1,0 +1,79 @@
+from .base import BaseECCBackend
+
+from evm.utils.ecdsa import (
+    encode_signature,
+)
+
+from evm.utils.numeric import (
+    big_endian_to_int,
+    safe_ord
+)
+
+from evm.constants import (
+    NULL_BYTE,
+)
+
+from evm.utils.keccak import (
+    keccak,
+)
+
+from evm.utils.secp256k1 import (
+    decode_public_key,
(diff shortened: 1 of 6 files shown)


evm/ecc/backends/coincurve.py
```diff
@@ -0,0 +1,77 @@
+from .base import BaseECCBackend
+
+from coincurve.keys import (
```
pipermerriam: I would request that all `coincurve` imports be positioned within the `CoinCurveECCBackend` class itself so that this module can be properly imported even if `coincurve` isn't available.  Then, within the `__init__` method of the `CoinCurveECCBackend` class, do something like the following to produce a nice friendly import error if the library isn't available.

```python
def __init__(....):
    try:
        import coincurve
    except ImportError:
        raise ImportError("The CoinCurveECCBackend requires the coincurve library which is not available for import.")
    ...
```

setup.py
```diff
@@ -32,6 +32,9 @@
         "ethereum-bloom>=0.4.0",
         "pyethash>=0.1.27",
     ],
+    extra_require=[
```
pipermerriam: This should be:

```python
extra_require={
    'coincurve': [
        "coincurve>=4.5.1",
    ],
}
```

Issue #32, opened 2017-07-21
Enable and fix remaining Transaction Tests


pipermerriam:
https://github.com/pipermerriam/py-evm/blob/master/tests/json-fixtures/test_transactions.py#L44

```python
def transaction_fixture_skip_fn(fixture_path, fixture_name, fixture):
    # TODO: enable all fixture tests
    return "ttTransactionTest.json" not in fixture_path
```

Removing this line from `tests/json-fixtures/test_transactions.py` will enable about 150 transaction tests that are currently being skipped.  These should be enabled and fixed if they are failing.

djrtwo:
Looking for an entry point into this project. This seems like a good place, right?

pipermerriam:
Yes, this is open and up for grabs.

PR #34, opened 2017-07-23
Enable and fix remaining transaction tests

DavidKnott:
This PR is in response to #32 
It enables and fixes the remaining json-fixture transaction tests

#### Cute Animal Picture
http://www.cutestpaw.com/wp-content/uploads/2012/01/baby-tiger.jpg

### DIFF
--- a/tests/json-fixtures/test_transactions.py
+++ b/tests/json-fixtures/test_transactions.py
@@ -26,6 +26,10 @@ from evm.vm.flavors import (
     MainnetEVM,
 )
 
+from evm.utils.ecdsa import (
+    BadSignature,
+)
+
 from evm.utils.fixture_tests import (
     find_fixtures,
     normalize_transactiontest_fixture,
@@ -39,15 +43,9 @@ ROOT_PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, 'fixtures', 'TransactionTests')
 
 
-def transaction_fixture_skip_fn(fixture_path, fixture_name, fixture):
-    # TODO: enable all fixture tests
-    return "ttTransactionTest.json" not in fixture_path
-
-
 FIXTURES = find_fixtures(
     BASE_FIXTURE_PATH,
     normalize_transactiontest_fixture,
-    transaction_fixture_skip_fn,
 )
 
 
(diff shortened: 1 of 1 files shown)


evm/utils/fixture_tests.py
```diff
@@ -284,7 +284,7 @@ def normalize_transactiontest_fixture(fixture):
     try:
         normalized_fixture['rlp'] = decode_hex(fixture['rlp'])
     except binascii.Error:
-        normalized_fixture['rlpHex'] = fixture['rlp']
```
pipermerriam: So this is actually intentional (and likely important) so that there is a clear and unambiguous differentiation between RLP that has been decoded to bytes from it's hex form, and RLP that failed decoding.

tests/json-fixtures/test_transactions.py
```diff
@@ -80,12 +78,15 @@ def test_transaction_fixtures(fixture_name, fixture):
...
+        try:
+            assert transaction.sender == sender
+        except BadSignature:
+            assert not (27 <= transaction.v <= 34)
     else:
         # check RLP correctness
         try:
             transaction = rlp.decode(fixture['rlp'], sedes=TransactionClass)
-        except rlp.exceptions.ObjectDeserializationError:
+        except (rlp.exceptions.ObjectDeserializationError, rlp.exceptions.DecodingError, TypeError):
```
pipermerriam: I'm a little worried that the tests passing are because this is just catching *all* of the errors.  I'm traveling today and don't have time to check this, but can you verify that the errors being caught here are appropriate error conditions by potentially logging out how many of the tests succeed because of the subsequent `return` statement.

evm/utils/fixture_tests.py
DavidKnott: @pipermerriam I changed this back and am now checking for for the `KeyError` when the tests try to call `fixture['rlp']` (when it's now `fixture['rlpHex']`)

tests/json-fixtures/test_transactions.py
DavidKnott: @pipermerriam I've attached the tests that are passing due to these exceptions, most of them have WrongRLP in the title and I believe the ones that don't should be caught here.  I've attached a list of the errors that are passing due to the aforementioned exceptions.
[passing_due_to_rlp_exception.txt](https://github.com/pipermerriam/py-evm/files/1170218/passing_due_to_rlp_exception.txt)

tests/json-fixtures/test_transactions.py
```diff
@@ -80,12 +78,21 @@ def test_transaction_fixtures(fixture_name, fixture):
...
     else:
         # check RLP correctness
         try:
             transaction = rlp.decode(fixture['rlp'], sedes=TransactionClass)
-        except rlp.exceptions.ObjectDeserializationError:
+        except KeyError:
+            assert fixture['rlpHex']
+            return
+        except TypeError as err:
+            assert err.args == ("'bytes' object cannot be interpreted as an integer",)
```
pipermerriam: This is curious to me?  What is causing this error?

pipermerriam: In fact, can you toss a comment into each of these except blocks with a brief explanation of what they are catching.

DavidKnott: Definitely, `TypeError` occurs on 4 tests all called `RLPElementIsListWhenItShouldntBe` and it occurs when it tries to change a list of bytes into an int
```
value = [b'dog', b'god', b'cat']
def big_endian_to_int(value):
    return int.from_bytes(value, byteorder='big')
E       TypeError: 'bytes' object cannot be interpreted as an integer
```

PR #35, opened 2017-07-24
Implement difficulty calculation for homestead

gsalgado:


evm/vm/flavors/homestead/blocks.py
```diff
@@ -8,3 +14,8 @@
 
 class HomesteadBlock(FrontierBlock):
     transaction_class = HomesteadTransaction
+    fields = [
+        ('header', BlockHeader),
+        ('transactions', CountableList(HomesteadTransaction)),
+        ('uncles', CountableList(BlockHeader))
+    ]
```
pipermerriam: Side note:

It would be nice if rlp objects supported *proper* inheritance in much the same way as django models work.  We'd need a metaclass to inherit from and a `Field` class of some sort for each field.  Would be a better/nicer interface and would allow more fine-grained subclassing

tests/json-fixtures/test_blockchain.py
```diff
@@ -116,7 +115,14 @@ def test_blockchain_fixtures(fixture_name, fixture):
     #     assert rlp.encode(genesis_header) == fixture['genesisRLP']
 
     db = MemoryDB()
-    evm = MainnetEVM.from_genesis(
+
+    evm = MainnetEVM
+    if fixture_name.startswith('Homestead'):
+        evm = EVM.configure(
+            'HomesteadEVM',
+            vm_configuration=[(0, HomesteadVM)])
```
pipermerriam: *note*

It would be great if we can figure out an API for *re-configuring* start block numbers that was more elegant.  Fully re-specifying the `vm_configuration` is not elegant.  It isn't clear what this looks like but seemed worth surfacing to get more people thinking about it.

pipermerriam:
Just looked over this.  Looks good.  Lemme know what the status is and whether you're ready for final review/merge.

evm/vm/flavors/homestead/blocks.py
gsalgado: Indeed, that'd be nice

tests/json-fixtures/test_blockchain.py
gsalgado: Yeah, I agree the current approach is not the most elegant, but if this turns out to be the only place where we'd use it, it may not be worth the effort. I'll add a TODO to make sure we re-evaluate the possibility of doing what you suggest later, if that's ok with you.

gsalgado:
I've added a TODO for the API to reconfigure start blocks; if you're ok with that I think this is ready for final review/merge

tests/json-fixtures/test_blockchain.py
pipermerriam: Sounds great.  Just leaving these things as I see them to try and keep both a record and a personal understanding of where some of the "bodies are buried".  Not even sure it's worth the TODO.

evm/vm/flavors/homestead/blocks.py
pipermerriam: I'm gonna open an issue over there.

pipermerriam: in case anyone cares https://github.com/ethereum/pyrlp/issues/44

pipermerriam:
will merge when it is green.  Thanks

PR #37, opened 2017-07-25
New VM flavor, with custom opcodes, for EIP150

gsalgado:


pipermerriam:
Ready for review?

evm/logic/system.py
```diff
@@ -52,7 +74,7 @@ def create(computation):
 
     call_data = computation.memory.read(start_position, size)
 
-    create_msg_gas = computation.gas_meter.gas_remaining
+    create_msg_gas = max_child_gas_modifier(computation.gas_meter.gas_remaining)
```
pipermerriam: This doesn't sit well with me.  Functions which pass around methods strike me as harder to follow than using a class.  Do you think there is a cleaner way to do this with a class abstraction similar to how it's done with the `Call` classes?

evm/vm/flavors/eip150/opcodes.py
```diff
@@ -0,0 +1,76 @@
...
+    context,
+    storage,
+    system,
+)
+
+from evm.vm.flavors.homestead.opcodes import HOMESTEAD_OPCODES
+
+
+NEW_OPCODES = {
+}
```
pipermerriam: Can we drop `NEW_OPCODES` since it's empty?

gsalgado: Sure, removed

evm/logic/system.py
gsalgado: The last commit changes it to use a class like Call. I don't find it significantly cleaner, but I don't feel strong about either way

Issue #33, opened 2017-07-21
Database Backends


pipermerriam:
### What is wrong?

Currently, PyEVM is using the `MemoryDB` from `py-trie` for it's database.

1. We need to use something "local" to the project since this is not a dependency we want to have.
2. We can't go forever using an in-memory database.

### How can it be fixed

This can be done in multiple steps, but the general plan should be:

1. Define the API for a database backend.  In general I think the `MemoryDB` class from the `py-trie` library is a good starting point.  This should go in `evm.db.backends.base`
2. Create a local `MemoryDB` in PyEVM at `evm.db.backends.memory` and replace all uses of the `py-trie` version with the one local to the project.
3. Create a new backend that is backed by something like RocksDB or LevelDB
    - http://rocksdb.org/
    - http://leveldb.org/
4. If it seems useful/helpful create an API for getting these backends in much the same way as the ECC backend system works.

Fill this in if you know how to fix it.

PR #41, opened 2017-07-26
Create local db backend

DavidKnott:
This PR is in response to #33 
Now py-evm uses a local `MemoryDb` as opposed to the one from `py-trie`

#### Cute Animal Picture
image

### DIFF
--- a/evm/db/__init__.py
+++ b/evm/db/__init__.py
@@ -1,0 +1,22 @@
+import os
+
+from evm.utils.module_loading import (
+    import_string,
+)
+
+
+DEFAULT_DB_BACKEND = 'evm.db.backends.memory.MemoryDB'
+
+
+def get_db_backend_class(import_path=None):
+    if import_path is None:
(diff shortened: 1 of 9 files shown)


evm/db/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+def get_db_backend_class(import_path=None):
+    if import_path is None:
+        import_path = os.environ.get(
+            'EVM_DB_BACKEND_CLASS',
+            DEFAULT_DB_BACKEND,
+        )
+    return import_string(import_path)
+
+
+def get_db_backend():
```
pipermerriam: I've typically allowed passing in the `import_path` into this function as well so that users can customize the backend with *either* the environment variable, or by directly passing through the backend class import path they want to use.

DavidKnott:
@pipermerriam Made the change you requested, gonna start working on a `leveldb` backend

PR #43, opened 2017-07-27
Implement DAO fork rules

gsalgado:


evm/vm/flavors/homestead/headers.py
```diff
@@ -55,4 +59,138 @@ def configure_homestead_header(vm, **header_params):
             parent_header,
             header_params['timestamp'],
         )
+
+    # In geth the modification of the state in the DAO fork block is performed
+    # before any transactions are applied, so doing it here is the closest we
+    # get to that. Another alternative would be to do it in Block.mine(), but
+    # there we'd need to manually instantiate the State and update
+    # header.state_root after we're done.
```
pipermerriam: What about doing it in `VM.create_header_from_parent()`?

gsalgado: That is a @classmethod, so we wouldn't have access to the state.

pipermerriam:
@gsalgado see https://github.com/pipermerriam/py-evm/issues/55 for continued discussion of this architecture.

PR #44, opened 2017-07-27
Create and test leveldb backend

DavidKnott:
Creates and tests leveldb backend in response to issues #33 
@pipermerriam I'm not to happy with how I'm implementing the snapshot and revert functionality but after reading the docs of multiple dbs it seems like the best way (leveldb and rockdb both takes snapshots that are read-only which is why I have to manually enter the information when reverting).

#### Cute Animal Picture
image

### DIFF
--- a/evm/db/__init__.py
+++ b/evm/db/__init__.py
@@ -17,6 +17,6 @@ def get_db_backend_class(import_path=None):
     return import_string(import_path)
 
 
-def get_db_backend(import_path=None):
+def get_db_backend(import_path=None, **init_kwargs):
     backend_class = get_db_backend_class(import_path)
-    return backend_class()
+    return backend_class(**init_kwargs)
--- a/evm/db/backends/level.py
+++ b/evm/db/backends/level.py
@@ -1,0 +1,60 @@
+import shutil
+from .base import (
+    BaseDB,
+)
+
+
+class LevelDB(BaseDB):
+
+    # Creates db as a class variable to avoid level db lock error
+    def __init__(self, db_path=None):
+        if not db_path:
+            raise TypeError("Please specifiy a valid path for your database.")
+        try:
+            import leveldb
+        except ImportError:
+            raise ImportError("LevelDB requires the leveldb \
+                               library which is not available for import.")
+        self.db_path = db_path
+        self.db = leveldb.LevelDB(db_path, create_if_missing=True, error_if_exists=False)
+
+    def get(self, key):
+        # 'Get' Returns a bytearray which needs to be converted to straight bytes
+        return bytes(self.db.Get(key))
+
+    def set(self, key, value):
+        self.db.Put(key, value)
+
+    # Returns False instead of KeyError if key doesn't exist
+    def exists(self, key):
+        return bool(self.db.Get(key, default=False))
+
+    def delete(self, key):
+        self.db.Delete(key)
+
+    #
+    # Snapshot API
+    #
+    def snapshot(self):
+        return Snapshot(self.db.CreateSnapshot())
+
+    def revert(self, snapshot):
+        for item in self.db.RangeIter(include_value=False):
+            self.db.Delete(item)
+        for key, val in snapshot.items():
+            self.db.Put(key, val)
+
+    # Clears the leveldb
+    def __del__(self):
+        shutil.rmtree(self.db_path, ignore_errors=True)
+
+
+class Snapshot(object):
+    def __init__(self, snapshot):
+        self.db = snapshot
+
+    def get(self, key):
+        return self.db.Get(key)
+
+    def items(self):
+        return self.db.RangeIter(include_value=True, reverse=True)
--- a/tests/level-db/test_leveldb_db_backend.py
+++ b/tests/level-db/test_leveldb_db_backend.py
@@ -1,0 +1,70 @@
+import pytest
+from evm.db.backends.memory import MemoryDB
+from evm.db import (
+    get_db_backend,
+)
+
+
+pytest.importorskip('leveldb')
+
+
+# Sets db backend to leveldb
+@pytest.fixture
+def config_env(monkeypatch):
+    monkeypatch.setenv('CHAIN_DB_BACKEND_CLASS',
+                       'evm.db.backends.level.LevelDB')
+
+
+@pytest.fixture
+def level_db(config_env, tmpdir):
+    level_db_path = str(tmpdir.mkdir("level_db_path"))
+    return get_db_backend(db_path=level_db_path)
+
+
+@pytest.fixture
+def memory_db():
+    return MemoryDB()
+
+
+def test_raises_if_db_path_is_not_specified(config_env):
+    with pytest.raises(TypeError):
+        get_db_backend()
+
+
+def test_set_and_get(memory_db, level_db):
+    level_db.set(b'1', b'1')
+    memory_db.set(b'1', b'1')
+    assert level_db.get(b'1') == memory_db.get(b'1')
+
+
+def test_set_on_existing_value(level_db, memory_db):
+    level_db.set(b'1', b'2')
+    level_db.set(b'1', b'3')
+    memory_db.set(b'1', b'2')
+    memory_db.set(b'1', b'3')
+    assert level_db.get(b'1') == memory_db.get(b'1')
+
+
+def test_exists(level_db, memory_db):
+    level_db.set(b'1', b'2')
+    memory_db.set(b'1', b'1')
+    level_db.exists(b'1') == memory_db.exists(b'1')
+
+
+def test_delete(level_db, memory_db):
+    level_db.set(b'1', b'1')
+    memory_db.set(b'1', b'1')
+    level_db.delete(b'1')
+    memory_db.delete(b'1')
+    assert level_db.exists(b'1') == memory_db.exists(b'1')
+
+
+def test_snapshot_and_revert(level_db):
+    snapshot = level_db.snapshot()
+    level_db.set(b'1', b'1')
+    assert level_db.get(b'1')
+    with pytest.raises(KeyError):
+        snapshot.get(b'1')
+    level_db.revert(snapshot)
+    with pytest.raises(KeyError):
+        level_db.get(b'1')
--- a/tox.ini
+++ b/tox.ini
@@ -1,6 +1,6 @@
 [tox]
 envlist=
-    py{27,34,35}-{core,coincurve,state-all,state-fast,state-slow,blockchain-fast,blockchain-slow,transactions,vm-all,vm-fast,vm-limits,vm-performance}
+    py{27,34,35}-{core,coincurve,leveldb, state-all,state-fast,state-slow,blockchain-fast,blockchain-slow,transactions,vm-all,vm-fast,vm-limits,vm-performance}
     flake8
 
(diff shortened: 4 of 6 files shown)


evm/db/backends/level.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # Creates db as a class variable to avoid level db lock error
+    def __init__(self):
+        try:
+            import leveldb
+        except ImportError:
+            raise ImportError("LevelDB requires the leveldb \
+                               library which is not available for import.")
+        if not LevelDB.db:
+            LevelDB.db = leveldb.LevelDB('./evm/db/backends/db',
```
pipermerriam: Just a note that this should be configurable at some point but it's not important immediately.

evm/db/backends/level.py
```diff
@@ -0,0 +1,48 @@
...
+                                         create_if_missing=True, error_if_exists=False)
+
+    def get(self, key):
+        return bytes(self.db.Get(key))
+
+    def set(self, key, value):
+        self.db.Put(key, value)
+
+    def exists(self, key):
+        return key in list(self.db.RangeIter(include_value=False))
```
pipermerriam: This strikes me as something that we can't do unless we have a way to lookup key existence in a way that is O(log(n)) or less. This is currently O(n).

evm/db/backends/level.py
```diff
@@ -0,0 +1,48 @@
...
+        self.db.Delete(key)
+
+    #
+    # Snapshot API
+    #
+    def snapshot(self):
+        return Snapshot(self.db.CreateSnapshot())
+
+    def revert(self, snapshot):
+        for item in list(self.db.RangeIter(include_value=False, reverse=True)):
```
pipermerriam: I'm curious why is this being cast to a `list` first?

tests/level-db/test_leveldb_db_backend.py
```diff
@@ -0,0 +1,42 @@
+import os
+import pytest
+from evm.db.backends.memory import MemoryDB
+from evm.db import (
+    get_db_backend,
+)
+
+
+pytest.importorskip('leveldb') 
+memoryDb = MemoryDB()
```
pipermerriam: This looks like it should be a pytest fixture as it will currently share state across all the tests.

```python
@pytest.fixture
def memory_db():
    return MemoryDB()
```

Then in a test:

```python
def test_get(memory_db):
    ...
```

This way you get a fresh memorydb for every test case.

tests/level-db/test_leveldb_db_backend.py
```diff
@@ -0,0 +1,42 @@
...
+
+def test_snapshot_and_revert():
+    snapshot = get_db_backend().snapshot()
+    get_db_backend().set(b'1', b'1')
+    assert get_db_backend().get(b'1')
+    with pytest.raises(KeyError): snapshot.get(b'1')
+    get_db_backend().revert(snapshot)
+    with pytest.raises(KeyError): get_db_backend().get(b'1')
+    # Switches db backend back to local memory for the rest of the tests
+    os.environ['EVM_DB_BACKEND_CLASS'] = 'evm.db.backends.memory.MemoryDB'
```
pipermerriam: Look up the `monkeypatch` fixture provided by pytest.  It has an API for setting environment variables in asuch a way that they are reset after the test is run.

.gitignore
```diff
@@ -41,6 +41,9 @@ output/*/index.html
 docs/_build
 docs/modules.rst
 
+# LevelDB
+evm/db/backends/db
```
pipermerriam: This doesn't seem appropriate as it will ignore anything found in this directory.  Strikes me that this should be under a different directory than `./evm` since that is the root source dir for this codebase.

evm/db/__init__.py
```diff
@@ -17,6 +17,7 @@ def get_db_backend_class(import_path=None):
     return import_string(import_path)
 
 
-def get_db_backend(import_path=None):
+def get_db_backend(import_path=None, db_path=None):
```
pipermerriam: I think this should be `get_db_backend(import_path=None, **init_kwargs)` and then instantiate the backend with `backend_class(**init_kwargs)`.

evm/db/backends/level.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def get(self, key):
+        return bytes(self.db.Get(key))
+
+    def set(self, key, value):
+        self.db.Put(key, value)
+
+    # Returns False instead of KeyError if key doesn't exist
+    def exists(self, key):
+        return bool(self.db.Get(key, False, True, False))
```
pipermerriam: Can you change this function signature to keyword arguments for whatever `False, True, False` are so that those reading this code can get an idea of what those arguments mean?

evm/db/backends/level.py
```diff
@@ -0,0 +1,51 @@
...
+        except ImportError:
+            raise ImportError("LevelDB requires the leveldb \
+                               library which is not available for import.")
+        if not LevelDB.db:
+            self.db_path = db_path
+            LevelDB.db = leveldb.LevelDB(db_path,
+                                         create_if_missing=True, error_if_exists=False)
+
+    def get(self, key):
+        return bytes(self.db.Get(key))
```
pipermerriam: I'm curious to why this is wrapped in bytes?  What type does `Get` return?

evm/db/backends/snapshot.py
```diff
@@ -0,0 +1,9 @@
+class Snapshot(object):
+    def __init__(self, snapshot):
+        self.db = snapshot
+
+    def get(self, key):
+        return self.db.Get(key)
```
pipermerriam: Given the use of `self.db.Get` this object looks specific to the leveldb backend.  If that is the case, I think it should get moved into the same module as the backend class.

tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 envlist=
-    py{27,34,35}-{core,coincurve,state-all,state-fast,state-slow,blockchain-fast,blockchain-slow,transactions,vm-all,vm-fast,vm-limits,vm-performance}
+    py{27,34,35}-{core,coincurve, leveldb, state-all,state-fast,state-slow,blockchain-fast,blockchain-slow,transactions,vm-all,vm-fast,vm-limits,vm-performance}
```
pipermerriam: can you remove the whitespace around `leveldb` to get this inline with the format of the rest of the list.

evm/db/backends/level.py
DavidKnott: `Get` Returns a `bytearray` while the return it should be is straight bytes

.gitignore
DavidKnott: @pipermerriam So you're saying the db files should be stored outside of `./evm` (the only thing which should be in `evm/db/backends/db` should be db files that are constantly being created and deleted in the `__delete__` method of `LevelDB`)?  Should the db files db created in a `db/` folder at the root level of the repo?

pipermerriam: So, `./evm` is the base source path for the `evm` python library.  That path is not an appropriate place for ephemeral files or installation specific files like a leveldb database.

After thinking this through, I don't think that the db path have a global default because there is no global default that is appropriate for all situations.  (see below).  This implies that the value should always be required.

Now there are some cases where default locations make sense.

* For testing:  Use the `tempdir` fixture that comes built in with pytest. See [this code in populus](https://github.com/pipermerriam/populus/blob/master/conftest.py#L34) for an example of how to use this fixture.
* For *production* chains, aka the mainnet, rinkeby, ropsten, use `$HOME/.pyevm/<chain-name>/chain-db`.  (this convention has already been established by the parity client as a location for the chain data.).

Worth pointing out that the go-ethereum client uses a slightly different convention, but I'm not inclined to follow their convention as the parity one seems more intuitive.

evm/db/backends/level.py
pipermerriam: This type of explanation would be a great comment to include above this line so that future developers who read this source can easily understand *why* since there is no clear context without an existing understanding of the LevelDB API.

DavidKnott:
@pipermerriam I'm pretty confused, travis is failing ([here](https://travis-ci.org/pipermerriam/py-evm/jobs/259625308)) but tox is passing locally.  Basically the snapshot is taken before a key was added should throw a `KeyError` and does locally. (Failure to raise happens [here](https://github.com/DavidKnott/py-evm/blob/create-leveldb-backend/tests/level-db/test_leveldb_db_backend.py#L64) )  I think it has something to do with using pytests tempdir (tempdir_factory didn't work either).  Do you have any suggestions as to what might be going on?

evm/db/backends/level.py
```diff
@@ -0,0 +1,64 @@
...
+    # Creates db as a class variable to avoid level db lock error
+    def __init__(self, db_path=None):
+        if not db_path:
+            raise TypeError("Please specifiy a valid path for your database.")
+        try:
+            import leveldb
+        except ImportError:
+            raise ImportError("LevelDB requires the leveldb \
+                               library which is not available for import.")
+        if not LevelDB.db:
```
pipermerriam: How come you're checking the class for the DB path here?  This seems like it should be `self.db`

evm/db/backends/level.py
```diff
@@ -0,0 +1,64 @@
...
+            self.db_path = db_path
+            LevelDB.db = leveldb.LevelDB(db_path,
```
pipermerriam: This will mutate the `LevelDB` class for all future instantiations of the `LevelDB` class which is *bad* as it will bleed state between different db instances.  This should most definitely be using `self.db`.

evm/db/backends/level.py
```diff
@@ -0,0 +1,64 @@
+import shutil
+from .base import (
+    BaseDB,
+)
+
+
+class LevelDB(BaseDB):
+    db = None
+    db_path = None
```
pipermerriam: Is the `db_path` not derivable from the `self.db` property?  If so, I think this should be removed in favor of just pulling this value from the leveldb instance itself.

tests/level-db/test_leveldb_db_backend.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+# Sets db backend to leveldb
+@pytest.fixture
+def config_env(monkeypatch):
+    monkeypatch.setenv('CHAIN_DB_BACKEND_CLASS',
+                       'evm.db.backends.level.LevelDB')
+
+
+@pytest.fixture(scope='session')
```
pipermerriam: The reason your test is failing is a combination from using `LevelDB.db` and scoping this fixture as `"session"`.

So, by using `LevelDB.db`, you're tests are all using the same database instance (meaning that any writes you make to the database will persist between test runs).  

Then, since you are using a *session* level fixture, pytest doesn't remove the database between tests (which it should since you want to test runs to be completely isolated.

So, to fix this, make the suggested changes to the `LevelDB` class code and remove the scoping of this fixture.

I suspect that the reason it fails on travis and not locally is that the tests are not being run in the same order on your local machine and on travis.

tests/level-db/test_leveldb_db_backend.py
```diff
@@ -0,0 +1,67 @@
...
+def temporary_dir(tmpdir_factory):
```
pipermerriam: I'd name this `level_db_path` as it's a slightly more specific name.

PR #47, opened 2017-07-27
Validation Tests

njgheorghita:
### What was wrong?
Not enough tests for `evm/validation.py`

### How was it fixed?
Writing tests

#### Cute Animal Picture

giphy 3


tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,408 @@
...
+    "value,is_valid",
+    (
+        (tuple(), False),
+        ([], False),
+        ({}, False),
+        (set(), False),
+        (1, False),
+        (True, False),
+        ('abc', False),
+        (b'abc', True),
```
pipermerriam: For this to work nicely with both python2 and 3 can you add a `from __future__ import unicode_literals` to the very top of this module?

tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,408 @@
...
+@pytest.mark.parametrize(
+    "value,is_valid",
+    (
+        (tuple(), False),
+        ([], False),
+        ({}, False),
+        (set(), False),
+        (1234, True),
+        ('abc', False),
+        # (True, False),
```
pipermerriam: If this is failing then the validation function should be fixed.  I suspect that the `is_integer` function from `eth_utils` is returning `True` for `is_integer(True)` which is a bug.  If this is the case, can you open up an issue in the [`ethereum-utils`](https://github.com/pipermerriam/ethereum-utils) repository.

tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,408 @@
...
+    "value,is_valid",
+    (
+        # (0, False),
+        (1, False),
+        (8, True),
+        (64, True),
+        (773, False),
+    ),
+)
+def test_validate_multiple_of_8(value, is_valid):
```
pipermerriam: Some quick grepping indicates this validation function is not used anywhere in the codebase.  Do you mind just removing it?

tests/core/validation/test_validation.py
njgheorghita: https://stackoverflow.com/questions/2764017/is-false-0-and-true-1-in-python-an-implementation-detail-or-is-it-guarante 
According to ^ it's because in python3 booleans are a subclass of integers (t == 1, f == 0)
so in `validation.py` `isinstance(value, int)` returns `True` w/ value as a boolean

i fixed it using `if type(value) is int:` in `validation.py` but not sure if that's the best solution?

evm/validation.py
```diff
@@ -16,7 +16,7 @@ def validate_is_bytes(value):
 
 
 def validate_is_integer(value):
-    if not isinstance(value, int):
+    if not type(value) is int:
```
pipermerriam: *mostly a nitpick*

can you change this to `if isinstance(value, int) and not isinstance(value, bool)`?

.gitignore
```diff
@@ -49,3 +49,5 @@ fixtures/**
 
 # profiling
 prof/**
+
+pyevm/
```
pipermerriam: can you explain this addition?

tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,391 @@
...
+    else:
+        with pytest.raises(ValidationError):
+            validate_length_lte(value, maximum_length)
+
+
+@pytest.mark.parametrize(
+    "value,minimum,is_valid",
+    (
+        (1, 2, False),
+        ('a','a', True),
```
pipermerriam: I think this should actually fail.  Python does allow this type of comparison between strings, but I think this function should *require* that it's inputs be numeric using something like the following code.

```python
import numbers

...
if not isinstance(value, Number) or isinstance(value, Bool):
    raise ValidationError("Value must be a non-boolean numeric type: got {0}".format(type(value)))
```

tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,391 @@
...
+    else: 
+        with pytest.raises(ValidationError):
+            validate_gte(value, minimum)
+
+
+@pytest.mark.parametrize(
+    "value,minimum,is_valid",
+    (
+        (1, 2, False),
+        ('a','a', False),
```
pipermerriam: same here with string comparisons.

tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,391 @@
...
+    else: 
+        with pytest.raises(ValidationError):
+            validate_gt(value, minimum)
+
+
+@pytest.mark.parametrize(
+    "value,maximum,is_valid",
+    (
+        (1, 2, True),
+        ('a', 'a', True),
```
pipermerriam: and here

tests/core/validation/test_validation.py
```diff
@@ -0,0 +1,391 @@
...
+    else: 
+        with pytest.raises(ValidationError):
+            validate_lte(value, maximum)
+
+
+@pytest.mark.parametrize(
+    "value,maximum,is_valid",
+    (
+        (1, 2, True),
+        ('a', 'a', False),
```
pipermerriam: And here

.gitignore
njgheorghita: oops, forgot to take that out

evm/validation.py
njgheorghita: `if not isinstance(value, int) or isinstance(value, bool):` seemed to work better

tests/core/validation/test_validation.py
njgheorghita: Rather than import Number, I just used `int`, is that cool? Everything seems to work fine.

njgheorghita:
hmm not quite sure what caused the failure. any thoughts?

evm/validation.py
njgheorghita: `if isinstance(value, int) and not isinstance(value, bool)` would break the core tests

pipermerriam:
Travis failure was my fault.  Master was broken for a few hours today.

Issue #40, opened 2017-07-26
Unit tests for evm/utils/blocks.py functions


gsalgado:
Those functions should be unit tested

DavidKnott:
@gsalgado  I'm on it!

DavidKnott:
@pipermerriam Okay, I will.  Thanks for the advice!

PR #48, opened 2017-07-27
Add unit tests to evm/utils/blocks

DavidKnott:
This PR unit tests `evm/utils/blocks.py` and is created in response to issue #40 

#### Cute Animal Picture
image

### DIFF
--- a/evm/vm/flavors/eip150/opcodes.py
+++ b/evm/vm/flavors/eip150/opcodes.py
@@ -66,6 +66,6 @@ UPDATED_OPCODES = {
 
 
 EIP150_OPCODES = {
-    **copy.deepcopy(HOMESTEAD_OPCODES),
+    **copy.deepcopy(HOMESTEAD_OPCODES),  # noqa: E999
     **UPDATED_OPCODES,
 }
--- a/evm/vm/flavors/frontier/__init__.py
+++ b/evm/vm/flavors/frontier/__init__.py
@@ -269,7 +269,7 @@ def _apply_frontier_create_message(vm, message):
                     contract_code_gas_cost,
                     reason="Write contract code for CREATE",
                 )
-            except OutOfGas as err:
+            except OutOfGas:
                 computation.output = b''
             else:
                 if vm.logger:
--- a/evm/vm/flavors/homestead/opcodes.py
+++ b/evm/vm/flavors/homestead/opcodes.py
@@ -26,7 +26,7 @@ UPDATED_OPCODES = {
 
 
 HOMESTEAD_OPCODES = {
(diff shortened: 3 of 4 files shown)


tests/core/utils/test_block.py
```diff
@@ -0,0 +1,61 @@
...
+from evm.rlp.headers import (
+    BlockHeader,
+)
+
+from evm.utils.db import (
+    make_block_hash_to_score_lookup_key,
+    make_block_number_to_hash_lookup_key,
+)
+
+db = get_db_backend()
```
pipermerriam: Same approach recommended here for these three globals.  I'd make them pytest fixtures.

```python
@pytest.fixture
def db():
    return get_db_backend()

@pytest.fixture
def header():
    return BlockHeader(...)

@pytest.fixture
def block(header, db):
    return FrontierBlock(header, db)

# Then in your tests
def test_something(db, block):
    ...
```

This ensures proper isolation between tests.

tests/core/utils/test_block.py
```diff
@@ -0,0 +1,61 @@
...
+    assert db.exists(block_to_hash_key)
+
+def test_get_score():
+    block_to_hash_key = make_block_hash_to_score_lookup_key(block.hash)
+    score = rlp.decode(db.get(block_to_hash_key), sedes=rlp.sedes.big_endian_int)
+    assert get_score(db, block.hash) == score
+
+def test_get_block_header_by_hash():
+    block_header = get_block_header_by_hash(db, block.hash)
+    assert block_header == header
```
pipermerriam: There is a utility I wrote for this in exact purpose that I recommend using.

It's currently living in `evm.utils.fixture_tests` but really it should be moved to `evm.utils.rlp`.  It's called `assert_rlp_equal` and it both ensures that all fields are equal as well as printing out nice error messages when they are not.

tests/core/utils/test_block.py
```diff
@@ -0,0 +1,61 @@
...
+    assert block_header.difficulty == 1
+    assert block_header.block_number == 1
+    assert block_header.gas_limit == 1
```
pipermerriam: I think if you move to the fixture based tests approach I recommended above that these assertions can be removed as they are encapulated in the previous `assert block_header == header` assertion.

tests/core/utils/test_block.py
```diff
@@ -0,0 +1,61 @@
...
+
+def test_lookup_block_hash():
+    block_hash = lookup_block_hash(db, 1)
+    assert block_hash == block.hash
+
```
pipermerriam: *style nitpick*

Can you remove this extra line.

PR #57, opened 2017-07-28
try to speed up travis runs

pipermerriam:
### What was wrong?

May be able to speed up travis test runs using xdist parallelization.

### How was it fixed?

Added `ADDOPTS="-n 3"` to split tests across 3 processes.  Lets see if it's faster.

#### Cute Animal Picture


pipermerriam:
Wooohooo double the speed!

PR #59, opened 2017-07-30
Unit tests for the Message object.

egong:
### What was wrong?
wrote some unit tests for the message object

### How was it fixed?
not fixed yet, at a stopping point for some feedback

#### Cute Animal Picture
image


tests/core/evm/vm/test_message.py
```diff
@@ -0,0 +1,32 @@
+import pytest
+import evm.vm.message as m
+
+message_object = m.Message(
```
pipermerriam: Typically it's best to avoid globals in test modules (an really in general).  would suggest the following change.

Instead of a global object, define a local "factory" function that you can use to easily generate message objects.  The following is probably a good starting point.

```python
def _create_message(gas=1, gas_price=1, to=b"...", sender=b"...", data=b"", code=b"", origin=b"..."):
    return Message(gas=gas, gas_price=gas_price, ...)
```

For `to`, `sender`, and `origin`, I would recommend using a non-zero, non-empty address such as `\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff`.  This is clearly a test address, but also a valid address.

Next, at the top of this module, you need to add `from __future__ import unicode_literals`.  This forces python2 to behave the same as python3 in terms of the string types it infers.

tests/core/evm/vm/test_message.py
```diff
@@ -0,0 +1,32 @@
...
+    b'',                    #to
+    '11111111111111111111', #sender
+    1,                      #value
+    '',                     #data
+    'asdf',                 #code
+    '22222222222222222222'  #origin
+    )                    
+
+def test_initilizes_properly():
+    # all __init__ parameters are properly validated.
```
pipermerriam: This would really be testing that all parameters are properly *assigned*.  The code style of this repository typically uses docstrings for this type of comment in the following format.

```python
def test_something():
    """
    Ensure that all initialization parameters are correctly assigned
    """
    ...
```

tests/core/evm/vm/test_message.py
```diff
@@ -0,0 +1,32 @@
...
+    assert message_object.gas == 1
+    assert message_object.gas_price == 1
+    assert message_object.to == b''
+    assert message_object.sender == '11111111111111111111'
+    assert message_object.value == 1
+    assert message_object.data == ''
+    assert message_object.code == 'asdf'
+    assert message_object.origin == '22222222222222222222'
+
+def test_is_origin():
```
pipermerriam: This tests the case where it is not the origin, but not the positive case where it is the origin.

There are two ways you could change this.

1. rename this test to `test_message_is_not_origin_when_sender_and_origin_address_are_different`, and then write another new test called `test_message_is_origin_when_sender_and_origin_are_same`.  Note the very verbose test names make it extremely clear what is being tested.
2. Use the `pytest.mark.parametrize` marker to write a generic tests that runs multiple times with different parameters.  See below.

```
@pytest.mark.parametrize(
    "origin,sender,should_be_origin",
    (
        ("0x01", "0x02", False),  # tests that when they are diffrent, it isn't the origin message.
        ("0x01", "0x01", True),  # tests that when they are the same, it is the origin message.
    )
)
def test_computed_is_origin_property(origin, sender, should_be_origin):
    message = _create_message(sender=sender, origin=origin)
    assert message.is_origin is should_be_origin
```

Note that you need to use different values than `"0x01"` and `"0x02"` as those are just meant to be illustrative.

tests/core/evm/vm/test_message.py
```diff
@@ -0,0 +1,32 @@
...
+    #correctly returns True/False for whether this message is the origin message
+    assert message_object.is_origin == False
+
+def origin():
```
pipermerriam: pytest only runs functions whose name begins with `test_`, so this won't be run.

pipermerriam:
Oh, and A+ on the cute animal picture

tests/core/evm/vm/test_message.py
egong: just added a commit 'added create message factory as commented' to address this comment.  Hitting each one individually.

tests/core/evm/vm/test_message.py
egong: Having a little trouble understanding how to properly _validate_ __init__ parameters without redoing the tests in validation.py.  They're automatically validated when the the class is instantiated.

Can you elaborate a little as to the methodology?

tests/core/evm/vm/test_message.py
```diff
@@ -0,0 +1,21 @@
+from __future__ import unicode_literals
+import pytest
+import evm.vm.message as m
+
+
+def _create_message(gas=1, gas_price=1, to=b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff", sender=b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff", value=1, data=b"", code=b"", origin="\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff"):
```
pipermerriam: Can you pull all of these long bytestrings out into top level module constants.

```python
ADDRESS = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xff'
```

Then you can just use them in the signature to make it readable.

```python
def _create_message(gas=1, to=ADDRESS, ...):
    ...
```

tests/core/evm/vm/test_message.py
```diff
@@ -0,0 +1,21 @@
...
+    return m.Message(gas=gas, gas_price=gas_price, to=to, sender=sender, value=value, data=data, code=code, origin=origin)
```
pipermerriam: For readability, this codebase is using the following style when calling or instantiating things with many arguments.  Generally, one per line.

```python
m.Message(
    gas=gas,
    gas_price=gas_price,
    to=to,
    ....
)
```

egong:
@pipermerriam can you please take another look?  the only test I think is incomplete is the first test, how to validate init parameters which I'd like some guidance.  the channelge is how to validate init parameters without redoing the tests in validation.py. They're automatically validated when the the class is instantiated.  thanks!

pipermerriam:
It would be good if you could get the test suite passing first, and then I'll take a look.

egong:
@pipermerriam all the other tests all pass for me.  The specific on I'd like more guidance on is how to test how __init__ _validates_ properly.  Can you share how you'd approach it?

pipermerriam:
Hey @egong 

Thank you very much for getting this started.  There were enough nit-picky things that I wanted changed that I went ahead and pushed them in myself.  You might take a look at the following commit to get a better sense for the code style of this project as well as ways to test the various things at play here.

https://github.com/pipermerriam/py-evm/pull/59/commits/9715a1542a4258d721d4b34a3f5837ca94cfac3f

Thanks again.  Hope to see you around the repository more often.  Let me know if I can help you with any other contributions.

egong:
@pipermerriam i'd like to learn from this experience and continue to contribute, thanks.

PR #61, opened 2017-07-31
Gas meter unit tests

njgheorghita:
### What was wrong?
No unit tests for `evm/vm/gas_meter.py`

### How was it fixed?
Wrote unit tests for GasMeter class

#### Cute Animal Picture
giphy 4


pipermerriam:
Oh, and you have some minor linting errors.  You can find these locally with `tox -e flake8` (you may need to install tox separately as it's not part of `requirements-dev.txt`).

njgheorghita:
Added the test, let me know if you want any other cases added on.

PR #63, opened 2017-08-02
Add a couple missing __init__.py files

gsalgado:
Any ideas how we could change our travis config to detect those?


pipermerriam:
If we start testing against python 2.7 we'll see errors in those modules.

PR #65, opened 2017-08-02
Write tests for stack object

njgheorghita:
### What was wrong?
No unit tests for stack object.

### How was it fixed?
Wrote unit tests for stack object.

#### Cute Animal Picture
giphy 7

### DIFF
--- a/tests/core/stack/test_stack.py
+++ b/tests/core/stack/test_stack.py
@@ -1,0 +1,125 @@
+import pytest
+
+from evm.vm.stack import (
+    Stack,
+)
+from evm.exceptions import (
+    FullStack,
+    InsufficientStack,
+    ValidationError,
+)
+from evm.constants import (
+    UINT256,
+    BYTES,
+    SECPK1_N,
+)
+
+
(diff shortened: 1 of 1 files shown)


tests/core/stack/test_stack.py
```diff
@@ -0,0 +1,125 @@
...
+    with pytest.raises(FullStack):
+        stack.dup(1)
+
+
+@pytest.mark.parametrize(
+    ("items,type_hint"),
+    (
+        ([1], UINT256),
+        ([1, 2, 3], UINT256),
+        ([b'1', b'10', b'101', b'1010'], BYTES)
```
pipermerriam: It would be good to have at least one of these for which the `type_hint` was different than the value being pushed onto the stack.

If it isn't clear, the `type_hint` tells the `stack` object what type it should return, so a good test would be pushing a `bytes` object onto the stack and pulling it back of as an `int`.

pipermerriam: Nevermind, I see you already wrote that test below.

PR #68, opened 2017-08-03
Write unit tests for memory object

njgheorghita:
### What was wrong?
No unit tests for memory object.

### How was it fixed?
Wrote unit tests for memory object.

#### Cute Animal Picture
giphy 9

### DIFF
--- a/tests/core/memory/test_memory.py
+++ b/tests/core/memory/test_memory.py
@@ -1,0 +1,73 @@
+import pytest
+
+from evm.vm.memory import (
+    Memory,
+)
+from evm.exceptions import (
+    ValidationError,
+)
+
+
+@pytest.fixture
+def memory():
+    return Memory()
+
+
+@pytest.fixture
+def memory32():
+    memory = Memory()
+    memory.extend(0, 32)
+    return memory
+
+
+def test_write(memory32):
+    # Test that write creates 32byte string == value padded with zeros
+    memory32.write(start_position=0, size=4, value=b'1010')
+    assert memory32.bytes == b'1010' + bytearray(28)
+
+
+@pytest.mark.parametrize("start_position", (-1, 2**256, 'a', b'1010'))
+def test_write_rejects_invalid_start_position(memory32, start_position):
+    with pytest.raises(ValidationError):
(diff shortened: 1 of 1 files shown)


njgheorghita:
@pipermerriam @gsalgado are these closer to the unit testing style you're looking for?

tests/core/memory/test_memory.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.fixture
+def memory32():
+    memory = Memory()
+    memory.extend(0, 32)
+    return memory
+
+
+@pytest.mark.parametrize("start_position,size,value", ((0, 4, b'1010'),))
```
pipermerriam: Not quite.  Parametrization with a single value strikes me as awkward since the idea behind parametrization is running the same test multiple times with different parameters.  In this case, if there is  only a single parameter, I think it's best just to have those values be part of the test.

tests/core/memory/test_memory.py
```diff
@@ -0,0 +1,75 @@
...
+        memory32.write(start_position, size, value)
+
+
+def test_extend_appropriately_extends_memory(memory):
+    memory.extend(0, 32)
+    assert memory.bytes == bytearray(32)
+    memory.extend(30, 32)
+    assert memory.bytes == bytearray(64)
+    memory.extend(48, 10)
+    assert memory.bytes == bytearray(64)
```
pipermerriam: This test could use some comments to explain what exactly is being tested in human readable terms.

tests/core/memory/test_memory.py
```diff
@@ -0,0 +1,75 @@
...
+def memory32():
+    memory = Memory()
+    memory.extend(0, 32)
+    return memory
+
+
+@pytest.mark.parametrize("start_position,size,value", ((0, 4, b'1010'),))
+def test_write(memory32, start_position, size, value):
+    memory32.write(start_position, size, value)
+    assert memory32.bytes == value + bytearray(32 - size)
```
pipermerriam: This is a confusing assertion for me.  It's not easy to understand what exactly is being tested.

Issue #54, opened 2017-07-28
Write unit test for the `Computation` object


pipermerriam:
### What is wrong?

There are no unit test for the `evm.vm.computation.Computation` object

### How can it be fixed

Write unit tests to verify that it:

- Basic smoke tests for `prepare_child_message`.  Check depth is increased, origin is preserved, etc.
- `expand_memory` extension correctly validates parameters and charges appropriate gas costs.
- `c.output` property only returns a value if the computation didn't error.
- `register_account_for_deletion` fails if an account is registered twice.
- `get_accounts_for_deletion()` returns empty if the computation errors.
- `add_log_entry` performs validation and adds the log entry.
- `get_log_entries` returns empty when computation has errored.
- `get_gas_refund` returns `0` when computation has errored.
- `get_gas_remaining` returns `0` when computation has errored.
- Use as a context manager appropriately catches and consumes VM errors raised inside of the context, resulting in an errored computation object.
- `get_gas_used` returns max gas when computation has errored.

DavidKnott:
@pipermerriam I'm on this!

PR #69, opened 2017-08-03
Create unit tests for Computation 

DavidKnott:
Creates unit tests for `Computation` class in response to #54 

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/vm/test_computation.py
+++ b/tests/core/vm/test_computation.py
@@ -1,0 +1,218 @@
+import pytest
+
+from evm.vm import (
+    Computation,
+    Message,
+)
+
+from eth_utils import (
+    to_canonical_address,
+)
+
+from evm.exceptions import (
+    ValidationError,
+)
+
+NORMALIZED_ADDRESS_A = "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6"
+NORMALIZED_ADDRESS_B = "0xcd1722f3947def4cf144679da39c4c32bdc35681"
+CANONICAL_ADDRESS_A = to_canonical_address("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6")
+CANONICAL_ADDRESS_B = to_canonical_address("0xcd1722f3947def4cf144679da39c4c32bdc35681")
+
+
+@pytest.fixture
+def message():
+    message = Message(
+        origin=CANONICAL_ADDRESS_B,
+        to=CANONICAL_ADDRESS_A,
+        sender=CANONICAL_ADDRESS_B,
+        value=100,
+        data=b'',
+        code=b'',
+        gas=100,
+        gas_price=1,
+    )
+    return message
+
+
+@pytest.fixture
+def computation(message):
+    computation = Computation(
+        vm=None,
+        message=message
+    )
+    return computation
+
+
+@pytest.fixture
+def create_error():
+    def computation_error(computation):
+        with computation:
+            computation.gas_meter.consume_gas(computation.get_gas_remaining() + 1,
+                                              'Create an out of gas error for testing')
+        return computation
+    return computation_error
+
+
+def test_prepare_child_message(computation):
+    child_message = computation.prepare_child_message(
+        gas=100,
+        to=CANONICAL_ADDRESS_B,
+        value=200,
+        data=b'',
+        code=b''
+    )
+    assert computation.msg.depth == 0
+    assert child_message.depth == 1
+    assert computation.msg.origin == child_message.origin
+    assert computation.msg.gas_price == child_message.gas_price
+    assert computation.msg.storage_address == child_message.sender
+
+
+def test_extend_memory_start_position_must_be_uint256(computation):
+    with pytest.raises(ValidationError):
+        computation.extend_memory("1", 1)
+    with pytest.raises(ValidationError):
+        computation.extend_memory(-1, 1)
+
+
+def test_extend_memory_size_must_be_uint256(computation):
+    with pytest.raises(ValidationError):
+        computation.extend_memory(1, "1")
+    with pytest.raises(ValidationError):
+        computation.extend_memory(1, -1)
+
+
+def test_extend_memory_stays_the_same_if_size_is_0(computation):
+    assert len(computation.memory.bytes) == 0
+    computation.extend_memory(1, 0)
+    assert len(computation.memory.bytes) == 0
+
+
+def test_extend_memory_increases_memory_by_32(computation):
+    assert computation.gas_meter.gas_remaining == 100
+    computation.extend_memory(0, 1)
+    assert len(computation.memory.bytes) == 32
+    # 32 bytes of memory cost 3 gas
+    assert computation.gas_meter.gas_remaining == 97
+
+
+def test_extend_memory_doesnt_increase_until_32_bytes_are_used(computation):
+    computation.extend_memory(0, 1)
+    computation.extend_memory(1, 1)
+    assert len(computation.memory.bytes) == 32
+    computation.extend_memory(2, 32)
+    assert len(computation.memory.bytes) == 64
+    assert computation.gas_meter.gas_remaining == 94
+
+
+def test_register_accounts_for_deletion_raises_if_address_isnt_canonical(computation):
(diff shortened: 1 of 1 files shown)


tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+        gas=100,
+        gas_price=1,
+    )
+    return message
+
+
+@pytest.fixture
+def computation(message):
+    computation = Computation(
+        vm=1,
```
pipermerriam: Just a note-to-self:

Giving the computation full access to the VM is probably an antipattern as computation should really only require very finite access to a finite number of APIs.  It would be very nice if we could get the `vm` object out of the computation in a clean way.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+        message=message
+    )
+    return computation
+
+
+@pytest.fixture
+def create_error():
+    def computation_error(computation):
+        computation.gas_meter.gas_remaining = 0
```
pipermerriam: I don't think this is a good approach.  Things I dislike are that 1) it directly mutates the `gas_meter` to produce an OOG error and 2) the error is not *authentic*.

It's pretty trivial to generate a VM error using some simple invalid bytecode.  The bytecode `b'\x50` which is a single `POP` instruction should throw an error as long as the stack is empty.  I'm not exactly sure how the best way to apply this to these tests is yet but leaving this here to at least surface the problem.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+        return computation
+    return computation_error
+
+
+def test_prepare_child_message(computation):
+    child_message = computation.prepare_child_message(
+        gas=100,
+        to=to_canonical_address(CANONICAL_ADDRESS_2),
+        value=200,
+        data=b"0x",
```
pipermerriam: Is the `0x` here instentional or are you trying to create an empty string?  If not, this should just be an empty string.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+    assert computation.msg.origin == child_message.origin
+    assert computation.msg.gas_price == child_message.gas_price
+    assert computation.msg.storage_address == child_message.sender
+
+
+def test_extend_memory(computation):
+    with pytest.raises(ValidationError):
+        # Start positions needs to be a uint256
+        computation.extend_memory("1", 1)
+        computation.extend_memory(-1, 1)
```
pipermerriam: This line (and all the other lines in this block) won't get executed as the first line throws the `ValidationError` and then exits this context.  You need individual `with` statements for each error you want to test.

Also, I prefer to avoid long tests like this which test many different things as they are prone to these types of errors.  Can you see about splitting this up into many smaller tests?

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+    # Log entries starts empty
+    assert computation.log_entries == []
+    with pytest.raises(ValidationError):
+        # Fails if address isn't canonical
+        computation.add_log_entry("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6", [1, 2, 3], b'1')
+        # Fails if topic elements aren't uint256
+        computation.add_log_entry(CANONICAL_ADDRESS_1, [-1, 2, 3], b'1')
+        computation.add_log_entry(CANONICAL_ADDRESS_1, ['1', 2, 3], b'1')
+        # Fails if data isn't bytes
+        computation.add_log_entry(CANONICAL_ADDRESS_1, [1, 2, 3], 1)
```
pipermerriam: It would be good if this tried to add a non-bytes type string as well to see that it rejects string types that are not `bytes`.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def test_add_log_entry(computation):
+    # Log entries starts empty
+    assert computation.log_entries == []
+    with pytest.raises(ValidationError):
+        # Fails if address isn't canonical
+        computation.add_log_entry("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6", [1, 2, 3], b'1')
+        # Fails if topic elements aren't uint256
+        computation.add_log_entry(CANONICAL_ADDRESS_1, [-1, 2, 3], b'1')
```
pipermerriam: Same problem here.  This statement isn't being run and needs it's own `with` context.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+
+def test_get_gas_remaining(computation, create_error):
+    assert computation.get_gas_remaining() == 100
+    # Trigger an out of gas error causing get gas remaining to be 0
+    computation = create_error(computation)
+    assert computation.get_gas_remaining() == 0
+
+
+def test_get_gas_used(computation, create_error):
+    computation.msg.gas = 150
```
pipermerriam: This type of mutation is error prone.  It would be much better to have the computation be create with this value.  One way to do this would be to stop using a static `computation` fixture and instead use a *factory*.

```python
def _computation_factory(....<put-some-sane-defaults-here>):
    return Computation(<boiler-plate-goes-here>)
```

Then you can easily construct a computation object within your test with custom parameters.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+    assert len(computation.memory.bytes) == 32
+    computation.extend_memory(2, 32)
+    assert len(computation.memory.bytes) == 64
+    assert computation.gas_meter.gas_remaining == 94
+
+
+def test_register_accounts_for_deletion(computation):
+    # Fails if address isn't canonical
+    with pytest.raises(ValidationError):
+        computation.register_account_for_deletion("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6")
```
pipermerriam: Rather than using `"0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6"` here, it seems like this value should be assigned as `NORMALIZED_ADDRESS_1` so that it's clear that it is related to `CANONICAL_ADDRESS_1`.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,184 @@
...
+
+from eth_utils import (
+    to_canonical_address,
+)
+
+from evm.exceptions import (
+    ValidationError,
+)
+
+CANONICAL_ADDRESS_1 = to_canonical_address("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6")
```
pipermerriam: This is a somewhat nit-picky naming convention thing, but, I've traditionally used the convention `thing_a` and `thing_b` over `thing_1` and `thing_2`. The reasoning is related to a bit of ocd with 0-indexed values and my brain thinking there should also be a `thing_0` as well as feeling like numeric ordering is much more heavily categorized as sequential (1 comes before 2), than alpha ordering.  With letters, there is less of a tendency for me at least to mentally line them up in a sequence.  They are just a set of labeled *things* without strict ordering.  Not sure if that makes sense and you don't need to update this if this approach doesn't hold water for you.

tests/core/vm/test_computation.py
DavidKnott: @pipermerriam Would it be okay if I made an error with `computation.extend_memory(32, 100000)`?  (causing it to actually run out of gas)

tests/core/vm/test_computation.py
DavidKnott: That makes a lot of sense!  Between 0 indexing and the tendency to see numeric ordering as sequential, I see why alpa labeling makes sense.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,217 @@
...
+        message=message
+    )
+    return computation
+
+
+@pytest.fixture
+def create_error():
+    def computation_error(computation):
+        with computation:
+            computation.extend_memory(32, 1000)
```
pipermerriam: I'm realizing that the simplest solution is to just replace this with `computation.gas_meter.consume_gas(computation.gas_meter.gas_remaining + 1)`.  Must less indirect than doing it by-proxy via memory extension.

PR #70, opened 2017-08-04
Write code-stream object unit tests

njgheorghita:
### What was wrong?
No unit tests for CodeStream object.

### How was it fixed?
Wrote unit tests for CodeStream object.

#### Cute Animal Picture
giphy 12

### DIFF
--- a/tests/core/code-stream/test_code_stream.py
+++ b/tests/core/code-stream/test_code_stream.py
@@ -1,0 +1,62 @@
+import pytest
+
+from evm import opcode_values
+from evm.vm.code_stream import (
+    CodeStream,
+)
+from evm.exceptions import (
+    ValidationError,
+)
+
+
+def test_code_stream_accepts_bytes():
+    code_stream = CodeStream(b'\x01')
+    assert len(code_stream.stream.getvalue()) == 1
+
+
+@pytest.mark.parametrize("code_bytes", (1010, '1010', True, bytearray(32)))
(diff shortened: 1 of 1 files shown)


pipermerriam:
These are some sexy tests.

Side note:  I haven't taken the time to evaluate whether my approach for checking if opcodes are valid is actually *fast*.  It tries to be efficient by *cleverly* working its way backwards through the bytecode but this may actually be slower than just processing the whole of the bytecode from start to finish.  If anyone is inclined to profile this and do that analysis, it would be valuable because that code will get called for *EVERY* `JUMPDEST` which is **a lot** of times.

njgheorghita:
giphy 13

 I'll give it a shot. Is there a min/max length for the bytecode strings that get evaluated?

pipermerriam:
> I'll give it a shot. 

Just to be clear, should be done in a different pull request.

> Is there a min/max length (32?) for the bytecode strings that get evaluated?

No, bytecode can be arbitrarily long.  The *worst* case performance scenario is really what we care about most since it's something someone can intentionally exploit.  This would look like *long* bytecode with a `JUMPDEST` very near the end.  This `JUMPDEST` would have a `PUSH32` somewhere in the 32 bytes before it.  Then there would be another `PUSH32` somewhere in the 32 bytes before that, and so on, until the very beginning of the bytecode.  The current implementation would have to trace it's way all the way to the beginning to find out whether the original `JUMPDEST` was valid.

PR #72, opened 2017-08-06
Padding unit test

egong:
### What was wrong?
padding.py needed unit tests

### How was it fixed?
Added tests to padding.py

#### Cute Animal Picture
image

### DIFF
--- a/tests/core/padding-utils/test_padding.py
+++ b/tests/core/padding-utils/test_padding.py
@@ -1,0 +1,58 @@
+import pytest
+
+from evm.utils.padding import (
+    pad_left,
+    pad_right,
+    pad32,
+    pad32r
+)
+
+padding_byte = b"\x00"
+
+
+@pytest.mark.parametrize(
+    "value, to_size, pad_with, expected",
+    (
+        (b"", 0, b"\x00", b""),
+        (b"", 1, b"\x00", b"\x00"),
+        (b"\x00", 1, b"\x01", b"\x00"),
+        (b"\x00", 2, b"\x01", b"\x01\x00")
+    )
+)
+def test_pad_left(value, to_size, pad_with, expected):
+    assert pad_left(value, to_size, pad_with) == expected
+
+
+@pytest.mark.parametrize(
+    "value, to_size, pad_with, expected",
+    (
+        (b"", 0, b"\x00", b""),
+        (b"", 1, b"\x00", b"\x00"),
+        (b"\x00", 1, b"\x01", b"\x00"),
+        (b"\x00", 2, b"\x01", b"\x00\x01")
+    )
+)
+def test_pad_right(value, to_size, pad_with, expected):
+    assert pad_right(value, to_size, pad_with) == expected
+
+
+@pytest.mark.parametrize(
+    "value, expected",
+    (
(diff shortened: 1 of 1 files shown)


egong:
@pipermerriam I got the build passing so please give your feedback on the tests.  However, I did run into two env issues which I'd like some feedback.

1- I had to change my tox.ini file (which I didn't check into git).  I changed basepython = py35: python 3.5 to python3.  Otherwise I get the interpreter error.

2- Virtualenv created a whole bunch of files that I'm attaching in a pic.  Do I gitignore these files or just never check them in?  I'm looking for the best practice here.  Thanks.
screen shot 2017-08-10 at 10 30 57 am

tests/core/padding-utils/test_padding.py
```diff
@@ -0,0 +1,34 @@
+from __future__ import unicode_literals
```
gsalgado: Why is this needed?

tests/core/padding-utils/test_padding.py
```diff
@@ -0,0 +1,34 @@
...
+    "value, to_size, pad_with, expected",
+    (
+        (b"", 0, b"\x00", b""),
+        (b"", 1, b"\x00", b"\x00"),
+        (b"\x00", 1, b"\x01", b"\x00"),
+        (b"\x00", 2, b"\x01", b"\x00\x01")
+    )
+)
+def test_pad_right(value, to_size, pad_with, expected):
+    assert pad_right(value, to_size, pad_with) == expected
```
gsalgado: Maybe add tests for pad32 and pad32r as well?

tests/core/padding-utils/test_padding.py
egong: Assumed it was for python 2/3 compatibility because it's imported in various places across the codebase.  Should I take it out?

gsalgado: I hadn't noticed it in other files before, so can't tell for sure. If the travis checks pass without it I'd say remove it

egong:
@gsalgado I've made the requested changes.  What do you think?

tests/core/padding-utils/test_padding.py
```diff
@@ -0,0 +1,59 @@
...
+
+from evm.utils.padding import (
+    pad_left,
+    pad_right,
+    pad32,
+    pad32r
+)
+
+test_31_length_byte_string = b"\x00" * 31
+test_32_length_byte_string = b"\x00" * 32
```
gsalgado: These variable names are not great as they don't really make it clear they're storing the padding byte repeated N times. Also, it feels a bit weird to have variables just for two seemingly arbitrary lengths: 31 and 32. Maybe instead you could define a padding_byte variable with the _b"\x00"_ value and then in tests use that with the length multiplier you want

gsalgado:
LGTM. Can you squash the commits before I merge?

egong:
@gsalgado Squashed the commits

gsalgado:
@egong care to reword that commit message? no need to mention the squashing there. https://chris.beams.io/posts/git-commit/ has some good tips on how to write good commit messages, but no need to follow them to the letter

PR #75, opened 2017-08-10
Update is_valid_opcode security

njgheorghita:
### What was wrong?
Codestream processing was insecure

### How was it fixed?
Updated codestream processing algorithm

#### Cute Animal Picture
giphy 15

### DIFF
--- a/evm/vm/code_stream.py
+++ b/evm/vm/code_stream.py
@@ -10,13 +10,15 @@ from evm.validation import (
 
 class CodeStream(object):
     stream = None
+    deepest = None
 
     logger = logging.getLogger('evm.vm.CodeStream')
 
     def __init__(self, code_bytes):
         validate_is_bytes(code_bytes)
         self.stream = io.BytesIO(code_bytes)
-        self._validity_cache = {}
+        self._validity_cache = set(range(0))
+        self.deepest = 0
 
     def read(self, size):
         return self.stream.read(size)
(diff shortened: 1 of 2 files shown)


njgheorghita:
Hey @pipermerriam still plenty of cleaning and refactoring to be done, but if you don't mind giving this a quick lookover to make sure i'm on the right track/didn't miss anything important, that would be awesome!

evm/vm/code_stream.py
```diff
@@ -10,13 +10,15 @@
...
     stream = None
+    deepest = None
 
     logger = logging.getLogger('evm.vm.CodeStream')
 
     def __init__(self, code_bytes):
         validate_is_bytes(code_bytes)
         self.stream = io.BytesIO(code_bytes)
-        self._validity_cache = {}
+        self._validity_cache = set(range(0))
```
pipermerriam: This should probably be changed to just `set()` as I think the `range(0)` part is not necessary.  And can we now rename this to `invalid_positions` since it's no longer really a cache.

evm/vm/code_stream.py
```diff
@@ -10,13 +10,15 @@
...
+        self.deepest = 0
```
pipermerriam: name nitpick:

Thoughts on renaming to `depth_processed` to be a bit more explicit?

evm/vm/code_stream.py
```diff
@@ -66,28 +68,35 @@ def seek(self, pc):
...
+            return False
+        # return true if position not in val_cache but has been parsed
+        if position <= self.deepest:
+            return True
+        else:
+            # set counter to deepest parsed position
+            i = self.deepest
+            while i <= position:
+                # get opcode
+                with self.seek(i):
```
pipermerriam: It'd be nice to do this without seeking into it.  Can we just add an `__getitem__` to this object so you can just do `self[i]` to grab the value at the given length?

evm/vm/code_stream.py
```diff
@@ -66,28 +68,35 @@ def seek(self, pc):
...
+                    opcode = self.next()
+                # if opcode = pushxx
+                if opcode >= opcode_values.PUSH1 and opcode <= opcode_values.PUSH32:
+                    # add range(xx) to val_cache
+                    self._validity_cache.update(range((i + 1), ((i + 1) + (opcode - 95))))
```
pipermerriam: For lines of code like this I've found that the following pattern makes a lot better readability

```
left_bound = ....
right_bound = ....
invalid_range = range(left_bound, right_bound)
self._validity_cache.update(invalid_range)
```
That way each line only contains a single *concept* which makes it easier to ingest one piece at a time.

Issue #67, opened 2017-08-03
VM.block and VM.state_db can easily get out of sync


gsalgado:
### What is wrong?

Having _VM.state_db_ and _VM.block_ is error prone as we need to ensure they're in sync, which is done by assigning _VM.state_db.root_hash_ to _VM.block.header.state_root_ whenever we update the state.

### How can it be fixed

We could have the state DB as a property on the Block class, with all state-related functionality wrapped in Block methods, thus completely abstracting the State DB (and the need to keep it in sync with the block) from external code

pipermerriam:
Looks like my email reply didn't make it in.

So one of my pet-peeves of the pyethereum codebase was that the `block` object had a ton of business logic baked into it.  I'd like to avoid that if at all possible and try to keep the RLP objects well isolated.

This gets me thinking of alternate ways to solve the same problem and it strikes me as similar to the problem of forgetting to close a file or database connection after opening it.  This gets me thinking about whether there is a context manager style API that would provide an elegant way to solve this so that we not longer have to remember to update the `header.state_root`.

Looking at where `vm.state_db` is accessed yields this:

```
evm/vm/computation.py:        return self.vm.state_db
evm/vm/flavors/frontier/__init__.py:    sender_balance = vm.state_db.get_balance(transaction.sender)
evm/vm/flavors/frontier/__init__.py:    vm.state_db.set_balance(transaction.sender, sender_balance - gas_cost)
evm/vm/flavors/frontier/__init__.py:    vm.state_db.increment_nonce(transaction.sender)
evm/vm/flavors/frontier/__init__.py:            vm.state_db.get_nonce(transaction.sender) - 1,
evm/vm/flavors/frontier/__init__.py:        code = vm.state_db.get_code(transaction.to)
evm/vm/flavors/frontier/__init__.py:        coinbase_balance = vm.state_db.get_balance(vm.block.header.coinbase)
evm/vm/flavors/frontier/__init__.py:        vm.state_db.set_balance(
evm/vm/flavors/frontier/__init__.py:            sender_balance = vm.state_db.get_balance(message.sender)
evm/vm/flavors/frontier/__init__.py:            vm.state_db.set_balance(message.sender, sender_balance + gas_refund_amount)
evm/vm/flavors/frontier/__init__.py:        coinbase_balance = vm.state_db.get_balance(vm.block.header.coinbase)
evm/vm/flavors/frontier/__init__.py:        vm.state_db.set_balance(
[...]

PR #76, opened 2017-08-11
VM: state_db is accessible only via a contextmanager now

gsalgado:
Issue #67

The state db is no longer stored as an instance variable anywehre; instead it is always accessed
via the VM.state_db() contextmanager, which takes care of updating the header's state_root on
exit.

One important thing to note is the fact that now the header's state_root is updated multiple times
during, say, the processing of a transaction (because there are multiple methods called to perform
that, and several of them make use of the state_db contextmanager), whereas before it'd only be
updated once, at the very end of the tx processing. Because of that, if we now have an uncaught
exception while adding a transaction to a block, we may end up with the wrong state_root, but
an uncaught exception there probably means we have a bug that prevents us from processing the
block anyway, so maybe ending up with the wrong state_root in this case is not an actual issue.

### DIFF
--- a/evm/chain.py
+++ b/evm/chain.py
@@ -96,12 +96,6 @@ class Chain(object):
     #
     # Convenience and Helpers
     #
-    def get_state_db(self):
-        """
-        Passthrough helper to the current VM class.
-        """
-        return self.get_vm().state_db
-
     def get_block(self):
         """
         Passthrough helper to the current VM class.
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import
 
+from contextlib import contextmanager
 import logging
 
 from evm.constants import (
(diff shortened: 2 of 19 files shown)


evm/chain.py
```diff
@@ -249,6 +249,27 @@ def import_block(self, block):
 
         return imported_block
 
+    def ensure_blocks_are_equal(self, block1, block2):
```
pipermerriam: This seems like it should be a stand-alone function, maybe renamed to `validate_block_equality`.  Doesn't appear to access `self` anywhere in the function body and it should easier to test in isolation that way.

evm/vm/base.py
```diff
@@ -66,19 +57,17 @@ def configure(cls,
...
-    @block.setter
-    def block(self, value):
-        self._block = value
-        self.state_db = State(db=self.db, root_hash=value.header.state_root)
+    @contextmanager
+    def state_db(self, read_only=False):
+        state = State(db=self.db, root_hash=self.block.header.state_root)
+        yield state
+        if read_only:
+            # TODO: This is a bit of a hack; ideally we should raise an error whenever the
```
pipermerriam: I think I like the idea of implementing this as a wrapper around the `self.db` which raises errors on any write operations.  Might be a good *easy pickings* ticket to open.

evm/vm/flavors/frontier/blocks.py
```diff
@@ -307,6 +306,9 @@ def add_uncle(self, uncle):
         self.header.uncles_hash = keccak(rlp.encode(self.uncles))
         return self
 
+    # TODO: Check with Piper what's the use case for having the mine() method allowing callsites
```
pipermerriam: I believe this functionality can be removed.  This was from an earlier architecture when I hadn't quite wrapped my head around what the block creation workflow was going to be.

evm/chain.py
gsalgado: The main reason why I kept it as a Chain method is so that we can easily [replace it in tests via the new Chain.configure()](https://github.com/pipermerriam/py-evm/pull/76/files#diff-e29dd936977fd333e7e2c113356591d8R90), but I guess with pytest's monkeypatch fixture we could achieve the same without having to worry about reverting after the test is done. I'm happy to give that a go if you think it's worth turning it into a standalone function

evm/vm/flavors/frontier/blocks.py
gsalgado: Cool, I'll submit a separate PR removing that

evm/vm/base.py
gsalgado: #80

PR #77, opened 2017-08-13
update coincurve

ofek:
https://github.com/ofek/coincurve#changelog

egleimv


ofek:
@pipermerriam Thanks! By the way, how is Coincurve working for your lib thus far?

PR #81, opened 2017-08-23
Fix encoding of ECDSA signatures and public keys

gsalgado:
We were using the Bitcoin format when encoding signatures, with the recovery id (v) at the first byte.
https://github.com/ethereum/go-ethereum/blob/master/crypto/signature_nocgo.go#L55

Also change decode_public_key() to accept only 64-bytes long public keys and encode_raw_public_key() to generate the 64-bytes version as well (without the \x04 prefix) as that's what we must use when generating addresses.

At this point we don't have many places using the non-raw ECCBackend APIs, so it's probably ok to leave it as is but at some point I think it'd be nice to create custom types for public/private keys and signatures, with better docs and validation checks so that there's less room for mistakes and we get meaningful errors when things go wrong


evm/utils/address.py
```diff
@@ -27,4 +27,8 @@ def private_key_to_address(private_key):
 
 
 def public_key_to_address(public_key):
-    return keccak(public_key[1:])[-20:]
+    if len(public_key) != 64:
+        raise ValueError(
```
pipermerriam: This strikes me as a good re-usable validation function since I bet we run into many more situations where we want to validate public keys.

I see this same code repeated in `evm/utils/secp256k1.py`.

tests/ecc/test_backends.py
```diff
@@ -0,0 +1,61 @@
...
+    MSGHASH,
+    SECRETS,
+)
+
+
+backends = [PurePythonECCBackend()]
+try:
+    backends.append(CoinCurveECCBackend())
+except ImportError:
+    pass
```
pipermerriam: This worries me as it could lead to a misconfiguration in CI leaving the coincurve backend untested because it simply gets skipped.

In the same vein, I dislike having `coincurve` be a development dependency because I want to be able to test the library **without** it installed to know that it isn't required.  Not sure what the best approach here is.

PR #83, opened 2017-08-28
add lru caching to validation methods

pipermerriam:
### What was wrong?

Potential opportunity for speeding some things up by memoizing the validation functions.

### How was it fixed?

Used the `cytoolz.functoolz.memoize` function in conjunction with the `pylru.lrucache` object to cache the most common validations that get passed through those functions.

#### Cute Animal Picture

pink-orchid-mantis

### DIFF
--- a/evm/validation.py
+++ b/evm/validation.py
@@ -1,11 +1,14 @@
 import functools
 
+import pylru
+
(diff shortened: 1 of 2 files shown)


pipermerriam:
Not as simple as I'd hoped. abort!

PR #85, opened 2017-08-28
Piper/p2p

pipermerriam:
Just looking at the diff for the networking code.

### DIFF
--- a/evm/p2p/discovery.py
+++ b/evm/p2p/discovery.py
@@ -1,0 +1,583 @@
+import asyncio
+import ipaddress
+import struct
+import time
+
+import rlp
+from rlp.utils import decode_hex, encode_hex, is_integer, str_to_bytes, safe_ord
+
+from repoze.lru import LRUCache
+
+from evm.p2p import kademlia
+from evm.p2p.p2p import host_port_pubkey_to_uri, host_port_pubkey_from_uri
+from evm.p2p.upnp import add_portmap, remove_portmap
+from evm.utils.secp256k1 import private_key_to_public_key
+from evm.utils.keccak import keccak
+from evm.utils.numeric import (
+    big_endian_to_int,
+    int_to_big_endian,
+)
+
+# XXX: Most of this code was lifted from pydevp2p
+
+
+class DefectiveMessage(Exception):
+    pass
+
+
+class WrongMAC(DefectiveMessage):
+    pass
+
+
+class PacketExpired(DefectiveMessage):
+    pass
+
+
+def int_to_big_endian4(integer):
+    ''' 4 bytes big endian integer'''
+    return struct.pack('>I', integer)
+
+
+def enc_port(p):
+    return int_to_big_endian4(p)[-2:]
+
+
+class Address(object):
+
+    def __init__(self, ip, udp_port, tcp_port=0, from_binary=False):
+        tcp_port = tcp_port or udp_port
+        if from_binary:
+            self.udp_port = big_endian_to_int(udp_port)
+            self.tcp_port = big_endian_to_int(tcp_port)
+        else:
+            assert is_integer(udp_port)
+            assert is_integer(tcp_port)
+            self.udp_port = udp_port
+            self.tcp_port = tcp_port
+        try:
+            self._ip = ipaddress.ip_address(ip)
+        except ipaddress.AddressValueError as e:
+            log.debug("failed to parse ip", error=e, ip=ip)
+            raise e
+
+    @property
+    def ip(self):
+        return str(self._ip)
+
+    def update(self, addr):
+        if not self.tcp_port:
+            self.tcp_port = addr.tcp_port
+
+    def __eq__(self, other):
+        # addresses equal if they share ip and udp_port
+        return (self.ip, self.udp_port) == (other.ip, other.udp_port)
+
+    def __repr__(self):
+        return 'Address(%s:%s)' % (self.ip, self.udp_port)
+
+    def to_dict(self):
+        return dict(ip=self.ip, udp_port=self.udp_port, tcp_port=self.tcp_port)
+
+    def to_binary(self):
+        """
+        struct Endpoint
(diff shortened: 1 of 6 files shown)


evm/p2p/kademlia.py
```diff
@@ -0,0 +1,446 @@
...
+
+def random_nodeid():
+    return random.randint(0, k_max_node_id)
+
+
+@total_ordering
+class Node(object):
+
+    def __init__(self, pubkey):
+        self.pubkey = pubkey
```
pipermerriam: This seems like it should get validated to ensure it's in the proper format.

evm/p2p/kademlia.py
```diff
@@ -0,0 +1,446 @@
...
+        if k_id_size == 512:
+            self.id = big_endian_to_int(pubkey)
+        else:
+            assert k_id_size == 256
```
pipermerriam: Assert statements can easily be disabled: https://stackoverflow.com/questions/1273211/disable-assertions-in-python

I've preferred explicit raising of a `ValueError` or `TypeError` instead.

evm/p2p/kademlia.py
```diff
@@ -0,0 +1,446 @@
...
+        """Least recently seen"""
+        return self.nodes[0]
+
+    @property
+    def depth(self):
+        """Depth is the prefix shared by all nodes in the bucket.
+
+        i.e. The number of shared leading bits.
+        """
+        def to_binary(x):  # left padded bit representation
```
pipermerriam: Can we pull this out to a utility somewhere (and test it!)

evm/p2p/kademlia.py
```diff
@@ -0,0 +1,446 @@
...
+            b = bin(x)[2:]
+            return '0' * (k_id_size - len(b)) + b
+
+        if len(self.nodes) < 2:
+            return k_id_size
+
+        bits = [to_binary(n.id) for n in self.nodes]
+        for i in range(1, k_id_size + 1):
+            if len(set(b[:i] for b in bits)) != 1:
```
pipermerriam: This looks like a good stand-alone function that could be well tested if we pulled it out to a utility.

evm/p2p/kademlia.py
```diff
@@ -0,0 +1,446 @@
...
+
+    def split_bucket(self, bucket):
+        a, b = bucket.split()
+        index = self.buckets.index(bucket)
+        self.buckets[index] = a
+        self.buckets.insert(index + 1, b)
+
+    @property
+    def idle_buckets(self):
+        one_hour_ago = time.time() - k_idle_bucket_refresh_interval
```
pipermerriam: naming nitpick:

The name `one_hour_ago` doesn't sit well with me.  Either the `k_idle_bucket_refresh_interval` value should be renamed to `one_hour_in_seconds` or we should rename this to `idle_cutoff_timestamp`.

evm/p2p/kademlia.py
```diff
@@ -0,0 +1,446 @@
...
+        return [b for b in self.buckets if b.last_updated < one_hour_ago]
+
+    @property
+    def not_full_buckets(self):
+        return [b for b in self.buckets if len(b) < k_bucket_size]
```
pipermerriam: Would it be better for this to use the `is_full` function on the bucket?

pipermerriam:
@gsalgado I rebased your `p2p` branch and left a few comments that came to mind as I read through.  Still need to parse the whole thing.  Looking good so far.

My biggest issue that I haven't wrapped my head around is the many mutable data structures in use and whether we can remove some of the mutability.

evm/p2p/kademlia.py
gsalgado: Yeah, I guess it's time to look into custom types for public/private keys as [I suggested recently](https://github.com/pipermerriam/py-evm/pull/81#issue-252268611)

evm/p2p/kademlia.py
gsalgado: These asserts are there to help debugging while I'm working on this code; there's a TODO at the top of the file to remind me to go through all the asserts and either replace them with exceptions or drop them ;-)

evm/p2p/kademlia.py
gsalgado: I've just started working on unit tests for this stuff, and that is going to lead to many refactorings, very possibly what you suggest or maybe a complete rewrite

evm/p2p/kademlia.py
gsalgado: Agreed

gsalgado:
I think it's a bit premature to review this code as there's a lot of stuff there that was added to help debugging and I've just started working on unit tests for it, which should lead to many refactorings. It may be a good idea to have a high-level/architecture review, though

evm/p2p/discovery.py
```diff
@@ -0,0 +1,600 @@
...
+    def listen(self, loop):
+        return loop.create_datagram_endpoint(
+            lambda: self, local_addr=(self.address.ip, self.address.udp_port))
+
+    def connection_made(self, transport):
+        self.transport = transport
+
+    @asyncio.coroutine
+    def bootstrap(self):
+        if self.bootstrapped or len(self.bootstrap_nodes) == 0:
```
pipermerriam: These both feel like anti-patterns.

* The `self.bootstrapped` state is something that I'd like to look for an alternate solution which didn't require storing state (low priority)
* The *silent* return  if the list of bootstrap nodes feels like a silent failure.

I don't have enough knowledge yet to know if this is a dumb solution but...

What do you think about modifying this so that it instead takes an iterable of bootstrap nodes as an argument.  This would remove the silent return when there are no bootstrap nodes and allow moving that check higher in the stack to allow the `DiscoveryProtocol` object to not care about that stuff as well as possibly removing the need to keep the `self.bootstrapped` state around?

gsalgado:
I believe for all change you suggested I've either done them or rewrote the code in a completely different (and hopefully clearer) way. Created #89 to get a second round of feedback, so closing this one

PR #88, opened 2017-09-05
Create public key validation function

egong:
### What was wrong?
The same code for validating public keys was used in two different places.

### How was it fixed?
That code was extracted into a new function in evm.validation.validate_raw_public_key

#### Cute Animal Picture
image

### DIFF
--- a/evm/utils/secp256k1.py
+++ b/evm/utils/secp256k1.py
@@ -17,15 +17,16 @@ from .padding import (
     pad32,
 )
 
+from evm.validation import (
+    validate_raw_public_key
+)
+
 
 def decode_public_key(public_key):
-    if len(public_key) != 64:
-        raise ValueError(
-            "Unexpected public key format: {}. Public keys must be 64 bytes long and must not "
-            "include the fixed \x04 prefix".format(public_key))
-    left = big_endian_to_int(public_key[0:32])
-    right = big_endian_to_int(public_key[32:64])
-    return left, right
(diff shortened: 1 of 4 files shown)


evm/utils/secp256k1.py
```diff
@@ -17,15 +17,16 @@
...
 
 def decode_public_key(public_key):
-    if len(public_key) != 64:
-        raise ValueError(
-            "Unexpected public key format: {}. Public keys must be 64 bytes long and must not "
-            "include the fixed \x04 prefix".format(public_key))
-    left = big_endian_to_int(public_key[0:32])
-    right = big_endian_to_int(public_key[32:64])
-    return left, right
+    if validate_raw_public_key(public_key):
```
pipermerriam: This should not be nested in an `if` statement.  Instead it should just be called at the beginning of the function.  Same with the other applications of this validator.

evm/validation.py
```diff
@@ -159,3 +159,8 @@ def validate_vm_block_numbers(vm_block_numbers):
 
     for block_number in vm_block_numbers:
         validate_block_number(block_number)
+
+
+def validate_raw_public_key(value):
+    if len(value) != 64 or validate_is_bytes(value):
```
pipermerriam: Can you remove the `validate_is_bytes` call from this statement and place it on it's own line above this line.

egong:
@pipermerriam Committed the requested changes.

PR #94, opened 2017-09-19
Piper/state db journalling

pipermerriam:
### What was wrong?

The existing snapshot and revert functionality wasn't going to scale with real blockchain data sizes.

### How was it fixed?

* Remove the snapshot functionality from the database backends
* Implemented a wrapper that journals the underlying db for the state db.

#### Cute Animal Picture

6a00e54ef9645388340134867cb589970c-800wi

### DIFF
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -59,27 +59,27 @@ class JournalDB(BaseDB):
     #
     # Snapshot API
     #
-    def _validate_snapshot(self, snapshot):
-        validate_is_integer(snapshot, title="Journal index")
-        validate_gte(snapshot, 0, title="Journal index")
-        validate_lte(snapshot, len(self.journal), title="Journal index")
+    def _validate_journal_index(self, journal_index):
+        validate_is_integer(journal_index, title="Journal index")
+        validate_gte(journal_index, 0, title="Journal index")
+        validate_lte(journal_index, len(self.journal), title="Journal index")
 
     def snapshot(self):
         return len(self.journal)
 
-    def revert(self, snapshot):
-        self._validate_snapshot(snapshot)
+    def revert(self, journal_index):
+        self._validate_journal_index(journal_index)
 
-        while len(self.journal) > snapshot:
+        while len(self.journal) > journal_index:
             key, previous_value = self.journal.pop()
             if previous_value is None:
                 self.wrapped_db.delete(key)
             else:
                 self.wrapped_db.set(key, previous_value)
 
-    def clear_journal(self, snapshot):
-        self._validate_snapshot(snapshot)
-        self.journal = self.journal[snapshot:]
+    def clear_journal(self, journal_index):
+        self._validate_journal_index(journal_index)
+        self.journal = self.journal[journal_index:]
 
     #
     # Dictionary API
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -13,11 +13,17 @@ class Journal(object):
     A Journal is an ordered list of checkpoints.  A checkpoint is a dictionary
     of database keys and values.  The values are the "original" value of that
     key at the time the checkpoint was created.
+
+    Checkpoints are referenced by a random uuid4.
     """
     checkpoints = None
 
     def __init__(self):
+        # contains an array of `uuid4` instances
         self.checkpoints = []
+        # contains a mapping from all of the `uuid4` in the `checkpoints` array
+        # to a dictionary of key:value pairs wher the `value` is the original
+        # value for the given key at the moment this checkpoint was created.
         self.journal_data = {}
 
     @property
@@ -56,7 +62,8 @@ class Journal(object):
 
     def create_checkpoint(self):
         """
-        Creates a new checkpoint.
+        Creates a new checkpoint.  Checkpoints are referenced by a random uuid4
+        to prevent collisions between multiple checkpoints.
         """
         checkpoint_id = uuid.uuid4()
         self.checkpoints.append(checkpoint_id)
@@ -65,7 +72,9 @@ class Journal(object):
 
     def pop_checkpoint(self, checkpoint_id):
         """
-        Returns all changes from the given checkpoint_id.
+        Returns all changes from the given checkpoint.  This includes all of
+        the changes from any subsequent checkpoints, giving precidence to
+        earlier checkpoints.
         """
         idx = self.checkpoints.index(checkpoint_id)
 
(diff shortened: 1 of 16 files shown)


pipermerriam:
@gsalgado this is really rough right now and it's not doing things correctly at all, but it's a starting point.  I still need to get the following in place.

* Ensure that journaling is *only* applied to state database operations.
* Better data structure handling.  I don't like that the journal is a simple list.
* Some sort of no-brainer snapshot/revert/commit API that I haven't figured out yet.

evm/db/journal.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.wrapped_db.set(key, value)
+
+    def exists(self, key):
+        return self.wrapped_db.exists(key)
+
+    def delete(self, key):
+        try:
+            current_value = self.wrapped_db.get(key)
+        except KeyError:
+            current_value = None
```
pipermerriam: This should just be a `pass`

evm/db/journal.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        - replacing an existing value
+        - setting a value that does not exist
+        """
+        try:
+            current_value = self.wrapped_db.get(key)
+        except KeyError:
+            current_value = None
+
+        self.journal.append((key, current_value))
```
pipermerriam: This should be conditional on the new value not equaling the current value.

evm/db/journal.py
```diff
@@ -0,0 +1,72 @@
...
+        return self.wrapped_db.delete(key)
+
+    #
+    # Snapshot API
+    #
+    def snapshot(self):
+        return len(self.journal)
+
+    def revert(self, snapshot):
+        while len(self.journal) > snapshot:
```
pipermerriam: This needs some bounds checking added for safety.

evm/vm/base.py
```diff
@@ -254,20 +258,21 @@ def configure_header(self, **header_params):
...
-        TODO: This needs to do more than just snapshot the state_db but this is a start.
         """
+        state_root, journal_index = snapshot
+
         with self.state_db() as state_db:
-            return state_db.revert(snapshot)
+            state_db.root_hash = state_root
+            self.journal_db.revert(journal_index)
+
+    def commit(self):
```
pipermerriam: This API feels awkward.  I'm not sure what to rename it to or how to fix it right now.

pipermerriam:
@gsalgado  this is ready for review

evm/vm/base.py
```diff
@@ -255,19 +259,25 @@ def snapshot(self):
...
+        state_root, journal_index = snapshot
+
         with self.state_db() as state_db:
-            return state_db.revert(snapshot)
+            # first revert the database state root.
+            state_db.root_hash = state_root
+            # now roll the underlying database back
+            self.journal_db.revert(journal_index)
+
+    def commit(self):
```
gsalgado: This doesn't seem to be used anywhere.

Also, this is the only place where clear_journal() is called, so unless I'm missing something the journal will never be cleared?

evm/db/state.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
```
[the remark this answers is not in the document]
pipermerriam: Yes and no

We aren't currently testing against python 2....  Maybe we just keep this python3 only since I'm trying to deprecate python2 support around the end of this year.

evm/vm/base.py
pipermerriam: Good catch.

pipermerriam:
@gsalgado I completely reworked the way that `JournalDB` works.  New approach has an effective `commit` mechanism and it should be more efficient as well.  Could be a fluke but it shaved 2 more minutes from the overall runtime.

Can you give this another look?

tests/database/test_journal_db.py
```diff
@@ -0,0 +1,97 @@
+import pytest
+from evm.db.backends.memory import MemoryDB
+from evm.db.journal import JournalDB
+
+
+@pytest.fixture
+def memory_db():
```
gsalgado: ISTM this fixture is not used anywhere other than the journal_db below, so maybe inline it there?

evm/vm/base.py
```diff
@@ -255,19 +260,38 @@ def snapshot(self):
         """
         Perform a full snapshot of the current state of the VM.
 
-        TODO: This needs to do more than just snapshot the state_db but this is a start.
+        Snapshots are a combination of the state_root at the time of the
+        snapshot and the indices of the journaled DB.
```
gsalgado: s/indices/uuids?

evm/db/journal.py
```diff
@@ -0,0 +1,210 @@
...
+            )
+
+    def __contains__(self, value):
+        return value in self.journal_data
+
+
+class JournalDB(BaseDB):
+    """
+    A wrapper around the basic DB objects that keeps a journal of all changes.
+    Snapshots are simply indices into the list of journal entries.  Reversion
```
gsalgado: This sentence is no longer accurate as now the journal entries are kept in a dict with uuids as keys

evm/db/journal.py
```diff
@@ -0,0 +1,210 @@
...
+            # overwrite it.
+            return
+        self.latest[key] = value
+
+    def create_checkpoint(self):
+        """
+        Creates a new checkpoint.
+        """
+        checkpoint_id = uuid.uuid4()
+        self.checkpoints.append(checkpoint_id)
```
gsalgado: I guess you now use uuids (instead of ints) to identify checkpoints to make it harder to accidentally revert/commit the wrong checkpoint?  Maybe it'd be worth adding a comment to __init__() describing that?

evm/db/journal.py
```diff
@@ -1,11 +1,106 @@
...
+from evm.exceptions import ValidationError
+
+
+class Journal(object):
+    """
+    A Journal is an ordered list of checkpoints.  A checkpoint is a dictionary
+    of database keys and values.  The values are the "original" value of that
+    key at the time the checkpoint was created.
+    """
+    checkpoints = None
```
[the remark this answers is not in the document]
pipermerriam: No.  It's a habit I got into a while back to help me keep track of the properties an object has.  It's been almost a year and I'm still not sure it's a good practice.  Thoughts?

evm/vm/flavors/homestead/__init__.py
```diff
@@ -33,6 +33,7 @@ def _apply_homestead_create_message(vm, message):
     computation = vm.apply_message(message)
 
     if computation.error:
+        vm.commit(snapshot)
```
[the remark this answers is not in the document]
pipermerriam: I think you're looking at `_apply_frontier_message`.  If you look at `_apply_frontier_create_message` (which is the frontier method that corresponds to this one) you'll see that it has no snapshot at all.

pipermerriam:
Issues addressed.  Lets see if it's still green.

PR #96, opened 2017-09-22
Implement P2P auth handshake

gsalgado:
I tried to use coincurve for the ecdh exchange, but apparently [it is not compatible with geth's implementation](https://github.com/ofek/coincurve/issues/9), so had to resort to pyca/cryptography.


evm/p2p/auth.py
```diff
@@ -0,0 +1,284 @@
...
+@asyncio.coroutine
+def handshake(remote, privkey):
+    """Perform the auth handshake with the given remote and return a Peer ready to be used.
+
+    The Peer will be configured with the shared secrets established during the handshake.
+    """
+    initiator = HandshakeInitiator(remote, privkey)
+    reader, writer = yield from initiator.connect()
+
+    initiator_nonce = keccak(int_to_big_endian(random.randint(0, 2 ** 256 - 1)))
```
pipermerriam: Is it safe for us to be using `random.random`?  My understanding has always been no.

With `pycryptodome` we can use `Crypto.Random.get_random_bytes(n)` which I understand to be appropriate for key generation.  thoughts?

evm/p2p/auth.py
```diff
@@ -0,0 +1,284 @@
...
+    auth_msg = initiator.create_auth_message(initiator_nonce)
+    auth_init = initiator.encrypt_auth_message(auth_msg)
+    writer.write(auth_init)
+
+    ack_data = yield from reader.read(encrypted_auth_ack_len)
+
+    ephemeral_pubkey, responder_nonce, auth_ack = initiator.decode_auth_ack_message(ack_data)
+    aes_secret, mac_secret, egress_mac, ingress_mac = initiator.derive_secrets(
+        initiator_nonce, responder_nonce, ephemeral_pubkey, auth_init, auth_ack)
```
pipermerriam: *formatting nitpick*

I find this style hard to read.  Can you drop these all onto their own lines and maybe even call it with keyword arguments to ensure that argument ordering doesn't ever get subtly borked.

evm/p2p/auth.py
```diff
@@ -0,0 +1,284 @@
...
+
+    peer = Peer(remote, privkey, reader, writer, aes_secret,
```
pipermerriam: I've always been a bit frightened of function calls and object instantaition that uses more than 3-4 positional arguments given the ease of transposing them or just plain getting the order wrong.  Thoughts on converting this to use keyword arguments?

evm/p2p/auth.py
```diff
@@ -0,0 +1,284 @@
...
+            nonce + ascii_chr(flag)
+        return auth_message
+
+    def decode_auth_ack_message(self, ciphertext):
+        try:
+            eph_pubkey, nonce, version = decode_ack_plain(ciphertext, self.privkey)
+        except AuthenticationError:
+            eph_pubkey, nonce, version = decode_ack_eip8(ciphertext, self.privkey)
+            self.got_eip8_auth = True
+        auth_ack = ciphertext
```
pipermerriam: this assignment struck me as strange.

evm/p2p/auth.py
```diff
@@ -0,0 +1,284 @@
...
+        S = encode_signature(
+            *get_ecc_backend().ecdsa_raw_sign(
+                secret_xor_nonce,
+                self.ephemeral_privkey)
+        )
+
+        # S || H(ephemeral-pubk) || pubk || nonce || 0x0
+        flag = 0x0
+        auth_message = S + keccak(self.ephemeral_pubkey) + self.pubkey + \
+            nonce + ascii_chr(flag)
```
pipermerriam: The `ascii_chr(flag` seems like strange indirection for the literal `b'\x00'`.

evm/p2p/auth.py
```diff
@@ -0,0 +1,284 @@
...
+            # Unused, according to EIP-8, but must be included nevertheless.
+            token_flag = b'\x00'
+            msg = self.ephemeral_pubkey + nonce + token_flag
+        return msg
+
+    def encrypt_auth_ack_message(self, ack_message):
+        if self.got_eip8_auth:
+            # The EIP-8 version has an authenticated length prefix.
+            prefix = struct.pack('>H', len(ack_message) + ecies.encrypt_overhead_length)
+            auth_ack = ecies.encrypt(ack_message, self.remote.pubkey, shared_mac_data=prefix)
```
pipermerriam: Maybe rename to `suffix` to avoid the overwriting of this value on the next line.

evm/p2p/ecies.py
```diff
@@ -0,0 +1,144 @@
...
+    decode_public_key,
+    private_key_to_public_key,
+)
+
+
+cipher = algorithms.AES
+mode = modes.CTR
+curve = ec.SECP256K1()
+# ECIES using AES256 and HMAC-SHA-256-32
+key_len = 32
```
pipermerriam: Thoughts on renaming this and the others above it to `KEY_LEN`? 

I find that casing helps distinguish top level module variables from vars in the local context.

evm/p2p/constants.py
```diff
@@ -0,0 +1,25 @@
...
+auth_msg_len = signature_len + hash_len + pubkey_len + hash_len + 1
+
+# Length of auth ack handshake message
+auth_ack_len = pubkey_len + hash_len + 1
+
+# Length of encrypted pre-EIP-8 initiator handshake
+encrypted_auth_msg_len = auth_msg_len + encrypt_overhead_length
+
+# Length of encrypted pre-EIP-8 handshake reply
+encrypted_auth_ack_len = auth_ack_len + encrypt_overhead_length
```
pipermerriam: Can all of these be changed to `CAP_CASE` so that it's a bit easier to tell they are constants when they are used in other modules.

evm/p2p/kademlia.py
```diff
@@ -81,7 +81,7 @@ def __init__(self, pubkey, address):
     @classmethod
     def from_uri(cls, uri):
         ip, port, pubkey = host_port_pubkey_from_uri(uri)
-        return cls(pubkey, Address(ip.decode(), int(port)))
+        return cls(pubkey, Address(ip.decode(), int(port), int(port)))
```
pipermerriam: Just one line but this seems like it could get extracted and merged separately.

evm/p2p/utils.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def ascii_chr(value):
+    return bytes([value])
+
+
+def sxor(s1, s2):
+    if len(s1) != len(s2):
+        raise ValueError("Cannot sxor strings of different length")
+    return b''.join(ascii_chr(safe_ord(a) ^ safe_ord(b)) for a, b in zip(s1, s2))
```
pipermerriam: I don't know if this is much better in terms of readability. but...

```python
bytes((itertools.starmap(operator.xor, zip(a, b))))
# or just
bytes(x ^ y for x, y in zip(a, b))
```

The `ascii_chr` and `safe_ord` are python2.7 compat right?  Those can be dropped?

evm/utils/keccak.py
```diff
@@ -10,4 +10,9 @@ def keccak(value):
     return keccak_256(value).digest()
 
 
+# FIXME: this is a terrible name.
+def keccak_updateable(value):
```
pipermerriam: why?

evm/p2p/auth.py
gsalgado: Yeah, I'm also a bit weary of them so I'm happy to do that. But I'll move this whole function and the creation of the Peer class to another branch I have, as this really belongs there and will mean less conflicts when I rebase the other branch after we're done with the code review

evm/p2p/auth.py
gsalgado: Right, fixed

evm/p2p/auth.py
gsalgado: Good catch; it makes no sense for this method to do this and then return it afterwards, as this is the very argument passed into the method. Fixed so it no longer returns auth_ack and dropped the assignment

evm/p2p/kademlia.py
gsalgado: This was left by accident, I'll remove it

evm/p2p/utils.py
gsalgado: Yep, leftover from stuff I copied from pydevp2p

evm/utils/keccak.py
gsalgado: I don't know, it just sounds like a weird name

evm/p2p/auth.py
gsalgado: Sure, but as I said below I'm moving this to another branch

evm/utils/keccak.py
pipermerriam: No, I mean why does it exist?

pipermerriam:
This should be rebased now that https://github.com/pipermerriam/py-evm/pull/95 is merged.

evm/utils/keccak.py
gsalgado: Because the existing one give us the .digest(), but for the egress/ingress macs (created in derive_secrets) we need a [hasher which allows us to continuously update its value before getting the digest](https://github.com/pipermerriam/py-evm/pull/98/files#diff-5dad8f91b2b80753e0030b15f47c3931R111)

gsalgado:
Rebased and I think I've addressed all your comments, so should be ready for another pass

evm/utils/keccak.py
pipermerriam: Got it.  I was somehow failing to map the term updatable to the hash.  I was thinking you were namespacing a mirrored version of keccak to be *updated* with some other function at a later time.  👍

evm/p2p/auth.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def create_auth_message(self, nonce):
+        ecdh_shared_secret = ecies.ecdh_agree(self.privkey, self.remote.pubkey)
+        secret_xor_nonce = sxor(ecdh_shared_secret, nonce)
+
+        # S(ephemeral-privk, ecdh-shared-secret ^ nonce)
+        S = self.ephemeral_privkey.sign_msg_hash(secret_xor_nonce).to_bytes()
+
+        # S || H(ephemeral-pubk) || pubk || nonce || 0x0
+        auth_message = S + keccak(self.ephemeral_pubkey.to_bytes()) + self.pubkey.to_bytes() + \
```
pipermerriam: I've found this format a bit more readable than using `\` to split across multiple lines.

```python
auth_message = (
    S +
    keccak(self.ephemeral_pubkey.to_bytes()) +
    self.pubkey.to_bytes() + 
    nonce +
    b'\x00'
)
```

Or:

```python
auth_message = b''.join((
    S,
    keccak(self.ephemeral_pubkey.to_bytes()),
    self.pubkey.to_bytes(),
    nonce,
    b'\x00',
))
```

evm/p2p/test_ecies.py
```diff
@@ -0,0 +1,83 @@
...
+     ),
+]
+
+
+def test_encrypt_decrypt():
+    msg = b'test yeah'
+    privkey = ecies.generate_privkey()
+    ciphertext = ecies.encrypt(msg, privkey.public_key)
+    decrypted = ecies.decrypt(ciphertext, privkey)
+    assert decrypted == msg
```
pipermerriam: This is the type of test that I've historically used hypothesis for (things that have a round trip operation like encrypt/decrypt, encode/decode)

evm/p2p/test_ecies.py
```diff
@@ -0,0 +1,83 @@
...
+
+    # message tag generated by geth
+    tag_secret = decode_hex("0xaf6623e52208c596e17c72cea6f1cb09")
+    tag_input = decode_hex("0x3461282bcedace970df2")
+    tag_expected = decode_hex("0xb3ce623bce08d5793677ba9441b22bb34d3e8a7de964206d26589df3e8eb5183")
+    hmac = ecies.hmac_sha256(tag_secret, tag_input)
+    assert hmac == tag_expected
+
+
+# FIXME: Document those values; this test was lifted from pydevp2p
```
pipermerriam: Couple of remaining `FIXME` left in.  Fine to fix them later but it might be good to at least provide a link here for now.

evm/p2p/utils.py
```diff
@@ -0,0 +1,4 @@
+def sxor(s1, s2):
+    if len(s1) != len(s2):
```
pipermerriam: Since this is used in cryptographic operations, should this use a constant time compare?

evm/utils/keccak.py
pipermerriam: Thinking that it might be better to just use `keccak_256` directly from the `sha3` library instead of this proxy.

evm/p2p/test_ecies.py
gsalgado: Oh, I see!

evm/utils/keccak.py
gsalgado: Agreed, I'll do that

evm/p2p/utils.py
gsalgado: Isn't that the case already, since we're comparing their lengths?

evm/p2p/test_ecies.py
gsalgado: Included the links in the comments

evm/p2p/utils.py
pipermerriam: I should not review code late at night. 👍

PR #98, opened 2017-09-25
Implement auth.handshake() and Peer

gsalgado:
Those together will perform the auth dance and exchange a hello packet with the remote, for a full handshake.

### DIFF
--- a/evm/p2p/auth.py
+++ b/evm/p2p/auth.py
@@ -13,6 +13,9 @@ from eth_keys import keys
 from evm.utils.keccak import (
     keccak,
 )
+from evm.utils.numeric import (
+    int_to_big_endian,
+)
 from evm.p2p import ecies
 from evm.p2p.constants import (
     AUTH_ACK_LEN,
@@ -24,9 +27,37 @@ from evm.p2p.constants import (
     SIGNATURE_LEN,
     SUPPORTED_RLPX_VERSION,
 )
+from evm.p2p.peer import Peer
 from evm.p2p.utils import (
     sxor,
 )
+
+
+@asyncio.coroutine
+def handshake(remote, privkey):
+    """Perform the auth handshake with the given remote and return a Peer ready to be used.
+
+    The Peer will be configured with the shared secrets established during the handshake.
+    """
+    initiator = HandshakeInitiator(remote, privkey)
+    reader, writer = yield from initiator.connect()
+
+    initiator_nonce = keccak(int_to_big_endian(random.randint(0, 2 ** 256 - 1)))
+    auth_msg = initiator.create_auth_message(initiator_nonce)
+    auth_init = initiator.encrypt_auth_message(auth_msg)
+    writer.write(auth_init)
+
+    auth_ack = yield from reader.read(ENCRYPTED_AUTH_ACK_LEN)
+
+    ephemeral_pubkey, responder_nonce = initiator.decode_auth_ack_message(auth_ack)
+    aes_secret, mac_secret, egress_mac, ingress_mac = initiator.derive_secrets(
+        initiator_nonce, responder_nonce, ephemeral_pubkey, auth_init, auth_ack)
+
+    peer = Peer(remote, privkey, reader, writer, aes_secret,
+                mac_secret, egress_mac, ingress_mac)
+    # TODO: Decode the hello packet we received to do the capabilities matching
+    # TODO: Use the peer created above to send a hello packet.
+    return peer
 
 
 class HandshakeBase:
--- a/evm/p2p/auth.py
+++ b/evm/p2p/auth.py
@@ -13,9 +13,6 @@ from eth_keys import keys
 from evm.utils.keccak import (
     keccak,
 )
-from evm.utils.numeric import (
-    int_to_big_endian,
-)
 from evm.p2p import ecies
 from evm.p2p.constants import (
     AUTH_ACK_LEN,
@@ -42,7 +39,7 @@ def handshake(remote, privkey):
     initiator = HandshakeInitiator(remote, privkey)
     reader, writer = yield from initiator.connect()
 
-    initiator_nonce = keccak(int_to_big_endian(random.randint(0, 2 ** 256 - 1)))
+    initiator_nonce = keccak(os.urandom(HASH_LEN))
     auth_msg = initiator.create_auth_message(initiator_nonce)
     auth_init = initiator.encrypt_auth_message(auth_msg)
     writer.write(auth_init)
@@ -51,12 +48,17 @@ def handshake(remote, privkey):
 
     ephemeral_pubkey, responder_nonce = initiator.decode_auth_ack_message(auth_ack)
     aes_secret, mac_secret, egress_mac, ingress_mac = initiator.derive_secrets(
-        initiator_nonce, responder_nonce, ephemeral_pubkey, auth_init, auth_ack)
+        initiator_nonce,
+        responder_nonce,
+        ephemeral_pubkey,
+        auth_init,
+        auth_ack
+    )
 
-    peer = Peer(remote, privkey, reader, writer, aes_secret,
-                mac_secret, egress_mac, ingress_mac)
-    # TODO: Decode the hello packet we received to do the capabilities matching
-    # TODO: Use the peer created above to send a hello packet.
+    peer = Peer(remote=remote, privkey=privkey, reader=reader, writer=writer,
+                aes_secret=aes_secret, mac_secret=mac_secret, egress_mac=egress_mac,
+                ingress_mac=ingress_mac)
+    peer.send_hello()
     return peer
 
 
--- a/evm/p2p/les.py
+++ b/evm/p2p/les.py
@@ -1,0 +1,9 @@
+from evm.p2p.protocol import (
+    Protocol,
+)
+
+
+class LESProtocol(Protocol):
+    name = b'les'
+    version = 1
+    commands = []
--- a/evm/p2p/peer.py
+++ b/evm/p2p/peer.py
@@ -1,5 +1,9 @@
 import asyncio
 import logging
+import struct
+
+import rlp
+from rlp import sedes
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives.ciphers import Cipher, algorithms, modes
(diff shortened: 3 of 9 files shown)


evm/p2p/auth.py
```diff
@@ -0,0 +1,285 @@
...
+@asyncio.coroutine
+def handshake(remote, privkey):
+    """Perform the auth handshake with the given remote and return a Peer ready to be used.
+
+    The Peer will be configured with the shared secrets established during the handshake.
+    """
+    initiator = HandshakeInitiator(remote, privkey)
+    reader, writer = yield from initiator.connect()
+
+    initiator_nonce = keccak(int_to_big_endian(random.randint(0, 2 ** 256 - 1)))
```
gsalgado: From #96:
Is it safe for us to be using random.random? My understanding has always been no.

With pycryptodome we can use Crypto.Random.get_random_bytes(n) which I understand to be appropriate for key generation. thoughts?

gsalgado: @pipermerriam On py3.6 we could use the new `secrets` module, but I guess we want to support 3.5? Anyway, [the pyca/cryptography docs suggest this alternative](https://cryptography.io/en/latest/random-numbers/): `int.from_bytes(os.urandom(20), byteorder="big")`

pipermerriam:
Same as #96: This should be rebased now that https://github.com/pipermerriam/py-evm/pull/95 is merrged.

evm/p2p/les.py
```diff
@@ -0,0 +1,9 @@
+from evm.p2p.protocol import (
+    Protocol,
+)
+
+
+class LESProtocol(Protocol):
```
pipermerriam: 🎂 ❗️

evm/p2p/peer.py
```diff
@@ -0,0 +1,228 @@
...
+            # hello['capabilities']
+            self.logger.debug("Got hello: {}".format(decoded_msg))
+
+    def encrypt(self, header, frame):
+        if len(header) != HEADER_LEN:
+            raise ValueError("Unexpected header length: {}".format(len(header)))
+
+        header_ciphertext = self.aes_enc.update(header)
+        mac_secret = self.egress_mac.digest()[:HEADER_LEN]
+        # egress-mac.update(aes(mac-secret,egress-mac) ^ header-ciphertext).digest
```
pipermerriam: leftover comment?

evm/p2p/peer.py
```diff
@@ -0,0 +1,228 @@
...
+        self.egress_mac.update(sxor(self.mac_enc(mac_secret), header_ciphertext))
+        header_mac = self.egress_mac.digest()[:HEADER_LEN]
+
+        frame_ciphertext = self.aes_enc.update(frame)
+        # egress-mac.update(aes(mac-secret,egress-mac) ^
+        # left128(egress-mac.update(frame-ciphertext).digest))
```
pipermerriam: leftover comments?

evm/p2p/peer.py
```diff
@@ -0,0 +1,228 @@
...
+                len(data), (read_size + MAC_LEN)))
+
+        frame_ciphertext = data[:read_size]
+        frame_mac = data[read_size:read_size + MAC_LEN]
+
+        self.ingress_mac.update(frame_ciphertext)
+        fmac_seed = self.ingress_mac.digest()[:MAC_LEN]
+        self.ingress_mac.update(sxor(self.mac_enc(fmac_seed), fmac_seed))
+        expected_frame_mac = self.ingress_mac.digest()[:MAC_LEN]
+        if frame_mac != expected_frame_mac:
```
pipermerriam: Constant time compare?

evm/p2p/peer.py
```diff
@@ -0,0 +1,228 @@
...
+        if len(data) != HEADER_LEN + MAC_LEN:
+            raise ValueError("Unexpected header length: {}".format(len(data)))
+
+        header_ciphertext = data[:HEADER_LEN]
+        header_mac = data[HEADER_LEN:]
+        mac_secret = self.ingress_mac.digest()[:HEADER_LEN]
+        aes = self.mac_enc(mac_secret)[:HEADER_LEN]
+        self.ingress_mac.update(sxor(aes, header_ciphertext))
+        expected_header_mac = self.ingress_mac.digest()[:HEADER_LEN]
+        if expected_header_mac != header_mac:
```
pipermerriam: constant time compare?

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+    def encode_payload(cls, data):
+        if isinstance(data, dict):  # convert dict to ordered list
+            if not isinstance(cls.structure, list):
+                raise ValueError("Command.structure must be a list when data is a dict")
+            data = [data[name] for (name, _) in cls.structure]
+        if isinstance(cls.structure, sedes.CountableList):
+            encoder = cls.structure
+        else:
+            assert len(data) == len(cls.structure)
+            encoder = sedes.List([type_ for (name, type_) in cls.structure])
```
pipermerriam: *nitpick*

Maybe replace `name` with `_` since it's unused.

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+        return rlp.encode(data, sedes=encoder)
+
+    @classmethod
+    def decode_payload(cls, rlp_data):
+        if isinstance(cls.structure, sedes.CountableList):
+            decoder = cls.structure
+        else:
+            decoder = sedes.List(
+                [type_ for (name, type_) in cls.structure], strict=cls.decode_strict)
```
pipermerriam: Same here with s/`name`/`_`

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+        data = rlp.decode(rlp_data, sedes=decoder)
+        if isinstance(cls.structure, sedes.CountableList):
+            return data
+        else:
+            return dict((cls.structure[i][0], v) for i, v in enumerate(data))
```
pipermerriam: Is this any easier to read?

```python
return {
    field_name: value 
    for ((field_name, _), value) 
    in zip(self.structure, data)
}
```

pipermerriam: Should there be a validation step to ensure `len(data) == len(self.structure)`

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+    id = None
+    decode_strict = True
+    structure = []
+
+    @classmethod
+    def encode_payload(cls, data):
+        if isinstance(data, dict):  # convert dict to ordered list
+            if not isinstance(cls.structure, list):
+                raise ValueError("Command.structure must be a list when data is a dict")
+            data = [data[name] for (name, _) in cls.structure]
```
pipermerriam: Feels weird that if `data` had a key in it that isn't in `cls.structure` that it would get silently dropped.  Unless there is a good reason to allow this, can you add validation that the the keys match?

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+        if isinstance(cls.structure, sedes.CountableList):
+            encoder = cls.structure
+        else:
+            assert len(data) == len(cls.structure)
```
pipermerriam: Should this raise an `rlp.EncodingError` instead?

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+
+def pad_to_16_byte_boundary(data):
+    """Pad the given data with NULL_BYTE up to the next 16-byte boundary."""
+    remainder = len(data) % 16
+    if remainder != 0:
+        data += NULL_BYTE * (16 - remainder)
+    return data
+
+
+def roundup_16(x):
```
pipermerriam: Since this is imported in `peer.py` can it get moved to a *utils* file of some sort so that it doesn't end up getting used as a public API?

evm/p2p/protocol.py
```diff
@@ -0,0 +1,95 @@
...
+
+    def process(self, cmd_id, msg):
+        cmd = self.cmd_by_id[cmd_id]
+        return cmd.handle(self, msg)
+
+    def send(self, header, body):
+        self.peer.send(header, body)
+
+
+def pad_to_16_byte_boundary(data):
```
pipermerriam: Mind prefixing this function name with a `_` to make it clear that it's private.

evm/p2p/test_auth.py
```diff
@@ -101,6 +111,35 @@ def test_handshake():
...
+        egress_mac=egress_mac, ingress_mac=ingress_mac)
+
+    # The hello msg sent by the initiator is going to be fed directly into the responder's
+    # reader, and thus read_msg() will return immediately.
+    initiator_peer.send_hello()
+    msg = yield from responder_peer.read_msg()
+
+    cmd_id = rlp.decode(msg[:1], sedes=sedes.big_endian_int)
+    assert cmd_id == Hello.id
+    responder_peer.process_msg(msg)
```
pipermerriam: feels odd to have the test end with something other than an assertion... is there something *after* this that could be asserted or a reason that the test ends with this statement?

evm/p2p/peer.py
```diff
@@ -0,0 +1,228 @@
...
+
+class Peer:
+    logger = logging.getLogger("evm.p2p.peer.Peer")
+    _sub_protocols = [LESProtocol]
+    # FIXME: Must be configurable.
+    listen_port = 30303
+
+    # We use keyword arguments here because there are too many arguments and it's very easy
+    # for callers to get them wrong.
+    def __init__(self, remote=None, privkey=None, reader=None, writer=None, aes_secret=None,
```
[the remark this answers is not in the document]
gsalgado: Yeah, I'm not a fan of **kwargs either, and indeed having the defaults here is not helping with anything so will do as you suggest

evm/p2p/peer.py
gsalgado: Yup, removed

evm/p2p/peer.py
gsalgado: Yup, removed

evm/p2p/peer.py
gsalgado: Right, fixed

evm/p2p/peer.py
gsalgado: Right, fixed

evm/p2p/test_auth.py
gsalgado: Right now there's nothing to assert after that; it's only checking that the message processing doesn't raise an exception. In the branch I'm currently working on I assert that the peer has been configured with sub-protocols matching the remote's capabilities, though

evm/p2p/protocol.py
gsalgado: No reason for that, added validation

evm/p2p/protocol.py
gsalgado: Yeah, definitely something other than an assert.

evm/p2p/protocol.py
gsalgado: I think so

evm/p2p/protocol.py
```diff
@@ -0,0 +1,92 @@
...
+    id = None
+    decode_strict = True
+    structure = []
+
+    @classmethod
+    def encode_payload(cls, data):
+        if isinstance(data, dict):  # convert dict to ordered list
+            if not isinstance(cls.structure, list):
+                raise ValueError("Command.structure must be a list when data is a dict")
+            data = [data[name] for name, _ in cls.structure]
```
pipermerriam: I would still really like to see validation that we have an exact key match before we reach into the dictionary.  Motivation here is to provide the most helpful error messages possible in the case of a mismatched key.  

Here are two functions copy/pasted from `ethereum-tester` that can be combined to do this.

```python
def validate_no_extra_keys(value, allowed_keys):
    extra_keys = tuple(sorted(set(value.keys()).difference(allowed_keys)))
    if extra_keys:
        raise ValidationError(
            "Only the keys '{0}' are allowed.  Got extra keys: '{1}'".format(
                "/".join(tuple(sorted(allowed_keys))),
                "/".join(extra_keys),
            )
        )


def validate_has_required_keys(value, required_keys):
    missing_keys = tuple(sorted(set(required_keys).difference(value.keys())))
    if missing_keys:
        raise ValidationError(
            "Blocks must contain all of the keys '{0}'.  Missing the keys: '{1}'".format(
                "/".join(tuple(sorted(required_keys))),
                "/".join(missing_keys),
            )
        )
```

gsalgado: I'm a bit uncomfortable using those validation functions here, mainly because they raise a ValidationError, which could end up being handled somewhere higher up the stack when in fact a mismatch in the keys here would mean a bug somewhere and we'd want the exception to bubble all the way up. Would you be ok with just comparing both sets of keys and raising a ValueError containing both sets in case they're not identical?

PR #92, opened 2017-09-08
Implement MainnetTesterChain

pipermerriam:
Blocker for: https://github.com/pipermerriam/ethereum-tester/pull/14

### What was wrong?

Py-EVM needs a *tester* chain that can be used with testing.  Primary features needed are:

* bypass POW validation
* easy configuration for starting blocks for the various hard forks
* easy initialization of the chain with custom state

### How was it fixed?

Added a new `Chain` subclass under `evm.vm.flavors.tester` with these properties.

#### Cute Animal Picture

531652_15a5d09296b0674673641d8e449a26d5_large

PR #103, opened 2017-09-27
Piper/add chain.mine block api

pipermerriam:
### What was wrong?

* The `Chain.ensure_blocks_are_equal` seemed to only exist because it needed to be overridden in some tests.
* Needed a `Chain.mine_block()` function for non-import based blockchain advancement.  This is needed in #92 
* The `configure_homestead_header` function delegated to the `configure_frontier_header` function, but only to perform the validation of the `header_params`.

### How was it fixed?

* Modified the test fixture that needed this override and moved this to a utility function.
* Added a `Chain.mine_block()` API
* Moved the `header_params` validation to a stand-alone function and decoupled the two frontier and homestead functions.

#### Cute Animal Picture

### DIFF
--- a/evm/validation.py
+++ b/evm/validation.py
@@ -204,3 +204,26 @@ def validate_vm_block_numbers(vm_block_numbers):
 
     for block_number in vm_block_numbers:
         validate_block_number(block_number)
+
+
+ALLOWED_HEADER_FIELDS = {
+    'coinbase',
+    'gas_limit',
+    'timestamp',
+    'extra_data',
+    'mix_hash',
+    'nonce',
+    'uncle_hash',
+}
+
+
+def validate_header_parames_for_configuration(header_params):
+    extra_fields = set(header_params.keys()).difference(ALLOWED_HEADER_FIELDS)
+    if extra_fields:
+        raise ValidationError(
+            "The `configure_header` method may only be used with the fields ({0}). "
+            "The provided fields ({1}) are not supported".format(
+                ", ".join(tuple(sorted(ALLOWED_HEADER_FIELDS))),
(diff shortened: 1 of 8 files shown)


pipermerriam:
@gsalgado Can you give this a look?  It's technically three things.  All small.  If you'd prefer I separate them into individual PRs I will.

evm/validation.py
```diff
@@ -204,3 +204,26 @@ def validate_vm_block_numbers(vm_block_numbers):
...
+    'gas_limit',
+    'timestamp',
+    'extra_data',
+    'mix_hash',
+    'nonce',
+    'uncles_hash',
+}
+
+
+def validate_header_parames_for_configuration(header_params):
```
gsalgado: s/parames/params?

tests/core/fixtures.py
```diff
@@ -76,6 +83,29 @@ def chain():
...
+        )
+
+    parent_chain = chain.get_parent_chain(block)
+    imported_block = parent_chain.get_vm().import_block(block)
+
+    persist_block_to_db(chain.db, imported_block)
+    if chain.should_be_canonical_chain_head(imported_block):
+        chain.add_to_canonical_chain_head(imported_block)
+
+    return imported_block
```
gsalgado: I'm a bit worried about duplicating the whole import_block() method here as there's nothing to ensure both implementations are identical. It could lead to tests that pass just because we're not exercising the actual production code. That's not an issue when we simply make the validation method a no-op

pipermerriam: Updated to use the kwarg approach we talked about.

pipermerriam:
@gsalgado Addressed your comments.  going to merge when green unless I hear otherwise.

PR #104, opened 2017-09-27
Remove unnecessary return value from decode_authentication()

gsalgado:
One of its return values was its only argument.

Also refactors a few things to do with decoding of auth msgs, to clarify
a few non-obvious things


evm/p2p/auth.py
```diff
@@ -206,18 +202,19 @@ def decode_ack_eip8(ciphertext, privkey):
 
     Returns the remote's ephemeral pubkey, nonce and protocol version.
     """
-    size = struct.unpack('>H', ciphertext[:2])[0] + 2
-    if len(ciphertext) != size:
-        raise ValueError("Invalid auth ack message size")
-    message = ecies.decrypt(ciphertext[2:size], privkey, shared_mac_data=ciphertext[:2])
+    # The length of the actual msg is stored in plaintext on the first two bytes.
+    encoded_size = ciphertext[:2]
+    (size,) = struct.unpack('>H', encoded_size)
```
pipermerriam: This `size` variable doesn't appear to be used anywhere...

evm/p2p/auth.py
```diff
@@ -230,10 +227,11 @@ def decode_auth_plain(ciphertext, privkey):
 
 def decode_auth_eip8(ciphertext, privkey):
     """Decode EIP-8 auth message format"""
-    size = struct.unpack('>H', ciphertext[:2])[0] + 2
-    if len(ciphertext) < size:
-        raise ValueError("Message shorter than specified size")
-    message = ecies.decrypt(ciphertext[2:size], privkey, shared_mac_data=ciphertext[:2])
+    # The length of the actual msg is stored in plaintext on the first two bytes.
+    encoded_size = ciphertext[:2]
+    (size,) = struct.unpack('>H', encoded_size)
```
pipermerriam: Same here.  `size` not used anywhere.

evm/p2p/auth.py
gsalgado: Indeed, removed

evm/p2p/auth.py
gsalgado: This one too

PR #108, opened 2017-09-30
Piper/rename flavor to fork

pipermerriam:
### What was wrong?

* While *flavor* was a clever name, *fork* is more accurate.
* The `Chain` subclasses were mixed in under the `VM` subclasses

### How was it fixed?

* Renamed *flavor* to *fork* for import paths.
* Moves the `MainnetChain` and `MainnetTesterChain` to live under `evm.chains` since thats what they are.

#### Cute Animal Picture

TODO


pipermerriam:
This should be strictly code re-organization, but extra 👀 to sanity check it would be nice.  @gsalgado

gsalgado:
:+1:

PR #110, opened 2017-10-04
New test fixture that gives us two directly connected p2p peers

gsalgado:
These peers are configured with matching auth secrets and have their
readers/writers connected directly so we can test roundtrips for
protocol-specific interactions.

### DIFF
--- a/evm/p2p/test_peer.py
+++ b/evm/p2p/test_peer.py
@@ -1,7 +1,106 @@
+import asyncio
+import os
+
+import pytest
+
+from evm.utils.keccak import (
+    keccak,
+)
+from evm.p2p import auth
+from evm.p2p import constants
+from evm.p2p import ecies
+from evm.p2p import kademlia
 from evm.p2p.les import LESProtocol
 from evm.p2p.peer import Peer
 from evm.p2p.protocol import Protocol
 from evm.p2p.p2p_proto import P2PProtocol
+
+
+@pytest.fixture
+@asyncio.coroutine
+def directly_linked_peers():
+    """Create two Peers with their readers/writers connected directly.
+
+    The first peer's reader will write directly to the second's writer, and vice-versa.
+    """
+    peer1_private_key = ecies.generate_privkey()
+    peer2_private_key = ecies.generate_privkey()
+    peer1_remote = kademlia.Node(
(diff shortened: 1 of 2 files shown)


evm/p2p/test_peer.py
```diff
@@ -1,9 +1,108 @@
...
+        peer1_private_key.public_key, kademlia.Address('0.0.0.0', 0, 0))
+    initiator = auth.HandshakeInitiator(peer1_remote, peer1_private_key)
+    peer2_reader = asyncio.StreamReader()
+    peer1_reader = asyncio.StreamReader()
+    # Link the peer1's writer to the peer2's reader, and the peer2's writer to the
+    # peer1's reader.
+    peer2_writer = type(
+        "mock-streamwriter",
+        (object,),
+        {"write": lambda data: peer1_reader.feed_data(data)}
```
pipermerriam: *curious*  

Instead of the lambda, could you use `staticmethod` here?

evm/p2p/test_peer.py
```diff
@@ -1,9 +1,108 @@
...
+        (object,),
+        {"write": lambda data: peer2_reader.feed_data(data)}
+    )
+
+    peer1, peer2 = None, None
+    handshake_finished = asyncio.Event()
+
+    @asyncio.coroutine
+    def do_handshake():
+        nonlocal peer1, peer2
```
pipermerriam: *fancy!*

evm/p2p/test_peer.py
gsalgado: For that I'd have to define an actual function beforehand, no? That's what I wanted to avoid here since it's a one-liner

pipermerriam: I just tested this an you can (if you want) replace this with

```python
{"write": staticmethod(peer1_reader.feed_data)}
```

Basic testing shows that you can wrap a method of an instance of another class in `staticmethod` and it works as expected.

gsalgado: Yeah, I'm fine with that but in this case why wrap it in a staticmethod() at all? I could use `peer1_reader.feed_data` directly:
```
{"write": peer1_reader.feed_data}
```

pipermerriam: Hrm, funny, I assumed that wouldn't work.

PR #112, opened 2017-10-05
Implement LES' header sync commands

gsalgado:


evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
+        'flowControl/MRR': sedes.big_endian_int,
+    }
+
+    def decode_payload(self, rlp_data):
+        data = super(Status, self).decode_payload(rlp_data)
+        # The LES/Status msg contains an arbitrary list of (key, value) pairs, where values can
+        # have different types and unknown keys should be ignored for forward compatibility
+        # reasons, so here we need an extra pass to deserialize each of the key/value pairs we
+        # know about.
+        decoded = {}
```
pipermerriam: Instead of building up local state here, can we instead use the `eth_utils.to_dict` decorator and change this to use `yield key, value` to build the return data.

evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
+        for key, value in data:
+            # The sedes.binary we use in .structure above will give us a bytes value here, but
+            # using bytes as dictionary keys makes it impossible to use the dict() constructor
```
pipermerriam: I'm not following what exactly is broken here?

```python
In [1]: x = {}

In [2]: x[b'arst'] = 3

In [3]: x
Out[3]: {b'arst': 3}

In [4]: x = {b'arst': 3}

In [5]: x
Out[5]: {b'arst': 3}
```

Seems to work as expected to me?  I must be missing something.

evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
+            if item_sedes is not None:
+                decoded[key] = item_sedes.deserialize(value)
+            else:
+                decoded[key] = value
+        return decoded
+
+    def encode_payload(self, data):
+        response = []
+        for key, value in data.items():
+            response.append((key, self.items_sedes[key].serialize(value)))
```
pipermerriam: Can you change this to a list comprehension to remove the local state?

```python
response = [
    (key, self.items_sedes[key].serialize(value))
    for key, value
    in data.items()
]
```

or with some fun/fancy `toolz` magic

```python
from cytoolz import itemmap
response = itemmap(lambda k, v: self.item_sedes[k].serialize(v), data)
```

evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
+            item_sedes = self.items_sedes[key]
+            if item_sedes is not None:
+                decoded[key] = item_sedes.deserialize(value)
+            else:
+                decoded[key] = value
+        return decoded
+
+    def encode_payload(self, data):
+        response = []
+        for key, value in data.items():
```
pipermerriam: What do you think about sorting this by keys so that our return value is always consistently ordered?

evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
 
 class LESProtocol(Protocol):
     name = b'les'
     version = 1
-    _commands = [Status]
-    # FIXME: Need to find out the correct value for this
-    cmd_length = 21
+    _commands = [Status, Announce, BlockHeaders]
+    _req_id = 0
+    max_headers = 10
```
pipermerriam: Where does this number come from?

evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
+        self.send(*cmd.encode(resp))
+        self.logger.debug("Sending LES/Status msg: {}".format(resp))
+
+    def process_handshake(self, decoded_msg):
+        # TODO: Possibly disconnect if any of
+        # 1. remote doesn't serve headers
+        # 2. genesis hash does not match
+        pass
+
+    def next_req_id(self):
```
pipermerriam: Thoughts on using `itertools.counter` for this so that you can just do `req_id = next(self.req_id_counter)`?

evm/p2p/les.py
```diff
@@ -6,14 +18,159 @@
...
+
+    def send_get_block_headers(self, block_number, reverse=True, max_headers=max_headers):
+        """Send a GetBlockHeaders msg to the remote.
+
+        This requests that the remote send us up to max_headers, starting from block_number if
+        reverse is False or ending at block_number if reverse is True.
+        """
+        cmd = GetBlockHeaders(self.cmd_id_offset)
+        req_id = self.next_req_id()
+        skip = 0
```
pipermerriam: Maybe a comment explaining what `skip` does here?  This is confusing to me since I don't know the protocol very well at this point.

evm/p2p/les.py
gsalgado: If I construct the dictionary using dict() and keyword arguments, then later use the values from the decoded payload, it won't work because the dictionary will have strings as keys whereas the decoded payload will be bytes. I did this just to standardise on strings as dict keys, as that's what I used in .items_sedes as well

```python
In [1]: d = dict(foo='bar')

In [2]: d[b'foo']
------- KeyError                                  Traceback (most recent call last) 
<ipython-input-2-d73dd6c877f2> in <module>()
----> 1 d[b'foo']

KeyError: b'foo'
```

evm/p2p/les.py
gsalgado: It's arbitrary, just the # of block headers we ask for when we send a GetBlockHeaders request. I'll add a FIXME comment

evm/p2p/les.py
gsalgado: Sounds good, but I'd prefer to have only one place responsible for incrementing the counter, so I kept `next_req_id` (although I've turned it into a `@property`)

evm/p2p/les.py
gsalgado: Done. It's what python APIs call step. I could rename it but I think it's better to use the same name used in the spec, with a comment

gsalgado:
My strategy for testing right now is to try and have as much code coverage as possible, with high-level functional tests and easy-to-run manual tests (e.g. running `python peer.py` against a local geth instance). I believe this gives us the best ROI at this point since there's a lot I'm still learning about ethereum/p2p and how these APIs I'm writing are going to be used, so I keep going back and refactoring them frequently, which would mean a lot of effort to maintain fine-grained unit tests up to date.

Here's the current coverage report, btw: https://pasteboard.co/GOg1eL4.png

PR #114, opened 2017-10-06
Implement BaseChainDB and move all db-related APIs there

gsalgado:


evm/utils/chaindb.py
```diff
@@ -0,0 +1,40 @@
...
+    lookup_block_hash,
+    persist_block_to_db,
+)
+from evm.utils.receipts import (
+    get_receipts_from_db,
+)
+
+
+# XXX: Had to place this under evm/utils because under evm/chains it creates circular dependencies
+# when we import it in evm/vm/base and evm/chains/__init__.py ends up importing VM from evm/vm.py
```
pipermerriam: Maybe just it's own place like `evm.db.chain`?

evm/vm/forks/frontier/blocks.py
```diff
@@ -243,6 +244,10 @@ def from_header(cls, header, db):
                 sedes=CountableList(BlockHeader),
             )
 
+        # XXX: This doesn't need to be obtained from chaindb because we have no need to override
```
pipermerriam: A brief perusal of the code suggest we could do the following.

* Alter the `VM` class to expect a `ChainDB` instance as it's `db` argument.
* Modify the various block and transaction lookups that happen internally in the `VM` class to use `ChainDB` methods.
* Move the `state_db` API onto the `ChainDB` class.

If I'm reading the code right, this would remove all direct db access by the `VM` class which seems good.  Thoughts?  Did I completely miss what the problem was?

gsalgado: I think that should work, but then we'd also need to either extend ChainDB to implement the BaseDB API, or pass ChainDB.db to JournalDB

gsalgado: Actually, to move the state_db API onto ChainDB, we'd need to move VM.journal_db there as well, which may be a good thing as then we could expose the snapshot/revert/commit APIs via ChainDB, which would allow us to get rid of VM.journal_db

gsalgado:
Pushed a new version with the change you suggested. There are a few XXXs that still need addressing, but I like the way it's looking. What do you think?

pipermerriam:
I'm really liking this as well.  The API for transactions and receipts feels awkward based on the arguments being passed into it.  I suspect we can clean that up with time (doesn't need to be done right away).

gsalgado:
Cool! Do you think it makes sense to get rid of the db-related functions from utils/[db,transactions,receipts].py by inlining them in ChainDB?

pipermerriam:
> Cool! Do you think it makes sense to get rid of the db-related functions from utils/[db,transactions,receipts].py by inlining them in ChainDB?

Maybe not inlining them exactly but yes, I think that all of the utils that directly access the `db` should now be moved somewhere that they won't get re-used since we want everything to access things using the `ChainDB` API

gsalgado:
So, my plan with this was to have a separate implementation of the BaseChainDB API which would use the LES client to fetch data on demand, but for them to be compatible, BaseChainDB's methods would have to be turned into `@asyncio.coroutine`s (together with the `Chain` ones), and that means we'd have to call them with `yield from` everywhere, which is not very nice. I'm thinking it makes more sense to have the new `LightChain` inherit from `Chain` and overwrite the methods that may need to fetch data on-demand, so only those are `@asyncio.coroutine`s and only users of `LightChain` have to deal with that.

Having said that, I think the BaseChainDB abstraction is a good thing by itself, so we may want to merge this anyway. 

Oh, and I've pushed another commit here to avoid storing two copies of the transactions/uncles in the DB -- it was a problem because when a new block was persisted to the DB without being mined/imported, the transactions were only saved keyed by their hashes, but when loading a block from the DB we expect to find them in the Trie rooted at `header.transaction_root`.

pipermerriam:
> BaseChainDB's methods would have to be turned into `@asyncio.coroutines` (together with the Chain ones), and that means we'd have to call them with `yield from` everywhere, which is not very nice.

Don't we have to do this anyways (eventually) assuming that we are going to support running as a full client at some point?

> I'm thinking it makes more sense to have the new `LightChain` inherit from `Chain` and overwrite the methods that may need to fetch data on-demand

This seems like a good intermediate solution.  If we do end up needing it in the main `Chain` object we can port those changes up to the base object.

> so we may want to merge this anyway.

Taking a look at the PR now to provide any feedback I can tome up with.  👍 on this plan in general.

evm/chains/chain.py
```diff
@@ -279,21 +275,21 @@ def get_chain_at_block_parent(self, block):
             ))
 
         init_header = self.create_header_from_parent(parent_header)
-        return type(self)(self.db, init_header)
+        return type(self)(self.chaindb.raw_db, init_header)
```
pipermerriam: Looking at this line as well as [this line from the `from_genesis` method](https://github.com/pipermerriam/py-evm/pull/114/files#diff-38ff1253a2c8ee641d11e159b1584c5fR208) makes me think that we might want to just pass in an instantiated `ChainDB` object rather than the underlying database backend.

tests/database/test_chaindb.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def test_add_block_number_to_hash_lookup(chaindb, block):
+    block_number_to_hash_key = make_block_number_to_hash_lookup_key(block.number)
+    assert not chaindb.exists(block_number_to_hash_key)
+    chaindb.add_block_number_to_hash_lookup(block)
+    assert chaindb.exists(block_number_to_hash_key)
+
+
+def test_perist_block_to_db(chaindb, block):
```
pipermerriam: s/perist/persist/

gsalgado:
>> BaseChainDB's methods would have to be turned into `@asyncio.coroutines` (together with the Chain ones), and that means we'd have to call them with `yield from` everywhere, which is not very nice.

> Don't we have to do this anyways (eventually) assuming that we are going to support running as a full client at some point?

Is the full client going to fetch data on-demand as well? I thought a full client would always have a fully-synced DB so it'd be able to answer anything with the data that is already there

evm/chains/chain.py
gsalgado: Yeah, makes sense, and then I can get rid of Chain.chaindb_class

PR #116, opened 2017-10-11
WIP: Implement PeerManager

gsalgado:
This kind of works currently, but it is very hard to find les/1 nodes on the same
network_id that don't have too many peers yet.

### DIFF
--- a/evm/p2p/peermanager.py
+++ b/evm/p2p/peermanager.py
@@ -1,0 +1,260 @@
+import asyncio
+import logging
+
+from eth_keys import keys
+from eth_utils import decode_hex
+
+from evm.p2p.constants import HANDSHAKE_TIMEOUT
+from evm.p2p.exceptions import (
+    PeerDisconnected,
+    UnreachablePeer,
+    UselessPeer,
+)
+from evm.p2p import ecies
+from evm.p2p.peer import (
+    handshake,
+    LESPeer,
+)
+
+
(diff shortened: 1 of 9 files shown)


pipermerriam:
@gsalgado would it help for you to have a dedicated LES node running somewhere that you can connect to?

gsalgado:
I have one running locally, which I use to test the Peer, but the PeerManager needs to be tested against the actual P2P network, to check that it can find useful nodes and connect to them. That works, but there seem to be few nodes out there that support LES/1 *and* don't have too many peers connected already, so most of the time it takes a while to find even just one of them. AIUI, the v5 of the discovery protocol is supposed to solve this

evm/p2p/peermanager.py
```diff
@@ -0,0 +1,131 @@
...
+    discovery_sleep = 3
+
+    def __init__(self, discovery):
+        self.discovery = discovery
+        self.connected_nodes = {}
+        # FIXME: Need a persistent DB to keep track of failed/successful peers or else every time
+        # we start it takes ages to find peers that support les/1 and are on the same network as
+        # us.
+        self.failed_remotes = []
+        self.useless_peers = []
```
pipermerriam: And a set for this as well?

gsalgado:
The `PeerPool` class, which was merged recently, supersedes this

PR #117, opened 2017-10-11
Implement peer.handshake()

gsalgado:
Similar to auth.handshake(), but this one performs the P2P handshake and
only returns a peer when there are matching sub-protocols.

Also use a timeout on Peer.read() and an asyncio.Event so that Peer.stop() only
returns when the Peer.start() coroutine exits.

### DIFF
--- a/evm/p2p/peer.py
+++ b/evm/p2p/peer.py
@@ -18,8 +18,11 @@ from eth_utils import (
 
 from eth_keys import keys
 
+from evm.p2p import auth
 from evm.p2p import ecies
 from evm.p2p.constants import (
+    CONN_IDLE_TIMEOUT,
+    HANDSHAKE_TIMEOUT,
     HEADER_LEN,
     MAC_LEN,
 )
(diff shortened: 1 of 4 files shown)


evm/p2p/peer.py
```diff
@@ -39,6 +44,27 @@
...
+    Return a Peer connected to that remote in case both handshakes are successful and at least one
+    of the sub-protocols supported by us is also supported by the remote.
+
+    Raises UnreachablePeer if we cannot connect to the peer or UselessPeer if none of the
+    sub-protocols supported by us is also supported by the remote.
+    """
+    try:
+        peer = yield from auth.handshake(remote, privkey)
+        msg = yield from peer.read_msg()
+    except (ConnectionRefusedError, OSError) as e:
```
pipermerriam: Are either of these exceptions specific to the lines in the `try` block or can both lines raise either exception?  If it's the former, can we split this into separate `try` statements?  

I suspect it's the later but since I can't find anything else in this PR to leave feedback on :)

gsalgado: Hmm, good catch; I'm pretty sure only the open_connection() call in auth.handshake() can raise those, so I'll move the read_msg() out of the try block

PR #120, opened 2017-10-13
Piper/backport statedb mutability fixes

pipermerriam:
### What was wrong?

The `vm.state_db` context manager had some pitfalls. 

* it was easy to continue interacting with the state_db after it had left it's context.
* if you were working with a `read_only` statedb, you would not get an error until the context exited.

### How was it fixed?

* Changed the `StateDB` class to have first class support for immutability via a database wrapper.
* Changed the context manager to render the `state_db` object useless once the context has exited.

#### Cute Animal Picture

### DIFF
--- a/evm/db/backends/base.py
+++ b/evm/db/backends/base.py
@@ -1,4 +1,4 @@
-class BaseDB(object):
+class BaseDB:
     def get(self, key):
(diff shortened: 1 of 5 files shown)


gsalgado:
:+1:

PR #121, opened 2017-10-18
WIP: Implement a LightChain/LightClient

gsalgado:


evm/db/chain.py
```diff
@@ -40,6 +40,9 @@
 )
 
 
+CANONICAL_HEAD_HASH_DB_KEY = b'canonical_head_hash'
```
pipermerriam: What do you think about going ahead and adding a version prefix to *all* of these keys that we generate since we're very likely to iterate on the database schema.

evm/constants.py
```diff
@@ -152,7 +152,7 @@
 GENESIS_GAS_LIMIT = 3141592
 GENESIS_PARENT_HASH = ZERO_HASH32
 GENESIS_COINBASE = ZERO_ADDRESS
-GENESIS_NONCE = b'\x00\x00\x00\x00\x00\x00\x00*'  # 42 encoded as big-endian-integer
+GENESIS_NONCE = b'\x00\x00\x00\x00\x00\x00\x00B'  # 42 encoded as big-endian-integer
```
pipermerriam: Why this?  Isn't this `66`?

evm/genesis.py
```diff
@@ -0,0 +1,42 @@
+from eth_utils import decode_hex
+
+from evm.constants import (
+    GENESIS_NONCE,
+    ZERO_HASH32,
+)
+from evm.rlp.headers import BlockHeader
+
+
+MAINNET_GENESIS_HEADER = BlockHeader(
```
pipermerriam: This seems like it should be located in `evm.chains.mainnet`

evm/genesis.py
```diff
@@ -0,0 +1,42 @@
...
+    block_number=0,
+    parent_hash=ZERO_HASH32,
+    receipt_root=decode_hex("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"),  # noqa: E501
+    uncles_hash=decode_hex("0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"),  # noqa: E501
+    state_root=decode_hex("0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544"),
+    timestamp=0,
+    transaction_root=decode_hex("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"),  # noqa: E501
+)
+
+ROPSTEN_GENESIS_HEADER = BlockHeader(
```
pipermerriam: This seems like it should be located in `evm.chains.ropsten` (which I don't believe exists currently)

evm/genesis.py
```diff
@@ -0,0 +1,42 @@
...
+    extra_data=decode_hex("0x3535353535353535353535353535353535353535353535353535353535353535"),
+    gas_limit=16777216,
+    gas_used=0,
+    bloom=0,
+    mix_hash=ZERO_HASH32,
+    nonce=GENESIS_NONCE,
+    block_number=0,
+    parent_hash=ZERO_HASH32,
+    receipt_root=decode_hex("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"),  # noqa: E501
+    uncles_hash=decode_hex("0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347"),  # noqa: E501
```
pipermerriam: This is the same as `evm.constants.EMPTY_UNCLE_HASH` in case you weren't aware that existed.

evm/genesis.py
```diff
@@ -0,0 +1,42 @@
...
+    difficulty=1048576,
+    extra_data=decode_hex("0x3535353535353535353535353535353535353535353535353535353535353535"),
+    gas_limit=16777216,
+    gas_used=0,
+    bloom=0,
+    mix_hash=ZERO_HASH32,
+    nonce=GENESIS_NONCE,
+    block_number=0,
+    parent_hash=ZERO_HASH32,
+    receipt_root=decode_hex("0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421"),  # noqa: E501
```
pipermerriam: This is the same as `evm.constants.BLANK_ROOT_HASH` in case you weren't aware that existed.

evm/constants.py
gsalgado: With the original value the hash for the genesis header does not match the expected value (and causes geth to refuse the peer). IIUC, the nonce should be 0x42 encoded as big-endian-int

```
In [4]: encode_hex(MAINNET_GENESIS_HEADER.hash)
Out[4]: '0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3'

In [5]: MAINNET_GENESIS_HEADER.nonce
Out[5]: b'\x00\x00\x00\x00\x00\x00\x00B'

In [6]: MAINNET_GENESIS_HEADER.nonce = b'\x00\x00\x00\x00\x00\x00\x00*'

In [7]: encode_hex(MAINNET_GENESIS_HEADER.hash)
Out[7]: '0x8fcee18008327b9b03f9001a8ad32e66bea58f6694a43df7ad47f388b29bc489'
```

pipermerriam: Looks good.  Thanks for catching the mistake.

evm/p2p/lightchain.py
```diff
@@ -0,0 +1,76 @@
...
+    def get_peer(self):
+        raise NotImplementedError("TODO")
+
+    @asyncio.coroutine
+    def stop(self):
+        raise NotImplementedError("TODO")
+
+
+class LightChain(Chain):
+    on_demand_data_backend_class = None
```
pipermerriam: Is this intended to be set to `None` or should it be the `OnDemandDataBackend`?

evm/p2p/lightchain.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, chaindb, header=None):
+        super(LightChain, self).__init__(chaindb, header=header)
+        self.on_demand_data_backend = self.on_demand_data_backend_class(chaindb)
```
pipermerriam: instead of using the `chaindb` as passed in through the `__init__` method do you think it would be good to use `self.chaindb` to respect anything that the parent class might do in setting that property.

evm/p2p/lightchain.py
```diff
@@ -0,0 +1,76 @@
...
+    @asyncio.coroutine
+    def stop(self):
+        yield from self.on_demand_data_backend.stop()
+
+    @asyncio.coroutine
+    def get_canonical_block_by_number(self, block_number):
+        try:
+            block_hash = self.chaindb.lookup_block_hash(block_number)
+        except KeyError:
+            raise BlockNotFound("No block with number {} found on canonical chain".format(
```
pipermerriam: Making sure I understand what is going on here:

This uses `self.chaindb` rather than `self.on_demand_data_backend` because it needs to return the *local* version of what the canonical block number is.  *If* we are still syncing block headers and we request a block number that is higher than the latest synced one, we'll hit this exception.

Does it make sense to have a secondary code path here for when `block_number` is less than what we *believe* the current highest block number is, but still greater than the latest block we've synced?

evm/rpc/server.py
```diff
@@ -0,0 +1,176 @@
...
+                head = self.chain.get_canonical_head()
+                number = head.block_number
+            else:
+                number = int(hash_or_number, 16)
+            block = yield from self.chain.get_canonical_block_by_number(number)
+        elif method == 'eth_getBlockByHash':
+            block_hash = decode_hex(hash_or_number)
+            block = yield from self.chain.get_block_by_hash(block_hash)
+        else:
+            raise HTTPMethodNotAllowed(method, self.allowed_methods)
```
pipermerriam: Maybe just add a `TODO` here that instead of raising an exception we should instead return an error response over the RPC server.

evm/rpc/server.py
```diff
@@ -0,0 +1,176 @@
...
+
+DemoLightChain = LightChain.configure(
+    'RPCDemoLightChain',
+    vm_configuration=MAINNET_VM_CONFIGURATION,
+    on_demand_data_backend_class=SinglePeerOnDemandDataBackend,
+    network_id=NETWORK_ID,
+)
+
+
+def quantity_encoder(i):
```
pipermerriam: Can't you just use the built-in `hex` function.  I think it behaves the same as this.

evm/p2p/lightchain.py
gsalgado: Yeah, good point, I'll do that

evm/rpc/server.py
gsalgado: The exception gets caught by aiohttp and turned into a 405 response. Ultimately, in web3.py we get a `HTTPError: 405 Client Error: Method Not Allowed for url: http://localhost:8080/`

evm/rpc/server.py
gsalgado: Indeed! Changed

evm/p2p/lightchain.py
gsalgado: I think eventually we'll provide a default, yeah, but since OnDemandDataBackend is currently incomplete, I thought it'd be better to set it to None and force us to specify a working one (which at this time will be hack-ish like the one used for the RPC server) via `.configure()`

evm/p2p/lightchain.py
gsalgado: For simplicity the LightChain currently assumes the headers and canonical block numbers are always synced, which is probably ok for now as it's only the initial sync that may take some time (15min or so), and after that the headers are synced almost instantaneously every time we get a new announce msg. If we could (and we should be able to) speed up the initial sync, I think it'd be fine to keep the LightChain always assuming headers are fully synced, for the time being, at least

pipermerriam:
:shipit:

PR #122, opened 2017-10-18
Use headers instead of blocks on most Chain methods

gsalgado:


evm/chains/chain.py
```diff
@@ -48,19 +49,21 @@ class Chain(object):
...
+    def __init__(self, chaindb, header=None):
         if not self.vms_by_range:
             raise ValueError(
                 "The Chain class cannot be instantiated with an empty `vms_by_range`"
             )
 
         self.chaindb = chaindb
         self.header = header
+        if self.header is None:
+            self.header = self.create_header_from_parent(chaindb.get_canonical_head())
```
pipermerriam: I like this change, however, *if* the `chaindb` doesn't have a canonical head saved, this will raise a `BlockNotFound` error yes?  If so, can we catch this, and re-raise a nicer error that makes it a bit clearer what went wrong.

evm/db/chain.py
```diff
@@ -40,6 +41,9 @@
 )
 
 
+CANONICAL_HEAD_HASH_DB_KEY = b'canonical_head_hash'
```
pipermerriam: Thoughts on adding a version prefix to this `b'v1:canonical_head_hash'`.  It seems very likely that we'll be iterating on our database schema and it might be nice to have a version namespace from the start.

evm/chains/chain.py
gsalgado: It will actually raise a CanonicalHeadNotFound, as that's what `chaindb.get_canonical_head()` raises.  It could also, in theory, raise a BlockNotFound if there's no header with the hash stored as the head, but `chaindb.set_canonical_head()` shouldn't permit that to happen

evm/db/chain.py
gsalgado: Sorry, forgot about this from the other PR. Yeah, I'm definitely fine with that

PR #124, opened 2017-10-19
Peer now fetches all block headers on startup and stays in sync

gsalgado:
Depends on #122


evm/p2p/peer.py
```diff
@@ -168,7 +220,13 @@ def process_msg(self, msg):
         if proto is None:
             raise UnknownProtocolCommand(
                 "No protocol found for cmd_id {}".format(cmd_id))
-        proto.process(cmd_id, msg)
+        decoded = proto.process(cmd_id, msg)
+        if decoded is None:
+            return
+        request_id = decoded.get('request_id')
+        if request_id is not None and request_id in self._pending_replies:
+            callback = self._pending_replies.pop(request_id)
```
pipermerriam: I think there is a subtle race condition here that may or may not be an issue.

If `request_id` gets `pop`'d between the point where we check it's presence in `self._pending_replies` and this line where we `pop` it, this will fail.  Thoughts on moving this to a `try/except` pattern like the following.

```python
if request_id is not None:
    try:
        cb = self._pending_replies.pop(request_id)
    except KeyError:
        pass
    else:
        callback(decoded)
```

evm/p2p/peer.py
```diff
@@ -280,48 +338,117 @@ def match_protocols(self, remote_capabilities):
...
+                request_id = gen_request_id()
+                target_head = min(head_number + MAX_HEADERS_FETCH, announced_head_number)
+                self.les_proto.send_get_block_headers(
+                    target_head, MAX_HEADERS_FETCH, request_id, reverse=True)
+                reply = yield from self.wait_for_reply(request_id)
+                for header in reversed(reply['headers']):
+                    self.chaindb.persist_header_to_db(header)
+                    # FIXME: The reference to the canonical chain head should not be stored in the
+                    # chaindb as that is shared by all peers -- instead, that should be kept on a
+                    # peer-specific db.
```
pipermerriam: Let me see if I'm following this correctly.  The idea is that we'd sync the chain for each peer, and then we'd merge the various peer-specific chains into a single canonical chain?

evm/p2p/peer.py
```diff
@@ -280,48 +338,117 @@ def match_protocols(self, remote_capabilities):
...
+
+        The available LESProtocol will be the highest version supported by this class and the
+        remote peer. It is available only after the initial P2P handshake.
+        """
+        assert len(self.enabled_sub_protocols) > 0
+        if self._les_proto is None:
+            for proto in self.enabled_sub_protocols:
+                if proto.name == LESProtocol.name:
+                    self._les_proto = proto
+        return self._les_proto
```
pipermerriam: This function seems to be missing handling for the case where none of the enabled sub-protocols are the `LESProtocol`.  Is the intention to return `None` in this case?  If so, it seems like a good comment to put in place to make it explicit.

evm/p2p/peer.py
```diff
@@ -280,48 +338,117 @@ def match_protocols(self, remote_capabilities):
...
-            print("Unexpected error during auth/p2p handhsake with {}: {}".format(remote, e))
+    genesis_header = ROPSTEN_GENESIS_HEADER
+    if args.db is not None:
+        chaindb = BaseChainDB(LevelDB(args.db))
+    else:
+        chaindb = BaseChainDB(MemoryDB())
+    try:
+        chaindb.get_canonical_head()
+        # If the above doesn't raise CanonicalHeadNotFound it means we're reusing an existing db.
+        chain = RopstenChain(chaindb)
```
pipermerriam: Can you move this to an `else` clause so that it's outside of the `try` block to remove the possibility of catching an exception we don't mean to.

evm/p2p/test_peer.py
```diff
@@ -174,3 +198,7 @@ class MockP2PProtocol(P2PProtocol):
 
     def send_handshake(self):
         pass
+
+
+class MockChain:
```
pipermerriam: This doesn't appear to be used anywhere.

evm/p2p/peer.py
gsalgado: Well, that can only happen if we have multiple threads; with asyncio context switches can only happen when you `yield from`. And none of this code is thread-safe at this point. If/when we need multiple threads we'd have to revisit a lot of this stuff to make it thread-safe

evm/p2p/peer.py
gsalgado: Yeah, my current thinking is that we can have multiple peers fetching headers and storing them in a shared (the main) DB, but the info about the canonical chain will have to be specific to every peer and in the LightChain/LightClient we'll need some way to reconcile the info from the various peers and provide a consistent view

evm/p2p/peer.py
gsalgado: Well, this is the LESPeer class, which only supports LESProtocoll and the handshake will fail if there's no matching sub-protocols. Maybe I could add a comment to clarify that?

evm/p2p/test_peer.py
gsalgado: Indeed, removed

evm/p2p/peer.py
gsalgado: :+1:

evm/p2p/peer.py
pipermerriam: Thanks for the reminder.  I keep forgetting the async model.

evm/p2p/peer.py
pipermerriam: In cases like this when the code path is *unreachable* I like to add a `raise Exception("Invariant")` so that if some future change makes it possible it doesn't fail silently.

pipermerriam:
:shipit:

evm/p2p/peer.py
gsalgado: Makes sense, I've added one

PR #126, opened 2017-10-19
Remove hard coded HALT opcodes in favor of exception pattern

pipermerriam:
### What was wrong?

* The `vm.apply_computation` implementation was standard enough that it wasn't being overridden by any of the VM subclasses.
* The opcodes which cause the VM to halt were hard-coded.

### How was it fixed?

* Created a new `Halt` exception that opcodes can raise if they wish to cause VM execution to halt.
* Move the implementation of `apply_computation` into the `BaseVM` class since it doesn't look like it will need to be overwritten.

#### Cute Animal Picture

daily updated funny birds and animals picdump 1

### DIFF
--- a/evm/exceptions.py
+++ b/evm/exceptions.py
@@ -33,6 +33,13 @@ class ValidationError(PyEVMError):
     pass
 
 
+class Halt(PyEVMError):
+    """
+    Raised by opcode function to halt vm execution.
+    """
+    pass
+
+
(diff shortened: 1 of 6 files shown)


gsalgado:
I was going to review this but most tests are failing, so probably best to wait?

gsalgado:
Can't find anything that could be improved here! :+1:

PR #128, opened 2017-10-24
Fix a bug in the calculation of a block's score

gsalgado:
All blocks were getting an extra GENESIS_DIFFICULTY added to their score
since BaseChainDB.get_score() considered that to be the score of the
genesis' parent, which is obviously wrong.

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -12,7 +12,6 @@ from eth_utils import (
 
 from evm.constants import (
     BLANK_ROOT_HASH,
-    GENESIS_DIFFICULTY,
     GENESIS_PARENT_HASH,
 )
 from evm.exceptions import (
@@ -66,11 +65,6 @@ class BaseChainDB:
         self.db.set(CANONICAL_HEAD_HASH_DB_KEY, head_hash)
 
     def get_score(self, block_hash):
-        # TODO: This may be a problem as not all chains will start with the same
-        # genesis difficulty. Maybe make it Chain aware so that it can pull the
-        # genesis block and pull the genesis difficulty from there.
-        if block_hash == GENESIS_PARENT_HASH:
-            return GENESIS_DIFFICULTY
         return rlp.decode(
             self.db.get(make_block_hash_to_score_lookup_key(block_hash)),
             sedes=rlp.sedes.big_endian_int)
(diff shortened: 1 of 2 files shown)


evm/db/chain.py
```diff
@@ -143,7 +137,10 @@ def persist_header_to_db(self, header):
             rlp.encode(header),
         )
 
-        score = self.get_score(header.parent_hash) + header.difficulty
+        if header.parent_hash == GENESIS_PARENT_HASH:
+            score = header.difficulty
+        else:
+            score = self.get_score(header.parent_hash) + header.difficulty
         self.db.set(
             make_block_hash_to_score_lookup_key(header.hash),
```
pipermerriam: This is somewhat orthogonal but I just realized that the `get_score(..)` function results in a write operation (setting the score).  This strikes me as un-expected behavior and I'm wondering if (separately from this PR) we should split this into two pieces or do something like `get_score(block_hash, set_missing_scores=True)` so that it can be called without accidentally mutating the database.

tests/database/test_chaindb.py
```diff
@@ -75,11 +75,22 @@ def test_persist_block_to_db(chaindb, block):
...
+
+    genesis_score_key = make_block_hash_to_score_lookup_key(genesis.hash)
+    genesis_score = rlp.decode(chaindb.db.get(genesis_score_key), sedes=rlp.sedes.big_endian_int)
+    assert genesis_score == 1
+    assert chaindb.get_score(genesis.hash) == 1
+
+    block1 = BlockHeader(difficulty=10, block_number=1, gas_limit=0, parent_hash=genesis.hash)
+    chaindb.persist_header_to_db(block1)
+
+    block1_score_key = make_block_hash_to_score_lookup_key(block1.hash)
```
pipermerriam: Not something that needs to be cleaned in this PR but:

Testing the value directly retrieved from the database feels like testing the implementation vs testing the behavior which typically results in fragile tests.  I suspect that we'll want to run these tests against multiple database backends at some point at which point this test will need to be re-worked.

No real question or request there.  Just a brain dump.

evm/db/chain.py
gsalgado: get_score() does not mutate the db, it simply does a `rlp.decode(self.db.get(...))`. The lines above are part of the `persist_header_to_db()` method

pipermerriam: 🤦‍♂️

tests/database/test_chaindb.py
gsalgado: The original test had this, and I believed it was because you wanted to check that the score was stored under the correct key in the DB, so I kept it. I'd be happy to drop that and rely only on asserting the result of `.get_score()` is what we expect

pipermerriam: nah, lets punt.  It'll fail and make us change it as soon as we start supporting more than one chainDB or database backend or whatever it is that makes this no-longer-appropriate and since we don't know exactly what that's going to look like, I think it's best to wait.

PR #129, opened 2017-10-27
Update canonical chain info whenever we store a header in the DB

gsalgado:
The Chain class used to do that, but now it's done as part of BaseChainDB.persist_header_to_db().

I needed this on another branch I'm working on (which changes the LightChain so that it can handle possibly diverging data from multiple peers), but this seemed like a worthwhile refactoring in itself so I'm submitting it separately

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -8,6 +8,7 @@ from trie import (
 
 from eth_utils import (
     to_list,
+    to_tuple,
 )
 
 from evm.constants import (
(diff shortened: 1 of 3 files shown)


evm/db/chain.py
```diff
@@ -131,6 +169,8 @@ def add_block_number_to_hash_lookup(self, header):
             rlp.encode(header.hash, sedes=rlp.sedes.binary),
         )
 
+    # TODO: This method sould take a chain of headers as that's the most common use case
+    # and it'd be much faster than inserting each header individually.
```
pipermerriam: What sort of optimizations do you think are likely to make this faster if this took a list of headers?

evm/db/chain.py
```diff
@@ -56,19 +57,56 @@ def get_canonical_head(self):
...
+        common with our canonical chain.
+        """
+        h = header
+        while True:
+            yield h
+            if h.parent_hash == GENESIS_PARENT_HASH:
+                break
+            try:
+                orig = self.get_canonical_block_header_by_number(h.block_number)
+                if orig.hash == h.hash:
```
pipermerriam: Can this section get moved to an `else` clause to minimize what's present in the `try` block.

gsalgado: Yep, sure

evm/db/chain.py
gsalgado: For every header we insert we check if it should be the head of the canonical chain, and if it does we have to go through its ancestors to find the chain leading up to the first one it has in common with the canonical chain, then update the block-number-to-hash-lookup entries in the DB for them. If we get a chain of headers here we would be able to do the should-be-canonical check *and* lookup-chain-up-to-common-ancestor just once instead of for every header.

pipermerriam: Ah, yes, that makes sense.

PR #130, opened 2017-10-27
Fix LESPeer to work when there are multiple instances running

gsalgado:
LESPeer's first version used to assume there was only one instance of it running and that our canonical chain was always in sync with that of the remote peer. Now it keeps track of headers fetched from each remote peer and uses that to decide which headers to fetch from them.


evm/p2p/peer.py
```diff
@@ -350,36 +364,96 @@ def __init__(self, remote, privkey, reader, writer, aes_secret, mac_secret,
...
+            headers = yield from self._fetch_headers_starting_at(oldest_ancestor_to_consider)
+            if len(headers) == 0 or not self.chaindb.header_exists(headers[0].hash):
+                self.logger.warn("No common ancestors between us and remote, disconnecting")
+                self.disconnect(DisconnectReason.other)
+                return
+            for header in headers:
+                self.chaindb.persist_header_to_db(header)
+            self.last_synced = chain_head.block_number
+        else:
+            self.last_synced = self.last_synced - reorg_depth
```
pipermerriam: What do you think about renaming `last_synced` to something a little more specific like `synced_block_height` or `synced_to_block_height`?

evm/p2p/peer.py
```diff
@@ -350,36 +364,96 @@ def __init__(self, remote, privkey, reader, writer, aes_secret, mac_secret,
...
+
+        while self.last_synced < head_number:
```
pipermerriam: At this point do we *know* that this peer has a header of at least `head_number` height?  I glanced but couldn't find it.  If not, do we need to add that validation to avoid this loop never terminating?

evm/p2p/peer.py
```diff
@@ -350,36 +364,96 @@ def __init__(self, remote, privkey, reader, writer, aes_secret, mac_secret,
...
+        """Fetches up to self.max_headers_fetch starting at start_block.
+
+        Returns a tuple containing those headers in ascending order of block number.
+        """
+        request_id = gen_request_id()
+        self.les_proto.send_get_block_headers(
+            start_block, self.max_headers_fetch, request_id, reverse=False)
+        reply = yield from self.wait_for_reply(request_id)
+        self.logger.info("fetched headers from {} to {}".format(
+            reply['headers'][0].block_number, reply['headers'][-1].block_number))
```
pipermerriam: Is it possible for `reply['headers']` to be empty?

evm/p2p/test_peer.py
```diff
@@ -20,23 +19,109 @@
...
+        self.les_proto.send_announce(
+            header.hash, header.block_number, total_difficulty, reorg_depth)
+
+    def handle_msg(self, cmd, msg):
+        if isinstance(cmd, GetBlockHeaders):
+            self.new_get_block_headers(msg)
+        else:
+            super(LESPeerServing, self).handle_msg(cmd, msg)
+
+    def new_get_block_headers(self, msg):
```
pipermerriam: What's the reason for calling this `new_get_block_headers` as opposed to just `get_block_headers`?

evm/p2p/test_peer.py
```diff
@@ -20,23 +19,109 @@
...
+        if query.reverse:
+            start = max(0, query.block - query.max_headers)
+            # Shift our range() limits by 1 because we want to include the requested block number
+            # in the list of block numbers.
+            block_numbers = reversed(range(start + 1, query.block + 1))
+        else:
+            end = min(self.head_number + 1, query.block + query.max_headers)
+            block_numbers = range(query.block, end)
+        for i in block_numbers:
+            headers.append(self.chaindb.get_canonical_block_header_by_number(i))
```
pipermerriam: Thoughts on changing this to:

```python
headers = tuple(
    self.chaindb.get_canonical_block_header_by_number(1)
    for i in block_number
)
```

Removes mutation and maintains a readability.  Would also be fine as a `map` call too.

evm/p2p/peer.py
```diff
@@ -129,6 +135,14 @@ def head_info(self):
             genesis_hash=genesis_header.hash,
         )
 
+    def handle_msg(self, cmd, msg):
+        """Peer-specific handling of incoming messages.
+
+        This method is called once the sub-protocol's process() method is done, so that custom
+        Peer implementations can appy extra processing to incoming messages.
+        """
+        pass
```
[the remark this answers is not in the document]
gsalgado: Because the subclass-specific extra processing is optional

evm/p2p/peer.py
gsalgado: Yeah, that's definitely better. Went with synced_block_height

evm/p2p/peer.py
gsalgado: Yeah, `head_number` is the head just announced by the remote, so it should be able to provide us all headers up to (inclusive) that

evm/p2p/peer.py
gsalgado: In theory, yes, but that would indicate either a bug in our code (`_sync()` should never call this with a `start_block` that would cause an empty list of headers in the response) or in the remote if it announces a head higher than its actual one. I can add a check and raise an exception here so that at least we don't get a cryptic KeyError if that happens

evm/p2p/test_peer.py
gsalgado: Because this is the handler for incoming LES/GetBlockHeaders messages, and I wanted the method name convey that, but it looks like I failed. Maybe `receive_get_block_headers()` would be better?

evm/p2p/peer.py
pipermerriam: I'd feel safer with that check in place.

evm/p2p/test_peer.py
pipermerriam: or `handle_get_block_headers_response`?  That sets up a clear naming convention for this type of thing that should work with other messages as well?

pipermerriam:
👍

evm/p2p/peer.py
gsalgado: I've added it already. Not super happy with the exception name, but maybe you can think of a better name?

evm/p2p/test_peer.py
gsalgado: Yeah, rather long but not something we'll have to type more than once so it makes sense. I'll change

gsalgado: But it actually should be `handle_get_block_headers` because it handles a GetBlockHeaders msg, which is a request, the response has a different msg type (BlockHeaders), so the handler for that would be `handle_block_headers`

PR #131, opened 2017-11-03
Some fixes to the Peer code

gsalgado:
I've been running a script that uses multiple Peers connected to different mainnet
remotes, all fetching headers and sharing a single chaindb, and that uncovered a
few issues which I'm fixing here.

### DIFF
--- a/evm/p2p/p2p_proto.py
+++ b/evm/p2p/p2p_proto.py
@@ -1,5 +1,7 @@
 import enum
 
+from cytoolz import assoc
+
 from rlp import sedes
 
 from evm.p2p.constants import (
@@ -54,12 +56,15 @@ class Disconnect(Command):
         except ValueError:
             return "unknown reason"
 
+    def decode(self, data):
+        raw_decoded = super(Disconnect, self).decode(data)
+        return assoc(raw_decoded, 'reason_name', self.get_reason_name(raw_decoded['reason']))
+
     def handle(self, proto, data):
         decoded = self.decode(data)
-        reason_name = self.get_reason_name(decoded['reason'])
         proto.logger.info(
-            "Peer {} disconnected; reason given: {}".format(proto.peer.remote, reason_name))
-        proto.peer.stop()
+            "{} disconnected; reason given: {}".format(proto.peer, decoded['reason_name']))
+        proto.peer.close()
         return decoded
 
 
--- a/evm/p2p/peer.py
+++ b/evm/p2p/peer.py
@@ -33,6 +33,7 @@ from evm.p2p.constants import (
 from evm.p2p.exceptions import (
     AuthenticationError,
     EmptyGetBlockHeadersReply,
+    PeerConnectionLost,
     PeerDisconnected,
     UnknownProtocolCommand,
     UnreachablePeer,
@@ -40,6 +41,7 @@ from evm.p2p.exceptions import (
 )
 from evm.p2p.utils import (
     gen_request_id,
+    get_devp2p_cmd_id,
     roundup_16,
     sxor,
 )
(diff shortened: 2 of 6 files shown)


pipermerriam:
Hard to separate from #130 so waiting for that to get merged before I dig into this.

gsalgado:
#130 has been merged and I've rebased this

evm/p2p/p2p_proto.py
```diff
@@ -56,10 +56,10 @@ def get_reason_name(self, reason_id):
 
     def handle(self, proto, data):
         decoded = self.decode(data)
-        reason_name = self.get_reason_name(decoded['reason'])
+        decoded['reason_name'] = self.get_reason_name(decoded['reason'])
```
pipermerriam: what do you think about just pulling the name out for the logging message, but not mutating the `decoded` value. 

Alternate solution would be to use `cytools.assoc` to do this in a non-mutating way

```
raw_decoded = self.decode(data)
decoded = assoc(raw_decoded, 'reason_name', self.get_reason_name(raw_decoded['reason']))
```

evm/p2p/peer.py
```diff
@@ -403,7 +418,10 @@ def sync(self, head_hash, head_number, reorg_depth=0):
                 # TODO: Implement a retry mechanism and only disconnect after retrying a few
                 # times.
                 self.logger.warn("Timeout when syncing, disconnecting from %s", self)
-                yield from self.stop()
+                yield from self.stop_and_wait_until_finished()
+            except Exception:
+                self.logger.error(
```
pipermerriam: Is there a reason this doesn't re-raise?

evm/p2p/utils.py
```diff
@@ -19,3 +21,12 @@ def roundup_16(x):
...
     return big_endian_to_int(os.urandom(8))
+
+
+def get_devp2p_cmd_id(msg):
+    """Return the cmd_id for the given devp2p msg.
+
+    The cmd_id, also known as the payload type, is always the first entry of the RLP, interpreted
+    as an integer.
+    """
+    return rlp.decode(msg[:1], sedes=rlp.sedes.big_endian_int)
```
pipermerriam: In the event that `msg == b''` this will raise a confusing error: `DecodingError: RLP string to short`

Does it make sense to special case that here or is there validation elsewhere for this.

evm/p2p/p2p_proto.py
gsalgado: I include it in the `decoded` dict because it is now used in `handshake()` so that it can include the disconnect reason in the exception it raises.

I don't understand what's the benefit of using cytools.assoc here?

evm/p2p/peer.py
gsalgado: Because this method is called with an asyncio.ensure_future(), so that exception would not be caught anywhere. This will change soon, though, as I'll implement a global (cross-peer) queue of sync requests to avoid fetching the same headers from multiple peers, and then those sync() calls will run within a loop where we'll be able to catch any exceptions they raise

evm/p2p/utils.py
gsalgado: Peer.read_msg() cannot return an empty bytes since it will always either read/decrypt a full, valid msg, or raise an exception.

evm/p2p/p2p_proto.py
pipermerriam: > I don't understand what's the benefit of using cytools.assoc here?

It's more about consistency in avoiding mutation.  In this case there is no direct benefit, but by sticking to the pattern it keeps the code consistent. Largely an ideology based recommendation.

An alternative approach would be to override `decode` on this object and add the inclusion of the `reason_name` there so that it's part of the decoding (which feels slightly more correct).

gsalgado: I'm fine with using cytooz.assoc, just wanted to understand the reasoning behind it. If we were to override `decode()` here you'd still prefer if we used assoc() there though?

gsalgado:
Cool, that's what I did before I merged

PR #134, opened 2017-11-08
Change from generator-based to native coroutines

gsalgado:
The generator-based coroutines have some limitations (e.g. not possible to yield values from them) and pitfalls that are not present on native coroutines. Since we don't intend to support python 3.4, there's no reason for us to use the generator-based approach


pipermerriam:
:shipit:

PR #136, opened 2017-11-09
BaseChainDB: do not persist a header if parent is unknown

gsalgado:


tests/database/test_chaindb.py
```diff
@@ -68,6 +76,12 @@ def test_persist_header_to_db(chaindb, header):
     assert chaindb.exists(number_to_hash_key)
 
 
+def test_persist_header_to_db_unknown_parent(chaindb, header):
+    header.parent_hash = keccak(os.urandom(32))
```
pipermerriam: I know it's silly but thoughts on replacing the randomness with a constant value here.  not really about the hash collision as much as just avoiding tests that rely on random data.

gsalgado: The randomness is intentional here, it ensures our code is not hard-coded to check for a specific value. [This post](https://www.julian-edwards.com/2017/01/16/random-inputs-in-unit-testing/) goes into more detail about that

PR #139, opened 2017-11-16
Implement PeerPool and use it in LightChain

gsalgado:


evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
+            self.logger.debug(
+                "Already connected to %s peers: %s; sleeping",
+                len(self.connected_nodes),
+                [remote for remote in self.connected_nodes])
+            return
+
+        for node in await self.get_nodes_to_connect():
+            self.logger.debug("Connecting to %s", node)
+            peer = await self.connect(node)
+            if peer is not None:
```
pipermerriam: This is a pattern I really don't like (checking return values).  I suspect it may be this way due to `self.connect(...)` being something that get's `await`-ed.  Is it possible for this to instead use a try/except pattern, or some other pattern that removes the return value checking?

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
+                self.logger.info("Successfully connected to %s", peer)
+                asyncio.ensure_future(peer.start(finished_callbabck=self.peer_finished))
+                self.connected_nodes[peer.remote] = peer
+
+    def peer_finished(self, peer):
+        self.logger.info("Peer finished: %s", peer)
+        if peer.remote in self.connected_nodes:
```
pipermerriam: This code has me confused.  Here we pop the peer out of the `connected_nodes` but above on line 126, we immediately place it into the `connected_nodes` once this finishes.  Not intuitive to me why this code does what it does.  Can you maybe add a comment?

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
+        await self.peer_pool.stop()
 
     async def get_canonical_block_by_number(self, block_number):
         try:
             block_hash = self.chaindb.lookup_block_hash(block_number)
         except KeyError:
-            raise BlockNotFound("No block with number {} found on canonical chain".format(
-                block_number))
+            raise BlockNotFound(
+                "No block with number %s found on local chain", block_number)
```
pipermerriam: I think this will raise an awkwardly formatted exception message.

```python
In [13]: raise BlockNotFound('No block with num % found', 3)
---------------------------------------------------------------------------
BlockNotFound                             Traceback (most recent call last)
<ipython-input-13-4f95849f050c> in <module>()
----> 1 raise BlockNotFound('No block with num % found', 3)

BlockNotFound: ('No block with num % found', 3)
```

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
 
+    @alru_cache(maxsize=1024)
     async def get_block_by_hash(self, block_hash):
+        peer = await self.peer_pool.get_best_peer()
+        self.logger.debug("Fetching block %s from %s", encode_hex(block_hash), peer)
+        request_id = gen_request_id()
+        peer.les_proto.send_get_block_bodies([block_hash], request_id)
+        reply = await peer.wait_for_reply(request_id)
+        if len(reply['bodies']) == 0:
+            raise BlockNotFound("No block with hash %s found", block_hash)
```
pipermerriam: Same here with formatting of the exception message.

evm/p2p/peer.py
```diff
@@ -193,14 +193,16 @@ def get_protocol_for(self, cmd_id):
             raise PeerConnectionLost()
         return data
 
-    async def start(self):
+    async def start(self, finished_callbabck=None):
```
pipermerriam: s/callbabck/callback

evm/p2p/peer.py
```diff
@@ -193,14 +193,16 @@ def get_protocol_for(self, cmd_id):
...
         try:
             await self.read_loop()
         except Exception as e:
             self.logger.error(
                 "Unexpected error when handling remote msg: {}".format(traceback.format_exc()))
         finally:
             self._finished.set()
+            if finished_callbabck is not None:
+                finished_callbabck(self)
```
pipermerriam: same spelling issue

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
+        """
+        while len(self.connected_nodes) == 0:
+            await asyncio.sleep(0.5)
+        peers = list(self.connected_nodes.values())
+        best = peers[0]
+        for peer in peers[1:]:
+            if peer.synced_block_height is None:
+                continue
+            if peer.synced_block_height > best.synced_block_height:
+                best = peer
```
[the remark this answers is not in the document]
carver: Or maybe:

```
def peer_block_height(peer):
    return peer.synced_block_height or -1

best = max(peers, key=peer_block_height)
```

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
         except KeyError:
-            raise BlockNotFound("No block with number {} found on canonical chain".format(
-                block_number))
+            raise BlockNotFound(
+                "No block with number %s found on local chain", block_number)
         return await self.get_block_by_hash(block_hash)
 
+    @alru_cache(maxsize=1024)
     async def get_block_by_hash(self, block_hash):
+        peer = await self.peer_pool.get_best_peer()
```
carver: As I read it, if we issue a bunch of requests and all of our peers are the same height, we would hammer the peer at the beginning of the list with all of the requests.

Am I reading it right? If so, maybe it's worth adding a tracking issue to spread the load across the peers.

pipermerriam: That suggests to me that the strategy for choosing the *best* peer should pick randomly from the set of peers who have the highest block height.

evm/p2p/lightchain.py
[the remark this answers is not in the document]
pipermerriam: I didn't know `max` took a `key` argument!

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
+                self.logger.error("Unexpected error (%s), restarting", traceback.format_exc())
+                await self.stop_all_peers()
+            # Wait self._connect_loop_sleep seconds, unless we're asked to stop.
+            await asyncio.wait([self._should_stop.wait()], timeout=self._connect_loop_sleep)
+        self._finished.set()
+
+    async def stop_all_peers(self):
+        self.logger.info("Stopping all peers ...")
+        await asyncio.gather(
+            *[peer.stop_and_wait_until_finished() for peer in self.connected_nodes.values()])
```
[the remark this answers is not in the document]
gsalgado: Well, we'll always iterate over all items and the number of connected nodes should always be relatively small

evm/p2p/lightchain.py
gsalgado: When connecting to a peer there are various possible exceptions that can be raised, and I decided to handle them all in connect() to keep things simple here. I could change connect() to raise something else, though? Any name suggestions for such a new exception?

evm/p2p/lightchain.py
gsalgado: This method is passed as the finished_callback arg to peer.start(). It is only called when the peer disconnects and peer.start() returns. I'd be happy to add a comment but not sure what it should say

evm/p2p/lightchain.py
gsalgado: Oops, I meant to change the logging calls to no longer use .format(), but not this. Reverting

evm/p2p/lightchain.py
gsalgado: Reverted as well

evm/p2p/lightchain.py
gsalgado: Yeah, makes sense. Since at this point it is used only for on-demand requests (and we only fetch blocks on demand), would it be ok if I add a TODO comment to address that later?

evm/p2p/lightchain.py
```diff
@@ -1,69 +1,270 @@
...
+        self.connected_nodes = {}
+        self._should_stop = asyncio.Event()
+        self._finished = asyncio.Event()
+
+    async def get_best_peer(self):
+        """
+        Return the peer with the highest announced block height.
+        """
+        while len(self.connected_nodes) == 0:
+            await asyncio.sleep(0.5)
```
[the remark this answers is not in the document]
gsalgado: That's quite an elegant solution and I really liked it at first, but upon thinking a bit more about it I came to realize config files are probably a better place (albeit not as elegant) to store sleep/timeout values, no? Those are the kind of thing devops will usually tweak, so having them in config files has several advantages: less chance of human error when changing, can be maintained in a separate repos, etc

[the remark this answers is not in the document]
pipermerriam: +1 simple yet expressive.  How do you feel about enabling all of that configuration through environment variables?

evm/p2p/lightchain.py
pipermerriam: I'd suggest establishing a common base exception, something like `PeerConnectionError` that all of the exceptions inherit from.  For the one `asyncio.TimeoutError` we can catch and re-raise our own exception (maybe `PeerTimeoutError`).

evm/p2p/lightchain.py
pipermerriam: > would it be ok if I add a TODO comment to address that later?

Yeah, that's fine.  Maybe even write up an issue for this and while I don't think it's an easy-pickings level thing, it's something that someone could come in and implement it without much base knowledge of the codebase.

evm/p2p/lightchain.py
pipermerriam: Maybe making it internal `_peer_finished()` makes it clearer that it's not meant to be useful on its own.

As for my confusion, I think I understand now.  This is *queued* to be run later.  Line 125 puts the remote in the `connected_nodes` list and then this removes it later when it's finished.  Maybe if line 124 and 125 were swapped this would be more intuitive?

pipermerriam:
:shipit:  goood to go at your discretion.

evm/p2p/lightchain.py
gsalgado: Sure, I've done that and also added a docstring to `_peer_finished()`

evm/p2p/lightchain.py
gsalgado: One thing to keep in mind is that we'll want to handle each of those exceptions differently; for example, a UselessPeer should probably not be retried any time soon, whereas a TimeoutError should. That's why I wanted to keep the handling of them in `connect()`. We could have `connect()` raise a `PeerConnectionError`, but that wouldn't feel right in the case of a `UselessPeer` (i.e. a peer with no matching sub-protocols)

pipermerriam: I'm comfortable with just adding a `# TODO` comment here to explore ways to remove return-value-checking from this API.

PR #147, opened 2017-11-16
Reorganize precompiles

pipermerriam:
### What was wrong?

All of the precompile functions were co-located in the same module under `evm/precompiles.py`.  Given a growing number of these, a more granular organization strategy was needed.

### How was it fixed?

Spit the precompiles out into individual modules.

#### Cute Animal Picture

tucker-the-pembroke-welsh-corgi_45257_2010-10-30_w450


evm/vm/forks/frontier/__init__.py
```diff
@@ -40,6 +39,14 @@
 )
 
 
+FRONTIER_PRECOMPILES = {
+    force_bytes_to_address(b'\x01'): precompiles.precompile_ecrecover,
+    force_bytes_to_address(b'\x02'): precompiles.precompile_sha256,
+    force_bytes_to_address(b'\x03'): precompiles.precompile_ripemd160,
+    force_bytes_to_address(b'\x04'): precompiles.precompile_identity,
```
gsalgado: What do you think about removing the 'precompile_' suffix from the function names, to avoid the stutter when using them?

PR #151, opened 2017-11-16
Move get_opcode_fn to Computation object

pipermerriam:
### What was wrong?

The `get_opcode_fn` was located on the VM class when it was better encapsulated by the `Computation` object.

### How was it fixed?

Moved `get_opcode_fn` to be a `Computation` object API.

#### Cute Animal Picture

5f2de153220780ba060741ef107a1c34

### DIFF
--- a/evm/vm/computation.py
+++ b/evm/vm/computation.py
@@ -8,6 +8,9 @@ from evm.constants import (
 from evm.exceptions import (
     VMError,
 )
+from evm.logic.invalid import (
+    InvalidOpcode,
+)
 from evm.validation import (
     validate_canonical_address,
     validate_uint256,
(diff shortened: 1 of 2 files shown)


evm/vm/computation.py
```diff
@@ -188,6 +194,12 @@ def add_log_entry(self, account, topics, data):
         validate_is_bytes(data, title="Log entry data")
         self.log_entries.append((account, topics, data))
 
+    def get_opcode_fn(self, opcode):
+        try:
+            return self.vm.opcodes[opcode]
```
gsalgado: I'm curious as to why you think it's better encapsulated in the Computation class since it has to look up the opcode by accessing a vm attribute.

pipermerriam: Now that you point it out, I'm not sure why I did this.  Going to close this issue.

PR #152, opened 2017-11-16
Refactory BaseVM initialization handling of chaindb

pipermerriam:
### What was wrong?

During initialization of the `BaseVM` objects there was a check for whether the `chaindb` was `None`.  This was a carry-over from a previous version where the db was part of the class definition.

### How was it fixed?

Removed the check.

#### Cute Animal Picture

395166785942e7f4098fb27b5dc8b3d2

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -35,18 +35,15 @@ class VM(object):
     individual VM classes for each fork of the protocol rules within that
     network.
     """
+    chaindb = None
     opcodes = None
     _block_class = None
     _precompiles = None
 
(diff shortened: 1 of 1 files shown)


carver:
I don't follow, is `chaindb` allowed to be `None` now?

pipermerriam:
A while back the signature for the VM class was:

```
def __init__(self, header, chaindb=None):
```

The `chaindb` was *sometimes* already present as a class property so there were cases where you didn't instantiate the `VM` class with a database because it already had one.  That's no longer a thing but this code is still hanging around.

PR #161, opened 2017-11-17
[WIP] Write initiation for Package class and tests

njgheorghita:
@pipermerriam Not sure if this is what you had in mind for the Package class, but I thought I'd take a stab at it. Feedback would be great. Especially on what to name the attributes of the `Package` class. And where/if to store the json samples.

### What was wrong?
No package class

### How was it fixed?
Wrote a package class

#### Cute Animal Picture
giphy 24


carver:
@njgheorghita I assume this PR was supposed to be opened against https://github.com/pipermerriam/ethereum-erc190 instead of py-evm...

njgheorghita:
@carver Oooh, yeah you're definitely right. I'll close this pr.

PR #162, opened 2017-11-18
Use MinGasLimit 5000 like other clients

nicksavers:
### What was wrong?

The minimum gas limit in the yellowpaper is `125000`, but all other clients have implemented `5000`. Either all clients should be updated, or py-evm could change the constant. 
Here is the original issue to report on this https://github.com/ethereum/yellowpaper/issues/317

### How was it fixed?

By changing the `GAS_LIMIT_MINIMUM` in `constants.py` to `5000`

### Note

Might be best to discuss this with other core devs before merging.


carver:
Merged in #222

Issue #132, opened 2017-11-06
Add type hints for all packages


gsalgado:
[Type hints](https://www.python.org/dev/peps/pep-0484/) allow us to perform static type checking, among other things. We could benefit a lot from using them.

[This stackoverflow answer](https://stackoverflow.com/questions/32557920/what-are-type-hints-in-python-3-5) does a great job at describing their main benefits.

[pyannotate](https://github.com/dropbox/pyannotate) would come in handy if it had support for py3-style annotations, but meanwhile we can add them manually

[monkeytype](https://engineering.instagram.com/let-your-code-type-hint-itself-introducing-open-source-monkeytype-a855c7284881) may actually be more useful than pyannotate as it generates py3-style annotations

### Definition of done

- type hints for all of `evm`, `p2p`, and `trinity` modules.
- individual `mypy` CI runs for each of these modules using `mypy --follow-imports=silent --ignore-missing-imports --check-untyped-defs --disallow-incomplete-defs -p <module>`

This should not be done as a single pull request, but rather many small ones to iteratively add type hints, slowly expanding the `mypy` coverage as hints are added.

pipermerriam:
Yes, this looks quite solid.  I'm supportive of adopting this pretty much across the board.

gitcoinbot:
__This issue now has a funding of 0.8 ETH (628.62 USD @ $785.78/ETH) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/132).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/132)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $7468.87 more Funded OSS Work Available at: https://gitcoin.co/explorer

gitcoinbot:
__Work has been started on the 0.8 ETH (628.62 USD @ $785.78/ETH) funding by__: 
 1. [@cburgdorf](https://gitcoin.co/profile/cburgdorf) 

 __Please work together__ and coordinate delivery of the issue scope. Gitcoin doesn't know enough about everyones skillsets / free time to say who should work on what, but we trust that the community is smart and well-intentioned enough to work together.  As a general rule; if you start work first, youll be at the top of the above list ^^, and should have 'dibs' as long as you follow through. 

[...]

PR #167, opened 2017-11-22
Add type annotations for evm.p2p

gsalgado:
Part of #132


evm/p2p/auth.py
```diff
@@ -170,18 +182,18 @@ def create_auth_ack_message(self, nonce):
             msg = self.ephemeral_pubkey.to_bytes() + nonce + token_flag
         return msg
 
-    def encrypt_auth_ack_message(self, ack_message):
+    def encrypt_auth_ack_message(self, ack_message: bytes) -> bytes:
         if self.got_eip8_auth:
             # The EIP-8 version has an authenticated length prefix.
-            prefix = struct.pack('>H', len(ack_message) + ecies.encrypt_overhead_length)
+            prefix = struct.pack('>H', len(ack_message) + ENCRYPT_OVERHEAD_LENGTH)
```
gsalgado: This was the first bug caught by mypy. It didn't show up when running the light client because by default we don't use the EIP8 handshake

pipermerriam: That's subtly really awesome

PR #168, opened 2017-11-23
More type annotations for evm.p2p

gsalgado:


evm/p2p/discovery.py
```diff
@@ -8,6 +8,7 @@
 import asyncio
 import logging
 import time
+from typing import Any, AnyStr, Callable, Generator, List, Tuple
```
pipermerriam: *nitpick*

My preference for imports is to do one per line:

```
from cat_in_the_hat import (
    thing_1,
    thing_2,
)

File: evm/p2p/discovery.py [hunk 2]
```diff
@@ -111,21 +115,23 @@ def connection_made(self, transport):
         self.logger.debug("boostrapping with {}".format(self.bootstrap_nodes))
         await self.kademlia.bootstrap(self.bootstrap_nodes)
 
-    def datagram_received(self, data, addr):
+    # FIXME: Enable type checking here once we have a mypy version that
+    # includes the fix for https://github.com/python/typeshed/pull/17400
```
pipermerriam: This URL returns a 404.

evm/p2p/kademlia.py
```diff
@@ -8,19 +8,30 @@
...
+    keys,
+)
 
 from evm.utils.keccak import keccak
 from evm.utils.numeric import big_endian_to_int
 
 
+# Workaround for import cycles caused by type annotations:
+# http://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles
+MYPY = False
```
pipermerriam: Can we use [`typing.TYPE_CHECKING`](https://docs.python.org/3.5/library/typing.html) here instead?

evm/p2p/kademlia.py
```diff
@@ -149,19 +158,19 @@ def split(self):
             bucket.replacement_cache.append(node)
         return lower, upper
 
-    def remove_node(self, node):
+    def remove_node(self, node: Node) -> None:
         if node not in self.nodes:
```
pipermerriam: Just a note that this should probably be `if node not in self` since we do define `__contains__` on this class.

evm/p2p/kademlia.py
```diff
@@ -111,32 +120,32 @@ def __hash__(self):
 
 
 @total_ordering
-class KBucket:
+class KBucket(Sized):
```
pipermerriam: *nitpick*

This also qualifies to be marked as `collections.abc.Container`.  If we were to add an `__iter__` we could just use `collections.abc.Set` which I think is actually appropriate.  We have at least one place that I could find where we iterate over the nodes in a bucket.

evm/p2p/kademlia.py
```diff
@@ -182,49 +191,50 @@ def add(self, node):
...
         return len(self.nodes)
 
-    def __lt__(self, other):
+    def __lt__(self, other: 'KBucket') -> bool:
         if not isinstance(other, self.__class__):
             raise TypeError("Cannot compare KBucket with type {}.".format(other.__class__))
         return self.end < other.start
 
 
 class RoutingTable:
```
pipermerriam: For consistency, do you want to add a `collections.abc` base class to this as well?

evm/p2p/lightchain.py
```diff
@@ -93,7 +100,7 @@ def peer_block_height(peer):
             peer = await asyncio.wait_for(
                 handshake(remote, self.privkey, self.peer_class, self.chaindb, self.network_id),
                 HANDSHAKE_TIMEOUT)
-            return peer
+            return cast(LESPeer, peer)
```
pipermerriam: Is the `cast` call here because the `handshake` method takes the `peer_class` and since it's dynamic we have to hint at what type this is?

evm/p2p/peer.py
```diff
@@ -109,10 +117,13 @@ class BasePeer:
     # FIXME: Must be configurable.
     listen_port = 30303
 
-    def __init__(self, remote, privkey, reader, writer, aes_secret, mac_secret,
-                 egress_mac, ingress_mac, chaindb, network_id):
+    def __init__(self, remote: Node, privkey: datatypes.PrivateKey, reader: asyncio.StreamReader,
```
pipermerriam: This is really hard to read, can we do one per line for these arguments?

evm/p2p/protocol.py
```diff
@@ -85,27 +95,28 @@ class Protocol:
     logger = logging.getLogger("evm.p2p.protocol.Protocol")
     name = None  # type: bytes
     version = None  # type: int
+    cmd_length = None  # type: int
     handshake_msg_type = None  # type: Type[Command]
     # List of Command classes that this protocol supports.
     _commands = []  # type: List[Type[Command]]
 
-    def __init__(self, peer, cmd_id_offset):
+    def __init__(self, peer: 'BasePeer', cmd_id_offset: int) -> None:
```
pipermerriam: Is `'BasePeer'` a string due to circular imports?

evm/p2p/discovery.py
gsalgado: Oops, there was an extra trailing 0 there

evm/p2p/kademlia.py
gsalgado: Oh, yeah, I had to use this in eth-keys because apparently there's no TYPE_CHECKING in py3.4, but here it should be ok. Will change

evm/p2p/kademlia.py
gsalgado: I wanted to use `collections.abc.Collection` but that was only added in 3.6, so used `Iterable` and  `Container`

evm/p2p/lightchain.py
gsalgado: Yep, although here I could have the `connect()` method returning a `BasePeer` as well and avoid the `cast`, but I think at this point it makes more sense to do this

evm/p2p/peer.py
gsalgado: Sure!

evm/p2p/protocol.py
gsalgado: Yep

evm/p2p/kademlia.py
gsalgado: So, it turns out that if I use a `collections.abc` base class here I also need to change the type annotations to match those of that base class: i.e. in `__contains__()` I need to use `object` instead of `Node` as the argument's type. I think I prefer to use the appropriate type, so not adding the base class here

PR #170, opened 2017-11-27
Organize constants

### LINKED ISSUE #163: Move fork and chain specific constants into their individual modules
### What is wrong?

The `evm.constants` module has a number of values that are *fork specific* such as `FRONTIER_DIFFICULTY_ADJUSTMENT`.  Having these in the `evm.constants` module feels *wrong* since the core Py-EVM library is supposed to be largely agnostic about specific fork rules.

### How can it be fixed

These variables would be more appropriately located if they were located under the fork module that they apply to.  Move each of the fork specific and chain specific constants into a new module at  `evm.vm.forks.<fork-name>.constants` or `evm.chains.mainnet.constants`.

ethanbennett:
### What was wrong?

There were several chain- and fork-specific constants in the `evm.constants` module.

### How was it fixed?

This PR creates `constants` modules for each fork that needs one, and it moves the existing constants accordingly.

#### Cute Animal Picture

Close #163


ethanbennett:
Travis is failing because of some `flake8` complaints that seem a bit strange (not connected to the changes in this PR). I'll have it passing soon.

evm/chains/mainnet/__init__.py
```diff
@@ -1,8 +1,20 @@
...
+    HOMESTEAD_MAINNET_BLOCK
+)
+from evm.vm.forks.eip150.constants import (
+    EIP150_MAINNET_BLOCK
+)
+from evm.vm.forks.spurious_dragon.constants import (
+    SPURIOUS_DRAGON_MAINNET_BLOCK
+)
+from evm.vm.forks.byzantium.constants import (
+    BYZANTIUM_MAINNET_BLOCK
```
pipermerriam: I think it might be good to just move these constants to a module here (`evm.chains.mainnet.constants`) since they are actually more specific to the mainnet chain than the forks themselves.

evm/vm/forks/eip160/constants.py
```diff
@@ -0,0 +1,3 @@
+# https://github.com/ethereum/EIPs/issues/160
+GAS_EXP_EIP160 = 10
+GAS_EXPBYTE_EIP160 = 50
```
pipermerriam: instead of `forks.eip160` this should be in `forks.byzantium` (since this is actually part of byzantium's fork rules)

evm/vm/forks/eip170/constants.py
```diff
@@ -0,0 +1,2 @@
+# https://github.com/ethereum/EIPs/issues/170
+EIP170_CODE_SIZE_LIMIT = 24577
```
pipermerriam: instead of `forks.eip160` this should be in `forks.byzantium` (since this is actually part of byzantium's fork rules)

evm/vm/forks/eip150/opcodes.py
```diff
@@ -2,7 +2,7 @@
 
 from cytoolz import merge
 
-from evm import constants
+from evm.vm.forks.eip150 import constants
```
pipermerriam: note to self: we should rename `forks.eip150` to be `forks.tangerine` since that's the official name of that fork.  Not something that should be done in this pull request.

evm/vm/forks/frontier/headers.py
```diff
@@ -4,14 +4,18 @@
...
+
 from evm.constants import (
     GENESIS_GAS_LIMIT,
     DIFFICULTY_ADJUSTMENT_DENOMINATOR,
     DIFFICULTY_MINIMUM,
     BOMB_EXPONENTIAL_PERIOD,
     BOMB_EXPONENTIAL_FREE_PERIODS,
-    FRONTIER_DIFFICULTY_ADJUSTMENT_CUTOFF,
 )
+from evm.vm.forks.frontier.constants import (
```
pipermerriam: I think for these imports within the `frontier` module importing another module from within the `frontier` module it may be more legible to use a relative import:

```
from .constants import (
    ...
)
```

Same applies across the other forks when they import from their own `constants` module.

evm/vm/forks/eip160/constants.py
ethanbennett: @pipermerriam When I try to consolidate the constants for `eip160` and `eip170` in `byzantium.constants`, it causes the `SpuriousDragonVM` to throw an error. I suspect it's a circular dependency between SD and Byzantium, which both call each other in `__init__`. What if we did a `forks.constants` module and put all the fork-specific constants there?

evm/chains/mainnet/__init__.py
ethanbennett: Good call. I created one for Ropsten with the Byzantium block, too, since it was in `evm.constants`, but it doesn't seem to be used in the project. Let me know if you want me to exclude it.

evm/vm/forks/eip170/constants.py
ethanbennett: The above comment applies to this change as well.

evm/vm/forks/eip160/constants.py
pipermerriam: This is because I'm dumb and told you to move it to byzantium when both this and the eip170 are spurious dragon.  I just tested and if you move them to `forks.spurious_dragon.constants` everything works fine.

ethanbennett: Haha ok, awesome. I didn't even think to check that, I just launched right into the python/rabbit hole.

pipermerriam: The clue for me was that an old fork was importing from a newer fork and that should really *never ever* happen.

ethanbennett: Nice, I'll keep an eye out for that in the future.

evm/p2p/lightchain.py
```diff
@@ -373,7 +367,6 @@ def peer_block_height(peer: LESPeer):
 
     async def get_canonical_block_by_number(self, block_number: int) -> BaseBlock:
         """Return the block with the given number from the canonical chain.
-
```
pipermerriam: Is this intentional or is your editor striping these blank lines.  

I think this is reducing readability of these docstrings.  Can you re-add them.

evm/vm/forks/byzantium/__init__.py
```diff
@@ -2,7 +2,8 @@
     merge,
 )
 
-from evm import constants
+from evm.constants import MAX_UNCLE_DEPTH
+from .constants import EIP649_BLOCK_REWARD
```
pipermerriam: Can you go take a look at [what the style guide has to say about import ordering](https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#import-statements).

These `.constants` imports should be further down below all of the `evm.` level imports since they are *local*.

evm/vm/forks/spurious_dragon/__init__.py
```diff
@@ -52,17 +53,17 @@ def _apply_spurious_dragon_create_message(vm, message):
     else:
         contract_code = computation.output
 
-        if contract_code and len(contract_code) >= constants.EIP170_CODE_SIZE_LIMIT:
-            computation._error = OutOfGas(
+        if contract_code and len(contract_code) >= EIP170_CODE_SIZE_LIMIT:
+            computation.error = OutOfGas(
```
pipermerriam: This change should be reverted.

evm/p2p/lightchain.py
ethanbennett: No, this wasn't intentional. I'll re-add them.

evm/vm/forks/spurious_dragon/__init__.py
ethanbennett: This definitely fixed it! Glad you noticed it--I must've done this somehow by mistake when changing the constant reference on the line above.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "python.pythonPath": "${workspaceRoot}/venv/bin/python"
+}
```
pipermerriam: tisk tisk: `git add .`

ethanbennett: Haha. The git problems I caused for myself on this branch are beyond `add .` unfortunately, but this won't be in the final version.

ethanbennett:
@pipermerriam This should be good to go now.

PR #172, opened 2017-11-28
[WIP] Implementation of collation

jannikluhn:


evm/rlp/transactions.py
```diff
@@ -16,6 +16,21 @@
 )
 
 
+class BaseAATransaction(rlp.Serializable):
```
pipermerriam: What's `AA` stand for here?

evm/vm/sharding/collations.py
```diff
@@ -0,0 +1,100 @@
...
+            header=header,
+            transactions=transactions
+        )
+
+    def validate(self):
+        if not self.chaindb.exists(self.header.state_root):
+            raise ValidationError(
+                "`state_root` was not found in the db.\n"
+                "- state_root: {0}".format(
+                    self.header.state_root,
```
pipermerriam: Probably good to go ahead and hex encode this so that the exception is legible.

PR #175, opened 2017-11-28
skip slowest tests during ci, run them in cron

pipermerriam:
### What was wrong?

There are some JSON fixture tests which take a *long* time to run.  This slows down the turnaround time in ci

### How was it fixed?

Skipped them during normal CI runs.  They will run in a cron job daily on travis.

#### Cute Animal Picture

dogs_wallpapers_267

### DIFF
--- a/tests/json-fixtures/test_state.py
+++ b/tests/json-fixtures/test_state.py
@@ -57,11 +57,79 @@ def expand_fixtures_forks(all_fixtures):
                 yield fixture_path, fixture_key, fixture_fork, post_state_index
 
 
+# These are the slowest 50 tests from the full statetest run.  This list should
+# be regenerated occasionally using `--durations 50`.
+SLOWEST_TESTS = {
+    ('stStaticCall/static_Call50000_sha256.json', 'static_Call50000_sha256', 'Byzantium', 0),
+    ('stStaticCall/static_Call50000_rip160.json', 'static_Call50000_rip160', 'Byzantium', 0),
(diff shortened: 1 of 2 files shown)


pipermerriam:
Travis is configured to run a daily cron job.  The test suite is configured to run the slow tests *if* the build is happening as part of cron using the environment variable that travis provides for this.

tests/json-fixtures/test_state.py
```diff
@@ -57,11 +57,79 @@ def expand_fixtures_forks(all_fixtures):
...
+    ('stSpecialTest/failed_tx_xcf416c53.json', 'failed_tx_xcf416c53', 'Byzantium', 0),
+    ('stRandom/randomStatetest159.json', 'randomStatetest159', 'Byzantium', 0),
+    ('stRandom/randomStatetest554.json', 'randomStatetest554', 'Byzantium', 0),
+}
+
+
+def should_run_slow_tests():
+    if os.environ.get('TRAVIS_EVENT_TYPE') == 'cron':
+        return True
+    return False
```
gsalgado: Maybe `return os.environ.get('TRAVIS_EVENT_TYPE') == 'cron'`?

pipermerriam: If you don't object, I'm going to leave it this way because I suspect we'll end up adding another mechanism for including these tests for things like local runs.  I intended to add it in this PR but clearly didn't.

Issue #164, opened 2017-11-20
Add new gas and return data API to Computation object.


pipermerriam:
### What is wrong?

With the introduction of the `REVERT` opcode in the byzantium fork rules we can no longer assume that the presence of `computation.error` means that gas has been fully depleted or that there is no return data.

This has resulted in the following lines of code being repeated in various places.

* https://github.com/pipermerriam/py-evm/blob/piper/byzantium/evm/vm/computation.py#L247
* https://github.com/pipermerriam/py-evm/blob/piper/byzantium/evm/vm/computation.py#L256
* https://github.com/pipermerriam/py-evm/blob/piper/byzantium/evm/logic/call.py#L114
* https://github.com/pipermerriam/py-evm/blob/piper/byzantium/evm/logic/call.py#L122
* https://github.com/pipermerriam/py-evm/blob/piper/byzantium/evm/vm/computation.py#L165

Some of this logic repetition would be better served by a dedicated API on the `Computation` object.

### How can it be fixed

Add the following two new APIs to the `Computation` object.

```python
def.should_burn_gas():
    return self.error and self.error.burns_gas

def should_erase_return_data():
    return self.error and self.error.zeros_return_data
```

*if* you would also like to cleanup the slighly confusing name `zeros_return_data` and change it to `erases_return_data` go for it.

carver:
I had to read this a few places before I realized `zero` was a verb here. Maybe `should_erase_return_data`?

frankcash:
The reference lines in the description are broken

pipermerriam:
fixed in #181

PR #181, opened 2017-12-01
[Issue #164] gas and return data helpers for Computation

djrtwo:
### What was wrong?
Issue #164 

### How was it fixed?
- added `should_burn_gas` to Computation and used throughout where applicable
- added `should_erase_return_data` to Computation and used throughout where applicable
- change `VMError` and `Revert` attribute `zeros_return_data`  to `erases_return_data`

#### Cute Animal Picture

### DIFF
--- a/evm/vm/computation.py
+++ b/evm/vm/computation.py
@@ -110,7 +110,7 @@ class Computation(object):
 
     @property
     def should_erase_return_data(self):
-        return self.error and self.error.zeros_return_data
+        return self.error and self.error.erases_return_data
 
     #
     # Execution
@@ -178,7 +178,7 @@ class Computation(object):
     #
     @property
     def output(self):
-        if self.error and self.error.zeros_return_data:
+        if self.should_erase_return_data:
             return b''
         else:
             return self._output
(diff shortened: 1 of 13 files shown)


evm/vm/computation.py
```diff
@@ -104,6 +104,14 @@ def is_success(self):
     def is_error(self):
         return not self.is_success
 
+    @property
+    def should_burn_gas(self):
+        return self.error and self.error.burns_gas
```
pipermerriam: Can you adjust this to use the `is_error` property and move `self.error` to be private at `self._error`.  Realizing that as part of this work, `self._error` should be clearly marked private to discourage using it.

pipermerriam:
I suspect you've seen it but: https://travis-ci.org/ethereum/py-evm/jobs/310112334

Some references to `zeros_return_data` still hanging around.

evm/vm/computation.py
djrtwo: on it

evm/vm/forks/frontier/__init__.py
```diff
@@ -118,7 +118,7 @@ def _execute_frontier_transaction(vm, transaction):
             # The address of the newly created contract has *somehow* collided
             # with an existing contract address.
             computation = Computation(vm, message)
-            computation.error = ContractCreationCollision(
+            computation._error = ContractCreationCollision(
```
djrtwo: explicit external setting of `_error` occurs in these fork `__init__.py` files. Violates the privateness of `_error`. Thoughts @pipermerriam?

pipermerriam: 🏈   <-- punt on it and mark it with a `# TODO` comment.

PR #182, opened 2017-12-01
Move header-syncing code from LESPeer to LightChain

gsalgado:
This allows us to avoid unnecessarily fetching the same headers from
all peers we're connected to.

Now we also retry a few times if we get a timeout when fetching headers
from a peer.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -18,12 +18,12 @@ Installation
 Syncing with Mainnet
 ---------------------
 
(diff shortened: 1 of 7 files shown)


gsalgado:
Various tests in test_peer.py can probably be simplified and should definitely be moved to a test_lightchain.py, but I'll do that in a separate branch as there are too many changes here already

gsalgado:
Yeah, makes sense. I'll have a go at it!

Issue #178, opened 2017-12-01
JSON-RPC server


pipermerriam:
We need to implement the basic API for JSON-RPC server.  A simplistic start can be found here: https://github.com/ethereum/py-evm/tree/master/evm/rpc

Lets flesh this out further to something more complete.  For now lets keep it as something that must be run as `python evm/rpc/server.py ....`

I suggest sticking to the *read-only* apis for now.

---
Compiled to-do list:

- [ ] support all the `eth_*` read-only methods
  - [x] `eth_protocolVersion`
  - [ ] `eth_syncing`
  - [ ] `eth_coinbase`
  - [ ] `eth_mining`
  - [ ] `eth_hashrate`
  - [ ] `eth_gasPrice`
  - [ ] `eth_accounts`
  - [x] `eth_blockNumber`
  - [x] `eth_getBalance`
  - [x] `eth_getStorageAt`
  - [x] `eth_getTransactionCount`
  - [x] `eth_getBlockTransactionCountByHash`
  - [x] `eth_getBlockTransactionCountByNumber`
  - [x] `eth_getUncleCountByBlockHash`
  - [x] `eth_getUncleCountByBlockNumber` block int
  - [x] `eth_getUncleCountByBlockNumber` block string
  - [x] `eth_getCode`
  - [ ] `eth_call`
  - [ ] `eth_estimateGas`
  - [x] `eth_getBlockByHash`
  - [x] `eth_getBlockByNumber` block int
  - [x] `eth_getBlockByNumber` block string
  - [ ] `eth_getTransactionByHash`
  - [x] `eth_getTransactionByBlockHashAndIndex`
  - [x] `eth_getTransactionByBlockNumberAndIndex`
  - [ ] `eth_getTransactionReceipt`
  - [x] `eth_getUncleByBlockHashAndIndex`
  - [x] `eth_getUncleByBlockNumberAndIndex` block int
  - [x] `eth_getUncleByBlockNumberAndIndex` block string
  - [ ] `eth_getCompilers`
  - [ ] `eth_getLogs`
  - [ ] `eth_getWork`
- [x] serve over socket
- [ ] make score available on block object, so that a returned block dict can be formatted without a `chain` reference ?

To save for another PR:

- [ ] server over HTTP

carver:
Reasonable approach: only serve json-rpc over sockets, plus an HTTP proxy like: https://github.com/chfast/json-rpc-proxy/blob/master/jsonrpcproxy.py

[...]

PR #183, opened 2017-12-02
Basic transport-agnostic RPC server

carver:
### What was wrong?

#178 JSON-RPC server

### How was it fixed?

It's not yet. Just a heads up that I'm starting down the path, and seeking early design feedback.

So far, this doesn't actually serve over any transport at all, it just accepts json strings and returns them.

It only supports `eth_getBlockByHash` at the moment, with a single happy-path test against from a pre-generated chain database.

TODO:

- [x] ~copy DB to ignored directory before loading in test~ use json-encoded MemoryDB fixture
- [x] move `evm.rpc.main.block_to_dict` to a separate formatting module
- [ ] make score available on block object, so that a returned block dict can be formatted without a `chain` reference ?
- [ ] support all the `eth_*` read-only methods
  - [x] `eth_protocolVersion`
  - [ ] `eth_syncing`
  - [ ] `eth_coinbase`
  - [ ] `eth_mining`
  - [ ] `eth_hashrate`
  - [ ] `eth_gasPrice`
  - [ ] `eth_accounts`
  - [x] `eth_blockNumber`
  - [x] `eth_getBalance`
  - [x] `eth_getStorageAt`
  - [x] `eth_getTransactionCount`
  - [ ] `eth_getBlockTransactionCountByHash`
  - [ ] `eth_getBlockTransactionCountByNumber`
  - [x] `eth_getUncleCountByBlockHash`
  - [x] `eth_getUncleCountByBlockNumber` block int
  - [ ] `eth_getUncleCountByBlockNumber` block string
  - [x] `eth_getCode`
  - [ ] `eth_call`
  - [ ] `eth_estimateGas`
  - [x] `eth_getBlockByHash`
  - [x] `eth_getBlockByNumber` block int
  - [x] `eth_getBlockByNumber` block string
  - [ ] `eth_getTransactionByHash`
  - [ ] `eth_getTransactionByBlockHashAndIndex`
  - [ ] `eth_getTransactionByBlockNumberAndIndex`
  - [ ] `eth_getTransactionReceipt`
  - [ ] `eth_getUncleByBlockHashAndIndex`
  - [ ] `eth_getUncleByBlockNumberAndIndex` block int
  - [ ] `eth_getUncleByBlockNumberAndIndex` block string
  - [ ] `eth_getCompilers`
  - [ ] `eth_getLogs`
  - [ ] `eth_getWork`

To save for another PR:

- [ ] serve over socket
- [ ] server over HTTP

#### Cute Animal Picture


.gitignore
```diff
@@ -59,3 +59,4 @@ prof/**
 docs/evm.*rst
 venv*
 .eggs/
+tests/rpc/mainnet_tester_chain.db/*
```
pipermerriam: I'm a big fan of relegating things to a `fixtures/` dir if you're into that sort of thing.

evm/rpc/main.py
```diff
@@ -0,0 +1,90 @@
...
+        The key entry point for all incoming requests
+        '''
+        try:
+            request = json.loads(request_json)
+        except json.decoder.JSONDecodeError as e:
+            raise ValueError("cannot parse json in %r" % request_json) from e
+        if request.get('jsonrpc', None) != '2.0':
+            raise NotImplemented("Only the 2.0 jsonrpc protocol is supported")
+        response = generate_response(request)
+        if not hasattr(self, request['method']):
```
pipermerriam: This looks like a recipe for a security vulnerability.  Can we have an explicit set of allowed methods or something?

tests/rpc/generate_mainnet_tester_chain.py
```diff
@@ -0,0 +1,126 @@
+import os
+import time
+
+from eth_keys import (
```
pipermerriam: you can just do `from eth_keys import keys` and it'll automatically *just* work with whatever the configured backend is.  (`keys` will be an instance of `KeyAPI`)

tests/rpc/generate_mainnet_tester_chain.py
```diff
@@ -0,0 +1,126 @@
...
+
+def build_chain(chain, account_keys):
+    num_blocks = chain.get_block().number - 1
+    if num_blocks < 1:
+        key1 = account_keys[0]
+        chain.mine_block(coinbase=key1.public_key.to_canonical_address())
+    if num_blocks < 2:
+        # mine 2nd block
+        # note that the db folder is .gitignore, so some extra work is required to save the data
+        #  - hacky short-term: remove from .gitignore, commit, then re-add to .gitignore
```
pipermerriam: `git add -f ./folder-that-is-ignored` or is it more than that?

pipermerriam: would it be better to use a memory db, serialize the kv to be hex encoded since json and bytes aren't necessarily friendly and just store it as a JSON document?

tests/rpc/rpc/test_rpc_eth_blocks.py
```diff
@@ -0,0 +1,77 @@
...
+    db = BaseChainDB(LevelDB(db_path))
+    return MainnetTesterChain(db)
+
+
+@pytest.fixture
+def rpc(chain):
+    return RPCServer(chain)
+
+
+def build_request(method, args):
```
pipermerriam: *naming nitpick*

`s/args/params`

tests/rpc/rpc/test_rpc_eth_blocks.py
```diff
@@ -0,0 +1,77 @@
...
+                'uncles': [],
+            },
+            None,
+        ),
+    ),
+)
+def test_eth_requests(rpc, rpc_request, expected_result, expected_error):
+    response_str = rpc.request(rpc_request)
+    response = json.loads(response_str)
+    if expected_result:
```
pipermerriam: Not sure if this pattern works for you but I've liked:

```python
if isinstance(expected_result, Exception):
  with pytest.raises(expected_result):
    ...
else:
   result = ...
   assert result == expected_result

Comment (carver) [reply tests/rpc/rpc/test_rpc_eth_blocks.py]: Yeah I like that pattern in general, but I don't think it applies here. The RPC server shouldn't raise exceptions, it should return json-encoded response with an error message embedded. Let's come back to it after I add an error message test. We might converge by then.

File: tests/rpc/generate_mainnet_tester_chain.py (hunk 4 above)
Comment (carver) [reply tests/rpc/generate_mainnet_tester_chain.py]: > git add -f ./folder-that-is-ignored or is it more than that?

leveldb keeps modifying files, even when just reading the db, so `git add -f` isn't enough to prevent the git status cruft. (because git wants to re-add the files that were previously added)

> would it be better to use a memory db, serialize the kv to be hex encoded since json and bytes aren't necessarily friendly and just store it as a JSON document?

Seems promising, I'll give it a shot.

File: tests/rpc/generate_mainnet_tester_chain.py (hunk 3 above)
Comment (carver) [reply tests/rpc/generate_mainnet_tester_chain.py]: TBH, this is copy-pasta from https://github.com/ethereum/eth-tester/blob/168c7a59/eth_tester/backends/pyevm/main.py

I'll take another pass through it, but since it was just for generating the test database, I wasn't holding it to the same standard as prod code.

File: tests/rpc/test_rpc_eth_blocks.py [hunk 7]
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.fixture
+def chain():
+    server_test_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+    db_path = os.path.join(server_test_path, 'fixtures', 'rpc_test_chain.db')
+    db = MemoryDB()
+    with open(db_path) as f:
+        key_val_hex = json.loads(f.read())
+        db.kv_store = {decode_hex(k): decode_hex(v) for k, v in key_val_hex.items()}
```
gsalgado: Does this mean we'll have to regenerate `rpc_test_chain.db` every time there's a change to the DB schema? Did you consider storing just the RLP data for the blocks and then here decoding/inserting them in the DB using the public APIs, like in https://github.com/ethereum/py-evm/blob/master/evm/p2p/test_peer.py#L452?

carver: How often is the schema changing? I assume that at some point the schema will be locked down, or that the database would have to handle its own schema upgrades. Otherwise, people's stored chain db could become corrupt at each `py-evm` upgrade.

I just tried loading that data set you linked to, and `chain.get_canonical_block_by_number(...)` works with 0, 1, and 2, but then on 3, I get an error. Any idea if this is a limitation of the approach or a corrupt database?

```
~/code/py-evm/evm/chains/chain.py in get_canonical_block_by_number(self, block_number)
    165         """
    166         validate_uint256(block_number, title="Block Number")
--> 167         return self.get_block_by_hash(self.chaindb.lookup_block_hash(block_number))
    168 
    169     def get_block_by_hash(self, block_hash):

~/code/py-evm/evm/chains/chain.py in get_block_by_hash(self, block_hash)
    173         validate_word(block_hash, title="Block Hash")
    174         block_header = self.get_block_header_by_hash(block_hash)
--> 175         vm = self.get_vm(block_header)
    176         return vm.get_block_by_header(block_header)
    177 

~/code/py-evm/evm/chains/chain.py in get_vm(self, header)
    135 
    136         vm_class = self.get_vm_class_for_block_number(header.block_number)
--> 137         return vm_class(header=header, chaindb=self.chaindb)
    138 
    139     #

~/code/py-evm/evm/vm/base.py in __init__(self, header, chaindb)
     40         self.chaindb = chaindb
     41         block_class = self.get_block_class()
---> 42         self.block = block_class.from_header(header=header, chaindb=self.chaindb)
     43 
     44     @classmethod

~/code/py-evm/evm/vm/forks/frontier/blocks.py in from_header(cls, header, chaindb)
    247             uncles = []
    248         else:
--> 249             uncles = chaindb.get_block_uncles(header.uncles_hash)
    250 
    251         transactions = chaindb.get_block_transactions(header, cls.get_transaction_class())

~/code/py-evm/evm/db/chain.py in get_block_uncles(self, uncles_hash)
    143     def get_block_uncles(self, uncles_hash):
    144         validate_word(uncles_hash, title="Uncles Hash")
--> 145         return rlp.decode(self.db.get(uncles_hash), sedes=rlp.sedes.CountableList(BlockHeader))
    146 
    147     @to_list

~/code/py-evm/evm/db/journal.py in get(self, key)
    135 
    136     def get(self, key):
--> 137         return self.wrapped_db.get(key)
    138 
    139     def set(self, key, value):

~/code/py-evm/evm/db/backends/memory.py in get(self, key)
     11 
     12     def get(self, key):
---> 13         return self.kv_store[key]
     14 
     15     def set(self, key, value):

KeyError: b"k\x17\xb98\xc6\xe4\xef\x18\xb2j\xd8\x1b\x9c\xa3Q_'\xfd\x9cN\x82\xaa\xc5j\x1f\xd8\xea\xb2\x88x^A"
```

gsalgado: At this early stage I wouldn't be surprised if we changed the db schema frequently. We'd probably version those schemas so that existing user chain data can continue to be used and/or migrated, but ideally we shouldn't have to worry about that in those tests and that's why I suggested an approach that is not locked to a specific db schema.

That data set I linked to contains only the header data (the README describes exactly what is there), so I'm surprised you were even able to use get_canonical_block_by_number() with it

evm/rpc/main.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def generate_response(request):
+    return {
+        'id': request['id'],
+        'jsonrpc': request['jsonrpc'],
+    }
+
+
+class RPCServer:
```
gsalgado: We want our RPC server to be asyncio-compatible since that's our approach to concurrency. The current `Chain` class doesn't do concurrency but eventually it will, just like `LightChain`. The existing one in rpc/server.py could be used as an example

carver: I started to think about how to be compatible with the current `Chain` and the future async chain. I didn't come up with anything good, but I don't know if that's because I'm an `asyncio` n00b or if there just isn't an elegant solution.

My current best idea is to go through and make all of the relevant `Chain` methods `async def` without any other changes. (I suppose I would also have to make any current Chain callers asyncio aware, too). Then I can start writing the server against a unified and future-proof async `Chain` interface.

Do you see any obvious problems with going this direction?

pipermerriam: @gsalgado can likely give a more definitive answer, but this seems at least worth spending a few hours on to validate it's viability.

carver: Yup, I think so, too. Starting to dig in...

gsalgado: Well, I believe that once we have a full client syncing blocks, it won't make sense to run the JSON-RPC with a LightChain, right? If that's the case, we can have this initial RPC server asyncio-compatible (by using aiohttp), and using the `await` keyword when calling `LightChain` methods, but when we switch it to using `Chain` we'd simply drop the `await` keyword in method calls

carver:
So here's what I'm thinking for the IPC server: 918c248

This way, the server can be async without making `RPCServer` and `Chain` async.

Thoughts @gsalgado @pipermerriam ?

evm/rpc/format.py
```diff
@@ -38,3 +40,24 @@ def block_to_dict(block, chain, include_transactions):
         block_dict['transactions'] = [encode_hex(tx.hash) for tx in block.transactions]
 
     return block_dict
+
+
+def format_params(*formatters):
+    def decorator(func):
+        @functools.wraps(func)
+        def formatted_func(self, *args):
+            assert len(formatters) == len(args), "could not apply %d formatters to %r" % (
```
pipermerriam: Assertion errors can be disabled.  Can we change to an explicit raise of something like `TypeError`?

evm/rpc/format.py
```diff
@@ -38,3 +40,24 @@ def block_to_dict(block, chain, include_transactions):
...
+                len(formatters),
+                args,
+            )
+            formatted = (formatter(arg) for formatter, arg in zip(formatters, args))
+            return func(self, *formatted)
+        return formatted_func
+    return decorator
+
+
+def hex_to_int(value):
```
pipermerriam: *naming nitpick*

Maybe `to_int_if_hex` would be a more appropriate name for this function?

evm/rpc/ipc.py
```diff
@@ -11,49 +12,81 @@
 from evm.rpc.main import RPCServer
 
 
+MAXIMUM_REQUEST_BYTES = 10000
+
+
 @curry
 async def connection_handler(execute_rpc, reader, writer):
+    '''
+    Catch fatal errors, log them, and close the connection
```
pipermerriam: I don't see any logging or connection closing...?

evm/rpc/modules/eth.py
```diff
@@ -1,15 +1,40 @@
...
+@contextmanager
+def state_at_block(chain, at_block, read_only=True):
+    if at_block == 'pending':
+        at_header = chain.header
+    elif at_block == 'latest':
+        at_header = chain.get_canonical_head()
+    elif at_block == 'earliest':
+        # TODO find if genesis block can be non-zero. Why does 'earliest' option even exist?
+        at_header = chain.get_canonical_block_by_number(0).header
+    else:
```
pipermerriam: We can potentially avoid some confusing errors down the line by making this an `elif is_hex_or_int(at_block):` and then putting an exception in the final `else` statement so that we are *sure* that `at_block` is well formed.

evm/rpc/modules/eth.py
```diff
@@ -20,51 +45,52 @@ class Eth(RPCModule):
     def accounts(self):
         raise NotImplementedError()
 
+    @apply_to_return_value(hex)
```
pipermerriam: Given the simplicity of this function, can we just change the return statement to be `hex(num)` and avoid the albeit minor overhead of the additional function call.

tests/rpc/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+from evm.rpc.ipc import (
+    get_test_chain,
+    start,
+)
+
+
+@pytest.fixture(scope='session')
+def ipc_pipe():
+    tmpdir = tempfile.gettempdir()
+    unique_id = datetime.datetime.timestamp(datetime.datetime.now())
```
pipermerriam: Unless you were intentionally going for a timestamped ipc socket you could use `uuid.uuid4()` for this as well.

tests/rpc/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    return os.path.join(tmpdir, 'test-%s.ipc' % unique_id)
+
+
+@pytest.fixture(scope='session', autouse=True)
+def ipc_server(ipc_pipe):
+    def serve_test_data():
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        start(ipc_pipe, get_test_chain())
+    thread = Thread(target=serve_test_data)
```
pipermerriam: in python3 the `Thread` object now takes a `daemon` kwarg!  much rejoicing 🎂

carver:
*winces* - posting this *very* in-progress part, as I wrap up for the night. Lots of obvious things to do to improve, but it shows the beginnings of the approach.

Obvious things to improve:

- [ ] remove hard-coded `chain_fixture` in `/tests/rpc/test_rpc_fixtures.py`
- [ ] load all fixtures from `/fixtures/BlockchainTests/*`
- [ ] test storage value
- [ ] test block header values
- [ ] probably just delete `rpc/test_rpc_eth_blocks.py` (or if not, refactor)
- [ ] refactor validation utilities out of `/tests/rpc/test_rpc_fixtures.py`

carver:
So I'm thinking maybe now is a good time to pinch off this long PR and merge it. With a test infrastructure in place, the rest of the json-rpc methods can be added in more focused PRs.

evm/rpc/ipc.py
```diff
@@ -0,0 +1,127 @@
...
+def strip_non_json_prefix(raw_request):
+    if raw_request and raw_request[0] != '{':
+        prefix, bracket, rest = raw_request.partition('{')
+        return prefix.strip(), bracket + rest
+    else:
+        return '', raw_request
+
+
+async def write_error(writer, message):
+    json_error = '{"error": "%s"}\n' % message
```
pipermerriam: I think constructing json using string formatting is asking for problems since the message could have characters that need to be escaped.  Can we use a legit JSON encoder here?

evm/rpc/main.py
```diff
@@ -0,0 +1,95 @@
...
+            logging.info("RPC method caused exception", exc_info=True)
+            response['error'] = str(exc)
+
+        if request['method'] == 'evm_resetToGenesisFixture' and 'result' in response:
+            self._set_chain(response['result'])
+            response['result'] = True
+
+        return json.dumps(response)
+
+    def _set_chain(self, new_chain):
```
pipermerriam: Thoughts on doing this as a `property` with a `setter`?  Would remove the duplication of code in the `__init__`.  Alternatively, you can just call this method in `__init__` to remove that duplication.

evm/rpc/modules/eth.py
```diff
@@ -0,0 +1,134 @@
...
+
+        block_dict = block_to_dict(block, self._chain, include_transactions)
+
+        return block_dict
+
+    @format_params(to_int_if_hex, identity)
+    def getBlockByNumber(self, block_number, include_transactions):
+        if isinstance(block_number, int):
+            block = self._chain.get_canonical_block_by_number(block_number)
+            assert block.number == block_number
```
pipermerriam: Doesn't have to happen now but [since assertions can be disabled](https://stackoverflow.com/questions/1273211/disable-assertions-in-python) I have tended to avoid using them as protection in production code.

tests/rpc/conftest.py
```diff
@@ -0,0 +1,26 @@
...
+
+from evm.rpc.ipc import (
+    start,
+)
+
+
+@pytest.fixture(scope='session')
+def ipc_pipe():
+    tmpdir = tempfile.gettempdir()
+    return os.path.join(tmpdir, 'test-%s.ipc' % uuid.uuid4())
```
pipermerriam: What do you think about changing this to a `yield` and doing an explicit cleanup check to be sure we're not leaving this file laying around?  Is that even a possible issue?  If not, please ignore.

evm/rpc/main.py
carver: The duplication being that we pass chain in as an argument when creating the module? I think I like that it's explicit there. We need to loop through and create module instances anyway, so it doesn't save us the loop.

I'll change it to a property setter anyway, since it looks better, but I'm not sure it buys us much in duplication.

evm/rpc/modules/eth.py
carver: I'm thinking of removing this altogether. I think it was left over from early testing. We haven't typically done checks like this in other places.

carver:
Ready for another look @pipermerriam

evm/rpc/main.py
```diff
@@ -0,0 +1,115 @@
...
+
+REQUIRED_REQUEST_KEYS = (
+    'id',
+    'jsonrpc',
+    'method',
+)
+
+
+def validate_request(request):
+    for key in REQUIRED_REQUEST_KEYS:
```
pipermerriam: *nitpick*

I have started gravitating to the following style because it captures all of the missing keys.

```python
missing_keys = set(REQUIRED_REQUEST_KEYS).difference(request.keys())
if missing_keys:
    raise ValueError("...")
```

evm/rpc/main.py
```diff
@@ -0,0 +1,115 @@
...
+    module_classes = (
+        Eth,
+        EVM,
+    )
+
+    def __init__(self, chain):
+        self.modules = {}
+        self.chain = chain
+        for M in self.module_classes:
+            self.modules[M.__name__.lower()] = M(chain)
```
pipermerriam: *nitpick*

Enforcing uniqueness on the set of module names would save us a potentially confusing edge case.

evm/rpc/main.py
```diff
@@ -0,0 +1,115 @@
...
+        try:
+            return getattr(module, method_name)
+        except AttributeError:
+            raise ValueError("Method not implemented: %r" % rpc_method)
+
+    def execute(self, request):
+        '''
+        The key entry point for all incoming requests
+        '''
+        response = generate_response(request)
```
pipermerriam: *mutability nitpick*

The places where `response` is mutated are bugging me.  Mostly that it's only through strict control flow that we don't end up with a response which has both an `"error"` and `"result"` key.

Suggestions to remove it.

```python
# remove mutation
base_response = generate_response(request)
...
if good:
    response = cytoolz.assoc(base_response, 'result', ...)
else:
    response = cytoolz.assoc(base_response, 'error', ...)

# or moving logic into `generate_response`
if good:
    response = generate_response(request, result=...)
else:
    response = generate_response(request, error=...)  
```

Second option should include validation that only one of error/result are provided or a separate response generation functions for error/result.

evm/rpc/main.py
```diff
@@ -0,0 +1,115 @@
...
+            validate_request(request)
+
+            if request.get('jsonrpc', None) != '2.0':
+                raise NotImplementedError("Only the 2.0 jsonrpc protocol is supported")
+
+            method = self._lookup_method(request['method'])
+            params = request.get('params', [])
+            response['result'] = method(*params)
+
+            if request['method'] == 'evm_resetToGenesisFixture':
```
pipermerriam: Suggest the following structure change to remove some of the juggling of the result in and out of the `response` object.

```python
result = method(*params)

if request['method'] == 'evm_resetToGenesisFixture':
    self.chain = result
    response = ...
else:
    response = ...
```

evm/rpc/main.py
```diff
@@ -0,0 +1,115 @@
...
+            response['error'] = str(exc)
+        except Exception as exc:
+            logging.info("RPC method caused exception", exc_info=True)
+            response['error'] = str(exc)
+
+        return json.dumps(response)
+
+    @property
+    def chain(self):
+        return self.__chain
```
pipermerriam: reason for the double underscore?

evm/rpc/modules/eth.py
```diff
@@ -0,0 +1,133 @@
...
+)
+
+
+def get_header(chain, at_block):
+    if at_block == 'pending':
+        at_header = chain.header
+    elif at_block == 'latest':
+        at_header = chain.get_canonical_head()
+    elif at_block == 'earliest':
+        # TODO find if genesis block can be non-zero. Why does 'earliest' option even exist?
```
pipermerriam: If you figure this out let me know because I've had the same question

evm/rpc/modules/main.py
```diff
@@ -0,0 +1,13 @@
+
+
```
pipermerriam: bonus whitespace?

tests/rpc/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.fixture(scope='session', autouse=True)
+def ipc_server(ipc_pipe, event_loop):
+    server = start(ipc_pipe, loop=event_loop)
+
+    yield
```
pipermerriam: Seems like you might as well `yield server` here just *because*

tests/rpc/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+
+    server.close()
+    event_loop.run_until_complete(server.wait_closed())
+    os.remove(ipc_pipe)
```
pipermerriam: This feels like a weird place for this.  Is there a reason to put it here rather than in the `ipc_pipe` fixture?

tests/rpc/test_rpc_fixtures.py
```diff
@@ -0,0 +1,225 @@
...
+        'transactions',
+        'uncles',
+    )
+    assert actual_block == expected
+
+
+def validate_block(rpc, block_fixture, at_block):
+    if is_hex(at_block) and len(at_block) == 66:
+        rpc_method = 'eth_getBlockByHash'
+    else:
```
pipermerriam: This pattern has bitten me enough that I avoid it.

```python
elif is_integer(at_block):
    ...
else:
    raise ValueError("bad `at_block` value")
```

evm/rpc/ipc.py
```diff
@@ -0,0 +1,129 @@
...
+    writer.close()
+
+
+async def connection_loop(execute_rpc, reader, writer, logger):
+    raw_request = ''
+    while True:
+        request_bytes = b''
+        try:
+            request_bytes = await reader.readuntil(b'}')
+        except asyncio.LimitOverrunError as e:
```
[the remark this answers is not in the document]
carver: Yes

evm/rpc/main.py
carver: Ok, I can clean up this logic. The issue I run into with validating that only one of error/result are provided is that it's already an error condition if that check fails. I don't want to raise an exception to crash the RPC server, but there's already an error being displayed. The best answer I have is to append an additional error message to the existing one.

evm/rpc/main.py
carver: I guess paranoia that someone would try to modify `_chain` internally and miss the setter. I can change it to single.

tests/rpc/conftest.py
carver: Is it unreasonable that I want to leave it out because it's not necessary?

I don't want someone (ie~ me in 3 months) to read this and wonder where server is being used. I think if it yields nothing, it's clearer that the point of this fixture is just to start up the server and then shut it down.

tests/rpc/conftest.py
carver: I'm going to rename the fixture `ipc_pipe` to `ipc_pipe_path` to clear that up. Since all that fixture does is provide the path, it doesn't really make sense for it to touch the fs at all.

tests/rpc/conftest.py
pipermerriam: No, not unreasonable though it might be good to leave a comment indicating this intent.

evm/rpc/main.py
pipermerriam: It's been a while since I read the docs, but python does something funky to double underscore variables.  Single seems suitable to say (leave this alone)

evm/rpc/main.py
pipermerriam: >  it's already an error condition if that check fails

I think it's ok to have a hard failure that crashes the server if `generate_response` is called with both of those keywords.  You could also just do two separate functions.

```python
# or moving logic into `generate_response`
if good:
    response = generate_success_response(request, result=...)
else:
    response = generate_error_response(request, error=...)  
```

pipermerriam:
Good to go with whatever fixes you put in place.  :shipit:

PR #184, opened 2017-12-02
docstrings: a bit more mine_block info

carver:
### What was wrong?

Type info about the keyword args to `mine_block` was not apparent in the low-level api docs.

### How was it fixed?

Adding the docstrings with parameter type info, to autopopulate the low-level api docs on RTD.

#### Cute Animal Picture

### DIFF
--- a/evm/vm/forks/frontier/blocks.py
+++ b/evm/vm/forks/frontier/blocks.py
@@ -288,16 +288,16 @@ class FrontierBlock(BaseBlock):
 
     def mine(self, **kwargs):
         """
-        - `coinbase`
-        - `uncles_hash`
-        - `state_root`
-        - `transaction_root`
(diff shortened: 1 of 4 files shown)


evm/vm/forks/frontier/blocks.py
```diff
@@ -288,16 +288,16 @@ def add_uncle(self, uncle):
...
-        - `uncles_hash`
-        - `state_root`
-        - `transaction_root`
-        - `receipt_root`
-        - `bloom`
-        - `gas_used`
-        - `extra_data`
-        - `mix_hash`
-        - `nonce`
+        :param bytes coinbase: 20-byte public address to receive block reward
```
pipermerriam: 👍  lets do this everywhere.

PR #189, opened 2017-12-04
lightchain_shell: run LightChain in the background of interactive shell

gsalgado:
Also extend the quickstart and building_chains docs, and add the asyncio
sphinx extension so that coroutines get special treatment in docs.

### DIFF
--- a/docs/building_chains.rst
+++ b/docs/building_chains.rst
@@ -22,11 +22,11 @@ class:
   from evm.vm.forks.homestead import HomesteadVM
 
   chain_class = Chain.configure(
-    name='Test Chain',
-    vm_configuration=(
-      (constants.GENESIS_BLOCK_NUMBER, FrontierVM),
-      (constants.HOMESTEAD_MAINNET_BLOCK, HomesteadVM),
-    ),
+      name='Test Chain',
+      vm_configuration=(
+          (constants.GENESIS_BLOCK_NUMBER, FrontierVM),
+          (constants.HOMESTEAD_MAINNET_BLOCK, HomesteadVM),
+      ),
   )
 
 Then to initialize, you can start it up with an in-memory database:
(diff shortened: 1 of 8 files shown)


docs/building_chains.rst
```diff
@@ -42,3 +42,41 @@ Then to initialize, you can start it up with an in-memory database:
...
+
+::
+
+  from evm.chains.mainnet import MAINNET_VM_CONFIGURATION, MAINNET_NETWORK_ID
+  from evm.p2p import ecies
+  from evm.p2p.lightchain import LightChain
+
+  DemoLightChain = LightChain.configure(
+      name='Demo LightChain',
+      privkey=ecies.generate_privkey(),
```
pipermerriam: doesn't need to be done here, but I think it's ok for us to start storing some data under `~/.py-evm/` which would include generating this value automatically for each chain if it's not already present.  This is how other clients do it yes?

docs/building_chains.rst
```diff
@@ -42,3 +42,41 @@ Then to initialize, you can start it up with an in-memory database:
...
+headers, you should tell asyncio to execute its run() method:
+
+::
+
+  import asyncio
+  from evm.db.backends.memory import MemoryDB
+  from evm.db.chain import BaseChainDB
+  from evm.chains.mainnet import MAINNET_GENESIS_HEADER
+
+  chain = DemoLightChain.from_genesis_header(BaseChainDB(MemoryDB()), MAINNET_GENESIS_HEADER)
```
pipermerriam: The amount of imports here make me wonder how we can establish some built-in `Chain` subclasses which have suitable defaults as well as potentially environment based configuration for the `ChainDB` class and the `BaseDB` class.  This could reduce this down to 

```
from evm.chains.demo import DemoMainnetLightChain

chain = DemoMainnetLightChain()
chain.run()   # can we put the `asyncio` loop setup into this or another method?
```

docs/building_chains.rst
gsalgado: Created #198

docs/building_chains.rst
gsalgado: Yeah, the amount of imports is annoying, and maybe we could have a helper for the asyncio loop, but it should not be in chain.run() because different services will use different asyncio APIs to run the chain (e.g. the interactive shell uses `run_until_complete()` whereas the RPC server uses `ensure_future()`)

PR #200, opened 2017-12-06
p2p and rpc: a few misc fixes

gsalgado:
- Log to a file in lightchain_shell
- Rename Peer.stop_and_wait_until_finished() to stop()
- lightchain.py, lightchain_shell.py and rpc/server.py now connect to ropsten by default
- peer.py no longer takes a db/mainnet arg as it no longer handles sync

### DIFF
--- a/evm/p2p/lightchain.py
+++ b/evm/p2p/lightchain.py
@@ -88,7 +88,7 @@ class PeerPool:
     async def stop_all_peers(self):
         self.logger.info("Stopping all peers ...")
         await asyncio.gather(
-            *[peer.stop_and_wait_until_finished() for peer in self.connected_nodes.values()])
+            *[peer.stop() for peer in self.connected_nodes.values()])
 
     async def stop(self):
         self._should_stop.set()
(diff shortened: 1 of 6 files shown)


evm/p2p/lightchain.py
```diff
@@ -265,9 +265,10 @@ def peer_block_height(peer: LESPeer):
         done, pending = await asyncio.wait(
             [self._announcement_queue.get(), wait_for_stop_event()],
             return_when=asyncio.FIRST_COMPLETED)
-        # The async call above returns as soon as one of our 2 coroutines complete, so we know
-        # for sure we'll have one task in the <done> and one task in the <pending> set.
-        pending.pop().cancel()
+        # The asyncio.wait() call above may return both tasks as done, so check that we have
+        # a pending task before attempting to cancel it.
+        if len(pending) > 0:
```
pipermerriam: Do you think `while pending:` is more legible?

gsalgado: `while pending:` or `if pending:`? There would never be more than one pending task here

gsalgado:
I guess you mean docs for rpc/server.py? I'll add them tomorrow

evm/p2p/lightchain.py
pipermerriam: Ok, now that I went and read what `asyncio.wait` does that makes sense.  

Would it be easy at some point in the future for a change to be made to the code above such that `pending` then has more than one *thing* in it at which point this would fail to stop the additional one?  Hence it seems appropriate to write this code in a way that deals with all items in the iterable rather than special casing it for a single item.

Suggestion was more about semantic correctness than the specifics of this functionality.  I'll defer to what you think is best.

pipermerriam:
:shipit: when ready.

Issue #192, opened 2017-12-05
STUB: Account access restriction


pipermerriam:
TODO: fill me in.

hwwhww:
Account access restriction #202

Issue #191, opened 2017-12-05
Account redesign


pipermerriam:
1. We need to support the account storage redesign.  This should be a configurable `StateDB` class that implements the following structure for accessing account information.
2. [A two-layer account trie inside a single-layer trie](https://ethresear.ch/t/a-two-layer-account-trie-inside-a-single-layer-trie/210)
3. [doc.md](https://github.com/ethereum/sharding/blob/develop/docs/doc.md#two-layer-trie-redesign)

```python
class State():
    def __init__(self, state_root, env, db):
        self.state_root, self.env. self.db = state_root, env, db
        cache = {}
        journal = []

    def get_balance(self, acct):
        key = sha3(acct) + b'\x00'
        if key not in self.cache:
            self.cache[key] = _trie_get(self.db, self.state_root)
        return self.cache[key]

    def get_nonce(self, acct): key = sha3(acct) + b'\x01' ...

    def get_code(self, acct): key = sha3(acct) + b'\x02' ...

    def get_storage(self, acct, k): key = sha3(acct) + b'\x03' + k ....
```

hwwhww:
@vbuterin Should we update the length of address in this ticket?
I remember is `20 bytes -> 24 bytes`?

hwwhww:
This ticket is assigned to @jannikluhn

jannikluhn:
Closed with #213

PR #202, opened 2017-12-06
Account access restriction

jannikluhn:
See #192 and https://ethresear.ch/t/account-read-write-lists/285


hwwhww:
This PR seems contains part of #191 or is based on assuming that #191 is ready?

jannikluhn:
@hwwhww The access prefix list is in a format which corresponds to how state will be stored according to #191. But it neither changes nor relies on how state is actually stored.

evm/db/state.py
```diff
@@ -44,13 +53,24 @@ class State:
...
+        read_list=None,
+        write_list=None
+    ):
         if read_only:
             self.db = ImmutableDB(db)
         else:
             self.db = db
         self._trie = HashTrie(Trie(self.db, root_hash))
 
+        self.check_access_restrictions = read_list is not None or write_list is not None
```
pipermerriam: *naming nitpick*

What do you think about changing this property to `is_access_restricted`

evm/logic/system.py
```diff
@@ -73,7 +81,10 @@ def selfdestruct_eip161(computation):
 
 
 def _selfdestruct(computation, beneficiary):
-    with computation.vm.state_db() as state_db:
+    with computation.vm.state_db(
+        read_list=computation.msg.read_list,
+        write_list=computation.msg.write_list
```
pipermerriam: This repetition makes me think that we should add a convenience method `Computation.get_state_db()` which auto-populates these values based on the current message.

tests/core/access-restrictions-utils/test_checks_and_conversion.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+TEST_ADDRESS1 = b'\xaa' * 20
+TEST_ADDRESS2 = b'\xbb' * 20
+TEST_PREFIX_LIST = to_prefix_list_form([
+    [TEST_ADDRESS1, b'\x00'],
+    [TEST_ADDRESS1, b'\x01\x00'],
+    [TEST_ADDRESS2, b'\xff' * 32]
+])
+print(TEST_PREFIX_LIST)
```
pipermerriam: Can you remove this print statement

evm/vm/message.py
```diff
@@ -46,6 +52,8 @@ def __init__(self,
                  data,
                  code,
                  origin=None,
+                 read_list=None,
+                 write_list=None,
```
pipermerriam: Is it valid to have only one of these lists present (read list but no write list or write list but no read list).  

If not, what do you think about changing this to be a single argument `read_and_write_list` which takes the form of a 2-tuple `(read_list, write_list)`.  Then within the constructor we can unpack it if it's not `None`:  `read_list, write_list = read_and_write_list`.

evm/utils/state_access_restriction.py
```diff
@@ -0,0 +1,36 @@
...
+            return True
+    return False
+
+
+def to_prefix_list_form(access_list):
+    """Expand an access list to a flat list of storage key prefixes.
+
+    As input a list of the form `[[address, prefix1, prefix2, ...], ...]` is expected.
+    """
+    prefix_list = []
```
pipermerriam: Most of py-evm avoids this pattern of creating a list, filling it, then returning it using the `eth_utils.to_tuple` decorator.

```python
def build_powers_with_mutation():
    the_values = []
    for i in range(10):
        the_values.append(2**i)
    return the_values

@to_tuple
def build_powers_no_mutation():
    for i in range(10):
        yield 2**i
```

evm/utils/state_access_restriction.py
```diff
@@ -0,0 +1,36 @@
...
+    for obj in access_list:
+        address, storage_prefixes = obj[0], obj[1:]
+        prefix_list.append(keccak(address) + NONCE_TRIE_PREFIX)
+        prefix_list.append(keccak(address) + BALANCE_TRIE_PREFIX)
+        prefix_list.append(keccak(address) + CODE_TRIE_PREFIX)
+        for prefix in storage_prefixes:
+            prefix_list.append(keccak(address) + STORAGE_TRIE_PREFIX + prefix)
+    # TODO: remove duplicates/redundancies?
```
pipermerriam: You can do this with the `eth_utils.to_set` decorator which will cast the return value of this function to a set, removing duplicates.

evm/validation.py
```diff
@@ -227,3 +227,19 @@ def validate_header_params_for_configuration(header_params):
                 ", ".join(tuple(sorted(extra_fields))),
             )
         )
+
+
+def validate_access_list(access_list):
+    for obj in access_list:
+        if len(obj) == 0:
+            raise ValidationError("Access list entry must at least specify an account address.")
+        address, prefixes = obj[0], obj[1:]
```
pipermerriam: *fancy python3 stuff nitpick*

We can use the new python 3 star unpacking syntax to do this

```python
address, *prefixes = obj
```

evm/logic/system.py
jannikluhn: Good idea, wasn't happy with this either. My first instinct was to add the defaults to `VM.state_db()` but this obviously doesn't work because `VM` doesn't know about the current tx/msg.

Naming convention wise, should it be `get_state_db` or `state_db`? Looks like the difference right now is that `VM.state_db` is a context manager whereas `ChainDB.get_state_db` is not. So `Computation.state_db` seems to be more consistent.

evm/utils/state_access_restriction.py
jannikluhn: Yes, but not completely (there can be redundant prefixes, e.g. `'\x00'`` and `'\x00\x11'`, won't matter in practice though I guess).

jannikluhn:
Alongside the improvements suggested by @pipermerriam I've also made two other changes:
- Conversion from access lists as specified in the transaction to read/write lists (via `to_prefix_list_form`) is removed from the Message object. I think this should happen in the Transaction object right after it gets decoded.
- Access restrictions are checked in `State.delete_account`, `State.account_has_code_or_nonce`, and `State.account_is_empty`. I'm not sure what to do with `State.account_exists` and `State.touch_account` -- what state do these functions actually access? I wouldn't be surprised if we had to make some changes in the VM implementation in that regard anyway so this might be out of scope of this PR.

pipermerriam:
> I'm not sure what to do with `State.account_exists` and `State.touch_account`

These should *in theory* read/write all parts of the account necessary for it *exist* in the state trie.  That behavior needs to be defined in whatever EIP this is part of as that could mean any number of things.  To me, the intuitive thing to do is to have read/write *all* parts of the account.

pipermerriam:
FYI, standard procedure for merging pull requests in this repo has been that once it's marked as approval, anyone is welcome to merge.  *Normally* that has been left to the person who opened the pull request.

We also have been using the  "Needs Review" and "Ready for Merge" labels for additional signalling in this department.

jannikluhn:
Will have to make some changes -- according to the spec there is only one "access list", not separate "read" and "write" lists.

evm/exceptions.py
```diff
@@ -134,3 +134,11 @@ class OutOfBoundsRead(VMError):
     boundaries of the buffer (such as with RETURNDATACOPY)
     """
     pass
+
+
+class UnannouncedStateAccess(VMError):
```
pipermerriam: Question about this exception.  *Should* raising this error result in a transaction failing with an OOG error or should it bubble beyond VM execution and cause a hard failure?  I'm assuming the former, but if it is the latter, this should not be a subclass of `VMError` as anything that inherits from that exception class will be captured as part of message execution.

evm/utils/state_access_restriction.py
```diff
@@ -0,0 +1,39 @@
...
+    As input a list of the form `[[address, prefix1, prefix2, ...], ...]` is expected.
+    """
+    for obj in access_list:
+        address, *storage_prefixes = obj
+        yield keccak(address) + NONCE_TRIE_PREFIX
+        yield keccak(address) + BALANCE_TRIE_PREFIX
+        yield keccak(address) + CODE_TRIE_PREFIX
+        for prefix in storage_prefixes:
+            yield keccak(address) + STORAGE_TRIE_PREFIX + prefix
+    # TODO: remove redundant prefixes?
```
pipermerriam: If *"redundant"* means duplicated then this comment can be removed because the `to_set` will remove duplicates.

tests/core/vm/test_state_access_restrictions.py
```diff
@@ -0,0 +1,109 @@
...
+from evm.exceptions import UnannouncedStateAccess
+
+from evm.utils.numeric import (
+    big_endian_to_int,
+)
+from evm.utils.state_access_restriction import (
+    to_prefix_list_form,
+)
+
+from tests.core.fixtures import chain  # noqa: F401
```
pipermerriam: Note to self: TODO: clean this up to live in `tests.core.conftest` and remove the fixture imports from tests.

evm/exceptions.py
jannikluhn: You're right, the former. Scenario: An attacker creates a transaction which does some heavy computation and only in the end accesses some state outside of the access list. In that case he needs to pay for the gas, otherwise the collator would have worked for free.

evm/utils/state_access_restriction.py
jannikluhn: No, what I meant is this: Prefix 1 `'eth'`, prefix 2 `'ethereum'`. Everything that matches prefix 2 matches prefix 1 as well, so adding prefix 2 is useless and it could be removed from the list.

pipermerriam: Would you mind opening an issue for this and labeling it "Good First Issue".  Should be well specified functionality that a newcomer could easily come along and do for us.

Issue #195, opened 2017-12-05
Stateless/pure transaction processing


pipermerriam:
We need a *pure* function for applying transactions.  

As *inputs* it would take:

- the current state (including state root, tx index, etc)
- the transaction to be processed
- a witness *proof* for all of the parts of the state that the transaction *can/will* touch.

It would *output* the resulting state.

Implementation is up-in-the-air but current leading idea is to process the transaction as normal, modifying the state as we go and then discarding the state changes, only returning the updated state.  There is an underlying assumption that our state database engine supports lazy fetching of parts of the overall state.

---
## Tasks
- [x] Adding a new db wrapper - TrackedDB: #204 
- [x] Removing the transaction logic from the Block objects
- [x] Implement VMState Object #236
- [x] Refactoring VM, VMState, and Computation

pipermerriam:
Thoughts on implementation.  I suspect that one of the *first* things this will need is a database wrapper which keeps track of all of the parts of the state trie which get touched as part of transaction processing.  I think it would be good to do this as a subclass of `evm.db.state.StateDB`.  It may also require some changes to the `ChainDB` class since it is the entry point for interacting with the state database.

* The `StateDB` would be responsible for collecting all the touched keys.
* The `ChainDB` would be responsible for persisting the set of all of the touched keys each time the statedb is used.

hwwhww:
@pipermerriam 

> The `ChainDB` would be responsible for persisting the set of all of the touched keys each time the statedb is used.

### Q1: Is the `ChainDB` here means `BaseChainDB` class? Or we have to abstract  `BaseChainDB` and create a new `ChainDB` subclass?

### Q2: `pure function for applying transactions` feature is  for both shard chain and main chain?
If we make both types of chains have purified add_transaction function, something different if that it returns `State` in main chain and `StateObj` in shard chain where `StateObj` only contains properties.

### Q3: What would purified function need
In our prerious Q&A, you mentioned about  the **minimal changes** to the current `apply_transaction` API: That statement was in reference to these pieces of functionality and anything that allows it to access external APIs directly like the `chaindb`. This would **not** be a significant architecture change.
* https://github.com/ethereum/py-evm/blob/master/evm/vm/forks/frontier/blocks.py#L213-L236
* https://github.com/ethereum/py-evm/blob/master/evm/vm/forks/frontier/blocks.py#L263-L280


Do you mean we modify `evm.vm.base.VM.apply_transaction`?
[...]

PR #204, opened 2017-12-07
Adding a new db wrapper - TrackedDB

hwwhww:
### What was wrong?
This PR is part of #195
The first step: database wrapper 

### How was it fixed?
Adding a new database wrapper `TrackedDB`. `TrackedDB` has two dicts: 
* `_writes`: keeps track of the nodes that are accessed by `TrackedDB.put(key, value)`
* `_reads`: keeps track of that are accessed by `TrackedDB.get(key)`

Some new class properties of `TrackedDB`:
- `@property def reads(self)`: returns the whole `_reads` dict
- `@property def writes(self)`: returns the whole `_writes` dict

#### Cute Animal Picture

bear3_small


evm/db/state.py
```diff
@@ -217,3 +218,86 @@ def _get_account(self, address):
 
     def _set_account(self, address, account):
         self._trie[address] = rlp.encode(account, sedes=Account)
+
+
+class StateDB(BaseDB):
```
pipermerriam: Can I suggest we move this to it's own module since it's a low level DB wrapper.

I would also suggest we rename this to something more descriptive.  Here are some names that come to mind.

- `StenographerDB` (A stenographer is the person in a courtroom who types everything that people say).
- `TrackedDB` - tracks changes, touches, etc.
- `LedgerDB` - keeps a ledger of the db read/write

hwwhww:
> Have you looked into how this is going to interact with the JournalDB or whether this should be integrated with that class in some way. If an internal call within a transaction touches some state and then those state changes are reverted due to an evm exception, do the touched parts still remain in the read/write list? My assumption is yes, but if not, we need to take that into account.

My assumption is that in the stateless client use case, `StateDB` is more like a **disposable** container because:
1. With pure function, we can use StateDB more flexibly.
2. We mainly need it for creating the `collation witness data`. So `reads` and `writes` data is no longer need for stateless client.
3. So if transaction fails, we just don't include the `reads` and `writes` data of this tx

p.s. I am writing some more detailed description and will comment on #195

evm/db/state.py
hwwhww: Sure!
In `pyethereum`, a similar db wrapper is called `ListeningDB`. IMO `StenographerDB` is cool but a bit too long. I like `TrackedDB`.

hwwhww:
I want to leave the modification of `ChainDB` to other PR, after I actually have some test cases of using pure `apply_transaction` function. So this PR is only for introducing `TrackedDB`.

evm/db/tracked.py
```diff
@@ -0,0 +1,84 @@
...
+        self._writes = {}
+
+    def get(self, key):
+        """
+        Return the value of specific key and update read dict.
+        """
+        try:
+            current_value = self.wrapped_db.get(key)
+        except KeyError:
+            raise
```
pipermerriam: This whole `try/except/else` block is equivalent to the same code without the `try/except/else` since `except` just re-raises.

```python
current_value = self.wrapped_db.get(key)
self._reads[key] = current_value
return current_value
```

evm/db/tracked.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        Delete the key and update writes dict.
+        """
+        try:
+            current_value = self.wrapped_db.get(key)
+        except KeyError:
+            # do not need to be added in writes
+            pass
+        else:
+            self._writes[key] = current_value
```
pipermerriam: Should this instead be set to something like `None` to indicate that it's been deleted?

evm/db/tracked.py
```diff
@@ -0,0 +1,84 @@
...
+
+        return self.wrapped_db.delete(key)
+
+    def get_reads(self, key=None):
+        """
+        Return the whole or the specific value of reads dict.
```
pipermerriam: I'm thinking that we can accomplish the same by just exposing `reads` as a property.

```python
all_reads = db.get_reads()
single_read = db.get_reads(key=b'cow')
# becomes the following if `reads` is a public property
all_reads = db.reads
single_read = db.reads.get(key=b'cow')
```

Alternatively, I'd be more comfortable with this API if it only did one of these two things (preferrably returning all reads).

evm/db/tracked.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        if key is None:
+            return self._reads
+        else:
+            return self._reads[key] if key in self._reads else None
+
+    def get_writes(self, key=None):
+        """
+        Return the whole or the specific value of writes dict.
```
pipermerriam: Same thoughts as with the `get_reads` API

pipermerriam:
if `clear_logs` is only used for testing, lets remove it and just manually reset the `obj._reads` and `obj._writes` on the db object.

evm/db/tracked.py
hwwhww: Good, using `None` would be more clear :)
And I think the `delete` function don't *need* to return something?
Can I just modify it to:
```python
     def delete(self, key):
        """
        Delete the key and update writes dict.
        """
        self.wrapped_db.delete(key)
        self._writes[key] = None
```

hwwhww:
1. Followed @pipermerriam 's suggestions:
	* `@property def reads(self)`: returns the whole `_reads` dict
	* `@property def writes(self)`: returns the whole `_writes` dict

2. Update the logic in `delete`
    * Set `self._writes[key] = None`
    * By the way, it seems during state transition, `delete` method would only be triggered for `SELFDESTRUCT` or `EIP161 (Spurious Dragon)`?

3. Update the logic in `exsits`
    * Set `self._reads[key] = self.wrapped_db.get(key) if self.wrapped_db.exists(key) else None`
    * It would call `wrapped_db.get(key)` twice: one is in `self.wrapped_db.get(key)`, another is in `self.wrapped_db.exists(key)`). It seems a wasted call, but also keeping the code simple and wrapped.

PR #205, opened 2017-12-08
Making Chain asynchronous

carver:
### What was wrong?

LightChain is already async, and Chain is going to be. Better do it now before too many other modules are written on top of it.

_I think that I would like challenge the presumption that all the Chain things should be async and that all the synchronous methods should go away. But I don't feel informed enough to make an alternate proposal yet. I'm taking it as given for now and marching on._

### How was it fixed?

Make all-the-things async. I dislike how async seems to infect everything around it, but maybe I just don't grok async yet... Seeking feedback.

Todo:

- [x] make get_block() async
- [x] make get_canonical_block_by_number() async
- [x] make get_block_by_hash() async
- [ ] make a list of methods to turn async

#### Cute Animal Picture

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -74,13 +74,17 @@ For example:
 
 ::
 
+  loop = asyncio.get_event_loop()
+
   highest_block_num = chain.get_canonical_head().block_number
 
-  block1 = chain.get_canonical_block_by_number(1)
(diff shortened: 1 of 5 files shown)


carver:
This seems like the wrong direction based on https://github.com/ethereum/py-evm/pull/183#discussion_r155804386 and my intuition. I'll leave it open until we have a clear, alternative design.

Issue #197, opened 2017-12-05
Validator manager contract utils


hwwhww:
Implement some APIs for shard chain validators to call the `validator manager contract` on main chain.

* contract code: https://github.com/ethereum/sharding/blob/develop/sharding/contracts/validator_manager.v.py
* previous implementation (with pyethereum): https://github.com/ethereum/sharding/blob/develop/sharding/validator_manager_utils.py

Implementation should probably remain *low* level.  A simplistic `Contract` class with the following API should suffice.

* Base class with `factory` class method which accepts `abi` and returns a `ContractClass`
* `ContractClass` as returned by `BaseContract.factory` can be instantiated with a single `address`.
* A method on the `ContractClass` for constructing a transaction (TODO: should this be a dictionary, a `Message` class, or something else?)

There are likely some other convenience APIs that should be added around this functionality to make common tasks easier as we find what those are.

I think this code should live in something like `evm.utils.contracts`.

pipermerriam:
Made some updates to try and add some clarity to how this might be implemented.

pipermerriam:
@carver looks like @mhchia may be working on this but you may be well suited to assist with implementation or just sussing out the API.

carver:
Interesting, the contract API starts to feel a lot like web3.py: http://web3py.readthedocs.io/en/latest/web3.eth.html#web3.eth.Eth.contract 

Maybe we could rip out some part of web3 into a `eth-contracts` library, for code reuse. Either way, I'm happy to chat about API. As you can see in the docs, web3.py uses a `dict` when constructing a transaction.

pipermerriam:
I'm going to backpedal hard on my original suggestion since I think I may have missed something crutial.

@mhchia does this interaction happen over the JSON-RPC interface?

mhchia:
@pipermerriam 
Yes it is. Previously I used the `web3.py` contract object.
But after, I changed to generate the tx data of the contract call transaction on our own, using `eth_abi`. (referenced from [here](https://github.com/ethereum/eth-tester/blob/99dfc5a8f39988316f6bd6ede3c6942d8a2bb389/eth_tester/utils/math_contract.py))

Now performing a contract call looks like [this](https://github.com/mhchia/sharding/blob/rpc/sharding/rpc/vmc_handler.py#L101-L108), specifying the `function_name` and `args`.
[...]

PR #206, opened 2017-12-08
[WIP] Sharding vmc handler

mhchia:
(Sorry that this PR is huge because of the existing contract code and the previous works)
### What was wrong?
As described in https://github.com/ethereum/py-evm/issues/197, to support basic sharding implementation, we're going to deploy Validator Manager Contract(vmc) on the mainchain via RPC. Besides we need tester to test its functionality. And because the existing code base relies on `pyethereum`, we have to remove them and turn to use `py-evm` and other packages.
After the `mainchain_handler` and `shardchains_handler`(another part, which still needs to be implemented) done, we can design a daemon simultaneously handles mainchain and shardchains.

### How was it fixed?
* Ported the existing vmc from repo `ethereum/sharding`, along with its utils functions, to `py-evm` repo.
* Removed the dependencies on `pyethereum` to other packages, like `eth_utils`, `eth_keys`, `eth_abi`
* Two backends for handling mainchain: `ChainHandler` using `web3.py` to manipulate nodes through RPC, and `TesterChainHandler` using `eth_tester`(whose backend is `evm.chains.tester.MainnetTesterChain`)

#### TODO
- [x] Reflect on the code review
    - [x] remove `viper` as the dependency?
    - [x] hardcoded `SpuriousDragonTransaction`
    - [x] Don't use globals in `vmc_utils.py`
    - [x] `get_code` in `is_vmc_deployed` is commented out
       - [x] Add `TODO` in `is_vmc_deployed()`
       - [x] Fix `get_code`, which is not implemented in `PyEVMBackend` in `eth_tester`.(May need to wait after https://github.com/ethereum/eth-tester/pull/31 is accepted)
    - [x] to-be-compiled code should be outside the function in test_handler?
    - [x] use eth_tester provider over `TesterChainHandler`
    - [x] use `print` over `logging`
    - [x] use “bytes" over "hex encoded" address
    - [x] `viper_rlp_decoder_addr` code convention to cytoolz.pipe
    - [x] `deploy_and_deposit`, …, should be in tests instead of vmc_handler
    - [x] why catch import keys
    - [x] in comment, `Transaction` should be modified to `SpuriousDragonTransaction`
    - [x] list appends to `yields` + `to_tuple`
    - [x] parameter indentation changes in `deposit`, ...
    - [x] `eth_utils.address` to `eth_utils`
    - [x] `pytest.fixture`
    - [x] `eth_utils.encode_hex` to `evm.utils.encode_hex`
    - [x] `sha3` should not be aliased
    - [x] remove `if __name__`
    - [x] `KeyAPI` to `keys`
- [x] flake8


evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+from evm.utils.address import generate_contract_address
+
+from evm.vm.forks.spurious_dragon.transactions import (
+    SpuriousDragonTransaction,
+    SpuriousDragonUnsignedTransaction,
+)
+
+from evm.chains.sharding.mainchain_handler.config import GASPRICE
+
+sha3 = eth_utils.crypto.keccak
```
pipermerriam: You can drop the `crypto` and make this just `eth_utils.keccak`

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
...6109ac5260806109ec51101561013b5760016109cc5161044001526001610a0c516109cc5161046001376001610a0c5101610a0c5260216109cc51016109cc52610281565b60b86109ec5110156101d15760806109ec51036109cc51610440015260806109ec51036001610a0c51016109cc51610460013760816109ec5114156101ac5760807f01000000000000000000000000000000000000000000000000000000000000006001610a0c5101350410151558575b607f6109ec5103610a0c5101610a0c5260606109ec51036109cc51016109cc52610280565b60c06109ec51101561027d576001610a0c51013560b76109ec510360200352600051610a2c526038610a2c5110157f01000000000000000000000000000000000000000000000000000000000000006001610a0c5101350402155857610a2c516109cc516104400152610a2c5160b66109ec5103610a0c51016109cc516104600137610a2c5160b66109ec5103610a0c510101610a0c526020610a2c51016109cc51016109cc5261027f565bfe5b5b5b81516001018083528114156100a4575b5050601f6109ac511115155857602060206109ac5102016109005260206109005103610a0c5261010060016064818352015b6000610a0c5112156102d45761030a565b61090051610a0c516040015101610a0c51610900516104400301526020610a0c5103610a0c5281516001018083528114156102c3575b50506109cc516109005101610420526109cc5161090051016109005161044003f35b61000461033003610004600039610004610330036000f31b2d4f"), Transaction)
```
pipermerriam: Looks like `Transaction` needs to be changed to `SpuriousDragonTransaction`.

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
...ecode(eth_utils.decode_hex("0xf90237808506fc23ac00830330888080b902246102128061000e60003961022056600060007f010000000000000000000000000000000000000000000000000000000000000060003504600060c082121515585760f882121561004d5760bf820336141558576001905061006e565b600181013560f783036020035260005160f6830301361415585760f6820390505b5b368112156101c2577f010000000000000000000000000000000000000000000000000000000000000081350483602086026040015260018501945060808112156100d55760018461044001526001828561046001376001820191506021840193506101bc565b60b881121561014357608081038461044001526080810360018301856104600137608181141561012e5760807f010000000000000000000000000000000000000000000000000000000000000060018401350412151558575b607f81038201915060608103840193506101bb565b60c08112156101b857600182013560b782036020035260005160388112157f010000000000000000000000000000000000000000000000000000000000000060018501350402155857808561044001528060b6838501038661046001378060b6830301830192506020810185019450506101ba565bfe5b5b5b5061006f565b601f841315155857602060208502016020810391505b6000821215156101fc578082604001510182826104400301526020820391506101d8565b808401610420528381018161044003f350505050505b6000f31b2d4f"), SpuriousDragonTransaction)
```
pipermerriam: Is there a reason for storing this data as an encoded transaction rather than just it's fields.  It's going to be problematic to hard-code references to the `SpuriousDragonTransaction` since each new set of fork rules will tend to introduce a new transaction rlp class.

Would it work to instead store the individual field values?  Alternatively, we can rlp decode this into a list and just pull the parameters from that list.

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+
+viper_rlp_decoder_addr = eth_utils.to_checksum_address(
+    generate_contract_address(eth_utils.to_canonical_address(viper_rlp_decoder_tx.sender), 0)
```
pipermerriam: Maybe slightly more legible in the following format?

```python
from cytoolz import pipe

viper_rlp_decoder_addr = pipe(
    viper_rlp_decoder_tx.sender,
    eth_utils.to_canonical_address,
    functools.partial(generate_contract_address, nonce=0),
    eth_utils.to_checksum_address,
)
```

or without the `cytoolz.pipe`

```
viper_rlp_decoder_addr = eth_utils.to_checksum_address(
    generate_contract_address(
        eth_utils.to_canonical_address(viper_rlp_decoder_tx.sender),
        0,
    )
)
```

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+    '''
+    keys = KeyAPI()
+    signature = keys.ecdsa_sign(message, privkey)
+    v, r, s = signature.vrs
+    v += 27
+    signature_bytes = b''.join([item.to_bytes(32, 'big') for item in (v, r, s)])
+    return signature_bytes
+
+def get_valmgr_abi():
+    global _valmgr_abi, _valmgr_code
```
pipermerriam: I would very much like to avoid the use of globals.  Can we restructure this so that these values are not lazily computed or that it's an error if they are not populated?

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+import rlp
+
+from viper import compiler
+
+from eth_abi import (
+    decode_abi,
+    encode_abi,
+)
+
+from eth_keys import KeyAPI
```
pipermerriam: You can modify this to use `from eth_keys import keys`.  The `key` object will be an instance of `KeyAPI` that you can use directly.

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+    )
+    _valmgr_sender_addr = eth_utils.to_checksum_address(valmgr_sender_addr)
+    _valmgr_addr = valmgr_addr
+    _valmgr_tx = tx
+
+def mk_initiating_contracts(sender_privkey, sender_starting_nonce, gasprice=GASPRICE):
+    """Make transactions of createing initial contracts
+    Including rlp_decoder, sighasher and validator_manager
+    """
+    o = []
```
pipermerriam: In general I'd like to avoid single letter variable names.

This function however can be adjusted to use `eth_utils.to_list`.  All of the places where you `o.append` or `o += [...]` can be turned into `yield` statements and the resulting return from this function will be all of those values that were yielded in a list.

```python
for tx in (....):
    yield SpuriousDragonUnsignedTransaction(...)
yield viper_rlp_decoder_tx
yield sighasher_tx
...
```

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+    return o
+
+def get_func_abi(func_name, contract_abi):
+    for func_abi in contract_abi:
+        if func_abi['name'] == func_name:
+            return func_abi
+    raise ValueError('ABI of function {} is not found in vmc'.format(func_name))
+
+def mk_contract_tx_obj(
+        func_name,
```
pipermerriam: Same request to reformat for consistency.

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,212 @@
...
+        contract_abi,
+        sender_addr,
+        value,
+        gas,
+        gas_price):
+    func_abi = get_func_abi(func_name, contract_abi)
+    arg_types = [arg_abi['type'] for arg_abi in func_abi['inputs']]
+    func_selector = eth_utils.function_abi_to_4byte_selector(func_abi)
+    data = func_selector + encode_abi(arg_types, args)
+    data = eth_utils.encode_hex(data)
```
pipermerriam: I would suggest using the `encode_hex` function from `evm.utils.hexadecimal.encode_hex`.  The `eth_utils.encode_hex` function may suffer a backwards incompatible breaking change in the coming months.

tests/sharding/test_tester_chain_handler.py
```diff
@@ -0,0 +1,80 @@
...
+    mk_contract_tx_obj,
+)
+
+from evm.chains.sharding.mainchain_handler.backends.tester_chain_handler import (
+    TesterChainHandler,
+)
+
+keys = get_default_account_keys()
+
+@pytest.fixture(scope='function')
```
pipermerriam: `'function'` scope is the default and doesn't have to be specified.  Makes for slightly cleaner looking fixtures.

```python
@pytest.fixture
def thing(...):
    ...
```

tests/sharding/test_tester_chain_handler.py
```diff
@@ -0,0 +1,80 @@
...
+def chain_handler():
+    return TesterChainHandler()
+
+def test_tester_chain_handler(chain_handler):
+    chain_handler.mine(1)
+    code = """
+num_test: public(num)
```
pipermerriam: For readability can this block of code get moved to a constant at the top level of the module so that the indentation isn't broken up here?

tests/sharding/test_vmc_handler.py
```diff
@@ -0,0 +1,167 @@
...
+    RPCChainHandler,
+)
+
+from evm.chains.sharding.mainchain_handler.backends.tester_chain_handler import (
+    TesterChainHandler,
+)
+
+keys = get_default_account_keys()
+
+sha3 = eth_utils.crypto.keccak
```
pipermerriam: Can we remove this alias and just use `keccak`?

tests/sharding/test_vmc_handler.py
```diff
@@ -0,0 +1,167 @@
...
+def chain_handler():
+    return TesterChainHandler()
+
+def print_current_contract_address(sender_address, nonce):
+    list_addresses = [
+        eth_utils.address.to_checksum_address(
+            generate_contract_address(keys[0].public_key.to_canonical_address(), i)
+        ) for i in range(nonce + 1)
+    ]
+    print(list_addresses)
```
pipermerriam: Same suggestion about using the `logging` API.

tests/sharding/test_vmc_handler.py
```diff
@@ -0,0 +1,167 @@
...
+        primary_addr,
+    )
+    vmc_handler.chain_handler.mine(1)
+    assert vmc_handler.get_receipt_value(0) == 1234567
+
+    do_withdraw(vmc_handler, validator_index)
+    vmc_handler.chain_handler.mine(1)
+    assert vmc_handler.sample(shard_id) == zero_addr
+
+if __name__ == '__main__':
```
pipermerriam: This should be removed as these tests should be runnable using `$ py.test tests/sharding/test_vmc_handler.py`

hwwhww:
nitpick suggestion: use `"""` for docstring instead of `'''`

mhchia:
@pipermerriam Thank you so much for so many suggestions. I'm working on fixing them.
About the use of hexidecimal encoded address, I think you're right, using bytes is more reasonable and consistent to the py-evm codebase. Previously we also used bytes as the address type in `pyethereum`. The reason why I used hex encoded address in this PR was because that `web3.py` and `eth_tester` take hex address in parameters. I will change them to bytes, only transforming them to hex when necessary.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,263 @@
...
+class VMCHandler:
+
+    def __init__(self, chain_handler, primary_addr):
+        self.chain_handler = chain_handler
+        self.primary_addr = primary_addr
+        self.init_vmc_attributes()
+
+    def init_vmc_attributes(self):
+        self._vmc_addr = get_valmgr_addr()
+        print("!@# vmc_addr={}".format(self._vmc_addr))
```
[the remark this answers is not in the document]
mhchia: Thank you. I will do it.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,263 @@
...
+    def get_receipt_value(self, receipt_id, sender_addr=None):
+        if sender_addr is None:
+            sender_addr = self.primary_addr
+        return self.call_vmc('get_receipts__value', [receipt_id], sender_addr=sender_addr)
+
+    # utils #######################################################
+
+    def is_vmc_deployed(self):
+        return (
+            # self.chain_handler.get_code(self._vmc_addr) != b'' and \
```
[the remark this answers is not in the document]
mhchia: Sorry for that. This code should not be commented out. I did it because for now I haven't finished the `get_code` in `TesterChainHandler`, which calls `PyEVMBackend.get_code`. But that part hasn't been implemented yet. I think it should be fixed when I changed to use `web3.py` tester? I will check for it. :)

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def sample(self, shard_id, sender_addr=None):
+        '''sample(shard_id: num) -> address
+        '''
+        if sender_addr is None:
+            sender_addr = self.primary_addr
+        return self.call_vmc('sample', [shard_id], sender_addr=sender_addr)
+
+    def deposit(
+            self,
```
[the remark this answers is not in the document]
mhchia: No problem!

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,263 @@
+from eth_tester.exceptions import ValidationError
+
+import eth_utils
+
+from evm.utils.address import generate_contract_address
+
+from viper import compiler
```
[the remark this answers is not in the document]
mhchia: I think it makes sense to remove viper as the dependency.
However, we are still modifying the validator manager contract sometimes.
Is it a good idea to do that in this situation?

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: Yes, thank you for reminding.

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: You're right. I'm sorry for that.
I will modify it to `yield` values and use `@to_tuple` to decorate the function.

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: No problem.

tests/sharding/test_vmc_handler.py
mhchia: no problem:)

tests/sharding/test_vmc_handler.py
mhchia: Yep sorry it's my fault. I forgot to remove it.

tests/sharding/test_tester_chain_handler.py
mhchia: no problem

tests/sharding/test_tester_chain_handler.py
mhchia: No problem. Thank you!

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: Actually it is because the `viper_rlp_decoder_tx` is hardcoded in [viper](https://github.com/ethereum/viper/blob/master/viper/utils.py#L106-L111).
Is it help to accept the transaction type as the input to decode each hardcoded transaction and to create the vmc tx?

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: no problem

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: cool! no problem

mhchia:
Sorry I have to go out for a while and have to respond the rest of the suggestions and questions later. I will keep working on these and respond to the questions as soon as I can:)
Thank you again for pointing all the problems out!

evm/chains/sharding/mainchain_handler/vmc_utils.py
pipermerriam: I'll dig into this again on Monday.  I suspect that we can figure out a way to structure this so that it doesn't create a hard dependency on a specific RLP object.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,263 @@
...
+
+    def deploy_initiating_contracts(self, privkey):
+        if not self.is_vmc_deployed():
+            addr = privkey.public_key.to_checksum_address()
+            self.chain_handler.unlock_account(addr)
+            nonce = self.chain_handler.get_nonce(addr)
+            txs = mk_initiating_contracts(privkey, nonce)
+            for tx in txs[:3]:
+                self.chain_handler.direct_tx(tx)
+            self.chain_handler.mine(1)
```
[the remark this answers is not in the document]
mhchia: Yep you're right. I think adding an extra parameter to `direct_tx` is a good idea.
Besides, IMO both `VMCHandler.deploy_valcode_and_deposit` and `VMCHandler. deploy_initiating_contracts` should be moved to the tests. Does it make sense?

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,263 @@
...
+    def first_setup_and_deposit(self, key):
+        self.deploy_valcode_and_deposit(key)
+        # TODO: error occurs when we don't mine so many blocks
+        self.chain_handler.mine(SHUFFLING_CYCLE_LENGTH)
+
+    def import_key_to_chain_handler(self, key):
+        try:
+            self.chain_handler.import_privkey(key.to_hex())
+        except (ValueError, ValidationError):
+            pass
```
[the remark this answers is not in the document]
mhchia: Both of them happens when a key is already imported, thrown from `web3.py` and `eth_tester` respectively. It is my fault. I think this testing function should not be here and will move it to the test.

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: I think `pipe` is cool. I didn't use this function before in python. 👍 
May I use the `pipe` one over the other one?
Just wondering if that is okay that I just import the package for only this code snippet.

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: Is it legible that `create_valmgr_tx` creates the vmc tx every time it is called, and we provide functions `get_abi(vmc_tx)`, `get_addr(vmc_tx)`, `get_sender_addr(vmc_tx)`, ... to get these information from the created vmc tx?
Or I can provide another class like `VMC`, which creates the transaction once it is initialized, and support methods for people to get those information?

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: No problem. Thank you for pointing it out. For the hardcoded RLP object, I also opened an [issue](https://github.com/ethereum/viper/issues/522) in viper before. Maybe we can also figure out what is the better way?

mhchia: And for the hard-code to `SpuriousDragonTransaction`, is it also happening in https://github.com/ethereum/py-evm/pull/206/files#diff-c1cdc70a34f7db7292b7adabab409d99R122 ?
If yes, is it helping that we have inject the dependency from outside, i.e. make the Transaction class as a parameter for `create_valmgr_tx`?
like
```python
+def create_valmgr_tx(gasprice=GASPRICE, TransactionClass):
    ...
    tx = TransactionClass(0, gasprice, 3000000, b'', 0, bytecode, v, r, s)
    ...
```

evm/chains/sharding/mainchain_handler/vmc_utils.py
pipermerriam: `cytoolz` is already a dependency and yes, it's your call what version you'd like to use.

evm/chains/sharding/mainchain_handler/vmc_utils.py
pipermerriam: I'll have to get back to you monday when I have some time to digest this and figure out where a good place to store/generate this stuff would be.

hwwhww:
Hint: Modify `tox.ini` to make `flake8` checking excluding contract.
https://github.com/ethereum/py-evm/blob/master/tox.ini#L42
`flake8 {toxinidir}/evm --exclude="./chains/sharding/contracts/"`

evm/chains/sharding/mainchain_handler/config.py
```diff
@@ -0,0 +1,9 @@
+# FIXME: should be 25 or 2500, fix it, only for faster testing
```
pipermerriam: This suggests that this  should be a default that's can be overridden in the places it's used.

evm/chains/sharding/mainchain_handler/config.py
```diff
@@ -0,0 +1,9 @@
...
+SHUFFLING_CYCLE_LENGTH = 5
+DEPOSIT_SIZE = 10 ** 20
+PERIOD_LENGTH = 5
+
+TX_GAS = 510000
+GASPRICE = 1
+PASSPHRASE = '123'
+DEFAULT_RPC_SERVER_URL = 'http://localhost:8545'
```
pipermerriam: I'm thinking that this and `PASSPHRASE` don't belong as constants.  Currently it looks like this is primarily a testing configuration value.  If you modify the `MainChainHandler` class to instead take a fully instantiated `Web3` instance then we gain the flexibility to run tests how we wish as well as allow use of any style of web3 connection.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def get_code(self, address):
+        address = to_checksum_address(address)
+        return self._w3.eth.getCode(address)
+
+    def get_nonce(self, address):
+        address = to_checksum_address(address)
+        return self._w3.eth.getTransactionCount(address)
+
+    def import_privkey(self, privkey, passphrase=PASSPHRASE):
```
pipermerriam: This looks like it's only used for testing.  If so, what do you think about doing this setup directly using the web3 instances during test setup.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def sample(self, shard_id, sender_addr=None):
+        """sample(shard_id: num) -> address
+        """
+        if sender_addr is None:
+            sender_addr = self.primary_addr
+        address_in_hex = self.call_vmc('sample', [shard_id], sender_addr=sender_addr)
+        # TODO: should see if there is a better way to automatically change the address result from
+        #       hex to bytes in. Maybe in `decode_contract_call_result`?
+        return to_canonical_address(address_in_hex)
```
pipermerriam: Given that we *know* that `address_in_hex` is well formed we can instead just use `evm.utils.hexadecimal.decode_hex` to convert to bytes.

tests/sharding/test_mainchain_handler.py
```diff
@@ -0,0 +1,76 @@
...
+num_test: public(num)
+
+@public
+def __init__():
+    self.num_test = 42
+
+@public
+def update_num_test(_num_test: num):
+    self.num_test = _num_test
+"""
```
pipermerriam: I would suggest pulling this block of viper code out to the module level so that it doesn't interupt the indentation of this function body.

pipermerriam:
Question.  Is there a reason you are not using the `web3.eth.contract` API?  It would allow you to do away entirely with `VMHandler.call_vmc` and `VMCHandgler.send_vmc_tx` as well as the majority of the rest of that class if the default mechanisms for calling contract methods are suitable.

evm/chains/sharding/mainchain_handler/config.py
mhchia: Get it. Since `SHUFFLING_CYCLE_LENGTH` is going to be deprecated in the new validator manager contract, I moved this to the test for explicit specification. Is it okay?

evm/chains/sharding/mainchain_handler/config.py
mhchia: Thank you for pointing this out. I moved `PASSPHRASE` to the tests and make `Web3` instance as the argument of the constructor of `MainchainHandler`.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
mhchia: Thank you! I remove the default value and force the passphrase to be specified when it is called

evm/chains/sharding/mainchain_handler/vmc_handler.py
mhchia: Have done it:) Thank you.

tests/sharding/test_mainchain_handler.py
mhchia: Done it. Will it even helpful that I replace the code with compiled bytecode, thus we can remove the dependency on viper?

mhchia:
@pipermerriam Sorry for the late response. At that time I was working on `send_raw_transaction` so I didn't use the API right away. I have used `web3.eth.contract` now. However, I still use `VMCHandler._mk_contract_tx_detail` to format the `tx_detail` which is passed to `contract.call()` and `contract.transact()`, for the purpose of preprocessing of transaction details. Is it okay to do that? or should I just remove that and inline all of them when doing the calls?

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,240 @@
...
+    def setup_vmc_instance(self):
+        self._vmc = self.mainchain_handler.contract(
+            to_checksum_address(self._vmc_addr),
+            self._vmc_abi,
+            self._vmc_bytecode,
+        )
+
+    # vmc utils #####################################################
+
+    def get_vmc(self):
```
pipermerriam: This getter method seems like it could be removed and we could simply move `self._vmc` to `self.vmc` and allow access as a public property.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,240 @@
...
+        if value is not None:
+            yield 'value', value
+        if gas_price is not None:
+            yield 'gas_price', gas_price
+        if data is not None:
+            yield 'data', data
+
+    # utils #######################################################
+
+    def is_vmc_deployed(self):
```
pipermerriam: Currently this is only used in tests.  If this is a testing only API I think it should be removed from this class.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def get_vmc(self):
+        """
+        :return: web3.eth.contract instance of vmc
+        """
+        return self._vmc
+
+    @to_dict
+    def _mk_contract_tx_detail(self,
+                               sender_addr=None,
```
pipermerriam: Since `sender_addr` is explicitly checked to not be `None` I think that this default should be removed.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,240 @@
...
+                               gas=TX_GAS,
+                               value=None,
+                               gas_price=None,
+                               data=None):
+        # Both 'from' and 'gas' are required in eth_tester
+        if sender_addr is None:
```
pipermerriam: This check should potentially be changed to check that `is_canonical_address(sender_addr)`

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,240 @@
...
+)
+
+from evm.chains.sharding.mainchain_handler.vmc_utils import (
+    create_vmc_tx,
+    get_contract_address_from_contract_tx,
+    get_vmc_json,
+)
+
+
+class VMCHandler:
```
pipermerriam: I'd like to suggest a different approach now that I'm starting to see how this is coming together.

Instead of writing a class that delegates to a web3 contract instance, I think it might be better to simply subclass the web3 contract class to add the necessary business logic.

```python
from web3.contract import Contract

class VMC(Contract):
    abi = <the-vmc-abi>
    bytecode = <the-vmc-bytecode>

    def sample(self, shard_id, sender_addr=None, gas=TX_GAS):
        txn_params = self.mk_txn_params(sender=sender, gas=gas)
        address_as_hex = self.call(tx_params).sample(shard_id)
        return decode_hex(address_in_hex)
```

I think this results in a cleaner and easier to read API, removing the initialization code.  Thoughts?

mhchia:
@pipermerriam I modified the `vmc_utils` to remove globals and removed dependencies on `viper`, by precompile all contracts. Could you help me to remove them if you have any spare time? Thank you so much!
Besides, may I ask how to add `web3` and `eth_tester` as the dependencies in the `install_requires` in `setup.py`? I added them but conflict happened, so for now I removed them.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def get_code(self, address):
+        address = to_checksum_address(address)
+        return self._w3.eth.getCode(address)
+
+    def get_nonce(self, address):
+        address = to_checksum_address(address)
+        return self._w3.eth.getTransactionCount(address)
+
+    def import_privkey(self, privkey, passphrase):
```
pipermerriam: Assuming `import_privkey` is an API that is only used in testing, I think it (and any others that are **only** for facilitating testing) should be removed.  We can still add a utiity to do these things elsewhere.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,102 @@
...
+
+from evm.chains.sharding.mainchain_handler.config import (
+    GASPRICE,
+    TX_GAS,
+)
+
+
+class MainchainHandler:
+
+    def __init__(self, web3_instance, use_eth_tester=False):
```
pipermerriam: I'm also thinking that the `use_eth_tester` param should be removed.  I see that there are some conditional logic forks in place, however, I believe all of them are in methods that are specifically for testing purposes and I've proposed removing those from this object.

evm/chains/sharding/mainchain_handler/vmc_handler.py
mhchia: Cool! I think it is better. Most of the new logics dealing with mainchain will be implemented in MainchainHandler. I think inheritance should be enough for `VMC`. I'm modifying this part. Thank you.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
mhchia: No problem. Will handle the tester logics outside the class, in the tests, for now.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
mhchia: You're right. I am moving it to the tests.

mhchia:
@pipermerriam Could you help me to review the recent changes? They are mainly about removing the unneeded contracts. And I correct the web3.py version therefore the failing build issue is solved. Thank you so much.

tox.ini
```diff
@@ -39,7 +40,7 @@ basepython =
 basepython=python
 deps=flake8==3.5.0
 commands=
-    flake8 {toxinidir}/evm
+    flake8 {toxinidir}/evm --exclude="{toxinidir}/evm/chains/sharding/contracts/"
```
pipermerriam: I'm assuming this is to avoid linting the viper files.  I would suggest changing this to `--exclude="*.v.py"`

evm/chains/sharding/mainchain_handler/config.py
```diff
@@ -0,0 +1,6 @@
+DEPOSIT_SIZE = 10 ** 20
+PERIOD_LENGTH = 5
+
+TX_GAS = 510000
+GASPRICE = 1
+DEFAULT_RPC_SERVER_URL = 'http://localhost:8545'
```
pipermerriam: This doesn't seem to be used anywhere.  Can you remove it?

evm/chains/sharding/mainchain_handler/geth_shell_scripts/launch_geth.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
pipermerriam: I question the inclusion of this file/script. Why is it present?  Why are we using geth?  I was somewhat assuming that we'd be interfacing with py-evm for this?

Mostly just looking for an explanation so that I can better understand the context of including these files.

tests/sharding/mainchain_handler/test_mainchain_handler.py
```diff
@@ -0,0 +1,63 @@
...
+def __init__():
+    self.num_test = 42
+
+@public
+def update_num_test(_num_test: num):
+    self.num_test = _num_test
+"""
+
+
+def test_tester_chain_handler(mainchain_handler):  # noqa: F811
```
pipermerriam: This test is titled `test_tester_chain_handler` but it's using the `mainchain_handler`.  Is that intentional?

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,78 @@
...
+    def mine(self, number):
+        expected_block_number = self.get_block_number() + number
+        self.w3.miner.start(1)
+        while self.get_block_number() < expected_block_number:
+            time.sleep(0.1)
+        self.w3.miner.stop()
+
+    def unlock_account(self, account, passphrase):
+        account = to_checksum_address(account)
+        self.w3.personal.unlockAccount(account, passphrase)
```
pipermerriam: Unlocking accounts is generally considered *bad*, especially over RPC where *anything* can get at it.  What do you think about removing the account unlocking and instead doing one of the following.

- Issue transactions over `web3.personal.sendTransaction`
- Using the [`web3.eth.account`](http://web3py.readthedocs.io/en/latest/web3.eth.account.html) API to sign transactions locally and send them over `web3.eth.sendRawTransaction`.  This should allow for code that is **significantly** more robust and should work more reliably across different backends.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,135 @@
...
+            yield 'data', data
+
+    # contract calls ##############################################
+
+    def sample(self, shard_id, sender_addr=None, gas=TX_GAS):
+        """sample(shard_id: num) -> address
+        """
+        tx_detail = self.mk_contract_tx_detail(sender_addr=sender_addr, gas=gas)
+        address_in_hex = self.call(tx_detail).sample(shard_id)
+        # TODO: should see if there is a better way to automatically change the address result from
```
pipermerriam: This comment looks like it can be removed.

tests/sharding/mainchain_handler/fixtures.py
```diff
@@ -0,0 +1,55 @@
...
+)
+
+from web3.providers.eth_tester import EthereumTesterProvider
+
+from eth_tester import EthereumTester
+
+from eth_tester.backends.pyevm import PyEVMBackend
+
+# from evm.chains.sharding.mainchain_handler.config import (
+#     DEFAULT_RPC_SERVER_URL,
```
pipermerriam: Remove?

tests/sharding/mainchain_handler/fixtures.py
```diff
@@ -0,0 +1,55 @@
+import pytest
+
+import rlp
+
+from web3 import (
+    Web3,
+    # HTTPProvider,
```
pipermerriam: remove?

tests/sharding/mainchain_handler/fixtures.py
```diff
@@ -0,0 +1,55 @@
...
+    def mine(self, number):
+        self.w3.testing.mine(number)
+
+    def direct_tx(self, tx):
+        raw_tx = rlp.encode(tx)
+        raw_tx_hex = self.w3.toHex(raw_tx)
+        try:
+            tx_hash = self.w3.eth.sendRawTransaction(raw_tx_hex)
+        except ValueError:
+            # FIXME: if `sendRawTransaction` is not implemented, `ValueError` is raised
```
pipermerriam: You should be good to go using `sendRawTransaction` directly with the latest `eth-tester` release.

tests/sharding/mainchain_handler/fixtures.py
```diff
@@ -0,0 +1,55 @@
...
+    # TODO: currently we only test with `TesterChainHandler` because it takes time to test with
+    #       real RPC.
+    #       Should see if there is a better way to test with RPCHandler.(maybe mock web3.py?)
+    eth_tester = EthereumTester(
+        backend=PyEVMBackend(),
+        auto_mine_transactions=False,
+    )
+    provider = EthereumTesterProvider(eth_tester)
+    # provider = HTTPProvider(DEFAULT_RPC_SERVER_URL)
+    web3_instance = Web3(provider)
```
pipermerriam: A convention we've been using is to use the variable name `w3` for web3 instances.

tests/sharding/mainchain_handler/fixtures.py
```diff
@@ -0,0 +1,55 @@
...
+            #        `self._eth_tester.backend.chain.apply_transaction`
+            tx_hash = self.w3.providers[0].ethereum_tester.backend.chain.apply_transaction(tx)
+        return tx_hash
+
+
+@pytest.fixture
+def mainchain_handler():
+    # TODO: currently we only test with `TesterChainHandler` because it takes time to test with
+    #       real RPC.
+    #       Should see if there is a better way to test with RPCHandler.(maybe mock web3.py?)
```
pipermerriam: I don't think you should need to test separately with the HTTPProvider.  Maybe remove this for now and focus on getting everything clean and working with the tester provider and then we can tackle geth/http-rpc integration separately?

tests/sharding/mainchain_handler/test_vmc_handler.py
```diff
@@ -0,0 +1,303 @@
...
+    signature = vmc_utils.sign(vmc_utils.WITHDRAW_HASH, privkey)
+    vmc_handler.withdraw(validator_index, signature, sender_addr)
+    chain_handler.mine(1)
+
+
+def deploy_valcode_and_deposit(vmc_handler, chain_handler, key):
+    """
+    Deploy validation code of and with the key, and do deposit
+
+    :param key: Key object
```
pipermerriam: Assuming this means `PrivateKey` object?

tests/sharding/mainchain_handler/test_vmc_handler.py
```diff
@@ -0,0 +1,303 @@
...
+        tx_list_root,
+        collation_coinbase,
+        post_state_root,
+        receipt_root,
+        number,
+        sig,
+    ])
+
+
+def test_vmc_contract_calls(mainchain_handler):  # noqa: F811
```
pipermerriam: This is a very long test body.  At minimum I think it would benefit from adding as many comment blocks as possible that read as a narrative of what is happening and why we are doing it.

evm/chains/sharding/mainchain_handler/geth_shell_scripts/launch_geth.sh
mhchia: You're right. I should not put the shell scripts here. I will remove them.
The idea is
> (ii) a "shard client" which talks to the main chain node via RPC (assumed to be trusted because it's also running on the user's computer)
in https://github.com/ethereum/sharding/blob/develop/docs/doc.md#introduction
So I should not limit the code to use only geth.

tests/sharding/mainchain_handler/fixtures.py
mhchia: May I remove this after https://github.com/ethereum/web3.py/pull/520 is merged? Otherwise we might probably get errors.

tests/sharding/mainchain_handler/test_vmc_handler.py
mhchia: Sure. Just want to confirm should I keep this long body, and just add enough comment blocks?
Because many of the steps depends on the previously made steps, like `deposit` + `mine(SHUFFLING_CYCLE_LENGTH)`. Originally because I test with two backends, `geth` and eth_tester, it takes too much time when testing with `geth`. So I compile the tests altogether. 
If we remove the logic of testing with `geth` from `test_vmc_handler.py`(thanks to your suggestion in the above), I can set the dependent steps as the fixture, and separate the long test body of `test_vmc_contract_calls` to multiple tests.

tox.ini
mhchia: Will do. Thank you for the reminder

tests/sharding/mainchain_handler/test_vmc_handler.py
mhchia: Yep it is. I will change it. Thank you:)

tests/sharding/mainchain_handler/test_mainchain_handler.py
mhchia: No it is not intentional. sorry I will rename it.

tests/sharding/mainchain_handler/fixtures.py
mhchia: No problem I will change it.

tests/sharding/mainchain_handler/fixtures.py
mhchia: Cool. This will be cleaner. I just saw the tests with geth/http-rpc in web3.py. Maybe we can integrate them in the same way in the next steps?

evm/chains/sharding/mainchain_handler/mainchain_handler.py
mhchia: @pipermerriam  I think you're right. It is more dangerous to use `unlockAccount` over other methods.
I'd like to change to any of the two solutions. However, it seems inevitable that we use `unlockAccount` when we do `eth.contract.transact()` using `eth.contract` if I don't understand it wrong.
In this condition can we get rid of `unlockAccount` with any solutions?
Thank you so much!

tests/sharding/mainchain_handler/test_vmc_handler.py
pipermerriam: Yes, for now we can keep the long body as long as it's well commented so that it's easier to follow.  I suspect we can trim this down into smaller pieces at some point but doesn't need to be now.

tests/sharding/mainchain_handler/fixtures.py
pipermerriam: Yes, lets see if we can't get @carver to merge and release that today.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
pipermerriam: I think that we can still get around this using the following APIs that I *think* are all avialable in web3.py today.

* Use `web3.eth.contract.buildTransaction()` to build a transaction dictionary
* Use `web3.eth.account.signTransaction()` to convert the txn dict to a raw transaction.
* Use `web3.eth.sendRawTransaction()` to send the transaction.

This approach should work unilaterally across all of the different RPC servers out there (with the assumption that the fixes for `eth-tester:py-evm` are in place.)

mhchia: Wow it's cool. I didn't even know this method `buildTransaction`. I Will try it right away. Really appreciate!

mhchia:
@pipermerriam I added a method `send_tx` in `VMC` and replaced all transaction sending contract calls with `send_tx`, to get rid of the `unlockAccount`. However, I was still struggling what is the better way to replace `web3.contract.call()`. Since `sendRawTransaction` actually sends the transaction. Do you have any idea about this?
Thank you!

pipermerriam:
> However, I was still struggling what is the better way to replace `web3.contract.call()`

It's not clear to me why you would need to replace this.  The need to replace `web3.contract.sendTransaction` was based on the need for the node to contain an unlocked account for signing.  `web3.eth.call` has no such requirement.  You should be ok just directly calling this with any transaction, *or* by proxy via `some_contract.call().someMethod()`

pipermerriam:
paging @carver 

Looks like current test failures are due to an older version of Web3.py in which the `eth-tester` provider still doesn't have `sendRawTransaction` implemented.  Is there a release that contains these changes that @mhchia can use?

evm/chains/sharding/mainchain_handler/config.py
```diff
@@ -0,0 +1,5 @@
+DEPOSIT_SIZE = 10 ** 20
```
pipermerriam: It would be good to have a comment explaining this *magic* number.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,78 @@
...
+
+from evm.chains.sharding.mainchain_handler.config import (
+    GASPRICE,
+    TX_GAS,
+)
+
+
+class MainchainHandler:
+
+    def __init__(self, web3_instance):
```
pipermerriam: *naming nitpick*

Can you rename the constructor variable from `web3_instance` to `w3` for consistency.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def mine(self, number):
+        expected_block_number = self.get_block_number() + number
+        self.w3.miner.start(1)
+        while self.get_block_number() < expected_block_number:
+            time.sleep(0.1)
+        self.w3.miner.stop()
+
+    def unlock_account(self, account, passphrase):
+        account = to_checksum_address(account)
```
pipermerriam: This is a pattern we have tried to avoid (overwriting variables).  In situations where an error occurs and you've overwritten a variable you lose the ability to see what the original value was.  I would suggest one of the following naming schemes.

- `raw_account/account`
- `account/checksummed_account`

*or* you can change the next line to `self.w3.personal.unlockAccount(to_checksum_address(account), ...)`

evm/chains/sharding/mainchain_handler/mainchain_handler.py
```diff
@@ -0,0 +1,78 @@
...
+        return self.w3.eth.getTransactionCount(address)
+
+    def mine(self, number):
+        expected_block_number = self.get_block_number() + number
+        self.w3.miner.start(1)
+        while self.get_block_number() < expected_block_number:
+            time.sleep(0.1)
+        self.w3.miner.stop()
+
+    def unlock_account(self, account, passphrase):
```
pipermerriam: Your latest comment suggested this was being removed.

mhchia: sorry for that. I have removed it.

evm/chains/sharding/mainchain_handler/mainchain_handler.py
mhchia: yep no problem.

evm/chains/sharding/mainchain_handler/config.py
mhchia: Have done it:) Thank you

evm/chains/sharding/mainchain_handler/mainchain_handler.py
mhchia: Thank you for the reminder. I will avoid this kind of overwriting variables. Have changed them.

mhchia:
@pipermerriam 
> It's not clear to me why you would need to replace this.
Sorry for the misleading. I just found that we can just do `contract.call()` with the address which is already in `eth.accounts` without performing `unlockAccount`
Before I thought `unlockAccount` were needed because when I changed the addresses given to `contract.call()`, errors occur. And it's just because `web3.HTTProvider` allows arbitrary address in `contract.call()`, `web3.EthereumTester` does not. The call in `EthereumTester` actually needs the corresponding key to sign the transaction, and `revert` after the transaction is applied.
So it is fine now. I just left the `contract.call` not modified, and add `send_transaction` for crafting transactions, to get rid of `unlockAccount`.
Thank you again for your help!

carver:
> Is there a release that contains these changes that @mhchia can use?

Not yet. 80% chance I can get it out today, 97% by tomorrow. Long tail of unknown unknowns.

pipermerriam:
> The call in EthereumTester actually needs the corresponding key to sign the transaction

This is a bug and should be fixed.  I'm opening an issue now to track it.

pipermerriam:
https://github.com/ethereum/eth-tester/issues/38

carver:
@mhchia v4-beta5 is released, should be good to go

mhchia:
@carver great! Thank you so much.

mhchia:
@pipermerriam I get it. Thank you!
I have one question is, I'm considering to remove `MainchainHandler`, is it okay that I just inline the logic inside to `test_vmc_handler.py`?

pipermerriam:
@mhchia 

> I'm considering to remove `MainchainHandler`, is it okay that I just inline the logics inside to `test_vmc_handler.py`?

Yes, I think this is a good simplification given how thin a wrapper it has become.

evm/chains/sharding/mainchain_handler/vmc_handler.py
```diff
@@ -0,0 +1,186 @@
...
+            yield 'value', value
+        if gas_price is not None:
+            yield 'gasPrice', gas_price
+        if data is not None:
+            yield 'data', data
+
+    def send_transaction(self,
+                         func_name,
+                         args,
+                         privkey,
```
pipermerriam: Seeing the private key passed around makes me think it would be cleaner if this was set as a property on the contract instance.

You can override the `__init__` as follows:

```python
class VMC(Contract):
    def __init__(self, *args, private_key, **kwargs):
        self.private_key = private_key
```

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,135 @@
...
+from evm.utils.address import (
+    generate_contract_address,
+)
+from evm.utils.hexadecimal import (  # use evm.utils over eth_utils for hex encode/decode
+    decode_hex,
+)
+
+from evm.chains.sharding.mainchain_handler.config import GASPRICE
+
+WITHDRAW_HASH = keccak("withdraw")
```
pipermerriam: This should be changed to `keccak(b"withdraw")` as well as changing to use `evm.utils.keccak.keccak` which only accepts bytes inputs.  Reason being that `evm.utils.keccak` is going to change in the near term future in a breaking way that will cause this way of calling it to break.

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,135 @@
...
+    return create_transaction_from_hex(viper_rlp_decoder_tx_hex, TransactionClass)
+
+
+@to_tuple
+def mk_initiating_contracts(sender_privkey,
```
pipermerriam: Since this is only used in testing, what do you think about moving it into the appropriate module under `tests/`

evm/chains/sharding/mainchain_handler/vmc_utils.py
```diff
@@ -0,0 +1,135 @@
...
+        functools.partial(rlp.decode, sedes=TransactionClass),
+    )
+
+
+def create_sighasher_tx(TransactionClass):
+    sighasher_tx_hex = "0xf9016d808506fc23ac0083026a508080b9015a6101488061000e6000396101565660007f01000000000000000000000000000000000000000000000000000000000000006000350460f8811215610038576001915061003f565b60f6810391505b508060005b368312156100c8577f01000000000000000000000000000000000000000000000000000000000000008335048391506080811215610087576001840193506100c2565b60b881121561009d57607f8103840193506100c1565b60c08112156100c05760b68103600185013560b783036020035260005101840193505b5b5b50610044565b81810360388112156100f4578060c00160005380836001378060010160002060e052602060e0f3610143565b61010081121561010557600161011b565b6201000081121561011757600261011a565b60035b5b8160005280601f038160f701815382856020378282600101018120610140526020610140f350505b505050505b6000f31b2d4f"  # noqa: E501
+    return create_transaction_from_hex(sighasher_tx_hex, TransactionClass)
+
+
+def create_viper_rlp_decoder_tx(TransactionClass):
```
pipermerriam: Since this is only used in testing, what do you think about moving it into the appropriate module under `tests/`

mhchia:
@pipermerriam 
> It appears that evm.chaints.sharding.mainchain_handler.vmc_utils.py is only used for testing. Can we move it, or at least all of the things from within it into an appropriate location under the tests/ directory?
You're right. I'm fine with it and will move it under `tests`.
I have one concern is about `create_vmc_tx`. It is used to specify the VMC transaction. In this case should we put it under `tests/`? If not, may I leave `create_vmc_tx` along with `get_vmc_json` in `evm.chaints.sharding.mainchain_handler.vmc_utils.py`, and move all the other functions to the module under `tests/`?
Big thanks for the review!

evm/chains/sharding/mainchain_handler/vmc_handler.py
mhchia: No problem. I will change it.

evm/chains/sharding/mainchain_handler/vmc_utils.py
mhchia: No problem I will change it. Thank you:)

mhchia:
@pipermerriam I haven't yet moved logics from `vmc_utils.py` to the a module under `tests/`. I want to do it after getting feedbacks from you about my questions. And there is no hurry at all, please review it after the vacation.
Thank you so much for all the feedbacks! They really help a lot for the cleaner code.
Happy vacation:)!

pipermerriam:
Sorry for the delay.  Just getting back from holiday and playing catch up.

> If not, may I leave create_vmc_tx along with `get_vmc_json` in `evm.chaints.sharding.mainchain_handler.vmc_utils.py,` and move all the other functions to the module under `tests/`?

Yes, this seems fine to me.

mhchia:
@pipermerriam It seems error is solved if I specify the version of `py-trie` to `0.3.2`. Should I do this for now or just wait until `pytrie` is fixed? :)

mhchia:
@pipermerriam Besides, I'm wondering if it would be better to put the files in this PR from the directory `evm.chains.sharding` to `evm.vm.forks.sharding`. Do you have any suggestion?
Thank you so much

pipermerriam:
> It seems error is solved if I specify the version of py-trie to 0.3.2. Should I do this for now or just wait until pytrie is fixed? :)

#249 fixes this.  I'll be rebasing the sharding branch as soon as that is merged.

mhchia:
The tests for sharding fails because eth-tester fails resulting from the modification of VM class. I made a PR https://github.com/ethereum/eth-tester/pull/41 , trying to solve the problem.

pipermerriam:
https://github.com/ethereum/web3.py/pull/541 fixes this at the web3.py level.

If you'd like to shortcut waiting for a web3.py release feel free to directly add the proper `eth-tester` dependency requirements.

mhchia:
@pipermerriam Thank you for the help!
Since the tests are resolved, may I merge it now and make another PR base on this?

pipermerriam:
I went ahead and squashed-and-merged given the extensive commit history.

mhchia:
Thank you:)

PR #213, opened 2017-12-12
Account redesign

jannikluhn:
This PR adds an alternative way of storing state as described in #191.

It depends on two not yet implemented trie methods:

- `Trie.delete_branch(key)` that should delete all keys that start with `key` (to delete the whole storage of an account)
- `Trie.branch_exists(key)` that checks if there is a key that starts with `key` (to check if an account exists)

State is always stored RLP encoded as that's the way it used to be in the two layer design. However, I don't really see a good motivation for that anymore as all entries are either integers (nonces, balances, storage entries) or binary (code), whereas in the old design we had a slightly more complex object (account).

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -25,7 +25,7 @@ from evm.db.journal import (
 )
 from evm.db.state import (
     State,
-    TwoLayerBackend,
+    NestedTrieBackend,
 )
 from evm.rlp.headers import (
     BlockHeader,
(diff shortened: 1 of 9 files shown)


NIC619:
I will add the two trie methods into binary trie.

jannikluhn:
This should now be ready, pending the two mentioned Trie methods.

jannikluhn:
Rebased. Will merge as soon as the two new trie methods have been implemented.

jannikluhn:
Here the hopefully final changes as discussed on Skype:
- renamed the trie backends to "flat" and "nested" trie backend
- raise `NotImplementedError`s for `delete_account`, `delete_storage` and `account_exists` (we probably don't need them)
Since this also removes the need for the two additional trie methods, this can be merged once reviewed.

Issue #212, opened 2017-12-11
Access prefix list redundancies


jannikluhn:
### What is wrong?

In an access prefix list (see https://ethresear.ch/t/account-read-write-lists/285), some prefixes can be unnecessary: If there is a prefix list consisting of `'eth'` and `'ethereum'`, everything that matches `'ethereum'` matches `'eth'` too. Therefore, `'ethereum'` can be safely removed from the list. This can potentially improve efficiency of access checks.

### How can it be fixed

Remove such redundancies from access prefix lists in `evm.utils.state_access_restrictions.to_prefix_list_form`.

gitcoinbot:
__This issue now has a funding of 0.121 ETH (76.91 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/212).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/212)

adtac:
I'd like to work on this! Getting started now.

adtac:
Here's a modified `to_prefix_list_form` that removes redundant prefixes before appending to `o`:

```diff
+def remove_redundant_prefixes(prefix_list):
+    class Node:
+        def __init__(self, char):
+            self.char = char
+            self.children = {}
+
+    root = Node(None)
+
+    result_list = []
+    for prefix in sorted(prefix_list, key=lambda prefix: -len(prefix)):
+        cur = root
+        created = False
+        for c in prefix:
+            if c not in cur.children:
+                cur.children[c] = Node(c)
+                created = True
+            else:
+                created = False
+            cur = cur.children[c]
+
+        if created:
+            result_list.append(prefix)
+    
+    return result_list
+
 def to_prefix_list_form(access_list):
     o = []
     for obj in access_list:
         o.append(sha3(addr) + b'\x00')
         o.append(sha3(addr) + b'\x01')
         o.append(sha3(addr) + b'\x02')
-        for prefix in storage_prefixes:
+        for prefix in remove_redundant_prefixes(storage_prefixes):
             o.append(sha3(addr) + b'\x03' + prefix)
     return o
```

owocki:
@adtac mind claiming this on gitcoin.co/explorer real quick?

adtac:
@owocki claimed with the google form option! (because of the metamask issue we were discussing the other day)

[...]

PR #217, opened 2017-12-13
Remove redundant prefixes in prefix list form

adtac:
This operates in a O(n*logn) time complexity because of the sort.

Closes https://github.com/ethereum/py-evm/issues/212

### What was wrong?

Previously, `eth` and `ethereum` were both allowed to be prefixes, for example. Since every string covered by `ethereum` is also covered by `eth`, just `eth` is sufficient.

### How was it fixed?

With a trie (based on nested dicts), the strings are inserted in the order of ascending length by keeping track of which nodes are the end of a string.

#### Cute Animal Picture

koala

A koala because I'm a maintainer at [coala](https://coala.io/) :)


evm/utils/state_access_restriction.py
```diff
@@ -23,6 +23,32 @@ def is_accessible(address, slot_as_key, access_prefix_list):
     return False
 
 
+def remove_redundant_prefixes(prefix_list):
+    """Given a list of strings, this removes redundant strings that are covered
+    by other strings in the list. For example, given `["eth", "ethereum"]`,
+    this function will return just `["ethereum"]` as every string with
```
jannikluhn: This is the wrong way around: `"eth"` should stay (more general), `"ethereum"` should be removed (more specific).

evm/utils/state_access_restriction.py
```diff
@@ -23,6 +23,32 @@ def is_accessible(address, slot_as_key, access_prefix_list):
...
+    `ethereum` prefix will also have `eth` as a prefix.
+    """
+    root = {}
+
+    result_list = []
```
pipermerriam: can you update this function to use the `to_set` decorator so that we can remove this list.  You can replace the `result_list.append(prefix)` with `yield prefix`.

adtac:
@jannikluhn @pipermerriam I've amended the commit with changes+tests.

evm/utils/state_access_restriction.py
adtac: Derp, dunno how I misunderstood the whole thing. Anyway, I've fixed this and added some tests as a sanity check.

pipermerriam:
@adtac I think we're really close.

We should still be able to do-away with the `Node` class if we go back to nested dictionaries and interpret an empty dictionary to be a word end.

evm/utils/state_access_restriction.py
```diff
@@ -23,6 +23,40 @@ def is_accessible(address, slot_as_key, access_prefix_list):
...
+            self.children = {}
+
+    root = Node(None)
+
+    for prefix in sorted(prefix_list, key=lambda prefix: len(prefix)):
+        cur = root
+        include = True
+
+        i = 0
+        while i < len(prefix):
```
pipermerriam: I think swapping this out for `for i in range(len(prefix))` will be more legible and won't require manually incrementing `i`.

adtac:
@pipermerriam Moved back to plain dicts with a `None` key marking word end. Also using a range instead of a `while` loop.

pipermerriam:
Looks good. I did one bit of reformating and rebasing to get tests passing.  Closing in favor of https://github.com/ethereum/py-evm/pull/219 which just has one commit on top of yours.  Thanks for getting this working.

adtac:
No problem, thanks! :)

PR #222, opened 2017-12-14
Set minimum gas limit to 5000 (duplicate for test re-run)

carver:
Re-attempting #162 to get a green check.

### What was wrong?

According to these sources, the yellow paper has a bug:

- https://github.com/ethereum/yellowpaper/issues/317#issuecomment-345458531

> This seems to have been accidentally changed on 16 Feb 2016 in the commit ethereum/yellowpaper@a3381de
The minGasLimit was 5000 at Frontier and hasn't changed in any of the hardforks.

- https://github.com/ethereum/tests/issues/383#issuecomment-351709445

> I think the yellow paper is wrong just because all others say 5000.

### How was it fixed?

Just repushed the change from #162 to reattempt passing tests.

... and then bugfixed some block gas limit issues.

#### Cute Animal Picture


carver:
I'm starting to pull on this thread, and more yarn is coming out than expected. It looks like some issues were being masked by the min gas limit at 125k. So far I've identified:

- [x] gas adjustment algorithm was missing a case (limit is below floor, but above protocol minimum)
- [ ] VM accepts new blocks with a much lower gas limit than previous block header. Test expects a validation error.
- [ ] more?

I don't see any open PRs looking at these issues, so I'll keep hacking on this till it's green. (It's causing RPC failures, so is blocking work on #183 ).

pipermerriam:
Ping me if you get overly confused or lost.

PR #223, opened 2017-12-14
New Transaction Format

jannikluhn:
This adds the sharding transaction class. Initially I tried to copy and adapt `vm/forks/frontier/*` as well, but at some point I realized that this is useless as everything would have to be rewritten as part of the account abstraction PR anyway. So this PR remains fairly basic.

One test is x-failing because of a bug in pyRLP, I'm gonna fix that over there.


jannikluhn:
Wait, that's not a bug in pyRLP, I'm just confused by too many brackets. Will fix tomorrow.

evm/rlp/sedes.py
```diff
@@ -1,11 +1,52 @@
...
+)
+
+
+class AccessListElement(CountableList):
+
+    def __init__(self):
+        super().__init__(Binary(max_length=32))
+
+    def serialize(self, obj):
+        result = super().serialize(obj)
```
pipermerriam: Can we move this to below the `if/else` so that it only happens *if* the validation checks pass.

jannikluhn: I'd prefer to leave it there, as `super().serialize(obj)` does a lot of checks already. In particular, that `obj` is actually a list and `obj[0]` is binary. If that weren't the case, the error messages would be misleading.

pipermerriam: That's fine.  I wasn't aware of the  checks that occur in the super call.

pipermerriam:
:shipit:

PR #229, opened 2017-12-19
lightchain: Implement get_account() method

gsalgado:


evm/p2p/lightchain.py
```diff
@@ -459,6 +471,17 @@ def peer_block_height(peer: LESPeer):
         return reply['codes'][0]
 
 
+def verify_trie_proof(root_hash: bytes, key: bytes, proof: List[bytes]) -> bytes:
```
pipermerriam: Just to mention it here.

This functionality should be available off of the `Trie` class.  Something like `Trie.verify_proof` and `Trie.construct_proof`.

gsalgado: Right, but the proof verification makes more sense as a standalone function since one does not need a Trie to use it, and nor does it return one: https://github.com/ethereum/py-trie/pull/15

Issue #194, opened 2017-12-05
STUB: New transaction format


pipermerriam:
See [sharding doc](https://github.com/ethereum/sharding/blob/develop/docs/doc.md#transaction-format)

- [x] Update `Transaction` format (depend on #203 and #192)
    The format of a transaction now becomes:
    ```python
    [
        chain_id,      # 1 on mainnet
        shard_id,      # the shard the transaction goes onto
        target,        # account the tx goes to
        data,          # transaction data
        start_gas,     # starting gas
        gasprice,      # gasprice
        access_list,   # access list (see below for specification)
        code           # initcode of the target (for account creation)
    ]
    ```
- [ ] Update `apply_transaction` function procedure

jannikluhn:
I'm gonna start working on this, but disregard any account abstraction stuff for the moment (as there's no clear specification yet). So the transaction object will include nonce and signature. Removing them should be trivial later.

jannikluhn:
Closing this with #223 (`apply_transaction` hasn't been updated yet, but this belongs more to #203 in my opinion)

hwwhww:
@jannikluhn  I will add a new ticket for `apply_transaction` with new transaction format.

jannikluhn:
@hwwhww This seems to be included in #238 already. I think it will need some updates concerning gas calculation / payment and maybe state access restrictions, but the basics seem to be there. Can you comment, @NIC619 ?

NIC619:
I updated the `execute_transaction` with the new transaction format in #238.
Regarding gas calculation, intrinsic gas cost is updated to also includes size of `tx.code`. Should we include the cost of reading access list?
Regarding payment, gas cost is updated to be paid by `tx.to`.
Regarding state access restriction, it seems to met that it's already be done in #202 ?

@jannikluhn @hwwhww  Please let me know if there's any update that's missing regarding the new transaction format.

jannikluhn:
@NIC619 I think the only thing wrong with the transaction format itself is that `gas_price` is still part of it (at least given https://ethresear.ch/t/tradeoffs-in-account-abstraction-proposals/263/23). Regarding transaction execution I can see three things which need to be changed long term, but are probably fine for now:

- gas subtraction at the protocol level (this should be solved in #234 I guess)
- there should be a way to stop execution after N gas has been consumed unless PAYGAS has been called (depending on the chosen account abstraction proposal)
[...]

PR #230, opened 2017-12-19
Standard user account

jannikluhn:
The standard user account is a contract parametrized by an address that
- Validates that the transaction is signed with the private key corresponding to the address,
- validates that the block number is in a certain range (given by the transaction),
- stores a nonce that is incremented at every call,
- pays for gas, and
- calls another contract.

It also has additional methods to
- extract the nonce and to
- validate an arbitrary signed message against the owner address.

Once the PR is finished it should contain
- the contract code both as Vyper LLL and compiled bytecode
- extensive tests for the contract code
- utility functions to construct transactions in a format understood by the contract
- utility functions to deploy such a contract (which might be split into a main and a validation contract)

References:
- https://ethresear.ch/t/tradeoffs-in-account-abstraction-proposals/263
- https://ethresear.ch/t/validation-code-for-abstraction/335/9


evm/auxiliary/user_account.py
```diff
@@ -0,0 +1,181 @@
...
+        self.nonce = nonce
+        self.gas_price = gas_price
+        self.msg_data = msg_data
+
+    #
+    # Properties for fields encoded in data
+    #
+
+    @property
+    def signature(self):
```
pipermerriam: what do you think about renaming this to `vrs` instead of `signature`?  I think that naming convention would match better with other portions of both this codebase and others (web3.py and eth-keys).

With the conventions we've been establishing, I would expect this to return a `eth_keys.Signature` object.

evm/auxiliary/user_account.py
```diff
@@ -0,0 +1,181 @@
...
+        return big_endian_to_int(self.data[VALUE_SLICE])
+
+    @value.setter
+    def value(self, value_):
+        validate_uint256(value_)
+        b = pad32(int_to_big_endian(value_))
+        self.data = self.data[:VALUE_SLICE.start] + b + self.data[VALUE_SLICE.stop:]
+
+    @property
+    def receiver(self):
```
pipermerriam: This fragments the terminology that we use for the *destination* of a transaction.  Elsewhere we're using `to`.  Any thoughts on whether it'd be better to conform to this naming convention?

evm/logic/context.py
```diff
@@ -176,3 +176,7 @@ def returndatacopy(computation):
     value = computation.return_data[returndata_start_position: returndata_start_position + size]
 
     computation.memory.write(mem_start_position, size, value)
+
+
+def sighash(computation):
+    pass
```
pipermerriam: thoughts on changing this to a `raise NotImplementedError`?

evm/logic/system.py
```diff
@@ -173,3 +173,7 @@ def __call__(self, computation):
         if computation.msg.is_static:
             raise WriteProtection("Cannot modify state while inside of a STATICCALL context")
         return super(CreateEIP150, self).__call__(computation)
+
+
+def paygas(computation):
+    pass
```
pipermerriam: thoughts on changing this to a `raise NotImplementedError`?

evm/vm/forks/sharding/opcodes.py
```diff
@@ -0,0 +1,33 @@
...
+from evm.vm.forks.byzantium import BYZANTIUM_OPCODES
+
+
+NEW_OPCODES = {
+    opcode_values.PAYGAS: as_opcode(
+        logic_fn=system.paygas,
+        mnemonic=mnemonics.PAYGAS,
+        gas_cost=0,
+    ),
+
```
pipermerriam: whitespace

tests/auxiliary/user-account/test_user_account.py
```diff
@@ -0,0 +1,172 @@
...
+    assert not unsigned_transaction.is_signed
+    private_key = keys.PrivateKey(b'\x22' * 32)
+    unsigned_transaction.sign(private_key)
+    assert unsigned_transaction.is_signed
+    assert unsigned_transaction.get_sender() == private_key.public_key.to_canonical_address()
+
+
+def test_valid_transfer(chain, tx1):
+    vm = chain.get_vm()
+    vm.apply_transaction(tx1)
```
pipermerriam: Are there any assertions that we could make here?

evm/auxiliary/user_account.py
```diff
@@ -0,0 +1,181 @@
...
+    big_endian_to_int,
+    int_to_big_endian,
+)
+from evm.utils.padding import (
+    pad32,
+)
+
+
+USER_ACCOUNT_CODE = decode_hex(
+    "0x3f595260805903608020600052602060006080600060006001610bb8f1506001600051"
```
pipermerriam: Will you add a comment referencing the *source code* that produced this bytecode and how it can be reproduced/verified.

evm/auxiliary/user_account.py
jannikluhn: Not sure... We used to have `to` and `sender`. Now we have `target`, `sender` and `receiver`.

- `to` and `target` refer to more or less the same thing I guess (the first account that is called)
- `sender` changes meaning: it used to be the one who signed the transaction, now it's the "owner" of the `target` account (and just happens to put a signature in the `data` field)
- `receiver` is the second hop (the account that's called by the `target` account)

So from a technical standpoint, I don't see a lot of overlap between original and new terms. I could imagine renaming `target` to `to` for consistency (although I'd prefer `target` in a vacuum). Renaming `receiver` to `to` could be misleading on the protocol level, but would make sense from a high level point of view in most cases. I don't see a good alternative to `sender`.

tests/auxiliary/user-account/test_user_account.py
jannikluhn: Definitely. That's one of the reasons why it's still WiP (this test doesn't even pass yet)

evm/auxiliary/user_account.py
jannikluhn: Do you prefer renaming it to `vrs` or making it return a `Signature` object? I have no preference in that regard.

evm/auxiliary/user_account.py
jannikluhn: Yes, this is not much more than a placeholder for now. Ideally I'd like to put the source code there directly and compile it in a test for comparison, but this will probably not work because Viper needs Python 3.6.

hwwhww: #206 has the same issue, it's better not import Viper even in Python 3.6. 

A (little complicated) suggestion of verifying Viper code:
1. A command for compiling Viper source code to bytecode.
2. If someone want to update the code, they have to upload `source code` and `bytecode` (can be in separate file or assigned to a global variable in this case)
3. Install Viper in CI testing.
4. Write a test for auto-compiling the source code and then comparing the compiled bytecode with `evm.auxiliary.user_account.USER_ACCOUNT_CODE`.

By the way, @pipermerriam what do you think about upgrading to Python 3.6?

jannikluhn: Perhaps the cleanest solution would be to put the code plus tests in a separate repository. The repository would contain packages with the bytecode for different languages that can be imported into client implementations such as Py-EVM.

evm/auxiliary/user_account.py
pipermerriam: Good explanation.  I think you are right to leave this as-is and not change the names from `target/sender/receiver`

pipermerriam: Does the name `destination` better capture the concept that `receiver` currently represents?

evm/auxiliary/user_account.py
pipermerriam: I think `vrs` is my preference since this is a low-level API

evm/auxiliary/user_account.py
pipermerriam: > By the way, @pipermerriam what do you think about upgrading to Python 3.6?

I'm in favor: https://github.com/ethereum/py-trie/pull/18

pipermerriam: I'm indifferent to the actual solution for the bytecode creation as long as it's documented and independently verifiable.

jannikluhn:
I've given up on testing the contract code for now, as it's not really possible at the moment. I'll come back to it once #238 is done, but for now I'd like to put this PR on hold.

pipermerriam:
I think that a large part of this diff will go away once this is rebased.

jannikluhn:
Updated the description. Sorry for the current state of the PR. Now that #238 is merged I'm back at writing tests/debugging the contract. In the meantime I've made a couple of changes in the contract. I'm going to push something later today.

jannikluhn:
Still not completely finished, but almost there I hope.

To run the tests, cherry-pick https://github.com/ethereum/py-evm/pull/259/commits/90d155d23ef95d18c24170e168414b634e9b7481, resolve the merge conflicts (accept both), and replace [this line](https://github.com/jannikluhn/py-evm/blob/6e76c1277a82fa6b2241501ecc5d71630fec205d/evm/logic/system.py#L196) with `pass`.

pipermerriam:
Sighash is now merged so this can be rebased (as I assume it needs sighash in it to work)

pipermerriam:
eh, should have said sighash is *ready for merge* :P

jannikluhn:
Merged the sighash PR and rebased. This should be ready for review now.

I'm not particularly happy with some names in this one, especially `ForwardingTransaction`. I called it `forwarding` because the "payload" data is passed through to the actual "destination" contract. But it's specific to one particular user contract, others might require a different format.

Test-wise I'd like to see tests that make sure that everything works as expected even if the gas runs out at different points during the execution of the contract. But this can only be done once PAYGAS is properly implemented.

jannikluhn:
Ooh, I missed one thing: `access_list` is left empty all the time. It should default to the whole storage space of the destination account and to the required parts of user and validation accounts (as suggested [here](https://github.com/ethereum/py-evm/issues/194#issuecomment-356503391). Tests are not failing because of #263 .

evm/auxiliary/user_account_contract/transaction.py
```diff
@@ -0,0 +1,210 @@
...
+VALUE_SLICE = slice32(CALLDATA_VALUE)
+MIN_BLOCK_SLICE = slice32(CALLDATA_MIN_BLOCK)
+MAX_BLOCK_SLICE = slice32(CALLDATA_MAX_BLOCK)
+DESTINATION_SLICE = slice32(CALLDATA_DESTINATION)
+DATA_SLICE = slice(CALLDATA_DATA, None)  # open ended
+
+EMPTY_DATA = b'\x00' * DESTINATION_SLICE.stop
+
+
+class ForwardingTransaction(ShardingTransaction):
```
pipermerriam: Other name ideas

- `StandardAccountTransaction`
- `ProxyTransaction`
- `EIP86Transaction` (or whatever the appropriate EIP number is)

evm/auxiliary/user_account_contract/transaction.py
```diff
@@ -0,0 +1,210 @@
...
+            access_list=access_list,
+            code=b'',
+        )
+        self.destination = destination
+        self.value = value
+        self.min_block = min_block
+        self.max_block = max_block
+        self.nonce = nonce
+        self.int_gas_price = gas_price
+        self.msg_data = msg_data
```
pipermerriam: It looks like this class needs to implement a `validate()` method to validate these additional parameters.

evm/auxiliary/user_account_contract/transaction.py
```diff
@@ -0,0 +1,210 @@
...
+    def is_signed(self):
+        return self.vrs != (0, 0, 0)
+
+    def get_message_for_signing(self):
+        return b''.join([
+            self.data[SIGNATURE_SLICE.stop:],
+            self.sig_hash,
+        ])
+
+    def sign(self, private_key):
```
pipermerriam: Was it intentional to deviate from the existing API for transaction signing which uses `as_signed_transaction` which returns a new object rather than mutating the existing object?

This question is from a broader question about the use of property setters.

- **if** the use of property setters was to embed validation as part of setting the value, then I would suggest changing to use a single `validate()` method which is inline with the other transaction classes.
- **if** the use of property setters was to allow mutation of the object, I would like to understand why since we've trended away from most forms of object mutation.

tests/auxiliary/user-account/test_contract.py
```diff
@@ -0,0 +1,372 @@
...
+    "chain_id": 1,
+    "shard_id": 1,
+    "to": ACCOUNT_ADDRESS,
+    "gas": 500000,
+    "gas_price": 0,
+    "access_list": [],
+    "code": b"",
+}
+
+DEFAULT_TX_PARAMS = {**DEFAULT_BASE_TX_PARAMS, **{
```
pipermerriam: This method of dictionary merging has always bugged me, mostly because it takes advantage of the implementation rather than using formal semantics or which keys take precedence.  I admit it could be argued that the way key precedence is handled in this case is part of the formal semantics for how this works, but either way, it's always felt a bit like magic.

What do you think about making use of something like the following.  (If you haven't figured it out by now, I have great love for the `toolz/cytoolz` library)

```python
from cytoolz import (
    merge,
    dissoc,
)

DEFAUT_TX_PARAMS = merge(
    dissoc(DEFAULT_BASE_TX_PARAMS, 'code'),
    {...<the overrides>},
}
```

Zero mutation and explicit about dictionary merging.

tests/auxiliary/user-account/test_contract.py
```diff
@@ -0,0 +1,372 @@
...
+
+        (v, r + 1, s),
+        (v, r - 1, s),
+        (v, 0, s),
+
+        (v, r, s + 1),
+        (v, r, s - 1),
+        (v, r, 0),
+
+        (27 if v == 28 else 28, r, SECPK1_N - s)
```
pipermerriam: can we parametrize these using `pytest.mark.parametrize` so that they each run in isolation?

tests/auxiliary/user-account/test_contract.py
```diff
@@ -0,0 +1,372 @@
...
+        "nonce": get_nonce(vm),
+        "destination": GAS_LOGGING_CONTRACT_ADDRESS,
+        "gas": 1 * 1000 * 1000,
+    }}).sign(PRIVATE_KEY)
+    computation = vm.apply_transaction(transaction)
+    assert computation.is_success
+
+    logs = computation.get_log_entries()
+    assert len(logs) == 1
+    assert big_endian_to_int(logs[0][-1]) > 900 * 1000  # some gas will have been consumed earlier
```
pipermerriam: A slightly easier to understand version of this assertion that I've used elsewhere is to assert that the delta is smaller than a certain value.

```
gas_used = big_endian_to_int(logs[0][-1])
assert transaction.gas - gas_used < 1000
```

tests/auxiliary/user-account/test_contract.py
```diff
@@ -0,0 +1,372 @@
...
+
+@pytest.fixture
+def vm():
+    header = BlockHeader(
+        coinbase=to_canonical_address("8888f1f195afa192cfee860698584c030f4c9db1"),
+        difficulty=131072,
+        block_number=10,
+        gas_limit=3141592,
+        timestamp=1422494849,
+        parent_hash=decode_hex("0000000000000000000000000000000000000000000000000000000000000000"),
```
hwwhww: You can use `evm.constants.ZERO_HASH32`.

evm/auxiliary/user_account_contract/transaction.py
jannikluhn: I added the setters to allow mutation. The other transaction objects are technically mutable as well and I wanted to mirror that. But if they aren't mutated in practice I'll be happy to remove them because all these setters look quite ugly. I'll also try to implement the `as_signed_transaction` pattern.

tests/auxiliary/user-account/test_contract.py
jannikluhn: Personally I don't prefer one way over the other, so I don't mind adopting this style.

evm/auxiliary/user_account_contract/transaction.py
jannikluhn: `StandardAccountTransaction` was the name before I switched to `ForwardingTransaction`. I dismissed it because I'm not sure if "standard account" is a good term either. It should probably be rather called "standard user account", and even that assumes that this contract will actually become "the" standard. But `StandardUserAccountTransaction` strikes me as too long.

General note on EIPs: As far as I know there are no EIPs for the sharding protocol itself. At least at the moment, I don't know if that will change.

evm/auxiliary/user_account_contract/transaction.py
```diff
@@ -71,18 +61,98 @@ def __init__(
...
             to=to,
-            data=EMPTY_DATA,
+            data=data,
             gas=gas,
             gas_price=gas_price,
             access_list=access_list,
             code=b'',
         )
+
+    def create_unsigned_transaction(self):
```
pipermerriam: To stick with the existing transaction API this should be a `classmethod`.  This API exists to reduce the cumbersome import of both the signed and unsigned classes, allowing you to create unsigned transactions from the main transaction class.  I'd be fine with this just being an `*args, **kwargs` passthrough to reduce size since there are 10+ arguments.

tests/auxiliary/user-account/test_contract.py
```diff
@@ -282,91 +286,107 @@ def test_call_checks_signature(vm):
...
+            min_block=transaction.min_block,
+            max_block=transaction.max_block,
+            gas_price=transaction.gas_price,
+            msg_data=transaction.msg_data,
+            v=v,
+            r=r,
+            s=s,
+        )
+        # don't detect invalidity before transaction execution
+        transaction.validate = lambda: None
```
pipermerriam: I think a more appropriate way to check this would be to assemble a `Message` and then to run `vm.get_computation(message).apply_message()`.

This should execute the transaction *below* the place where the call to `transaction.validate()` occurs, allowing you to directly test that the EVM code appropriately validates the signature without having to do any monkeypatching of the transaction object.

hwwhww:
I was checking with @vbuterin  about renaming some parameters and also mentioned about 
`ShardingTransaction.to` or `ShardingTransaction.target`, he clarified his intention:
> so the reason I wanted to rename it to target is that in real life, the transaction.to field will be the sender of the money (and/or function call)
as it would be the forwarding contract that controls the account from which the funds get sent and the receiver would be specified in the tx.data under some encoding
I think it would cause too much long-term risk of devs getting confused for `transaction.to` to be the sender
hence my idea to rename it `target`

jannikluhn:
Addressed comments and rebased.

@vbuterin Do you have time to review the contract code? https://github.com/ethereum/py-evm/pull/230/files#diff-05178e170e27782a10a1498ec2aefdef

evm/auxiliary/user_account_contract/contract.py
```diff
@@ -0,0 +1,211 @@
...
+
+def generate_account_lll_code(validation_code_address):
+    return [
+        'seq',
+
+        # handle get_nonce if called
+        ['if',
+            ['eq',
+                ['calldataload', CALLDATA_FUNCTION_ID],
+                NONCE_GETTER_ID],
```
vbuterin: Shouldn't NONCE_GETTER_ID be multiplied by 2^224? It's the first four bytes that are the function sig, so getting the first 32 bytes would return that with 28 zero bytes at the end.

BTW are there tests for all of this? If so, the tests should catch this.

evm/auxiliary/user_account_contract/contract.py
```diff
@@ -0,0 +1,211 @@
...
+        # copy full call data to memory
+        ['calldatacopy', MEMORY_CALLDATA, 0, ['calldatasize']],
+
+        # handle get_validation_code if called
+        ['if',
+            ['eq',
+                ['calldataload', CALLDATA_FUNCTION_ID],
+                VALIDATION_CODE_GETTER_ID,
+                # return validation code address
+                ['mstore', 0, validation_code_address],
```
vbuterin: Now that I think about it, what *would* the validation code address be? We don't want to create two contracts per account. I would actually recommend just not bothering with this `get_validation_code` feature at this point to keep it simpler.

evm/auxiliary/user_account_contract/contract.py
jannikluhn: You're right, will fix this.

Tests are here: https://github.com/jannikluhn/py-evm/blob/07587ac50734c62a7d8862c072fbcaeedaa93c97/tests/auxiliary/user-account/test_contract.py

They cheat a bit by importing the function ids, that's why they didn't catch this. Will add a test comparing against a manually generated one (using the Vyper code).

evm/auxiliary/user_account_contract/contract.py
```diff
@@ -0,0 +1,211 @@
...
+                VALIDATION_CODE_GAS,
+                validation_code_address,
+                0,
+                # input data (spans sighash and signature)
+                MEMORY_SIGHASH,
+                128,
+                # output data (discarded)
+                0,
+                0
+            ]],
```
vbuterin: Let's not do things this way. This would require calling 2 accounts to send from this account, which would blow up witness size by a factor of 2. IMO just do ecrecover here.

evm/auxiliary/user_account_contract/contract.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # Verify and increment nonce
+        ['assert', ['eq', ['calldataload', CALLDATA_NONCE], ['sload', STORAGE_NONCE]]],
+        ['sstore', STORAGE_NONCE, ['add', ['calldataload', CALLDATA_NONCE], 1]],
+
+        # Assert that we won't run out of gas from here on
+        ['assert', ['gt', 'gas', GAS_RESERVE + GAS_RESERVE_OFFSET]],
```
vbuterin: Is this line still necessary?

jannikluhn: Depends: If we run out of gas after the PAYGAS call, the transaction can be replayed infinitely and the account would get drained. This line makes sure that this can never happen. If we trust the user to never create transactions with not enough gas, the check isn't necessary. But at least I would feel much safer with it.

jannikluhn:
Anything else or can we merge?

pipermerriam:
I can get onboard with the rename from `to` to `target` with that explanation.

jannikluhn:
I'm also in favor of renaming `to` to `target`, but I think this should be done in a separate pull request.

jannikluhn:
Rebased. Is this `Ready-for-merge` when tests have passed?

evm/auxiliary/user_account_contract/contract.py
```diff
@@ -0,0 +1,157 @@
...
+    big_endian_to_int,
+)
+
+
+# S <= N // 2
+MAX_ALLOWED_S = SECPK1_N // 2
+VALIDATION_CODE_GAS = 3500
+GAS_RESERVE = 4500  # amount of gas reserved for returning
+GAS_RESERVE_OFFSET = 200
+NONCE_GETTER_ID = 0x141b5b48
```
hwwhww: maybe add a comment to describe
`# NONCE_GETTER_ID = big_endian_to_int(keccak(b"get_nonce()")[:4])` ?

hwwhww:
They are green now  🎉

Issue #231, opened 2017-12-19
missing: LICENSE file


themusicgod1:
* py-evm Version: v0.2.0-alpha.7
* OS: n/a
* Python Version (python --version): n/a
* Environment (output of `pip freeze`): n/a

### What is wrong?

While setup.py claims that the project is under
license='MIT' / 'License :: OSI Approved :: MIT License', the project does not appear to specify, in a human-readible(well, outside-of-code at least way) license.

### How can it be fixed

Including said OSI approved MIT license instead of just referring to it in the code itself.

(If there's one thing that will get py-evm into Debian quicker, it's having all the ducks lined up in a row on license issues)

themusicgod1:
lgtm

PR #232, opened 2017-12-19
included license file

themusicgod1:
### What was wrong?

No LICENSE file. See #231 

### How was it fixed?

Added a MIT license under the name specified in setup.py

#### Cute Animal Picture

do cartoons count?

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,21 @@
+The MIT License (MIT)
+
 Copyright 2017 Piper Merriam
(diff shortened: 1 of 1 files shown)


pipermerriam:
FYI, I updated with a copy/paste from one of my repos since I didn't want to compare word/for/word accuracy.  Thanks for identifying this issue and fixing.  will merge here as soon as tests are green.

PR #235, opened 2017-12-20
Pure apply_transaction

hwwhww:
#### Description

#195 subtask
Pure apply_transaction function

1. Moved the logic of `Block.make_receipt(transaction, computation)` to `VM.make_receipt(transaction, computation)` 
2. I found that it’s easier to using `VM` as a container for state transition than separating input `block.chaindb`, `vm.block.chaindb`, and `vm.chaindb`. So I only moved the logic of  `Block.add_transaction(Block)` to `VM.add_transaction(transaction, computation)`.
3. Added `VM.apply_transaction_to_block(transaction, block, chaindb)` method. This method has similar function as `VM.apply_transaction(tx)`, but instead of using the member variable, it `deepcopy` and apply the given `block` and `chaindb`.
4. Added `vm.state_transition_helper.apply_transaction` function. It’s a pure function that won’t mutate the given input and returns `success, reads, writes, block`, where:
	* `success`: check if the state transition succeeded
	* `reads`: the reads dict during state transition
	* `writes`: the writes dict during state transition
	* `block`: the applied block

#### TODO
- Add more tests to verify  `reads` and `writes` dicts.
- Our original goal is `apply_transaction(stateobj, db, blockdata, tx) -> state_obj', reads, writes`, so I'm still looking for using simpler StateObj object instead of VM as the state transition container.
	1. Should we remove `self.block` from `VM`?
	2. Should we remove `self.chaindb` from `Block`?
	3. Risk of using too many `deepcopy` (Looking for other method) 

#### Cute Animal Picture

cute-animal-red-panda-2

### DIFF
--- a/evm/db/backends/memory.py
+++ b/evm/db/backends/memory.py
@@ -6,8 +6,8 @@ from .base import (
 class MemoryDB(BaseDB):
     kv_store = None
 
-    def __init__(self):
-        self.kv_store = {}
+    def __init__(self, kv_store=None):
+        self.kv_store = {} if kv_store is None else kv_store
 
     def get(self, key):
         return self.kv_store[key]
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -23,6 +23,9 @@ from evm.exceptions import (
 from evm.db.journal import (
     JournalDB,
 )
+from evm.db.tracked import (
+    TrackedDB
+)
 from evm.db.state import (
     State,
     NestedTrieBackend,
@@ -244,9 +247,10 @@ class BaseChainDB:
     def clear(self):
         self.db.clear()
 
-    def get_state_db(self, state_root, read_only, access_list=None):
+    def get_state_db(self, state_root, read_only, access_list=None, is_stateless=False):
+        db = TrackedDB(self.db) if is_stateless else self.db
         return State(
-            db=self.db,
+            db=db,
             root_hash=state_root,
             read_only=read_only,
             access_list=access_list,
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -90,15 +90,12 @@ class VM(object):
     #
     # Execution
     #
-
     def add_transaction_to_block(self, transaction, block, computation):
         """
+        Add a transaction to the given block.
         """
-        with self.state_db(read_only=True) as state_db:
-            self.logger.info(
-                "1.1 block.coinbase: %s",
-                state_db.get_balance(b'\xa9OSt\xfc\xe5\xed\xbc\x8e*\x86\x97\xc1S1g~n\xbf\x0b'),
-            )
+        # [FIXME]: initialize self.block?
+        # [FIXME]: don't mutate the given block
 
         receipt = block.make_receipt(transaction, computation)
 
@@ -108,8 +105,8 @@ class VM(object):
 
         self.block.transactions.append(transaction)
 
-        tx_root_hash = self.chaindb.add_transaction(block.header, index_key, transaction)
-        receipt_root_hash = self.chaindb.add_receipt(block.header, index_key, receipt)
+        tx_root_hash = self.block.chaindb.add_transaction(block.header, index_key, transaction)
+        receipt_root_hash = self.block.chaindb.add_receipt(block.header, index_key, receipt)
 
         self.block.bloom_filter |= receipt.bloom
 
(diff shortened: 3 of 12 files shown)


evm/db/backends/memory.py
```diff
@@ -6,8 +6,8 @@
 class MemoryDB(BaseDB):
     kv_store = None
 
-    def __init__(self):
-        self.kv_store = {}
+    def __init__(self, kv_store=None):
+        self.kv_store = {} if kv_store is None else kv_store
```
pipermerriam: *nitpick*

Inline turnary statements aren't great readability.  Do you mind expanding this to a fill `if/else` across 4 lines.

evm/db/chain.py
```diff
@@ -244,9 +247,10 @@ def commit(self, checkpoint):
     def clear(self):
         self.db.clear()
 
-    def get_state_db(self, state_root, read_only, access_list=None):
+    def get_state_db(self, state_root, read_only, access_list=None, is_stateless=False):
+        db = TrackedDB(self.db) if is_stateless else self.db
```
pipermerriam: Same thought here about avoiding inline turnary statements in favor of readability.

evm/vm/base.py
```diff
@@ -57,9 +72,17 @@ def configure(cls,
                 )
         return type(name, (cls,), overrides)
 
+    def set_stateless(self, is_stateless):
```
pipermerriam: Hrm, is it possible for us to avoid needing this setter and to require statelessness be declared on instantiation?

evm/vm/base.py
```diff
@@ -90,13 +121,62 @@ def logger(self):
...
+        )
+
+        return computation
+
+    def apply_transaction_to_block(self, transaction, block, chaindb):
+        """
+        Apply the transaction to the vm in the current block.
+        """
+        # (semi) pure function setting
+        self.chaindb = copy.deepcopy(chaindb)
```
pipermerriam: I suspect the use of `deepcopy` here is not going to be a workable solution since we almost certainly cannot write our `chaindb` class in a way that allows for arbitrary copying of the database.  Solutions that come to mind are:

- Something like the `OverlayDB` class.
- An API on the `BaseChainDB` class such as `chaindb.get_overlay_copy()` which wrapps the underlying db in an `OverlayDB`.

evm/vm/state_transition_helper.py
```diff
@@ -0,0 +1,47 @@
...
+    except Exception as e:
+        # FIXME: any specific exception type in this function?
+        logger.error(
+            "Unexpected error when handling remote msg: {}".format(traceback.format_exc()))
+        success = False
+
+    reads = vm.reads
+    writes = vm.writes
+
+    return success, reads, writes, vm.block
```
pipermerriam: Instead of success/failure can we return the computation object itself?  Reasoning is that it should still be a fine representation of success/failure but it also allows for deeper inspection of what happened during computation which I think is valuable and was one of the things missing in pyethereum that I really wanted.  Things like accessing the `output` of computation or really anything else that's exposed through the computation object.

evm/vm/state_transition_helper.py
```diff
@@ -0,0 +1,47 @@
+import copy
+import logging
+import traceback
+
+
+logger = logging.getLogger("evm.research.state_transition_helper")
+
+
+def apply_transaction(transaction, block, prev_vm, chaindb):
```
vbuterin: Are we sure that this is actually a pure function? It's clear that prev_vm is not being modified; but could chaindb be modified anywhere? Likely not but I'm not well-versed enough in the code to be able to tell.

hwwhww: Currently, I only checked `state_root` hasn't been changed (`chain.get_vm()` uses `chain.chaindb` to initialize `VM`):
https://github.com/ethereum/py-evm/pull/235/files#diff-555a9b3a5d79c67dccb0a54b0f1649ddR51
I will add more tests after #236 .

evm/vm/state_transition_helper.py
hwwhww: Yes, I will change to returning `computation`.

hwwhww:
The pure `apply_trasaction` function would be established on #236 and other refactorings on `master` branch. This PR would be split into other PRs.

Issue #234, opened 2017-12-20
Add PAYGAS opcode


NIC619:
### What is wrong?

Add `PAYGAS` opcode to abstract away `gasprice`

### How can it be fixed

PAYGAS opcode takes one input `gasprice` off the stack and deduct `gasprice*tx.gas` from the contract's balance

NIC619:
Closed via #285

Issue #224, opened 2017-12-15
Implement CREATE2 opcode


hwwhww:
Subtask of #203 
see https://github.com/ethereum/EIPs/issues/86
https://github.com/ethereum/EIPs/blob/bd136e662fca4154787b44cded8d2a29b993be66/EIPS/abstraction.md

PR #238, opened 2017-12-21
Sharding[VM,Block,Transaction,Message] and sharding transaction/message execution logic

NIC619:
### What was wrong?

With new transaction format and to introduce ENTRY_POINT address(b'ffff….ffff'), some logics in `vm.execute_transaction` needed to be changed.
Then CREATE2(#224), PAYGAS(#234) and PANIC opcodes can be built on top.

### How was it fixed?

- [x] `ShardingMessage`, `ShardingTransaction`, `ShardingVM`
    - [x]  `ShardingTransaction.intrinsic_gas`
    - [x] `validate_sharding_transaction`
- [x]  Account Abstraction
    - [x] `ENTRY_POINT`
    - [x] `generate_create2_contract_address`

#### Cute Animal Picture


evm/vm/forks/sharding/validation.py
```diff
@@ -0,0 +1,14 @@
...
+from evm.vm.forks.frontier.validation import (
+    validate_frontier_transaction,
+)
+
+
+def validate_sharding_transaction(evm, transaction):
+    # TODO:Update transaction validation logic in Sharding
+    # e.g. checking shard_id < SHARD_COUNT
+
+    validate_frontier_transaction(evm, transaction)
```
hwwhww: Add a comment for using `validate_frontier_transaction` but `validate_homestead_transaction` on purpose would lesser ambiguity. (`validate_homestead_transaction` doesn't apply for sharding)

evm/vm/forks/sharding/validation.py
```diff
@@ -0,0 +1,14 @@
+from evm.exceptions import (
+    ValidationError,
```
hwwhww: Never been used.

evm/rlp/transactions.py
```diff
@@ -151,7 +151,7 @@ class BaseShardingTransaction(rlp.Serializable):
         ('shard_id', big_endian_int),
         ('target', address),
         ('data', binary),
-        ('start_gas', big_endian_int),
+        ('gas', big_endian_int),
```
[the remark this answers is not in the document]
NIC619: `start_gas` is only used in `GasMeter` while other Transaction classes use `gas`. So I change it for uniformity, but I have no strong preference for either one.

evm/vm/forks/sharding/blocks.py
```diff
@@ -0,0 +1,21 @@
...
+)
+from evm.vm.forks.homestead.blocks import (
+    HomesteadBlock,
+)
+from .transactions import (
+    ShardingTransaction,
+)
+
+
+class ShardingBlock(HomesteadBlock):
```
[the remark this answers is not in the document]
NIC619: yeah, the name will be changed later but for now I keep it this way for simplicity.

I suppose with new transaction class, there would have to be new block class?
the new block class will be part of the `ShardingVM`
```
ShardingVM = ByzantiumVM.configure(
    name='ShardingVM',
    _block_class=ShardingBlock,
)
```

evm/rlp/transactions.py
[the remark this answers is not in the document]
pipermerriam: I've been using the term `gas` in the other `Transaction` classes to be more inline with the conventions used elsewhere such as web3.py or the JSON-RPC spec.  I *could* be convinced to change it if someone has a compelling argument.

evm/vm/forks/sharding/blocks.py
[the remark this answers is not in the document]
pipermerriam: I think it was discussed but I can't recall the specifics.  

What do you think about just writing an entirely separate `Shard` class so that we aren't overloading the `block` based API's, trying to cram the concepts of collations and blocks into the same API.  While the two have a lot of similarities, I think they are going to be sufficiently different that we'll need a separate abstraction.

I have some deeper thoughts on the subject that I'll try to write-up since I'm starting to get the feeling that we will need a `Shard` class which mirrors the `Chain` class and than we'll want a `ShardVM` which mirrors the `VM` class.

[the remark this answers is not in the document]
NIC619: Agree that it's better to write a `Shard` class. But maybe it can be replaced later? My goal for this PR is to change Transaction related code. (sorry about the misleading `ShardingBlock`, I figured if I want to test the modification on Transaction I need to also set up `VM` and `Block` class for `Sharding`.)

evm/rlp/transactions.py
```diff
@@ -160,3 +160,32 @@ class BaseShardingTransaction(rlp.Serializable):
     @property
     def hash(self):
         return keccak(rlp.encode(self))
+    
+    @property
+    def intrensic_gas(self):
```
pipermerriam: *organization nitpick*

This seems like it should be next to the `get_intrensic_gas` method.

pipermerriam:
Please get tests passing (looks like a linting error?).

NIC619:
@pipermerriam @hwwhww I update the code along with PR detail. Please have a look, thanks.

evm/vm/message.py
```diff
@@ -142,3 +143,73 @@ def storage_address(self, value):
...
+        self.gas_price = gas_price
+
+        validate_canonical_address(to, title="Message.to")
+        self.to = to
+
+        if sender != ENTRY_POINT:
+            validate_canonical_address(sender, title="Message.sender")
+        self.sender = sender
+
+        validate_uint256(value, title="Message.value")
```
pipermerriam: I think you can eliminate the majority of the body of this function by inserting a `super` call to the `Messag.__init__` method and then only dealing with what looks like the *only* difference which is `access_list` here in this method.

evm/vm/message.py
```diff
@@ -142,3 +143,73 @@ def storage_address(self, value):
...
+        validate_uint256(gas, title="Message.gas")
+        self.gas = gas
+
+        validate_uint256(gas_price, title="Message.gas_price")
+        self.gas_price = gas_price
+
+        validate_canonical_address(to, title="Message.to")
+        self.to = to
+
+        if sender != ENTRY_POINT:
```
pipermerriam: This check shouldn't be required since the `ENTRY_POINT` is a valid canonical address (at least I think it is)

NIC619: yeah `ENTRY_POINT` is a valid canonical address so we only validate if `sender` is not `ENTRY_POINT`

evm/vm/message.py
NIC619: ah yeah you are right, I thought they differ in how they handle `to` and `sender` but the effects are actually the same.

evm/vm/message.py
pipermerriam: Silly me.  I saw `==` even though it's clearly `!=`

NIC619:
@pipermerriam I add tests for `is_create` and intrensic_gas`. Will figure out how to test `execute_transaction`, `apply_message` and `apply_create_message` next. Or I can pick transaction and message execution and their tests out and put them in next PR if this PR is considered fat enough.

evm/utils/address.py
```diff
@@ -16,3 +20,15 @@ def force_bytes_to_address(value):
 
 def generate_contract_address(address, nonce):
     return keccak(rlp.encode([address, nonce]))[-20:]
+
+
+def generate_create2_contract_address(salt, code):
+    """
+    If contract is created by transaction, `salt` should be empty.
```
jannikluhn: Shouldn't in this case the salt be specified by the transaction somehow (maybe the first 32 bytes of `data`)? Otherwise one can't deploy the same contract twice, or am I missing something?

evm/vm/forks/sharding/validation.py
```diff
@@ -0,0 +1,14 @@
...
+
+def validate_sharding_transaction(evm, transaction):
+    # TODO:Update transaction validation logic in Sharding
+    # e.g. checking shard_id < SHARD_COUNT
+
+    # `validate_homestead_transaction` assumes underlying signature scheme
+    # to be ECDSA and checks signatures accordingly. So it's skipped for
+    # the reason that there won't be built-in signature scheme in Account
+    # Abstraction.
+    validate_frontier_transaction(evm, transaction)
```
jannikluhn: `validate_frontier_transaction` checks the balance of `transaction.sender` which would fail for `ShardingTransaction`s. It also ignores `transaction.access_list`. So I think we can't use this.

evm/utils/address.py
NIC619: The `salt` could be transaction nonce if `nonce` is not abstracted away eventually. Or if it is, one could simply append data to `code` and treat it as `salt`.

evm/vm/forks/sharding/validation.py
NIC619: yeah you are right. I will update it. Thanks!

evm/vm/forks/sharding/__init__.py
```diff
@@ -0,0 +1,266 @@
...
+        value=0,
+        data=data,
+        code=code,
+        create_address=contract_address,
+    )
+
+    #
+    # 2) Apply the message to the VM.
+    #
+    if message.is_create:
```
jannikluhn: Right now `message.is_create` is `True` even if `transaction.code` is empty (as in this case `message.code` is set to the contract code in line 70). So I think there's something wrong here, but I'm not sure what's the best way to fix it.

NIC619: ah that's also true, it's my mistake. And I don't see a clear and simple way around this either. Perhaps have `message.is_create` set by `execute_transaction` and `CREATE*` opcodes?

jannikluhn: Yeah, that sounds reasonable.

NIC619:
btw, @jannikluhn @pipermerriam I was thinking 1. adding nonce back to transaction format OR 2. completely abstract nonce away.
Since nonce abstraction is still under discussion, I keep it where it was(as part of account's data). So right now nonce is not checked in transaction validation and also not incremented either.

I would prefer 1 and if nonce is eventually going away we can simply remove the codes altogether.

evm/vm/forks/sharding/__init__.py
pipermerriam: I'm not sure if I'm reading this right, but if what you're suggesting involves mutating the `Message` object I would like to find another way as I think `Message` should be treated as immutable.

pipermerriam:
@NIC619 my comment about testing was not intended to mean that you needed to provide those tests now (though I don't mind if you do), just to say that those code paths are un-tested.  Maybe just a single *smoke* test for very basic transaction execution to know that there are no obvious major broken parts?  I assume that the JSON fixture based tests will be updated at some point and then we'll use those to test.

evm/vm/forks/sharding/validation.py
```diff
@@ -0,0 +1,31 @@
...
+    validate_transaction_access_list,
+)
+
+
+def validate_sharding_transaction(vm, transaction):
+    # `validate_homestead_transaction` assumes underlying signature scheme
+    # to be ECDSA and checks signatures accordingly.
+    # `validate_frontier_transaction` checks for the balance of transaction
+    # sender.
+    # Both of which are in confilct with Account Abstraction implementation.
```
pipermerriam: I think this whole comment can be removed with the understanding that this is just a new validation function.

evm/vm/message.py
```diff
@@ -142,3 +138,45 @@ def storage_address(self, value):
...
+            is_static=is_static,
+        )
+
+        if access_list is not None:
+            validate_access_list(access_list)
+        self.access_list = access_list
+
+    @property
+    def is_create(self):
+        return self.code not in (None, b'')
```
pipermerriam: Hrm, it'd be nice to have a single representation of *empty*.  What do you think about changing `message.code` to always be a bytestring so that `b''` is the only *empty* value and we can just `return bool(self.code)`

evm/vm/forks/sharding/__init__.py
NIC619: That makes sense. How about setting the `message.is_create` flag during message initialization? Since we can already pass in `create_address`, we can 
1. simply add a new `is_create` parameter 
OR
2. tell by checking the value of `create_address`
OR
3. just replace `create_address` with `is_create` (since in Account Abstraction, instead of `CREATE_CONTRACT_ADDRESS`, `to` should be set as the address of the new contract. Setting `storage_address` to `create_address` would be the same as leave it empty).

evm/rlp/transactions.py
```diff
@@ -160,3 +160,32 @@ class BaseShardingTransaction(rlp.Serializable):
...
+
+    #
+    # Validation
+    #
+    def validate(self):
+        """
+        Hook called during instantiation to ensure that all transaction
+        parameters pass validation rules.
+        """
+        if self.intrensic_gas > self.gas:
```
vbuterin: Correct spelling is "intrinsic".

evm/vm/forks/sharding/__init__.py
vbuterin: I'd say go for 2 or 3; keep it simple.

evm/rlp/transactions.py
NIC619: @pipermerriam I will update the naming in sharding branch along with this PR and update the one in master branch with a new PR.

pipermerriam:
We should wait for #244 to be merged as this will need to be refactored to the new architecture introduced in that PR.

tests/core/message-object/test_message_object.py
```diff
@@ -42,6 +44,27 @@ def _create_message(gas=1,
     )
 
 
+def _create_sharding_message(
```
pipermerriam: This doesn't appear to be used anywhere.

NIC619:
@pipermerriam rebased on #250 and updated to new computation logic(#244). If everything looks fine I will go ahead and merge this.

PR #239, opened 2017-12-21
Start implementing LES/2

gsalgado:
Depends on #229 

After next geth's release we should see a lot more LES/2 nodes in the network and then we should be able to drop LES/1


evm/p2p/les.py
```diff
@@ -223,9 +224,20 @@ class Proofs(Command):
...
         ('buffer_value', sedes.big_endian_int),
-        ('nodes', sedes.CountableList(sedes.CountableList(sedes.raw))),
+        ('proofs', sedes.CountableList(sedes.CountableList(sedes.raw))),
     ]
 
+    def decode_payload(self, rlp_data: bytes) -> _DecodedMsgType:
+        decoded = super().decode_payload(rlp_data)
+        # This is just to make Proofs messages compatible with ProofsV2, so that LightChain
+        # doesn't have to special-case them. Soon we should be able to drop support for LES/1
+        # anyway, and then all this code will go away.
```
pipermerriam: It's not clear to me which of the following is LES/1 and which is LES/2.  Can you explicitely label them.

evm/p2p/les.py
```diff
@@ -343,3 +355,54 @@ def send_get_contract_code(self, block_hash: bytes, key: bytes, request_id: int)
...
+            'protocolVersion': self.version,
+            'networkId': self.peer.network_id,
+            'headTd': head_info.total_difficulty,
+            'headHash': head_info.block_hash,
+            'headNum': head_info.block_number,
+            'genesisHash': head_info.genesis_hash,
+        }
+        cmd = StatusV2(self.cmd_id_offset)
+        self.send(*cmd.encode(resp))
+        self.logger.debug("Sending LES/Status msg: %s", resp)
```
pipermerriam: *nitpick*

I would expect this logging statement to happen *before* the `self.send(...)` call both based on the wording of the message, and a personal convention of placing the logging statement before the action if it's announcing that the action is *happening* and after if it's announcing that the action *happened*.  Reason here is that in the event that the action fails, we still get the logging message which is slightly more informative of what happened.

evm/p2p/les.py
```diff
@@ -343,3 +355,54 @@ def send_get_contract_code(self, block_hash: bytes, key: bytes, request_id: int)
...
+
+    def send_get_proof(self, block_hash: bytes, account_key: bytes, key: bytes, from_level: int,
```
pipermerriam: *linting nitpick*

Can we drop all of these to their own line.  I'm finding that signatures with type annotations and more than 2-3 arguments are really hard to read when they are on the same line.

evm/p2p/lightchain.py
```diff
@@ -438,15 +442,20 @@ def peer_block_height(peer: LESPeer):
         return reply['receipts'][0]
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
```
pipermerriam: Do you know if `alru_cache` appropriately handles decorating methods.  I know that other such decorators sometimes have trouble with the `self` argument since it normally shouldn't be included in the cache.

evm/p2p/les.py
gsalgado: This method is under the Proofs class, and there's another class named ProofsV2, so I thought that'd make it clear enough. I guess it wasn't clear enough here because of lack of context?

evm/p2p/les.py
gsalgado: Makes sense

evm/p2p/lightchain.py
gsalgado: It does include `self` in the cache's key, but that is the behaviour I'd expect in most cases, I think. Can you think of a situation where we'd want to reuse the cache across instances?

Anyway, I think it may not matter here as the LightChain instance is meant to be a singleton and trying to have multiple instances of it is not supported

evm/p2p/les.py
pipermerriam: Sorry, I mis-understood what I was looking at.  Disregard.

evm/p2p/lightchain.py
pipermerriam: Sounds good to me.  I just wanted to make sure that we weren't going to lose all of the cache benefits for some reason because of this.  👍

Issue #236, opened 2017-12-20
Implement `State` object


pipermerriam:
### What is wrong?

Currently, the `VM` class is part of the constructor for the `Computation` object.  Here are the places where it is currently used.

```
evm/logic/block.py:7:    block_hash = computation.vm.get_ancestor_hash(block_number)
evm/logic/block.py:13:    computation.stack.push(computation.vm.block.header.coinbase)
evm/logic/block.py:17:    computation.stack.push(computation.vm.block.header.timestamp)
evm/logic/block.py:21:    computation.stack.push(computation.vm.block.header.block_number)
evm/logic/block.py:25:    computation.stack.push(computation.vm.block.header.difficulty)
evm/logic/block.py:29:    computation.stack.push(computation.vm.block.header.gas_limit)
evm/logic/system.py:87:    computation.vm.logger.debug(
evm/logic/system.py:141:            computation.vm.logger.debug(
evm/vm/forks/spurious_dragon/utils.py:22:        yield computation.vm.block.header.coinbase
evm/vm/computation.py:76:        self.vm = vm
evm/vm/computation.py:197:            child_computation = self.vm.apply_create_message(child_msg)
evm/vm/computation.py:199:            child_computation = self.vm.apply_message(child_msg)
evm/vm/computation.py:280:        with self.vm.state_db(read_only, self.msg.access_list) as state_db:
```

Giving computation direct access to the `VM` is problematic for a number of reasons.

- poor isolation for what data and APIs are available during EVM execution.
- makes implementation of stateless clients difficult.
- makes VM execution inherently non-pure.

### How can it be fixed

To fix this, lets create a new `State` object.  Similar to how the `Transaction` class is specified for each `VM`, the `State` object should be similarly overrideable.

- `VM.get_state_class()`
- `VM.get_state()`
- `VM.state` (this could be a convenience `property` that just delegates to `VM.get_state()`.

The `State` object will need the following things.

- Access to the current block header information (`coinbase/timestamp/block_number/difficulty/gas_limit`.
- The ancestry hashes for the current block number.
- Access to the `StateDB`

The `State` object will also need the following APIs available to it.

- `VM.apply_message`
- `VM.apply_create_message`

I think the best way to accomplish this is to relocate the following APIs to live directly on the `State` object itself, removing them from the `VM` object.

- `VM.apply_message`
- `VM.apply_create_message`
- `VM.apply_computation`

[...]

PR #240, opened 2017-12-22
State Object

hwwhww:
### What was wrong?
This PR is for Implement `State` object #236

### How was it fixed?
1. Basically, followed #236, and some class methods for `State` usage only are also moved from `VM` class to `State` class.
    1. Snapshot and Revert
    2. `get_ancestor_hash()`
    3. `get_opcode_fn()`
2. `opcodes` and  `precompiles` are assigned in `@classmethod State.create_state(chaindb, block_header, opcodes, precompiles)`, and become instance objects of `State`.

#### Cute Animal Picture

cute


evm/logic/block.py
```diff
@@ -4,26 +4,26 @@
...
 
 
 def timestamp(computation):
-    computation.stack.push(computation.vm.block.header.timestamp)
+    computation.stack.push(computation.state.timestamp)
 
 
 def number(computation):
-    computation.stack.push(computation.vm.block.header.block_number)
+    computation.stack.push(computation.state.number)
```
pipermerriam: *naming nitpick*

Can we keep the `block_` prefix here so it's explicit that we're referring to a block number?

(does this get funky with the introduction of collations later?)

evm/logic/block.py
```diff
@@ -4,26 +4,26 @@
...
 
 
 def difficulty(computation):
-    computation.stack.push(computation.vm.block.header.difficulty)
+    computation.stack.push(computation.state.difficulty)
 
 
 def gaslimit(computation):
-    computation.stack.push(computation.vm.block.header.gas_limit)
+    computation.stack.push(computation.state.gaslimit)
```
pipermerriam: *naming nitpick*

Can we keep this as `gas_price` for naming consistency?

evm/state.py
```diff
@@ -0,0 +1,197 @@
...
+    # Snapshot and Revert
+    #
+    def snapshot(self):
+        """
+        Perform a full snapshot of the current state.
+
+        Snapshots are a combination of the state_root at the time of the
+        snapshot and the checkpoint_id returned from the journaled DB.
+        """
+        return (self.block_header.state_root, self.chaindb.snapshot())
```
pipermerriam: Idle musing/thought

Might be a nice API if snapshotting returned another `State` object.  This gets confusing/complex with the underlying `chaindb` which is *inherently* mutable.

Just something to think on.

evm/state.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        Revert the VM to the state at the snapshot
+        """
+        state_root, checkpoint_id = snapshot
+
+        with self.state_db() as state_db:
+            # first revert the database state root.
+            state_db.root_hash = state_root
+            # now roll the underlying database back
+            self.chaindb.revert(checkpoint_id)
```
pipermerriam: Can we dedent this line since it doesn't need to be within the context of the `state_db` block.

evm/state.py
```diff
@@ -0,0 +1,197 @@
...
+        while header.block_number != block_number:
+            header = self.chaindb.get_block_header_by_hash(header.parent_hash)
+        return header.hash
+
+    #
+    # classmethod
+    #
+    @classmethod
+    def create_state(cls, chaindb, block_header, opcodes, precompiles):
+        return cls(chaindb, block_header, opcodes, precompiles)
```
pipermerriam: This looks like a direct wrapper around the `__init__` method.  Seems like anywhere that `State.create_state(...)` is used you can just use `State(...)`

evm/vm/base.py
```diff
@@ -329,10 +222,30 @@ def clear_journal(self):
...
+        """
+        if cls._state_class is None:
+            raise AttributeError("No `_state_class` has been set for this VM")
+
+        return cls._state_class
+
+    def get_state(self):
+        """Return state object
+        """
+        return self.get_state_class().create_state(
```
pipermerriam: This should be identical to `self.get_state_class()(...)`

evm/vm/forks/frontier/state.py
```diff
@@ -0,0 +1,82 @@
...
+from evm.utils.hexadecimal import (
+    encode_hex,
+)
+from evm.utils.keccak import (
+    keccak,
+)
+
+
+class FrontierState(BaseState):
+    def apply_message(self, message):
```
pipermerriam: oooh, I think I like that we can now declare all of these methods on an actual object!

pipermerriam:
@gsalgado curious to hear your thoughts on this API change.

evm/logic/block.py
hwwhww: Yes, I will modify it and keep `block_` prefix.

(In current design, we haven't added collation header to `State` yet, but I think it's possible in the future.)

pipermerriam:
This is good to :shipit: 

I'd like to figure out fully removing access to the `chaindb` from the `VMState` but I think that's better done in a different pull request.

evm/logic/block.py
```diff
@@ -4,26 +4,26 @@
...
 
-    block_hash = computation.vm.get_ancestor_hash(block_number)
+    block_hash = computation.state.get_ancestor_hash(block_number)
 
     computation.stack.push(block_hash)
 
 
 def coinbase(computation):
-    computation.stack.push(computation.vm.block.header.coinbase)
+    computation.stack.push(computation.state.coinbase)
```
gsalgado: No longer having to reach that deep through the class hierarchy to get the current block header is a very nice improvement, but nevertheless I wouldn't expect to find it under an attribute named `state`. Same goes for `state.precompiles` and `state.opcodes`, so I'm wondering if we couldn't instead pass those things (block, opcodes, precompiles) separately to `Computation.__init__()`, move `apply_computation()` there and have a configurable `_computation_class` instead of `_state_class` in the `VM`

pipermerriam: Hrm, initially I like this but..  here are my thoughts.

- `Computation` class represents the VM state while the VM is *running*
- `VMState` class represents the VM state while the VM is *at rest*

I think maintaining this separation is valuable.

Maybe there is a middle ground here where the `VMState` is treated as immutable and all of the mutability and state transformation code moves into the `Computation`.  That implies the following API.

- `VMState.coinbase`, `VMState.block_number`, ...
- `VMState.get_block_hash(number)`
- `VMState.apply_transaction(...) -> next_vm_state, computation_result`  <---- this is a pure function that returns the subsequent `VMState` after computation and the computation result... or something like that.
- `Computation.apply_computation(vm_state, message)`
- `Computation.apply_message(...)`
- `Computation.apply_create_message(...)`
- `Computation.opcodes`
- `Computation.precompiles`

This is a really rough sketch of what this might look like but it seems potentially promising.

hwwhww: I like the idea if we can make `VMState` to be is treated as immutable by these changes.
(Before it, I was trying to figure out how to use another new immutable object to handle state transition.)
I can start to move these functions to be under `Computation`.

pipermerriam: > I can start to move these functions to be under `Computation`.

If you think this will work, lets try it.  

FYI, I'm still on *"sort-of-vacation"* time until next week but I'm setting aside time each day to check email and do code review to make sure I'm not holding anyone back.  Keep up the great work.

hwwhww:
New PR: #244

PR #241, opened 2017-12-25
Rename EIP150 to Tangerine Whistle.  Fixes #221

### LINKED ISSUE #221: Rename EIP150 fork to Tangerine Whistle
### What is wrong?

The `evm.vm.forks.eip150` should really be named `evm.vm.forks.tangerine_whistle` to conform to the current naming convention for the fork modules.

### How can it be fixed

Rename the module and all of the various places where imports happen from that module, as well as anything that's named as EIP150 which isn't specific to the actual EIP150 rules but rather refers to the fork.

NicolaiSoeborg:
### What was wrong?
The Tangerine Whistle hardfork was called EIP150 in the code.

### How was it fixed?
Refactoring.

#### Cute Animal Picture

x-mas cat


pipermerriam:
Thanks @NicolaiSoeborg

PR #247, opened 2018-01-03
Refactored VM, VMState, and Block + new apply_transaction

hwwhww:
### What was wrong?
For #195: pure apply_transaction + idea of #236
 
### How was it fixed?

The original goal is to implement:
```python
state_obj', reads, writes = apply_transaction(stateobj, transaction, blockdata, db)
```

Now we have:
```python
            vm_state = cls.get_state_class()(
                chaindb=witness_db,
                block_header=block.header,
                prev_headers=prev_headers,
                receipts=receipts,
            )
            computation, result_block, _ = vm_state.apply_transaction(
                transaction=transaction,
                block=block,
                is_stateless=True,
            )
```
* `computation.access_logs.reads` and `computation.access_logs.reads` are the reads and writes sets.
* `witness_db` is `db`. 
* `stateobj` information would be acquired from `vm_state` and `block`
* `blockdata` information would be be acquired from `block`

#### 3. Refactored `VM`, `VMState`,  and `Block`
1. Added `VMState` forks
2. Set `computation_class` in `VMState` and moved `VM.get_computation()` to `VMState.get_computation()`
3.  From `Block` to `VM`
	* Moved `Block.mine(**kwargs)` into `VM.pack_block(block, *args, **kwargs)`
4. From `Block` to `VMState` 
	* Moved `VM.execute_transaction(transaction)` function into `VMState.execute_transaction(transaction)`
5. From `Block` to `VMState`
	* Moved `Block.make_receipt(transaction, computation)` into `VMState.make_receipt(transaction, computation)`
6. Refactored and redesigned `apply_transaction` and `add_transaction`
	1. Moved `VM.apply_transaction(transaction)` function into `VMState.apply_transactionn(transaction)`
	2. Moved `Block.add_transaction(transaction, computation)` into `VMState. add_transaction(vm_state, transaction, computation)`
	* I keep old logic for verifying the “stateless” state transition works. Basically, the stateless mode is set on `VM._is_stateless` flag; if `VM.is_stateless` is `True`, use witness data as `chaindb`; otherwise, use  `Chain.chaindb` as before.
	* Assigned  `Block.db = BaseChainDB(MemoryDB()).db`, only for generating transaction_root and receipt_root in `VMState` level.
	* `VMState.apply_transaction` returns the updated `block` and `trie_data` to `VM.apply_transaction`, where `trie_data` is the key-value list of transactions and receipts data that should be store in `VM.chaindb` for non-stateless clients.
	* If `apply_transaction is tiggered` by `Chain`, the path would be: 
```python
Chain.apply_transaction(transaction) -> return computation, self.block
    └─VM.apply_transaction(transaction) 
        └─[is_stateless]
            └─VMState.apply_transaction(vm_state, transaction, block, is_stateless, witness_db) -> return computation, block, trie_data
                └─VMState.execute_transaction(transaction) -> return computation, block_header
                    └─Computation.apply_message(message) or Computation.apply_create_message(message) -> return computation
                └─VMState.add_transaction(vm_state, transaction, computation, block) -> return block, trie_data
        └─[not is_stateless]
            └─VMState.apply_transaction(vm_state, transaction, block, is_stateless, witness_db)
                └─VMState.execute_transaction(transaction) -> return computation, block_header
                    └─Computation.apply_message(message) or Computation.apply_create_message(message) -> return computation
            └─VM.add_transaction(transaction, computation) -> return computation, None, None
                └─VMState.make_receipt(transaction, computation) -> return receipt

```

#### 5. Create block (unfinished)
* To verify and demonstrate stateless state transition, there’s a new function `@classmethod create_block(transactions, transaction_witnesses, prev_state_root, parent_header, coinbase)`  to create a block with the given `transaction_witness`.
* Please see `tests/core/vm/test_vm.py::test_create_block`.
#### 9. Moved `VMState.get_parent_header(block_header, db)` and `VMState.get_prev_headers(last_block_hash, db)` into standalone functions

p.s. Sorry for that I reset and squash some commits and then my `git mv` turns into mv, so it's unclear to diff `evm/vm/vm_state.py` and `evm/vm_state.py`.

#### Cute Animal Picture

cute

---
TODOs for other PRs
1. Decouple `VMState.block_header` the `VMState` properties (blockhash, timestamp, block_number, difficulty, and gas_limit), because shard chains use `period_start_prevhash` block properties of main chain in contracts, but `VMState.state_db()` should use `VMState.collation_header`'s state_root.
2. Remove `is_stateless` flag, make it only one route.

I will open new issues later.


evm/computation.py
```diff
@@ -407,6 +399,13 @@ def apply_computation(cls, vm_state, message, opcodes, precompiles):
     #
     # Opcode API
     #
+    @property
+    def precompiles(self):
+        if self._precompiles is None:
+            return set()
```
pipermerriam: This should be a `dict` instead of a `set`

evm/db/state.py
```diff
@@ -49,6 +51,7 @@ def __init__(self, db, root_hash=BLANK_ROOT_HASH, read_only=False):
             self.db = ImmutableDB(db)
         else:
             self.db = db
+        self.db = TrackedDB(self.db)
```
pipermerriam: Instead of overwriting `self.db` here maybe move this up into the previous `if/else`

evm/rlp/blocks.py
```diff
@@ -1,9 +1,21 @@
...
 
+from evm.db.backends.memory import MemoryDB
+from evm.db.chain import BaseChainDB
+
 
 class BaseBlock(rlp.Serializable):
     db = None
 
+    def __init__(self, header, transactions=None, uncles=None):
+        self.db = BaseChainDB(MemoryDB()).db  # for generating transaction_root and receipt_root
```
pipermerriam: I'm not sure I understand the need for this.  Where is it used in the code?

evm/vm/base.py
```diff
@@ -147,6 +189,131 @@ def mine_block(self, *args, **kwargs):
...
+
+        header = block.header
+        provided_fields = set(kwargs.keys())
+        known_fields = set(tuple(zip(*BlockHeader.fields))[0])
+        unknown_fields = provided_fields.difference(known_fields)
+
+        if unknown_fields:
+            raise AttributeError(
+                "Unable to set the field(s) {0} on the `BlockHeader` class. "
+                "Received the following unexpected fields: {0}.".format(
```
pipermerriam: The `{0}` here should be `{1}` to correctly format the string.

evm/vm/base.py
```diff
@@ -147,6 +189,131 @@ def mine_block(self, *args, **kwargs):
...
+                )
+        block.state_root = vm_state.block_header.state_root
+
+        return block, vm_state
+
+    @classmethod
+    def create_block(
+            cls,
+            transactions,
+            transaction_witnesses,
```
pipermerriam: not sure if this is an improvement but...

If I understand correctly, we require that `len(transaction) == len(transaction_witnesses)`.  Rather than passing these in as separate lists, it might be good to use a data structure like the following:

```
transactions_and_witnesses = (
    (transaction_0, transaction_witness_0),
    (transaction_1, transaction_witness_1),
    (transaction_2, transaction_witness_2),
    ...
)
```

This approach builds in some minimal validation since you can then loop over them like this:

```python
for index, (transaction, transaction_witness) in enumerate(transactions_and_witnesses):
    ...
```

Not sure this is a huge improvement, but since the two arguments are inherently linked together, it might be nice to accept them already linked.

evm/vm/forks/byzantium/computation.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class ByzantiumComputation(SpuriousDragonComputation):
+    def __init__(self, vm_state, message):
+        super(ByzantiumComputation, self).__init__(
+            vm_state,
+            message,
+        )
+        # Override
+        self.opcodes = BYZANTIUM_OPCODES
```
pipermerriam: Is there a reason you are doing this in the `__init__` rather than setting them as class properties?

evm/vm/forks/frontier/computation.py
```diff
@@ -14,8 +18,26 @@
...
+
 
 class FrontierComputation(BaseComputation):
+    def __init__(self, vm_state, message):
+        super(FrontierComputation, self).__init__(
+            vm_state,
+            message,
+        )
+        # Override
+        self.opcodes = FRONTIER_OPCODES
```
pipermerriam: Same suggestion here to remove these from `__init__` and set them as class properties.

evm/vm_state.py
```diff
@@ -0,0 +1,299 @@
...
+from evm.db.tracked import (
+    AccessLogs,
+)
+
+
+class BaseVMState(object):
+    _chaindb = None
+    block_header = None
+    computation_class = None
+    access_logs = AccessLogs()
```
pipermerriam: This is probably problematic since the object will be shared across all instances of this class and the `access_logs` is mutable.  This should instead be initialized in the `__init__` method.

evm/vm_state.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        return self._chaindb.get_block_header_by_hash(block_hash)
+
+    def get_parent_header(self, block_header):
+        """
+        Returns the header for the parent block.
+        """
+        return self.get_block_header_by_hash(block_header.parent_hash)
+
+    def is_key_exsits(self, key):
```
pipermerriam: spelling/typo

evm/vm_state.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        return self._chaindb.exists(key)
+
+    #
+    # Computation
+    #
+    def get_computation(self, message):
+        """Return state object
+        """
+        if self.computation_class is not None:
```
pipermerriam: Slightly more readable version of this.

```python
if self.computation_class is None:
    raise AttributeError("No `computation_class` has been set for this VMState")
else:
    return self.computation_class(self, message)
```

evm/vm_state.py
```diff
@@ -0,0 +1,299 @@
...
+    #
+    # Execution
+    #
+    @classmethod
+    def apply_transaction(
+            cls,
+            vm_state,
+            transaction,
+            block,
+            is_stateless=True,
```
pipermerriam: I think there is a route for us to remove the need for this flag.  We can have `VM.apply_transaction` **always** be pure/stateless, and then modify `Chain.apply_transaction` to update the local vm state before returning the `computation` object.

Assuming there is nothing wrong with that approach I like it a lot more than using a flag.

evm/vm_state.py
```diff
@@ -0,0 +1,299 @@
...
+            transaction,
+            block.db,
+        )
+        receipt_root_hash, receipt_db = cls.add_trie_node_to_db(
+            block.header.receipt_root,
+            index_key,
+            receipt,
+            block.db,
+        )
+        trie_data = {}
```
pipermerriam: You can replace these three lines (and remove the mutation) with:

```
trie_data = cytoolz.merge(tx_db.wrapped_db.kv_store, receipt_db.wrapped_db.kv_store)
```

setup.py
```diff
@@ -25,7 +25,7 @@
         "py-ecc==1.4.2",
         "rlp==0.4.7",
         "eth-keys==0.1.0b3",
-        "trie>=0.3.2",
+        "trie==0.3.2",
```
pipermerriam: What was the breakage with `1.0.0`?  I thought it was backwards compatible.

hwwhww: I haven't dug into the root cause. The error message is like:
https://travis-ci.org/hwwhww/py-evm/jobs/324634985
```
tests/json-fixtures/test_state.py:247: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
evm/vm/base.py:39: in __init__
    self.block = block_class.from_header(header=header, chaindb=self.chaindb)
evm/vm/forks/frontier/blocks.py:94: in from_header
    transactions = chaindb.get_block_transactions(header, cls.get_transaction_class())
.tox/py35-native-state-byzantium/lib/python3.5/site-packages/eth_utils/functional.py:22: in inner
    return callback(fn(*args, **kwargs))
evm/db/chain.py:163: in get_block_transactions
    if transaction_key in transaction_db:
.tox/py35-native-state-byzantium/lib/python3.5/site-packages/trie/hexary.py:397: in __contains__
    return self.exists(key)
.tox/py35-native-state-byzantium/lib/python3.5/site-packages/trie/hexary.py:106: in exists
    return self.get(key) != BLANK_NODE
.tox/py35-native-state-byzantium/lib/python3.5/site-packages/trie/hexary.py:62: in get
    root_node = self._get_node(self.root_hash)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <trie.Trie object at 0x7f6c2cc46908>, node_hash = None
    def _get_node(self, node_hash):
        if node_hash == BLANK_NODE:
            return BLANK_NODE
        elif node_hash == BLANK_NODE_HASH:
            return BLANK_NODE
    
>       if len(node_hash) < 32:
E       TypeError: object of type 'NoneType' has no len()
```

evm/rlp/blocks.py
hwwhww: In non-stateless mode, the transaction trie is rebuilt via `BaseChainDB.add_transaction( block_header, index_key, transaction)`:
```python
    def add_transaction(self, block_header, index_key, transaction):
        transaction_db = Trie(self.db, root_hash=block_header.transaction_root)
        transaction_db[index_key] = rlp.encode(transaction)
        return transaction_db.root_hash
```
And it is triggered in `VM.add_transaction(transaction, computation)`:
https://github.com/hwwhww/py-evm/blob/ec9ff9c63ff96e484150bc62ab862aed5505a03d/evm/vm/base.py#L79

For stateless mode, `BaseBlock.db` is only used for storing transaction trie and receipt trie in `VMState` level 
during `create_block` because I'd want to remove `chaindb` from `VMState` eventually.

In `VMState.add_transaction`:
https://github.com/hwwhww/py-evm/blob/ec9ff9c63ff96e484150bc62ab862aed5505a03d/evm/vm_state.py#L223
```python
        # Get trie roots and changed key-values.
        tx_root_hash, tx_db = cls.add_trie_node_to_db(
            block.header.transaction_root,
            index_key,
            transaction,
            block.db,
        )
        receipt_root_hash, receipt_db = cls.add_trie_node_to_db(
            block.header.receipt_root,
            index_key,
            receipt,
            block.db,
        )
```

And then insert the transaction to `block.db`:
https://github.com/hwwhww/py-evm/blob/ec9ff9c63ff96e484150bc62ab862aed5505a03d/evm/vm_state.py#L247-L254
```python
    @staticmethod
    def add_trie_node_to_db(root_hash, index_key, node, db):
        """
        Add transaction or receipt to the given db.
        """
        trie_db = Trie(db, root_hash=root_hash)
        trie_db[index_key] = rlp.encode(node)
        return trie_db.root_hash, trie_db.db
```

evm/vm/base.py
hwwhww: Yes, we do require that `len(transaction) == len(transaction_witnesses)`. 
From sharding doc:
```python
Transaction package format
    [
        [nonce, acct, data....],    # transaction body (see below for specification)
        [node1, node2, node3....]   # witness
    ]
```
I will modify it to the form you suggested, thank you. :)

evm/vm/forks/byzantium/computation.py
hwwhww: oh, sorry, `BYZANTIUM_OPCODES` was passed by `VMState` during my previous implementation and I forgot to modify this part. I'll modify it to set them to class properties.

setup.py
gsalgado: That's because `Trie.__init__()` doesn't upcall -- it just issues a warning

evm/vm_state.py
hwwhww: Agree! Now the non-stateless mode is only for verifying the stateless mode result in case I screw up. Could we remove the non-stateless mode later?

setup.py
pipermerriam: fixed in #249

evm/vm_state.py
pipermerriam: Yes, doing it separately is fine.  Mind opening an issue to track it and linking it here.

evm/rlp/blocks.py
pipermerriam: I'm wondering if we can take a different approach to decouple some of this.

The [`mk_receipt_root`](https://github.com/ethereum/pyethereum/blob/953a8a6cdc0bbcd8a10733ca6935525674b9c544/ethereum/common.py#L183) function in the pyethereum codebase is probably the utility we need to bring over.

Then, rather than keeping track of the trie for receipts and transactions we could very easily just rebuild it on demand each time we add a new transaction/receipt.  Thoughts?

hwwhww: I did rewrite `mk_receipt_root` for debugging during implementing, it would be much cleaner if we use `mk_receipt_root` to get root. The trade-offs are:

1. Since we want to get the "current block" after each time we call `VMState.apply_transaction`, if `import_block` is called, the trie-rebuilding will insert the all the previous receipts in this block during each `VMState.apply_transaction` calling. Would it be more inefficient?

2. I find that if we keep the current code, **maybe** we don't need to store `VMState.receipts` in `VMState` level. I'm still checking if it's true, right now it seems that we only require maintaining `VMState.receipts` for rebuilding receipt trie.

If 2 is true, it won't be a big deal to maintain `VMState.receipts`.

pipermerriam: I'm comfortable with the overhead with rebuilding the trie each time.  If the overhead turns out to be significant we can address it at that time.

pipermerriam: by the way, you do what you think is best.  This idea isn't coming from me having a fully formed solution, but rather trying to further simplify the `Block` object and remove some coupling and database access.  It may very well be that it doesn't provide a worthwhile improvement.

hwwhww: Totally agree your point. Either is fine with me for now, let me figure out it after more refactorings of `create_block`.

hwwhww:
#### Changelog
1. Added VMState.prev_headers class instance, now all the parent_header queries in `VMState` level are using `prev_header` data.
2. Fixed `VM.create_block()`
3. Added more tests of `test_vm.py` and `test_vm_state.py`
4. Use `VMState.receipts` to rebuild the trie root in every `apply_transaction` - cleaner!
5. Refactored reward methods: moved from `base.py` to Frontier
6. Added sample ReStructuredText comment: https://github.com/ethereum/py-evm/pull/247/commits/a1e13a4c6a89836371d3cd1bb9990c2fea2be579

#### Seeking for @pipermerriam 's advice and wisdom about mutating `VMState` & `chaindb`:

In this commit https://github.com/ethereum/py-evm/pull/247/commits/e1c854c97540b63ebd41f196c83a35dad, I’m thinking about making `VMState` & `chaindb` mutable would let the code simpler.

Because:
1. I hope to remove `copy.deepcopy(vm_state)` and `copy.deepcopy(witness_db)` from `VMState.apply_transaction`.
2. I find that we initialize a new vm_state every time before we call `VMState.apply_transaction()`.
3. If `computation.is_error`, the `chaindb` would revert anyway.
So I think it seems okay to mutate these `vm_state` and `vm.chaindb` (== `vm.state.chaindb`) in `apply_transaction` in our current use cases?

But, on the other hand:
1. If the clients want to **generate transaction witness** and then broadcast the transaction. package with witness, they have to revert themselves or give a copy.deepcopy(vm.chaindb) before calling `vm_state.apply_transaction()`?
2. Alternatively, `VMState` can accept **kv_store dict** as parameter, and then initialize the `VMState.chaindb` in `VMState` Level. **kv_store dict** equals ** transaction_witness** in `VM.create_block` function, the dict won’t be too much in `VM.create_block` since it’s only transaction witness, but will be huge in `VM.apply_transaction()`.  https://github.com/hwwhww/py-evm/blob/81b4b1ae1c854c97540b63ebd41f196c83a35dad/evm/vm/base.py#L255
3.  It’s not on the table now, but my rough though is that separating `vm.chaindb`from `VMState` entirely would help for parallelizability.

As I saw in `create_block`, this change won’t affect on stateless clients too much because they use transaction witnesses. But what do you think of the memory issue or other possible issues for non-stateless client? Thank you.

pipermerriam:
Both options seem to have comprable trade-offs so I'd say go with whichever seems best.  

Allowing the mutation to remain for the time being is fince with the understanding that we'll be looking for a clean way to remove it in the future.  I'd rather give us time to figure out a solution we like rather than push forward with a solution we're not sure about.

hwwhww:
@pipermerriam 
Thank you for sharing your thought, let's keep it mutable now. 👍

evm/utils/headers.py
```diff
@@ -65,3 +69,29 @@ def compute_gas_limit(parent_header, gas_limit_floor):
...
+    Generate BlockHeader from state_root and parent_header
+    """
+    header = BlockHeader(
+        difficulty=compute_difficulty(parent_header, timestamp),
+        block_number=(parent_header.block_number + 1),
+        gas_limit=compute_gas_limit(
+            parent_header,
+            gas_limit_floor=GENESIS_GAS_LIMIT,
+        ),
+        timestamp=max(timestamp, parent_header.timestamp + 1),
```
pipermerriam: I think that rather than using `max` here this should throw an exception if `timestamp <= parent_header.timestamp`.  It feels wrong to override the `timestamp` argument in this way.

evm/utils/headers.py
```diff
@@ -65,3 +69,29 @@ def compute_gas_limit(parent_header, gas_limit_floor):
         return parent_header.gas_limit + decay
     else:
         return gas_limit
+
+
+def generate_header_from_parent_header(
+        compute_difficulty,
+        parent_header,
+        timestamp,
+        coinbase=b'\x35' * 20,
```
pipermerriam: Where does `b'\x35` * 20` come from and can we get it moved into a constant variable somewhere?

evm/utils/headers.py
```diff
@@ -65,3 +69,29 @@ def compute_gas_limit(parent_header, gas_limit_floor):
         return parent_header.gas_limit + decay
     else:
         return gas_limit
+
+
+def generate_header_from_parent_header(
+        compute_difficulty,
```
pipermerriam: naming nitpick.  Might be a bit clearer that this is a callable if it was named `compute_difficulty_fn` or something else with a clear indicator that it's a callable.

evm/utils/state.py
```diff
@@ -30,3 +37,16 @@ def diff_state_db(expected_state, state_db):
                     actual_storage_value,
                     expected_storage_value,
                 )
+
+
+# Make the root of a receipt tree
+def make_trie_root_and_nodes(transactions):
```
pipermerriam: Since it looks like we're introducing binary tries, do you think it's worth changing the name of this DB to something like `make_hexary_trie_root_and_nodes`?

evm/vm/forks/byzantium/computation.py
```diff
@@ -0,0 +1,28 @@
...
+)
+from evm.vm.forks.frontier.computation import FRONTIER_PRECOMPILES
+from evm.vm.forks.spurious_dragon.computation import SpuriousDragonComputation
+
+from .opcodes import BYZANTIUM_OPCODES
+
+BYZANTIUM_PRECOMPILES = merge(
+    FRONTIER_PRECOMPILES,
+    {
+        force_bytes_to_address(b'\x05'): precompiles.modexp,
```
pipermerriam: Note to self.  This function might be better named something like:

- `expand_bytes_to_address` 
- `pad_bytes_to_address`
- `pad_to_20_byte_address` <-- maybe most explicit

evm/vm/forks/frontier/__init__.py
```diff
@@ -1,202 +1,49 @@
...
+        UNCLE_DEPTH_PENALTY_FACTOR + uncle.block_number - block_number
+    ) // UNCLE_DEPTH_PENALTY_FACTOR
 
-            # TODO: this balance setting is likely superflous and can be
-            # removed since `delete_account` does this.
-            state_db.set_balance(account, 0)
-            state_db.delete_account(account)
 
-    return computation
+def _frontier_get_nephew_reward(cls):
```
pipermerriam: Good choice moving these into the frontier namespace.

evm/vm/forks/frontier/vm_state.py
```diff
@@ -0,0 +1,320 @@
...
+                raise ValidationError(
+                    "`timestamp` is equal to the parent block's timestamp\n"
+                    "- block : {0}\n"
+                    "- parent: {1}. ".format(
+                        block.header.timestamp,
+                        parent_header.timestamp,
+                    )
+                )
+
+        # XXX: Should these and some other checks be moved into
```
pipermerriam: Maybe open an issue to revisit this separately.

evm/vm_state.py
```diff
@@ -0,0 +1,325 @@
...
+        return self.block_header.gas_limit
+
+    #
+    # chaindb
+    #
+    # @property
+    # def chaindb(self):
+    #     return self._chaindb
+
+    def set_chaindb(self, db):
```
pipermerriam: This doesn't appear to be used anywhere.

evm/vm_state.py
```diff
@@ -0,0 +1,325 @@
...
+            self.block_header.state_root = state.root_hash
+
+        self.access_logs.reads.update(state.db.access_logs.reads)
+        self.access_logs.writes.update(state.db.access_logs.writes)
+
+        # remove the reference to the underlying `db` object to ensure that no
+        # further modifications can occur using the `State` object after
+        # leaving the context.
+        state.db = None
+        state._trie = None
```
pipermerriam: Note to self:

There was a `decomission()` method added to the state database object to remove the need to access this private variable.  We should backport that improvement from the sharding branch.

evm/vm_state.py
```diff
@@ -0,0 +1,325 @@
...
+    def revert(self, snapshot):
+        """
+        Revert the VM to the state at the snapshot
+        """
+        state_root, checkpoint_id = snapshot
+
+        with self.state_db() as state_db:
+            # first revert the database state root.
+            state_db.root_hash = state_root
+            # now roll the underlying database back
```
pipermerriam: This comment placement seems wrong, assuming I'm correct that it applies to the next statement `self._chaindb.revert(checkpoint_id)`.

evm/vm_state.py
```diff
@@ -0,0 +1,325 @@
...
+        header = self.get_block_header_by_hash(self.block_header.parent_hash)
+        while header.block_number != block_number:
+            header = self.get_block_header_by_hash(header.parent_hash)
+        return header.hash
+
+    def get_block_header_by_hash(self, block_hash):
+        """
+        Returns the block header by hash.
+        """
+        for value in self.prev_headers:
```
pipermerriam: It looks like we could store `self.prev_headers` as a mapping from `hash => header` so that we can do O(1) lookups rather than repeatedly enumerating this list to find the necessary header.

evm/utils/state.py
hwwhww: How about passing Trie class as a parameter? Otherwise, we will need to refactor and implement another `make_binary_trie_root_and_nodes` later.
 
```python
def make_trie_root_and_nodes(transactions, trie_class=HexaryTrie):
    chaindb = BaseChainDB(MemoryDB())
    db = chaindb.db
    transaction_db = trie_class(db)
    ...
```

evm/utils/headers.py
hwwhww: I'll remove it.

evm/vm/base.py
```diff
@@ -188,8 +360,53 @@ def get_block_class(cls):
...
+
+    @staticmethod
+    def get_parent_header(block_header, db):
+        """
+        Returns the header for the parent block.
+        """
+        return db.get_block_header_by_hash(block_header.parent_hash)
+
+    @staticmethod
+    def get_block_header_by_hash(block_hash, db):
```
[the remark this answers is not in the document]
hwwhww: Not really, I just left them with `@classmethod def get_block_by_header(cls, block_header, db)`. Should I move `@staticmethod def get_parent_header(block_header, db)` and `@classmethod def get_prev_headers(cls, last_block_hash, db)` to `evm.utils.db`?

evm/vm/base.py
```diff
@@ -234,40 +471,31 @@ def get_state_class(cls):
 
         return cls._state_class
 
-    def get_state(self):
+    def get_state(self, chaindb=None, block_header=None, prev_headers=None):
         """Return state object
         """
+        if chaindb is None:
+            chaindb = self.chaindb
+        if block_header is None:  # TODO: remove
```
gsalgado: Can you maybe explain (in the comment) why we want to remove this?

evm/vm/base.py
```diff
@@ -234,40 +471,31 @@ def get_state_class(cls):
 
         return cls._state_class
 
-    def get_state(self):
+    def get_state(self, chaindb=None, block_header=None, prev_headers=None):
```
gsalgado: No callsites seem to pass a prev_headers argument to this method. Would it make sense to remove it?

evm/vm/forks/byzantium/vm_state.py
```diff
@@ -0,0 +1,22 @@
...
+
+class ByzantiumVMState(SpuriousDragonVMState):
+    computation_class = ByzantiumComputation
+
+    @staticmethod
+    def make_receipt(vm_state, transaction, computation):
+        old_receipt = _make_frontier_receipt(vm_state, transaction, computation)
+
+        receipt = Receipt(
+            state_root=b'' if computation.is_error else b'\x01',
```
gsalgado: What does b'\x01' mean here? Should it be a constant?

evm/vm_state.py
```diff
@@ -0,0 +1,325 @@
...
+            computation = self.computation_class(self, message)
+        return computation
+
+    #
+    # Execution
+    #
+    @classmethod
+    def apply_transaction(
+            cls,
+            vm_state,
```
gsalgado: It feels weird to have a `BaseVMState` method that takes a `vm_state` argument, but maybe I'm missing something?

evm/vm_state.py
```diff
@@ -0,0 +1,325 @@
...
+            transaction,
+            block,
+            is_stateless=True):
+        """
+        Apply transaction to the given block
+
+        :param vm_state: the VMState object
+        :param transaction: the transaction need to be applied
+        :param block: the block which the transaction applies on
+        :param is_stateless: if is_stateless, call VMState.add_transactionto set block
```
gsalgado: It actually calls `cls.add_transaction()`

tests/core/vm/test_vm_state.py
```diff
@@ -47,3 +57,163 @@ def test_state_db(state):  # noqa: F811
...
+    vm_state1 = FrontierVMState(
+        chaindb=chaindb1,
+        block_header=block_header1,
+        prev_headers=prev_headers,
+        receipts=[],
+    )
+    parent_header = copy.deepcopy(prev_headers[0])
+
+    computation, block, _ = FrontierVMState.apply_transaction(
+        vm_state1,
```
gsalgado: This is what I meant; it doesn't feel right to call a classmethod passing an instance of that same class as its first argument. I guess that's because we wanted to make `apply_transaction()` a pure function, but then maybe we should make it into a standalone function as well, to avoid this weirdness when calling it?

evm/utils/headers.py
hwwhww: Yes, you're right, it was supposed to be checking current time (`time.time()`) and `parent_header.timestamp + 1` if `timestamp is None`. I'll update it. Thank you.

evm/vm/forks/frontier/vm_state.py
hwwhww: >  XXX: Should these and some other checks be moved into VM.validate_block(), as they apply to all block flavours?

Ah, these two lines and this function were copied from `Block.validate()`. `VMState.validate_block()` became kind of `applying to all block flavors` now.
Another future issue is that this function would need to be refactored for sharding #256 since there's no `Collation.uncle`.

evm/vm_state.py
hwwhww: I found that I missed that if we store `self.prev_headers` as a list is good for access `block.blockhash(uint blockNumber)` in contract code, we could make `VMState.get_ancestor_hash(block_number)` get block_hash via block_number directly since the `prev_headers` should be chained already:
```python
    def get_ancestor_hash(self, block_number):
        """
        Return the hash of the ancestor with the given block number.
        """
        ancestor_depth = self.block_header.block_number - block_number - 1
        if (ancestor_depth >= MAX_PREV_HEADER_DEPTH or
                ancestor_depth < 0 or
                ancestor_depth >= len(self.prev_headers)):            
            return b''
        header = self.prev_headers[ancestor_depth]
        return header.hash
```
Then we get O(1) for BLOCKHASH opcode.

And since `VMState` has its block properties, we can make `VMState.get_parent_header(block_header)` into `@property parent_header()` for 
 `VMState.validate_block(block)`:
```python
    @property
    def parent_header(self):
        return self.prev_headers[0]
```

But, unfortunately, `VMState.validate_uncle(uncle)` still requires calling `VMState.get_block_header_by_hash(uncle.parent_hash)`:
https://github.com/hwwhww/py-evm/blob/81b4b1ae1c854c97540b63ebd41f196c83a35dad/evm/vm/forks/frontier/vm_state.py#L305

Look on the bright side, `MAX_UNCLE_DEPTH` is 6.

With these changes, I'd vote for accessing via `block_number` directly for sharding clients. But for non-sharding clients, they may call `validate_uncle` more than call BLOCKHASH(many-ancestors-ago).

p.s. EIP 96 would make `BLOCKHASH` very different: https://github.com/ethereum/EIPs/pull/210

evm/vm/forks/byzantium/vm_state.py
hwwhww: It was copied from `evm.vm.forks.byzantium.blocks`:
https://github.com/ethereum/py-evm/blob/master/evm/vm/forks/byzantium/blocks.py#L30

This protocol change is from [EIP 98/658](https://github.com/Arachnid/EIPs/blob/d1ae915d079293480bd6abb0187976c230d57903/EIPS/eip-658.md) Embedding transaction status code in receipts, 1 for success, 0 for failure.

tests/core/vm/test_vm_state.py
hwwhww: True! Actually, I think after the discussion in https://github.com/ethereum/py-evm/pull/247#issuecomment-356301856 and keeping VMState mutable, we can change them back to regular instance method `def apply_transaction(self, transaction, block, is_stateless=True)`.

hwwhww: True!
Actually, I think after the discussion in https://github.com/ethereum/py-evm/pull/247#issuecomment-356301856 and keeping VMState mutable, another way we can do is to change them back to regular instance method `def apply_transaction(self, transaction, block, is_stateless=True)`. And so do `VMState.execute_transaction(...)` and `VMState.make_receipt(...)` methods.

@pipermerriam what do you think?

evm/vm/forks/byzantium/vm_state.py
pipermerriam: I think I support moving these two values into named constants to improve readability.

tests/core/vm/test_vm_state.py
pipermerriam: Yes, going back to regular instance methods seems to make the most sense.

evm/vm/base.py
hwwhww: Ah sorry, I was thinking about if I can store `block_header` in `prev_headers` in some cases, but it turns out because `block_header` is more like a container object in `VMState`, it's more reasonable to leave `block_header` alone now. I'll remove this TODO tag.

Furthermore, there's another TODO issue I mention in this PR and #256, for sharding clients, we need to decouple `VMState.block_header` the `VMState` properties (`blockhash, timestamp, block_number, difficulty, and gas_limit)`, in that case, the parameters of `VMState.__init__` may become something like:
```python
def __init__(self, chaindb, state_root, block_header, prev_headers, receipts=[])
```
Thanks!

hwwhww:
#### Changelog
1. Squashed some old commits
2. Passing `trie_class` to `make_trie_root_and_nodes`. `trie_class` could be `HexaryTrie` or `BinaryTrie`.
3. Moved `VMState.get_parent_header(block_header, db)` and `VMState.get_prev_headers(last_block_hash, db)` into standalone functions
4. Refactored `VMState.get_ancestor_hash(block_number)` and made `VMState.get_parent_header(block_header)` into "the" VMState.parent_header
5. Changed some class methods of `VMState` back to regular instance methods
6. Clean up

@pipermerriam @gsalgado 
Please take a look, thank you!

pipermerriam:
:shipit:

Issue #246, opened 2018-01-02
Set sharding protocol constants and configuration


hwwhww:
### What is wrong?
Set sharding protocol constants and configuration

### How can it be fixed

1. Protocol constants are listed here: https://github.com/ethereum/sharding/blob/develop/docs/doc.md#constants
  Find an appropriate place to set these parameters.

2. Node configuration
    * Related issue: Related to https://github.com/ethereum/py-evm/issues/198
    * Sharding only configuration: (Ref: https://github.com/ethereum/sharding/blob/develop/docs/doc.md#client-logic)
        * `validator_address`: "0x..." OR null
        * `watching`: [list of shard IDs]
    * DB setting (TBD): one DB for each shard

We can start `1. Protocol constants` now and leave `2. Node configuration` after #198 is solved or after we start `ShardManager` implementation.

pipermerriam:
> Find an appropriate place to set these parameters.

* `evm.constants` - module is getting a bit full but a fine default location to start with.
* `evm.vm.forks.sharding.constants` - where the rest of the sharding code lives.
* `evm.sharding.constants` - assumes creation of a new `evm.sharding` module which might be appropriate.

Also, if these values need to be configurable and not *true* constants, I would propose making them configurable using environment variables.

```python
LOOKAHEAD_PERIODS = int(os.environ.get('PYEVM_SHARING_LOOKAHEAD_PERIODS', 4))
```

This pattern is fine, however, it will execute once at module load and be done.  I'd suggest something more akin to this so that it can be changed at runtime.

```python
def get_sharding_config():
    return {
        'LOOKAHEAD_PERIODS': int(os.environ.get('PYEVM_SHARING_LOOKAHEAD_PERIODS', 4)),
        ...
    {
```

This way anything that needs these constants, or to modify them at runtime can do so.

hwwhww:
**Part 1: Protocol constants(config)** is done via #248.
Move **Part 2: Protocol constants** to Stage 2.

PR #248, opened 2018-01-04
Add config.py for sharding

mhchia:
### What was wrong?
As https://github.com/ethereum/py-evm/issues/246 pointed, there isn't a constant file for sharding.

### How was it fixed?
I think `evm.vm.forks.sharding.constants` is a suitable path for constants for the sharding protocol.
For the future needed configs specified in https://github.com/ethereum/sharding/blob/develop/docs/doc.md#client-logic , maybe after that part is implemented, we can add them to another file `config.py` in the same directory.

#### Cute Animal Picture

Hedgehog


evm/vm/forks/sharding/constants.py
```diff
@@ -0,0 +1,21 @@
...
+# the number of blocks in one `period`
+PERIOD_LENGTH = 5
+# the gas limit of one collation
+COLLATION_GASLIMIT = 10 ** 7
+# the number of shards
+SHARD_COUNT = 100
+# the gas limit of verifying a signature
+SIG_GASLIMIT = 40000
+# the reward for creating a collation
+COLLATOR_REWARD = int(0.001 * denoms.ether)
```
pipermerriam: floating point math seems like a bad choice here.  Can you change this to either specify an exact integer value **or** use `eth_utils.to_wei('0.001', 'ether')`

pipermerriam:
https://github.com/ethereum/py-evm/pull/249 should fix the CI failures.

evm/vm/forks/sharding/constants.py
mhchia: sorry for that. I prefer to use `eth_utils.to_wei('0.001', 'ether')`. Thank you:)

vbuterin:
Shouldn't this technically be ideally implemented as part of a config file?

mhchia:
@vbuterin @pipermerriam I have changed the file to config.py, using the pattern introduced in #246 :) Please see if there is any problem if you have time.
Thank you!

pipermerriam:
Thoughts on adopting a library to do the environment variable extraction in a slightly more readable way?  This is a library that I wrote and used pretty extensively for a number of years.

https://github.com/simpleenergy/env-excavator

Rather than adopting it as a dependency, I think it'd be reasonable to just vendor a minimalistic version of that utility into the `py-evm` codebase.  My *plans* for py-evm involve a **lot** more environment variable based configuration so concise utilities for extracting typed values from environment variables seems useful.

mhchia:
@pipermerriam I think that would be great. I'm really not sure how minimal should it be, though I will only need `env_int` in `config.py`. So if I'm going to introduce the utility in this PR, in the first thought I will put the whole https://github.com/simpleenergy/env-excavator/blob/master/excavator/utils.py in `evm.utils` and rename it to `excavator.py`.
Is this a reasonable way ? Thank you:)

pipermerriam:
I think it's reasonable to pull all of those utils.  If I were to exclude any it might be the iso8601 datetime one since I don't know of any use cases we'd have for that (yet).  Names for the utils module that come to mind:

- `evm.utils.env`
- `evm.utils.config`
- `evm.utils.excavator`

I think `env` is my favorite.  I think `config` might end up being confusing.  `excavator` is clever but I think just calling it `env` is a bit more direct/explicit.

mhchia:
@pipermerriam I think `evm.utils.env`  is good! Thank you for the suggestion.

mhchia:
@pipermerriam One question is, I am considering whether to copy the tests in excavator to here, since there are so many tests in `excavator`. My thought is since it is tested in the original repo, if we are not going to change it in `py-evm`, it seems no need to copy them here. Do you have any suggestion?
Thank you!

pipermerriam:
@mhchia the test suite should be brought over with it.  At some point we'll change something and then we'll need a test suite to know it works.

mhchia:
@pipermerriam No problem I get it. Have added them:)
Thank you!

pipermerriam:
Good to merge when tests pass.  The environment variable extraction code should get pulled up to master at some point.

pipermerriam:
I restarted the travis test that failed due to network issues.  Lets see if things go green.

mhchia:
@pipermerriam I changed some code in the latest commit after the rebase. Could you please help me to see if the code is fine or not? Thank you so much!

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -15,11 +15,14 @@
...
-    DEPOSIT_SIZE,
-    GASPRICE,
-    TX_GAS,
+    get_sharding_config,
 )
 
+sharding_config = get_sharding_config()
+DEPOSIT_SIZE = sharding_config['DEPOSIT_SIZE']
+GAS_PRICE = sharding_config['GAS_PRICE']
+DEFAULT_GAS = sharding_config['DEFAULT_GAS']
```
pipermerriam: Rather than storing these at the module level, I would recommend storing the `sharding_config` as a property of the `VMC` instance.  Populate it in the `__init__` method and then pull values from it from whatever property you set it to:

```python
def sample(self, shard_id, gas=None):
    if gas is None:
        gas = self.config['DEFAULT_GAS']
    ...
```

Reason for this pattern is:

* Allow for two independent VMC contract instances which use different config values.
* Allow for environment configuration to be set at runtime *after* module loading has occurred.

evm/vm/forks/sharding/vmc_utils.py
```diff
@@ -15,7 +20,7 @@ def get_vmc_json():
     return json.loads(vmc_json_str)
 
 
-def create_vmc_tx(TransactionClass, gasprice=GASPRICE):
+def create_vmc_tx(TransactionClass, gasprice=GAS_PRICE):
```
pipermerriam: two things.

- Naming nitpick to rename to `gas_price` for naming consistency
- Maybe just make `gasprice` required (remove the default value) to avoid the need to have this module be aware of config values.

evm/utils/env.py
```diff
@@ -0,0 +1,301 @@
...
+        raise ValueError(
+            "`env_timestamp` requires either a default value to be specified, "
+            "or for the variable to be present in the environment"
+        )
+
+    timestamp = float(value)
+    return datetime.datetime.fromtimestamp(timestamp)
+
+
+def env_iso8601(name, required=False, default=empty):
```
pipermerriam: I'm skeptical that we'll use this utility.  What do you think about removing it (and the related tests).

mhchia: I think it's fine to remove it. I cannot figure out a scenario to make use of it.

evm/vm/forks/sharding/vmc_utils.py
mhchia: No problem. Will do.

evm/vm/forks/sharding/vmc_handler.py
mhchia: That's a great idea! I got you're point. In my code it was hard for two contract to use different config, since `sharding_config` is set when the module is executed, not when the VMC instance is created. 
Will modify it. Thank you:)

evm/utils/env.py
```diff
@@ -0,0 +1,256 @@
...
+    :type default: bool
+    """
+    value = get_env_value(name, required=required, default=default)
+    if value is empty:
+        return []
+    # wrapped in list to force evaluation in python 3
+    return list(filter(bool, [v.strip() for v in value.split(separator)]))
+
+
+def env_timestamp(name, required=False, default=empty):
```
pipermerriam: I glossed over this on the first one not realizing it returned `datetime.datetime` objects.  I think we could remove it as well for the same reasons as `iso8601`

mhchia: No problem.:)

mhchia:
No problem. Thank you!

PR #252, opened 2018-01-08
Add sharding transaction test

NIC619:
### How was it fixed?
Add sharing transaction test.

Steps are:
1. calculate contract address to be deployed
2. initiate a shard chain(without block validation) with the address derived above funded
3. deploy the contract
4. send transaction to the contract

I'm working on the contract code right now which is similar to the [user account contract](https://github.com/ethereum/py-evm/pull/230/files#diff-05178e170e27782a10a1498ec2aefdefR42) but simpler(parse the transaction data and transfer `amount` of ether to `recipient`).


hwwhww:
As the discussion in #238,  if we are going to create a `Shard` class which mirrors the `Chain` class and a `ShardVM` which mirrors the `VM` class, how about creating a new `test_shard_vm.py` file for these tests?

tests/core/vm/test_vm.py
```diff
@@ -46,3 +53,28 @@ def test_import_block(chain_without_block_validation):  # noqa: F811
...
     block = parent_vm.import_block(vm.block)
     assert block.transactions == [tx]
+
+
+def test_sharding_transaction(shard_chain_without_block_validation):  # noqa: F811
+    # Target contract, i.e. funded_address, is a simple ether transfer contract which transfer
+    # specified amount of ether to recipient.
+    # TODO: update contract_deployment_code
+    contract_deployment_code = b''
+    contract_addr = generate_create2_contract_address(b'', contract_deployment_code)
```
pipermerriam: What do you think about changing the name to `generate_CREATE2_contract_address` to make it a bit clearer that we're referring to the `CREATE2` opcode.

tests/core/fixtures.py
```diff
@@ -83,6 +84,57 @@ def shard_chain():
...
+    genesis_state = {
+        funded_addr: {
+            'balance': initial_balance,
+            'nonce': 0,
+            'code': b'',
+            'storage': {},
+        }
+    }
+    chain = klass.from_genesis(shard_chaindb, genesis_params, genesis_state)
+    chain.funded_address = funded_addr
```
pipermerriam: Instead of storing this on the `chain` object, can we just use the` funded_address` fixture in any place where we need access to this address.  I'm worried that setting a precedent for tacking values onto objects will lead to confusing tests down the road when it's hard for someone to differentiate what is a formal `Chain` API and what is just part of the test setup.

tests/core/vm/test_vm.py
```diff
@@ -46,3 +53,28 @@ def test_import_block(chain_without_block_validation):  # noqa: F811
...
+    assert not computation.is_error
+
+    # Transfer ether to recipient
+    recipient = decode_hex('0xa94f5374fce5edbc8e2a8697c15331677e6ebf0c')
+    amount = 100
+    tx_initiator = chain.funded_address
+    transfer_tx = new_sharding_transaction(tx_initiator, recipient, amount, b'', b'', b'')
+    computation = vm.execute_transaction(transfer_tx)
+    assert not computation.is_error
+    with vm.state.state_db(read_only=True) as state_db:
```
pipermerriam: Just had the thought that we could provide access to a read-only statedb without the context manager as part of the `state` object, and have the context manager API be only necessary for when write operations are necessary.

tests/core/vm/test_vm.py
NIC619: I will update it in CREATE2 PR.

tests/core/vm/test_vm.py
NIC619: That makes sense. Just to make sure, context manager's job here is to (1) double check no state were written in read_only mode and (2) decommission the state to prevent further modification to db, correct?
Does it involve managing(opening/closing/cleaning) the actual data base file?

jannikluhn:
The `simple_transfer_contract` is intended just for testing, correct? If so I'd not put it into `evm/auxiliary` but move it somewhere to the `tests` subdirectory.

NIC619:
@pipermerriam I made some updates and it is ready for review/merge, please have a look, thank you.

Updates:
1. Switch to NestedTrieBackend for shard chain testing for now to circumvent account data access restriction. Will update it once access list generating helper function is out.
2. Update contract code fixture for testing. The contract is a simple contract which takes in `destination`, `value`, each padded to 32 bytes, and transfer `value` amount of ether to `destination`. Steps are, firstly, fund the contract address in genesis, then deploy the contract and finally execute the transfer transaction.

NIC619:
@jannikluhn ahh you are right. I will move them to `tests` directory.

tests/core/fixtures.py
```diff
@@ -77,12 +83,70 @@ def chain(chaindb):
...
+
+    This Chain does not perform any validation when importing new blocks.
+
+    The Chain's state includes one funded account which is where the simple transfer
+
+    contract will be deployed at.
+    """
+    # TODO: Comment this out to use NestedTrieBackend to prevent access list validation.
+    # Uncomment this to use FlatTrieBackend once access list generation is implemented.
+    # shard_chaindb = BaseChainDB(get_db_backend(), state_backend_class=FlatTrieBackend)
```
pipermerriam: I'm not sure if this is what you want, but assuming you want to run this using *both* state backends you can do that in a very clean way as follows.

```python
@pytest.fixture(params=(FlatTrieBackend, NestedTrieBackend))
def shard_chain_without_block_validation(request):
    shard_chaindb = BaseChainDB(get_db_backend(), state_backend_class=request.param)
    ...
```

tests/core/fixtures.py
```diff
@@ -77,12 +83,70 @@ def chain(chaindb):
...
+        'difficulty': constants.GENESIS_DIFFICULTY,
+        'gas_limit': constants.GENESIS_GAS_LIMIT,
+        'parent_hash': constants.GENESIS_PARENT_HASH,
+        'coinbase': constants.GENESIS_COINBASE,
+        'nonce': constants.GENESIS_NONCE,
+        'mix_hash': constants.GENESIS_MIX_HASH,
+        'extra_data': constants.GENESIS_EXTRA_DATA,
+        'timestamp': 1501851927,
+        # 'state_root': decode_hex(
+        #     '0x9d354f9b5ba851a35eced279ef377111387197581429cfcc7f744ef89a30b5d4')
```
pipermerriam: Reason for this being commented out?  Can it be removed?

tests/core/vm/contract_fixture.py
```diff
@@ -0,0 +1,18 @@
...
+                                    ['assert', ['call', 0, ['calldataload', 0], ['calldataload', 32], 0, 0, 0, 0]],  # noqa: E501
+                                    'stop'],
+                            0]]]  # noqa: E128
+
+
+# compiled byte code
+contract_bytecode = b'0x61003e567401000000000000000000000000000000000000000060205260003560205181101558575060006000600060006020356000356000f1155857005b61000461003e0361000460003961000461003e036000f3'  # noqa: E501
+
+# address where this contract will be deployed
+contract_address = b'\xdb\xcd\xfc\xf2\xea!\xcd\x0c\xa5d\xaay\x8b;\xf0\xe4\xe2\x98S\xca'
```
pipermerriam: I assume this value is based on the contract bytecode?  If so, it seems like this value is likely to end up out-of-sync if the bytecode changes.  I'd suggest one of:

- make this a computed value.
- put a big **loud** comment here indicating how to compute this value and at the bytecode above indicating that if the bytecode is changed, then this value needs to be changed also.

tests/core/fixtures.py
NIC619: Sorry my comment was misleading. What I mean is to comment out FlatTrieBackend and use NestedTrieBackend instead for now. After the helper function which generates access list for a transaction is implemented, we can uncomment it and use FlatTrieBackend. I will update this part.

tests/core/fixtures.py
NIC619: You are right. This can be removed.

tests/core/vm/contract_fixture.py
NIC619: Will update it to a computed value.

PR #259, opened 2018-01-10
SIGHASH opcode

### LINKED ISSUE #226: Add SIGHASH opcode
### What is wrong?
Checking transaction signatures in account abstraction (#225) involves calculating the hash of the transaction, excluding the `data` field containing the signature itself. This is not possible right now, as there are no opcodes for all transaction fields.

### How can it be fixed

Add an opcode `SIGHASH` with byte value `0x3f` that gets `sha3(rlp(tx \ {data}))`, where `tx \ {data}` refers to the transaction without the data field.

jannikluhn:
Adds the SIGHASH opcode that puts the hash of the transaction excluding the data field on the stack.

Closes #226


evm/vm/message.py
```diff
@@ -178,6 +180,32 @@ def __init__(self,
...
+                              data,
+                              code,
+                              **kwargs):
+        kwargs.setdefault('sender', self.msg.storage_address)
+
+        child_message = ShardingMessage(
+            gas=gas,
+            gas_price=self.msg.gas_price,
+            origin=self.msg.origin,
+            sig_hash=self.msg.sig_hash,
```
pipermerriam: This has a "bad smell".

The `sighash` falls into the same category as the block data which indicates there should be a single location this value is stored.  Yet it is copied for each level of computation.

After some investigation, the same is true for `msg.gas_price` and `msg.origin`.  This isn't something we need to fix now, but I think it's worth investigating how we can remove this duplication. 

Issue opened here: https://github.com/ethereum/py-evm/issues/262

evm/logic/context.py
```diff
@@ -189,3 +189,7 @@ def returndatacopy(computation):
     value = computation.return_data[returndata_start_position: returndata_start_position + size]
 
     computation.memory.write(mem_start_position, size, value)
+
+
+def sighash(computation):
```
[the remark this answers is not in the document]
jannikluhn: There isn't, unfortunately.

evm/vm/message.py
jannikluhn: I agree. I must admit that I simply copied the `msg.origin` behavior.

Issue #242, opened 2017-12-27
Add `get_next_log` and `fetch_candidate_head`


mhchia:
### What is wrong?
Add `get_next_log` and `fetch_candidate_head` described in https://github.com/ethereum/sharding/blob/develop/docs/doc.md#fetch-candidate-heads-in-reverse-sorted-order

### How can it be fixed
Implement `getNextLog` with web3.py(essentially `LogFilter.get_new_entries` acts like that) and `fetch_candidate_head` in `MainchainHandler`.

mhchia:
Closed by #265

PR #265, opened 2018-01-12
Add `fetch_candidate_head`

mhchia:
### What was wrong?
As https://github.com/ethereum/py-evm/issues/242 described, `get_next_log` and `fetch_candidate_head` is needed.

### How was it fixed?
`get_next_log`: since there is a `get_new_entries()` in `web3.eth.filter`, which gives latest "unread" logs in the RPC server, when we fetch new logs from `get_new_entries()`, we are sure that they are latest and unseen. Besides, because `get_next_log` should only get **one** latest log, we have to save the other logs fetched altogether from `get_new_entries` to a stack, and treat them as "new" logs, waiting to be chosen by `get_new_log` in the future.  The way to maintain the stack(`VMC.new_collation_added_logs`) is, if `get_next_log` is called, first call `get_new_entries` and put the new logs on top of the stack(`VMC.new_collation_added_logs`). Then, return the log on the top, ensuring we get the latest and unseen log.

`fetch_candidate_head`: implement the same way as it is described in doc.md

#### TODOs
- [x] Use `eth_getLogs` over `filter.xxx`, maintaining new logs ourselves

#### Cute Animal Picture


mhchia:
Remark: The reason I currently use `raw_log` over `CollationAdded.log` is because `bytes` is only allowed to have at most 32 bytes in the log declaration. However the size of headers is far bigger than that.

tests/sharding/test_vmc_handler.py
```diff
@@ -304,13 +307,56 @@ def get_testing_colhdr(vmc_handler,
...
+        ),
+        (
+            (1, 2, 3, 2, 2, 2),
+            (True, True, True, False, False, False),
+            (3, 2, 2, 2, 2, 1),
+            (True, True, False, False, False, True),
+        ),
+    )
+)
+def test_vmc_fetch_candidate_head(vmc,
```
hwwhww: It's better to be aligned with https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#function-signatures

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -8,28 +14,146 @@
...
+        header_values = rlp.decode(header_bytes, sedes=sedes)
+        yield 'header', header_values
+        yield 'is_new_head', is_new_head
+        yield 'score', score
+
+    @to_tuple
+    def _get_new_logs(self, shard_id):
+        # use `get_new_entries` over `get_all_entries`
+        #   1. Prevent from the increasing size of logs
+        #   2. Leave the efforts maintaining `new_logs` in RPC servers
```
pipermerriam: I have a pattern I've been wanting to try out that this might be a good candidate for.  It's meant to emulate `get_new_entries` without any of the complication of installing the filter and tracking the `filter_id`.  It looks roughly like this.

```
from cytoolz import (
    merge,
    sliding_window,
)


def log_filter(web3, filter_params, page_size=10):
    # TODO: handle pending/latest/earliest
    from_block = filter_params['fromBlock']
    to_block = filter_params['toBlock']

    # TODO: handle dynamic block ranges to enforce reasonable request sizes.
    for block_number in range(from_block, to_block + 1):
        params = merge(filter_params, {'fromBlock': block_number, 'toBlock': block_number})
        logs = web3.eth.getLogs(params)

        for page in sliding_window(page_size, logs):
            yield page
```

Issue in web3.py here: https://github.com/ethereum/web3.py/issues/551

I suspect you could implement a specialized version of this that would be more robust than relying on filters on the server, eliminating the need to handle things like re-installing the filter if it gets lost.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -8,28 +14,146 @@
...
+        if shard_id not in self.collation_added_filter:
+            raise FilterNotFound(
+                "CollationAdded filter haven't been set up in shard {}".format(shard_id)
+            )
+        new_logs = self.collation_added_filter[shard_id].get_new_entries()
+        for log in new_logs:
+            yield pipe(
+                log['data'],
+                decode_hex,
+                self.parse_collation_added_data,
```
pipermerriam: Seems like moving the `decode_hex` into `parse_collation_added_data` would make this a lot easier to read.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -8,28 +14,146 @@
...
+        return self.new_collation_added_logs[shard_id].pop()
+
+    def fetch_candidate_head(self, shard_id):
+        # Try to return a log that has the score that we are checking for,
+        # checking in order of oldest to most recent.
+        if shard_id not in self.collation_added_filter:
+            raise FilterNotFound(
+                "CollationAdded filter haven't been set up in shard {}".format(shard_id)
+            )
+        for i in range(len(self.unchecked_collation_added_logs[shard_id]) - 1, -1, -1):
```
pipermerriam: I *think* this can be simplified with:

```
for log_entry in reversed(self.unchecked_collation_added_logs[shard_id]):
    if log_entry == ...:
        ...
```

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -8,28 +14,146 @@
...
+
+
 class VMC(Contract):
 
     logger = logging.getLogger("evm.chain.sharding.mainchain_handler.VMC")
 
+    # For handling logs filtering
+    # Event:
+    #   CollationAdded(indexed uint256 shard, bytes collationHeader, bool isNewHead, uint256 score)
+    collation_added_topic = "0x" + keccak(b"CollationAdded(int128,bytes4096,bool,int128)").hex()
```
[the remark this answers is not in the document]
mhchia: No problem. Will do.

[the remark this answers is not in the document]
mhchia: It's cool! will change to it.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -8,28 +14,146 @@
...
+                self.collation_added_topic,
+                shard_id_topic,
+            ],
+        })
+        self.new_collation_added_logs[shard_id] = []
+        self.unchecked_collation_added_logs[shard_id] = []
+        self.current_checking_score[shard_id] = None
+
+    @to_dict
+    def parse_collation_added_data(self, data):
```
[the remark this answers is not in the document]
mhchia: Sorry I don't quite understand it. Do you mean it should be moved outside the class? and can be tested in isolation?

evm/vm/forks/sharding/vmc_handler.py
mhchia: yep you're right. I have changed it. Thank you:)

evm/vm/forks/sharding/vmc_handler.py
mhchia: Thank you for pointing it out. Have changed it.

evm/vm/forks/sharding/vmc_handler.py
mhchia: No problem.
Thank you for letting me know of this. I didn't consider the situation that filters get lost in servers.
I will do it. :)

mhchia: Since I just found `getLogs` has not been implemented in eth_tester, I copied and pasted some code from `eth_tester.create_log_filter` and `eth_tester._process_block_logs` to a new function [`eth_tester.get_logs`](https://gist.github.com/mhchia/45e7fd2ada48918eb2bbb0c6a549017a), and made `web3.py` calls it. Now it seems to work fine.(still need testing).
One question is, for now I plan to save `block_number` every time I get new logs with `eth_getLogs`, to identify the latest block which our new logs have been synced to. Should I handle the situation when forks happen? If we take forks in mainchain into consideration, it seems to be more complicated, and we need to store more things.
Thank you for your wisdom:)

mhchia: @pipermerriam I'm sorry that after thinking for a while, I still have some questions about the code.
1. We should specify the `fromBlock` and `toBlock` to ensure the logs we are going to get are in the blocks range.
2. Why not just
```
    for block_number in range(from_block, to_block + 1):
        params = merge(filter_params, {'fromBlock': block_number, 'toBlock': block_number})
        logs = web3.eth.getLogs(params)
        for page in logs:
            yield page
```
instead of 
```
    for block_number in range(from_block, to_block + 1):
        params = merge(filter_params, {'fromBlock': block_number, 'toBlock': block_number})
        logs = web3.eth.getLogs(params)
        for page in sliding_window(page_size, logs):
            yield page
```
? What is purpose to use the `sliding_window`?
Thank you so much!

pipermerriam: I have no idea why I put `sliding_window` in there.  I think I got my wires crossed while hammering out that example code.  Will try to get you answers to other questions asap but I'm out of time today.

pipermerriam: Instead of sliding window, it should be [`toolz.partition_all`](http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition_all)

mhchia: hmm that makes sense.
Thank you so much.
Please take your time. It is no hurry at all!

pipermerriam: > I have one question is, for now I plan to save `block_number` every time I get new logs with `eth_getLogs`, to identify the latest block which our new logs have been synced to. Should I handle the situation when forks happen? If we take forks in mainchain into consideration, it seems to be more complicated, and we need to store more things.

Here is some untested code that won't actually work but should demonstrate the idea:

```python
def get_previous_headers(w3, history_size=256):
    block = w3.eth.getBlock('latest')
    recent_hashes = [block['hash']]

    # initialize the list of recent hashes
    for _ in range(history_size - 1):
        block = w3.eth.getBlock(block['parentHash'])
        recent_hashes.append(block['hash'])

        # break the loop if we hit the genesis block.
        if block['number'] == 0:
            break

    recent_hashes = tuple(reversed(_recent_hashes))
    return recent_hashes


def check_chain_head(w3, known_previous_headers):
    block = w3.eth.getBlock('latest')

    new_blocks = []

    for _ in range(history_size):
        if block['hash'] in recent_hashes:
            break
        new_block_hashs.append(block['hash'])
        block = w3.eth.getBlock(block['parentHash'])
    else:
        raise Exception('No common ancestor found for block: {0}'.format(block['hash']))

    first_common_ancestor_idx = recent_hashes.index(block['hash'])

    unchanged_hashes = recent_hashes[:first_common_ancestor_idx]
    revoked_hashes = recent_hashes[first_common_ancestor_idx:]

    recent_hashes.extend(new_block_hashs)
    # this won't actually work because of variable scoping
    recent_hashs = recent_hashs[-1 * history_size:]
    return revoked_hashes, new_block_hashs


```

The idea is to build a utility that will track the most recent hashes in the head of the chain in such a way that it can provide you with two values.

- the block hashes that were revoked from the head of the chain due to re-org
- the new block hashes that we haven't yet seen

You can use the revoked hashes to roll back any logs that you processed which are no longer valid (assuming you need to do that).

You can use the new hashes to define the `fromBlock` and `toBlock` for fetching any new logs that you haven't yet seen.

Questions?  Does this make sense?  The code above could probably more easily be written as a class.

mhchia: It's a beautiful and simple idea! In this way we can focus on the `revoked_changes` and `unseen_and_canonical_changes`, and no need to worry about the other re-orgs between them. I think I can soon implement the utility based on your code.
Still one question: Why is `history_size` set to `256`? Should it be larger or smaller?
I really appreciate your help and patience!

pipermerriam: `256` was a totally arbitrary number.  Should probably be configurable.

mhchia:
@pipermerriam I implemented your algorithm, and made a PR for eth_getLogs https://github.com/ethereum/eth-tester/pull/50
Could you help me to see whether there is anything wrong if you have spare time?
If there is no issue in that PR, I will make a PR in web3.py to enable `eth_getLogs` for eth_tester in web3.py
Thank you!

pipermerriam:
👍 will get your `eth-tester` pr reviewed and merged as soon as the travis-ci log-jam clears.

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+        self.recent_block_hashes = unchanged_hashes + reversed_new_block_hashes
+        # keep len(self.recent_block_hashes) <= self.history_size
+        self.recent_block_hashes = self.recent_block_hashes[(-1 * self.history_size):]
+        return revoked_hashes, reversed_new_block_hashes
+
+    def preprocess_block_param(self, block_param):
+        if isinstance(block_param, int):
+            return block_param
+        elif not isinstance(block_param, str):
+            raise ValueError("block parameter {0} is in wrong type".format(block_param))
```
pipermerriam: Instead of saying, *"wrong type"* it would likely make for a more helpful error message to invert this and say *"must be `str` type"*

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+        # reverse it to comply with the order of `self.recent_block_hashes`
+        reversed_new_block_hashes = tuple(reversed(new_block_hashes))
+        # append new blocks to `unchanged_hashes`, and move revoked ones out of
+        # `self.recent_block_hashes`
+        self.recent_block_hashes = unchanged_hashes + reversed_new_block_hashes
+        # keep len(self.recent_block_hashes) <= self.history_size
+        self.recent_block_hashes = self.recent_block_hashes[(-1 * self.history_size):]
+        return revoked_hashes, reversed_new_block_hashes
+
+    def preprocess_block_param(self, block_param):
```
pipermerriam: This also would make for a good stand-alone function.  Should be pretty easy to test all of the logic branches that way.

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+    @to_dict
+    def mk_filter_params(self, from_block_number, to_block_number, address=None, topics=None):
+        yield 'fromBlock', self.preprocess_block_param(from_block_number)
+        yield 'toBlock', self.preprocess_block_param(to_block_number)
+        if address is not None:
+            yield 'address', address
+        if topics is not None:
+            yield 'topics', topics
+
+    def filter_logs(self, filter_params):
```
pipermerriam: Given the direct passthrough nature of the method, can we just remove it an call into `self.w3` directly at any places which use `self.filter_logs(...)`

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+        logs = self.w3.eth.getLogs(filter_params)
+        return logs
+
+    def get_new_logs(self, address=None, topics=None):
+        _, new_block_hashes = self.check_chain_head()
```
pipermerriam: A continuation of my previous comment suggesting that most of the `check_chain_head` functionality be moved to stand-alone functions.  Here, is the place where the example code from above would go, after which you can update whatever local variable you're using to store the most recent hashes.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+
+    # For handling logs filtering
+    # Event:
+    #   CollationAdded(indexed uint256 shard, bytes collationHeader, bool isNewHead, uint256 score)
+    COLLATION_ADDED_TOPIC = event_signature_to_log_topic(
+        "CollationAdded(int128,bytes4096,bool,int128)"
+    )
 
-    def __init__(self, *args, default_privkey, **kwargs):
+    shards = set()
```
pipermerriam: You shouldn't use mutable objects as class properties.  This should be set in the constructor as otherwise, all instances of this class will share this variable meaning if it's changed in one, it'll be changed in all of them.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+    # shard_id -> list
+    # older <---------------> newer
+    new_collation_added_logs = {}
```
pipermerriam: Same here about mutable data structures and setting this in the `__init__`

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+    # shard_id -> list
+    # newer <---------------> older
+    unchecked_collation_added_logs = {}
+    # shard_id -> score
+    current_checking_score = {}
```
pipermerriam: Same here about mutable data structures and setting this in the `__init__`

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+    )
 
-    def __init__(self, *args, default_privkey, **kwargs):
+    shards = set()
+    # shard_id -> list
+    # older <---------------> newer
+    new_collation_added_logs = {}
+    # shard_id -> list
+    # newer <---------------> older
+    unchecked_collation_added_logs = {}
```
pipermerriam: Same here about mutable data structures and setting this in the `__init__`

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
         self.default_privkey = default_privkey
         self.default_sender_address = default_privkey.public_key.to_canonical_address()
         self.config = get_sharding_config()
         super().__init__(*args, **kwargs)
 
+    def init_shard_variables(self, shard_id):
+        self.shards.add(shard_id)
+        self.unchecked_collation_added_logs[shard_id] = []
+        self.new_collation_added_logs[shard_id] = []
+        self.current_checking_score[shard_id] = None
```
pipermerriam: I think that you can remove all of this initialization if you use `collection.defaultdict` instances for all of these three data structures.

```python
self.unchecked_collation_added_logs = collections.defaultdict(list)
self.new_collation_added_logs = collections.defaultdict(list)
self.current_checking_score = {}
```

Then you can access both `self.unchecked_collation_added_logs[shard_id]` and `self.new_collation_added_logs[shard_id]` as if they have already been initialized thanks to the magic of defaultdicts.

This should allow you to remove all of the initialization and `ensure_` initialization code.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+                encode_hex(self.COLLATION_ADDED_TOPIC),
+                shard_id_topic_hex,
+            ],
+        )
+        for log in new_logs:
+            yield parse_collation_added_data(log['data'])
+
+    def get_next_log(self, shard_id):
+        new_logs = self._get_new_logs(shard_id)
+        self.new_collation_added_logs[shard_id] += new_logs
```
pipermerriam: I think `.extend(new_logs)` is the *more* canonical way to do this.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+        if len(self.new_collation_added_logs[shard_id]) == 0:
+            raise NextLogUnavailable("No more next logs")
+        return self.new_collation_added_logs[shard_id].pop()
+
+    def fetch_candidate_head(self, shard_id):
+        # Try to return a log that has the score that we are checking for,
+        # checking in order of oldest to most recent.
+        self.ensure_shard_variables_initialied(shard_id)
+
+        for i in reversed(range(len(self.unchecked_collation_added_logs[shard_id]))):
```
pipermerriam: I think this is more readable as:

```python
# tuple(..) call is to force the iterator to fully evaluate so that the later `.pop(..)` doesn't cause problems.
unchecked_logs = tuple(reversed(enumerate(self.unchecked_collation_added_logs[shard_id])))
current_score = self.current_checking_score[shard_id]

for idx, log_entry in unchecked_logs:
    if log_entry['score'] == current_score:
        return self.unchecked_collation_added_logs[shard_id].pop(idx)
```

It's really just moving things around, but it removes one index based lookup, as well as storing the current score in a local variable to make the inner loop more legible and easier to follow.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,145 @@
...
+    def __init__(self, *args, log_handler, default_privkey, **kwargs):
+        self.log_handler = log_handler
         self.default_privkey = default_privkey
         self.default_sender_address = default_privkey.public_key.to_canonical_address()
         self.config = get_sharding_config()
         super().__init__(*args, **kwargs)
 
+    def init_shard_variables(self, shard_id):
+        self.shards.add(shard_id)
+        self.unchecked_collation_added_logs[shard_id] = []
```
pipermerriam: What do you think about renaming this variable to `self.unchecked_logs` and `self.new_logs` and `current_score`.  Should make a lot of the lines in this class way shorter and easier to read.

evm/vm/forks/sharding/vmc_handler.py
mhchia: Sorry it's my fault. Thank you for pointing out. Will change them.

evm/vm/forks/sharding/vmc_handler.py
mhchia: Will do. Actually the reason why I was naming this long name is I thought there would be multiple types of logs to be handled, so I specified the name.
But since they are not handled yet, I think it's a good idea to renaming them to shorter ones for now. Thank you.

evm/vm/forks/sharding/vmc_handler.py
mhchia: It's a good idea. Thank you for pointing out. I will change it.

evm/vm/forks/sharding/vmc_handler.py
mhchia: Thank you for the suggestion! It is way more readable. Will change it.

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+            return block_param
+        elif not isinstance(block_param, str):
+            raise ValueError("block parameter {0} is in wrong type".format(block_param))
+        current_block_number = self.w3.eth.blockNumber
+        mapping = {
+            'earliest': 0,
+            'latest': current_block_number,
+            'pending': current_block_number + 1,
+        }
+        return mapping.get(block_param, block_param)
```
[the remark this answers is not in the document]
mhchia: Ok no problem:) Will fix it.

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+        for _ in range(self.history_size - 1):
+            # break the loop if we hit the genesis block.
+            if block['number'] == 0:
+                break
+            block = self.w3.eth.getBlock(block['parentHash'])
+            recent_hashes.append(block['hash'])
+        reversed_recent_hashes = tuple(reversed(recent_hashes))
+        return reversed_recent_hashes
+
+    def check_chain_head(self):
```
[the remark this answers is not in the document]
mhchia: Have done it. Thank you for the wise instruction.
May I ask how can I determine which functions should stay in the class, while the others should be put outside the class `LogHandler`? Is the reason that `get_new_logs` stays in the class because it is the only interface used by `VMCHandler`?
Is the reason `check_chain_head`, `get_recent_block_hashes`, `preprocess_block_param` can be put outside the class `LogHandler` because they only depend on `web3` and can be used directly without the logics of logs?
Sorry for asking this newbie question. Just want to know what is the better criteria to design the class.

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,99 @@
...
+
+    logger = logging.getLogger("evm.chain.sharding.LogHandler")
+
+    def __init__(self, w3, history_size=256):
+        self.history_size = history_size
+        self.w3 = w3
+        # ----------> higher score
+        self.recent_block_hashes = self.get_recent_block_hashes()
+
+    def get_recent_block_hashes(self):
```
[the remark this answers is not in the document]
hwwhww: There's the `VM. get_prev_hashes(last_block_hash, db)` function for the main chain client to build the *previous-256-headers-hashes-list*:

https://github.com/ethereum/py-evm/blob/10d29c4a3e189de85ec3c96e949bf8ebca52c9de/evm/vm/base.py#L418-L431

evm/vm/forks/sharding/log_handler.py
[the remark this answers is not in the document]
pipermerriam: No need to apologize and I'm all for asking questions of any kind.

My general rule of thumb is the following loose collection of axioms and guidelines.

- classes are typically stateful and statefullness is hard to manage.
- functions (especially pure ones) are easier to reason about than classes.
- a class with only one function is just a function disguised as a class.  This also regularly applies to classes with only 2-3 functions.
- a method that doesn't use `self` should almost always be a function.
- it is *ok* to include convenience functions using `@staticmethod`, but normally should only be done if the convenience is *external*, meaning that outside actors who use this class almost always need the convenience function.  If it's only needed internally, it should just be used as a stand-alone function.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,147 @@
...
+    def __init__(self, *args, log_handler, default_privkey, **kwargs):
+        self.log_handler = log_handler
         self.default_privkey = default_privkey
         self.default_sender_address = default_privkey.public_key.to_canonical_address()
         self.config = get_sharding_config()
+        # shard_id -> list
+        # older <---------------> newer
+        self.new_logs = defaultdict(list)
+        # shard_id -> list
+        # newer <---------------> older
```
pipermerriam: I think this isn't accurate.

- Lets say we get logs `1, 2, 3, 4, 5`
- Log 1 is the oldest
- Log 5 is the newest
- They come in as two batches, `1, 2, 3` and then when we query again after some blocks, `4, 5`

In this case, I think that `unchecked_logs` will be `3, 2, 1, 5, 4`.  Maybe I'm reading the code wrong...?

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,147 @@
...
+        current_score = self.current_score[shard_id]
+
+        for idx, logs_entry in unchecked_logs:
+            if logs_entry['score'] == current_score:
+                return self.unchecked_logs[shard_id].pop(idx)
+        # If no further recorded but unchecked logs exist, go to the next
+        # is_new_head = true log
+        while True:
+            # TODO: currently just raise when there is no log anymore
+            self.unchecked_logs[shard_id].append(self.get_next_log(shard_id))
```
pipermerriam: Related to my comment about the logs potentially not being maintained in the right order, I think this is the offending line.  It seems like the logs should be pushed onto the front of the list rather than apppended to the end.  If this is the case, then you might consider using a `collections.deque` which has fast inserts and pops to both the beginning and end of the list.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,147 @@
...
+    # For handling logs filtering
+    # Event:
+    #   CollationAdded(indexed uint256 shard, bytes collationHeader, bool isNewHead, uint256 score)
+    COLLATION_ADDED_TOPIC = event_signature_to_log_topic(
+        "CollationAdded(int128,bytes4096,bool,int128)"
+    )
 
-    def __init__(self, *args, default_privkey, **kwargs):
+    def __init__(self, *args, log_handler, default_privkey, **kwargs):
+        self.log_handler = log_handler
```
pipermerriam: I think you're going to need one log handler per shard.  Since the log handler is stateful, each time you ask for more logs, it only returns the ones for the given `shard_id`.  If you're tracking multiple `shard_id`s then I think you're going to miss logs.

evm/vm/forks/sharding/vmc_handler.py
mhchia: I think you're right. It is not accurate. I will remove this comment.

evm/vm/forks/sharding/vmc_handler.py
mhchia: Sorry about the late reply. I think I need to think for a while of it tomorrow.
Will soon give an explanation and what to revise. Thank you for pointing these out!

evm/vm/forks/sharding/log_handler.py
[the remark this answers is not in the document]
mhchia: Thank you:)

evm/vm/forks/sharding/log_handler.py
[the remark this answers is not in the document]
mhchia: Thank you so much for giving me these guidelines! They are really helpful.
Just want to conclude,
is it correct that: In general, we should use functions over classes(as pointed in https://github.com/pipermerriam/ethereum-dev-tactical-manual#the-zen-of-the-python-ethereum-team). And it seems the right timing to choose a class over functions is when we have to maintain states besides a set of functions. (and should move functions which is not necessarily resided inside the class to stand-alone functions if possible?)

evm/vm/forks/sharding/vmc_handler.py
mhchia: Ahh I didn't thought of this case. Thank you for discovering that.
Will fix it.

mhchia: Actually I was thinking that we could use maintain one `recent_block_hashes` for all shards, to share it over shards. But in this way we will need to store all logs from all shards. It seems not better than keeping one log_handler for each shard.

evm/vm/forks/sharding/vmc_handler.py
mhchia: For the algorithm `fetch_candidate_head`, it is from https://github.com/ethereum/sharding/blob/develop/docs/doc.md#fetch-candidate-heads-in-reverse-sorted-order .
It's idea is
> The idea is that this algorithm is guaranteed to check potential head candidates in highest-to-lowest sorted order of score, with the second priority being oldest to most recent.

I am thinking one example, which follows what you assume above.
Assume we have `batch1 = [1, 2, 3]` and `batch2 = [4, 5]`.
First, `batch1` arrives and we have `new_logs = [1, 2, 3]`.
Before `batch2` arrives, the outcome of `fetch_candidate_head` will be `3`, `2`, `1`. Anytime as long as `batch2` arrives, the result of `fetch_candidate_head` would be the  element in `batch2`, instead of the one in `batch1`.
So I think for this case the algorithm is right.

However, if we assume the `batch1 = [1, 1]` and `batch2 = [2]`. As long as batch2 arrives after `batch1` and before `fetch_candidate_head` is called, the order of the outcome would be `1, 1, 2`, which seems not consistent with the idea of the algorithm "The idea is that this algorithm is guaranteed to check potential head candidates in highest-to-lowest sorted order of score, with the second priority being oldest to most recent.".
Example code is in [here](https://gist.github.com/mhchia/dfa9961de7c3fd59f08afaf97552883a)
About this I think I should ask @vbuterin for its detail.

vbuterin: The intended principle is that you want to get logs in reverse logical order, so latest block to oldest block, and within a block last transaction to first transaction, and within a transaction last to first created.

evm/vm/forks/sharding/vmc_handler.py
pipermerriam: Yeah, one handler per shard seems cleanest despite the added overhead.  We can address the performance issues later if needed.

mhchia:
Note: Integration tests failed. Will look into it tomorrow.

mhchia:
@pipermerriam 
After discussed with @vbuterin , it seems `fetch_candidate_head` is fine to work in that way, if I didn't understand wrong. Since a node calls `fecth_candidate_head` inside `GUESS_HEAD` after it find itself is a **collator** in some period. At that time, the node can save the current status of `new_logs`, `unchecked_logs`, and `current_score`. If there is a new block arriving, just restore them to the saved status, append new logs to `new_logs`, and re-run `GUESS_HEAD`.
Is this scheme acceptable to you? and do you have any suggestion on this?

Thank you!

pipermerriam:
I'm good with whatever you guys have settled on regarding the proper functionality.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,34 +1,161 @@
...
+        self.current_score = {}
+
         super().__init__(*args, **kwargs)
 
+    def init_log_variables(self, shard_id):
+        self.new_logs[shard_id] = []
+        # shard_id -> list
+        self.unchecked_logs[shard_id] = []
+        # shard_id -> score
+        self.current_score[shard_id] = None
```
pipermerriam: I think that this entire class would be cleaned up by implementing one more data structure to house all of the shard data.

```python
class ShardTracker:
    shard_id = ...
    log_handler = ...
    new_logs = ...
    unchecked_logs = ...
    current_score = ...
```

I suspect that some, if not most of the business logic for updating the shard state could then be moved into that object at which point this class would just manage all of the individual instances of that class.

Thoughts?  I think it would clean this class up in a significant way.

mhchia: I think that is a good idea. I will do it.
Just two questions about it.
1. Should I put `ShardTracker` in another file `shard_tracker.py`?
2. Since all logics regarding shards logs are moved to `ShardTracker`(including `get_next_log`, `fetch_candidate_head`), if there is a need for an API in `VMCHandler` like `fetch_candidate_head`, may I just set a same name method `fetch_candidate_head` in `VMCHandler` and simply calls `ShardTracker.fetch_candidate_head`? (Though I think there might only be new methods like `guess_head` in VMC which will be implemented in the future)
Very much appreciate!

mhchia: Moreover, may I request a bumpversion in `web3.py`, to resolve the error in CI?
Thank you!

pipermerriam: 1. It's probably fine to start with it here in this file or to put it in it's own module.  Up to you.  `shard_tracker.py` seems like a fine choice as well.

2. If this makes `VMCHandler` obsolete and you think you're able to do what you need with only the `ShardTracker` class, then it's probably a good idea to do away with it.  Really depends on how these APIs will be used and whether you need an abstraction to track *many* shards and whether it still makes sense for that to be a class.

pipermerriam:
I believe you can bump the web3 version locally here in your branch to fix CI

evm/vm/forks/sharding/vmc_handler.py
mhchia: @pipermerriam Thank you for the great suggestions! I have moved those logics to `ShardTracker`, currently in the same module `vmc_handler` as `VMCHandler`.
For now I just make `get_next_log`, `fetch_candidate_head` directly calls the corresponding `shard_tracker`. I think I can determine what to remove and how to restructure between `ShardTracker` and `VMCHandler` after `GUESS_HEAD` and `CREATE_COLLATION` are implemented, which will be done in the next stage.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -118,44 +103,79 @@ def _get_new_logs(self, shard_id):
...
-                return self.unchecked_logs[shard_id].pop(idx)
+                return self.unchecked_logs.pop(idx)
         # If no further recorded but unchecked logs exist, go to the next
         # is_new_head = true log
         while True:
             # TODO: currently just raise when there is no log anymore
-            self.unchecked_logs[shard_id].append(self.get_next_log(shard_id))
-            if self.unchecked_logs[shard_id][-1]['is_new_head'] is True:
+            self.unchecked_logs.append(self.get_next_log())
+            if self.unchecked_logs[-1]['is_new_head'] is True:
```
pipermerriam: If i'm reading this logic right.

- retrieve the new log with `get_next_log()`
- append to `self.unchecked_logs`
- if newest log `is_new_head == True`
    - break from loop and pop off newest log.

I think this is simpler written as:

```python
while True:
    log_entry = self.get_new_log()
    if log_entry['is_new_head']:
        break
    else:
        self.unchecked_logs.append(log_entry)

...
```

It avoids appending and then popping by only appending in the case that the new log isn't the new head.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -118,44 +103,79 @@ def _get_new_logs(self, shard_id):
...
+        self.shard_trackers[shard_id] = shard_tracker
+
+    def get_shard_tracker(self, shard_id):
+        if shard_id not in self.shard_trackers:
+            raise ShardNotTracked('Shard {} is not tracked'.format(shard_id))
+        return self.shard_trackers[shard_id]
+
+    # TODO: currently just calls `shard_tracker.get_next_log`
+    def get_next_log(self, shard_id):
+        if shard_id not in self.shard_trackers:
```
pipermerriam: Instead of repeating this logic, I would suggest using `self.get_shard_tracker(shard_id)` here to retrieve the tracker.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -118,44 +103,79 @@ def _get_new_logs(self, shard_id):
...
+            raise ShardNotTracked('Shard {} is not tracked'.format(shard_id))
+        shard_tracker = self.shard_trackers[shard_id]
+        return shard_tracker.get_next_log()
+
+    # TODO: currently just calls `shard_tracker.fetch_candidate_head`
+    def fetch_candidate_head(self, shard_id):
+        if shard_id not in self.shard_trackers:
```
pipermerriam: Same here about removing this logic in favor of using `self.get_shard_tracker(shard_id)`

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,117 @@
...
+    def mk_filter_params(self, from_block_number, to_block_number, address=None, topics=None):
+        yield 'fromBlock', preprocess_block_param(self.w3, from_block_number)
+        yield 'toBlock', preprocess_block_param(self.w3, to_block_number)
+        if address is not None:
+            yield 'address', address
+        if topics is not None:
+            yield 'topics', topics
+
+    def get_new_logs(self, address=None, topics=None):
+        revoked_hashes, new_block_hashes = check_chain_head(
```
pipermerriam: I may have already asked this question.  Does something need to be done with the revoked hashes as well?  Should the tracker be doing anything with these to revert anything in the event that one of it's log entries for which is updated it's state is revoked?

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -40,6 +40,10 @@ class NextLogUnavailable(Exception):
     pass
 
 
+class ShardNotTracked(Exception):
```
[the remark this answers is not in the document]
mhchia: It's better, Thank you:)

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -65,51 +69,32 @@ def parse_collation_added_data(data_hex):
...
-        # shard_id -> list
-        self.unchecked_logs[shard_id] = []
-        # shard_id -> score
-        self.current_score[shard_id] = None
-
-    def setup_log_handler(self, log_handler, shard_id):
-        self.log_handlers[shard_id] = log_handler
-        self.init_log_variables(shard_id)
+        self.new_logs = []
+        self.unchecked_logs = []
```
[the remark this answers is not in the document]
mhchia: I think they are in constructor already?
I think setting them as `None` is a good idea.
Is the purpose of doing this is to explicitly show their existence?

evm/vm/forks/sharding/vmc_handler.py
mhchia: Thank you for pointing out.
Will change it.

evm/vm/forks/sharding/vmc_handler.py
mhchia: Will change it. Thank you:)

evm/vm/forks/sharding/vmc_handler.py
mhchia: you're right. Thank you for pointing it out.
I've modified the code.

evm/vm/forks/sharding/vmc_handler.py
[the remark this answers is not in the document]
pipermerriam: Ignore my comment.  I didn't see that this was within `__init__.py` (thought I was looking at the class body).

The reason for setting them to `None` on the class is entirely convention to improve readability.  I find it helpful to be able to look at the class definition and get a full view of all of the various properties the class has.  It's harder to gather that information if you have to dig it out of code, even simple code like this in the constructor.

evm/vm/forks/sharding/log_handler.py
mhchia: Actually I'm not sure about this. Please let me think about it. Sorry about this.
IMO now, If we just want to mimic the behavior of the version using `filter.get_new_entries`, I think we don't need to do the revoke.
Do you have any suggestion?

pipermerriam: Unfortunately, I'm not familiar enough with the underlying algorithm to be able to provide a suggestion on what the *right* thing to do here is.  Just noticed that the revoked data wasn't used and wanted be sure that wasn't an oversight.  Feel free to mark it with a `TODO` comment and open an issue to track it and we can merge with the understanding that we'll need to address it down the road.

hwwhww:
Mind writing a brief document in the PR description?

tests/sharding/fixtures.py
```diff
@@ -25,11 +30,6 @@
     get_default_account_keys,
 )
 
-from eth_utils import (
-    to_canonical_address,
-    to_checksum_address,
-)
-
 from evm.vm.forks.spurious_dragon.transactions import (
     SpuriousDragonTransaction,
```
hwwhww: Just curious, why use `SpuriousDragonTransaction` instead of `ByzantiumTransaction`?

evm/vm/forks/sharding/log_handler.py
```diff
@@ -0,0 +1,117 @@
...
+            )
+        )
+
+
+class LogHandler:
+
+    logger = logging.getLogger("evm.chain.sharding.LogHandler")
+
+    def __init__(self, w3, history_size=256):
+        self.history_size = history_size
```
hwwhww: For `LogHandler` or higher levels (`ShardTracker`, `VMC`) APIs, is the `LogHandler.history_size` a *fixed number* of previous block hashes/headers capacity from the head block and *it is only set in initialization*?
If it's true, we may need a more flexible method to enable us to query for *older* block hashes (more than previous 256 block headers) *dynamically* during `guess_head` algorithm. The reason is that we want to verify as many collations as we can.

tests/sharding/fixtures.py
mhchia: Thank you for pointing out!
I will change it.

evm/vm/forks/sharding/log_handler.py
mhchia: That's fine. Thank you so much for keeping update with this part.
Yeah you're right it is worth keeping notice of this. I marked it with `TODO` and opened the issue https://github.com/ethereum/py-evm/issues/344.

PR #269, opened 2018-01-14
Add create2

NIC619:
### How was it fixed?
Add CREATE2 opcode
- [x] update intrinsic gas calculation of sharing transaction
- [x] add CREATE2 address check in `execute_transaction` to make sure `transaction.to` is the same as computed contract address
- [x] add CREATE2 opcode and remove CREATE from sharding opcodes
- [x] test


hwwhww:
link #224

evm/vm/forks/sharding/transactions.py
```diff
@@ -43,5 +44,6 @@ def _get_sharding_intrinsic_gas(transaction_data, transaction_code):
     return (
         GAS_TX +
         num_zero_bytes * GAS_TXDATAZERO +
-        num_non_zero_bytes * GAS_TXDATANONZERO
+        num_non_zero_bytes * GAS_TXDATANONZERO +
+        GAS_TXCREATE
```
hwwhww: I may miss something, could you tell me why `GAS_TXCREATE` is included in every sharding tx?

evm/logic/system.py
```diff
@@ -190,3 +191,69 @@ def __call__(self, computation):
         if computation.msg.is_static:
             raise WriteProtection("Cannot modify state while inside of a STATICCALL context")
         return super(CreateEIP150, self).__call__(computation)
+
+
+class Create2(CreateEIP150):
```
hwwhww: Is there's a reason for inheriting `CreateEIP150` instead of `CreateByzantium`?

NIC619: `CreateByzantium` and `Create2` both override `__call__` function so I make `Create2` inherit from `CreateEIP150` and copy the `is_static` check from `CreateByzantium`.

evm/vm/forks/sharding/transactions.py
NIC619: You are right. It should not be included in every tx.

tests/core/vm/test_sharding_vm.py
```diff
@@ -7,26 +12,64 @@
...
-    contract_address,
+    simple_transfer_contract_bytecode,
+    simple_transfer_contract_address,
+    simple_contract_factory_bytecode,
+    CREATE2_contract_bytecode,
+    CREATE2_contract_address,
 )
 
 
 def test_sharding_transaction(shard_chain_without_block_validation):  # noqa: F811
```
hwwhww: Since it's a test in `test_vm_state.py`, mind updating the test name to `test_sharding_apply_transaction`?

NIC619: Sure. I will update the name. Do you mean also moving it to `test_vm_state.py`?

hwwhww: ohhhhh sorry, I meant `test_sharding_vm.py`!

NIC619:
Tests updated. Thanks.

evm/vm/forks/sharding/opcodes.py
```diff
@@ -18,10 +20,18 @@
...
 }
 
 
 SHARDING_OPCODES = merge(
     copy.deepcopy(BYZANTIUM_OPCODES),
     NEW_OPCODES
 )
+
+
+SHARDING_OPCODES.pop(opcode_values.CREATE)
```
pipermerriam: Instead of popping this value out, can we do the following a few lines up.

```python
from cytoolz import dissoc

SHARDING_OPCODES = merge(
     dissoc(copy.deepcopy(BYZANTIUM_OPCODES), opcode_values.CREATE),
     NEW_OPCODES
 )
```

tests/core/vm/test_sharding_vm.py
```diff
@@ -7,26 +12,117 @@
...
+    gas_used = vm.block.header.gas_used - last_gas_used
+    assert gas_used > transfer_tx.intrinsic_gas
+    last_gas_used = vm.block.header.gas_used
     with vm.state.state_db(read_only=True) as state_db:
         assert state_db.get_balance(recipient) == amount
+
+    # Second test: contract that deploy new contract with CREATE2
+    second_deploy_tx = new_sharding_transaction(
+        CREATE2_contract_address,
+        b'',
```
pipermerriam: Can you change this to use keyword arguments so that it's clear what all of these similar parameters are.

tests/core/vm/test_sharding_vm.py
```diff
@@ -7,26 +12,117 @@
...
+        assert state_db.get_code(newly_deployed_contract_address) == b'\xbe\xef'
+        assert state_db.get_storage(CREATE2_contract_address, 0) == 1
+
+
+def test_CREATE2_deploy_contract_edge_cases(shard_chain_without_block_validation):  # noqa: F811
+    # First case: computed contract address not the same as provided in `transaction.to`
+    chain = shard_chain_without_block_validation
+    first_failed_deploy_tx = new_sharding_transaction(
+        simple_transfer_contract_address,
+        b'',
```
pipermerriam: Same request here regarding keyword arguments.

tests/core/vm/test_sharding_vm.py
```diff
@@ -7,26 +12,117 @@
...
+    computation = vm.apply_transaction(first_failed_deploy_tx)
+    assert computation.is_error
+    gas_used = vm.block.header.gas_used
+    assert gas_used > first_failed_deploy_tx.intrinsic_gas
+    last_gas_used = gas_used
+
+    # Next, complete deploying the contract
+    successful_deploy_tx = new_sharding_transaction(
+        simple_transfer_contract_address,
+        b'',
```
pipermerriam: and here

tests/core/vm/test_sharding_vm.py
```diff
@@ -7,26 +12,117 @@
...
+    computation = vm.apply_transaction(successful_deploy_tx)
+    assert not computation.is_error
+    gas_used = vm.block.header.gas_used - last_gas_used
+    assert gas_used > successful_deploy_tx.intrinsic_gas
+    last_gas_used = gas_used
+
+    # Second case: deploy to existing account
+    second_failed_deploy_tx = successful_deploy_tx
+    computation = vm.apply_transaction(second_failed_deploy_tx)
+    assert computation.is_error
```
pipermerriam: Maybe add a check that `isinstance(computation._error, ContractCreationCollision)`

tests/core/vm/test_sharding_vm.py
```diff
@@ -7,26 +12,117 @@
...
+        b'',
+        0,
+        b'',
+        b'',
+        '0xf3',
+    )
+
+    vm = chain.get_vm()
+    computation = vm.apply_transaction(first_failed_deploy_tx)
+    assert computation.is_error
```
pipermerriam: Maybe add a check for `isinstance(computation._error, IncorrectContractCreationAddress)`

PR #272, opened 2018-01-15
p2p: Implement ETH PV63 sub-protocol commands

gsalgado:


evm/p2p/eth.py
```diff
@@ -0,0 +1,206 @@
...
+from rlp import sedes
+
+from eth_utils import encode_hex
+
+from .constants import MAX_HEADERS_FETCH
+from evm.rlp.headers import BlockHeader
+from evm.rlp.receipts import Receipt
+from evm.rlp.transactions import BaseTransaction
+from evm.p2p.exceptions import HandshakeFailure
+from evm.p2p.les import HashOrNumber
```
pipermerriam: Seems like this should get moved to a separate `sedes` file rather than it living in `les` now that it's being used in multiple places.

evm/p2p/peer.py
```diff
@@ -449,6 +452,23 @@ def les_proto(self):
         return self._les_proto
 
 
+class ETHPeer(BasePeer):
+    _eth_proto = None
+    _supported_sub_protocols = [ETHProtocol]
+
+    @property
+    def eth_proto(self):
+        assert len(self.enabled_sub_protocols) > 0
```
pipermerriam: *nitpick*

explicit exception over assertions in non-test code.

evm/p2p/peer.py
```diff
@@ -449,6 +452,23 @@ def les_proto(self):
...
+        if self._eth_proto is None:
+            for proto in self.enabled_sub_protocols:
+                if proto.name == ETHProtocol.name:
+                    self._eth_proto = proto
```
pipermerriam: It feels like this loop is missing a `break` statement.  I *think* that there's an implicit understanding that there aren't multiple values in `self.enabled_sub_protocols` which have the same name.  However, it bothers me that this code *can* set `self._eth_proto` multiple times (albeit to the same value) before exiting.

If we add a break, we can also change the next line from being `if self._eth_proto is None` to be an `else` clause  for this `for` loop which I think is *slightly* more semantically correct for how this code flows.

* Loop over proto names and set the local `_eth_proto`
* Break when you find the proper protocol
* If the loop never breaks, then an error has occurred.

evm/p2p/peer.py
```diff
@@ -449,6 +452,23 @@ def les_proto(self):
...
+            if self._eth_proto is None:
+                raise Exception(
```
pipermerriam: Anything better than a bare *Exception* that can be raised here?

evm/p2p/peer.py
```diff
@@ -490,8 +514,22 @@ def __init__(self, block_number, block_hash, total_difficulty, genesis_hash):
     try:
         peer = loop.run_until_complete(
             asyncio.wait_for(
-                handshake(remote, ecies.generate_privkey(), LESPeer, chaindb, network_id),
+                handshake(remote, ecies.generate_privkey(), peer_class, chaindb, network_id),
                 HANDSHAKE_TIMEOUT))
+        # Request some stuff, just as a basic test.
+        block_hash = decode_hex(
+            '0x59af08ab31822c992bb3dad92ddb68d820aa4c69e9560f07081fa53f1009b152')
```
pipermerriam: Assuming this is going to remain beyond just your testing, can you add a comment indicating what block this is

evm/p2p/peer.py
gsalgado: There can be multiple values with the same name in `.enabled_sub_protocols`, but they will have different versions, and we actually want to continue in the loop until the last protocol that matches `ETHProtocol.name` because that's going to be the one with the highest version. This is all a bit of a hack, really; we need a generic mechanism to access the protocols supported by a given Peer instance, but I wanted to wait until we have more use cases before attempting to implement it.
Would it be ok with you if I just added a FIXME comment here for now?

evm/p2p/peer.py
gsalgado: This is inside the `if __name__ == "__main__":` block, so it's all just used when testing manually, but I'll add a link to the block this hash belongs to

gsalgado:
Most of the TODOs cannot be really addressed yet as we don't have full-sync capabilities and thus won't have the data needed. My plan was to try and get the full sync working, and then implement the remaining eth commands

evm/p2p/peer.py
pipermerriam: >  Would it be ok with you if I just added a FIXME comment here for now?

Yes, along with the explanation about there being multiple values and why it's written this way.

pipermerriam:
:shipit:

evm/p2p/peer.py
gsalgado: I'm not changing it here as this whole thing is a temporary hack anyway.

Issue #262, opened 2018-01-11
Deduplication of `Message.gas_price` and `Message.origin`


pipermerriam:
### What is wrong?

The `Message.gas_price` and `Message.origin` are copied for every level of computation during VM execution, despite them remaining *constant* for the entire context of the VM execution.

It would be nice to remove this duplication in favor of having these values stored in *one* location.

Also see: `sighash` from https://github.com/ethereum/py-evm/pull/259/files

### How can it be fixed

Implement a `TransactionContext` object, similar to the `ExecutionContext` object introduced in #273 

This would be an object that we treat as immutable which houses information like `gas_price`, and `sighash` for the in-progress sharding work.

Since the info for this object will change over time with changes to the VM rules we should use the following API to allow for changing out the class which contains this data.

```
# vm_state.py

class VMState:
    transaction_context_class = None  # must be set in individual VMState classes

    @classmethod
    def get_transaction_context_class(cls):
        # error check to be sure it's set
        return self.transaction_context_class
```

Then, within the `VMState.execute_transaction` implementation, after the `message` instance is created, we should also generate the transaction context instance.

Then, we modify `vm_state.get_computation` to take both `message` and `transaction_context` as arguments.   The `transaction_context` should be stored on the `Computation` object.  When child computations are spawned, this object should be passed into those instances.

Any opcodes which previously accessed values from the `Message` which now reside on the `TransactionContext` object will need to be updated.

pipermerriam:
@gsalgado got any ideas on this one?

gsalgado:
I've had a look at that code and couldn't think of anything other than what you suggest, but I'm not very familiar with that area of the code. I'll sleep on it and maybe I'll come up with something...

pipermerriam:
With respect to #273 

The `BlockInfo` object there holds *constants* that are true for the entire execution context of a block.  We could follow the same pattern and introduce a `TransactionInfo` object that mirrors this for data that is constant through the execution of a transaction.

pipermerriam:
@lrettig this would be a good issue for you to look into.

lrettig:
Do we want to include `message.is_static` here as well? Per https://github.com/ethereum/EIPs/blob/master/EIPS/eip-214.md, this appears to be static in sub-calls.

PR #273, opened 2018-01-15
Decouple blockchain opcodes and state mutable data

### LINKED ISSUE #270: Decouple blockchain opcodes and state mutable data
### What is wrong?

1. Decouple `VMState.block_header` and `VMState` properties (`timestamp, block_number, difficulty, and gas_limit)`, because shard chains use `period_start_prevhash` block properties of main chain in contracts, but `VMState.state_db()` should use the `state_root` of `VMState.collation_header` to build the state root trie.
2. Note that only `COINBASE` opcode refers to the shard coinbase.

### How can it be fixed

1. As v suggested, implement a simple `BlockInfo` class:

    ```python
    class BlockInfo():
        # For shard chain, refer to the shard coinbase.
        _coinbase = None

        # For shard chian, block data of period_start_prevhash.
        _timestamp = None
        _number = None
        _difficulty = None
        _gas_limit = None
        _prev_headers = None

        def __init__(
                self,
                coinbase,
                timestamp,
                number,
                difficulty,
                gas_limit,
                prev_headers):
            self._coinbase = coinbase
            self._timestamp = timestamp
            self._number = number
            self._difficulty = difficulty
            self._gas_limit = gas_limit
            self._prev_headers = prev_headers

        @property
        def coinbase(self):
            return self._coinbase

        @property
        def timestamp(self):
            return self._timestamp

        @property
        def number(self):
            return self._number

        @property
        def difficulty(self):
            return self._difficulty

        @property
        def gas_limit(self):
            return self._gas_limit

        @property
        def prev_headers(self):
            return self._prev_headers
    ```

2. Pass  `block_info` to initialize `VMState`.
3. Assign `coinbase` separately.
4. Arrange/Refactor the temporary mutable data during state transition: 
	* Part of  the given block’s header from `VMState.__init__(..., block_header, ...)` and replace with  `VMState.apply_transaction(...,block,...)`
		* `state_root`
		* `gas_used` - update in `VMState.make_receipt`
	* `receipts`  - now is part of `VMState`
	* **Solution 1**: generalize `VMState.block_header` to `VMState.header` and make sure it’s compatible for both `BlockHeader` and `CollationHeader` structure.
	* **Solution 2**:
		1. pass `block` to `VMState.make_receipt` function so that we just use the given block to maintain `gas_used`.
		2. pass `state_root` and `receipts` to `VMState.__init__(...,state_root, receipts, ...)`
	* **Solution 3**: apply Vitalik’s suggestion to arrange these variables to a new simple `StateObj` structure. And the apply transaction would become:

    ```python
    # In create_block level function
    # state_obj: previous state_obj or new StateObj

    state_obj = StateObj(
        state_root=state_obj.state_root,
        gas_used=state_obj.gas_used,
        receipts=state_obj.receipts,
    )
    block_info = BlockInfo(
        coinbase=block.header.coinbase,    # or collation.coinbase
        timestamp=block.header.timestamp,
        number=block.header.number,
        difficulty=block.header.difficulty,
        gas_limit=block.header.gas_limit,
        prev_headers=prev_headers,
    )
    vm_state = cls.get_state_class()(
        chaindb=witness_db,
        state_obj=state_obj,
        block_info=block_info,
    )
    computation, result_block, _ = vm_state.apply_transaction(
        transaction=transaction,
        block=block,
        is_stateless=True,
    )
    state_obj = computation.vm_state.state_obj
    ```
    I'd like to try **Solution 2** first because I want keep the `VMState` simple, and I'm also fine with **Solution 3**.

5. Check with @pipermerriam : do we design that `VMState` only be used for handling the “transaction level” state transition? Or we would introduce “block level” function like `VMState.apply_block` function?
 
6. About uncle validation, there are: 
	* `Chain.mine_block` or `Chain.import_block(block)`
	    └─ `Chain.validate_block(block)`
			    └─ `Chain.validate_uncles(block)`
	*  `VM.pack_block(block)`
	    └─  `VMState.validate_block(block)`
			    └─`VMState.validate_uncle(block, uncle)`

    Regarding (5):
    * If `VMState` only handle “transaction level” state transition:
	    * The  `VMState.validate_block(block)` should be move to `VM`.
	    * With this change, we can also modify `VMState.prev_header` to simpler `VMState.prev_hashes`.
    * If `VMState` also handle “block level” state transition, should we move some more functions from `VM` to `VMState`?

hwwhww:
### What was wrong?
To generalize `VMState` for both sharding client and non-sharding client in the early stage, we need to fix #270: decouple blockchain opcodes and state mutable data.

updated:
As discussion in #270, it makes sense to move `validate_block` from `VMState` to `VM` since `VM.import_block(block)` and  `VM.mine_block()` are in `VM` now.

### How was it fixed?
- [x] Add BlockInfo object to hold the block information opcode values
- [x] Refactor `VMState.block_header` to other format (see #270 for more details)
- [x] Move `validate_block` from `VMState` to `VM`
- [x] Now `VMState` doesn't necessarily need the whole previous 256 headers, the hashes are enough to provide the context. So this PR also changed the `prev_headers` to `prev_hashes` and removed the unused methods.

#### Cute Animal Picture
levi


evm/vm/block_info.py
```diff
@@ -0,0 +1,62 @@
...
+            prev_headers):
+        self._coinbase = coinbase
+        self._timestamp = timestamp
+        self._block_number = block_number
+        self._difficulty = difficulty
+        self._gas_limit = gas_limit
+        self._prev_headers = prev_headers
+
+    @classmethod
+    def generate_by_block_header(cls, block_header, prev_headers, coinbase=None):
```
pipermerriam: This would closer align with existing naming conventions if it was `generate_from_block_header` or more concisely, just `from_block_header`

evm/vm/block_info.py
```diff
@@ -0,0 +1,62 @@
+class BlockInfo():
+    # For shard chain, refer to the shard coinbase.
+    _coinbase = None
+
+    # For shard chian, block info of period_start_prevhash.
+    _timestamp = None
+    _number = None
+    _difficulty = None
+    _gas_limit = None
+    _prev_headers = None
```
pipermerriam: Given all of these are written as direct passthrough `property` methods with no setters, I think it might be better to remove the underscore prefix and property methods and expect them to be accessed directly.

evm/vm/block_info.py
```diff
@@ -0,0 +1,62 @@
+class BlockInfo():
```
pipermerriam: Wondering if it would be better to not use the name *block* and instead to call this something like

* VMContext
* ExecutionContext

I'm partially thinking about https://github.com/ethereum/py-evm/issues/262 and whether the changes in this pull request can help us fix that issue as well (but not in this PR)

evm/vm/block_info.py
hwwhww: I think the purpose of this strategy *is* making these fields won't be changed after initialization. In this way, it will raise `AttributeError` when someone tries to modify these properties. But maybe there's a better approach for this purpose? I'd love to hear your thoughts.

evm/vm/block_info.py
hwwhww: My carelessness, thank you.

hwwhww:
I try to pass `VM.block.header.state_root` to initialize `VMState`. Since the `VM.block.block_header` no more be used as a mutable object passed to `VMState.__init__(..., block_header,...)`, it turns out that there is a new side effect: it needs to maintain `VMState.block.header.state_root` whenever modifying the `state_db` in `VM` level. For example:
https://github.com/hwwhww/py-evm/blob/a963f0fdad552f9eade8e1cdcee88dd2d0db8e97/evm/vm/forks/homestead/headers.py#L81-L89
```python
        with vm.state.state_db() as state_db:
            for account in dao_drain_list:
                account = decode_hex(account)
                balance = state_db.get_balance(account)
                state_db.delta_balance(dao_refund_contract, balance)
                state_db.set_balance(account, 0)

+           # Update state_root manually
+           vm.block.header.state_root = state_db.root_hash
```
I'm still investigating any better solution.

hwwhww:
#### Updated PR description:
As discussion in #270, it makes sense to move `validate_block` from `VMState` to `VM` since `VM.import_block(block)` and  `VM.mine_block()` are in `VM` now.

#### Changelogs:
1. Moved `validate_block` from `VMState` to `VM`.
2. Now `VMState` doesn't necessarily need the whole previous 256 headers, the hashes are enough to provide the context. So this PR also changed the `prev_headers` to `prev_hashes` and removed the unused methods.

evm/vm/block_info.py
pipermerriam: I don't think that these will throw errors if modified.  I believe python will happily let you assign over top of `@property` methods.

- The pragmatic part of me says to just use direct properties and we can have *soft* enforcement of not overwriting values (as in, in code reviews and things, we have a pretty strong *culture* of immutability and well defined boundaries when we require mutability).
- The purist likes this idea and suggests making our own `read_only_property` decorator which includes a `setter` which raises such an attribute error.

I'm leaner towards *soft* enforcement, but I'm willing to hear any arguments you have for the later.

evm/vm/base.py
```diff
@@ -379,21 +477,21 @@ def get_block_by_header(cls, block_header, db):
         return cls.get_block_class().from_header(block_header, db)
 
     @classmethod
-    def get_prev_headers(cls, last_block_hash, db):
-        prev_headers = []
+    def get_prev_hashes(cls, last_block_hash, db):
+        prev_hashes = []
```
pipermerriam: Can we use the `to_tuple` decorator here to avoid the `prev_hashes.append` calls.

evm/vm/execution_context.py
```diff
@@ -0,0 +1,62 @@
+class ExecutionContext():
```
pipermerriam: *style nitpick*

You can drop the `()`

evm/vm/forks/frontier/validation.py
```diff
@@ -18,7 +18,11 @@ def validate_frontier_transaction(vm_state, transaction):
     if sender_balance < total_cost:
         raise ValidationError("Sender account balance cannot afford txn")
 
-    if vm_state.block_header.gas_used + transaction.gas > vm_state.block_header.gas_limit:
+    if vm_state.receipts:
+        gas_used = vm_state.receipts[-1].gas_used
```
pipermerriam: What do you think about adding a`@propery` method for the `vm_state` class since this block of logic is repeated at least one more time elsewhere in the code.

```python
@propert
def gas_used(seelf):
    if vm_state.receipts:
        return ...
    else:
        return 0
```

evm/vm_state.py
```diff
@@ -141,36 +134,19 @@ def is_key_exists(self, key):
...
     def get_ancestor_hash(self, block_number):
         """
         Return the hash of the ancestor with the given block number.
         """
-        ancestor_depth = self.block_header.block_number - block_number - 1
+        ancestor_depth = self.block_number - block_number - 1
         if (ancestor_depth >= MAX_PREV_HEADER_DEPTH or
                 ancestor_depth < 0 or
-                ancestor_depth >= len(self.prev_headers)):
+                ancestor_depth >= len(self.execution_context.prev_hashes)):
```
pipermerriam: We could improve readability here by pulling the full contents of the `if` statement and storing in a variable and then just using the variable in the `if` statement.

```python
is_ancestor_depth_out_of_range = (
    ancestor_depth >= MAX_PREV_HEADER_DEPTH or
    ancestor_depth < 0 or
    ancestor_depth >= len(self.execution_context.prev_hashes)
))
```

hwwhww:
#### Changelogs
Refactored `VM` and `VMState` for avoiding to update `state_db` in `VM` level.
1. Moved the reward methods from `VM` to `VMState`.
2. Pick partial logic from `VM.mine_block()` to `VMState.finalize_block(block)` method.

* Now the only one special case of "updating `state_root` manually in VM level” under `evm` is [the DAO fork header initialization](https://github.com/hwwhww/py-evm/blob/18578f1ca25e0a2dc45f003eac9e8f4579be8dc5/evm/vm/forks/homestead/headers.py#L89-L90).
* Under `tests`, I also added 3 lines to update `state_root` manually in VM level for setup the JSON fixtures test cases, but I think it's fine if only for the testing setup.
* With these changes, we also got clear `VM.create_block(block)`.

hwwhww:
This branch has passed Travis CI in my repository, so I go :shipit:

evm/vm/execution_context.py
```diff
@@ -0,0 +1,62 @@
+class ExecutionContext:
```
lrettig: Is there a reason this is an old-style python object?

pipermerriam:
```python
class AwesomeNewstyleOnlyInPython3:
    ...
```

It's only old-style in python2.  In python 3 this is equivalent to `class Thing(object)`.

lrettig: 🤯

PR #274, opened 2018-01-16
Convert configure to base class

### LINKED ISSUE #245: Add a new `Configurable` base class
### What is wrong?

One of the standard patterns that has developed in this codebase is classes which can be configured.

```python
class VM(object):
    @classmethod
    def configure(cls,
                  name=None,
                  **overrides):
        if name is None:
            name = cls.__name__

        for key in overrides:
            if not hasattr(cls, key):
                raise TypeError(
                    "The VM.configure cannot set attributes that are not "
                    "already present on the base class.  The attribute `{0}` was "
                    "not found on the base class `{1}`".format(key, cls)
                )
        return type(name, (cls,), overrides)
```

This pattern is found in:

- `evm.rlp.blocks.BaseBlock`
- `evm.vm.base.VM`
- `evm.chains.chain.Chain`

As well as in the not yet merged `evm.computation.Computation` and `evm.vm.vm_state.VMState` from #244 

### How can it be fixed

Create a new base class that contains this functionality `evm.utils.datatypes.Configurable` which implements a standard implementation of this `configure` method.   Then modify all of the classes which implement this pattern to instead inherit from this base class.

njgheorghita:
### What was wrong?
`Configure` classmethod was repeated across different classes. 

### How was it fixed?
Wrote `evm.utils.datatypes.Configurable` base class for classes to inherit from. 
Classes changed. . .
`evm.rlp.blocks.BaseBlock`
`evm.vm.base.VM`
`evm.chains.chain.Chain`
`evm.computation.Computation`
`evm.vm_state.VMState`

#### Cute Animal Picture
giphy 44

closes #245

### DIFF
--- a/evm/utils/datatypes.py
+++ b/evm/utils/datatypes.py
@@ -1,0 +1,21 @@
+class Configurable(object):
+    """
+    Base class for simple inline subclassing
+    """
+    @classmethod
+    def configure(cls,
+                  name=None,
+                  **overrides):
+
+        if name is None:
+            name = cls.__name__
+
+        for key in overrides:
+            if not hasattr(cls, key):
+                raise TypeError(
+                    "The {0}.configure cannot set attributes that are not "
+                    "already present on the base class. The attribute `{1}` was "
+                    "not found on the base class `{2}`".format(cls.__name__, key, cls)
+                )
+
+        return type(name, (cls,), overrides)
--- a/evm/chains/chain.py
+++ b/evm/chains/chain.py
@@ -26,13 +26,14 @@ from evm.validation import (
     validate_uint256,
     validate_word,
 )
-
 from evm.rlp.headers import (
     BlockHeader,
 )
-
 from evm.utils.chain import (
     generate_vms_by_range,
+)
+from evm.utils.datatypes import (
+    Configurable,
 )
 from evm.utils.headers import (
     compute_gas_limit_bounds,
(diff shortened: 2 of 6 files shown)


evm/utils/datatypes.py
```diff
@@ -0,0 +1,34 @@
...
+        if name is None:
+            name = cls.__name__
+
+        if 'vms_by_range' in overrides:
+            raise ValueError("Cannot override vms_by_range.")
+
+        # Organize the Chain classes by their starting blocks.
+        if 'vm_configuration' in overrides:
+            vm_configuration = overrides['vm_configuration']
+            overrides['vms_by_range'] = generate_vms_by_range(vm_configuration)
```
pipermerriam: I think rather than including this in the base class since it's specific to the `VM` class, it should instead be implemented in the `VM` class using a `super` call.

```python
# in the VM class
@classmethod
def configure(cls, name, vm_configuration, **overrides):
    if 'vms_by_range' in overrides:
        raise ValueError(...)
    overrides['vms_by_range'] = generate_vms_by_range(vm_configuration)
    return super().configure(name, **overrides)
```

njgheorghita: @pipermerriam makes sense, though correct me if i'm wrong - shouldn't that go into `evm.chains.chain.Chain` not `evm.vm.base.VM`? if i'm wrong, i'll switch it over tomorrow

pipermerriam: You are correct.  I crossed wires while banging that example out.

pipermerriam:
@njgheorghita time to practice rebasing!?

njgheorghita:
@pipermerriam haha always, should be good to go now once travis completes

pipermerriam:
> once travis completes

Lately that means something like tomorrow :P :(

Issue #263, opened 2018-01-11
Access restrictions are not honored when executing sharding transactions


jannikluhn:
### What is wrong?

Sharding transactions carry an access list. If state is accessed that has not been specified there execution should fail. This is not happening right now.

### How can it be fixed

Pass the access list to `computation.vm_state.state_db()`.

PR #276, opened 2018-01-16
Check access restrictions during tx execution

jannikluhn:
### What was wrong?

#263 

### How was it fixed?

- Fix `Computation.state_db`
- Convert `transaction.access_list` to prefix list form during initialization and store the result in `transaction.prefix_list`.
- Whenever a `Computation` object is available (i.e. in all opcode functions) use `computation.state_db` instead of `computation.vm_state.state_db` to access state.
- In all other places use `vm.state.state_db` but pass `transaction.prefix_list` explicitly.

#### Cute Animal Picture


jannikluhn:
Now that the Travis tests are running again I realized that `test_sharding_vm` is failing. Reasons are that it uses `NestedTrieBackend`instead of `FlatTrieBackend` and because it doesn't specify an access list. Fixing this is not straightforward though because the test creates a contract and contract creation currently calls `state_db.touch_account` which isn't supported in `FlatTrieBackend` (on purpose -- if I remember correctly we concluded that they are superfluous since SpuriousDragon).

I'm unsure how to continue there. I'm leaning towards fixing the test in so far as to switch to `FlatTrieBackend` and to specify the appropriate access list, but let it xfail. Then, once contract creation is implemented properly remove the xfail. Other opinions on that?

hwwhww:
@jannikluhn  I recall that the reason why @NIC619 used `NestedTrieBackend` in `testing_sharding_vm.py` is also because the access list generator hasn't been implemented.
So if we switch to `FlatTrieBackend`, the exception would only be raised when contract creation? IMO, I think that's okay to do xfail for now, just remember to edit #263 or open a new issue for follow-up.

Also, to finish the contract creation, I guess we also need the #269, you can wait for #269 to be 
 merged and then adjust these tests.

jannikluhn:
> I recall that the reason why @NIC619 used NestedTrieBackend in testing_sharding_vm.py is also because the access list generator hasn't been implemented.

As I understood it access list generation is supposed to be handled by contract programming languages, so that's why I started writing them manually for the tests.

> So if we switch to FlatTrieBackend, the exception would only be raised when contract creation?

Yes, exactly.

> Also, to finish the contract creation, I guess we also need the #269, you can wait for #269 to be
merged and then adjust these tests.

#269 won't change anything about this unfortunately.

I'll go ahead and implement the xfail solution, as well as open an issue for that.

hwwhww:
@jannikluhn oh, sad. #269 also added some new test cases, just wondering if we merge #269 first, could help for reducing conflicts for this PR?

jannikluhn:
Looks like the tests in #269 would run into the same problem if this PR (#276) is merged first. So I'd vote to merge #269 first, then I can fix these issues here (specifying correct access lists and applying `@xfail`'s).

NIC619:
I will get the #269 merged as soon as tests passed.

jannikluhn:
Rebased and added access lists to the tests of #269 . Also marked them with xfail because of #281 and old gas payment style. The tests pass if the following lines are commented out:

- https://github.com/jannikluhn/py-evm/blob/340e97840dcdf54d3d477d851d6a0863902878f5/evm/vm/forks/sharding/__init__.py#L174-L175
- https://github.com/jannikluhn/py-evm/blob/340e97840dcdf54d3d477d851d6a0863902878f5/evm/vm/forks/sharding/computation.py#L51-L52
- https://github.com/jannikluhn/py-evm/blob/340e97840dcdf54d3d477d851d6a0863902878f5/evm/logic/call.py#L289

I'll apply similar fixes to #230 once it is merged, that's why I'm setting this as work in progress again.

pipermerriam:
@jannikluhn FYI

This is a common pytest pitfall.  `xfail` is intended for tests that are unreliable, rather than tests that are *required* to fail.  A test marked with `xfail` that passes doesn't cause the test suite to fail.

In cases where we have an explicit failure, I'd rather stay away from `xfail` and instead write the test with a `with pytest.raises(TheExpectedError)` in the place that it's expected to fail.  Then at least when we fix the error, we'll have a clearly failing test so that we know that section needs to be updated as well.

In general, I think we should stay way from `xfail` entirely because it's not that useful.

jannikluhn:
The [pytest documentation](https://docs.pytest.org/en/latest/skipping.html) says the following:

> A common example [for xfail] is a test for a feature not yet implemented, or a bug not yet fixed.

so I think xfail is correct here. I'm not a fan of adding `pytest.raises` because that would change the test itself, although it is correct (it's the tested code that's broken).

I agree though that it would be nicer if the test would fail if it passes unexpectedly. `xfail` seems to have a `strict` flag which does exactly that. What do you think about keeping the `xfail`s and adding the `strict` flag?

pipermerriam:
@jannikluhn eating my words.  Looks like `xfail` has been improved since I last used it and now does what we/you need.  Feel free to continue using it.  I think we should add the `xfail_strict=true` to our `pytest.ini` to ensure that this is always the case.

Also, I have a preference towards using `raises` in conjunction with `reason` since it's more explicit about *how* the test is expected to fail.

jannikluhn:
Replaced by #313.

PR #279, opened 2018-01-17
p2p: Implement StateDownloader

gsalgado:
A service that uses the GetNodeData command of ETH/PV63 to download the
whole state starting from a given root hash.


evm/p2p/state.py
```diff
@@ -0,0 +1,187 @@
...
+                # from _pending_nodes.
+                if node_key in self._pending_nodes:
+                    self._pending_nodes.pop(node_key)
+
+    # FIXME: Need a better criteria to select peers here.
+    async def get_random_peer(self) -> ETHPeer:
+        while len(self.peer_pool.peers) == 0:
+            self.logger.debug("No connected peers, sleeping a bit")
+            await asyncio.sleep(0.5)
+        peer = self.peer_pool.peers[random.randrange(0, len(self.peer_pool.peers))]
```
pipermerriam: I think `random.choice(self.peer_pool.peers)` does this.

evm/p2p/state.py
```diff
@@ -0,0 +1,187 @@
...
+                except SyncRequestAlreadyProcessed:
+                    # This means we received a node more than once, which can happen when we retry
+                    # after a timeout.
+                    pass
+                # Since a node may be received more than once, check before trying to remove it
+                # from _pending_nodes.
+                if node_key in self._pending_nodes:
+                    self._pending_nodes.pop(node_key)
+
+    # FIXME: Need a better criteria to select peers here.
```
pipermerriam: I've imagined us storing a local database of peers with somewhat simple metrics attached to them so that we could rank peers by some sort of simple score based on how reliable and fast the peer is.

gsalgado: Yep, something like that is what I had in mind as well, but here we might want to rank peers by latency based solely on our interactions with them during the downloading of the state and not from previous interactions as well.

pipermerriam: Think you understand the problem sufficiently well to write it up in an issue? I can write up something generic but I don't have a good sense for what parts of the code it would fit into.

gsalgado:
I'm not sure using hypothesis here adds enough value to justify the extra complexity, as accounts with different balances/nonces/etc will be treated exactly the same. Where I think hypothesis adds some value is in the TrieSync tests (in py-trie), since there we want to exercise the code with all possible node types/sizes

evm/p2p/state.py
```diff
@@ -0,0 +1,187 @@
...
+class StateDownloader:
+    logger = logging.getLogger("evm.p2p.state.StateDownloader")
+    _pending_nodes = {}  # type: Dict[Any, float]
+    _total_processed_nodes = 0
+    _report_interval = 10  # Number of seconds between progress reports.
+    # TODO: Experiment with different timeout/max_pending values to find the combination that
+    # yields the best results.
+    _max_pending = 5 * MAX_STATE_FETCH
+    _reply_timeout = 5  # seconds
+    _last_report_time = time.mktime(time.gmtime(0))
```
pipermerriam: interesting.  Had to look this up as my longer way to do this was always:

```python
datetime.datetime.utcfromtimestamp(0).timestamp()
```

Yours is much more concise.  That said.  Can we just initialize this at zero and call it a readability/complexity win?  Glancing around, I think that it'll be have exactly the same.

evm/p2p/state.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(self, chaindb, state_db, root_hash, network_id, privkey):
+        self.peer_pool = PeerPool(ETHPeer, chaindb, network_id, privkey, self.msg_handler)
+        self.root_hash = root_hash
+        self.scheduler = StateSync(root_hash, state_db, self.logger)
+
+    def msg_handler(self, peer: BasePeer, cmd: protocol.Command,
+                    msg: protocol._DecodedMsgType) -> None:
+        """The callback passed to BasePeer, called for every incoming message."""
+        peer = cast(ETHPeer, peer)
+        if isinstance(cmd, eth.NodeData):
```
pipermerriam: Is it intentional that this `if` has no `elif/else` causing the function to silently exit when this is false?

evm/p2p/state.py
```diff
@@ -0,0 +1,187 @@
...
+                try:
+                    self.scheduler.process([(node_key, node)])
+                except SyncRequestAlreadyProcessed:
+                    # This means we received a node more than once, which can happen when we retry
+                    # after a timeout.
+                    pass
+                # Since a node may be received more than once, check before trying to remove it
+                # from _pending_nodes.
+                if node_key in self._pending_nodes:
+                    self._pending_nodes.pop(node_key)
```
pipermerriam: Maybe rather than doing the condition check we can do one of

```
self._pending_nodes.pop(node_key, None)
# or
try:
    self._pending_nodes.pop(node_key)
except KeyError:
    pass
```

I'm sure the `in` check is cheap but I think we gain a tiny bit of performance this way.

evm/p2p/state.py
```diff
@@ -0,0 +1,187 @@
...
+                # Yield control to ensure the Peer's msg_handler callback is called to process any
+                # nodes we may have received already. Otherwise we spin too fast and don't process
+                # received nodes often enough.
+                await asyncio.sleep(0)
+
+            self._maybe_report_progress()
+
+        self.logger.info("Finished state sync with root hash %s" % encode_hex(self.root_hash))
+
+    def _maybe_report_progress(self):
```
pipermerriam: Eventually, it seems like it would be more appropriate for this type of reporting to happen at an orchestration level higher up in the stack which is handling all the various long-running processes that need to be run.

evm/p2p/state.py
gsalgado: Yep, it is because during the state download all we care about are the NodeData messages sent in response to our requests.

evm/p2p/state.py
gsalgado: I'd prefer the former, but since this whole thing is pretty heavy on network round trips (since geth limits the maximum number of nodes sent in one message to 384), I'm afraid any performance improvements like this will be completely irrelevant

evm/p2p/state.py
gsalgado: Yeah, that might make sense. Another alternative would be to handle it in the base class for asyncio services (#280), so subclasses just have to implement, say, a `report_progress()` method

evm/p2p/state.py
gsalgado: Do you think #282 is a good start?

PR #283, opened 2018-01-18
Two new gas estimators

### LINKED ISSUE #264: Implement gas estimation
### What is wrong?

We need to implement gas estimation for the JSON-RPC interface.

An ideal gas estimation API should take into account the following.

- total gas used isn't always right due to gas refunds.
- transactions can alter execution flow based on the txn gas.

### How can it be fixed

This should be implemented using a swappable backend API.  Accurate gas estimations are likely to be computationally expensive where-as naive estimations based on gas used are likely to work for most use cases and should be computationally cheap.

The API can be something as simple as a function which takes `VM` instance and a `Transaction` instance and returns the gas estimation.  It *might* be appropriate to include this as an API on the `Chain` and `VM` as `Chain.estimate_transaction_gas()` and `VM.estimate_transaction_gas()`

#### Strategies

We should first implement a `PeekGasUsedStrategy` which uses the peek gas usage

1. run the computation with the gas limit set at the full gas limit for the next pending block.
2. return the `computation.gas_meter.start_gas - computation.gas_metea.gas_remaining` for the outer computation.

Then we should implement what I believe is the current strategy used by go-ethereum which starts with running the transaction using the gas limit for the next pending block and then performs a binary search, re-running the transaction at each level of the search until it finds the threshold for which the transaction succeeds/fails.

> Note that the binary search can be performed at a higher granularity than 1 gas.  From @carver comment, a 1024 granularity should produce approximately a 2x speedup.

carver:
### What was wrong?

#264 - Want to easily estimate gas usage of a transaction.

### How was it fixed?

Now fully fixes #264 

- Add two gas estimators to `Chain`.
  - The simplest one just runs the transaction to see how much gas was used, and multiplies the estimate by 2. The estimator cannot return an amount higher than the gas set in the transaction. It might under-guess the gas if there is a long chain of calls (because it doesn't account for the 63/64 gas passed to subcall)
  - The more complicated one does binary search, (roughly) as defined in #264
- Add `gas_used_by(computation)` API to Transaction (for now, just intrinsic cost + gas used)
- Generate the state of a temporary block on top of the current header. This allows us to estimate gas taking into account the pending state, but still have a full block's worth of gas available to test with.
- Some flake8 warning suppression cleanup
- gas_estimator is added to `Chain`. It takes a `VMState` and a `Transaction`. It loads an estimator function from the environment variable `GAS_ESTIMATOR_BACKEND_FUNC`. If none is set, it loads the `binary_search_intrinsic_tolerance` estimator, which runs binary search to find the minimum gas required to within 21k gas.
- Added a `utils.spoof.SpoofTransaction` class which neatly lets you alter some attributes of a signed transaction without changing the sender, using `spoofed = SpoofTransaction(transaction, gas=123456)`.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/chain.py
+++ b/evm/chains/chain.py
@@ -15,6 +15,9 @@ from evm.constants import (
     BLANK_ROOT_HASH,
     MAX_UNCLE_DEPTH,
 )
+from evm.estimators import (
+    get_gas_estimator,
+)
 from evm.exceptions import (
     BlockNotFound,
     TransactionNotFound,
@@ -56,6 +59,7 @@ class Chain(object):
     header = None
     network_id = None
     vms_by_range = None
+    gas_estimator = None
 
     def __init__(self, chaindb, header=None):
         if not self.vms_by_range:
@@ -67,6 +71,8 @@ class Chain(object):
         self.header = header
         if self.header is None:
             self.header = self.create_header_from_parent(self.get_canonical_head())
+        if self.gas_estimator is None:
+            self.gas_estimator = get_gas_estimator()
 
     @classmethod
     def configure(cls, name, vm_configuration, **overrides):
@@ -267,6 +273,12 @@ class Chain(object):
         vm = self.get_vm()
         computation, _ = vm.apply_transaction(transaction)
         return computation
+
+    def estimate_gas(self, transaction, at_header=None):
+        if at_header is None:
+            at_header = self.get_canonical_head()
+        with self.get_vm(at_header).state_in_temp_block() as state:
+            return self.gas_estimator(state, transaction)
 
     def import_block(self, block, perform_validation=True):
         """
--- a/evm/db/__init__.py
+++ b/evm/db/__init__.py
@@ -8,15 +8,14 @@ from evm.utils.module_loading import (
 DEFAULT_DB_BACKEND = 'evm.db.backends.memory.MemoryDB'
 
 
-def get_db_backend_class(import_path=None):
-    if import_path is None:
-        import_path = os.environ.get(
-            'CHAIN_DB_BACKEND_CLASS',
-            DEFAULT_DB_BACKEND,
-        )
+def get_db_backend_class():
+    import_path = os.environ.get(
+        'CHAIN_DB_BACKEND_CLASS',
+        DEFAULT_DB_BACKEND,
+    )
     return import_string(import_path)
 
 
-def get_db_backend(import_path=None, **init_kwargs):
-    backend_class = get_db_backend_class(import_path)
+def get_db_backend(**init_kwargs):
+    backend_class = get_db_backend_class()
     return backend_class(**init_kwargs)
--- a/evm/estimators/__init__.py
+++ b/evm/estimators/__init__.py
@@ -8,6 +8,6 @@ from evm.utils.module_loading import (
 def get_gas_estimator():
     import_path = os.environ.get(
         'GAS_ESTIMATOR_BACKEND_FUNC',
-        'evm.estimators.gas.binary_search_intrinsic_tolerance',
+        'evm.estimators.gas.binary_gas_search_intrinsic_tolerance',
     )
     return import_string(import_path)
--- a/evm/estimators/__init__.py
+++ b/evm/estimators/__init__.py
@@ -5,10 +5,9 @@ from evm.utils.module_loading import (
 )
 
 
-def get_gas_estimator(import_path=None):
-    if not import_path:
-        import_path = os.environ.get(
-            'GAS_ESTIMATOR_BACKEND_FUNC',
-            'evm.estimators.gas.binary_search_intrinsic_tolerance',
-        )
+def get_gas_estimator():
+    import_path = os.environ.get(
+        'GAS_ESTIMATOR_BACKEND_FUNC',
+        'evm.estimators.gas.binary_search_intrinsic_tolerance',
+    )
     return import_string(import_path)
--- a/evm/estimators/__init__.py
+++ b/evm/estimators/__init__.py
@@ -1,0 +1,14 @@
+import os
+
+from evm.utils.module_loading import (
+    import_string,
+)
+
+
+def get_gas_estimator(import_path=None):
+    if not import_path:
+        import_path = os.environ.get(
+            'GAS_ESTIMATOR_BACKEND_FUNC',
+            'evm.estimators.gas.double_execution_cost',
+        )
+    return import_string(import_path)
--- a/evm/estimators/__init__.py
+++ b/evm/estimators/__init__.py
@@ -9,6 +9,6 @@ def get_gas_estimator(import_path=None):
     if not import_path:
         import_path = os.environ.get(
             'GAS_ESTIMATOR_BACKEND_FUNC',
-            'evm.estimators.gas.double_execution_cost',
+            'evm.estimators.gas.binary_search_intrinsic_tolerance',
         )
     return import_string(import_path)
--- a/evm/estimators/gas.py
+++ b/evm/estimators/gas.py
@@ -3,6 +3,14 @@ from cytoolz import (
 )
 
 
+from evm.exceptions import (
+    OutOfGas,
+)
+from evm.utils.spoof import (
+    SpoofTransaction,
+)
+
+
 @curry
 def execute_plus_buffer(multiplier, state, transaction):
     computation = state.execute_transaction(transaction)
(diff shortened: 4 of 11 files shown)


pipermerriam:
Ideas for how to get around the transaction mutability:

- execute via the `apply_computation` API.  
- override some APIs on the transaction object to disable the signature checking, or to make modifying the gas not change the *message* that it checks the signature against.
- override the `transaction.get_sender()` to return a hard coded value and sign the modified transaction using a different key.

All of these probably have pitfalls.

evm/estimators/gas.py
```diff
@@ -0,0 +1,15 @@
...
+@curry
+def execute_plus_buffer(multiplier, state, transaction):
+    computation, _ = state.execute_transaction(transaction)
+
+    if computation.is_error:
+        raise computation._error
+
+    gas_used = transaction.gas_used_by(computation)
+
+    return int(gas_used * multiplier)
```
pipermerriam: Shouldn't this value be constrained by the gas limit?

evm/chains/chain.py
```diff
@@ -142,6 +146,15 @@ def create_header_from_parent(self, parent_header, **header_params):
             block_number=parent_header.block_number + 1,
         ).create_header_from_parent(parent_header, **header_params)
 
+    def set_gas_estimator(self, gas_estimator_func):
```
pipermerriam: I'm not sure this needs a setter.  I'd prefer this be something that is configured at the class level rather than the instance level.  In addition, I'd like it if it was configurable using environment variables (like the database backend).

Something like:

```python
class Chain:
    gas_estimation_backend = None

    def get_gas_estimation_backend(self):
        if self.gas_estimation_backend is None:
            # check whether a path has been declared using an environment variable.  If not, return the "default" backend.
            ...
        return self.gas_estimation_backend()
```

Whatcha think?

evm/vm/base.py
```diff
@@ -235,6 +236,13 @@ def pack_block(self, block, *args, **kwargs):
 
         return block
 
+    @contextlib.contextmanager
+    def state_in_temp_block(self):
+        header = self.block.header
+        temp_block = self.generate_block_from_parent_header_and_coinbase(header, header.coinbase)
+        prev_headers = [self.block.header] + self.prev_headers
+        yield self.get_state(block_header=temp_block.header, prev_headers=prev_headers)
```
carver: Do I need to do more cleanup here to undo changes that happen in the `with` context? I don't understand all the internals, yet.

carver: After another look, maybe it should be something like:

```py
state = self.get_state(block_header=temp_block.header, prev_headers=prev_headers)
snapshot = state.snapshot()
yield state
state.revert(snapshot)
```

pipermerriam: This looks closer to correct.  Assuming the context manager API for state is gone, this should be fine.  If it's not, then you need to nest the `yield` in it's own `with` block.

carver: I see a `state_db` context manager, but it's invoked several times inside the various VM execute transactions, with no way to pass a `read_only` flag. On a first-pass read-through of the code, `state.snapshot()` and `state.revert()` should reverse any of those changes, though.

pipermerriam: I believe you are correct.  I think @hwwhww was looking into removing the context manager API because it no longer makes sense with some of the recent changes, but that may not have happened yet.

evm/estimators/gas.py
```diff
@@ -0,0 +1,85 @@
...
+double_execution_cost = execute_plus_buffer(2)
+
+
+def _is_enough_gas(state, transaction):
+    snapshot = state.snapshot()
+    computation = state.execute_transaction(transaction)
+    state.revert(snapshot)
+
+    if computation.is_error:
+        return not isinstance(computation._error, OutOfGas)
```
pipermerriam: I'm not sure that we should only be checking for `OutOfGas` here.  I got annoyed that pyethereum only had one exception type (which is accurate in terms of VM semantics).  Hence, py-evm has fine grained exceptions for all the various ways that errors can blow up your VM execution.  However, as far as the EVM is concerned, every single `VMError` subclass is an out-of-gas error.

I think it's more appropriate to check against the base `VMError`.  I can think of at least one contract in the ethereum alarm clock contracts which would throw an `InvalidJumpDestination` error in py-evm if the gas allotment for the transaction was too low.

evm/estimators/gas.py
```diff
@@ -0,0 +1,85 @@
...
+    :return int: The smallest confirmed gas to not throw an OutOfGas exception,
+        subject to tolerance. If OutOfGas is thrown at block limit, return block limit.
+    """
+    minimum_transaction = SpoofTransaction(transaction, gas=transaction.intrinsic_gas)
+    if _is_enough_gas(state, minimum_transaction):
+        return transaction.intrinsic_gas
+
+    maximum_transaction = SpoofTransaction(transaction, gas=state.gas_limit)
+    if not _is_enough_gas(state, maximum_transaction):
+        return state.gas_limit
```
pipermerriam: I think this should be an error state that throws it's own exception.

evm/estimators/gas.py
```diff
@@ -0,0 +1,85 @@
...
+        if _is_enough_gas(state, test_transaction):
+            minimum_viable = midpoint
+        else:
+            maximum_out_of_gas = midpoint
+
+    return minimum_viable
+
+
+# Estimate in increments of intrinsic gas usage
+binary_search_intrinsic_tolerance = binary_search(21000)
```
pipermerriam: Maybe move the `tolerance` keyword to the end of the arguments so that you can do this as `binary_search(tolerance=21000)` to be a bit easier to see what argument is being curried at a glance?

tests/core/chain-object/test_chain.py
```diff
@@ -72,6 +76,52 @@ def test_empty_transaction_lookups(chain):
         chain.get_pending_transaction(b'\0' * 32)
 
 
+@pytest.mark.parametrize(
+    'data, gas_estimator, to, on_pending, expected',
```
pipermerriam: FYI, `pytest.mark.parametrize` can be stacked to produce a product of the two parameters

```
import pytest


@pytest.mark.parametrize('a', [1, 2])
@pytest.mark.parametrize('b', [3, 4])
def test_1(a, b):
    print(a, b)
```

This results in four test cases `[1, 3], [2, 3], [1, 4], [2, 4]`

evm/estimators/gas.py
```diff
@@ -0,0 +1,85 @@
...
+    state.revert(snapshot)
+
+    if computation.is_error:
+        return not isinstance(computation._error, OutOfGas)
+    else:
+        return True
+
+
+@curry
+def binary_search(tolerance, state, transaction):
```
pipermerriam: Maybe use slightly more explicit name of something like `binary_search_required_gas` or `binary_gas_search`

evm/estimators/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+)
+
+
+def get_gas_estimator(import_path=None):
+    if not import_path:
+        import_path = os.environ.get(
+            'GAS_ESTIMATOR_BACKEND_FUNC',
+            'evm.estimators.gas.binary_search_intrinsic_tolerance',
+        )
+    return import_string(import_path)
```
pipermerriam: Note to self:  this pattern is repeated at least one other place.  We should write a util in `evm.utils.env` to do the combination of pulling a value from the environment and then importing it.

carver: I thought about this a little more, and I can't come up with a time when we would want to pass in an `import_path` here (or in `get_db_backend()`). If we want to specify either one explicitly, we can just import and use it directly. `get_db_backend()` is never called with an argument in the code base, although it supports it. I think supporting an `import_path` here is a YAGNI feature, which means this just simplifies down to:

```py
def get_gas_estimator():
    import_path = os.environ.get(
        'GAS_ESTIMATOR_BACKEND_FUNC',
        'evm.estimators.gas.binary_search_intrinsic_tolerance',
    )
```

I'm not sure there's much left there to generalize.

evm/estimators/gas.py
carver: > I'm not sure that we should only be checking for `OutOfGas` here

In theory, I liked the idea that a throw or other kind of error would not cause this estimate to choose the block maximum.

>  I can think of at least one contract in the ethereum alarm clock contracts which would throw an InvalidJumpDestination error in py-evm if the gas allotment for the transaction was too low.

... which ruins the whole plan above. I think the best solution is to throw if there is an error at the block limit (as you mentioned in another comment), and let the caller decide what's appropriate to do. Then all later errors can be considered out-of-gas related (even if not `OutOfGas`), since the transaction has succeeded once, at the block limit.

tests/core/chain-object/test_chain.py
carver: This is a good reminder, I had forgotten about that. In this case, I actually don't want to run all the permutations, because some of these tests are pretty slow. I just want a sampling of them. Running one test at 1000 tolerance takes 1-2 seconds, for example.

pipermerriam: Only to clarify my original comment:  I was suggesting it for the `on_pending` parameter which if I recall, each test was duplicated with this as `True/False`.  Don't care if you actually change anything here.

carver:
I think that covers everything in the review @pipermerriam

evm/db/__init__.py
```diff
@@ -8,15 +8,14 @@
 DEFAULT_DB_BACKEND = 'evm.db.backends.memory.MemoryDB'
 
 
-def get_db_backend_class(import_path=None):
-    if import_path is None:
-        import_path = os.environ.get(
-            'CHAIN_DB_BACKEND_CLASS',
-            DEFAULT_DB_BACKEND,
-        )
+def get_db_backend_class():
```
pipermerriam: I am using the previous API in my trinity branch :(

carver:
> I am using the previous API in my trinity branch :(

Aw, sorry. YAGNI => YNI, this time...

> I can cleanup the db_backend thing in my branch.

Ok, that will save a test cycle, thanks.

> I'm open to different ways to do this if you have ideas.

I don't see where you use it in #300 yet, but I can guess.

The following is an alternative, but I don't really have any objections to adding the argument back. I just thought I was removing unused cruft.


```py
def infer_db_import_str():
    return os.environ.get(
        'CHAIN_DB_BACKEND_CLASS',
        DEFAULT_DB_BACKEND,
    )

Database = import_string(user_supplied_db_import_str or infer_db_import_str())
```

pipermerriam:
I lied, it's in a yet-to-be-opened-for-PR branch.

PR #285, opened 2018-01-18
Add PAYGAS opcode

NIC619:
### What was wrong?
Implementation of #234 

#### Add `PAYGAS` opcode
* Carry additional informations such as `transaction_gas_limit`, `paygas_gasprice ` and `paygas_total_fee` in `ShardingMessage` object
    * `transaction_gas_limit` is the same as `transaction.gas`
    * `paygas_gasprice ` is set by `PAYGAS` opcode and only under two conditions:
        1. in a top level call(i.e., `computation.msg.depth == 0`)
        2. the value has not been set before during this transaction
    * If the two conditions are not met, current action is to do nothing.
    * `paygas_total_fee` is equal to `transaction_gas_limit * paygas_gasprice`
* Once triggered, deduct `paygas_total_fee` from transaction initiator(`transaction.to`)'s balance
* Refund fee is calculated base on `paygas_gasprice` instead of `transaction.gas_price`
* Opcode value is currently set as ~~`0x5c`~~ `0xf5`, placed in ~~Stack, Memory, Storage and Flow Operations~~ Systems section. I preferred to put it in Environment Information section but it's full already.

- [x] 1. Remove coinbase from `execute_transaction` and ~transfer the fee directly to `PAYGAS_TEMPORARY_ACCOUNT` of which the account data is enforced to be permanently stored in database hence it's witness data is not required.~ store the tx fee amount due in `block.tx_fee_sum` and transfer all fees to coinbase in `finalize_block` function
- ~2. Add a function that generates a transaction to transfer fees collected from all transactions of this block to coinbase.~
- 3. Abstract `gas_price` away, removing it from transaction format.(Moved to another PR)
- 4. Add tests.(More to be added in another PR)

### DIFF
--- a/evm/exceptions.py
+++ b/evm/exceptions.py
@@ -166,7 +166,7 @@ class GasPriceAlreadySet(Exception):
     pass
 
 
-class NotTopLevelCall(VMError):
+class NotTopLevelCall(Exception):
     """
     Error raised if it's not executing in the top level call(i.e, msg.depth == 0)
     """
--- a/evm/exceptions.py
+++ b/evm/exceptions.py
@@ -164,3 +164,10 @@ class GasPriceAlreadySet(Exception):
     Error raised if trying to reset the already set gas price.
     """
     pass
+
+
+class NotTopLevelCall(VMError):
+    """
+    Error raised if it's not executing in the top level call(i.e, msg.depth == 0)
+    """
+    pass
--- a/evm/exceptions.py
+++ b/evm/exceptions.py
@@ -157,3 +157,10 @@ class UnannouncedStateAccess(VMError):
     been specified in the transaction's read or write list, respectively.
     """
     pass
+
+
+class GasPriceAlreadySet(Exception):
+    """
+    Error raised if trying to reset the already set gas price.
+    """
+    pass
--- a/evm/logic/call.py
+++ b/evm/logic/call.py
@@ -96,6 +96,7 @@ class BaseCall(Opcode):
                 'to': to,
                 'data': call_data,
                 'code': code,
+                'transaction_gas_limit': computation.msg.transaction_gas_limit,
                 'code_address': code_address,
                 'should_transfer_value': should_transfer_value,
                 'is_static': is_static,
--- a/evm/logic/call.py
+++ b/evm/logic/call.py
@@ -96,7 +96,6 @@ class BaseCall(Opcode):
                 'to': to,
                 'data': call_data,
                 'code': code,
-                'transaction_gas_limit': computation.msg.transaction_gas_limit,
                 'code_address': code_address,
                 'should_transfer_value': should_transfer_value,
                 'is_static': is_static,
@@ -343,3 +342,101 @@ class CallSharding(CallByzantium):
         transfer_gas_fee = constants.GAS_CALLVALUE if value else 0
         create_gas_fee = constants.GAS_NEWACCOUNT if not account_is_empty else 0
         return transfer_gas_fee + create_gas_fee
+
+    def __call__(self, computation):
+        computation.gas_meter.consume_gas(
+            self.gas_cost,
+            reason=self.mnemonic,
+        )
+
+        (
+            gas,
+            value,
+            to,
+            sender,
+            code_address,
+            memory_input_start_position,
+            memory_input_size,
+            memory_output_start_position,
+            memory_output_size,
+            should_transfer_value,
+            is_static,
+        ) = self.get_call_params(computation)
+
+        computation.extend_memory(memory_input_start_position, memory_input_size)
+        computation.extend_memory(memory_output_start_position, memory_output_size)
+
+        call_data = computation.memory.read(memory_input_start_position, memory_input_size)
+
+        #
+        # Message gas allocation and fees
+        #
+        child_msg_gas, child_msg_gas_fee = self.compute_msg_gas(computation, gas, to, value)
+        computation.gas_meter.consume_gas(child_msg_gas_fee, reason=self.mnemonic)
+
+        # Pre-call checks
+        with computation.state_db(read_only=True) as state_db:
+            sender_balance = state_db.get_balance(computation.msg.storage_address)
+
+        insufficient_funds = should_transfer_value and sender_balance < value
+        stack_too_deep = computation.msg.depth + 1 > constants.STACK_DEPTH_LIMIT
+
+        if insufficient_funds or stack_too_deep:
+            computation.return_data = b''
+            if insufficient_funds:
+                err_message = "Insufficient Funds: have: {0} | need: {1}".format(
+                    sender_balance,
+                    value,
+                )
+            elif stack_too_deep:
+                err_message = "Stack Limit Reached"
+            else:
+                raise Exception("Invariant: Unreachable code path")
+
+            self.logger.debug(
+                "%s failure: %s",
+                self.mnemonic,
+                err_message,
+            )
+            computation.gas_meter.return_gas(child_msg_gas)
+            computation.stack.push(0)
+        else:
+            with computation.state_db(read_only=True) as state_db:
+                if code_address:
+                    code = state_db.get_code(code_address)
+                else:
+                    code = state_db.get_code(to)
+
+            child_msg_kwargs = {
+                'gas': child_msg_gas,
+                'value': value,
+                'to': to,
+                'data': call_data,
+                'code': code,
+                'transaction_gas_limit': computation.msg.transaction_gas_limit,
+                'code_address': code_address,
+                'should_transfer_value': should_transfer_value,
+                'is_static': is_static,
+            }
+            if sender is not None:
+                child_msg_kwargs['sender'] = sender
+
+            child_msg = computation.prepare_child_message(**child_msg_kwargs)
+
+            child_computation = computation.apply_child_computation(child_msg)
+
+            if child_computation.is_error:
+                computation.stack.push(0)
+            else:
+                computation.stack.push(1)
+
+            if not child_computation.should_erase_return_data:
+                actual_output_size = min(memory_output_size, len(child_computation.output))
+                computation.memory.write(
+                    memory_output_start_position,
+                    actual_output_size,
+                    child_computation.output[:actual_output_size],
+                )
+
+            if not child_computation.should_burn_gas:
+                computation.gas_meter.return_gas(child_computation.gas_meter.gas_remaining)
--- a/evm/logic/system.py
+++ b/evm/logic/system.py
@@ -4,6 +4,7 @@ from evm.exceptions import (
     Halt,
     Revert,
     WriteProtection,
+    GasPriceAlreadySet,
 )
 
 from evm.opcode import (
@@ -241,4 +242,34 @@ class Create2(CreateEIP150):
 
 
 def paygas(computation):
-    raise NotImplementedError("PAYGAS opcode is not yet implemented")
+    gas_price = computation.stack.pop(type_hint=constants.UINT256)
+
+    # Only valid if (1) triggered in a top level call and
+    # (2) not been set already during this transaction execution
+    if computation.msg.depth == 0:
+        try:
+            computation.set_PAYGAS_gas_price(gas_price)
+        except GasPriceAlreadySet:
+            computation.stack.push(0)
+        else:
+            with computation.state_db(read_only=False) as state_db:
+                tx_initiator = computation.msg.to
+                tx_initiator_balance = state_db.get_balance(tx_initiator)
+
+                PAYGAS_gasprice = computation.get_PAYGAS_gas_price()
+                if PAYGAS_gasprice is None:
+                    PAYGAS_gasprice = 0
+                fee_to_be_charged = PAYGAS_gasprice * computation.msg.transaction_gas_limit
+
+                if tx_initiator_balance < fee_to_be_charged:
+                    raise InsufficientFunds(
+                        "Insufficient funds: {0} < {1}".format(
+                            tx_initiator_balance,
+                            fee_to_be_charged
+                        )
+                    )
+
+                state_db.delta_balance(tx_initiator, -1 * fee_to_be_charged)
+            computation.stack.push(1)
+    else:
+        computation.stack.push(0)
--- a/evm/logic/system.py
+++ b/evm/logic/system.py
@@ -5,6 +5,7 @@ from evm.exceptions import (
     Revert,
     WriteProtection,
     GasPriceAlreadySet,
+    InsufficientFunds,
 )
 
 from evm.opcode import (
@@ -229,6 +230,7 @@ class Create2(CreateEIP150):
             value=value,
             data=b'',
             code=call_data,
+            transaction_gas_limit=computation.msg.transaction_gas_limit,
             is_create=True,
         )
 
--- a/evm/logic/system.py
+++ b/evm/logic/system.py
@@ -250,7 +250,7 @@ def paygas(computation):
     # (2) not been set already during this transaction execution
     if computation.msg.depth == 0:
         try:
-            computation.set_PAYGAS_gas_price(gas_price)
+            computation.set_PAYGAS_gasprice(gas_price)
         except GasPriceAlreadySet:
             computation.stack.push(0)
         else:
--- a/evm/logic/system.py
+++ b/evm/logic/system.py
@@ -4,8 +4,9 @@ from evm.exceptions import (
     Halt,
     Revert,
     WriteProtection,
-    GasPriceAlreadySet,
     InsufficientFunds,
+    GasPriceAlreadySet,
+    NotTopLevelCall
 )
 
 from evm.opcode import (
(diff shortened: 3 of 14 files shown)


NIC619:
@vbuterin does the `PAYGAS_TEMPORARY_ACCOUNT` scheme in WIP item.1 and 2. makes sense?

evm/logic/context.py
```diff
@@ -193,3 +194,26 @@ def returndatacopy(computation):
 
 def sighash(computation):
     computation.stack.push(computation.msg.sig_hash)
+
+
+def paygas(computation):
+    gasprice = computation.stack.pop(type_hint=constants.UINT256)
```
pipermerriam: *naming nitpick*

`gas_price` for conformity.

evm/logic/context.py
```diff
@@ -193,3 +194,26 @@ def returndatacopy(computation):
...
+
+    # Only valid if (1) triggered in a top level call and
+    # (2) not been set already during this transaction execution
+    if computation.msg.depth == 0 and computation.msg.paygas_gasprice == 0:
```
pipermerriam: `computation.msg.paygas_gasprice == 0` seems problematic because it disallows a gas price of zero which at least historically, has been allowed.  Maybe use `None` as the *not-set* value?

evm/vm/message.py
```diff
@@ -187,12 +193,29 @@ def __init__(self,
...
+
+    @property
+    def paygas_gasprice(self):
+        if self._paygas_gasprice is not None:
+            return self._paygas_gasprice
+        else:
+            return 0
+
+    @paygas_gasprice.setter
+    def paygas_gasprice(self, gas_price):
```
pipermerriam: probably good to add validation for this value here to be **really** sure that it's valid.

evm/vm/message.py
```diff
@@ -187,12 +193,29 @@ def __init__(self,
...
+        if self.paygas_gasprice == 0:
```
pipermerriam: It seems like there should be a logic branch here which throws an exception (not a `VMException` but something that will bubble) if the value is already set, because if I understand correctly, we shouldn't even try to set the value if it's already set.

evm/vm/message.py
```diff
@@ -144,6 +144,8 @@ def is_create(self):
 class ShardingMessage(Message):
 
     is_create = False
+    transaction_gas_limit = None
```
pipermerriam: Thoughts on removing the `transaction_` prefix here?  I'm not sure if my thinking is accurate, but historically, the `Message` object has been intentionally naive about the transaction.  Would it still be accurate if this was simply titled `gas_limit`?

Also, how does do `gas_limit` and `gas` differ?  Now that I'm asking makes me think that this is in-fact appropriately prefixed with `transaction_` assuming that's the difference, one is the message gas, the other is the outer transaction gas limit.

NIC619: Yes, you are correct. One is the outer transaction gas limit and one is message gas. And hence the prefix to tell the difference between them.

evm/vm/message.py
NIC619: What about returning `True` of `False` in `paygas_gasprice.setter` to indicate if the value is successfully set or not.
```python
if self.paygas_gasprice is None:
    self._paygas_gasprice = gas_price
    return True
else:
    return False
```

and in `PAYGAS` opcode check against this returned value and push `1` or `0` to the stack to indicate the outcome of this `PAYGAS` operation?
```python
def paygas(computation):
    ...
    if computation.msg.depth == 0 and computation.msg.paygas_gasprice(gas_price):
        …
        computation.stack.push(1)
    else:
        computation.stack.push(0)
```

It's like fail of value transfer in a call. It seems better to me that it does not result in failure of transaction but I'm not that confident about this. Could it be the situation that gas were paid in top level call but later during some calls to other contract, the contract trigger `PAYGAS` again?

vbuterin: I like pushing 1 on success and 0 on failure.

pipermerriam: So my suggestion was not focused on VM semantics, but rather our python API semantics.  Here's the situation I'm trying to avoid.

```python
# somewhere we set the gas price.
msg.paygas_gasprice = 10

# elsewhere in the code, we try to set it again.
msg.paygas_price = 100  # this will silently fail
```

What I would like is that the later case cannot silently fail, hence, suggesting that we raise an exception.  Then, if we ever have a location in the code which tries to set the gas price when it's already set which doesn't explicitely handle whatever exception we raise, then we'll get a hard failure in our code, rather than a silent failure that only manifests later at some point when the gas accounting is off.

pipermerriam: @NIC619 

Look here: https://github.com/ethereum/py-evm/blob/master/evm/computation.py#L316-L338

Anything inside of vm execution which throws any exception that is a subclass of `VMError` gets handled there, resulting in `computation.is_error` being `True`.

What I'm proposing is something that throws something like a `TypeError` or another custom exception that **is not** a `VMError` subclass.  This way, if we write code which sets `msg.paygas_gasprice` without implementing error handling **and** the `msg.paygas_gasprice` has already been set... an error will be thrown in such a way that it won't be caught in the linked code above.

Does that make sense?

NIC619:
@pipermerriam Made some updates. 

Now the exception `PAYGASResetGasprice` will be raised if trying to reset the already set `msg.paygas_gasprice`. And in `PAYGAS` opcode it would use try/except to check if `msg.paygas_gasprice` has been set already. Push `1` to the stack if not and `0` otherwise.

evm/exceptions.py
```diff
@@ -150,3 +150,7 @@ class UnannouncedStateAccess(VMError):
     been specified in the transaction's read or write list, respectively.
     """
     pass
+
+
+class PAYGASResetGasprice(Exception):
```
pipermerriam: other name suggestions:

- `GasPriceAlreadySet`
- `GasPriceOverwriteProtection`

I'm thinking the inclusing of `PAYGAS` in the name isn't needed and is a little visually confusing.

evm/logic/context.py
```diff
@@ -193,3 +195,30 @@ def returndatacopy(computation):
...
 def sighash(computation):
     computation.stack.push(computation.msg.sig_hash)
+
+
+def paygas(computation):
+    gas_price = computation.stack.pop(type_hint=constants.UINT256)
+
+    # Only valid if (1) triggered in a top level call and
+    # (2) not been set already during this transaction execution
+    if computation.msg.depth == 0:
```
pipermerriam: The way this is written without an `else` clause this does nothing if triggered from `depth > 0`.  Is this intended?  Should it push `0` onto the stack if `depth > 0`?

evm/logic/context.py
```diff
@@ -193,3 +195,30 @@ def returndatacopy(computation):
...
+        try:
+            computation.msg.paygas_gasprice(gas_price)
```
pipermerriam: This looks like an error.  You are calling a property which I believe should raise an exception stating that either `NoneType` or the integers are not callable.  

I do however like using a setter function rather than assigning to a property that throws an exception.  Maybe rename to something like `computation.msg.set_PAYGAS_gasprice(...)`?

evm/vm/message.py
```diff
@@ -187,12 +196,34 @@ def __init__(self,
...
+
+    @property
+    def paygas_gasprice(self):
+        if self._paygas_gasprice is None:
+            return 0
+        else:
+            return self._paygas_gasprice
+
+    @paygas_gasprice.setter
+    def paygas_gasprice(self, gas_price):
```
pipermerriam: Continuation of previous suggestion to remove the `setter` and just make this a method `set_PAYGAS_gasprice`.  Seems a bit more intuitive/correct/something.

evm/vm/message.py
```diff
@@ -187,12 +196,34 @@ def __init__(self,
...
         self.access_list = access_list
 
+    @property
+    def paygas_total_fee(self):
+        return self.paygas_gasprice * self.transaction_gas_limit
+
+    @property
+    def paygas_gasprice(self):
+        if self._paygas_gasprice is None:
+            return 0
```
pipermerriam: I'm still curious about returning `0` instead of `None` here.  It seems valuable to be able to distinguish between *not-set* and *set-to-0* which cannot be done in the current implementation.

evm/logic/context.py
NIC619: My mistake here. I will update the setter to `set_PAYGAS_gasprice` function.

evm/vm/message.py
NIC619: My rationale was that from inside(`Message` object) the two cases are distinguished by `msg._paygas_gasprice` and from outside one indeed can not tell the difference. But since when `msg.paygas_gasprice` is used to compute gas charge/refund (from outside) the two cases are basically the same.

Also if `msg.paygas_gasprice` could return `None`, we need to add a condition check every time it's used to compute gas charge/refund.
```python
gas_refund_amount = (gas_refund + gas_remaining) * computation.msg.paygas_gasprice if computation.msg.paygas_gasprice is not None else 0
```

evm/logic/context.py
NIC619: You are right. It should push `0` onto the stack if `depth > 0`.

evm/vm/message.py
pipermerriam: So i'm on the fence about this.  Your call.  Here's some extra thoughts on why it *might* still be good to use `None`.

The `Computation` object has an existing convention of having `get_XXX` methods for things to account for these little bits of internal logic.  I think the following API is consisten with the rest of the class and clean.

```python
class Computation:
    ...
    def get_PAYGAS_gas_price(self):
        if self._paygas_gasprice is None:
            return 0
        else:
            return self._paygas_gasprice

# and then used elsewhere
gas_refund_amount = (gas_refund + gas_remaining) * computation.get_paygas_gasprice()
```

It also has the nice benefit of not reaching down into the message object.

Again, your call.  I'm +0 on this change if you don't think there is a valid use case for differentiating between set to 0 and unset.

evm/logic/context.py
pipermerriam: Do you think it's worth implementing a few minimal test cases for this opcode to hit these code paths?  I know the JSON test suite will cover these eventually, but

- we'd probably benefit from having basic test cases for opcodes that were not buried in the hard to decipher JSON fixture tests.
- until we have JSON fixture tests for this it'd be nice to know it works (from a testing/CI perspective)

NIC619:
Since this had to be rebased and some part of this PR is already been taken cared of in other new PR, I just re-do it along with PR feedbacks.

I changed it back so that `paygas_gasprice` will return `None` if not set yet because the other case doesn't seem to have much advantage over this one.

I moved `paygas_garprice` and it's getter/setter to `Computation`. But at this moment of writing I feel like I might have misunderstood what you mean. @pipermerriam 
Do you mean (1)keep the `paygas_gasprice` in `Message` and use getter/setter in `Computation` to get/set this value, or (2)move `paygas_gasprice` and it's getter/setter together to `Computation`?

NIC619:
Another thing is transferring the transaction fee to `coinbase`. I was thinking about hooking this information on to `computation` object(`computation.tx_fee`) since it's what's returned after executing transaction. Then add the fee information to `Block.tx_fee_sum` and finally transfer all the fees collected to `coinbase` in `finalize_block`.

```python
def execute_transaction(self, transaction):
    …
    transaction_fee = (transaction.gas - gas_remaining - gas_refund) * PAYGAS_gasprice
    computation.tx_fee = transaction_fee
    …

def add_transaction(self, transaction, computation, block):
    …
    if block.tx_fee_sum:
        block.tx_fee_sum += computation.tx_fee
    else:
        block.tx_fee_sum = computation.tx_fee
    …

def finalize_block(self, block):
    …
    with self.state_db() as state_db:
            state_db.delta_balance(block.header.coinbase, block.tx_fee_sum)
            self.logger.debug(
                "TOTAL TRANSACTON FEE: %s -> %s",
                block.tx_fee_sum,
                block.header.coinbase,
            )
    …
```

evm/vm/forks/sharding/computation.py
```diff
@@ -28,9 +29,24 @@
...
 class ShardingComputation(SpuriousDragonComputation):
+    _paygas_gasprice = None
+
     # Override
     opcodes = SHARDING_OPCODES
 
+    def get_PAYGAS_gas_price(self):
+        return self._paygas_gasprice
+
+    def set_PAYGAS_gasprice(self, gas_price):
```
pipermerriam: I think that this is a good move (moving the value and the setters co `Computation`).  This function could use one extra validation check to enforce `depth == 0` within the setter itself.

NIC619: @pipermerriam 
Added the `depth == 0` validation check along with PAYGAS testing contract fixture and test suite.

hwwhww:
According to our last discussion, the `GAS_PAYGAS` is `3`, please update it.

evm/vm/forks/sharding/vm_state.py
```diff
@@ -160,14 +167,12 @@ def execute_transaction(self, transaction):
...
+        transaction_fee = (transaction.gas - gas_remaining - gas_refund) * PAYGAS_gasprice
         self.logger.debug(
-            'TRANSACTION FEE: %s -> %s',
+            'TRANSACTION FEE: %s',
             transaction_fee,
-            encode_hex(self.coinbase),
         )
-        with state_db_cm() as state_db:
-            state_db.delta_balance(self.coinbase, transaction_fee)
+        computation.tx_fee = transaction_fee
```
pipermerriam: Rather than storing this value on the computation object what do you think about just re-calculating it within `add_transaction`?  It looks like we have all of the necessary information already available in the context of that function to do the computations.

tests/core/helpers.py
```diff
@@ -32,7 +32,8 @@ def new_sharding_transaction(
 
     [destination, value, msg_data, vrs].
     """
-    tx_data = pad32(data_destination) + pad32(bytes([data_value])) + pad32(data_msgdata) + pad32(data_vrs)  # noqa: E501
+    assert len(data_vrs) in (0, 65)
+    tx_data = pad32(data_destination) + pad32(bytes([data_value])) + pad32(data_msgdata) + pad32(data_vrs[:32]) + pad32(data_vrs[32:64]) + bytes(data_vrs[64:])  # noqa: E501
```
pipermerriam: suggest reformatting to 

```python
tx_data = (
    pad32(data_destination) +
    pad32(bytes([data_value])) +
    pad32(data_msgdata) +
    pad32(data_vrs[:32]) +
    pad32(data_vrs[32:64]) +
    bytes(data_vrs[64:]
)
```

evm/logic/system.py
```diff
@@ -228,6 +230,7 @@ def __call__(self, computation):
             value=value,
             data=b'',
             code=call_data,
+            transaction_gas_limit=computation.msg.transaction_gas_limit,
```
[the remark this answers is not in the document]
NIC619: Will do. 👍

evm/vm/forks/sharding/vm_state.py
NIC619: You are right. Totally makes sense.

evm/logic/call.py
```diff
@@ -342,3 +342,101 @@ def compute_msg_extra_gas(self, computation, gas, to, value):
...
+
+            if not child_computation.should_erase_return_data:
+                actual_output_size = min(memory_output_size, len(child_computation.output))
+                computation.memory.write(
+                    memory_output_start_position,
+                    actual_output_size,
+                    child_computation.output[:actual_output_size],
+                )
+
+            if not child_computation.should_burn_gas:
```
pipermerriam: Note to self.

this is a confusing `if` statement because of the need for the `not`.  We might make a second property which returns the inverse of this as `computation.should_return_gas`.

evm/vm/forks/sharding/vm_state.py
```diff
@@ -185,3 +189,97 @@ def execute_transaction(self, transaction):
...
+
+        # Calculate gas price
+        tx_PAYGAS_gasprice = computation.get_PAYGAS_gas_price()
+        if tx_PAYGAS_gasprice is None:
+            tx_PAYGAS_gasprice = 0
+        # Calculate gas usage
+        gas_remaining = computation.get_gas_remaining()
+        gas_refunded = computation.get_gas_refund()
+        gas_used = transaction.gas - gas_remaining
+        gas_refund = min(gas_refunded, gas_used // 2)
```
pipermerriam: Hrm, rather than repeating this block of code both here and above, I think we should dig it out into a utility of some sort.

```python
def compute_transaction_gas_amounts(computation, transaction_gas, gas_price):
    gas_remaining = computation.get_gas_remaining()
    gas_refunded = computation.get_gas_refund()
    gas_used = transaction_gas - gas_remaining
    gas_refund = min(gas_refunded, gas_used // 2)
    tx_fee = (transaction_gas - gas_remaining - gas_refund) * gas_price
    return tx_fee, gas_used, gas_refund
```

Something like this.  Could also be implemented as a method on the `Computation` class which would allow removing the `gas_price` argument from the call signature.

evm/vm/forks/sharding/vm_state.py
```diff
@@ -185,3 +189,97 @@ def execute_transaction(self, transaction):
...
+        # Calculate transaction fee
+        tx_fee = (transaction.gas - gas_remaining - gas_refund) * tx_PAYGAS_gasprice
+        # Bookkeep this transaction fee
+        if hasattr(block, "tx_fee_sum"):
+            block.tx_fee_sum += tx_fee
```
pipermerriam: I'd like to figure out a way to do this which doesn't involve storing data on the block object.  Looking around at the code however, it's not very clear where that would be.

For now, lets just formalize this property on the block object so that we don't have to do `hasattr` checks.

evm/logic/call.py
NIC619: Noted. Will make a PR to update this.

Issue #257, opened 2018-01-09
Implement `Collation` class


hwwhww:
### What is wrong?
See [sharding doc](https://github.com/ethereum/sharding/blob/develop/docs/doc.md#collation-header). 

Note that some fields are renamed.

### How can it be fixed
We can reuse https://github.com/ethereum/py-evm/pull/172.
Since the `Block` and `BlockHeader` RLP object are much simpler now, I guess it's easier to implement the simple `Collation` and `CollationHeader` classes.

hwwhww:
I just recalled that I forgot to add `CollationHeader.number` field in doc.md last year. Let's review again in the current paradigm:

Currently, the sequential number of collation is exactly the score of collation.

Should we add `number` field in `CollationHeader`?

The following evaluation is made in September last year with `pyethereum` structure.

Cons
1. Taking more space for storing this one new field.
2. The cost of switching shard branch would increase by additional data management:
if we want to use a function like `ShardChain.get_collation_by_number(collation_number)`, it may takes more computational time upon receiving a new block and reorganization the collation head since we have to reorganize database values like: `self.db.put(b'coll:%d' % collation.header.number, collation.header.hash)`
    * ^^^^^ Won't be an issue for stateless client shuffling.

Pros
1. It’s easier to delete old junk collations in database with `CollationHeader.number` since it’s easier to locate the old collation by their number.
2. It’s easier to peek the score of an orphan collation with `CollationHeader.number`:
it would help the shard validators to determine if they fell behind when the given collation’s parent is not in the database, and then they can ask for the parent collations. Otherwise, the shard validators have to call validator manager contract to check the score after they received the block which includes the given collation.

Look over again, I think it's still worth to add `number` field for archival nodes.

hwwhww:
closed via #288

PR #288, opened 2018-01-18
Add Collation and CollationHeader classes

jannikluhn:
This pull request adds a Collation and CollationHeader classes. I used the Block and BlockHeader classes in the current master as a model, so they are very slim. Only functionality added is the ability to get witnesses from `BaseChainDB` (using `py-trie`s `get_witness_for_key_prefix`) which is needed for `Collation.from_header`. I wrote a test for this, which unfortunately fails. I believe the reason for this is that `get_witness_for_key_prefix` has been written for `BinaryTrie` instead of `HexaryTrie` (which is still in use), but would appreciate feedback by @NIC619 on that matter. If this is confirmed I'd mark it as `xfail`ing (as switching to `BinaryTrie` should be done in another PR).

@hwwhww Is there anything missing?

#### Cute Animal Picture

### DIFF
--- a/evm/rlp/headers.py
+++ b/evm/rlp/headers.py
@@ -193,7 +193,7 @@ class CollationHeader(rlp.Serializable):
     def from_parent(cls,
                     parent,
                     period_start_prevhash,
-                    expected_period_number=None,
+                    expected_period_number,
                     coinbase=ZERO_ADDRESS,
                     sig=b""):
         """
@@ -202,7 +202,7 @@ class CollationHeader(rlp.Serializable):
         """
         header_kwargs = {
             "shard_id": parent.shard_id,
-            "expected_period_number": expected_period_number or parent.expected_period_number + 1,
+            "expected_period_number": expected_period_number,
             "period_start_prevhash": period_start_prevhash,
             "parent_hash": parent.hash,
             "state_root": parent.state_root,
--- a/evm/rlp/headers.py
+++ b/evm/rlp/headers.py
@@ -141,3 +141,76 @@ class BlockHeader(rlp.Serializable):
             for field_name
             in tuple(zip(*self.fields))[0]
         })
+
+
+class CollationHeader(rlp.Serializable):
+    fields = [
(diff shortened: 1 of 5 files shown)


evm/rlp/headers.py
```diff
@@ -133,3 +133,76 @@ def from_parent(cls,
...
+
+
+class CollationHeader(rlp.Serializable):
+    fields = [
+        ("shard_id", big_endian_int),
+        ("expected_period_number", big_endian_int),
+        ("period_start_prevhash", hash32),
+        ("parent_hash", hash32),
+        ("coinbase", address),
+        ("transaction_root", hash32),
```
hwwhww: The order of `coinbase` and `transaction_root` is wrong. It needs to follow spec for RLP objects.

evm/rlp/headers.py
```diff
@@ -133,3 +133,76 @@ def from_parent(cls,
...
+                    expected_period_number=None,
+                    coinbase=ZERO_ADDRESS,
+                    sig=b""):
+        """
+        Initialize a new collation header with the `parent` header as the collation's
+        parent hash.
+        """
+        header_kwargs = {
+            "shard_id": parent.shard_id,
+            "expected_period_number": expected_period_number or parent.expected_period_number + 1,
```
hwwhww: `parent.expected_period_number + 1` is not true for all circumstances. There may be no collation in one period. I'd say that the input parameter `expected_period_number` can't be `None`.

hwwhww:
According to https://github.com/ethereum/py-evm/issues/257#issuecomment-358889420, the `number` field is added: https://github.com/ethereum/sharding/blob/develop/docs/doc.md#collation-header

Please add it to align with VMC, thanks!

jannikluhn:
This looks finished to me. Only issue is it requires #274 which is merged in master but not yet in sharding. Shall I remove the `Configurable` dependency or are we going to rebase?

hwwhww:
Something is missing in the tests, looks like you have to replace `FlatTrieBackend`?
Rebase +1

PR #290, opened 2018-01-20
Code cleanup of evm.p2p with pylint

h00701350103:
I was gonna work on the static typing issue to get acquainted with the codebase, but started playing around with pylint and couldn't help myself upon seeing sub-perfect code. I've done a few changes I think are strict improvements on the code, but I've held up on more controversial or bigger-scoped changes, instead writing them down with my thoughts on them. If OK'd I'll [attempt] fixing them.
Comments are not necessarily exhaustive for modules, my investigations have been semi-random.

### Module evm.p2p.test_peer
`W: 93, 4: Unused variable 'peer1_ephemeral_pubkey' (unused-variable)`
`W: 93,28: Unused variable 'peer1_nonce' (unused-variable)`

These look completely unused, but I don't know if they *should* be used or can be straight-up removed.

### Module evm.p2p.test_kademlia
W:222,11: Using possibly undefined loop variable 'i' (undefined-loop-variable)`

```python
for i in range(1000):
    assert table.add_node(random_node()) is None
assert i == len(table) - 1 
```
Unclear if misplaced assert (which surprised me to see it working at all, TIL), or intentionally optimizing to only check length when finished.
If that's the case, should IMO be written by specifying length as a variable that's used in the range(), and then checked against with the assert to be clear about intent.
I moved the length assert into the loop for now.

### Module evm.p2p.discovery
`C:295, 0: Line too long (180/100) (line-too-long)`
As mentioned in gitter there are a couple of too long lines with bytestrings containing long urls and pylint doesn't respect flake8's notation for not complaining.
I think a cleaner solution is to use multiline strings, but will refrain from changing as that's probably been considered when playing with flake8.

`W: 55,34: Redefining built-in 'id' (redefined-builtin)`
id is used as a parameter name which conflicts with the built-in function of the same name. Should maybe be named index or similar, unclear if you also want to rename the class variable for consistency as it doesn't directly collide with the built-in.

`W:122,26: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)`
https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation
Holding off on changing this one as it's mostly a question of optimization, it's sorta ugly, and pylint starts complaining with logging-format-truncated if you actually follow the given standard so would require disabling.

### Module evm.p2p.exceptions
`W: 55, 4: __init__ method from base class 'Exception' is not called (super-init-not-called)`
While I don't think this is an actual error, rewriting it to call super saves us from having to override __str__. I'm not super comfortable with enums+exceptions+inheritance, so might be wrong on this one.

### Module evm.p2p.kademlia
`R:560, 4: Method could be a function (no-self-use)`
This method should probably be moved outside the class, but holding off on having to change all calls to it until confirmed.

### Module evm.p2p.les
`C:235,11: Do not use `len(SEQUENCE)` to determine if a sequence is empty (len-as-condition)`
took me a minute to distinguish that 'proofs' and 'proof' were different. Fixed the len() warning but this code doesn't look very readable at all and could probably warrant a look by somebody that knows what it
should do.

While pylint is crazy verbal and complain about pretty much everything, every time I start running it against code I learn new stuff about the proper way to do various different things and this pass definitely catched some broken stuff (mostly missing asserts). Might be worth adding to tox if disabling a bunch of warnings that aren't useful.

#### Cute Animal Picture

This is Sixten, one of my two pet rats :3

### DIFF
--- a/evm/p2p/auth.py
+++ b/evm/p2p/auth.py
@@ -2,9 +2,10 @@ import asyncio
 import logging
 import os
 import random
-import sha3
 import struct
 from typing import Tuple
+
+import sha3
 
 import rlp
 from rlp import sedes
--- a/evm/p2p/auth.py
+++ b/evm/p2p/auth.py
@@ -160,9 +160,9 @@ class HandshakeInitiator(HandshakeBase):
         if len(ciphertext) < ENCRYPTED_AUTH_ACK_LEN:
             raise ValueError("Auth ack msg too short: {}".format(len(ciphertext)))
         elif len(ciphertext) == ENCRYPTED_AUTH_ACK_LEN:
-            eph_pubkey, nonce, version = decode_ack_plain(ciphertext, self.privkey)
+            eph_pubkey, nonce, __ = decode_ack_plain(ciphertext, self.privkey)
         else:
-            eph_pubkey, nonce, version = decode_ack_eip8(ciphertext, self.privkey)
+            eph_pubkey, nonce, __ = decode_ack_eip8(ciphertext, self.privkey)
             self.got_eip8_auth = True
         return eph_pubkey, nonce
 
@@ -202,10 +202,10 @@ class HandshakeResponder(HandshakeBase):
         if len(ciphertext) < ENCRYPTED_AUTH_MSG_LEN:
             raise ValueError("Auth msg too short: {}".format(len(ciphertext)))
         elif len(ciphertext) == ENCRYPTED_AUTH_MSG_LEN:
-            sig, initiator_pubkey, initiator_nonce, version = decode_auth_plain(
+            sig, initiator_pubkey, initiator_nonce, __ = decode_auth_plain(
                 ciphertext, self.privkey)
         else:
-            sig, initiator_pubkey, initiator_nonce, version = decode_auth_eip8(
+            sig, initiator_pubkey, initiator_nonce, __ = decode_auth_eip8(
                 ciphertext, self.privkey)
             self.got_eip8_auth = True
 
--- a/evm/p2p/auth.py
+++ b/evm/p2p/auth.py
@@ -161,9 +161,9 @@ class HandshakeInitiator(HandshakeBase):
         if len(ciphertext) < ENCRYPTED_AUTH_ACK_LEN:
             raise ValueError("Auth ack msg too short: {}".format(len(ciphertext)))
         elif len(ciphertext) == ENCRYPTED_AUTH_ACK_LEN:
-            eph_pubkey, nonce, __ = decode_ack_plain(ciphertext, self.privkey)
+            eph_pubkey, nonce, _ = decode_ack_plain(ciphertext, self.privkey)
         else:
-            eph_pubkey, nonce, __ = decode_ack_eip8(ciphertext, self.privkey)
+            eph_pubkey, nonce, _ = decode_ack_eip8(ciphertext, self.privkey)
             self.got_eip8_auth = True
         return eph_pubkey, nonce
 
@@ -203,10 +203,10 @@ class HandshakeResponder(HandshakeBase):
         if len(ciphertext) < ENCRYPTED_AUTH_MSG_LEN:
             raise ValueError("Auth msg too short: {}".format(len(ciphertext)))
         elif len(ciphertext) == ENCRYPTED_AUTH_MSG_LEN:
-            sig, initiator_pubkey, initiator_nonce, __ = decode_auth_plain(
+            sig, initiator_pubkey, initiator_nonce, _ = decode_auth_plain(
                 ciphertext, self.privkey)
         else:
-            sig, initiator_pubkey, initiator_nonce, __ = decode_auth_eip8(
+            sig, initiator_pubkey, initiator_nonce, _ = decode_auth_eip8(
                 ciphertext, self.privkey)
             self.got_eip8_auth = True
 
--- a/evm/p2p/discovery.py
+++ b/evm/p2p/discovery.py
@@ -276,7 +276,7 @@ def _unpack(message: AnyStr) -> Tuple[datatypes.PublicKey, int, List[Any], AnySt
     return remote_pubkey, cmd_id, payload, message_hash
 
 
-if __name__ == "__main__":
+def main():
     async def show_tasks():
         while True:
             tasks = []
@@ -331,3 +331,7 @@ if __name__ == "__main__":
     discovery.stop()
     # logger.info("Pending tasks at exit: {}".format(asyncio.Task.all_tasks(loop)))
     loop.close()
+
+
+if __name__ == "__main__":
+    main()
--- a/evm/p2p/discovery.py
+++ b/evm/p2p/discovery.py
@@ -119,7 +119,7 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
             # FIXME: Instead of sleeping here to wait until connection_made() is called to set
             # .transport we should instead only call it after we know it's been set.
             await asyncio.sleep(1)
-        self.logger.debug("boostrapping with {}".format(self.bootstrap_nodes))
+        self.logger.debug("boostrapping with %s", self.bootstrap_nodes)
         await self.kademlia.bootstrap(self.bootstrap_nodes)
 
     # FIXME: Enable type checking here once we have a mypy version that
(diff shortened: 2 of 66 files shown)


evm/p2p/kademlia.py
```diff
@@ -516,7 +516,7 @@ def ping(self, node: Node) -> bytes:
         async def _find_node(node_id, remote):
             self.wire.send_find_node(remote, node_id)
             candidates = await self.wait_neighbours(remote)
-            if len(candidates) == 0:
+            if not candidates:
```
gsalgado: I'm not sure I agree with pylint here; the original check makes it clear that we expect candidates to be of type sequence, so we'll get an exception if by accident we get a different type whereas the new check may fail silently if we get something unexpected here.

evm/p2p/les.py
```diff
@@ -232,7 +232,7 @@ def decode_payload(self, rlp_data: bytes) -> _DecodedMsgType:
         # This is just to make Proofs messages compatible with ProofsV2, so that LightChain
         # doesn't have to special-case them. Soon we should be able to drop support for LES/1
         # anyway, and then all this code will go away.
-        if len(decoded['proofs']) == 0:
+        if not decoded['proofs']:
```
gsalgado: Ditto

gsalgado:
> W:122,26: Use % formatting in logging functions and pass the % parameters as arguments (logging-format-interpolation)
These should be changed. I've updated a bunch of them but there are many left still

gsalgado:
> W: 93, 4: Unused variable 'peer1_ephemeral_pubkey' (unused-variable)
> W: 93,28: Unused variable 'peer1_nonce' (unused-variable)
> These look completely unused, but I don't know if they should be used or can be straight-up removed.

We could remove line 93 completely, but it might be better to just discard its return value and add a comment explaining we're keeping it to ensure the responder can decode the auth msg. Ideally we'd assert something on the return values but that's not possible so we have to rely only on the fact that `responder.decode_authentication()` does not raise any exceptions

evm/p2p/discovery.py
```diff
@@ -276,7 +276,7 @@ def _unpack(message: AnyStr) -> Tuple[datatypes.PublicKey, int, List[Any], AnySt
     return remote_pubkey, cmd_id, payload, message_hash
 
 
-if __name__ == "__main__":
+def main():
```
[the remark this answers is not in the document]
h00701350103: The actual problem being fixed here is that all variables declared in the "main function" (before I made it a function) become global variables and are reachable by all functions being called from it. The actual warnings from pylint was that every single variable declared didn't conform to UPPER_CASE naming style (because they're global), and that many other random functions that used the same variable names got "Redefining name <name> from outer scope".
If we want to indicate that the main function shouldn't be called I can name it something less inviting, but if we really don't want the code in a function the variables would need to be named __PRIVKEY or similar to hope that other functions don't use them.
Could move the main function into a separate file such that you'd have to explicitly import it (and place a comment in it that it's not to be imported and used) and don't automatically get it when importing discovery.py

evm/p2p/exceptions.py
```diff
@@ -53,7 +53,4 @@ class StopRequested(Exception):
 class HandshakeFailure(Exception):
 
     def __init__(self, reason: 'DisconnectReason') -> None:
-        self.reason = reason
-
-    def __str__(self):
-        return self.reason.name
+        super().__init__(reason.name)
```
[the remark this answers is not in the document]
h00701350103: okay, I'll change back so it's still saved in the object, but keep the super().__init__ call instead of the __str__ override.

evm/p2p/kademlia.py
h00701350103: If we're keeping it for type safety, wouldn't it then be even better to do
```python
if candidates == []:
```
To nail down the type completely and avoid all accidents? with len() == 0 it'll work with dicts, strings, etc.

The argument for not checking type at all is that this code is already covered with static type checks, so changing return type will fail mypy tests, and it is redundant to check the type several times.

evm/p2p/discovery.py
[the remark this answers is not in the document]
gsalgado: Well, all of this is only a problem if we intend to run pylint as part of our CI checks, right? Is there actually much value in running pylint as well as flake8?

evm/p2p/kademlia.py
gsalgado: I seem to remember working on projects that preferred explicit length checks instead of relying on the fact that empty sequences are false, and I also personally prefer that, but it looks like PEP-8 recommends otherwise[1], so it's probably better if we stick to it

[1] For sequences, (strings, lists, tuples), use the fact that empty sequences are false.

evm/p2p/discovery.py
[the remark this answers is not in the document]
h00701350103: Isn't it a problem that the variables are global? It would e.g. make it easy to forget to define the variable yourself in an underlying function and instead accidentally use the global variable.
I picked up pylint mostly for the fun of it, so getting rid of specific warnings is probably not a priority atm and I will not bother adding pylint's version of noqa for now.
Can probably evaluate later whether the stuff that I find with pylint is enough to warrant integrating it into the CI or should just do it as a one-off.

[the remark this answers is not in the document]
gsalgado: Yeah, that could be a problem and it doesn't look like any of our CI checks would prevent that. Maybe just using a name like `_test()` for those functions would be enough?

evm/p2p/auth.py
```diff
@@ -160,9 +161,9 @@ def decode_auth_ack_message(self, ciphertext: bytes) -> Tuple[datatypes.PublicKe
         if len(ciphertext) < ENCRYPTED_AUTH_ACK_LEN:
             raise ValueError("Auth ack msg too short: {}".format(len(ciphertext)))
         elif len(ciphertext) == ENCRYPTED_AUTH_ACK_LEN:
-            eph_pubkey, nonce, version = decode_ack_plain(ciphertext, self.privkey)
+            eph_pubkey, nonce, __ = decode_ack_plain(ciphertext, self.privkey)
```
pipermerriam: Any reason for using a double underscore over a single underscore?  We have an existing convention of using single underscore as the *not-used* variable.

h00701350103: I found https://stackoverflow.com/questions/5893163/what-is-the-purpose-of-the-single-underscore-variable-in-python/5893946#5893946

But I'll switch to single underscore if that's the convention in this project.

pipermerriam: Thnx @h00701350103 , I'd like to stick with single underscore for now.

h00701350103:
Finally got down to continue working on this. All earlier comments should be addressed.

Stuff I wasn't sure about:

## evm/p2p/protocol
### E: 51,22: No value for argument 'obj' in constructor call (no-value-for-parameter)
```python
raise rlp.EncodingError(
                    "Keys in data dict ({}) do not match expected keys ({})".format(
                        data_keys, expected_keys))
```
obj should be "the object that could not be encoded". Not entirely sure this is the correct exception to throw here as we haven't yet calculated the data that will be decoded later and therefore can't pass it to EncodingError.

## evm/p2p/test_peer
### W:184, 4: __init__ method from base class 'LESPeer' is not called (super-init-not-called)
I tried touching this and failed to get anything working out. Not sure if it's a problem or not.


After this I think most obvious stuff in evm/p2p is fixed. For future linters I used the following disable's to preserve my sanity:
missing-docstring,no-member,cyclic-import,too-many-locals,too-many-arguments,protected-access,unbalanced-tuple-unpacking,bad-continuation,too-few-public-methods,line-too-long,invalid-name,fixme,redefined-builtin,no-self-use,unused-import,too-many-public-methods,too-many-instance-attributes,duplicate-code

These are a combination of bugs in pylint (e.g. no-member), systematic deviations from Best Practices (e.g. missing-docstring, invalid-name, redefined-builtin), noqa's for flake8 not registered by pylint (e.g. line-too-long), and pylint not being smart enough to detect some correct code (e.g. unused-import)

evm/p2p/les.py
h00701350103: Not fixed as per the other comment chain (and this change done in many other places).

h00701350103:
Merged upstream changes, changed the rlp.EncodingError to a ValueError. This should be all ready to merge now.

pipermerriam:
@h00701350103 can you rebase this from master.  It should clean up the diff as it currently seems to be showing more than just your changes.

pipermerriam:
@h00701350103 ignore my comment about rebasing.  I wasn't looking at the right diff.

Issue #281, opened 2018-01-17
Missing `account_exists` implementation prevents execution of sharding transactions


jannikluhn:
### What is wrong?

Executing sharding transactions currently does not work with the FlatTrieBackend as it doesn't implement `account_exists`.

Note that it is not related to account creation transactions per se (as I assumed earlier), normal transactions are affected in the same way.

### How can it be fixed

Either implement `account_exists` or don't use it. It is called in some SELFDESTUCT (e.g. https://github.com/ethereum/py-evm/blob/362266c323c83a851f7dff0a2a59796abaa3f318/evm/logic/system.py#L64) and some CALL (e.g. https://github.com/ethereum/py-evm/blob/362266c323c83a851f7dff0a2a59796abaa3f318/evm/logic/call.py#L287) opcodes. It changes from fork to fork though and I'm not sure if this is still needed, especially as at least one one motivation seems to have been the dust clean up in Spurious Dragon. Also, it seems to be used for gas calculation only which will be different with stateless clients anyway.

It is also called indirectly via `touch_account` for every message, but this can unproblematically be replaced by setting nonce/balance/code to zero unless they are non-zero.

Implementing `account_exists` is probably the easier solution (with `check_if_branch_exist` in py-trie), but I don't like this as much. But that's mainly due to me not seeing why such a function would be necessary.

hwwhww:
Checked with V:

* We can remove `SELFDESTRUCT` opcode from sharding now.
* For `CALL` opcode, we may have to override it by adding `CallSharding` class:
    * Only check `account_is_empty`, not `account_exists`.
    * Look over the logic of `CALL`, the `nonce` part may be very different from the account abstraction model?

jannikluhn:
> Look over the logic of CALL, would the nonce part become very different with the account abstraction model?

I was under the impression that contract nonces would be removed altogether. In its current implementation CREATE2 uses the contract nonce though, so I'm not sure about this. But I don't think it matters a lot, the only thing that would be affected is the definition of an "empty account" (`code/balance/nonce = 0` to `code/balance = 0`).

NIC619:
I can get my hands on `CallSharding`, removing `SELFDESTRUCT`, `account_is_empty`, `account_exist` and `touch_account` after `PAYGAS` opcode is done.

jannikluhn:
I don't think the account methods need to be removed, rather changed slightly:

- `account_is_empty` should not check the nonce
[...]

Issue #289, opened 2018-01-19
Switch to `BinaryTrie` state


hwwhww:
### What is wrong?
Shard chains use `BinaryTrie` for better efficiency.
See https://github.com/ethereum/wiki/wiki/Wishlist#trie

### How can it be fixed
1. Set `BinaryTrie` as the default trie of `FlatTrieBackend`.
2. Apply `FlatTrieBackend` in most sharding tests.

Will switch transaction trie and receipt trie in other tickets.

NIC619:
@jannikluhn @hwwhww @pipermerriam Since gas consumption of creating a new account depends on `account_exist`:
```python
class Call(BaseCall):
    def compute_msg_extra_gas(self, computation, gas, to, value):
        …
        create_gas_fee = constants.GAS_NEWACCOUNT if not account_exists else 0
        ...
```
Unless we implement `account_exist` in FlatTrieBackend(or replace it's logic with `account_is_empty`), it seems that `FlatTrieBackend` is not backward compatible.
And if `FlatTrieBackend` is not backward compatible, I would say we can remove(or disable) `touch_account` as well.

hwwhww:
hmm, so part of this issue also depends on https://github.com/ethereum/py-evm/issues/281 ?

> `FlatTrieBackend` is not backward compatible

It's could be compatible theoretically if we implement these functions, but practically, we don't need them. How about just raising `NotImplementedError` exceptions for these function + not testing `FlatTrieBackend` with `class Call`, just testing with the new `CallSharding`? Would that work?

NIC619:
Yeah I think replacing with `CallSharding` would work for both this issue and #281 .

NIC619:
My bad. `touch_account` is actually used in `ShardingComputation.apply_message` so `ShardingComputation.apply_message` needs to be changed too.

hwwhww:
@NIC619  @jannikluhn  Could we modify `FlatTrieBackend.touch_account(address)` from 
```python
def touch_account(self, address):
        if not self.account_exists(address):
            self.set_nonce(address, 0)
            self.set_balance(address, 0)
            self.set_code(address, b'')
```

to

```python
def touch_account(self, address):
        if not self.account_is_empty(address):
            self.set_nonce(address, 0)
            self.set_balance(address, 0)
            self.set_code(address, b'')
```
?

jannikluhn:
Yes, did this in https://github.com/ethereum/py-evm/pull/291

[...]

PR #291, opened 2018-01-22
Avoid usage of account_exists

jannikluhn:
### What was wrong?

#281

### How was it fixed?

- [x] SELFDESTRUCT was removed
- [x] the implementation of CALL was updated to not check for account existence, but emptiness instead (only used for gas calculation)
- [x] `touch_account` was reimplemented to check for "zero-ness" instead of existence before touching.

#### Cute Animal Picture

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -26,8 +26,7 @@ from evm.db.journal import (
     JournalDB,
 )
 from evm.db.state import (
-    AccountStateDB,
-    NestedTrieBackend,
+    MainAccountStateDB,
 )
 from evm.rlp.headers import (
     BlockHeader,
@@ -59,9 +58,9 @@ class TransactionKey(rlp.Serializable):
 
 class BaseChainDB:
 
-    def __init__(self, db, state_backend_class=NestedTrieBackend):
+    def __init__(self, db, account_state_class=MainAccountStateDB):
         self.db = JournalDB(db)
-        self.state_backend_class = state_backend_class
+        self.account_state_class = account_state_class
 
     def exists(self, key):
         return self.db.exists(key)
@@ -376,10 +375,11 @@ class BaseChainDB:
         self.db.clear()
 
     def get_state_db(self, state_root, read_only, access_list=None):
-        return AccountStateDB(
-            db=self.db,
-            root_hash=state_root,
-            read_only=read_only,
-            access_list=access_list,
-            backend_class=self.state_backend_class,
-        )
+        kwargs = {
+            "db": self.db,
+            "root_hash": state_root,
+            "read_only": read_only,
+        }
+        if access_list is not None:
+            kwargs["access_list"] = access_list
+        return self.account_state_class(**kwargs)
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -375,11 +375,12 @@ class BaseChainDB:
         self.db.clear()
 
     def get_state_db(self, state_root, read_only, access_list=None):
-        kwargs = {
-            "db": self.db,
-            "root_hash": state_root,
-            "read_only": read_only,
-        }
+        extra_kwargs = {}
         if access_list is not None:
-            kwargs["access_list"] = access_list
-        return self.account_state_class(**kwargs)
+            extra_kwargs["access_list"] = access_list
+        return self.account_state_class(
+            db=self.db,
+            root_hash=state_root,
+            read_only=read_only,
+            **extra_kwargs
+        )
--- a/evm/db/state.py
+++ b/evm/db/state.py
@@ -40,10 +40,10 @@ from evm.utils.padding import (
     pad32,
 )
 from evm.utils.state_access_restriction import (
-    code_key,
-    balance_key,
-    nonce_key,
-    storage_key,
+    get_code_key,
+    get_balance_key,
+    get_nonce_key,
+    get_storage_key,
 )
 
 from .hash_trie import HashTrie
@@ -319,7 +319,7 @@ class ShardingAccountStateDB(BaseAccountStateDB):
         validate_canonical_address(address, title="Storage Address")
         validate_uint256(slot, title="Storage Slot")
 
-        key = storage_key(address, slot)
+        key = get_storage_key(address, slot)
         self._check_accessibility(key)
 
         if key in self._trie:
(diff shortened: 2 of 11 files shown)


evm/logic/call.py
```diff
@@ -348,3 +348,13 @@ def get_call_params(self, computation):
...
         return call_params
+
+
+class CallSharding(CallByzantium):
+    def compute_msg_extra_gas(self, computation, gas, to, value):
+        with computation.state_db(read_only=True) as state_db:
+            account_is_empty = state_db.account_is_empty(to)
+
+        transfer_gas_fee = constants.GAS_CALLVALUE if value else 0
+        create_gas_fee = constants.GAS_NEWACCOUNT if not account_is_empty else 0
```
NIC619: It seems `create_gas_fee` should only be charged if the account is empty?

But with `FlatTrieBackend` IMHO we should probably review/modify this `create_gas_fee` as described here: https://github.com/ethereum/py-evm/issues/289#issuecomment-359379247

evm/db/state.py
```diff
@@ -300,7 +300,8 @@ def account_has_code_or_nonce(self, address):
         return self.get_code(address) != b'' or self.get_nonce(address) != 0
 
     def touch_account(self, address):
-        if not self.account_exists(address):
+        do_not_touch = self.account_has_code_or_nonce(address) or self.get_balance(address) != 0
```
pipermerriam: can we invert this.

```python
if self.account_has_code_or_nonce(address):
    should_touch = False
elif self.get_balance(address) != 0:
    should_touch = False
else:
    should_touch = True:

if should_touch:
    ...
```

It's longer and more explicit, but I think it's easier to follow the logic.
should_touch = 
    not self.account_has_
)

jannikluhn:
As discussed [here](https://github.com/ethereum/py-evm/issues/281#issuecomment-359521568), I refactored away the state DB backend abstraction. Instead, there's now a `MainAccountStateDB` (for main chain) and a `ShardingAccountStateDB` (for shards). `touch_account` was removed completely. This makes some tests fail, which I've marked as `xfail`ing. I will fix them in #276 as far as possible because I think it fits slightly (but only slightly) better there.

@pipermerriam Could you review and in particular comment on the `xfail` solution (see also https://github.com/ethereum/py-evm/pull/276#issuecomment-358903670)? If you don't like it I can switch to `pytest.raises` in this PR as well as #276 and #288.

evm/db/chain.py
```diff
@@ -245,10 +244,11 @@ def clear(self):
...
-            root_hash=state_root,
-            read_only=read_only,
-            access_list=access_list,
-            backend_class=self.state_backend_class
-        )
+        kwargs = {
+            "db": self.db,
+            "root_hash": state_root,
+            "read_only": read_only,
+        }
```
pipermerriam: What if instead of constructing the kwargs here you simply made this a passthrough like the following?

```python
def get_state_db(self, state_root, read_only, **extra_kwargs):
    return self.account_state_class(db=self.db, state_root=state_root, read_only=read_only, **extra_kwargs)
```

I'll admit that layers of passthrough `**kwargs` can get really confusing really fast but here it might be clean enough.

evm/utils/state_access_restriction.py
```diff
@@ -60,8 +67,28 @@ def to_prefix_list_form(access_list):
...
-        yield keccak(address) + BALANCE_TRIE_PREFIX
-        yield keccak(address) + CODE_TRIE_PREFIX
+        yield nonce_key(address)
+        yield balance_key(address)
+        yield code_key(address)
         for prefix in remove_redundant_prefixes(storage_prefixes):
             yield keccak(address) + STORAGE_TRIE_PREFIX + prefix
+
+
+def storage_key(address, slot):
```
pipermerriam: *naming nitpick*

I had to scroll down to this module to look at  the implementation of these to find out if they were values or functions.  Could we prefix these names with something to indicate they are actions:

- `mk_storage_key`
- `construct_storage_key`
- `get_storage_key`

evm/db/state.py
```diff
@@ -252,169 +320,62 @@ def get_balance(self, address):
...
-        else:
-            return 0
-
-    def set_code(self, address, code):
-        key = self.code_key(address)
-        self._check_accessibility(key)
+        self._trie[key] = int_to_big_endian(nonce)
 
-        self._trie[key] = code
+    def increment_nonce(self, address):
```
pipermerriam: The implementations of some of these *meta* APIs are the same across these two backends.  Might be suitable to have a `BaseXXXX` backend class whith these common implementations for `increment_nonec, delta_balances,account_has_code_or_nonce,account_is_empty`

tests/auxiliary/user-account/test_contract.py
```diff
@@ -347,6 +355,7 @@ def test_no_call_if_not_enough_gas(vm):
     assert computation.gas_meter.gas_remaining > 0
 
 
+@pytest.mark.xfail(reason="obsolete gas payment mechanism, #234, and #281", strict=True)
```
pipermerriam: I'm fine with leaving these this way *if* you will 1) create an issue to track the cleanup, and 2) you expect to address this in the near future.

Otherwise I'd be worried about us forgetting about these, them no longer actually failing, but the test suite continuing to pass.

tests/core/vm/test_sharding_vm.py
```diff
@@ -90,7 +93,8 @@ def test_sharding_apply_transaction(shard_chain_without_block_validation):  # no
         assert state_db.get_storage(CREATE2_contract_address, 0) == 1
 
 
-def test_CREATE2_deploy_contract_edge_cases(shard_chain_without_block_validation):  # noqa: F811
+@pytest.mark.xfail(reason="#281", strict=True)  # noqa: F811
```
pipermerriam: I'm gonna eat my words and say that this is fine now that I know about the `strict` paramter.  That's great.

pipermerriam: Oooh! it looks like we can enable this globally by default: by adding to our `pytest.ini`

```
[pytest]
xfail_strict=true
```

pipermerriam: And... it might be nice to add the `raises` parameter to any of these that you expect to hang around for a while so that we also know they are failing for the expected reason.

https://docs.pytest.org/en/latest/skipping.html#raises-parameter

jannikluhn:
Addressed the feedback, will merge after rebase on #307

PR #292, opened 2018-01-22
Set gas cost of SIGHASH

jannikluhn:
### What was wrong?

Gas cost of the new SIGHASH opcode was set to 0.

### How was it fixed?

Gas cost was set to GAS_BASE (2). This seems reasonable to me as similar environmental constants have the same cost (TIMESTAMP, CALLER,  GASPRICE, ...)

#### Cute Animal Picture

### DIFF
--- a/evm/vm/forks/sharding/opcodes.py
+++ b/evm/vm/forks/sharding/opcodes.py
@@ -21,7 +21,7 @@ NEW_OPCODES = {
     opcode_values.SIGHASH: as_opcode(
         logic_fn=context.sighash,
         mnemonic=mnemonics.SIGHASH,
-        gas_cost=0,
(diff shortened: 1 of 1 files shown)


hwwhww:
I'm not sure if it should be higher than `GAS_BASE(2)` since this opcode actually does a `SHA3` operation, and `GAS_SHA3` is 30.
Page @vbuterin, we need it to be defined in the incoming account abstraction spec.

pipermerriam:
Does this need a `TODO` comment to note that the current price may change once this is better defined in the account abstraction spec?  Worried that we'll forget (maybe even needs to be tracked with an issue?)

hwwhww:
Checked with V, he said keeping it to `GAS_BASE(2)`. I think it's ready to merge.

PR #296, opened 2018-01-23
Fix CI errors

mhchia:
### What was wrong?
* In the latest `eth-tester`, there is some usage of the latest syntax in py-evm master branch, which is not yet merged(rebased) to py-evm sharding branch. This cause CI errors in sharding branch.

* Moreover, since `TransactionFailed` is added in `eth-tester` recently and is not caught by `web3.py`, which causes tests to fail.

### How was it fixed?
* Specify the version of `eth-tester` to `0.1.0b12`
* Modify the test code to catch `eth_tester.exceptions.TransactionFailed`, instead of `web3.exceptions.BadFunctionCallOutput`

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,7 @@ setup(
         "eth-keys==0.1.0b3",
         "trie>=0.3.1",
         "trie>=1.0.1,<2.0.0",
-        "eth-tester>=0.1.0b11",
+        "eth-tester==0.1.0b12",
         "web3>=4.0.0b5",
     ],
     extra_require={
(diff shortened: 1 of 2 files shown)


pipermerriam:
I pushed `0.1.0b13` yesterday evening.  It has some additional fixes in it if you want to go ahead and bump the version one more.

mhchia:
Sorry but it seems if we want to use `eth-tester==0.1.0b13` instead, we have to rebase the sharding branch onto master first, to avoid errors. So this PR is just intended to fix the CI error in the short-term, before the rebase is done.

pipermerriam:
:shipit:

mhchia:
Thank you:)

Issue #297, opened 2018-01-23
Add IPython Support To Default Py-EVM shell


amitt001:
### What is wrong?

Nothing is wrong with current flow. This will be an improvement to current Py-EVM lightchain_shell.py module interactive mode shell.

IPython is very widely used Python REPL. It provides tab completion/hinting, command history, command autocomplete, history navigation and many such handy features that help in development.

### How can it be fixed

lightchain_shell.py file can be modified to add support for running shell from the code. By default run the IPython shell if it is installed on the current PYTHONPATH. If an import error is raised run the default Python REPL with improvements like tab completion and history navigation through up key.

PR #298, opened 2018-01-23
Added support for IPython shell and improved default shell, updated docs

amitt001:
### What was wrong?

Implementation of https://github.com/ethereum/py-evm/issues/297

### How was it fixed?

Added support for IPython shell in the lightchain_shell.py module.

- Run `IPython.terminal.embed.InteractiveShellEmbed` if IPython <= v1.0.
- Run `IPython.frontend.terminal.embed.InteractiveShellEmbed` if IPython > v1.0.
- If IPython is installed run the default Python shell.

By default, IPython is run if it is available. To run the default Python shell pass `-default_shell` as an argument on command line. Ex:

`python -m evm.lightchain_shell -db /tmp/testnet.db -default_shell`

Also updated the docs for changes and leveldb requirement with lighchain_shell as note.

#### Cute Animal Picture

choco

### DIFF
--- a/evm/lightchain_shell.py
+++ b/evm/lightchain_shell.py
@@ -29,7 +29,8 @@ LOGLEVEL = logging.INFO
 parser = argparse.ArgumentParser()
 parser.add_argument('-db', type=str, required=True)
 parser.add_argument('-debug', action='store_true')
-parser.add_argument('-default_shell', action='store_true', default=False)
+# pass -vanilla_shell at command line to run a Python REPL
+parser.add_argument('-vanilla_shell', action='store_true', default=False)
 args = parser.parse_args()
 
 print("Logging to", LOGFILE)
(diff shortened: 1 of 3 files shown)


evm/lightchain_shell.py
```diff
@@ -28,6 +29,7 @@
 parser = argparse.ArgumentParser()
 parser.add_argument('-db', type=str, required=True)
 parser.add_argument('-debug', action='store_true')
+parser.add_argument('-default_shell', action='store_true', default=False)
```
gsalgado: It was only after I read the implementation that I understood what default means here. Maybe vanilla_shell would be a better name for it?

evm/lightchain_shell.py
```diff
@@ -74,3 +76,81 @@ def cleanup():
 
 
 atexit.register(cleanup)
+
+# This is to start IPython or Python REPL.
+
+def ipython_shell_1(namespace=None, banner=None, debug=False):
```
gsalgado: Should we drop the banner and namespace args from all those functions and add it if/when we have a use for it?

evm/lightchain_shell.py
```diff
@@ -74,3 +76,81 @@ def cleanup():
...
+            traceback.print_exc()
+        return None
+    kwargs = dict(user_ns=namespace)
+    if banner:
+        kwargs = dict(banner1=banner)
+    return InteractiveShellEmbed.instance(**kwargs)
+
+
+def ipython_shell(namespace=None, banner=None, debug=False):
+    """IPython version > 1.0"""
```
gsalgado: Did you consider combining those two functions in a single one? You could try importing from `IPython.terminal.embed` first and, if that fails, try from `IPython.frontend.terminal.embed`. That would avoid some duplication and we could also get rid of `start_ipython_shell` and do something like this in `start_shell()`

```
if use_ipython:
    ipython_shell = ipython_shell(...)
    if ipython_shell is not None:
        ipython_shell()
        return
    else:
        print("IPython not available")
...
```

evm/lightchain_shell.py
amitt001: Rename done.

evm/lightchain_shell.py
amitt001: I think we can keep it as None for now. If we ever need it we can just pass it w/o much modification to the existing code.

amitt001:
@gsalgado Did the necessary refactoring.

Removed the `ipython_shell_1` and `ipython_shell_4` functions. Keeping `start_ipython_shell` and `start_python_shell` to put the start shell logic for Python and IPython in a separate place. `start_shell` calls both the methods based on arguments passed.

Do the changes look ok now?

amitt001:
@pipermerriam Yes sure, I will be happy to work on it. I will take a look at #300 and start porting things from here. I have a couple of questions.

Is #300 ready to be merged into Trinity branch?
I think this makes #297 invalid. Shall I close it?
Is Trinity CLI going to be the default(and only) CLI, to eventually replace current lighchain_shell?

pipermerriam:
#300 is not merged but will be merged within a few hours.  #297 will need to be closed (which I'm doing) so that you can re-open against the trinity branch.  You should rebase on top of #300 with the knowledge that you'll need to rebase again when #300 is merged.

> Is Trinity CLI going to be the default(and only) CLI, to eventually replace current lighchain_shell or Py-EVM will have it's own CLI?

Trinity will be the only CLI interface for running a node.  Py-EVM will *eventually* be an EVM only library, with `trinity` containing all of the networking logic (and eventually split out to it's own package).  For now we're keeping everything bundled to reduce development overhead.

Issue #198, opened 2017-12-05
Decide where to store config and chain data


gsalgado:
Currently py-evm doesn't have a standard place to store its configs and chain data. We need to agree on one and start storing stuff there to avoid having to hard-code or [regenerate values on every run](https://github.com/ethereum/py-evm/pull/189#discussion_r154775751)

I believe we should use $XDG_DATA_HOME and $XDG_CONFIG_HOME as per https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

pipermerriam:
@gsalgado I took a quick gander at that spec.  Would you mind expanding this issue to contain the following.

* What data do we need to store that is *global*
* What data do we need to store for each *chain*
* What folder structure we might use for this data.
* What base directories we would use according the the XDG spec from above.

gsalgado:
I cannot think of any global data at this point. Can you think of anything?

Chain specific:
- block/state (one for lightchain and one for full chain) 
- private keys for accounts
- P2P node's private key (used to generate the node's unique ID and for authentication)
- peer statistics

Since all of that is actually _data_ (and not user-maintained _config_) generated/updated by py-evm, I believe it should go under _$XDG_DATA_HOME_ (which defaults to _$HOME/.local/share_).

As for the directory structure, I think the following would be a good start:

```
├── py-evm
│   └── mainnet
│       ├── chain         # LevelDB
│       ├── keys          # KeyStore, whatever format we chose
│       ├── lightchain    # LevelDB
│       ├── nodekey       # a text file containing the node's P2P private key
│       └── peers         # LevelDB
```

That also leaves room for any global data to be stored directly under _$XDG_DATA_HOME/py-evm_

pipermerriam:
I like it and happy to learn about `$XDG_DATA_HOME` which is a standard I wasn't aware of.

pipermerriam:
Oh, I think we might benefit from **not** implementing a keystore.  I don't want to manage user's private keys (at least not using py-evm).  I'm imagining this as something that moves into the web3.py layer with the `web3.account` API.  We can still read from keystore files and whatnot, I just don't want to do it at the py-evm level.

pipermerriam:
This is fixed in #300 which is about to get merged.  Closing.

PR #300, opened 2018-01-23
Initial setup of trinity CLI

pipermerriam:
### What was wrong?

We need a *real* command line interface for running the node, and it's time we start calling it by it's true name......... **Trinity**!!!!!

### How was it fixed?

Rather than doing this in a separate repo, for now I figure we should develop side-by-side and extract things when we start getting close to the end of our alpha cycle.

This adds a CLI entry point named `trinity` which accepts a single option `--log-level` which can be set to either `info` or `debug`.  

The chain sync is run in it's own process using `multiprocessing`.  The reasoning here is to prevent having the same problem that pyethapp had.  By being single threaded, and being kind of slow since this is python, we need to be sure that the time it takes to process blocks/transactions doesn't cause timeouts.  To accomplish this, we'll partition all of the networking code to live in one process and the EVM execution to be separated into a different process.

The logging is handled using built-in utilities in the `multiprocessing` library so that the loggers from within each process pass their logs through a `multiprocessing.Queue` which is read from from the primary process which aggregates all of the logs.

This also implements the solution laid out in https://github.com/ethereum/py-evm/issues/198 , storing the chaindata in `$XDG_DATA_HOME/trinity/`.

#### Cute Animal Picture

tiny-animals-31


trinity/main.py
```diff
@@ -0,0 +1,110 @@
...
+
+    try:
+        process.start()
+        process.join()
+    except KeyboardInterrupt:
+        logger.info('Keyboard Interrupt: Stopping')
+        process.terminate()
+
+
+def ropsten_light_node_sync(log_queue, db_path):
```
pipermerriam: If you're ok with it, ignore this entire function body.  I expect to fully rework this code and it's only here so that there is something for the CLI to run.

.travis.yml
```diff
@@ -11,98 +11,98 @@ matrix:
...
+    - # - python: "3.6"
+      # - env: TOX_POSARGS="-e py36-transactions"
+    - # - python: "3.6"
+      # - env: TOX_POSARGS="-e py36-p2p"
+    - # - python: "3.6"
+      # - env: TOX_POSARGS="-e py36-database"
+    - # - python: "3.6"
+      # - env: TOX_POSARGS="-e py36-rpc-state-quadratic"
+    #-# -  python: "3.6"
+    #-# -  env: TOX_POSARGS="-e py36-vm-performance"
```
carver: Hah, that's one way to easily pass the tests. Just a reminder to revert this before merging...

Side thought: maybe we could only run the py36 tests in the cron job? Which reminds me that that is still probably broken from long-running tests that I said I would clean up. Adding a note to myself...

trinity/README.md
```diff
@@ -0,0 +1,10 @@
+
+
+
+## Environment Configuration
+
+- `TRINITY_MP_CONTEXT` - The context that new processes will be spawned from the python `multiprocessing` library.
+- `TRINITY_HOME` - Base directory where trinity stores data
+- `TRINITY_CHAIN_DIR` - The base directory where data will be store for the currently running chain.
+- `TRINITY_DATA_DIR` - The directory where the chain data will be stored for the currently running chain.
```
carver: I think this means that one of `('TRINITY_CHAIN_DIR', 'TRINITY_DATA_DIR')` contains the in-chain state, and the other contains metadata about the topology of the chain, but wasn't sure which was which.

trinity/utils/xdg.py
```diff
@@ -0,0 +1,63 @@
+import os
+
+
+XDG_CACHE_HOME = os.environ.get(
```
carver: What is XDG? The only things I came up with are Dogecoin and X Desktop group.

trinity/main.py
```diff
@@ -0,0 +1,110 @@
...
+    '-l',
+    '--log-level',
+    choices=LOG_LEVEL_CHOICES,
+    default=DEFAULT_LOG_LEVEL,
+)
+
+
+DemoLightChain = LightChain.configure(
+    name='Demo LightChain',
+    privkey=ecies.generate_privkey(),  # TODO: get from stored key.
```
carver: This will use `get_nodekey_path()`?

trinity/main.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+DEFAULT_LOG_LEVEL = 'info'
+LOG_LEVEL_CHOICES = (
+    'debug',
+    'info',
+)
+
+
+parser = argparse.ArgumentParser(description='Trinity')
```
carver: So we landed on argparse? :-/

Ok, just had to get in one last *grumble*. :)

trinity/utils/xdg.py
step21: A standard to specify home directory and similar, where to look for it etc... https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.html

trinity/main.py
pipermerriam: Yeah, I've got another branch in flight which deals with graceful detection of whether or not a chain directory has been initialized (db with canonical chain head, nodekey, genesis info) and initializing the things that need initializing on-the-fly when spinning up for the first time.

trinity/main.py
pipermerriam: No, but it's the easiest one to replace with the other tools and I want to see if we can get by with it.  So far it looks like it's going to be fine.

trinity/README.md
pipermerriam: This is piss poor documentation at-the-moment.  What I'm really doing here is keeping a running list of the environment variables so that when the time comes to document them *for-real* we don't have to excavate them from the codebase.

trinity/utils/xdg.py
pipermerriam: See https://github.com/ethereum/py-evm/issues/198

trinity/main.py
carver: Ok, fair enough. In the abstract, I do prefer using stdlib tools, so if it's not broken then :+1: .

trinity/utils/filesystem.py
```diff
@@ -0,0 +1,9 @@
+import os
+
+
+def ensure_path_exists(dir_path):
+    """
+    Make sure that a path exists
+    """
+    if not os.path.exists(dir_path):
+        os.makedirs(dir_path)
```
gsalgado: You can replace this with `os.makedirs(dir_path, exist_ok=True)`

trinity/main.py
```diff
@@ -0,0 +1,110 @@
...
+def main():
+    args = parser.parse_args()
+
+    logger, log_queue, listener = setup_trinity_logging(args.log_level.upper())
+
+    # start the listener thread to handle logs produced by other processes in
+    # the local logger.
+    listener.start()
+
+    db_path = get_data_dir('ropsten')
```
gsalgado: We should make a constant for this

trinity/main.py
```diff
@@ -0,0 +1,110 @@
...
+
+from trinity.utils.filesystem import (
+    ensure_path_exists,
+)
+from trinity.utils.logging import (
+    setup_trinity_logging,
+    setup_queue_logging,
+)
+from trinity.utils.mp import (
+    ctx,
```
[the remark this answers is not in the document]
pipermerriam: `multiprocessing`.  In a lot of the examples in the docs they do `import multiprocessing as mp`

trinity/main.py
```diff
@@ -0,0 +1,110 @@
...
+parser.add_argument(
+    '-l',
+    '--log-level',
+    choices=LOG_LEVEL_CHOICES,
+    default=DEFAULT_LOG_LEVEL,
+)
+
+
+DemoLightChain = LightChain.configure(
+    name='Demo LightChain',
```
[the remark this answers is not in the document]
pipermerriam: Yeah, though I'm taking care of that in the next PR if you don't mind.  Going with the convention: `RopsentLightChain`

trinity/main.py
pipermerriam: If you mean the string `'ropsten'`, yes, 👍

PR #307, opened 2018-01-25
Mend sharding branch after rebasing

hwwhww:
### What was wrong?
It's easier to do the refactoring after dealing with conflicts.

### How was it fixed?
1. Refactored `ShardingVM` like what we did to other VMs in `master` branch.
2. Made it compatible with the new `VMState` and `TrackedDB` interface.

p.s. Note that Travis CI doesn't check `pytest tests/auxiliary` now and it would fail. Leave it to be fixed later. cc @jannikluhn 

#### Cute Animal Picture

cat-sweet-kitty-animals-57416


evm/db/state.py
```diff
@@ -153,6 +153,9 @@ def delete_account(self, address):
     def account_exists(self, address):
         return bool(self._trie[address])
 
+    def account_has_code_or_nonce(self, address):
+        return self.get_code_hash(address) != EMPTY_SHA3 or self.get_nonce(address) != 0
```
pipermerriam: trivial but inverting these checks has a performance gain that we should probably take advantage of since this can end up being called numerous times during txn execution.


- `self.get_code_hash`  - 2 tries lookups and a keccak
- `self.get_nonce` - 1 trie lookup

Issue #301, opened 2018-01-24
Proper `Ropsten` configuration


pipermerriam:
### What is wrong?

The `evm.chains.ropsten.RopstenChain` currently uses the block number configurations from the mainnet.

https://github.com/ethereum/py-evm/blob/30c3e97e6050a9f04ad78b18e816764400242598/evm/chains/ropsten/__init__.py#L14

### How can it be fixed

Dig up the proper block numbers for the fork transitions and use the correct block numbers.

PR #308, opened 2018-01-25
Fix hard fork block numbers for Ropsten chain

lrettig:
These block numbers come from
https://github.com/ethereum/cpp-ethereum/blob/develop/libethashseal/genesis/ropsten.cpp.

Update comments and standardize constant names across both Ropsten and
Mainnet chains.

Closes https://github.com/ethereum/py-evm/issues/301.

### What was wrong?

Ropsten chain was using Mainnet chain fork block numbers.

### How was it fixed?

By pulling in the correct Ropsten chain block numbers from cpp-ethereum.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/ropsten/__init__.py
+++ b/evm/chains/ropsten/__init__.py
@@ -3,6 +3,7 @@ from eth_utils import decode_hex
 from .constants import (
     BYZANTIUM_ROPSTEN_BLOCK,
     SPURIOUS_DRAGON_ROPSTEN_BLOCK,
+    TANGERINE_WHISTLE_ROPSTEN_BLOCK,
 )
 from evm import constants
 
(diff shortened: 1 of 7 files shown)


pipermerriam:
VMs disallow duplicate block numbers.  You can merely remove the earlier VM's from the configuration if their block numbers are duplicated.

evm/chains/ropsten/__init__.py
```diff
@@ -1,17 +1,35 @@
...
+    ByzantiumVM,
+    FrontierVM,
+    SpuriousDragonVM,
+)
+
+
+ROPSTEN_VM_CONFIGURATION = (
+    # Note: Homestead and Tangerine Whistle are excluded here since VMs disallow duplicate block
+    # numbers and they are all zero for Ropsten.
+    (0, FrontierVM),
```
pipermerriam: This should actually be the `TangerineWhistleVM` since the ropsten network *starts* at that VM, effectively skipping over Frontier and Homestead.

lrettig:
Logged #323 for the strange import issue that came up.

PR #309, opened 2018-01-25
p2p: Redesign sub-protocol matching and msg processing

gsalgado:
P2P message processing was scattered across methods from multiple classes (Protocol,
Peer, LightChain).

Also, the sub-protocol matching was designed to allow a connection to a single remote
node to use various different protocols, but that's not useful now and may never be.

Started consolidating msg processing on Peer methods. Now Protocol/Command methods
only deal with encoding/decoding of messages

Changed sub-protocol matching to pick just one sub-protocol

Better to review both commits separately, I believe

### DIFF
--- a/evm/p2p/peer.py
+++ b/evm/p2p/peer.py
@@ -17,6 +17,7 @@ from cryptography.hazmat.primitives.constant_time import bytes_eq
 
 from eth_utils import (
     decode_hex,
+    encode_hex,
 )
 
 from eth_keys import (
@@ -34,7 +35,6 @@ from evm.p2p import protocol  # noqa: F401
 from evm.p2p.exceptions import (
     AuthenticationError,
     EmptyGetBlockHeadersReply,
-    HandshakeFailure,
     PeerConnectionLost,
     PeerDisconnected,
     UnknownProtocolCommand,
@@ -47,14 +47,8 @@ from evm.p2p.utils import (
     roundup_16,
     sxor,
 )
-from evm.p2p.eth import ETHProtocol
-from evm.p2p.les import (  # noqa: F401
-    Announce,
-    HeadInfo,
-    LESProtocol,
-    LESProtocolV2,
-    Status,
-)
+from evm.p2p import eth
+from evm.p2p import les
 from evm.p2p.p2p_proto import (
     Disconnect,
     DisconnectReason,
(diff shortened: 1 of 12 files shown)


evm/p2p/state.py
```diff
@@ -36,6 +36,7 @@ class StateDownloader:
     _report_interval = 10  # Number of seconds between progress reports.
     # TODO: Experiment with different timeout/max_pending values to find the combination that
     # yields the best results.
+    # FIXME: Should use the # of peers times MAX_STATE_FETCH here
```
pipermerriam: Is this as simple as adding a property method for this which computes this value?  If so... sounds like a good issue to create and label with "Good First Issue"

evm/p2p/peer.py
```diff
@@ -304,16 +290,16 @@ def process_p2p_handshake(self, msg: bytes) -> None:
             raise PeerDisconnected("Peer disconnected before completing handshake: {}".format(
                 decoded_msg['reason_name']))
         remote_capabilities = decoded_msg['capabilities']
-        self.match_protocols(remote_capabilities)
-        if len(self.enabled_sub_protocols) == 0:
+        self.select_sub_protocol(remote_capabilities)
```
pipermerriam: The *side-effect* nature of this stood out to me.  Is there a way that we can restructure so that it does something like return the selected subprotocol and then check it locally prior to setting it to `self.sub_proto`.

```python
self.sub_proto = self.select_sub_protocol(...)
if self.sub_proto is None:
    ...
```

evm/p2p/peer.py
```diff
@@ -696,7 +636,7 @@ def __init__(self, block_number, block_hash, total_difficulty, genesis_hash):
         self.genesis_hash = genesis_hash
 
 
-if __name__ == "__main__":
+def _test():
```
pipermerriam: Is this temporary for development?  Is there a plan to convert this to a proper test?  Convenience of some sort?

evm/p2p/peer.py
gsalgado: Yeah, that's a nice improvement

evm/p2p/peer.py
gsalgado: It could become an integration test, similar to the one for the lightchain, but in this branch I'm just moving it inside a function to avoid having all the variables defined here in the module's global namespace

evm/p2p/state.py
gsalgado: It should be, but I plan to tackle it soon when I refactor the `PeerPool` so that a single instance of it can be shared between the state and chain downloaders

PR #310, opened 2018-01-25
Add automatic chain initialization of chain directories to trinity.

pipermerriam:
### What was wrong?

There are a small number of things that need to be in place in order to run a chain.

- chain directory structure created
- chaindb initialized wtih genesis data
- nodekey

### How was it fixed?

Added detection for whether these things are initialized and code to do intelligent initialization.  Note that this is just a start, as it won't handle cases where it's a custom chain with custom genesis data, etc.

This PR currently special cases a number of things that will need to be polished out (such as only supporting the ropsten chain and light sync mode.

#### Cute Animal Picture

cute-tiny-animals-6

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -42,11 +42,10 @@ commands=
 deps = -r{toxinidir}/requirements-dev.txt
     coincurve
     database: leveldb
+    trinity: leveldb
 basepython =
     py35: python3.5
     py36: python3.6
-extras =
-    trinity: trinity
 
 [testenv:flake8]
 basepython=python
--- a/tox.ini
+++ b/tox.ini
@@ -45,6 +45,8 @@ deps = -r{toxinidir}/requirements-dev.txt
 basepython =
     py35: python3.5
     py36: python3.6
+extras =
+    trinity: trinity
 
 [testenv:flake8]
 basepython=python
--- a/trinity/chains/__init__.py
+++ b/trinity/chains/__init__.py
@@ -34,14 +34,14 @@ def is_chain_initialized(chain_config):
 
     if chain_config.nodekey_path is None:
         # has an explicitely defined nodekey
-        return True
+        pass
     elif not os.path.exists(chain_config.nodekey_path):
         return False
 
     if chain_config.nodekey is None:
         return False
 
-    chaindb = get_chain_db(chain_config.data_dir)
+    chaindb = get_chain_db(chain_config.database_dir)
     try:
         chaindb.get_canonical_head()
     except CanonicalHeadNotFound:
@@ -77,7 +77,7 @@ def initialize_chain(chain_config, sync_mode):
     chain_class = get_chain_protocol_class(chain_config, sync_mode)
 
     # Database Initialization
-    chaindb = get_chain_db(chain_config.data_dir)
+    chaindb = get_chain_db(chain_config.database_dir)
     try:
         chaindb.get_canonical_head()
     except CanonicalHeadNotFound:
--- a/trinity/chains/__init__.py
+++ b/trinity/chains/__init__.py
@@ -1,0 +1,128 @@
+import os
+
+from eth_utils import (
+    to_dict,
+)
+
+from evm.chains.ropsten import ROPSTEN_GENESIS_HEADER
+from evm.p2p import ecies
+from evm.exceptions import CanonicalHeadNotFound
+
+from trinity.constants import (
+    ROPSTEN,
+)
+from trinity.utils.db import (
+    get_chain_db,
+)
+from trinity.utils.filesystem import (
+    ensure_path_exists,
+)
+from trinity.utils.xdg import (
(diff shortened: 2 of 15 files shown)


pipermerriam:
@carver looking for review on this.  So much of this is really rough so think of this as a high level review.  Guessing much of this will still morph and change as I figure things out.

tox.ini
```diff
@@ -42,11 +42,10 @@ commands=
 deps = -r{toxinidir}/requirements-dev.txt
     coincurve
     database: leveldb
+    trinity: leveldb
 basepython =
     py35: python3.5
     py36: python3.6
-extras =
-    trinity: trinity
```
carver: Huh, it's annoying that this didn't work as expected. The extras-style setup ends up being a lot more portable, with less repeated requirements in configs. I don't have any idea why this failed though.

trinity/chains/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    if not os.path.exists(chain_config.data_dir):
+        return False
+
+    if not os.path.exists(chain_config.database_dir):
+        return False
+
+    if chain_config.nodekey_path is None:
+        # has an explicitely defined nodekey
+        return True
```
carver: I don't understand what is happening here. Why is this short-circuiting if the path is none, and doesn't care if the chain is initialized?

trinity/chains/__init__.py
```diff
@@ -0,0 +1,108 @@
...
+    # Nodekey
+    if chain_config.nodekey is None:
+        nodekey = ecies.generate_privkey()
+        with open(chain_config.nodekey_path, 'wb') as nodekey_file:
+            nodekey_file.write(nodekey.to_bytes())
+
+    chain_class = get_chain_protocol_class(chain_config, sync_mode)
+
+    # Database Initialization
+    chaindb = get_chain_db(chain_config.data_dir)
```
carver: what goes in the `database_dir` if the database goes in the `data_dir`? :squint:

trinity/main.py
```diff
@@ -79,21 +125,26 @@ def main():
...
-    """
-    Runs an LES node against the Ropsten network.
-    """
-    setup_queue_logging(log_queue)
+@with_queued_logging
+def run_chain(chain_config, sync_mode):
+    if not is_chain_initialized(chain_config):
+        # TODO: this will only work as is for chains with known genesis
+        # parameters.  Need to flesh out how genesis parameters for custom
+        # chains are defined and passed around.
```
carver: Can it just be the json-decoded dict of the standard genesis.json format?

trinity/main.py
```diff
@@ -79,21 +125,26 @@ def main():
...
+        chain_class = initialize_chain(chain_config, sync_mode=sync_mode)
+    else:
+        chain_class = get_chain_protocol_class(chain_config, sync_mode=sync_mode)
 
-    chaindb = BaseChainDB(LevelDB(db_path))
+    # TODO: this should probably be something that is passed in to allow
+    # specifying the db_path via the CLI as well as the db class.
```
carver: Seems like we should be choosing the class based on other configs they choose, like `--syncmode light`.

tox.ini
pipermerriam: Yeah, I was excited about this.  didn't futz with it for long so maybe re-visit at another time (newer tox version?? dunno)

trinity/chains/__init__.py
pipermerriam: You're lack of understanding is because this is broken (and not covered in the tests apparently).  This logic is wrong.  `nodekey_path == None` indicates that the nodekey was explicitly specified.  This should be:

```python
if chain_config.nodekey_path is not None and not os.path.exists(chain_config.nodekey_path):
   return False
```

Will fix

pipermerriam: Good catch.  Should be `database_dir`.

trinity/chains/__init__.py
pipermerriam: Good catch.  Should be `database_dir`.

trinity/main.py
pipermerriam: I think this comment is outdated, but you are right that the class we choose will be based on various CLI options.

trinity/chains/__init__.py
pipermerriam: You're lack of understanding is because this is broken (and not covered in the tests apparently).  This logic is wrong.  `nodekey_path == None` indicates that the nodekey was explicitly specified.  This should be:

```python
if chain_config.nodekey_path is not None and not os.path.exists(chain_config.nodekey_path):
   return False
```

Will fix

trinity/main.py
pipermerriam: Yes, this comment is more to indicate that we need to support a `--genesis` argument for specifying the genesis data.

PR #313, opened 2018-01-26
Fix transaction execution

jannikluhn:
Replaces #276 (I suck at rebasing and thought rewriting this would be faster -- probably mistakenly, but well...).

1) transaction access list is converted to prefix list at initialization (that's the format used in the rest of the code)
2) state accessibility is checked during transaction execution (both during pre- and post-processing and for all opcodes), using `computation.state_db` when available.
3) access lists are passed down from transaction to message and from message to child message (see also #262)
4) `test_sharding_vm` and user account contract tests have been updated to provide transactions with correct access lists

When fixing 3) I noticed the use of `prepare_child_message` was not consistent: `BaseComputation` has methods `prepare_child_message` and `prepare_child_sharding_message`. `ShardingMessage` has an additional `prepare_child_message`.

What I did was removing all `prepare_child_sharding_message`s and implementing `ShardingComputation.prepare_child_message`. So now only the computation objects have a `prepare_child_message`, and the sharding specific functionality (passing on access list and sighash) is implemented in `ShardingComputation`. I don't see anything wrong with this, just mentioning it explicitly because it's probably the only noteworthy difference to #276.

#### Cute Animal Picture

### DIFF
--- a/evm/computation.py
+++ b/evm/computation.py
@@ -35,7 +35,6 @@ from evm.vm.memory import (
 )
 from evm.vm.message import (
     Message,
-    ShardingMessage,
 )
 from evm.vm.stack import (
     Stack,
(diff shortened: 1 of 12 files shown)


evm/computation.py
```diff
@@ -145,31 +144,6 @@ def prepare_child_message(self,
         )
         return child_message
 
-    def prepare_child_sharding_message(self,
```
pipermerriam: this looks like a good change.

evm/rlp/transactions.py
```diff
@@ -161,6 +164,10 @@ class BaseShardingTransaction(rlp.Serializable):
         ('code', binary),
     ]
 
+    def __init__(self, chain_id, shard_id, to, data, gas, gas_price, access_list, code):
+        super().__init__(chain_id, shard_id, to, data, gas, gas_price, access_list, code)
```
pipermerriam: You could `*args, **kwargs` both this call and the `__init__` signature here since you aren't using any of these variables in the local context.

hwwhww:
Maybe update `tox.ini` for testing `tests/auxiliary` in this PR?

jannikluhn:
Will do that in a separate PR (in my experience, fixing Travis requires some trial and error and I'd like to keep this one clean).

Issue #306, opened 2018-01-25
Function to Execute Raw Bytecode


cwinsloweaton:
Was asked to open an issue here, hope this is alright.

It would be pretty useful to have a function that takes bytecode and data and executes the bytecode along with the data. Ethereumjs-vm has a function [runCode()](https://github.com/ethereumjs/ethereumjs-vm#vmruncodeopts-cb) that does this and works pretty well, but I'd rather be using Python.

pipermerriam:
This strikes me as a nice convenience API to have available.  I think this fits well as a new API on the `VM` class or as a stand-alone that takes a VM class as one of it's arguments.  There are a few additional parameters that would need to be filled in, though I suspect we can supply sane defaults for most of them (`sender/coinbase/anything-else-on-the-message-object`).

pipermerriam:
```python
def execute_raw_bytecode(vm, bytecode, data=b'', ...):
    ...
```

Extra arguments would be something like every property of the `evm.vm.message.Message` object, with all of them defaulted to empty values or whatever other defaults we can come up with.

lrettig:
This is up my alley, given the work I've been doing on eWASM. I'll take a crack at it.

lrettig:
How's this for a set of defaults?

These mostly come from https://github.com/ethereumjs/ethereumjs-vm/blob/master/lib/runCode.js (except that it does not specify a default `gasPrice`):

- gas: 0xffffff
- gas_price: 0x01
- to: 0x0
- sender: 0x0
- value: 0x0
- data: None
- code: None

These defaults are already specified in `evm.vm.message.Message`:
- origin: None
- depth: 0
- create_address=None
- code_address=None
- should_transfer_value=True
- is_static=False

pipermerriam:
gas should probably default to the current lock gas limit.

everything else looks good.

lrettig:
Question: how much validation do we need to do on these args? E.g., do we care about gas metering or that the sender account contains at least the value being sent (plus gas)?

pipermerriam:
I think that it's fine to do minimal validation (or near zero validation).  If they send value, we could credit the `to` address, but only deduct from the `sender` if they have appropriate funds.  Or we could just not do the transfer at all, or we could only credit the `to` address....  I'm not sure what is right because I'm not sure all the ways this would be used.  Maybe better to lay out the known use cases and then design with those in mind.

lrettig:
[...]

PR #315, opened 2018-01-26
Add VM function to execute raw bytecode

lrettig:
CF #306

### DIFF
--- a/tests/json-fixtures/test_virtual_machine.py
+++ b/tests/json-fixtures/test_virtual_machine.py
@@ -135,23 +135,7 @@ def vm_class(request):
         assert False, "Unsupported VM: {0}".format(request.param)
 
 
-def test_vm_fixtures(fixture, vm_class):
-    chaindb = ChainDB(get_db_backend())
-    header = BlockHeader(
-        coinbase=fixture['env']['currentCoinbase'],
-        difficulty=fixture['env']['currentDifficulty'],
-        block_number=fixture['env']['currentNumber'],
-        gas_limit=fixture['env']['currentGasLimit'],
-        timestamp=fixture['env']['currentTimestamp'],
-    )
-    vm = vm_class(header=header, chaindb=chaindb)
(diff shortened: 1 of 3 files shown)


lrettig:
This is a pretty minimal prototype. It's basically just a thin wrapper over the `Message` object, leaving out a few details (`origin`, `create_address`, `code_address`, `should_transfer_value`, `is_static`).

For now, for testing, I'm running the full suite of VM tests through this new entrance. Not sure if this is what we want to do--we may want to also test different combinations of parameters.

@cwinsloweaton please take a look, maybe give this a shot and see if it suits your purposes.

(Just realized I still need to add `code_address` per https://github.com/ethereum/py-evm/issues/306#issuecomment-362027669.)

pipermerriam:
@lrettig can you rebase?

lrettig:
Rebased and added `code_address`. I think it might be good to add a test for this too, e.g., to try reading the state of an existing contract account using arbitrary code.

tests/json-fixtures/test_virtual_machine.py
```diff
@@ -164,60 +148,95 @@ def test_vm_fixtures(fixture, vm_class):
...
+        to=fixture['exec']['address'],
+        sender=fixture['exec']['caller'],
+        value=fixture['exec']['value'],
+        data=fixture['exec']['data'],
+        code=code,
+    )
+
+
+def test_vm_fixtures(fixture, vm_class):
+    for computation_getter in (fixture_to_computation, fixture_to_bytecode_computation):
```
pipermerriam: Instead of having a `for` loop in the test, I would prefer parametrizing the test itself using `pytest.mark.parametrize` to keep the two methods of execution separated.

```python
@pytest.mark.parametrize(
    'computation_getter',
    (
        fixture_to_bytecode_computation,
        fixture_to_computation,
    ),
)
def test_vm_fixtures(fixture, vm_class, computation_getter):
    ...
```

PR #316, opened 2018-01-26
Trinity filesystem utils

pipermerriam:
### What was wrong?

There are some filesystem utility functions that are needed in trinity.

### How was it fixed?

Separated them out into their own pull request to reduce overall size of https://github.com/ethereum/py-evm/pull/310 

#### Cute Animal Picture

screen shot 2018-01-26 at 3 09 10 pm


trinity/utils/filesystem.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def is_under_path(base_path, path, strict=True):
+    if is_same_path(base_path, path):
+        if strict:
+            return False
+        else:
+            return True
+    absolute_base_path = os.path.abspath(base_path)
+    absolute_path = os.path.abspath(path)
```
carver: Probably want to `expanduser` here, too?

Looks like a `_normalize_path = compose(os.path.abspath, os.path.expanduser)` might be useful in this module.

trinity/utils/filesystem.py
```diff
@@ -0,0 +1,22 @@
+import os
+
+
+def is_same_path(p1, p2):
+    n_p1 = os.path.abspath(os.path.expanduser(p1))
+    n_p2 = os.path.abspath(os.path.expanduser(p2))
```
carver: *I'm rambling to myself, below. I think I resolved this to: take no action, but just in case I missed something...*

> `expandvars` doesn't make sense here, right? Anything supplied with a command line argument or variable with something like `$HOME` would get expanded before we see it. Plus, `expandvars` runs the risk of someone actually having a folder called `$HOME` and messing things up. Ok, I talked myself out of it making sense.

trinity/utils/filesystem.py
```diff
@@ -0,0 +1,22 @@
...
+
+    try:
+        return os.path.samefile(n_p1, n_p2)
+    except FileNotFoundError:
+        return n_p1 == n_p2
+
+
+def is_under_path(base_path, path, strict=True):
```
carver: `base_path` was a confusing name for me here. I thought it meant the opposite, on the first read. Maybe part of the reason is that I was expecting arguments in reverse order: `(parent_path, maybe_child_path, strict=True)`

How will this get used eventually? Now that you've got me looking for currying everywhere, I imagine the more curry-friendly version would be reverse the order of these args. Like `is_in_data_dir = is_under_path(data_dir)`.

pipermerriam:
This should be against the `trinity` branch.

PR #317, opened 2018-01-26
Trinity hex encoding decoding utils

pipermerriam:
### What was wrong?

Trinity needs it's own hexadecimal encoding/decoding utils

### How was it fixed?

Copy/paste from the py-evm ones.  This is a stand alone to help make https://github.com/ethereum/py-evm/pull/310 smaller.

#### Cute Animal Picture

cute-tiny-animals05


trinity/utils/hexadecimal.py
```diff
@@ -0,0 +1,12 @@
...
+import codecs
+
+
+def encode_hex(value):
+    return '0x' + codecs.decode(codecs.encode(value, 'hex'), 'utf8')
+
+
+def decode_hex(value):
+    _, _, hex_part = value.rpartition('x')
+    return codecs.decode(hex_part, 'hex')
```
carver: Why can't we use the eth-utils ones here (and in py-evm)? Seems like a lot of test and implementation duplication.

trinity/utils/filesystem.py
```diff
@@ -0,0 +1,8 @@
+import os
+
+
+def ensure_path_exists(dir_path):
+    """
+    Make sure that a path exists
+    """
+    os.makedirs(dir_path, exist_ok=True)
```
carver: This seems so small that it's worth inlining. Then we can delete the tests, too.

pipermerriam: Yeah, this was a carry-over from python2 support that was necessary but is no longer.  👍

trinity/utils/hexadecimal.py
pipermerriam: I've been considering the `eth_utils` ones off-limits until we convert API with the new non-`force_text` based versions.  Also, because of the extra handling for both text and binary types they have extra overhead which was why I kept them out of py-evm (may not be necessary anymore, decision from a long time ago).

I don't love all the duplicate utils, but hate it less than I hate using the `eth_utils` versions prior to converting them....  dunno, I could go either way.  I'll sleep on it and if you feel strongly I'll take your lead.

pipermerriam:
This should be against the `trinity` branch.

PR #327, opened 2018-01-29
Add transaction_context object to Computation

lrettig:
Creating a PR to track this work in progress. From #262.

### DIFF
--- a/evm/transaction_context.py
+++ b/evm/transaction_context.py
@@ -12,11 +12,10 @@ class BaseTransactionContext(object):
     _gas_price = None
     _origin = None
 
-    def __init__(self, gas_price, origin=None):
+    def __init__(self, gas_price, origin):
         validate_uint256(gas_price, title="TransactionContext.gas_price")
         self._gas_price = gas_price
-        if origin is not None:
-            validate_canonical_address(origin, title="TransactionContext.origin")
+        validate_canonical_address(origin, title="TransactionContext.origin")
         self._origin = origin
 
     @property
--- a/evm/transaction_context.py
+++ b/evm/transaction_context.py
@@ -1,0 +1,28 @@
+from evm.validation import (
+    validate_canonical_address,
+    validate_uint256,
+)
+
+
+class BaseTransactionContext(object):
+    """
+    This immutable object houses information that remains constant for the entire context of the VM
+    execution.
+    """
+    _gas_price = None
+    _origin = None
+
+    def __init__(self, gas_price, origin=None):
+        validate_uint256(gas_price, title="TransactionContext.gas_price")
+        self._gas_price = gas_price
+        if origin is not None:
+            validate_canonical_address(origin, title="TransactionContext.origin")
+        self._origin = origin
+
+    @property
+    def gas_price(self):
+        return self._gas_price
+
+    @property
+    def origin(self):
+        return self._origin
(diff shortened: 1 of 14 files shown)


evm/transaction_context.py
```diff
@@ -0,0 +1,28 @@
...
+)
+
+
+class BaseTransactionContext(object):
+    """
+    This immutable object houses information that remains constant for the entire context of the VM
+    execution.
+    """
+    _gas_price = None
+    _origin = None
```
pipermerriam: I don't think we need the `@property` versions of these.  Should be good just using them directly as properties.

lrettig: I copied the behavior of https://github.com/hwwhww/py-evm/blob/66a858f6ed1b4a02e35c311ed3255440499c8f9a/evm/vm/execution_context.py. Should this be different for some reason?

hwwhww: The reason for why we protect fields of `ExecutionContext` is that we want to make sure the fields **only be assigned in `__init__`** and won't be changed in other places.
But I'm not sure what's the use cases of `TransactionContext`.

lrettig: @pipermerriam originally said:

> This would be an object that we treat as immutable

So it should be the same, no?

pipermerriam: @hwwhww `@property` methods don't actually provide any protection from mutation *unless* you create a `setter` which throws.  I'd be good with doing that but it should probably be done with our own decorator that automatically creates the `setter` function.  Something like a new decorator:

```python
@immutable_property
def something(self):
    ...
```

hwwhww: @pipermerriam 
So sorry for that I just realized I missed and haven't seen this comment(https://github.com/ethereum/py-evm/pull/273#discussion_r161866203) 15 days ago! Agree with you about using explicit `@read_only_property` or `@immutable_property` decorator.

pipermerriam: No big deal.  Easy enough for us to make that cleanup now (mind opening an issue for it?)

@lrettig I'm fine with leaving current `@property` methods in place and cleaning them up with a stand-alone PR with the `@immutable_property/read_only_property` decorator.

lrettig: Sounds good, thanks for the clarification

tests/json-fixtures/test_state.py
```diff
@@ -295,6 +302,8 @@ def test_state_fixtures(fixture, fixture_vm_class):
         computation, _ = vm.apply_transaction(transaction)
     except ValidationError as err:
         transaction_error = err
+        LOGGER.warn("Got transaction error:")
```
lrettig: @pipermerriam This is a little messy, not sure if this is how logging should happen inside tests, but this made my life much easier when running tests as these errors were being swallowed silently. Let me know if this is okay.

I also want to understand: can a transaction raise a `TransactionError` but the test still pass? Are there tests which expect an error?

pipermerriam: There is no `TransactionError` in py-evm that I know of so I'm not entirely sure what you're referring to here.

In the case of this section of code, yes, transactions *can* error and the tests still pass **because** some of the tests are designed to throw errors, testing invalid transactions do not succeed to execute in the EVM.

lrettig:
Perfect. Thanks. That's what I meant.

PR #328, opened 2018-01-30
p2p: Redesign sub-protocol msg processing

gsalgado:
Now BasePeer only handles P2P msgs (via the handle_p2p_msg()) method, and
places sub-proto messages onto a queue to be consumed by our services (e.g.
LightChain)

Also get rid of the Command.handle() method; last step in order to have
Protocol/Command responsible only for encoding/decoding of data.

### DIFF
--- a/evm/p2p/lightchain.py
+++ b/evm/p2p/lightchain.py
@@ -36,12 +36,14 @@ from evm.rlp.receipts import Receipt
 from evm.utils.keccak import keccak
 from evm.p2p.exceptions import (
     EmptyGetBlockHeadersReply,
+    HandshakeFailure,
     LESAnnouncementProcessingError,
+    PeerFinished,
     StopRequested,
     TooManyTimeouts,
+    UnexpectedMessage,
 )
 from evm.p2p import les
-from evm.p2p import protocol
 from evm.p2p.peer import (
     BasePeer,
     LESPeer,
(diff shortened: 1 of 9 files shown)


evm/p2p/lightchain.py
```diff
@@ -59,25 +60,43 @@ class LightChain(Chain):
...
+        """
+        try:
+            await peer.do_sub_proto_handshake()
+        except HandshakeFailure as e:
+            self.logger.info(str(e))
+            return
+        await self._handle_peer(peer)
+        self.logger.info("Peer %s finished", peer)
+
+    async def _handle_peer(self, peer: LESPeer) -> None:
```
pipermerriam: This function is confusing me.  I can't tell what it's supposed to be doing or why it's doing what it is doing.

It seems to only handle the `les.Announce` command, silently ignoring all others and discarding them indefinitely untill `PeerFinished` kills the loop ad disconnects the peer.  The silently ignoring other commands is the confusing part for me.

evm/p2p/lightchain.py
```diff
@@ -59,25 +60,43 @@ class LightChain(Chain):
...
+
+    async def handle_peer(self, peer: LESPeer) -> None:
+        """Handle the lifetime of the given peer.
+
+        Returns when the peer is finished or when the LightChain is asked to stop.
+        """
+        try:
+            await peer.do_sub_proto_handshake()
+        except HandshakeFailure as e:
+            self.logger.info(str(e))
```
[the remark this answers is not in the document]
gsalgado: Yeah, handshake failures are expected and happen all the time (different chains, too many peers, etc), so `loger.debug` is more appropriate

evm/p2p/lightchain.py
gsalgado: That's because (apart from `Announce`) we implement only the LES commands for retrieving data, and since those will always come in response to requests we make they are consumed by `LESPeer.handle_sub_proto_msg()`. I'll add a comment to that effect, and change the code to raise an `UnexpectedMessage` if we get anything other than an `Announce` as that would indicate a bug in the peer's handling of sub-proto msgs

PR #334, opened 2018-01-31
p2p: PeerPool now has subscribers that get notified of new peers

gsalgado:
Instead of taking a single callback to be called when a new peer is
added, PeerPool now can have multiple subscribers that will be notified
when new peers are added.

Also LightChain is no longer responsible for creating/running a PeerPool
itself

This way we can share a single PeerPool among multiple p2p services.

### DIFF
--- a/evm/p2p/lightchain.py
+++ b/evm/p2p/lightchain.py
@@ -7,6 +7,7 @@ from typing import (  # noqa: F401
     Dict,
     List,
     Optional,
+    Set,
     Tuple,
     Type,
 )
@@ -33,7 +34,6 @@ from evm.rlp.headers import BlockHeader
 from evm.rlp.receipts import Receipt
 from evm.p2p.exceptions import (
     EmptyGetBlockHeadersReply,
-    HandshakeFailure,
     LESAnnouncementProcessingError,
     PeerFinished,
     StopRequested,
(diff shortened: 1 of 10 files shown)


evm/p2p/peer.py
```diff
@@ -710,8 +722,8 @@ def __init__(self,
             if peer is not None:
                 self.logger.info("Successfully connected to %s", peer)
                 self.connected_nodes[peer.remote] = peer
-                asyncio.ensure_future(peer.run(finished_callback=self._peer_finished))
-                self.new_peer_callback(peer)
+                for subscriber in self._subscribers:
+                    subscriber.register_peer(peer)
```
pipermerriam: Any thoughts on allowing subscribers to unsubscribe themselves by doing something like raising a certain exception.  No clue if it's even useful, but it did jump out to me that there is no mechanism for removing a subscription.

gsalgado: Yeah, that will be necessary as we'll have services that will be short-lived (e.g the `StateDownloader`, which I just realized has been broken by those changes), but I believe it'd be better for them to explicitly unsubscribe once they're finished.

I'll try to come up with a simple test for `StateDownloader`, fix it and see what an unsubscribe API would look like.

gsalgado:
Unit testing `StateDownloader` is going to take a nontrivial amount of work, and at this stage I'm not sure it's worth it, so I've added the type annotations that were missing (which would've helped catch the breakage introduced by this `PeerPool` redesign). I hope you're ok with that

Also, while manually testing it I realized it makes more sense to perform the sub-proto handshake as part of the handshake() function, so that we can only call `Peer.run()` once we've completed both the P2P and sub-proto handshakes. And I've moved the decoding of messages to `Peer.read_msg()`, to avoid duplication.

Given those extra changes, it'd be great if you could have another look at this, @pipermerriam

pipermerriam:
> I hope you're ok with that

I am.  Fast and loose for now with the understanding that backfilling tests when we actually know how to test the things will be more productive (and that we're in an alpha stage so it's kind of ok if things break in weird ways right now).

evm/p2p/lightchain.py
```diff
@@ -331,28 +336,41 @@ def _test():
...
     try:
         chaindb.get_canonical_head()
     except CanonicalHeadNotFound:
         # We're starting with a fresh DB.
-        chain = DemoLightChain.from_genesis_header(chaindb, GENESIS_HEADER)
+        chain = DemoLightChain.from_genesis_header(chaindb, GENESIS_HEADER, peer_pool)
     else:
         # We're reusing an existing db.
-        chain = DemoLightChain(chaindb)
+        chain = DemoLightChain(chaindb, peer_pool)
```
pipermerriam: Note to self.  This `try/except/else` is a bit cumbersome and I found myself replicating it in like 2-3 places so far.  going to try and figure out a more elegant/expressive/intuitive way to implement this logic.

evm/p2p/state.py
```diff
@@ -43,28 +46,39 @@ class StateDownloader:
...
+
+    async def handle_peer(self, peer: ETHPeer) -> None:
+        """Handle the lifecycle of the given peer."""
+        self._running_peers.add(peer)
+        while not self._should_stop.is_set():
+            try:
+                cmd, msg = await peer.read_sub_proto_msg()
+            except PeerFinished:
+                break
+            if isinstance(cmd, eth.NodeData):
```
pipermerriam: Just another case of an `if` with no `else`.  Should we at least debug log something when `cmd` is not `eth.NodeData`?  If it's truly a case where we want to ignore the other messages, maybe we log them anyways and add a comment explaining why we ignore them?

evm/p2p/state.py
```diff
@@ -43,28 +46,39 @@ class StateDownloader:
...
+                    node_key = keccak(node)
+                    try:
+                        self.scheduler.process([(node_key, node)])
+                    except SyncRequestAlreadyProcessed:
+                        # This means we received a node more than once, which can happen when we
+                        # retry after a timeout.
+                        pass
+                    # A node may be received more than once, so pop() with a default value.
+                    self._pending_nodes.pop(node_key, None)
+        self._running_peers.remove(peer)
```
pipermerriam: Just curious if it makes sense for this to be embedded in a `try/finally`.  Not sure how critical it is that this get removed from the `_running_peers` in the case of an unforseen error?

evm/p2p/state.py
gsalgado: When performing a state download we want to pause the chain sync anyway, so for simplicity I'm considering the possibility of ignoring everything that is not a NodeData here. Not sure that's going to work out, but either way I agree we should at least log and have a comment explaining that

evm/p2p/state.py
gsalgado: Yeah, it definitely makes sense. Fixed

evm/p2p/lightchain.py
carver: Maybe something like?

```py
chain = DemoLightChain.load(chaindb, peer_pool, default_genesis=ROPSTEN_GENESIS_HEADER)
```

PR #341, opened 2018-02-01
Piper/trinity hex encoding decoding utils

pipermerriam:
Replaces https://github.com/ethereum/py-evm/pull/317

### What was wrong?

Trinity needs it's own hexadecimal encoding/decoding utils

### How was it fixed?

Copy/paste from the py-evm ones.  This is a stand alone to help make https://github.com/ethereum/py-evm/pull/310 smaller.

#### Cute Animal Picture

cute-tiny-animals05

### DIFF
--- a/tests/trinity/hexadecimal-utils/test_encode_and_decode.py
+++ b/tests/trinity/hexadecimal-utils/test_encode_and_decode.py
@@ -1,0 +1,62 @@
+from __future__ import unicode_literals
+
+import pytest
+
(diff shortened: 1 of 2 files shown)


carver:
why not use the eth-utils version?

pipermerriam:
> why not use the eth-utils version?

I'll convert it to use them as soon as the `eth-utils` versions aren't using `force_text` anywhere (which I think is underway)

PR #345, opened 2018-02-02
Implement CancelToken and use it everywhere

gsalgado:
CancelTokens should be passed to any async operation, and whenever
    they are trigger()ed they cause those operations to be cancelled. It was
    inspired by http://goo.gl/MUFMKU and right now is used mainly to ensure
    we cancel pending operations when terminating services

### DIFF
--- a/evm/p2p/cancel_token.py
+++ b/evm/p2p/cancel_token.py
@@ -1,0 +1,97 @@
+import asyncio
+from typing import Any, Awaitable, List  # noqa: F401
+
+from evm.p2p.exceptions import OperationCancelled
+
+
+class CancelToken:
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+        self._chain = []  # type: List['CancelToken']
+        self._triggered = asyncio.Event()
+
+    def chain(self, token: 'CancelToken') -> 'CancelToken':
+        """Return a new CancelToken chaining this and the given token.
+
+        The new CancelToken's triggered will return True if trigger() has been
+        called on either of the chained tokens, but calling trigger() on the new token
+        has no effect on either of the chained tokens.
+        """
+        chain_name = ":".join([self.name, token.name])
+        chain = CancelToken(chain_name)
+        chain._chain.extend([self, token])
+        return chain
+
+    def trigger(self) -> None:
+        self._triggered.set()
+
+    @property
+    def triggered_token(self) -> 'CancelToken':
+        if self._triggered.is_set():
+            return self
+        for token in self._chain:
+            if token.triggered:
+                # Use token.triggered_token here to make the lookup recursive as self._chain may
+                # contain other chains.
+                return token.triggered_token
+        return None
+
+    @property
+    def triggered(self) -> bool:
+        if self._triggered.is_set():
+            return True
+        return any(token.triggered for token in self._chain)
+
+    async def wait(self) -> None:
+        if self.triggered_token is not None:
+            return
+
+        futures = [asyncio.ensure_future(self._triggered.wait())]
+        for token in self._chain:
+            futures.append(asyncio.ensure_future(token.wait()))
+
+        def cancel_pending(f):
+            for future in futures:
+                if not future.done():
+                    future.cancel()
+
+        fut = asyncio.ensure_future(_wait_for_first(futures))
+        fut.add_done_callback(cancel_pending)
+        await fut
+
+    def __str__(self):
+        return self.name
+
+
+async def _wait_for_first(futures):
+    for future in asyncio.as_completed(futures):
+        await future
+        return
+
+
+async def wait_with_token(future: Awaitable,
+                          cancel_token: CancelToken,
+                          timeout: float = None) -> Any:
+    """Wait for future to complete, unless we timeout or the cancel token is triggered.
+
+    Returns the future's result in case it completes.
+
+    Raises TimeoutError if we timeout or OperationCancelled if the cancel token is triggered.
+    """
+    done, pending = await asyncio.wait(
+        [future, cancel_token.wait()],
+        timeout=timeout,
+        return_when=asyncio.FIRST_COMPLETED)
+    for task in pending:
+        task.cancel()
+    if not done:
+        raise TimeoutError()
+    if cancel_token.triggered_token is not None:
+        # We've been asked to cancel so we don't care about our future, but we must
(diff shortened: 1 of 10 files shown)


evm/p2p/cancel_token.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class CancelToken:
+
+    def __init__(self, name, other_token=None):
+        self.name = name
+        self._cancel_requested = asyncio.Event()
+        self.other_token = other_token
+
+    # XXX: Maybe rename to chain()?
```
pipermerriam: 👍

evm/p2p/cancel_token.py
```diff
@@ -0,0 +1,59 @@
...
+    def combine(self, other_token):
+        """Return a new CancelToken combining this and the other token.
+
+        The new CancelToken's cancel_requested will return True if request_cancellation() has been
+        called on either of the combined tokens, but calling request_cancellation() on this token
+        has no effect on either of the combined tokens.
+        """
+        return CancelToken(self.name + other_token.name, other_token)
```
pipermerriam: Maybe join the names with a separator `|` or `:` seem appropriate.

pipermerriam:
Just noticed it's still WIP so @hwwhww @jannikluhn you can hold off for now.

gsalgado:
It is ready for review now, @pipermerriam. The only change since your review was a reorg of the arguments in `LESPeer._get_proof()`, to address an XXX comment. @hwwhww or @jannikluhn, feel free to review whenever you have time :-)

evm/p2p/cancel_token.py
```diff
@@ -0,0 +1,90 @@
...
+        timeout=timeout,
+        return_when=asyncio.FIRST_COMPLETED)
+    for task in pending:
+        task.cancel()
+    if not done:
+        raise TimeoutError()
+    if cancel_token.triggered_token is not None:
+        # We've been asked to cancel so we don't care about our future, but we must
+        # consume its exception or else asyncio will emit warnings.
+        [task.exception() for task in done]
```
pipermerriam: I think this would be better expanded into a two line `for` loop.  Changes this from a *side-effect* based operation to something more explicit.

evm/p2p/lightchain.py
```diff
@@ -106,10 +103,13 @@ def register_peer(self, peer: BasePeer) -> None:
         await peer.stop()
 
     async def wait_until_finished(self):
+        count = 0
         while self._running_peers:
```
pipermerriam: What do you think about this.

```python
for _ in range(50):
    if not self._running_peers:
        break
    self.logger.debug("Waiting for %d running peers to finish", len(self._running_peers))
    await asyncio.sleep(0.1)
else:
    self.logger.info("Waited too long for peers to finish, exiting anyway")
```

pipermerriam: Alternate to remove the *weird* magic number `50`

```python
start_at = time.time()
while time.time() < start_at + 5:
    ...
else:
    ...
```

Makes it a bit more explicit how long the shutdown process is allowed to wait for.

evm/p2p/peer.py
```diff
@@ -489,65 +494,72 @@ def callback(r):
...
-        header = await self.get_block_header_by_hash(block_hash)
+        proof = await self._get_proof(cancel_token, block_hash, account_key=b'', key=key)
+        header = await self.get_block_header_by_hash(block_hash, cancel_token)
         rlp_account = HexaryTrie.get_from_proof(header.state_root, key, proof)
         return rlp.decode(rlp_account, sedes=Account)
 
-    async def _get_proof(self, block_hash: bytes, account_key: bytes, key: bytes,
-                         from_level: int = 0) -> List[bytes]:
+    async def _get_proof(
+            self, cancel_token: CancelToken, block_hash: bytes, account_key: bytes, key: bytes,
```
pipermerriam: *readability*

split these to one argument per line?

evm/p2p/cancel_token.py
```diff
@@ -0,0 +1,90 @@
...
+        self._triggered = asyncio.Event()
+
+    def chain(self, next_token: 'CancelToken') -> 'CancelToken':
+        """Return a new CancelToken chaining this and next_token.
+
+        The new CancelToken's triggered will return True if trigger() has been
+        called on either of the chained tokens, but calling trigger() on this token
+        has no effect on either of the chained tokens.
+        """
+        return CancelToken(":".join([self.name, next_token.name]), next_token)
```
jannikluhn: It's confusing to me that the resulting token seems to have no reference to `self` (besides the name of course). Shouldn't chaining be just something like `self.next = next_token`?

evm/p2p/cancel_token.py
```diff
@@ -0,0 +1,90 @@
...
+        def cancel_pending(f):
+            for future in futures:
+                if not future.done():
+                    future.cancel()
+
+        fut = asyncio.ensure_future(self._wait_for_first(futures))
+        fut.add_done_callback(cancel_pending)
+        await fut
+
+    async def _wait_for_first(self, futures):
```
jannikluhn: Can this be a standalone utils function or does the `async` mean that it has to be part of the CancelToken class?

pipermerriam: the `async def` is what makes it a coroutine which is in-turn what allows you to call `await token._wait_for_first(...)`.

The `await` is what yields execution to other processes allow them to continue execution while this function would normally block.

@gsalgado can probably correct me if I'm wrong here.

evm/p2p/lightchain.py
gsalgado: Went with the latter

evm/p2p/cancel_token.py
gsalgado: @pipermerriam's explanation is spot on, but this method can indeed be a standalone function, although I'd prefer to keep it in this module for now

evm/p2p/cancel_token.py
gsalgado: Good catch; this is indeed a bug! It was not caught by the tests and didn't affect our code because they were always triggering the last token on the chain. I've added more tests and changed `CancelToken` to keep a list of chained tokens instead of a single reference to the next item on the chain, as when chaining we want to create a new token that can be triggered separately from the original one

gsalgado:
@jannikluhn @pipermerriam would you guys like to have another look or is this good to go once CI is working again?

Issue #304, opened 2018-01-24
Create EVM Process for Trinity


pipermerriam:
### What is wrong?

When Trinity is running a chain it will likely have the following *processes*.

- DevP2P
  - Handles networking
  - Makes requests to the EVM process to read existing chain state.
  - Submits new data for import to the EVM process.
- JSON-RPC
  - Serves requests to the JSON-RPC server over the IPC pipe
  - Makes requests to the EVM process to read existing chain state.
  - Submits new transactions to the DevP2P process for broadcast to the network.
  - Submits requests for EVM execution for things like `eth_call` and `eth_estimateGas`.
- EVM
  - Serves requests from the JSON-RPC process for thing like `eth_call` and `eth_estimateGas`.
  - Serves requests from the DevP2P process for reading chain data.
  - Handle new chain data submitted by the DevP2P process.

There may be a fourth process for the transaction pool as well as others we haven't thought of yet.

### How can it be fixed

Each of these process will be it's own `multiprocessing.Process` instance.  We have limited communication options between these processes using the primatives from the `multiprocessing` module (`Pipe/Queue/..?..`).

Exactly how this all looks still needs to be hashed out, after which we can fill this section in further.

pipermerriam:
Here are some extra thoughts on this topic.

The EVM process will be a loose wrapper around the `Chain` object and the underlying `ChainDB`.  The wrapper should probably be written using all of the nice `async` primatives so that it'll sleep while requests and responses are processing.

I believe most of the inter-process communication is going to be over `multiprocessing.Pipe` connections.  We'll need to setup pipes between all of the processes which need to communicate with each other prior to initializing the processes.  Then, on each side, we'll need something to manage the connection which dispatches incoming requests to the proper APIs and returns appropriate responses to each request.

One issue that we'll need to address is that objects which pass through pipes are pickled for transport.  We should verify that a cleverly constructed transaction or block (or other externally received data) cannot exploit this for remote code execution.  At a glance, the documentation appears to say that it's safe, but we should be **very** sure.

pipermerriam:
This is probably a dependency for #303 being completed but I suspect some of it can happen in parallel.

pipermerriam:
Also, the reason for using `multiprocessing.Process` instances for everything is to pro-actively prevent the systemic problem that pyethapp faced.

[...]

Issue #302, opened 2018-01-24
Add CLI command for entering into a python REPL


pipermerriam:
### What is wrong?

We need a CLI command for entering into a REPL connected to the chain.

### How can it be fixed

This functionality is likely to evolve but initially, anything that leverages the foundation laid out in #300 to allow the user to use `$ trinity console` which drops them into a python REPL with some objects in the local context which allow them to interact with the chain works for now.

- detection of IPython is a plus.
- ensure that importing local python modules works.
- some minimal approach to testing would be nice.  Doesn't need to be comprehensive, but ideally something that demonstrates the basic *happy-path* functions as expected.

gitcoinbot:
__This issue now has a funding of 0.25 ETH (256.74 USD) attached to it.__

 * If you would like to work on this issue you can claim it [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/302).
 * If you've completed this issue and want to claim the bounty you can do so [here](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/302)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $16114.21 more Funded OSS Work Available at: https://gitcoin.co/explorer

pipermerriam:
I expect testing this is going to be messy.  Please feel free to run a plan by my before spending the time implementing it to be sure it's something we can merge.

pipermerriam:
@amitt001 please claim this using the gitcoin link above at your ealiest convenience

amitt001:
@pipermerriam Getting a very poor internet connection as I am travelling for next 2 days. I will claim this and also discuss the plan with you as soon as I get a stable network.

gitcoinbot:
__The funding of 0.25 ETH (266.0 USD) attached has been shown interest by: [@amitt001](https://gitcoin.co/profile/amitt001).__ 

 If you are interested, please leave a comment to let the funder  and the other parties involved why you're interested in working on this issue and your plans to resolve it.  If you don't leave a comment, the funder may not think you're too interested. 

 * Learn more [on the gitcoin issue page](https://gitcoin.co/funding/details?url=https://github.com/ethereum/py-evm/issues/302)
 * Questions? Get help on the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
 * $41154.38 more Funded OSS Work Available at: https://gitcoin.co/explorer

[...]

PR #346, opened 2018-02-03
Initial support for trinity REPL

amitt001:
### What was wrong?
Implements #302. Ports the initial work done in #298 

### How was it fixed?
Added support for Python REPL to trinity CLI. Running `trinity console` starts either an IPython, if it is installed, or a Python REPL. The shell comes loaded with chain variable that can be used while it's syncing. Running `trinity --log-level info console --vanilla-shell` sets the log level as `INFO` and run the default Python shell even if IPythin is installed on the system.

Chain syncing continues in the background while the shell is running in the foreground. Starting the REPL starts the chain syncing in a thread instead of a process. This is unlike trinity CLI which starts syncing in a separate process.

The reason for using multithreaded syncing instead of multiprocess is when we use a Python REPL we need to share the same `chain` object across sync process and the REPL so that user can interact with the `chain`. The `Chain` object contains coroutines which run in an event loop which is owned by the process that runs it. So a complex object like chain can't be shared across processes through `PIPE/QUEUE` or a [BaseManager](https://docs.python.org/2/library/multiprocessing.html#multiprocessing.managers.BaseManager). 

This might change after things clear out around #304.

#### Cute Animal Picture

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -88,6 +88,17 @@ parser.add_argument(
     )
 )
 
+# Add console sub-command to trinity CLI.
+subparser = parser.add_subparsers(dest='subcommand')
+console_parser = subparser.add_parser('console', help='start the trinity REPL')
+console_parser.add_argument(
+    '--vanilla-shell',
+    action='store_true',
+    default=False,
+    help='start a native Python shell'
+)
+console_parser.set_defaults(func=console)
+
 
 def chain_obj(chain_config, sync_mode):
     if not is_chain_initialized(chain_config):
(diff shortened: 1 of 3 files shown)


trinity/main.py
```diff
@@ -43,6 +45,8 @@
     choices=LOG_LEVEL_CHOICES,
     default=DEFAULT_LOG_LEVEL,
 )
+parser.add_argument('console', nargs='?', default='')
```
pipermerriam: Can you update this to leverage the `add_subparser` API

https://docs.python.org/3.5/library/argparse.html#argparse.ArgumentParser.add_subparsers

amitt001:
@pipermerriam done the review changes.

amitt001:
@pipermerriam Thanks for the review and approval. Happy that my 1st PR is going to be in py-evm/trinity. Looking forward to doing more of it.

PR #347, opened 2018-02-04
Add paygas test cases

NIC619:
### How was it fixed?
Add PAYGAS opcode tests
- Normal case
- Edge cases
    - PAYGAS not triggered
    - triggered with 0 gas price
    - triggered but not in a top level call
    - triggered twice

#### Cute Animal Picture


tests/core/vm/test_PAYGAS_tx_fee.py
```diff
@@ -34,16 +35,162 @@ def test_trigger_PAYGAS(shard_chain_without_block_validation):  # noqa: F811
...
+        data_value=0,
+        data_msgdata=b'',
+        data_vrs=b'',
+        code=PAYGAS_contract_normal['bytecode'],
+    )
+    computation, _ = vm.apply_transaction(normal_PAYGAS_contract_deploy_tx)
+    assert not computation.is_error
+
+    # Trigger the PAYGAS contract with gas price set to 0
+    recipient = decode_hex('0xa94f5374fce5edbc8e2a8697c15331677e6ebf0c')
```
pipermerriam: I'd like to see these divided up into individual test cases rather than the current format (which I think is error prone and harder to follow).

pipermerriam: (or even using @jannikluhn 's fancy new test case format from #350 if you're feeling brave and @jannikluhn doesn't think it's a terrible idea.  Best way to validate an API is to use it in-my-experience.

tests/core/vm/contract_fixture.py
```diff
@@ -4,27 +4,25 @@
...
-simple_transfer_contract_bytecode = b'0x61003e567401000000000000000000000000000000000000000060205260003560205181101558575060006000600060006020356000356000f1155857005b61000461003e0361000460003961000461003e036000f3'  # noqa: E501
-
-
-# address where this contract will be deployed
-simple_transfer_contract_address = generate_CREATE2_contract_address(
-    b'',
-    decode_hex(simple_transfer_contract_bytecode)
-)
+simple_transfer_contract = {
+    'lll_code': ['seq',
```
pipermerriam: I'm wondering if we shouldn't relegate these to their own files in a `tests/fixtures` directory somewhere along with a script to compile them and generate a JSON document with the compiled assets.

If it's possible, we could re-generate those in the CI environment to be sure we're not working with stale fixtures.  However, it'd be fine with me to commit that generated file to git in the short term as long as we have an issue to the the CI setup eventually.

tests/core/vm/contract_fixture.py
```diff
@@ -35,51 +33,96 @@
...
+                        ['lll',
+                            ['seq',
+                                ['mstore', 32, ['sload', 0]],
+                                ['mstore', 64, 118167469832824678325960435],
+                                ['clamp_nonzero', ['create2', ['mload', 32], ['mload', 32], 85, 11]],  # noqa: E501
+                                ['sstore', 0, ['add', ['mload', 32], 1]],
+                                'stop'],
+                        0]]],  # noqa: E128
+    # compiled byte code
+    'bytecode': b'0x610039566000546020526a61beef6020526002603ef3604052600b6055602051602051fe8061002957600080fd5b50600160205101600055005b61000461003903610004600039610004610039036000f3',  # noqa: E501
```
hwwhww: Just confirm with you, are the bytecodes generated by the latest Vyper sharding branch?

tests/core/vm/contract_fixture.py
```diff
@@ -35,51 +33,96 @@
...
+                            ['seq',
+                                ['paygas', ['calldataload', 64]],
+                                # Transfer value
+                                ['mstore', 128, 1461501637330902918203684832716283019655932542976],  # noqa: E501
+                                ['uclamplt', ['calldataload', 0], ['mload', 128]],  # noqa: E501
+                                ['assert', ['call', ['sub', ['gas'], 100000], ['calldataload', 0], ['calldataload', 32], 0, 0, 0, 0]],  # noqa: E501
+                                'stop'],
+                        0]]],  # noqa: E128
+    # compiled byte code
+    'bytecode': b'0x61005356604035f55074010000000000000000000000000000000000000000608052600035608051811061002e57600080fd5b506000600060006000602035600035620186a05a03f161004d57600080fd5b005b61000461005303610004600039610004610053036000f3',  # noqa: E501
```
hwwhww: ditto

tests/core/fixtures.py
```diff
@@ -229,7 +227,6 @@ def chain_without_block_validation():
     genesis_state = {
         funded_addr: {
             'balance': initial_balance,
-            'nonce': 0,
```
hwwhww: It's main chain, so the nonce can't be removed.

tests/core/vm/contract_fixture.py
NIC619: Ah, no, I don't think so. I will have them compiled using Vyper sharding branch.

tests/core/fixtures.py
NIC619: Sorry about that. Will fix it.

tests/core/vm/contract_fixture.py
NIC619: I move these to `tests/contract_fixtures` directory.(Since there's another name `tests/fixtures.py` in conflict. For now I use the name `contract_fixtures` as the fixtures put here at the moment are mainly contract code.)

And add a compile script(`contract_fixture_compile_script.py`) to generate json documents. But I don't know exactly how to automate the process of pre-run the compile script in tox. And also `Vyper` doesn't seem to support Python 3.5. So I currently commit the generated json files together.

Issue #325, opened 2018-01-29
STUB: Remove shuffling cycle from VMC


hwwhww:
### What is wrong?
STUB

### How can it be fixed
STUB

hwwhww:
@mhchia I think this issue has already been fixed via #348? If not, please reopen it.

PR #348, opened 2018-02-04
Remove validation code

mhchia:
### What was wrong?
As the item#3 in https://github.com/ethereum/sharding/issues/37 described, we remove the validation code logics from VMC, to reduce gas consumption.

### How was it fixed?
* Remove validation code construction
* Modify `add_header` to accept parameters directly, instead of the rlp encoded bytes.
* Verify `msg.sender` in `add_header` and `withdraw`, instead of using validation code.

#### Cute Animal Picture


evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -46,8 +41,8 @@
 # Gas limit of the signature validation code
 sig_gas_limit: num
```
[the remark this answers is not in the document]
mhchia: Thank you for pointing out. Will remove it

evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -363,21 +266,28 @@ def add_header(header: bytes <= 4096) -> bool:
...
+    # Emit a log which is equivalent to
+    # CollationAdded: __log__({shard_id: indexed(num), collation_header: bytes <= 4096, is_new_head: bool, score: num})
+    # TODO: should be replaced by `log.CollationAdded`
+    if is_new_head:
+        new_head_in_num = 1
+    else:
+        new_head_in_num = 0
+    raw_log(
+        [
+            sha3("CollationAdded(int128,bytes4096,bool,int128)"),
```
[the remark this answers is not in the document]
mhchia: Yes, actually it matches log signature `CollationAdded(indexed int128 shard, bytes4096 collationHeader, bool isNewHead, int128 score)`.
The reason why I use `int128`(i.e. `num` in vyper) is it is more friendly for calculation. And the reason why I use `raw_log` instead of `log.CollationAdded()` is previously vyper doesn't support bytes type whose length > 32. Actually there is a TODO comment in line 67

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -266,45 +281,58 @@ def get_eligible_proposer(self, shard_id, period=None, gas=None):
...
         return decode_hex(address_in_hex)
 
-    def deposit(self,
-                validation_code_addr,
-                return_addr,
-                gas=None,
-                gas_price=None):
-        """deposit(validation_code_addr: address, return_addr: address) -> num
+    def deposit(self, gas=None, gas_price=None):
+        """deposit() -> num
```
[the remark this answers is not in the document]
mhchia: Will change it. Thank you!

tests/sharding/test_vmc_handler.py
```diff
@@ -90,167 +82,68 @@ def send_raw_transaction(vmc_handler, raw_transaction):
...
-    vmc_handler.withdraw(validator_index, signature)
+    vmc_handler.withdraw(validator_index)
     mine(vmc_handler, 1)
 
 
-def deploy_valcode_and_deposit(vmc_handler, privkey):
+def do_deposit(vmc_handler, privkey):
     """
-    Deploy validation code of and with the privkey, and do deposit
+    Deposit a validator
```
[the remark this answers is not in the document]
mhchia: Ok no problem

tests/sharding/test_vmc_handler.py
```diff
@@ -277,38 +170,42 @@ def mk_testing_colhdr(vmc_handler,
...
+                pad_bytes_to_bytes32(collation_coinbase),
+                post_state_root,
+                receipt_root,
+                int_to_bytes32(number),
+            )
+        )
     )
-    sig = sign(sighash, privkey)
-    return rlp.encode([
+    header_tuple = (
```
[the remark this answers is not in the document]
mhchia: Thank you for pointing it out!
I think it's a good idea to use `evm.rlp.headers.CollationHeader` to build the collation header from logs. I will change it.
But actually the result of `mk_testing_colhdr` is designed to be passed to `VMC.add_header`. How do you think in this condition?
Personally I prefer to keep this behavior here.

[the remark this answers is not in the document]
hwwhww: I'm thinking when the collator calls `VMC.add_header`, the collator would have built a real collation before. So there may be a property in `CollationHeader`:
```python
@property
def add_header_params:
    # returns header fields except `sig`
    ...
```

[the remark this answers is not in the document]
hwwhww: > But actually the result of mk_testing_colhdr is designed to be passed to VMC.add_header. How do you think in this condition?
> Personally I prefer to keep this behavior here.
> Will it help if I change the function name to mk_testing_colhdr_params?

The second output (`header_hash`) of `mk_testing_colhdr` is not the input of `VMC.add_header` but the input of next `mk_testing_colhdr` call.
`header_hash` can be another property of `CollationHeader`. In this way, this `mk_testing_colhdr` function could just return `CollationHeader`, and these two properties could be accessed in `test_vmc_contract_calls`.

hwwhww:
I think this PR also closes #325. ✌🏻

tests/sharding/test_vmc_handler.py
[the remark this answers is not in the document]
mhchia: > header_hash can be another property of CollationHeader. In this way, this mk_testing_colhdr function could just return CollationHeader object, and these two properties could be accessed in test_vmc_contract_calls.

It makes sense. I will change it. Thank you!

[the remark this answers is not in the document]
mhchia: Hold on. So in this case should I add another `header_hash`, which simply returns `keccak(to_bytes32(shard_id) ++ to_bytes32(expected_period_number) ++ ...)`, is that correct?

evm/rlp/headers.py
```diff
@@ -190,15 +197,33 @@ def __repr__(self):
...
     @property
     def hash(self):
-        return keccak(rlp.encode(self))
+        int_to_bytes32 = compose(
+            pad32,
+            int_to_big_endian,
+        )
+        header_hash = keccak(
+            b''.join(
+                (
```
pipermerriam: *whitespace nitpick*

This should be a bit easier to read if you write this as:

```python
b''.join((
    thing_a,
    thing_b,
))
```

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -44,27 +46,35 @@ class UnknownShard(Exception):
     pass
 
 
+@to_dict
+def deserialize_header_bytes(header_bytes):
```
pipermerriam: This function feels like something that should be a `classmethod` on the `CollationHeader` class.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -44,27 +46,35 @@ class UnknownShard(Exception):
...
+    header_sedes = CollationHeader.get_sedes()
+    # assume all fields are padded to 32 bytes
+    obj_size = 32
+    assert len(header_bytes) == obj_size * len(header_sedes)
+    for idx, field in enumerate(CollationHeader.fields):
+        field_name, field_type = field
+        start_index = idx * obj_size
+        field_bytes = header_bytes[start_index:(start_index + obj_size)]
+        if field_type == rlp.sedes.big_endian_int:
+            # remove the leading zeros, to avoid `not minimal length` error in deserialization
```
pipermerriam: is this a bug in the `rlp` library?  If so can you open an issue for it in that repository?

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -158,6 +168,9 @@ def __init__(self, *args, default_privkey, **kwargs):
 
         super().__init__(*args, **kwargs)
 
+    def get_default_sender_address(self):
```
pipermerriam: This function doesn't appear to be used anywhere.

evm/vm/forks/sharding/contracts/validator_manager.v.py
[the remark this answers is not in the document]
enriquefynn: Can the `collationHeader` argument from `CollationAdded` be unpacked?

So instead of using `CollationAdded(indexed int128 shard, bytes4096 collationHeader, bool isNewHead, int128 score)`
Use `CollationAdded(indexed int128 shard, int128 expectedPeriodNumber, bytes32 periodStartPrevHash, bytes32 _parentCollationHash, bytes32 txListRoot, address collationCoinbase, bytes32 postStateRoot, bytes32 receiptRoot, int128 collationNumber, bool isNewHead, int128 score)`

The benefits would be: 1 int less for the log: `shard`; easier to use with the ABI.

evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -242,115 +191,65 @@ def get_eligible_proposer(shard_id: num, period: num) -> address:
...
+        period_start_prevhash,
+        parent_hash,
+        transaction_root,
+        as_bytes32(collation_coinbase),
+        state_root,
+        receipt_root,
+        as_bytes32(collation_number),
+    )
+    entire_header_hash = sha3(header_bytes)
     assert entire_header_hash != as_bytes32(0)
```
enriquefynn: Unnecessary assert

mhchia: You're right. Thank you for pointing out.

evm/vm/forks/sharding/contracts/validator_manager.v.py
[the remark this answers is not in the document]
mhchia: I think it's a good point. I will see if there are specific reason for using `CollationAdded(indexed int128 shard, bytes4096 collationHeader, bool isNewHead, int128 score)`.
Thank you for the suggestion.

evm/vm/forks/sharding/vmc_handler.py
mhchia: I'm not sure. It just requires the `big_endian_int` to be in minimal length (e.g. `b'\x00\x01'` should be `b'\x01'`)
The error is like this
```shell
self = <rlp.sedes.big_endian_int.BigEndianInt object at 0x111d6bb38>
serial = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x06'

    def deserialize(self, serial):
        if self.l is not None and len(serial) != self.l:
            raise DeserializationError('Invalid serialization (wrong size)',
                                       serial)
        if self.l is None and len(serial) > 0 and serial[0:1] == ascii_chr(0):
            raise DeserializationError('Invalid serialization (not minimal '
>                                      'length)', serial)
E           rlp.exceptions.DeserializationError: Invalid serialization (not minimal length)

/home/xxx/.pyenv/versions/py36env/lib/python3.6/site-packages/rlp/sedes/big_endian_int.py:40: DeserializationError
```
Will dig more into it to see if it is an issue.
Thank you:)

evm/vm/forks/sharding/vmc_handler.py
mhchia: Actually I used them in `test_vmc_handler.py`, to get the `default_sender_address` for doing transactions.
Should I remove them? I can replace them by pass the address in use as argument in those testing functions in `test_vmc_handler.py`

evm/vm/forks/sharding/vmc_handler.py
mhchia: Thank you for this great suggestion! I will change it.

evm/rlp/headers.py
mhchia: You're right. I will change it.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -266,45 +253,58 @@ def get_eligible_proposer(self, shard_id, period=None, gas=None):
...
-                sig,
             ],
             gas=gas,
             gas_price=gas_price,
         )
         return tx_hash
 
-    def add_header(self, header, gas=None, gas_price=None):
-        """add_header(header: bytes <= 4096) -> bool
+    def add_header(self,
```
hwwhww: `VMC.add_header` takes a lot of parameters and combines `add_header` inputs and gas setting in the same level. What do you think about pass `CollationHeader` to this function and move logic of `add_header_send_tx` to here so that the params would be clearer:
```python
def add_header(self, collation_header, gas=None, gas_price=None)
```

mhchia: Good idea! I will change it. Thank you.

tests/sharding/test_vmc_handler.py
```diff
@@ -90,167 +79,67 @@ def send_raw_transaction(vmc_handler, raw_transaction):
...
-def mk_initiating_contracts(sender_privkey,
-                            sender_starting_nonce,
-                            TransactionClass,
-                            gas_price):
+def mk_initiating_transactions(sender_privkey,
+                               sender_starting_nonce,
+                               TransactionClass,
+                               gas_price):
     """Make transactions of createing initial contracts
     Including rlp_decoder, sighasher and validator_manager
```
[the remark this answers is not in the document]
mhchia: No problem. Thank you for discovering it.

mhchia:
@enriquefynn I have changed the signature of log `CollationAdded`. Please check it out if you have spare time.
Thank you for the great suggestion!

evm/rlp/headers.py
```diff
@@ -196,15 +207,57 @@ def __repr__(self):
 
     @property
     def hash(self):
-        return keccak(rlp.encode(self))
+        int_to_bytes32 = compose(
```
pipermerriam: I'd be good with this living in `evm.utils.numeric`.  I suspect there are other places in the code where we'd use it.

evm/rlp/headers.py
```diff
@@ -196,15 +207,57 @@ def __repr__(self):
...
+        )
+        return header_hash
+
+    @classmethod
+    @to_dict
+    def _deserialize_header_bytes_to_dict(cls, header_bytes):
+        header_sedes = cls.get_sedes()
+        # assume all fields are padded to 32 bytes
+        obj_size = 32
+        assert len(header_bytes) == obj_size * len(header_sedes)
```
pipermerriam: Assertions are generally bad for runtime code as they can be easily disabled.  This should probably raise a `ValidationError`

evm/vm/forks/sharding/vmc_handler.py
pipermerriam: I must have missed it.  disregard.

evm/rlp/headers.py
mhchia: Have changed it. Thank you!

evm/rlp/headers.py
mhchia: I have moved it to `evm.utils.numeric`, and added two tests for it.
Hope that would help.

mhchia:
@pipermerriam Could you help me to review the newly added commits if you have spare time? Since I added the tests for `int_to_bytes32`, not sure if they are okay for merging.
Really appreciate it. :)

evm/rlp/headers.py
```diff
@@ -229,10 +224,12 @@ def hash(self):
     @classmethod
     @to_dict
     def _deserialize_header_bytes_to_dict(cls, header_bytes):
-        header_sedes = cls.get_sedes()
         # assume all fields are padded to 32 bytes
         obj_size = 32
-        assert len(header_bytes) == obj_size * len(header_sedes)
+        if len(header_bytes) != obj_size * len(cls.fields):
+            raise ValidationError(
+                "Length of header bytes should be equal to obj_size * len(cls.fields)"
```
pipermerriam: I would actually be very explicit here as it's frustrating to have to go dig when there is code mis-behaving.

```
raise Validation Error(
    "Expected header bytes to be of length: {0}.  Got length {1} instead.\n- {2}".format(
        obj_size * len(cls.fields), len(header_bytes), encode_hex(header_bytes),
    )
)
```

Good error messages are soooo valuable when things start going [pear shaped](https://en.wikipedia.org/wiki/Pear-shaped).

evm/utils/numeric.py
```diff
@@ -21,6 +29,17 @@ def int_to_byte(value):
...
 
 
+def int_to_bytes32(value):
+    value_bytes = pipe(
+        value,
+        int_to_big_endian,
+        pad32,
+    )
+    if len(value_bytes) > 32:
+        raise ValueError("value is too large")
```
pipermerriam: Instead of doing *post* validation, how about just adding a pre-condition check that it's a valid `uint256` and raising a `ValueError` if it isn't.  (hint, I don't think we should use `evm.validation.validate_uint256` here since it's outside of `evm.utils`.  Instead, just inline the same validation code.

Btw, I suggest `ValueError` for the same reason, not importing from outside of `evm.utils` from within `evm.utils`.

evm/rlp/headers.py
mhchia: Thank you sooo much! I didn't think of this.
I will keep reminding myself with this suggestion.

evm/utils/numeric.py
mhchia: Thank you for suggesting this.
Actually I'm curious how to know when to import one function from outside directory or just inline it?
I can guess in this case the validation is simple so the cost(DRY) is not expensive. But if it is in other case, is it not good to import function in another layer?

Issue #332, opened 2018-01-31
JSON-fixture tests builder tool


hwwhww:
### What is wrong?
Since we made (and may make more) protocol changes for Ethereum 2.0, it's better to be well-tested with more various JSON-fixture tests.

### How can it be fixed
As @pipermerriam suggested, we can write a tool to record EVM execution and export it in the JSON-fixture tests format.

jannikluhn:
What do you think of the following scheme:

```
test1 = StateTest(
    name="CallIncrementsNonce"

    env={  # should have a sensible default because most tests won't touch it
        # ...
    },
    pre_state={
        ACCOUNT_ADDRESS: ACCOUNT_CODE,
    },

    transaction=UnsignedUserAccountTransaction(
        **DEFAULT_TX_PARAMS
    ).as_signed_transaction(PRIVATE_KEY),

    output=None,
    state_updates=[
        (ACCOUNT_ADDRESS, "storage", 0, 1),
    ],
    logs=[],
    gas_payment=0,
)

test2 = StateTest(...)

test_group = TestGroup([test1, test2])
print(test_group.to_json())
```

Initially I thought we'd want a more complex solution, but now I don't think this would make test creation more clear or concise. So basically, most fields from the JSON file are specified directly. Only the transaction is given by a transaction object, and instead of the complete final state only state changes are given. Changes can be either given as a dictionary (similar to `pre_state`) or as a flat list to avoid unnecessary nesting especially for storage. Now that I think about it, `pre_state` could be such a list as well.

`TestGroup` groups the tests (obviously) and creates the JSON representation. For this, the `post_state` is constructed and the state root is calculated (by setting the state directly, not by executing the transaction). If multiple tests share the same name they are required to have the same pre_state as well. In this case a single "parametrized" test is created (see "indexes" in the state test files).

If many tests share the same `pre_state` (one of the cases I thought a more complex design might be useful), `functools.partial` could be used to avoid repetition. To test a sequence of transactions, multiple tests have to be generated (because state tests allow only for a single transaction), but for that a helper can be used (saves setting env and pre_state):

```
test3 = test2.extend(
    # same parameters as StateTest, excluding env and pre_state
)
```

VMTests have mostly the same fields, but "transaction" is replaced with "exec":

```
test1 = VMTest(
    name="PAYGAS",
    env={
        # ...
    },
    pre_state={
        # ...
    },
    exec={
        "code": [PUSH1, b"\x00", PAYGAS],
    },
[...]

PR #350, opened 2018-02-04
JSON test builder

jannikluhn:
Referencing #332 

This is an unfinished first iteration of a JSON state test builder with the goal to explore if this approach makes sense. For sharding some changes in the test runner and the JSON format are required, so to keep things simple I started with replicating main chain tests.

I posted a usage example the resulting JSON file [here](https://gist.github.com/jannikluhn/208707c73e96bf9f7ad8bdbff6154e88).

For comparison, from what I understood the original JSON tests are generated from a hand written JSON file (or automatically generated using some pattern) which is subsequently "filled" (i.e. expanded and the resulting state is calculated). Compared to my approach this is safer as it doesn't use the tested code to generate the tests (the example above for instance uses the `BaseTransaction` class). But it makes writing tests more cumbersome and less clear (especially when encoding transaction data for, e.g., user account transactions).

Anyways, I'm very interested in your opinions and other ideas on this, @pipermerriam @hwwhww @NIC619.

#### Cute Animal Picture


evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+
+
+def normalize_state(state):
+    if isinstance(state, Iterable) and not isinstance(state, Mapping):
+        state = state_list_to_dict(state)
+    for address, account_state_dict in state.items():
+        account_state_dict.setdefault("balance", 0)
+        account_state_dict.setdefault("nonce", 0)
+        account_state_dict.setdefault("code", b"")
+        account_state_dict.setdefault("storage", OrderedDict())
```
pipermerriam: This will mutate the original state dictionary which is very likely to cause un-expected errors.  Can you change this to a non-mutating pattern?

evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+    formatted_items = apply_formatter_to_array(state_item_formatter, items)
+    return dict(formatted_items)
+
+
+class StateTest:
+
+    def __init__(
+        self,
+        name,
+        *,
```
pipermerriam: I'm not familiar with this pattern.  What's it do?

evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+        else:
+            self.filled_with = FILLED_WITH_TEMPLATE.format(version=get_version_from_git())
+        self.info = OrderedDict([
+            ("comment", self.comment),
+            ("filledwith", self.filled_with),
+        ])
+        self.forks = forks
+
+        # pre state
+        self.env = merge(DEFAULT_ENV, env or OrderedDict())
```
pipermerriam: You could just change `OrderedDict()` to `{}` here for slightly improved readability.  Same with other places that you use an empty ordered dict (unless your relying on the underlying `merge` to return an `OrderedDict` at which point please ignore (does it do that?).

evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+        return env
+
+    @property
+    def pre_state(self):
+        pre_state = self.sub_tests[0].pre_state
+        assert all(test.pre_state == pre_state for test in self.sub_tests)
+        return pre_state
+
+    def to_dict(self):
+        d = OrderedDict()
```
pipermerriam: `eth-utils` has an `eth_utils.to_ordered_dict` which will let you write this using the `yield` pattern (if you're inclined to).

evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+    new_state = copy.deepcopy(state2)
+    for key, value in state1.items():
+        if key not in new_state:
+            new_state[key] = copy.deepcopy(value)
+        elif isinstance(value, Mapping) and isinstance(new_state[key], Mapping):
+            new_state[key] = merge_states(new_state[key], value)
+    return new_state
+
+
+def without_duplicates(l):
```
pipermerriam: Available from [`toolz.unique`](http://toolz.readthedocs.io/en/latest/api.html)

evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+    return list(OrderedDict.fromkeys(l))
+
+
+def flatten(l):
```
pipermerriam: Available from [`toolz.concat`](http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.concat) though it doesn't return a `list` type but rather the return type from `itertools.chain.from_iterable`.

evm/utils/test_builder.py
```diff
@@ -0,0 +1,309 @@
...
+def int_to_hex(i):
+    return encode_hex(int_to_big_endian(i))
+
+
+def get_version_from_git():
+    version = subprocess.check_output(["git", "describe"]).strip()
+    return force_text(version)
+
+
+env_formatter = curried.apply_formatters_to_dict({
```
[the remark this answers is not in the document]
jannikluhn: I love the formatters, they made the code much cleaner! The only one I missed was a dict key formatter (basically what `storage_formatter` and `state_formatter` do manually now), but I'm not sure if it's a good idea to add them to the library as it could destroy uniqueness of keys if one's not careful.

evm/utils/test_builder.py
jannikluhn: It requires the following arguments to be given by keyword (https://www.python.org/dev/peps/pep-3102/). This has the side effect that you can add parameters with no default after other keyword arguments (`transaction` and `secret_key`).

evm/utils/test_builder.py
jannikluhn: Hmm, I assumed it does, but now I checked and I'm wrong... I tried to use `OrderedDict`s to keep a sensible order in the final JSON file (pre before transaction before post). But apparently the `ethereum/tests` standard is ordering alphabetically, so I don't think it's worth the effort. I'm gonna switch to plain dictionaries.

evm/utils/test_builder.py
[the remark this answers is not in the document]
pipermerriam: You're looking for `eth_utils.apply_key_map`

evm/utils/test_builder.py
pipermerriam: I love this so much.  I'd like to backfill this pattern onto a bunch of places in py-evm and also the p2p code (cc @gsalgado).

evm/utils/test_builder.py
pipermerriam: In this same vein, we *should* use `sort_keys=True` if this actually ends up dumping JSON strings at any point.

jannikluhn:
I started over completely:

1) Test generation is now split into two steps:
  - test specification in high level "filler" files
  - "compilation" of the filler file to the actual test file
2) removed classes, the JSON file itself (or rather, the corresponding dict) is generated step by step in a functional style with some helper functions

The motivation for 1) was to resemble the existing process of JSON test generation over at ethereum/tests as closely as possible. This should make integration with their and other tooling easier (e.g. filler files could be generated with py-EVM and then filled with different implementations to check that the state roots match).

I may have gone overboard with 2 a bit (it's just so much fun!), but in general it makes sense I think. The dict stores exactly the same data as the class, but with less clutter (got rid of these huge argument lists as well). Also, this should be more flexible and allow for manual modification if necessary.

As always, here's a usage example, with resulting filler and test file: https://gist.github.com/jannikluhn/5a8e6cd0e85e9f27c9679d2458fe8fe0

evm/utils/test_builder/normalization.py
```diff
@@ -0,0 +1,103 @@
...
+        - `(address, "balance", 1)`
+        - `(address, "nonce", 2)`
+        - `(address, "code", b"3")`
+        - `(address, "storage", 4, 5)`
+
+    For storage, the second to last entry specifies the slot and the last one the value.
+    """
+    d = {}
+    for state_item in state_list:
+        d = assoc_in(d, state_item[:-1], state_item[-1])
```
pipermerriam: *functional nonsense*

```python
return cytoolz.merge(*(
    assoc_in({}, state_item[:-1], state_item[-1])
    for state_item
    in state_list
))
```

pipermerriam: It *might* be good to pre-validate that `state_list` is well formed.  Otherwise this is going to produce some very confusing results that might be hard to track down.  Similarly, enforcing uniqueness of all `state_item[:-1]` would fit into that validation to ensure that no duplicates are present.

evm/utils/test_builder/test_builder.py
```diff
@@ -0,0 +1,218 @@
...
+    environment = merge(DEFAULT_ENVIRONMENT, normalize_environment(environment or {}))
+    return {name: {
+        "env": environment,
+    }}
+
+
+@curry
+def pre_state(pre_state, filler):
+    assert len(filler) == 1
+    test_name = next(iter(filler.keys()))
```
pipermerriam: These two lines are duplicated a few times.  Maybe write a small utility function  **or** change the signature to expect two arguments `pre_state(pre_state, test_name, state)`

evm/utils/test_builder/builder_utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+from eth_utils import (
+    force_text,
+    int_to_big_endian,
+)
+
+from eth_keys import keys
+
+
+def merge_nested(*dicts):
```
pipermerriam: *functional nonsense*

Slightly more concise and uses fully non-mutation based approach

```python
from cytoolz import merge_with

def merge_if_dicts(values):
    if all(isinstance(item, Mapping) for item in values):
        return merge_with(merge_if_dicts, *values)
    else:
        return values[-1]

def deep_merge(*dicts):
    return merge_with(merge_if_dicts, *dicts)
```

jannikluhn: 👍 for `deep_merge` over `merge_nested`, as its analogous to `copy.deepcopy`

evm/utils/test_builder/normalization.py
jannikluhn: I like it, that's nicer (but needs `deep_merge` instead of `merge`). With 
```
deep_merge(*(
    assoc_in({}, state_item[:-1], state_item[-1]) if isinstance(state_item, Mapping) else state_item
    for state_item
    in state_list
))
```
this would be even more flexible (then one can switch from the list to dict definition at any level:
```
[
    ("0xaabb", "balance", 3),
    ("0xaabb", "storage", {
        4: 5,
    }),
    {"0xbbcc", {
        "balance": 6,
        "nonce": 7
    }
]
```

pipermerriam:
Could we get a small `README.md` document somewhere in the fillers that documents how this stuff works at some basic level of granularity.  Doesn't have to be utterly thorough documentation but something to get others started building on this API.

evm/utils/fixture_tests.py
```diff
@@ -287,37 +290,34 @@ def generate_fixture_tests(metafunc,
...
     elif 'v' in transaction and 'r' in transaction and 's' in transaction:
         yield 'vrs', (
             to_int(transaction['v']),
             to_int(transaction['r']),
             to_int(transaction['s']),
         )
+        is_main_transaction = True
     else:
-        raise KeyError("transaction missing secret key or vrs values")
+        is_main_transaction = False
```
carver: I think it will help prevent future errors if we validate the shard transaction fields. Something like:

```py

def is_valid_shard_transaction(transaction):
    return all(key in transaction for key in {'shardID', 'accessList'})

if 'secretKey' in transaction:
...
elif is_valid_shard_transaction(transaction):
     is_main_transaction = False
else:
    raise ValueError("transaction must be a signed main chain transaction, or a shard transaction.")

**Comment by jannikluhn** [comment]
Updates:

- rebased
- paygas tests fixed (they pass now when copied to the fixture directory)
- more strict normalization (missing and additional keys in dictionary should be catched now)
- https://github.com/ethereum/py-evm/pull/350/commits/65d7b24d718b1aa1238ba72a385d16baa93498e7: fixed shard VM instantiation which failed due to the differences between blocks and collations. I don't particularly like my change of `persist_block_to_db`, but it's the least invasive way to make it work (and I assume that the chain class has to be revised anyway to support the different types of clients).

Please review one (hopefully) last time.

Also, @NIC619 could you have a look at https://github.com/jannikluhn/py-evm/blob/aa7343e5d16ddf0292f0fd06d878f70651228133/tests/fillers/vm_fillers/paygas_fillers.py and check if we need additional test cases for PAYGAS?

File: evm/db/chain.py [hunk 12]
```diff
@@ -403,10 +403,11 @@ def persist_block_to_db(self, block):
                 self._add_transaction_to_canonical_chain(transaction_hash, header, index)
 
         # Persist the uncles list
-        self.db.set(
```
pipermerriam: Opened https://github.com/ethereum/py-evm/issues/410 to track this issue.

evm/utils/test_builder/README.md
```diff
@@ -0,0 +1,183 @@
+# JSON test builder
+
+
+The JSON test builder found in `evm.utils.test_builder.test_builder` is a library that
```
pipermerriam: I recently introduced `evm.tools` into the module.  I think that `test_builder` should probably be moved there.

evm/utils/test_builder/README.md
```diff
@@ -0,0 +1,183 @@
...
+  key                      | description
+  -------------------------|--------------------------------------------------------------------
+  `"currentCoinbase"`      | the coinbase address
+  `"currentNumber"`        | the collation number
+  `"previousHash"`         | the hash of the parent collation
+  `"shardID"`              | the ID number of the shard
+  `"expectedPeriodNumber"` | the period number in which the collation is supposed to be included
+  `"periodStartHash"`      | the hash of the block starting the current period.
+
+Use `setup_main_filler` or `setup_sharding_filler` with the same signature, but correct
```
pipermerriam: This sentence appears to be truncated.

evm/utils/test_builder/builder_utils.py
```diff
@@ -0,0 +1,126 @@
...
+)
+
+from evm.db.backends.memory import MemoryDB
+
+from cytoolz import (
+    first,
+    merge_with,
+)
+from eth_utils import (
+    force_text,
```
pipermerriam: `force_text` shouldn't be used.  It's been/being deprecated and removed.  If you need to coerce things from `bytes -> text` you should use `eth_utils.to_text` in combination with `eth_utils.text_if_str`

```python
from eth_utils.curried import text_if_str
from eth_utils import to_text

force_text = text_if_str(to_text)
```

The behavior isn't quite identical, but I think it's identical in all the ways that matter.  Note, that it **will not** be able to handle any arbitrary bytestring since not all 0-255 byte values are `utf8` encodable.

evm/utils/test_builder/builder_utils.py
```diff
@@ -0,0 +1,126 @@
...
+    else:
+        return values[-1]
+
+
+def deep_merge(*dicts):
+    return merge_with(merge_if_dicts, *dicts)
+
+
+def is_cleanly_mergable(*dicts):
+    if len(dicts) <= 1:
```
pipermerriam: This function, while not a public API might benefit from a docstring since it's not entirely clear what *cleanly mergable* means without reading the implementation.  A human readable description of this function would go a long way to helping future readers understand it's intended functionality.

evm/utils/test_builder/formatters.py
```diff
@@ -0,0 +1,133 @@
...
+    combine_argument_formatters,
+    encode_hex,
+    int_to_big_endian,
+    to_checksum_address,
+)
+import cytoolz.curried
+import eth_utils.curried
+
+
+int_to_hex = compose(encode_hex, int_to_big_endian)
```
pipermerriam: this can be replaced with `eth_utils.to_hex`

evm/utils/test_builder/normalization.py
```diff
@@ -0,0 +1,278 @@
...
+                state_item[:-1],
+                state_item[-1]
+            ) if not isinstance(state_item, Mapping) else state_item
+            for state_item
+            in state_definition
+        ]
+        assert is_cleanly_mergable(*state_dicts)
+        state_dict = deep_merge(*state_dicts)
+    else:
+        assert False
```
pipermerriam: It'd be nice to get a human readable error here that exposed the allowed types and the type that was observed.

evm/utils/test_builder/normalization.py
```diff
@@ -0,0 +1,278 @@
...
+        state_dicts = [
+            assoc_in(
+                {},
+                state_item[:-1],
+                state_item[-1]
+            ) if not isinstance(state_item, Mapping) else state_item
+            for state_item
+            in state_definition
+        ]
+        assert is_cleanly_mergable(*state_dicts)
```
pipermerriam: It'd be nice to get an exception here that showed what key or maybe keys where found to not be cleanly mergable.

evm/utils/test_builder/normalization.py
```diff
@@ -0,0 +1,278 @@
...
+        state_dict = deep_merge(*state_dicts)
+    else:
+        assert False
+
+    second_layer_keys = concat(d.keys() for d in state_dict.values())
+    assert all(key in ["balance", "nonce", "storage", "code"] for key in second_layer_keys)
```
pipermerriam: *minor performance thing*

- casting `second_layer_keys` to a `set` will remove duplicates so you don't check the same key multiple times.
- the set of allowed keys *should* be a `set` since we're using it purely to check for presence within the set.
- this could be done in one go with `bad_keys = set(seen_keys).difference({"balance", "nonce", ...})`

I'd recommend changing to use the `bad_keys = ` pattern because then you can raise an explicit error showing what *extra* keys were found.

evm/utils/test_builder/test_builder.py
```diff
@@ -0,0 +1,371 @@
...
+}
+
+
+DEFAULT_MAIN_TRANSACTION = {
+    "data": b"",
+    "gasLimit": 100000,
+    "gasPrice": 0,
+    "nonce": 0,
+    "secretKey": decode_hex("0x45a915e4d060149eb4365960e6a7a45f334393093061116b197e3240065ff2d8"),
+    "to": to_canonical_address("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6"),
```
pipermerriam: Did this address come from anywhere in particular?

evm/vm/forks/sharding/collations.py
```diff
@@ -26,6 +26,8 @@ def __init__(self, header, transactions=None, witness_nodes=None):
         if witness_nodes is None:
             witness_nodes = []
 
+        self.transaction_fee_sum = 0
```
pipermerriam: This doesn't appear to be used (as researched by a quick ctrl+f)

evm/utils/test_builder/builder_utils.py
carver: Or, if you know the input will always be a `bytes` value, then `to_text(bytes_val)` is sufficient.

evm/vm/forks/sharding/collations.py
NIC619: `block.transaction_fee_sum`(or `collation.transaction_fee_sum`) is used to record the accumulated transaction fee paid by each transaction that's included in the block(collation).

```python
def add_transaction(self, transaction, computation, block):
    …
    # Calculate transaction fee
    transaction_fee, _ = computation.compute_transaction_fee_and_refund()
    # Bookkeep this transaction fee
    block.transaction_fee_sum += transaction_fee
    ...
```

NIC619:
@jannikluhn What do you think about adding a negative gas price test case?

evm/vm/forks/sharding/collations.py
pipermerriam: Ah, I see.

I've opened #412 to track the task of trying to find a new way to do this.

I need to dig in and understand a bit better why this is there in the first place, but I suspect we can find a different mechanism.

evm/utils/test_builder/test_builder.py
jannikluhn: It's used in a lot of existing JSON tests (https://github.com/ethereum/tests/search?utf8=%E2%9C%93&q=0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6&type=), but I guess they just generated it randomly.

jannikluhn:
> @jannikluhn What do you think about adding a negative gas price test case?

The EVM doesn't know negative numbers, so we don't have to (and can't, even) write such a test here. But it's something the Solidity/Vyper guys should do in their tests.

evm/rlp/headers.py
```diff
@@ -273,3 +288,21 @@ def from_parent(cls,
         }
         header = cls(**header_kwargs)
         return header
+
+    def clone(self):
+        # Create a new CollationHeader object with the same fields.
+        return self.__class__(**{
+            field_name: getattr(self, field_name)
+            for field_name
+            in tuple(zip(*self.fields))[0]
```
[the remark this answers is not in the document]
jannikluhn: Looks like `unzip` is only in `toolz`, not `cytoolz`, unfortunately. But I'll change the `tuple(..)[0]` to `cytoolz.first(..)` which is arguably a bit nicer.

jannikluhn:
Addressed the feedback and made some minor changes to make the tests pass. Will merge when tests are green (hope I caught everything this time).

evm/rlp/headers.py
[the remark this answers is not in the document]
pipermerriam: Nice to know that `first` works on generator/iterator types.

Issue #331, opened 2018-01-31
Switch to `BinaryTrie` for transactions and receipts


hwwhww:
### What is wrong?
For sharding, we switch all `HexaryTrie` to `BinaryTrie`.

### How can it be fixed
Most changes are in `evm.db.chain.BaseChainDB`. I suggest we divide it into two parts:
1. `master` branch:
    * Make `trie_class` to be configurable in `BaseChainDB.__init__`.
    * Default: `HexaryTrie`

2. Rebase `sharding` branch and switch sharding-related tries to `BinaryTrie`.

hwwhww:
close via #355

PR #351, opened 2018-02-05
Configurable trie_class in master branch - part2

hwwhww:
### What was wrong?
Another patch for switching transaction and receipt tries (#331) , and also prepare for switching State trie (#289).

### How was it fixed?
The empty `HexaryTrie` root is `BLANK_ROOT_HASH` and empty `BinaryTrie` root is `EMPTY_SHA3`, so it requires setting the default root by `chaindb.trie_class`.

#### Cute Animal Picture

dog

### DIFF
--- a/evm/rlp/headers.py
+++ b/evm/rlp/headers.py
@@ -114,8 +114,7 @@ class BlockHeader(rlp.Serializable):
                     nonce=None,
                     extra_data=None,
                     transaction_root=None,
-                    receipt_root=None,
-                    state_root=None):
+                    receipt_root=None):
         """
         Initialize a new block header with the `parent` header as the block's
         parent hash.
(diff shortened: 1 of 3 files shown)


evm/rlp/headers.py
```diff
@@ -112,7 +112,10 @@ def from_parent(cls,
                     timestamp,
                     coinbase=ZERO_ADDRESS,
                     nonce=None,
-                    extra_data=None):
+                    extra_data=None,
+                    transaction_root=None,
+                    receipt_root=None,
+                    state_root=None):
```
pipermerriam: I'm curious about these additions.  They don't  appear to be used anywhere in this PR.

hwwhww: @pipermerriam Since the default empty root is `BLANK_ROOT_HASH`, these additions will be used in `ShardingVM.create_header_from_parent` to use `EMPTY_SHA3` as the empty root:

```python
def create_sharding_header_from_parent(parent_header, **header_params):
    if 'transaction_root' not in header_params:
        header_params['transaction_root'] = EMPTY_SHA3
    if 'receipt_root' not in header_params:
        header_params['receipt_root'] = EMPTY_SHA3
    return create_byzantium_header_from_parent(parent_header, **header_params)
```

And after you pointed out, I found that I shouldn't add `state_root=None`, it shouldn't be overwritten here. Will remove `state_root` parameter, thank you!

PR #352, opened 2018-02-05
memoize multiple utility functions

pipermerriam:
STUB: been meaning to test whether this has a noticeable speed improvement.

#### Cute Animal Picture

### DIFF
--- a/evm/utils/address.py
+++ b/evm/utils/address.py
@@ -1,5 +1,8 @@
 import rlp
 
+from cytoolz import memoize
+import lru
+
 from .keccak import (
     keccak,
 )
@@ -8,6 +11,7 @@ from .padding import (
 )
 
 
+@memoize(cache=lru.LRU(64))
 def force_bytes_to_address(value):
     trimmed_value = value[-20:]
     padded_value = pad_left(trimmed_value, 20, b'\x00')
--- a/evm/utils/keccak.py
+++ b/evm/utils/keccak.py
@@ -9,7 +9,7 @@ except ImportError:
     from sha3 import sha3_256 as keccak_256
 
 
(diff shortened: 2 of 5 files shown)


pipermerriam:
@carver @dylanjw thoughts on this?

Looking at the test run times I don't think I can objectively say this produced a significant speed improvement, but it *shouldn't* have slowed anything down.  Memory footprint of memoizing these function calls should be reasonably low.

My *gut* says to hold off until we can implement some sort of global caching framework so that it's up to users what sort of memory footprint they want to allow for this sort of thing.

carver:
Hm, without a clear speed improvement, I'm a very slight 👎 . Out of curiosity, can we get a measurable speedup if you massively increase cache size?

pipermerriam:
> Out of curiosity, can we get a measurable speedup if you massively increase cache size?

I doubt it.  There are roughly two or three optimizations at play here depending on how you look at it.

1. JSON-fixture tests use many of the same addresses for contracts, transactions, etc.
2. There are a few values that just natively get used a lot like 32 null bytes or empty byte string.
3. There are a few storage values that are going to be common, like lower integers.

The optimizations happen wtih

- `evm.db.hash_trie.HashTrie` which uses `keccak(key)` as the key when storing in the trie.
- `evm.vm.stack` which moves values back and forth between integer and bytes representations (`int_to_big_endian/big_endian_to_int`).
- `evm.db.state` which uses `pad32` for it's storage keys.

I'm going to try bumping the cache sizes.  Maybe `64` was a bit too small and these common values were constantly being evicted.  Will check run-times in the morning and see what things look like.

pipermerriam:
Not seeing any noticeable speed increases. 👎

PR #354, opened 2018-02-06
relegate low level DB to it's own process

pipermerriam:
### What was wrong?

We needed a way to share database access across the processes.

### How was it fixed?

A **REALLY** simple process which handles the low level database connection, running in a loop.

TODO: this needs to be written using the async primatives as well as accepting *many* pipes, so that we can give the database one pipe per *other* process which needs to read from the database.  Otherwise I suspect we'll run into some weird problems (but we should be sure before making it more complicated than necessary).

#### Cute Animal Picture

tiny-baby-animals-4jpg


trinity/main.py
```diff
@@ -65,27 +73,85 @@ def main():
...
+        chain_process.join()
     except KeyboardInterrupt:
         logger.info('Keyboard Interrupt: Stopping')
-        process.terminate()
+        chain_process.terminate()
+        db_process.terminate()
 
 
-def ropsten_light_node_sync(log_queue, db_path):
+def base_db_process(db_class, init_args, mp_pipe, log_queue):
```
carver: I expect it to be a pretty large refactor to handle multiple pipes and async, so this is a note to self that I'd be curious to see how that plays out. When I did the IPC server, I "cheated" and used a built-in server, but I don't think that has any explicit multiprocess support. For reference: https://github.com/ethereum/py-evm/blob/master/evm/rpc/ipc.py#L99-L104

trinity/db/mp.py
```diff
@@ -0,0 +1,66 @@
...
+        req_id = uuid.uuid4()
+        self.mp_connection.send([req_id, DELETE, key])
+
+        resp_id, response = self.mp_pipe.recv()
+        if resp_id != req_id:
+            raise ValueError('Request/Response id mismatch')
+
+        if isinstance(response, Exception):
+            raise response
+        return response
```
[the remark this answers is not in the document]
pipermerriam: Got any bright ideas?  They all look the same but nothing jumped out at me as a way to de-dup.

trinity/main.py
pipermerriam: I *think* that wiring this in will be as simple as modifying the RPCServer to use the `PipeDB` (previously `MPDB`) class for it's backend db connection.

trinity/db/mp.py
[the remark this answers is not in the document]
carver: I'm imagining something like:
```py
def run_command(mp_pipe, command, *params):
    req_id = uuid.uuid4()
    mp_pipe.send([req_id, command] + params)

    resp_id, response = mp_pipe.recv()
    if resp_id != req_id:
        raise ValueError('Request/Response id mismatch')

    if isinstance(response, Exception):
        raise response
    return response

class MPDB(BaseDB):
    def set(self, key, value):
        return run_command(self.mp_pipe, SET, key, value)

    def delete(self, key):
        return run_command(self.mp_pipe, DELETE, key)
    ...
```

trinity/db/mp.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def delete(self, key):
+        req_id = uuid.uuid4()
+        self.mp_connection.send([req_id, DELETE, key])
```
carver: `mp_connection` => `mp_pipe`, presumably

tests/trinity/chain-management/test_is_data_dir_initialized.py
```diff
@@ -46,41 +39,37 @@ def nodekey(chain_config, data_dir):
...
-    # Database Initialization
-    chaindb = get_chain_db(chain_config.database_dir)
-    chain = RopstenLightChain.from_genesis_header(chaindb, ROPSTEN_GENESIS_HEADER)
 
-    # we need to delete the chaindb in order to release the leveldb LOCK.
-    del chain
-    del chaindb.db
-    del chaindb
+def test_full_initialized_data_dir_with_custom_nodekey():
+    return ChainConfig('test_chain', nodekey=NODEKEY)
```
carver: ~~return~~ `chain_config = ` ?

pipermerriam: ?!

pipermerriam: also fixed.

trinity/main.py
```diff
@@ -134,14 +119,19 @@ def main():
...
 
     # if console command, run the trinity CLI
     if args.subcommand == 'console':
         use_ipython = not args.vanilla_shell
         debug = args.log_level.upper() == 'DEBUG'
 
-        chain = chain_obj(chain_config, sync_mode)
+        chain_class = get_chain_protocol_class(chain_config, sync_mode)
+        chaindb = BaseChainDB(LevelDB(chain_config.database_dir))
+        chain = chain_class(chaindb)
```
carver: Looks like a reasonable solution to the earlier `sync_mode` question 👍

tests/trinity/chain-management/test_is_data_dir_initialized.py
carver: haha, yeah, how did that even run?

tests/trinity/chain-management/test_is_database_initialized.py
```diff
@@ -0,0 +1,39 @@
...
+    return BaseChainDB(LevelDB(db_path=chain_config.database_dir))
+
+
+def test_database_dir_not_initialized_without_canonical_head_block(chaindb):
+    assert not is_database_initialized(chaindb)
+
+
+def test_fully_initialized_database_dir(chaindb):
+    assert not is_database_initialized(chaindb)
+    RopstenLightChain.from_genesis_header(chaindb, ROPSTEN_GENESIS_HEADER)
```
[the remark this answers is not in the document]
pipermerriam: Realized I can just swap this for `chaindb.persist_header_to_db(...)`

trinity/main.py
pipermerriam: Yeah, it cleaned itself up on it's own once I separated initialization of the data-dir from initialization of the database.

tests/trinity/chain-management/test_is_data_dir_initialized.py
pipermerriam: Tests are just functions.  It just returned from the test function without hitting any of the assertion statements.

Issue #356, opened 2018-02-06
Add tests for `PipeDB` and corresponding process


pipermerriam:
### What is wrong?

#354 introduces a new database wrapper `trinity.db.pipe.PipeDB` which accesses a database behind a `multiprocessing.Pipe` connection.  There are no tests.

### How can it be fixed

Fill in a simple test suite against an in memory database.  We should spin up a new process in the tests running the `db_over_pipe` process and then test that we can interact with the database as expected using a `PipeDB` instance.

PR #358, opened 2018-02-07
Change trinity db connection to listener model

pipermerriam:
### What was wrong?

The previous model for database connections required setting up `multiprocessing.Pipe` connections in advance.  This is cumbersome as well as restrictive.

### How was it fixed?

- Modified the database process to use a `multiprocessing.connection.Listener` which will accept socket connections from other processes.
- Added a new socket connection `data-dir/db.ipc` for connecting to the db server.
- Changed `PipeDB` to use `multiprocessing.connection.Client` to connect to the database server.

#### Cute Animal Picture

cat-monkey-shutterstock-76988095-webonly


pipermerriam:
closes #356

trinity/db/pipe.py
```diff
@@ -35,69 +43,136 @@ def run_command(pipe_db, *params, command):
...
+            args=(listener, connections),
+            daemon=True,
+        )
+        thread.start()
+
+        while True:
+            # If we have no connections, sleep for a moment, otherwise `wait`
+            # blocks indefinitely on an empty list.
+            if not connections:
+                time.sleep(0.1)
```
carver: Hm, this sleep and check stuff makes me think we might be better off with an `asyncio.start_unix_server(...)` style server.

Like https://github.com/ethereum/py-evm/blob/master/evm/rpc/ipc.py#L99

Doesn't seem urgent if this works, though. Just an idea.

trinity/main.py
```diff
@@ -103,6 +111,12 @@
 )
 console_parser.set_defaults(func=console)
 
+"""
+chaindb = ChainDB(LevelDB(args.db))
+peer_pool = PeerPool(LESPeer, chaindb, ROPSTEN_NETWORK_ID, ecies.generate_privkey())
+chain = DemoLightChain(chaindb, peer_pool)
+"""
```
carver: cruft?

trinity/utils/mp.py
```diff
@@ -1,9 +1,51 @@
...
+
+
+def wait_for_ipc(ipc_path, timeout=1):
+    start_at = time.time()
+    while time.time() - start_at < timeout:
+        if os.path.exists(ipc_path):
+            break
+
+
+def kill_processes_gracefully(*processes, logger=None, SIGINT_timeout=5, SIGTERM_timeout=3):
```
carver: nice

pipermerriam:
> The sleep after the wait_for_ipc is the only thing

Looks like they weren't needed.  Removed in #359

tests/trinity/database/test_db_over_mp_connection.py
```diff
@@ -0,0 +1,50 @@
...
+        ipc_path = os.path.join(temp_dir, 'db.ipc')
+
+        db_server_process = multiprocessing.Process(
+            target=db_server,
+            args=(base_db, ipc_path),
+        )
+        db_server_process.start()
+
+        wait_for_ipc(ipc_path)
+        time.sleep(0.1)
```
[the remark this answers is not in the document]
pipermerriam: removed in #359

trinity/db/pipe.py
pipermerriam: I looked into this to some degree.  What I couldn't figure out was how to get the friendly API that `multiprocessing.Connection` has of automatic pickling and unpickling of objects which is key to making this work without a lot more overhead.

pipermerriam: Also found this: https://github.com/dano/aioprocessing

Doesn't look super promising, but if nothing else we could lift limited implementations of the necessary components from there.

trinity/chains/__init__.py
```diff
@@ -97,5 +97,5 @@ def get_chain_protocol_class(chain_config, sync_mode):
         raise NotImplementedError("Ropsten is the only chain currently supported.")
 
     return RopstenLightChain.configure(
-        privkey=chain_config.nodekey,
+        #privkey=chain_config.nodekey,
```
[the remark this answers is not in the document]
pipermerriam: Yeah should be removed entirely.  `privkey` is no longer a chain attribute with the latest p2p code(lives in `PeerPool` now)

trinity/db/pipe.py
carver: > What I couldn't figure out was how to get the friendly API that multiprocessing.Connection has of automatic pickling and unpickling of objects 

Ah, yes. `multiprocessing` is better for now, then.

> https://github.com/dano/aioprocessing
>
> Doesn't look super promising, but if nothing else we could lift limited implementations of the necessary components from there.

Yeah, a failing build and no activity for a year is a no-go for direct usage. Since they call out `multiprocessing.Connection` specifically, it seems worth a look at some point. Still doesn't seem urgent, though.

pipermerriam:
I'm going to clean these up in #359

PR #359, opened 2018-02-07
Relegate chaindb access to independent process

pipermerriam:
### What was wrong?

In order to prevent EVM execution from blocking execution and causing the networking threads to time out, we need to segregate EVM execution from the networking execution.

### How was it fixed?

Created an abstraction around the `ChainDB` class which delegates requests over a `multiprocessing.Connection` to a separate `ChainDB` instance in it's own `multiprocessing.Process`.

#### Cute Animal Picture

cropped-happy-smug-sloth


trinity/utils/ipc.py
```diff
@@ -0,0 +1,188 @@
...
+import threading
+import time
+import uuid
+
+
+def wait_for_ipc(ipc_path, timeout=1):
+    start_at = time.time()
+    while time.time() - start_at < timeout:
+        if os.path.exists(ipc_path):
+            break
```
carver: Wow that's a tight loop. Maybe a tiny `else: sleep(0.05)` is reasonable to avoid slamming disk/cpu.

tests/trinity/database/test_core_db_over_ipc.py
```diff
@@ -0,0 +1,56 @@
...
+    assert db[b'key-b'] == b'value-b'
+
+    assert b'key-c' not in db
+    db[b'key-c'] = b'value-c'
+    assert b'key-c' in db
+    assert db[b'key-c'] == b'value-c'
+
+    assert b'key-b' in db
+    del db[b'key-b']
+    assert b'key-b' not in db
```
carver: Might be nice to show exceptions in action, like:

```py
with pytest.raises(KeyError):
    db[b'spoon']
```

trinity/db/chaindb.py
```diff
@@ -0,0 +1,62 @@
...
+    BaseChainDB,
+)
+
+from trinity.utils.ipc import (
+    ObjectOverIPC,
+    IPCMethod,
+)
+
+
+class PipeChainDB(ObjectOverIPC, BaseChainDB):
```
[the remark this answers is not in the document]
pipermerriam: I almost went this far but I'm planning to iterate towards that in the next release and then probably pull it out into a stand-alone library since my quick googling didn't yield any results.

evm/db/chain.py
```diff
@@ -205,7 +183,7 @@ def clear(self):
     # State Database API
     #
     def get_state_db(self, state_root, read_only):
-        return AccountStateDB(db=self.db, root_hash=state_root, read_only=read_only)
+        raise NotImplementedError("ChainDB classes must implement this method")
```
pipermerriam: These changes will need to be backported to `master`

pipermerriam:
@carver I'm going to completely rework this to use [`multiprocessing.Manager`] and [`multiprocessing.Proxy`] which are built-in parts of the `multiprocessing` library which do exactly what my ipc utilities do.

https://docs.python.org/3.5/library/multiprocessing.html#multiprocessing-managers

carver:
> After some experimenting with converting ethereum/py-evm#359 to use the builtin multiprocessing.managers.Manager and Proxy abstractions to accomplish the same thing that the trinity.utils.ipc utilities are doing, I'm punting. About an hour of tinkeing, including digesting the full source of the Manager and underlying Server classes, I'm unable to get them to shut down cleanly.

>Also, I'm quite comfortable with my current implementation and I fully understand what it's doing. The Manager and Proxy architecture uses a lot of sub-threading and new multiprocesses that I think are not entirely necessary for our use case.

👍 Works for me. Good to go.

PR #360, opened 2018-02-07
Rename BaseChainDB to ChainDB and establish base API

pipermerriam:
### What was wrong?

When I see an object named `BaseXXX` I generally think of it as something that requires inheritance, or at least that you build on top of.  The way that our `BaseChainDB` object is used does not fit this model and the `Base` prefix adds some level of confusion (in-my-humble-ish-opinion)

### How was it fixed?

Established a `BaseChainDB` class which has stub methods for all of the *public* API.

Renamed the original `BaseChainDB -> ChainDB` and re-organized the class methods to be grouped by functionality.

> Note that this is in preparation for the work in trinity to relegate the ChainDB execution to it's own process.

#### Cute Animal Picture

cute animal friendship 007

### DIFF
--- a/tests/database/test_chaindb.py
+++ b/tests/database/test_chaindb.py
@@ -62,21 +62,21 @@ def block(request, header):
 
 def test_add_block_number_to_hash_lookup(chaindb, block):
     block_number_to_hash_key = make_block_number_to_hash_lookup_key(block.number)
-    assert not chaindb.db.exists(block_number_to_hash_key)
+    assert not chaindb.exists(block_number_to_hash_key)
     chaindb._add_block_number_to_hash_lookup(block.header)
-    assert chaindb.db.exists(block_number_to_hash_key)
+    assert chaindb.exists(block_number_to_hash_key)
 
 
 def test_persist_header_to_db(chaindb, header):
     with pytest.raises(BlockNotFound):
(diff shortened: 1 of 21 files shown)


tests/database/test_chaindb.py
```diff
@@ -62,21 +62,21 @@ def block(request, header):
 
 def test_add_block_number_to_hash_lookup(chaindb, block):
     block_number_to_hash_key = make_block_number_to_hash_lookup_key(block.number)
-    assert not chaindb.exists(block_number_to_hash_key)
-    chaindb.add_block_number_to_hash_lookup(block.header)
-    assert chaindb.exists(block_number_to_hash_key)
+    assert not chaindb.db.exists(block_number_to_hash_key)
```
carver: As far as I can tell, this change isn't necessary, because chaindb still has the `exists` proxy. Why the change?

Should we encourage callers to call `exists` one way or the other? My instinct is to suggest that they not reach inside the underlying db like this, but it's a soft opinion.

pipermerriam: Leftover from me thinking I could remove the `exists()` method.

pipermerriam: Changed back in f75c376

PR #365, opened 2018-02-09
Use chaindb over multiprocessing Manager for lightchain access to ChainDB

pipermerriam:
### What was wrong?

* The networking process was not actually using the chaindb over the IPC connection.
* When converted to use the chaindb over ipc connection all of the chaindb calls need to be made async so that they don't block the event loop.
* When the chaindb calls were wrapped in async coroutines, things started breaking.

Breakage manifested as broken client connections and was very difficult to debug in a meaningful way.  This indicated that there was a fundamental flaw in my approach in `trinity.utils.ipc` which isn't surprising.

### How was it fixed?

Converted to use the `multiprocessing.managers.Manager` model with `multiprocessing.Proxy` objects.  This has the same effect of delegating execution of the methods to the process where the manager is running.  After doing this and wrapping the networking calls to the chaindb in asyncio, the errors stopped and things began working.

> One side effect of this is that I don't currently know how to shut the process down in a clean manner.  When a `KeyboardInterrupt` is issued a big pile of stacktraces shows up which I haven't yet traced down how to make that graceful.

Bonus part is that the `Manager` pattern is really clean and requires very little boilerplate.

#### Cute Animal Picture

TODO.....


pipermerriam:
Tests have not yet been updated.  I'm planning on taking the `aioprocessing` library for a spin to see if it'll give me async access to the chaindb methods without having to wrap them in anything (which I'm pretty sure it does).

pipermerriam:
Ended up going with a simple async wrapper around the `persist_header_to_db` method.   I figure we can extend this pattern to other methods which *might* have long processing times to ensure we don't block the db *or* we could just apply this to all the chaindb functions. (experimenting with this now)

trinity/utils/ipc.py
```diff
@@ -20,13 +11,12 @@ def wait_for_ipc(ipc_path, timeout=1):
         time.sleep(0.05)
 
 
-def kill_processes_gracefully(*processes, logger=None, SIGINT_timeout=5, SIGTERM_timeout=3):
+def kill_processes_gracefully(process, logger=None, SIGINT_timeout=5, SIGTERM_timeout=3):
```
carver: Looks like the name should update to `kill_process_gracefully`?

pipermerriam:
Still something leaving open file descriptors in the RPC tests.  Haven't had time to dig for it yet.

trinity/db/chain.py
```diff
@@ -0,0 +1,40 @@
...
+        return await loop.run_in_executor(
+            None,
+            functools.partial(self._callmethod, kwds=kwargs),
+            method_name,
+            args,
+        )
+    return method
+
+
+def sync_method(method_name):
```
pipermerriam: This is deserving of a metaclass which does these two things automatically but I'm going to set that aside as an issue someone else can come through and do the implementation for since this is totally functional as is.

PR #366, opened 2018-02-10
Add transaction-context object

lrettig:
Cf. #262, and #327 where this was added to master.

Creating a PR to track this work in progress.

### DIFF
--- a/evm/logic/system.py
+++ b/evm/logic/system.py
@@ -231,7 +231,6 @@ class Create2(CreateEIP150):
             value=value,
             data=b'',
             code=call_data,
-            transaction_gas_limit=computation.msg.transaction_gas_limit,
             is_create=True,
         )
 
@@ -261,7 +260,9 @@ def paygas(computation):
             PAYGAS_gasprice = computation.get_PAYGAS_gas_price()
             if PAYGAS_gasprice is None:
                 PAYGAS_gasprice = 0
-            fee_to_be_charged = PAYGAS_gasprice * computation.msg.transaction_gas_limit
+            fee_to_be_charged = (
+                PAYGAS_gasprice * computation.transaction_context.transaction_gas_limit
+            )
 
             if tx_initiator_balance < fee_to_be_charged:
                 raise InsufficientFunds(
--- a/evm/transaction_context.py
+++ b/evm/transaction_context.py
@@ -1,0 +1,27 @@
+from evm.validation import (
+    validate_canonical_address,
+    validate_uint256,
+)
+
+
+class BaseTransactionContext(object):
+    """
+    This immutable object houses information that remains constant for the entire context of the VM
+    execution.
+    """
+    _gas_price = None
+    _origin = None
+
+    def __init__(self, gas_price, origin):
+        validate_uint256(gas_price, title="TransactionContext.gas_price")
+        self._gas_price = gas_price
+        validate_canonical_address(origin, title="TransactionContext.origin")
+        self._origin = origin
+
+    @property
+    def gas_price(self):
+        return self._gas_price
+
+    @property
+    def origin(self):
+        return self._origin
--- a/evm/transaction_context.py
+++ b/evm/transaction_context.py
@@ -4,7 +4,7 @@ from evm.validation import (
 )
 
 
-class BaseTransactionContext(object):
+class BaseTransactionContext:
     """
     This immutable object houses information that remains constant for the entire context of the VM
     execution.
--- a/tests/json-fixtures/test_state.py
+++ b/tests/json-fixtures/test_state.py
@@ -2,8 +2,6 @@ import os
 
 import pytest
 
-import traceback
-
 from eth_keys import keys
 
 from evm.db import (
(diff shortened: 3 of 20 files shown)


hwwhww:
I think it will be done when `sharding` rebases from `master`.

lrettig:
There's at least one sharding-specific tweak here, adding [sighash](https://github.com/ethereum/py-evm/pull/259/files#diff-1f5d08c6bc59d96f4db9335ec1eb94fbR152). @pipermerriam suggested we PR this into the sharding branch now. What do you think?

lrettig:
Does `transaction_gas_limit` belong in the static transaction context too?

NIC619:
@lrettig Yes, `transaction_gas_limit` belongs in the static transaction context too. Thanks for the effort! And if you have spare time to squash the commits by the way, that would really make rebasing much easier!

lrettig:
Thanks @NIC619 will do as you suggest.

evm/logic/system.py
```diff
@@ -261,7 +260,8 @@ def paygas(computation):
             PAYGAS_gasprice = computation.get_PAYGAS_gas_price()
             if PAYGAS_gasprice is None:
                 PAYGAS_gasprice = 0
-            fee_to_be_charged = PAYGAS_gasprice * computation.msg.transaction_gas_limit
+            fee_to_be_charged = PAYGAS_gasprice * \
```
pipermerriam: We've currently got zero `\` based multi-line statements (that I know of).  Prefered code style for this is.

```python
fee_to_be_charged = (
    PAYGAS_gasprice * computation.transaction_context.transaction_gas_limit
)
# or
gas_limit = computation.transaction_context.transaction_gas_limit
fee_to_be_charged = PAYGAS_gasprice * gas_limit
```

evm/transaction_context.py
```diff
@@ -0,0 +1,48 @@
+from evm.validation import (
+    validate_canonical_address,
+    validate_sig_hash,
+    validate_uint256,
+)
+
+
+class BaseTransactionContext(object):
```
pipermerriam: Can you drop the `(object)`

evm/transaction_context.py
```diff
@@ -0,0 +1,48 @@
...
+    @property
+    def gas_price(self):
+        return self._gas_price
+
+    @property
+    def origin(self):
+        return self._origin
+
+
+class ShardingTransactionContext(BaseTransactionContext):
```
pipermerriam: I think this should be located in `evm.vm.forks.sharding.transaction_context`

evm/transaction_context.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, gas_price, origin, sig_hash, transaction_gas_limit):
+        super(ShardingTransactionContext, self).__init__(gas_price, origin)
```
pipermerriam: Lets use python3 super calls.

```python
super().__init__(gas_price, origin)
```

tests/json-fixtures/test_state.py
```diff
@@ -295,6 +302,8 @@ def test_state_fixtures(fixture, fixture_vm_class):
         computation, _ = vm.apply_transaction(transaction)
     except ValidationError as err:
         transaction_error = err
+        LOGGER.warn("Got transaction error:")
```
pipermerriam: Was this left in on purpose.  I'm not sure that these logging statements are needed and I'm worried that they could cause confusion since errors here are expected and part of the tests.

At minimum, I'd like a comment added explaining as much, but I also think simply reverting this change should be fine.

tests/json-fixtures/test_state.py
```diff
@@ -295,6 +302,8 @@ def test_state_fixtures(fixture, fixture_vm_class):
...
+        LOGGER.warn(traceback.format_exc())
```
pipermerriam: Just FYI

```python
logger.warn("Got transaction error", exc_info=True)
```

This can be used anywhere inside an `except` block to include a traceback of the exception in the log message (at whatever log level you specify).

evm/logic/system.py
lrettig: Thanks. I actually looked for this in the tactical manual and couldn't find it :)

evm/transaction_context.py
lrettig: Sure. I think I was confused because `evm/vm/message.py` contains `ShardingMessage`.

tests/json-fixtures/test_state.py
lrettig: This went into master in #327 (https://github.com/ethereum/py-evm/pull/327/files#diff-412113be18fef671d701f02096a24e07). I found this invaluable when I was testing, since tests (which were supposed to pass) were failing, and the exceptions were being swallowed silently. I agree this is tricky. Maybe the thing to do here is to switch this from `warn` to `info` or `debug` and add a comment? We should probably do the same to master.

tests/json-fixtures/test_state.py
lrettig: Good to know, thanks!

evm/transaction_context.py
lrettig: Should backport this to master as well

lrettig:
Made the requested changes. If this looks good I'll finish squashing the commits!

evm/transaction_context.py
pipermerriam: Yeah, that's simply misplaced.  Should be moved into sharding at some point.

pipermerriam:
when CI shows green we can squash and merge.

lrettig:
Squashed commits, tests are passing, this should be ready for merge!

PR #367, opened 2018-02-10
Add probe_PAYGAS_mode flag to vm_state.execute_transaction

NIC619:
### What was wrong?
Miners need to check if they are getting paid before execute the whole transaction.

### How was it fixed?
Add a mode to test if `PAYGAS` is triggered in the transaction.
This is my first instinct of how to implement this test mode, but there could be better ways to do this instead of adding a flag to `vm_state.execute_transaction`. I was trying not to introduce too many changes.
If `probe_PAYGAS_mode` is on, the message is allocated with `probe_mode_gas_allocated` instead of `transaction.gas` and return immediately once the computation is done. Miner will either tell by `computation.get_PAYGAS_gas_price()` or `OutOfGas` error. 

There are a few things needed to be worked out.
- Stop the computation immediately once `PAYGAS` is triggered. By adding a flag to `PAYGAS` function and maybe raise an vm exception once triggered? or create a new way to exit computation for this. But since the intention is only to avoid further computation waste(Because we have to execute the transaction again, for real.) and gas allocated for this test mode is rather low, this might not be necessary.
- The occasion to run this function. We should not run this test mode for the transactions that are already included in a mined block because it's allowed to not paying gas to the miner(`coinbase`).

#### Cute Animal Picture

### DIFF
--- a/evm/vm/forks/sharding/vm_state.py
+++ b/evm/vm/forks/sharding/vm_state.py
@@ -46,7 +46,7 @@ class ShardingVMState(ByzantiumVMState):
     computation_class = ShardingComputation
     trie_class = BinaryTrie
 
-    def execute_transaction(self, transaction):
+    def execute_transaction(self, transaction, probe_PAYGAS_mode=False, probe_mode_gas_allocated=100000):
         # state_db ontext manager that restricts access as specified in the transacion
         state_db_cm = functools.partial(self.state_db, access_list=transaction.prefix_list)
 
@@ -61,7 +61,10 @@ class ShardingVMState(ByzantiumVMState):
 
         with state_db_cm() as state_db:
             # Setup VM Message
-            message_gas = transaction.gas - transaction.intrinsic_gas
+            if probe_PAYGAS_mode:
+                message_gas = probe_mode_gas_allocated
+            else:
+                message_gas = transaction.gas - transaction.intrinsic_gas
 
             if transaction.code:
                 contract_address = generate_CREATE2_contract_address(
@@ -142,6 +145,8 @@ class ShardingVMState(ByzantiumVMState):
         else:
             computation = self.get_computation(message).apply_message()
 
+        if probe_PAYGAS_mode:
+            return computation
         #
         # 2) Post Computation
         #


pipermerriam:
This feels like something that deserves it's own dedicated API.

(Also I'm generally a strong advocate against *flag* based functionality)

Also, this doesn't feel like EVM functionality, but rather *meta* functionality that those who mine blocks would use.

What do you think about implementing this so that it's not tightly coupled with the VMState.  Maybe as a stand-alone function or a mixin-class which can be added to the VM state which adds this extra functionality.

Alternate idea is for us to go ahead and write a basic abstraction for the `execute_transaction` functionality.  It's currently cut into 3 distinct logical sections.  It would be pretty straight forward to create a class which implements hooks for each of these sections.

```python
class TransactionExecution:
    def __call__(self, vm_state, transaction):
        message = self.run_pre_computation(vm_state, transaction)
        computation = self.run_computation(vm_state, message)
        return self.run_post_computation(vm_state, computation)
```

> Note that the actual *correct* division of logic could be different, this is just mirroring the current way the function body is divided up.

Thoughts?

hwwhww:
@NIC619  "Miners need to check if they are getting paid before execute the whole transaction." - what if they don't check and they pick the *bad* transactions? Would that cause any consensus fault? If not, is it a nice-to-have API?

I like the idea of `TransactionExecution` (3 distinct logical sections) for benefiting long-term Trinity production! But if the `TransactionExecution` *only* makes more sense because the procedure of `ShardingVMState.execute_transaction` consists overlapping logic of `FrontierVMState.execute_transaction`, I think this refactoring is low-priority here if "a stand-alone function or a mixin-class" can fix the problem.

pipermerriam:
>  I think this refactoring is low-priority here if "a stand-alone function or a mixin-class" can fix the problem.

Agreed, I'll open an issue for the transaction execution abstraction to be done separately.

NIC619:
> Also, this doesn't feel like EVM functionality, but rather meta functionality that those who mine blocks would use.

You are right. It's does not feel like EVM functionality.
What do you mean by meta here? I'm looking for where to place this API but can't find a proper location and now have a feeling that it's not really among the top priorities since it's only needed for miners. Maybe close this for now and open an TODO issue for this functionality instead?

@hwwhww By bad transaction, do you mean transactions that don't pay fees? Transaction with no fee is allowed so I don't think this will result in consensus failure. Miners can choose whether to activate this or not but I think this is needed for the change that `PAYGAS` brings.

pipermerriam:
> What do you mean by meta here?

Something that *can* exist as a layer on top of py-evm (or maybe trinity).

> I'm looking for where to place this API but can't find a proper location and now have a feeling that it's not really among the top priorities since it's only needed for miners.

You're not missing anything.  We don't have a place for that type of thing currently.  I'd be fine with it being added into `evm.utils` somewhere, or like your said.  If you don't have an immediate use for it, maybe punt and wait until it's needed before adding it to the codebase.

PR #369, opened 2018-02-12
Add nonce abstraction tests

NIC619:
### How was it fixed?
Add tests for nonce abstraction.
- One contract keeps track of the nonce in it's storage and check the validity of the value passed in. Thus transaction replayed with old transaction will throw exception before doing anything else(paying transaction fee or transfer value etc.)
- The other one does not. So it's valid to replay old transactions and drain the contract.

#### Cute Animal Picture


tests/core/vm/test_nonce_abstraction.py
```diff
@@ -0,0 +1,139 @@
...
+from evm.utils.padding import pad32
+
+from tests.core.helpers import (
+    new_sharding_transaction,
+)
+
+
+DIR = os.path.dirname(__file__)
+nonce_tracking_contracts = json.load(
+    open(os.path.join(DIR, '../contract_fixtures/nonce_tracking_contracts.json'))
```
pipermerriam: Note to self:

This would make a good generic fixture.  Something that loads all of the json files from `contract_fixtures` and exposes them as a generic fixture so that they can be used across the entire test suite without any boilerplate.

tests/core/fixtures.py
```diff
@@ -139,6 +146,16 @@ def chain(chaindb, funded_address, funded_address_initial_balance):  # noqa: F81
...
     },
+    {
+        "contract_code": nonce_tracking_contracts['nonce_tracking_contract']['bytecode'],
+        "deployed_address": nonce_tracking_contracts['nonce_tracking_contract']['address'],
+        "initial_balance": 100000000,
+    },
+    {
+        "contract_code": nonce_tracking_contracts['no_nonce_tracking_contract']['bytecode'],
+        "deployed_address": nonce_tracking_contracts['no_nonce_tracking_contract']['address'],
+        "initial_balance": 100000000,
```
hwwhww: You can replace all `100000000` with `funded_address_initial_balance` fixture now.

tests/core/contract_fixtures/nonce_tracking_contract.py
```diff
@@ -0,0 +1,23 @@
...
+                                ['lll',
+                                    ['seq',
+                                        ['assert', ['eq', ['calldataload', 32], ['sload', 0]]],
+                                        ['paygas', ['calldataload', 64]],
+                                        ['sstore', 0, ['add', ['sload', 0], 1]],
+                                        'stop'],
+                                0]]]  # noqa: E128
+
+
+# Contract that keeps track of it's nonce
```
hwwhww: According to the variable name `no_nonce_tracking_lll_code` and codes, I think this comment should be "Contract that doesn't keep track of its nonce".

tests/core/contract_fixtures/nonce_tracking_contract.py
```diff
@@ -0,0 +1,23 @@
...
```
hwwhww: it's -> its

Issue #312, opened 2018-01-26
Apply the new witness structure to `create_block` function


hwwhww:
### What is wrong?
Apply new protocol changes to the stateless `create_block` function.
https://github.com/ethereum/py-evm/blob/sharding/evm/vm/base.py#L182-L233

### How can it be fixed
- [ ] Some witness utils are implemented in #288, see if we can use it to generate the witness data.
- [ ] Return the block (collation) witness
- [ ] Better to move the API entry to `Chain` object

hwwhww:
### Updates status
A workable messy branch: https://github.com/ethereum/py-evm/compare/sharding...hwwhww:create_block3

Basically, this branch will be torn to two parts:
1. `master` branch:
    * Add the patch for switching trie
    * Remove `BaseVM.create_block` <--- Don't waste time to gather the `HexaryTrie` witness.
2. `sharding` branch:
    * Implement `ShardVM.create_block`
    * Implement `Shard.apply_block_with_witness` and `ShardVM.apply_block_with_witness` APIs <-- probably needs a better naming. 
    * Refactor the tests
    * Refine the witness APIs

PR #372, opened 2018-02-15
Some clean up for sharding

hwwhww:
### What was wrong?
As mentioned at https://github.com/ethereum/py-evm/issues/312#issuecomment-365233568, this is the part 1 (for master branch).

### How was it fixed?
1. Backported from sharding branch:
    * fixture refactoring
2. Removed `create_block` in master branch. Will update a `BinaryTrie` version in `sharding` branch.
3. Refactored and fixed trie switching
4.  `evm.utils.db.get_empty_root_hash function` -> `chaindb.empty_root_hash`

#### Cute Animal Picture

### DIFF
--- a/evm/utils/db.py
+++ b/evm/utils/db.py
@@ -1,14 +1,3 @@
-from trie import (
-    BinaryTrie,
-    HexaryTrie,
-)
-
-from evm.constants import (
-    BLANK_ROOT_HASH,
-    EMPTY_SHA3,
-)
-
-
 def make_block_number_to_hash_lookup_key(block_number):
     number_to_hash_key = b'block-number-to-hash:%d' % block_number
     return number_to_hash_key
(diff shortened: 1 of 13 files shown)


evm/utils/db.py
```diff
@@ -30,3 +41,20 @@ def get_block_header_by_hash(block_hash, db):
     Returns the header for the parent block.
     """
     return db.get_block_header_by_hash(block_hash)
+
+
+def get_empty_root_hash(db):
```
pipermerriam: This strikes me as a good `classmethod` for the `ChainDB` class.  That's basically what it currently is doing as I believe that you could do the following and it would just work:

```python
class ChainDB:
    get_empty_root_hash = classmethod(get_empty_root_hash)
```

I'm not proposing we do the above, but more stating that the current implementation is equivalent to just implementing this as a class method.

hwwhww:
@pipermerriam 
Thank you for the advice! I looked the `get_empty_root_hash` util usage again and found out that maybe it would be clearer with these modifications in https://github.com/ethereum/py-evm/pull/372/commits/7c55bf7abb261d0213adcc829c0238f4d15b7b51:
1. make `BaseChainDB.set_trie(trie_class)` function that sets `ChainDB.trie_class` and `ChainDB.empty_root_hash` in `ChainDB.__init__(db, trie_class)`.
2. Instead of calling `get_empty_root_hash(chaindb)`, just calling `chaindb.empty_root_hash`.

PR #376, opened 2018-02-16
DOC: building-chains: syncing documentation with source code

lacabra:
### What was wrong?

the documentation available at http://py-evm.readthedocs.io/en/latest/building_chains.html is out of sync with the source code on how to setup a LightChain

### How was it fixed?

Documentation was updated providing instructions that work with the actual codebase

#### Cute Animal Picture

### DIFF
--- a/docs/building_chains.rst
+++ b/docs/building_chains.rst
@@ -58,10 +58,10 @@ with a vm_configuration, but it also requires a network_id and privkey:
   from evm.chains.mainnet import MAINNET_VM_CONFIGURATION, MAINNET_NETWORK_ID
   from evm.p2p import ecies
   from evm.p2p.lightchain import LightChain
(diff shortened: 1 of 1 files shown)


carver:
Maybe this would be a good time to try out a doctest, so we catch it when the docs fall out of sync.

pipermerriam:
@lacabra you can check out how @carver implemented doctests in the `eth-hash` codebase (I think that's where it was).

carver:
It was in the local signing code of web3: https://raw.githubusercontent.com/ethereum/web3.py/master/docs/web3.eth.account.rst

pipermerriam:
Going ahead and merging as-is.  Some doctests would still be great @lacabra if you want to take that on.  Otherwise, thanks for getting this updated.

PR #378, opened 2018-02-19
dirty

pipermerriam:
Trying to chase down whatever is leaving open file descriptors in the test runs.


pipermerriam:
Looks like RPC test failures are present on the base trinity branch....

PR #379, opened 2018-02-20
Initial Trinity Implementation

pipermerriam:
STUB


pipermerriam:
@gsalgado I want to go ahead and merge this into master and continue work from there.  Everything in here was reviewed at some point, but I think another pass would be good before it gets merged.

trinity/db/chain.py
```diff
@@ -0,0 +1,61 @@
...
+
+    get_block_header_by_hash = sync_method('get_block_header_by_hash')
+    get_canonical_head = sync_method('get_canonical_head')
+    get_score = sync_method('get_score')
+    header_exists = sync_method('header_exists')
+    lookup_block_hash = sync_method('lookup_block_hash')
+    persist_header_to_db = sync_method('persist_header_to_db')
+
+
+class AsyncChainDB(ChainDB):
```
gsalgado: I don't understand why we need those async methods wrapping the sync ones. Can't we just keep using the sync ones?

pipermerriam: This object is only used in testing.  The object above this `ChainDBProxy` is what is actually used when the lightchain is run.  It delegates all of it's calls to an instance of `ChainDB` that's running in another process.

gsalgado: Would it make sense to move AsyncChainDB to test_peer.py, then?

trinity/utils/db.py
```diff
@@ -0,0 +1,35 @@
+class MemoryDB:
```
gsalgado: Can't we use `evm.db.backends.MemoryDB`?

trinity/main.py
```diff
@@ -0,0 +1,249 @@
...
+# Add console sub-command to trinity CLI.
+subparser = parser.add_subparsers(dest='subcommand')
+console_parser = subparser.add_parser('console', help='start the trinity REPL')
+console_parser.add_argument(
+    '--vanilla-shell',
+    action='store_true',
+    default=False,
+    help='start a native Python shell'
+)
+console_parser.set_defaults(func=console)
```
gsalgado: IIUC, this will run the `LightChain` in a thread and drop into a python shell, right? In that case, should we remove evm/lightchain_shell.py which was just  a quick hack anyway?

trinity/main.py
```diff
@@ -0,0 +1,249 @@
...
+            await peer_pool.stop()
+            await chain.stop()
+
+    chain = chain_class(chaindb, peer_pool)
+
+    loop = asyncio.get_event_loop()
+
+    try:
+        loop.run_until_complete(run())
+    except KeyboardInterrupt:
```
gsalgado: Will we ever hit this code path given that this is running in a sub-process? ISTM a KeyboardInterrupt would happen only in the main() function above. Also, instead of catching KeyboardInterrupt it'd be better to install a signal handler as that'd cover both a user hitting Ctrl-C in the terminal as well as the process being `kill`ed

trinity/db/chain.py
pipermerriam: > Would it make sense to move AsyncChainDB to test_peer.py, then?

Yup, will move.

trinity/main.py
pipermerriam: Yes, we can remove `evm/lightchain_shell.py`.
 
Also, anything console related can largely be ignored in this PR as it will be significantly overhauled very soon to instead use IPC connections to access the database.

trinity/main.py
pipermerriam: I just saw some of the signal handler stuff in one of your recent branches.  I'm going to open an issue to track this as handling keyboard interrupts is currently not working well and needs to be overhauled to use that mechanism.

pipermerriam:
Going to merge and then clean up individual things in smaller subsequent pull requests.  This Pr is too big.

PR #380, opened 2018-02-20
pin version of dependencies within major version ranges

pipermerriam:
### What was wrong?

Some of the versions were not pinned under a maximum version.

### How was it fixed?

Added additional `<` version pins to the version numbers.

#### Cute Animal Picture

jon-snow


pipermerriam:
Needs https://github.com/ethereum/eth-keys/pull/34 to be merged and released first.

pipermerriam:
Needs https://github.com/ethereum/eth-hash/pull/4 to be merged and released.

tox.ini
```diff
@@ -40,9 +40,9 @@ commands=
     native-state-constantinople: py.test {posargs:tests/json-fixtures/test_state.py -k Constantinople}
     native-state-metropolis: py.test {posargs:tests/json-fixtures/test_state.py -k Metropolis}
 deps = -r{toxinidir}/requirements-dev.txt
-    coincurve
-    database: leveldb
-    trinity: leveldb
+    coincurve>=7.0.0,<8.0.0
+    database: leveldb>=0.194,<1.0.0
+    trinity: leveldb>=0.194,<1.0.0
```
carver: Or move these to the extras config, so we don't have to duplicate the version numbers from setup.py.

```
extras = 
    coincurve
    database: leveldb
    trinity: trinity
```

evm/p2p/auth.py
```diff
@@ -5,17 +5,25 @@
...
 
-from eth_keys import keys
-from eth_keys import datatypes
+from eth_keys import (
+    datatypes,
+    keys,
+)
+
+from eth_hash.main import (
+    Digest,
```
carver: This is called Preimage now, right?

evm/utils/fixture_tests.py
```diff
@@ -205,8 +206,14 @@ def normalize_to_address(value):
 
 def robust_decode_hex(value):
     unprefixed_value = remove_0x_prefix(value)
-    if len(unprefixed_value) % 2:
-        return decode_hex(pad_left(unprefixed_value, len(unprefixed_value) + 1, b'0'))
+
+    if is_bytes(value):
+        text_value = codecs.decode(unprefixed_value, 'utf8')
```
carver: _(if I'm wrong about the whole method being replaceable by `hexstr_if_str(to_bytes)`)_

I don't think we want to decode to text using utf-8 here. Guessing from the function context, I think we would want to convert to a hex string.

evm/utils/fixture_tests.py
```diff
@@ -205,8 +206,14 @@ def normalize_to_address(value):
 
 def robust_decode_hex(value):
     unprefixed_value = remove_0x_prefix(value)
-    if len(unprefixed_value) % 2:
-        return decode_hex(pad_left(unprefixed_value, len(unprefixed_value) + 1, b'0'))
+
+    if is_bytes(value):
```
carver: `remove_0x_prefix` will crash on a `bytes` value before we get here.

I think this method definition is equivalent to `robust_decode_hex = eth_utils.curried.hexstr_if_str(to_bytes)`, which is equivalent to: `to_bytes(hexstr=value) if is_string(value) else to_bytes(value)`

evm/utils/fixture_tests.py
pipermerriam: you were correct, was able to replace the whole thing.

pipermerriam:
replaced by #389

PR #382, opened 2018-02-20
Add JSON-RPC server process to trinity

pipermerriam:
### What was wrong?

We need a process run within trinity to serve JSON-RPC requests.

### How was it fixed?

Temporary solution to run it alongside the networking process.

This also adds `trinity attach` and `trinity console` commands.

#### Cute Animal Picture

flying_cat_by_stevewaite-d4212cp


pipermerriam:
@gsalgado I think this is going to depend on some of the cancel token work that you're doing.  Basic local testing has the following things working in this PR.

- Run the chain normally `$ trinity`
- Run the chain and a python shell `$ trinity console` (this is really unusable due to the amount of logging output)
- Run the chain in one shell `$ trinity` and then attach from another shell `$ trinity attach`

The console has been changed to not expose the `chain` object directly, but rather a `Web3` instance via the variable `w3`.  Let me know what you think.  This all feels a little hacky but I'm inclined to move forward with some hacky solutions to finally get the first public alpha out the door and then to try and solidify some of this based on user feedback.

trinity/main.py
```diff
@@ -175,5 +170,22 @@ class DBManager(BaseManager):
...
 
-    loop.run_until_complete(run_lightchain(chain))
+    # setup the JSON-RPC server
+    rpc = RPCServer(chain)
+    asyncio.ensure_future(asyncio.start_unix_server(
+        connection_handler(rpc.execute),
+        chain_config.json_rpc_ipc_path,
+        loop=loop,
+        limit=MAXIMUM_REQUEST_BYTES,
+    ), loop=loop)
```
gsalgado: I guess this is where the `CancelToken` would be needed? I believe we could move this into a coroutine method under `RPCServer`, where we could do something like

```
async def run(self):
    self.srv = await asyncio.start_unix_server(...)
    self.srv.close()

async def wait_closed(self):
    await self.srv.wait_closed()
```

Then `run_chain()` would become something like:

```
async def run_chain():
    try:
        asyncio.ensure_future(chain.peer_pool.run())
        asyncio.ensure_future(rpc.run())
        await chain.run()
    finally:
        ...
        rpc.cancel_token.trigger()
        await rpc.wait_closed()
```
Or, if you don't want to _infect_ the `RPCServer` with asyncio code, I guess those could be standalone functions here

gsalgado: Oh, I just realized you'll probably also need to pass the `CancelToken` to `connection_handler()` and from there to `connection_loop()`, where you'd then use `wait_with_token()` and catch `OperationCancelled` to cleanly terminate pending coroutines when we stop the RPC server

pipermerriam: Good catch, will do.

pipermerriam:
Ping @gsalgado need review on this again now that I *maybe* got everything working.

trinity/rpc/ipc.py
```diff
@@ -1,33 +1,46 @@
...
-async def connection_handler(execute_rpc, reader, writer):
+async def connection_handler(execute_rpc, cancel_token, reader, writer):
     '''
     Catch fatal errors, log them, and close the connection
     '''
     logger = logging.getLogger('trinity.rpc.ipc')
+
     try:
-        await connection_loop(execute_rpc, reader, writer, logger)
+        await wait_with_token(
```
pipermerriam: Is this enough? is this in the right place?

gsalgado: Nope, you need to pass the token all the way down until the function/method that calls the asyncio APIs, and there you use `wait_with_token()`. In this case that seems to be `connection_loop()` and `write_error()`

trinity/rpc/ipc.py
```diff
@@ -88,44 +102,27 @@ def strip_non_json_prefix(raw_request):
...
+    def __init__(self, rpc, ipc_path):
+        self.rpc = rpc
+        self.ipc_path = ipc_path
+        self.cancel_token = CancelToken('IPCServer')
+
+    async def run(self, loop=None):
+        self.server = await asyncio.start_unix_server(
+            connection_handler(self.rpc.execute, self.cancel_token),
+            self.ipc_path,
+            loop=loop,
```
gsalgado: Although I usually prefer explicit over implicit, I've been a bit reluctant about ading the `loop` argument to our APIs as ISTM at this point it's only needed for tests, and we can work around that simply by having our fixtures depend on the `event_loop` fixture (even though we don't use the `event_loop`, like in https://github.com/ethereum/py-evm/blob/master/tests/p2p/test_kademlia.py#L22-L26). If you're ok with that I'd be happy to take care of documenting this properly and seeing if there's any way to provide better error messages when our tests fail because we end up accidentally using different loops

tests/trinity/conftest.py
```diff
@@ -56,10 +59,12 @@ def ipc_server(jsonrpc_ipc_pipe_path, event_loop):
     the course of all tests. It never needs to be actually
     used as a fixture, so it doesn't return (yield) a value.
     '''
-    server = start(jsonrpc_ipc_pipe_path, loop=event_loop)
+    rpc = RPCServer(None)
+    ipc_server = IPCServer(rpc, jsonrpc_ipc_pipe_path)
+
+    asyncio.ensure_future(ipc_server.run(loop=event_loop), loop=event_loop)
```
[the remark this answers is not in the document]
pipermerriam: Removing these causes tests to start failing.

- removal from `ensure_future`  causes the IPC socket to never open
- removal from `run` yells about the event loop already being closed.

pipermerriam:
When I tried passing `cancel_token` down into `write_error` and wrapping `writer.drain()` in `wait_with_token`, things started breaking.  Take a look at 6ac2ec4 to see the changes I made.

gsalgado:
I think this should be enough as `write_error()` only uses `await` on `writer.drain()`

trinity/utils/chains.py
```diff
@@ -81,6 +81,21 @@ def get_database_socket_path(chain_identifier: str, data_dir: str=None) -> str:
     )
 
 
+JSONRPC_SOCKET_FILENAME = 'jsonrpc.ipc'
+
+
+def get_jsonrpc_socket_path(chain_identifier, data_dir=None):
```
cburgdorf: There are some method signatures in this PR like this one that are missing type hints. I guess we want all new Trinity code to land as typed right from the start.

pipermerriam: I originally did this PR including the addition of type hints, but I had a bunch of hard-to-debug failures and I reverted everything to the minimal implementation that worked.  I'm going to do type hints for this stuff right away after this is merged (assuming it passes CI).

cburgdorf: :+1:

Issue #368, opened 2018-02-11
Create generic transaction execution abstraction


pipermerriam:
### What is wrong?

Currently, there is a lot of code duplication in the `execute_transaction` functions.

- [Frontier](https://github.com/ethereum/py-evm/blob/9df7a3d02bc8a1a2d898d010c63ab7d567950d19/evm/vm/forks/frontier/vm_state.py#L45-L185)
- [Sharding](https://github.com/ethereum/py-evm/blob/sharding/evm/vm/forks/sharding/vm_state.py#L49-L184)

### How can it be fixed

A basic abstraction for the `execute_transaction` functionality. 

The functions are currently divided into 3 distinct logical sections.  It would be pretty straight forward to create a class which implements hooks for each of these sections.

```python
class BaseTransactionExecution:
    def __call__(self, vm_state, transaction):
        message = self.run_pre_computation(vm_state, transaction)
        computation = self.run_computation(vm_state, message)
        return self.run_post_computation(vm_state, computation)

    def run_pre_computation(vm_state, transaction):
        raise NotImplementedError(...)

    def self.run_computation(vm_state, message):
        raise NotImplementedError(...)

    def run_post_computation(vm_state, computation):
        raise NotImplementedError(...)
```

onyb:
@pipermerriam I plan to work on this. Do we want the abstraction only on `sharding` branch, or also want it backported to `master`?

pipermerriam:
@onyb 👍 

Lets do it in `master`.  We should have 

- a stubbed out `Base` implementation that only raises `NotImplementedErrors`.  
- A `FrontierXXX` implementation with the current frontier logic.
- A `SpuriousDragonXXX` implementation with the small bit of *extra* logic in the `run_post_computation` portion.

PR #383, opened 2018-02-21
Generic transaction executor

onyb:
**Note:** Purpose of this pull-request is to get the ball rolling on issue #368 and come up with an initial refactoring so as to formalize the API.

### What was wrong?

`execute_transaction` functionality for different forks could benefit from some code re-use.

### How was it fixed?

I created a `BaseTransactionExecutor` to be implemented by Frontier and SpuriousDragon forks (for now). The implementations, overriding the `run_pre_computation`, `run_computation`, and `run_post_computation` methods, are used as mixins in the corresponding VM state classes.

My API of `BaseTransactionExecutor` does not comply with what was suggested in #368. This is largely due to strong coupling between the three computation phases, thereby not allowing me to split them in a clean fashion. This is the API in this pull-request:
- `self.execute_transaction(transaction)`
- `self.run_pre_computation(transaction)`
- `self.run_computation(transaction, message)`
- `self.run_post_computation(transaction, computation)`

There is some scope for improvement in the above API, notably that we shouldn't need to pass `transaction` as the first argument to `self.run_computation` and `self.run_post_computation`. Below are some ideas that I would like to discuss.

- Move `gas_price` to `evm.vm.message.Message`. This will allow us to avoid coupling due to `transaction_context`. I believe this was the case on `sharding` branch at some point, but was later changed.
- Specify the message class as a class attribute of the transaction executor (example below). This will be useful for the executor in the sharding fork, since the message class is something else (`ShardingMessage`).

    ```py
    from evm.vm.message import Message

    class FrontierTransactionExecutor(BaseTransactionExecutor):
        message_cls = Message

        def run_pre_computation(self, transaction):
            return self.message_cls(...)
    ```
- We should probably also try to enumerate refactoring goals in sharding/frontier code so as to re-use (parts of) the frontier txn executor in the sharding fork. For example, ironing out the difference in usage of `self.state_db` in the transaction executor logic of frontier vs sharding?

#### Cute Animal Picture

### DIFF
--- a/evm/vm_state.py
+++ b/evm/vm_state.py
@@ -303,13 +303,6 @@ class BaseVMState(Configurable, metaclass=ABCMeta):
         self.receipts.append(receipt)
 
     @abstractmethod
-    def execute_transaction(self, transaction):
-        """
-        Execute the transaction in the vm.
-        """
-        raise NotImplementedError("Must be implemented by subclasses")
-
-    @abstractmethod
     def make_receipt(self, transaction, computation):
         """
         Make receipt.
--- a/evm/vm_state.py
+++ b/evm/vm_state.py
@@ -360,3 +360,25 @@ class BaseVMState(Configurable, metaclass=ABCMeta):
     @abstractmethod
     def get_nephew_reward(cls):
         raise NotImplementedError("Must be implemented by subclasses")
+
+
+class BaseTransactionExecutor(metaclass=ABCMeta):
+    def execute_transaction(self, transaction):
+        """
+        Execute the transaction in the vm.
+        """
+        message = self.run_pre_computation(transaction)
+        computation = self.run_computation(transaction, message)
+        return self.run_post_computation(transaction, computation)
+
+    @abstractmethod
+    def run_pre_computation(self, transaction):
+        raise NotImplementedError()
+
+    @abstractmethod
+    def run_computation(self, transaction, message):
+        raise NotImplementedError()
+
+    @abstractmethod
+    def run_post_computation(self, transaction, computation):
+        raise NotImplementedError()
--- a/evm/vm/forks/frontier/vm_state.py
+++ b/evm/vm/forks/frontier/vm_state.py
@@ -27,6 +27,7 @@ from evm.vm.message import (
 )
 from evm.vm_state import (
     BaseVMState,
+    BaseTransactionExecutor,
 )
 
 from evm.utils.address import (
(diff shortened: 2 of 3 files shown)


pipermerriam:
@onyb fyi, our CI is failing on master right now.  should be fixed soon-ish.  Test failures are not your fault.

dylanjw:
@onyb I am hoping to make some changes that build on this pull request, so it would be nice to get this merged. I went ahead and rebased this with `master`, the result of which is here: https://github.com/dylanjw/py-evm/tree/generic-txn-exec.  If you have time would you like to update this PR with those changes? Let me know, otherwise I will create a new pull request with your rebased changes.

dylanjw:
> Specify the message class as a class attribute of the transaction executor (example below). This will be useful for the executor in the sharding fork, since the message class is something else (ShardingMessage).

What about removing the message creation from the Executor/VMState class entirely, and putting it in the transaction allowing something like: `transaction.as_message()`, and removing the transaction argument from `TransactionExecutor`, which could be renamed something more generic like `StateChange`, `StateTransition`. My hope is to have the executor class be re-usable for `call` and `estimateGas` methods, by moving out any transaction specific validation.

evm/vm_state.py
```diff
@@ -324,3 +318,19 @@ def get_uncle_reward(block_number, uncle):
...
         raise NotImplementedError("Must be implemented by subclasses")
+
+
+class BaseTransactionExecutor:
+    def execute_transaction(self, transaction):
+        message = self.run_pre_computation(transaction)
+        computation = self.run_computation(transaction, message)
+        return self.run_post_computation(transaction, computation)
+
+    def run_pre_computation(self, transaction):
```
pipermerriam: can you update these to be decorated with `abc.abstractmethod` as was recently added elsewhere in #467

evm/vm_state.py
```diff
@@ -324,3 +318,19 @@ def get_uncle_reward(block_number, uncle):
     @classmethod
     def get_nephew_reward(cls):
         raise NotImplementedError("Must be implemented by subclasses")
+
+
+class BaseTransactionExecutor:
```
pipermerriam: Can you update this to use the `abc.ABCMeta` metaclass:

```python
class BaseTraansactinExecutor(metaclass=ABCMeta):
    pass
```

onyb:
@dylanjw Done. Thanks!

@pipermerriam I rebased on master, and addressed your review comments.

onyb:
@dylanjw 

> What about removing the transaction to message conversion from the Executor/VMState class entirely? It could move to the transaction class. Something like: transaction.as_message(). Then the transaction argument in TransactionExecutor could be replaced with a Message argument.

It would be great if we can do this. However, I don't think the transaction class has enough context to generate the complete `Message` object, as required by the `run_computation` method. This is mainly because all the important pieces of information in the `Message` object that is not already present in the transaction, require access to the `self.state_db`.

Do you have ideas to get around this limitation?

Issue #388, opened 2018-02-21
Remove `evm.utils.padding` in favor of built-in ljust and rjust methods.


pipermerriam:
### What is wrong?

The utility functions `pad_left` and `pad_right` are not necessary and are likely causing a performance hit.

### How can it be fixed

Replace all usage of them with calls to `str.ljust` `str.rjust`, `bytes.ljust`, `bytes.rjust`.

cburgdorf:
I think this can be closed now (forgot to close it from the commit message)

PR #390, opened 2018-02-22
Refactor pad utilities to use Python internals (#388)

cburgdorf:
I tried to adress #388 but I'm still very much a Python noob (in fact, this is my first contribution to a Python project ever)

@pipermerriam I hope I refactored it the way you had it in mind. I first wasn't sure about the curried functions but think it turned out well (again, bloody Python noob here!).

### DIFF
--- a/evm/logic/context.py
+++ b/evm/logic/context.py
@@ -9,9 +9,6 @@ from evm.utils.address import (
 from evm.utils.numeric import (
     ceil32,
 )
-from evm.utils.padding import (
-    pad_right,
-)
 
 
 def balance(computation):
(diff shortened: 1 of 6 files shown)


cburgdorf:
Ok, turns out. This still has plenty of tests failing. Looking into it...

pipermerriam:
Looks like you did the following replacement:

- `pad_left -> ljust`
- `pad_right > rjust`

It should actually be the other way around.

- `pad_left > rjust`
- `pad_right > ljust`

cburgdorf:
Ah yes, thank you! So that's what happens when you blindly replace stuff. Should be fixed now.

PR #393, opened 2018-02-22
Fix fallout from trinity changes

gsalgado:


evm/p2p/lightchain.py
```diff
@@ -325,7 +325,7 @@ def _test():
         network_id=NETWORK_ID,
     )
 
-    chaindb = ChainDB(LevelDB(args.db))
+    chaindb = AsyncChainDB(LevelDB(args.db))
```
pipermerriam: What do you think about removing this whole `_test` section and relying on just the `trinity` CLI.  With a bit of added CLI functionality you should be able to add the equivalent functionality for the `LocalGethPeerPool` via standard CLI parameters.

evm/db/chain.py
```diff
@@ -545,3 +545,24 @@ def clear(self):
     #
     def get_state_db(self, state_root, read_only):
         return AccountStateDB(db=self.db, root_hash=state_root, read_only=read_only)
+
+
+# XXX: This should be just an interface, the actual implementation should be moved somewhere else.
```
pipermerriam: Do I understand this correcty to mean:

- change the class name to be something like `BaseAsyncChainDB`
- change these methods to raise `NotImplementedError` exceptions.
- add an implementation that looks like this one somewhere in trinity, maybe in the trinity tests, for use in cases where we don't want to deal with the overhead of setting up the database process/manager.
- probably add a docstring with a disclaimer/warning to that implementation explaining why this implementation **cannot** be used for production networking connections due to it blocking all of the networking processes.

gsalgado: Yep, that's the plan. And if we can keep the implementation in a `test_` file, it shouldn't even need the disclaimer/warning

evm/p2p/lightchain.py
gsalgado: Makes sense, I'll do that

Issue #374, opened 2018-02-16
Merge sharding branch into master branch


hwwhww:
### What is wrong?
We can merge `sharding` branch into `master` branch to reduce additional overheads.

### How can it be fixed
1. Rebase sharding branch
2. Clean up
3. Make a PR to make sure the sharding stuff is isolated enough

hwwhww:
close via  #397.

PR #397, opened 2018-02-23
Merge sharding into master branch

hwwhww:
### What was wrong?
#374

A glimpse of what we need to clean.

#### Cute Animal Picture


evm/chains/chain.py
```diff
@@ -1,405 +1,7 @@
...
-)
-from evm.utils.hexadecimal import (
-    encode_hex,
-)
-from evm.utils.rlp import (
-    ensure_imported_block_unchanged,
-)
-
-
-class Chain(Configurable):
```
hwwhww: Note to self: make sure that the latest and `Chain` is moved to `BaseChain` completely.

evm/rlp/headers.py
```diff
@@ -148,3 +159,117 @@ def clone(self):
...
+                 shard_id,
+                 expected_period_number,
+                 period_start_prevhash,
+                 parent_hash,
+                 number,
+                 transaction_root=EMPTY_SHA3,
+                 coinbase=ZERO_ADDRESS,
+                 state_root=EMPTY_SHA3,
+                 receipt_root=EMPTY_SHA3,
+                 sig=b""):
```
hwwhww: TODO: No `sig` now.

evm/utils/db.py
```diff
@@ -30,3 +41,22 @@ def get_block_header_by_hash(block_hash, db):
     Returns the header for the parent block.
     """
     return db.get_block_header_by_hash(block_hash)
+
+
+def get_empty_root_hash(db):
+    # TODO: Backport and refactor evm.chains.base.BaseChain and
```
hwwhww: TODO: Remove this comment.

evm/vm/forks/sharding/computation.py
```diff
@@ -0,0 +1,174 @@
...
+)
+from evm.utils.hexadecimal import (
+    encode_hex,
+)
+from evm.validation import (
+    validate_uint256,
+)
+
+from ..spurious_dragon.computation import (
+    SpuriousDragonComputation,
```
hwwhww: TODO: Use `ByzantiumComputation`.

evm/exceptions.py
```diff
@@ -141,3 +149,25 @@ class OutOfBoundsRead(VMError):
...
+
+
+class GasPriceAlreadySet(Exception):
+    """
+    Error raised if trying to reset the already set gas price.
+    """
+    pass
+
+
+class NotTopLevelCall(Exception):
```
hwwhww: @NIC619  ditto

evm/exceptions.py
```diff
@@ -141,3 +149,25 @@ class OutOfBoundsRead(VMError):
...
+
+class UnannouncedStateAccess(VMError):
+    """
+    Error raised if a storage slot or account field is accessed that has not
+    been specified in the transaction's read or write list, respectively.
+    """
+    pass
+
+
+class GasPriceAlreadySet(Exception):
```
hwwhww: @NIC619  Should it be a `VMError`?
```python
class GasPriceAlreadySet(Exception):
```

evm/vm_state.py
```diff
@@ -97,14 +101,13 @@ def state_db(self, read_only=False):
         elif self.state_root != state.root_hash:
             self.set_state_root(state.root_hash)
 
-        self.access_logs.reads.update(state.db.access_logs.reads)
-        self.access_logs.writes.update(state.db.access_logs.writes)
+            self.access_logs.reads.update(state.db.access_logs.reads)
```
hwwhww: Now I think `self.access_logs.reads.update(state.db.access_logs.writes)` should be in `elif self.state_root != state.root_hash` section while `self.access_logs.writes.update(state.db.access_logs.reads)` isn't.

evm/exceptions.py
pipermerriam: I think it's intentionally not a VM error to ensure that mis-use of this API results in a hard failure, requiring EVM code that uses this API to correctly handle all of the cases (such as the price already being set).

evm/vm_state.py
NIC619: Seems like a typo? self.access_logs.**reads**.update(state.db.access_logs.**writes**)
I suppose `self.access_logs.reads.update(state.db.access_logs.reads)` should be updated regardless of whether it's read or write?

evm/computation.py
```diff
@@ -303,6 +304,11 @@ def get_gas_remaining(self):
         else:
             return self.gas_meter.gas_remaining
 
+    @contextmanager
+    def state_db(self, read_only=False):
```
NIC619: Note: #405 split `state_db` into `read_only_state_db` and `mutable_state_db`.
I can wait for this one to get merged first and make consequent updates.

evm/rlp/headers.py
hwwhww: ^^^^ After merging, use #420 to fix it.

evm/exceptions.py
pipermerriam: The way it's used, I don't believe so.  This isn't something that halts VM execution, but rather to signal to the opcode logic that it's being called incorrectly and to push `0` onto the stack.

tox.ini
```diff
@@ -1,6 +1,7 @@
 [tox]
 envlist=
     py{35,36}-{core,p2p,database,transactions,vm,native-blockchain}
+    py{35,36}-{auxiliary, sharding}
```
cburgdorf: nitpick: I think this should go into the line above, no?

evm/logic/call.py
```diff
@@ -58,8 +58,9 @@ def __call__(self, computation):
         computation.gas_meter.consume_gas(child_msg_gas_fee, reason=self.mnemonic)
 
         # Pre-call checks
-        state_db = computation.vm_state.read_only_state_db
-        sender_balance = state_db.get_balance(computation.msg.storage_address)
+        with computation.state_db(read_only=True) as state_db:
+            sender_balance = state_db.get_balance(computation.msg.storage_address)
+
```
NIC619: TODO: After this is merged, I will make a PR to remove `Message.access_list` and `Computation.state_db` which should only apply to `ShardingVM` and another PR to convert this `state_db(read_only=True)`/`state_db()` back to `read_only_state_db`/`mutable_state_db`.

tox.ini
hwwhww: yeah, I think so.  👍
(we were trying to isolate sharding stuff, but they haven't been separated in `core` anyway)
I will modify it after merging.

PR #404, opened 2018-02-23
MAINT: updated requirements-dev.txt w/ leveldb

lacabra:
### What was wrong?

trinity.rpc.server imports evm.db.backends.level which in turns imports the leveldb package, but package is missing in local environment because it's not listed as a requirement

### How was it fixed?

added leveldb  package into requirements-dev.txt

#### Cute Animal Picture

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,7 +1,7 @@
 bumpversion==0.5.3
 flake8==3.5.0
 hypothesis==3.44.26
-leveldb==0.194
+leveldb>=0.194,<1.0.0
 pytest-asyncio==0.8.0
 pytest-cov==2.5.1
 pytest-logging>=2015.11.4
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,7 @@
 bumpversion==0.5.3
 flake8==3.5.0
 hypothesis==3.44.26
+leveldb==0.194
 pytest-asyncio==0.8.0
 pytest-cov==2.5.1
 pytest-logging>=2015.11.4


carver:
This dependency is already defined in some extras. Pinning the version in requirements-dev can make upgrading more error-prone down the road.

Maybe there's a doc that needs updating, to say that py-evm should be installed with `pip install py-evm[leveldb]`? (or really, probably `pip install py-evm[trinity]`)

lacabra:
ok, this makes sense. Feel free to reject this PR if that's the case. 

I believe then that `pip install py-evm[trinity]` should be documented in `docs/quickstart.rst` as you suggest.

A note to your comment on making upgrading more error-prone: almost other packages listed in `requirements-dev.txt` are pinned to a specific version, so I submitted this PR with a pinned version for consistency. I'm updating it to the same notation set in the setup file you reference, in case you would merge. 

Up to you.

carver:
> I believe then that pip install py-evm[trinity] should be documented in docs/quickstart.rst as you suggest.

That sounds great. Any interest in submitting the change?

> almost other packages listed in requirements-dev.txt are pinned to a specific version, so I submitted this PR with a pinned version for consistency.

Yeah, I can imagine a pretty strong case for pinning versions exactly in development environments (to minimize differences between devs). The issue I foresee with upgrading is because the same package is required in both `requirements-dev.txt` and `setup.py`. It is easy to accidentally only change one when trying to trigger an upgrade.

> Up to you.

I'll leave the PR open for now, in case you want to wipe the commits and make the change to the docs instead.

lacabra:
Yes, leave it open, I'll probably tackle it early next week.

pipermerriam:
Going to close since commit-wise there should be no overlap.  Feel free to copy/paste descriptions and whatnot when you open new pr.

PR #407, opened 2018-02-24
 Type Annotations for Trinity/utils

### LINKED ISSUE #386: Add type annotations to Trinity code
### What is wrong?

Need to add type annotations before Trinity code gets any larger.

### How can it be fixed

Add them

cburgdorf:
Related to #386 this PR brings type annotations for trinity/utils to kick things off.

### DIFF
--- a/trinity/utils/chains.py
+++ b/trinity/utils/chains.py
@@ -23,11 +23,13 @@ from .xdg import (
     get_xdg_trinity_root,
 )
 
+from typing import Union
+
 
 #
 # Filesystem path utils
 #
-def get_default_data_dir(chain_identifier):
+def get_default_data_dir(chain_identifier: str) -> str:
     """
     Returns the base directory path where data for a given chain will be stored.
     """
@@ -40,7 +42,7 @@ def get_default_data_dir(chain_identifier):
 DATABASE_DIR_NAME = 'chain'
 
 
-def get_database_dir(chain_identifier, data_dir=None):
+def get_database_dir(chain_identifier: str, data_dir: str=None) -> str:
     """
     Returns the directory path where chain data will be stored.
     """
@@ -52,7 +54,7 @@ def get_database_dir(chain_identifier, data_dir=None):
 NODEKEY_FILENAME = 'nodekey'
 
 
-def get_nodekey_path(chain_identifier, data_dir=None):
+def get_nodekey_path(chain_identifier: str, data_dir: str=None) -> str:
     """
     Returns the path to the private key used for devp2p connections.
     """
@@ -67,7 +69,7 @@ def get_nodekey_path(chain_identifier, data_dir=None):
 DATABASE_SOCKET_FILENAME = 'db.ipc'
 
 
-def get_database_socket_path(chain_identifier, data_dir=None):
+def get_database_socket_path(chain_identifier: str, data_dir: str=None) -> str:
     """
     Returns the path to the private key used for devp2p connections.
     """
@@ -82,7 +84,7 @@ def get_database_socket_path(chain_identifier, data_dir=None):
 #
 # Nodekey loading
 #
-def load_nodekey(nodekey_path):
+def load_nodekey(nodekey_path: str) -> PrivateKey:
     with open(nodekey_path, 'rb') as nodekey_file:
         nodekey_raw = nodekey_file.read()
     nodekey = keys.PrivateKey(nodekey_raw)
@@ -106,11 +108,11 @@ class ChainConfig:
     _network_id = None
 
     def __init__(self,
-                 chain_identifier,
-                 data_dir=None,
-                 nodekey_path=None,
-                 nodekey=None,
-                 network_id=None):
+                 chain_identifier: str,
+                 data_dir: str=None,
+                 nodekey_path: str=None,
+                 nodekey: PrivateKey=None,
+                 network_id: int=None) -> None:
         # validation
         if nodekey is not None and nodekey_path is not None:
             raise ValueError("It is invalid to provide both a `nodekey` and a `nodekey_path`")
@@ -127,7 +129,7 @@ class ChainConfig:
             self.nodekey = nodekey
 
     @property
-    def data_dir(self):
+    def data_dir(self) -> str:
         """
         The data_dir is the base directory that all chain specific information
         for a given chain is stored.  All other chain directories are by
@@ -139,19 +141,19 @@ class ChainConfig:
             return self._data_dir
 
     @data_dir.setter
-    def data_dir(self, value):
+    def data_dir(self, value: str) -> None:
         self._data_dir = os.path.abspath(value)
 
     @property
-    def database_dir(self):
+    def database_dir(self) -> str:
         return get_database_dir(self.chain_identifier, self.data_dir)
 
     @property
-    def database_ipc_path(self):
+    def database_ipc_path(self) -> str:
         return get_database_socket_path(self.chain_identifier, self.data_dir)
 
     @property
-    def nodekey_path(self):
+    def nodekey_path(self) -> str:
         if self._nodekey_path is None:
             if self._nodekey is not None:
                 return None
@@ -161,11 +163,11 @@ class ChainConfig:
             return self._nodekey_path
 
     @nodekey_path.setter
-    def nodekey_path(self, value):
+    def nodekey_path(self, value: str) -> None:
         self._nodekey_path = os.path.abspath(value)
 
     @property
-    def nodekey(self):
+    def nodekey(self) -> PrivateKey:
         if self._nodekey is None:
             try:
                 return load_nodekey(self.nodekey_path)
@@ -180,7 +182,7 @@ class ChainConfig:
             return self._nodekey
 
     @nodekey.setter
-    def nodekey(self, value):
+    def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
         if isinstance(value, bytes):
             self._nodekey = keys.PrivateKey(value)
         elif isinstance(value, PrivateKey):
@@ -192,7 +194,7 @@ class ChainConfig:
             )
 
     @property
-    def network_id(self):
+    def network_id(self) -> int:
         if self._network_id is not None:
             return self._network_id
 
--- a/trinity/utils/db.py
+++ b/trinity/utils/db.py
@@ -1,35 +1,38 @@
-class MemoryDB:
-    kv_store = None
+from typing import Dict
 
-    def __init__(self, kv_store=None):
+
+class MemoryDB:
+    kv_store = None  # type: Dict[bytes, bytes]
+
+    def __init__(self, kv_store: Dict[bytes, bytes]=None) -> None:
         if kv_store is None:
             self.kv_store = {}
         else:
             self.kv_store = kv_store
 
-    def get(self, key):
+    def get(self, key: bytes) -> bytes:
         return self.kv_store[key]
 
-    def set(self, key, value):
+    def set(self, key: bytes, value: bytes) -> None:
         self.kv_store[key] = value
 
-    def exists(self, key):
+    def exists(self, key: bytes) -> bool:
         return key in self.kv_store
 
-    def delete(self, key):
+    def delete(self, key: bytes) -> None:
         del self.kv_store[key]
 
     #
     # Dictionary API
     #
-    def __getitem__(self, key):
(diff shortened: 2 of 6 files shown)


trinity/utils/db.py
```diff
@@ -1,35 +1,40 @@
-class MemoryDB:
-    kv_store = None
+from typing import TypeVar, Generic, Dict
 
-    def __init__(self, kv_store=None):
+K = TypeVar('K')
+V = TypeVar('V')
+
+class MemoryDB(Generic[K, V]):
```
cburgdorf: @pipermerriam @gsalgado I went all in with typings here and made this generic. Is this how you would like to have it or would you rather want this stuff to be non-generic and annotated with `object`? Personally, I'd vote for making it generic but basically because I'm a fan of Generics in other languages. I'm a bloody Python newbie and don't know about the implications that may have.

gsalgado: I think type annotations should be as specific as possible, so here it'd be preferable to use `class MemoryDB(Mapping[bytes, bytes]):`. Although I've just noticed trinity's MemoryDB is not used anywhere, so we should probably remove it

cburgdorf: oh. So the `MemoryDB` sounded like something generic to me where the type of the `key` and `value` may depend on the use case. If you pin it down to `[bytes, bytes]` we'll only be able to use it as such (which may be fine if it's really only meant to be used in that way) but with `K` and `V` being generic one could use a `MemoryDB` somewhere where the key is a `str` and the` value` is a `byte` whereas somewhere else it would be used with both `key` and `value` being of type `byte` for instance. All that while still maintaining full type safety for both use cases.

But if the `MemoryDB` isn't used anyway, I'll just remove it in a separate commit.

gsalgado: Because of python's dynamic typing nature, any container type will be _generic_, which can be problematic if the actual implementation cannot handle elements of _all_ possible types. That's one of the cases where type hints are useful, as they help ensure those containers are only used with types they know how to handle. Also, at this point the type hints are not used to provide actual type safety -- they're only used by static analysis tools like mypy. You might want to read https://www.python.org/dev/peps/pep-0484/ to understand the background and goals for type hints

cburgdorf: Yes, I get that. That's the same with TypeScript which just adds type hints on top of JavaScript. But still `mypy` (or `tsc` in the TypeScript case) will enforce the type constraints at **development time** and save you tons of trouble at least. This already gives you much more safety.

Let's imagine you have a `Stack`. If you don't make the `Stack` generic you either:

1. Lose type safety and allow for this to happen

```
my_stack.push('foo')
my_stack.push(3)
```

Which will may blow up at runtime if the purpose of this stack instance was to hold only strings.

or

2. You have to create a `StringStack`, `ByteStack`, `IntStack` etc to get type safety at the cost of code duplication. This is essentially what you are doing when you say you'd would like to fix `MemoryDB` to `[byte, byte]`.

But with generics you can make a `Stack<T>` (sorry, using TypeScript syntax for my personal convenience :sweat_smile: :)

and then you can have many different stacks that are - **at development time** - constraint to be type checked.

So the following would be disallowed by mypy/tsc.

```
my_stack = new Stack<string>();
my_stack.push('foo')
// blows up at *development time*
my_stack.push(3)
```

Still you can use your `Stack<T>` to make another *instance* of the stack that is strongly typed to only serve ints.

```
my_stack = new Stack<int>();
my_stack.push(3)
```

That's the beauty of generics which are possible to enforce at *development time* with Python 3.6 type hints.

gsalgado:
FWIW, #409 changes our mypy invocation so that it checks untyped defs as well. You probably want to rebase your work on top of that

trinity/utils/db.py
pipermerriam: We've chosen to lean towards the *strict* side of things.  Our database objects like the one in question will always be `bytes -> bytes`.

pipermerriam:
#409 was just merged so this can now simply be rebased from `master`

trinity/utils/chains.py
```diff
@@ -100,21 +100,20 @@ def load_nodekey(nodekey_path):
...
+    _network_id:int = None
 
     def __init__(self,
-                 chain_identifier,
-                 data_dir=None,
-                 nodekey_path=None,
-                 nodekey=None,
-                 network_id=None):
+                 chain_identifier:str,
+                 data_dir:str=None,
```
pipermerriam: for consistent formatting, the convention setup thus far in the `p2p` module is for there to be a space after the `:` here.

```
def fn(self,
       chain_identifier: str,
       ...
```

cburgdorf: No problem. Is there a linter I can run locally?

trinity/utils/db.py
cburgdorf: Ok, no problem. But do we agree that I should simply nuke the `MemoryDB` in a separate commit since it isn't used at all across the code base.

cburgdorf:
>#409 was just merged so this can now simply be rebased from `master`

Will do. I'll batch some more changes together before I clean things up with the intention to get properly reviewed and merged if that sounds ok.

trinity/utils/db.py
cburgdorf: Ok, I noticed that even though `MemoryDB` isn't used in trinity yet, there's at least a test that uses it so I brought it back and added the `[bytes, bytes]` type hints you suggested.

cburgdorf:
Ok, so I learned that with Python 3.5 the only way to have type hints for variables are these ugly comment annotations.

```
self._chain = []  # type: List['CancelToken']
```

This may be a dumb question but being new to Python I just have to ask those: Do we need to stay Python 3.5 compatible?

And if the answer is yes, do you want me to add these (ugly! *cough*) comment annotations or rather leave out the type annotations for variables / instance members?

cburgdorf:
I went ahead and dropped the type annotations from all variables (including instance members) where they could be dropped. There was one case where I had to add one, so I went for the comment type annotation to stay Python 3.5 compatible.

https://github.com/cburgdorf/py-evm/blob/8d565f43b3522b4b543ec5f8c9930d9ebda899f5/trinity/utils/db.py#L4

Still, I think if there's no strong reason to keep the Python 3.5 support I'd rather drop it and went for proper type annotations.

cburgdorf:
@pipermerriam This went green now.

Notice that this only adds type annotations for Trinity/utils. If it's cool with you, we could take this PR and I send follow up PRs with batched type annotations. Not too many but batching makes it possible to have things more digestible on my end since I'm

a.) juggling between different projects

b.) still working my way through Python.

I used the following strategy for now:

- add type annotations for all function params and return values
- don't add type annotations for variables / member variables
- if scenarios where type annotations for variables could not be omitted I used the comment annotation to stay compatible with python 3.5

pipermerriam:
> Do we need to stay Python 3.5 compatible?

Yes, py35 compat is required

> And if the answer is yes, do you want me to add these (ugly! cough) comment annotations

Yes, that is the appropriate style to use for them.  I would *prefer* we annotate variables but I don't have a direct reason why other than that it seems like it *should* help detect certain classes of issues.

> If it's cool with you, we could take this PR and I send follow up PRs with batched type annotations.

This approach seems good.  I like many smaller PRs rather than one monstrous one.

cburgdorf:
>Yes, py35 compat is required

Understand. Do you mind to elaborate on the reasoning? Not trying to argue against it. I just want to learn.

>I would prefer we annotate variables

Here's how one would usually approach this in TypeScript. It's quite a fair comparison because TypeScript also adds type annotations to a dynamic language (JS) and has similar features as `mypy` such as type inference.

I wouldn't add type annotations in cases like this:

```
foo = 'str'
```
Due to the type inference, `mypy` is smart enough to disallow this.

```
foo = 'str'
foo = 5
```
Raises an error as expected
```
$ mypy test.py 
test.py:16: error: Incompatible types in assignment (expression has type "int", variable has type "str")
```
So, adding type annotations doesn't add any safety in this case. I'd consider it rather noisy tbh.

Especially when you keep in mind that we have to use Py35 compatible comment annotations. Comment type annotations are rather fragile compared to proper ones.

Consider this:

```
foo = 4 # ttype: str
```

`mypy` checks this without errors even though we obviously intended this to be of type `str`. It also doesn't catch that we mistyped anything. It will simply see this as an `int`.

Let's take a look at another case and imagine we have a class with an instance variable that is initially set to `None`. This case is more tricky as `None` doesn't let `mypy` infer the type.

However, `mypy` can still infer the type based on the context.

```
class Foo ():
  expected_to_be_int = None

  def blows_up(self) -> float:
    if self.expected_to_be_int is not None:
      return self.expected_to_be_int / 2
    
    return 0

val = Foo()
val.expected_to_be_int = 'bar'
```

This blows up as expected

```
$ mypy test.py 
test.py:11: error: Incompatible types in assignment (expression has type "str", variable has type "None")
```

Also `mypy` will tell us when it needs our help because it can't infer the type.

```
val = []
```

```
$ mypy test.py 
test.py:1: error: Need type annotation for variable
```

With all that in mind, I recommend to only add the comment type annotations when `mypy` explicitly asks for them. 

If you agree with that, this PR could probably get merged. If you insist on adding comment type annotations for all variables I'll update the PR. Your call :)

pipermerriam:
> py35 compat is required: Do you mind to elaborate on the reasoning?

Not at all.  And now that you've said it, I realize that py35 support is only required for py-evm.  Trinity can be py36 only and I think I'm good with that.  (ping @carver @gsalgado )

Reason that py-evm has to support py35 is that Web3.py supports python3.5 and it's `EthTester` provider is based on the `eth-tester` library, which has a `py-evm` backend, and thus, py-evm requires support for py35.

Also, we have sort of been taking our cues **loosely** from Django in terms of dropping support for older versions of python.

> With all that in mind, I recommend to only add the comment type annotations when mypy explicitly asks for them.

I'm good with that.  I'm kind of new to type annotations so I do appreciate you laying that out for me as I wasn't aware of any of that TBH.   ping @gsalgado to get a chance to veto this as he knows more about this domain than I currently do.

carver:
> And now that you've said it, I realize that py35 support is only required for py-evm. Trinity can be py36 only and I think I'm good with that.

My other reference point is: what is the python installation in the most recent Ubuntu LTS distro. I expect the new one in April will be py36, since the interim releases are already on py36. So I'm good with 36 for Trinity. :+1:

gsalgado:
I completely agree there's no need to add type annotations for variables when mypy can infer the type -- in the same way we don't add type annotations for the `self` argument of methods. That's what I've been doing for the `p2p` package

cburgdorf:
Sounds good! Then I suggest we merge this and I'll send a follow up PR to drop Python 3.5 support for trinity and use proper type annotations (as in, not the comment annotations) for variables except for the cases where its obvious for `mypy` to infer the type.

cburgdorf:
@pipermerriam 

>Reason that py-evm has to support py35 is that Web3.py supports python3.5 and it's EthTester provider is based on the eth-tester library, which has a py-evm backend, and thus, py-evm requires support for py35.

Makes sense. Thanks your for explaining it to me.

>I'm kind of new to type annotations so I do appreciate you laying that out for me as I wasn't aware of any of that TBH

I'm super happy I can be helpful here. There's tons of stuff I'm lacking to know when it comes to Python.

cburgdorf:
Noticed something strange. It seems the only reason why this PR went green on Travis is because the `flake8` job did not check trinity.

```
flake8 runtests: commands[0] | flake8 /home/travis/build/ethereum/py-evm/evm
flake8 runtests: commands[1] | flake8 /home/travis/build/ethereum/py-evm/tests --exclude=
```

https://travis-ci.org/ethereum/py-evm/jobs/347174579

I'm investigating this as I'm working on #423. So, maybe we hold back merging until I'll figured out what's wrong here.

cburgdorf:
Ah, ok. This PR isn't rebased on top of latest master and flake8 checks for Trinity where just recently enabled: https://github.com/ethereum/py-evm/commit/64d1662d5b2e48ae0dc853e5ea70fc9b394bfd7f

Will rebase and fix.

cburgdorf:
Ok, I rebased this on latest master and the problem seems solved. That said, the build is still failing but I think that's a general problem of master being broken right now as I see this (merged) commit failed because of the same reason.

https://travis-ci.org/ethereum/py-evm/jobs/347464971

pipermerriam:
#421 has now landed if you want to rebase.

cburgdorf:
@pipermerriam rebased and green :)

PR #409, opened 2018-02-26
Make mypy check bodies of untyped functions as well

gsalgado:
By default it only checks the code of a func/method if it is typed, so
I changed the command line we use in TOX to ask it to check untyped defs
as well.

Also added an extra switch to disallow incomplete type annotations (i.e.
require func/methods to annotate all args and return value).

Those new checks caught a few bugs, that I've fixed as well


cburgdorf:
Great to see these rules could be tweaked!

gsalgado:
Yeah, although it doesn't seem to work very reliably: https://github.com/python/mypy/issues/4637 :-(

Issue #398, opened 2018-02-23
trinity crashes when attempting to run the database server


gsalgado:
I get this traceback when I try to run trinity:

```
Traceback (most recent call last):
  File "/home/salgado/virtualenvs/pyevm/bin/trinity", line 11, in <module>
    load_entry_point('py-evm', 'console_scripts', 'trinity')()
  File "/home/salgado/src/py-evm/trinity/main.py", line 191, in main
    database_server_process.start()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 105, in start
    self._popen = self._Popen(self)
  File "/usr/lib/python3.6/multiprocessing/context.py", line 284, in _Popen
    return Popen(process_obj)
  File "/usr/lib/python3.6/multiprocessing/popen_spawn_posix.py", line 32, in __init__
    super().__init__(process_obj)
  File "/usr/lib/python3.6/multiprocessing/popen_fork.py", line 20, in __init__
    self._launch(process_obj)
  File "/usr/lib/python3.6/multiprocessing/popen_spawn_posix.py", line 47, in _launch
    reduction.dump(process_obj, fp)
  File "/usr/lib/python3.6/multiprocessing/reduction.py", line 60, in dump
    ForkingPickler(file, protocol).dump(obj)
_pickle.PicklingError: Can't pickle <class 'abc.NativeECCBackendPrivateKey'>: attribute lookup NativeECCBackendPrivateKey on abc failed
```

pipermerriam:
That makes sense.  Because of the dynamic class generation the `PrivateKey` objects returned by `eth-keys` aren't picklable.  This really needs to be fixed on the eth-keys side as the dynamic class generation is proving problematic.

pipermerriam:
Odd that I haven't seen this error, anything non-standard going on? code changes?  hard coded private keys?

pipermerriam:
link https://github.com/ethereum/eth-keys/issues/26

pipermerriam:
proposed fix: https://github.com/ethereum/eth-keys/pull/36

gsalgado:
No code changes, but I was passing a `nodekey` arg to trinity instead of using the nodekey-file

pipermerriam:
that's something I think we should support, passing a *raw* nodekey via something like `--nodekey 0xabcd1234...`.

gsalgado:
Right, and that works with your changes to eth-keys -- that's why I closed the issue

PR #415, opened 2018-02-27
Bump eth-keys version

pipermerriam:
### What was wrong?

upstream `eth-keys` used dynamically generated classes.  This was problematic because they couldn't be pickled.

### How was it fixed?

Upstream fix in the `eth-keys` library to make objects pickleable.

#### Cute Animal Picture

maxresdefault


pipermerriam:
link #398

Issue #357, opened 2018-02-07
Implement `GUESS_HEAD`


mhchia:
### What is wrong?
As [sharding spec](https://github.com/ethereum/sharding/blob/develop/docs/doc.md#create_collation) pointed, we need `GUESS_HEAD` to  "pick the longest valid chain, check validity as far as possible, and if you find it's invalid then switch to the next-highest-scoring valid chain you know about", once we find ourselves the eligible proposer.

### How can it be fixed
- Implement it with the mocked `fetch_and_verify_collation`.
- Start with one process(At first we don't consider the case with multiprocess)
    - Use `timer` or `blockFilter` to see when to return from the function

#### TODO list (To be completed)
- [ ] Might need to solve https://github.com/ethereum/py-evm/issues/344

PR #416, opened 2018-02-28
Guess head and remove xfails in tests

mhchia:
### What was wrong?
1. `GUESS_HEAD` in [sharding doc](https://github.com/ethereum/sharding/blob/develop/docs/doc.md#create_collation)
https://github.com/ethereum/py-evm/issues/357

2. Remove xfails
Need to wait until https://github.com/ethereum/eth-tester/pull/67 is merged

#### TODO

- [x] GuessHeadStateManager, with (GuessHeadIncrement?)
- [x] add tests for several case
- [x] choose between snapshot or reset
    - reset costs a lot
- [x] refactor `test_vmc_handler`, currently there is too many DRYs

### How was it fixed?
- Implement the algorithm with `fetch_candidate_head` (now we use another more stateless `candidate_heads` to replace it)
- Reset `ShardTracker` in the begin of `GUESS_HEAD`, to ensure the result is correct

#### Cute Animal Picture


carver:
@mhchia https://github.com/ethereum/eth-tester/pull/67 is merged, so you should be unblocked there. Let me know if you run into anything else.

mhchia:
@carver Thank you for the reminder.
Would you mind bump a version of eth-tester when it is ready?
No rush at all since I can already test my code with the latest eth-tester.
Thank you :)

carver:
Ah, we need to apply #421 to the `sharding` branch (plus there was a unique dependency on the old `force_text` in sharding that had to be fixed).  See: #426

mhchia:
Note: need to make sure the behavior of `GUESS_HEAD` is correct or not.
Currently the tests for `GUESS_HEAD` regulate the behavior, but maybe need modifying if there is anything wrong

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -33,14 +37,37 @@
 )
 
 
+GENESIS_COLLATION_HASH = b'\x00' * 32
```
pipermerriam: Convention has been that this would live in `evm.vm.forks.sharding.constants`

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -212,9 +305,10 @@ def send_transaction(self,
             gas_price=gas_price,
             data=data,
         )
-        build_transaction_instance = self.buildTransaction(build_transaction_detail)
-        func_instance = getattr(build_transaction_instance, func_name)
-        unsigned_transaction = func_instance(*args)
+        func_instance = getattr(self.functions, func_name)
```
pipermerriam: @carver we should make it possible to do `self.functions[func_name]` in web3.py contracts.

tests/sharding/test_vmc_handler.py
```diff
@@ -135,26 +147,131 @@ def deploy_initiating_contracts(vmc_handler, privkey):
...
         'deploy_initiating_contracts: vmc_tx_hash=%s',
         w3.eth.getTransactionReceipt(encode_hex(txs[-1].hash)),
     )
 
 
-def import_key(vmc_handler, privkey):
+def deploy_vmc(vmc_handler):
+    # test the deployment of vmc ######################################
+    # deploy vmc if it is not deployed yet.
+    if not is_vmc_deployed(vmc_handler):
```
pipermerriam: I'm seeing this pattern here and in `import_key` and I believe we can replace it with a `pytest.fixture`

It looks like `tests.sharding.fixtures.vmc` should be renamed to `vmc_handler`

Then define:

```python
@pytest.fixture
def primary_key(vmc_handler):
    # the body of the current `import_key` without the `try/except` logic.

@pytest.fixture
def vmc(vmc_handler, primary_key):
    # the body of `deploy_vmc` without the `if` block.
```

Since we know pytest will only run these *once* for each tests we should get strong guarantees that they won't cause errors.  The one exception *may* be the importing of the key as I'm not 100% sure what the logic behind the scenes is for how imported accounts are managed and if they are reset on each test run (I think they should be).

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -166,6 +198,67 @@ def fetch_candidate_head(self, shard_id):
         shard_tracker = self.get_shard_tracker(shard_id)
         return shard_tracker.fetch_candidate_head()
 
+    def memoized_fetch_and_verify_collation(self, collation_hash):
```
pipermerriam: I'm not sure the `memoize_` prefix is needed here.  The fact that it's memoized is an implementation detail.

tests/sharding/test_vmc_handler.py
```diff
@@ -142,7 +142,7 @@ def deploy_initiating_contracts(vmc_handler, privkey):
     for tx in txs:
         tx_hash = send_raw_transaction(vmc_handler, tx)
         mine(vmc_handler, 1)
-        print("!@# tx_hash={}".format(w3.eth.getTransactionReceipt(tx_hash)))
+        assert w3.eth.getTransactionReceipt(tx_hash) != None
```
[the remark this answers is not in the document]
mhchia: Thank you for pointing this out. Actually I have changed this in the previous commits. It is more clear now after your explanation.

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -240,10 +250,14 @@ def guess_head(self, shard_id):
 
         :return: returns the hash of the guessed head collation
         """
+        # TODO: this should be change due to the performance issue
+        self.reset_shard_tracker(shard_id)
```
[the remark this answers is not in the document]
mhchia: Now I have removed this line.
Currently `guess_head` is changed like this https://github.com/ethereum/py-evm/pull/416/files#diff-dd73c470e87a612ab8844d87c3a5d172R229
At the end of the function I still need to clean up the logs in `shard_tracker`. Since it seems(still need confirming) [`guess_head`](https://github.com/ethereum/sharding/blob/develop/docs/doc.md#create_collation) should work only only with new arriving logs, since the last time `guess_head` is called.
In other words, after calling `guess_head`, new logs fetched from the network by `get_next_log` remaining in the shard_tracker should be cleaned, to avoid they being used in the future `guess_head`

evm/vm/forks/sharding/vmc_handler.py
mhchia: Will move it. Thank you so much:)

evm/vm/forks/sharding/vmc_handler.py
mhchia: You're right.  Thank you for pointing this out.
Is it okay that I rename this method to the same name as the function `fetch_and_verify_collation` outside?
Or do you have any suggestion on this? thank you:)

mhchia:
Note: Removed the "Need review" tag, since I'm making big change of the `GUESS_HEAD`. And will have a lot to refactor and refine.

mhchia:
@pipermerriam Do you have any suggestion if I need the program to asynchronously download collations from other peers and keep running other logics?
Is it fine to use threads here? or should I use other ways like `async/await`?
Thank you so much:)

tests/sharding/test_vmc_handler.py
mhchia: Will do it. Thank you!:)

pipermerriam:
> Is it fine to use threads here? or should I use other ways like async/await?

I would strongly prefer we use an `asyncio` based approach.  I think that a thread based approach may bite us down the road as writing threadsafe code is much harder than writing `asyncio` code (which doesn't require thread safety).

mhchia:
No problem. I will check it out.
Thank you:)

evm/vm/forks/sharding/vmc_handler.py
carver: :+1: tracking in https://github.com/ethereum/web3.py/issues/705

mhchia:
Note: still resolving conflict, will update the rebased code and documentation tomorrow. sorry for the late updates.

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        # ignore tasks that are done
+        return [
+            self.collation_verifying_task[collation_hash]
+            for collation_hash in self.chain_collations[head_collation_hash]
+        ]
+
+    async def guess_head(self, enable_timing_up=False):
+        '''
+        Perform windback process.
+        returns None if there is no candidate head available in this period
```
hwwhww: Returns the head collation hash if it meets the windback condition. Returns None If there's no qualified candidate head.

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        while True:
+            # discard old logs if we're in the new period
+            if self.get_current_period() > self.last_period_fetching_candidate_head:
+                self.shard_tracker.clean_logs()
+            head_collation_hash = self.fetch_candidate_head_hash()
+            if head_collation_hash is None:
+                break
+            current_collation_hash = head_collation_hash
+            current_depth = 0
+            while self.chain_validity[head_collation_hash]:
```
hwwhww: Feeling a little nervous to see three `while` loops in this function.  And the code snippet:
```python
while self.chain_validity[head_collation_hash]:
    # if running out of time
    if enable_timing_up and self.is_late_collator_period():
        return head_collation_hash
```
is duplicate.

mhchia: Yes you're right. Thank you for pointing out.

Actually I have 2 alternatives for now.
1. Use `for i in range(self.vmc.config['WINDBACK_LENGTH']):` and move the most inner while to the same level as `for`
2. Use two layer `while`, however, in the inner while loop, we don't do `current_collation_hash = self.vmc.get_collation_parent_hash(...)` every time. Instead, need to check if the `current_collation_hash == GENESIS_COLLATION_HASH`, if so, check if the collations of the chain `head_collation` is totally finished.

I prefer the first one since the logic is more clear and readable, I think. And I have a version of this  already.

Do you have any suggestion about which one to choose? or other solutions? Thank you.

evm/vm/forks/sharding/shard_tracker.py
```diff
@@ -0,0 +1,141 @@
...
+        #       to make one log_handler shared over all shards.
+        self.shard_id = shard_id
+        self.log_handler = log_handler
+        self.vmc_address = vmc_address
+        self.current_score = None
+        self.new_logs = []
+        self.unchecked_logs = []
+
+    @to_tuple
+    def peek_new_logs(self):
```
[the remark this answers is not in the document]
mhchia: No, thank you for pointing out. I will remove it.

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+    return await verify_collation(collation, collation_hash)
+
+
+def clean_done_task(tasks):
+    return [task for task in tasks if not task.done()]
+
+
+class GuessHeadStateManager:
+
+    TIME_ONE_STEP = 0.1
```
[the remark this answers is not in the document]
mhchia: it refers to the time yield to unfinished tasks for now.

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+    chain_collations = None
+
+    def __init__(self, vmc, shard_id, shard_tracker, my_address):
+        # a well-setup vmc handler with a shard_tracker in shard `shard_id`
+        self.vmc = vmc
+        self.shard_id = shard_id
+        self.shard_tracker = shard_tracker
+        self.my_address = my_address
+
+        self.last_period_fetching_candidate_head = 0
```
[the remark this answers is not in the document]
mhchia: This one is better. Will change it. Thank you!

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+    chain_validity = None
+
+    unfinished_verifying_tasks = None
+    collation_verifying_task = None
+    chain_collations = None
+
+    def __init__(self, vmc, shard_id, shard_tracker, my_address):
+        # a well-setup vmc handler with a shard_tracker in shard `shard_id`
+        self.vmc = vmc
+        self.shard_id = shard_id
```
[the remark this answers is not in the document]
mhchia: Yes it can. I will change it. Thank you:)

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        current_block_num = self.vmc.web3.eth.blockNumber
+        last_block_num_in_current_period = current_period * self.vmc.config['PERIOD_LENGTH']
+        return (
+            self.is_collator_in_period(current_period) and
+            (current_block_num + NUM_RESERVED_BLOCKS >= last_block_num_in_current_period)
+        )
+
+    # guess_head process related
+
+    async def verify_collation(self, head_collation_hash, collation_hash):
```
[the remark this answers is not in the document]
mhchia: It sounds good to me. Should I then move the content from `fetch_and_verify_collation` to `verify_collation`, and remove `fetch_and_verify_collation`? 
Another question i'm concerning about is, if I use `coro` as the prefix, should I further add the `coro` to all `async` functions?

evm/vm/forks/sharding/guess_head_state_manager.py
mhchia: Good suggestion! Should I add a "time's up" condition for this additionally?

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+            current_collation_hash = head_collation_hash
+            current_depth = 0
+            while self.chain_validity[head_collation_hash]:
+                # if running out of time
+                if enable_timing_up and self.is_late_collator_period():
+                    return head_collation_hash
+
+                # when `WINDBACK_LENGTH` or GENESIS_COLLATION is reached
+                # TODO: assume the whole chain for now
+                if ((current_depth == self.vmc.config['WINDBACK_LENGTH']) or
```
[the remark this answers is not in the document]
mhchia: Thank you for the suggestion. Will change it!

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+                    self.shard_id,
+                    current_collation_hash,
+                )
+                current_depth += 1
+
+                # clean up the finished tasks, yield CPU to tasks
+                if len(self.unfinished_verifying_tasks) != 0:
+                    await asyncio.wait(self.unfinished_verifying_tasks, timeout=self.TIME_ONE_STEP)
+                    not_finished_tasks = clean_done_task(self.unfinished_verifying_tasks)
+                    self.unfinished_verifying_tasks = not_finished_tasks
```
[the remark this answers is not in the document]
mhchia: You're right. I missed it. Thank you

evm/vm/forks/sharding/guess_head_state_manager.py
mhchia: The code snippet becomes https://gist.github.com/mhchia/8da88064a3efd1014c7d334c8045e370

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    collation = await download_collation(collation_hash)
+    return await verify_collation(collation, collation_hash)
+
+
+def clean_done_task(tasks):
+    return [task for task in tasks if not task.done()]
+
+
+class GuessHeadStateManager:
```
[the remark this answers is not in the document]
mhchia: > IMO:
This class name is too long.
It's unclear what's the "state" for.
ghs_manager looks like a ghost manager.

I agree. Will change them.

> It's inefficient if you are setting each collator has to maintain SHARD_COUNT GuessHeadStateManagers. Does this process be initialized for each known period that the collator applied lookahead and knows it's the eligible collator?

This class is only initialized once, and able to be reused in the future `guess_head`s. My thought about the inefficiency issue is: we need one `log_handler` for each shard, because `get_new_logs` is called for each shard at different moment. And every `shard_tracker` needs one `log_handler`. Even though we move `guess_head` to `shard_tracker` and collator related logics to class `Collator`, we still need `SHARD_COUNT` `shard_tracker`s for one collator. Do you have any solution in mind already? You're right It might be able to be further optimized.

> If yes, it's like a worker to do its job - windback? If so, how about WindbackWorker...?

I'm fine with this name. It's better than the original one.

evm/vm/forks/sharding/guess_head_state_manager.py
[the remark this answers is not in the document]
mhchia: It means the yielded time from the `guess_head` loop to the `unfinished_verifying_tasks` currently.
Or should I rename it to `yielded_time` or something else?

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        return self.vmc.web3.eth.blockNumber // self.vmc.config['PERIOD_LENGTH']
+
+    def is_collator_in_period(self, period):
+        collator_address = self.vmc.get_eligible_proposer(
+            self.shard_id,
+            period,
+        )
+        return collator_address == self.my_address
+
+    def is_collator_in_lookahead_periods(self):
```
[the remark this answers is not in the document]
mhchia: You're right. I will remove the lookahead related logics outside this class.
Actually I need an indicator of "time's up" in this class.
I'm thinking to add a bool member variable in this class, indicating if currently we are running out of time, and check this variable in every loop. If it is set to `True`, which means running out of time, then just break the loop and return the result of `guess_head`. This is a relatively easy way to do "time's up" in `guess_head`, without introducing collator related logics. And is safe because we use async tasks instead of threads.

evm/vm/forks/sharding/guess_head_state_manager.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        See if we are going to be a collator in the future periods
+        """
+        result = False
+        for future_periods in range(self.vmc.config['LOOKAHEAD_PERIODS']):
+            lookahead_period = self.get_current_period() + future_periods
+            result |= self.is_collator_in_period(lookahead_period)
+        return result
+
+    def is_late_collator_period(self):
```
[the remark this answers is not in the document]
mhchia: My thought of this method is "check if it is close to the end of the period when we are collator", which may need to assume we are in the last blocks in that period.

evm/vm/forks/sharding/guess_head_state_manager.py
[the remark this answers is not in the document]
mhchia: I changed the name of `GuessHeadStateManager.verify_collation` to `process_collation`, moved the content inside `fetch_and_verify_collation` to it, and removed `fetch_and_verify_collation`. Does it look better now?

mhchia:
@hwwhww I just updated the code, corresponding your suggestions:) Thank you!
About the **CollatorProcess**,
> Apply Lookahead. If it found it’s an eligible collator of period period, initialize a WindbackWorker(vmc, shard_tracker, period, address) to find the head_collation. Check if it’s now the period that being the eligible collator. If yes:

The current design is, we initialize `WindbackWorker` only once, when the class `Collator`(not implemented yet, still need designing) is initialized. And `WindbackWorker.guess_head`(or any other more suitable name) is called every period when the node finds itself a eligible collator in the lookahead periods.


> Option 1, active ProposalPool: Apply ProposalPool.collect_proposals to select the proposals.
Option 2, passive ProposalPool: A ProposalPool process like the TxPool. Initialize a ProposalPool during the windback, only retain the proposals with the top 3 highest bids.

Sorry I didn't get why we need two kinds of pools. Why not just have a `ProposalPool` which keeps collecting all proposal, which we can filter out the proposal in the current period, and remove the proposals which are outdated?

jannikluhn:
> The main loop can keep running CheckBlockNumber coroutine (polling or registering event) 

Minor comment: In the POC I found it very helpful having the SMC handler do the polling. It can provide `asyncio.Event`s for new blocks, new periods, and new headers. The advantage is that the code becomes easier to read and the timing assumptions are made more clear. There can also helper coroutines like `SMCHandler.wait_for_period(period)` or maybe even `SMCHandler.wait_for_time_in_period(period, time)` to wait for, say, 20s into some period.

evm/vm/forks/sharding/log_handler.py
```diff
@@ -117,3 +123,7 @@ def get_new_logs(self, address=None, topics=None):
             topics,
         )
         return self.w3.eth.getLogs(filter_params)
+
+    def reset(self):
```
pipermerriam: This function is at least not used at all in the changelog of this pull request.  

Also, I find it odd that it resets to genesis.  Intuitively I would expect it to reset to it's original state (as in the state it was in when instantiated, or maybe just the most recent blockhashes).  Maybe rename to `reset_to_genesis`?

evm/vm/forks/sharding/shard_tracker.py
```diff
@@ -0,0 +1,157 @@
...
+    shard_id_bytes32 = log['topics'][1]
+    data_hex = log['data']
+    data_bytes = decode_hex(data_hex)
+    score = big_endian_to_int(data_bytes[-32:])
+    is_new_head = bool(big_endian_to_int(data_bytes[-64:-32]))
+    header_bytes = shard_id_bytes32 + data_bytes[:-64]
+    collation_header = CollationHeader.from_bytes(header_bytes)
+    yield 'header', collation_header
+    yield 'is_new_head', is_new_head
+    yield 'score', score
```
pipermerriam: Since this function has no conditional keys, I think the use of `to_dict` reduces readability.  Maybe switch to just building the dictionary locally and returing it.

mhchia:
@jannikluhn @hwwhww @NIC619 @pipermerriam 
In the latest 2 commits, I change to the @jannikluhn version of `fetch_candidate_head`. I think it is simpler and cleaner now. It will be helpful if there is any suggestion.
For now I have several things need discussing regarding this approach:
* If we want to introduce the sliding window for `headers_by_height`, we still need to figure out the size of sliding window of `headers_by_height`. Since we need to consider not only the case of the chain head with the max height, but also other candidate heads as many as possible, to guarantee we possess `WINDBACK_LENGTH + 1` collations of those candidate heads.
* I use `copy.deepcopy` in `fetch_candidate_heads_generator`, to ensure the generator is not affected by the concurrently executed `add_log`, which might be introduced later.

evm/vm/forks/sharding/log_handler.py
mhchia: So sorry for this. You're right it is indeed not used anymore. I think I can just remove it.
Thank you so much for pointing it out.

evm/vm/forks/sharding/shard_tracker.py
mhchia: No problem. I have changed it. Thank you

evm/vm/forks/sharding/guess_head_state_manager.py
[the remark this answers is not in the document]
hwwhww: If `LOOKAHEAD_LENGTH := 4`, let's say the collator disconnect after one more period, the maximum concurrently existing `log_handler` is 5, I think it's tolerable if the separated-period-workers could simplify many sharing memory problems.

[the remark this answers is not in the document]
mhchia: Yes you're right. If you're concern is `SHARD_COUNT` `log_handler`s, actually we only store the recent `256` blockhashes. So roughly there will be 256*32 bytes*100 = 819200 bytes <= 1 MB memory consumption, if a collator synchronizes all shards.
IMO it seems for now it is not a problem? I can also mark a TODO to indicate this issue so that we can do further optimization later?

evm/vm/forks/sharding/shard_tracker.py
```diff
@@ -0,0 +1,161 @@
...
+            self.headers_by_height.append([])
+        self.headers_by_height[collation_height].append(log_entry['header'])
+
+    def fetch_candidate_heads_generator(self, windback_length):
+        new_logs = self._get_new_logs()
+        for log_entry in new_logs:
+            # TODO: we can use another coroutine to sync the logs concurrently, instead of doing
+            #       that in this function
+            self.add_log(log_entry)
+        # TODO: deepcopy seems to be costly
```
jannikluhn: Considering this, it might be a bad idea to make `candidate_heads` a generator (whose state might change in between). As an alternative, we could make it a simple function and add a predicate argument. `candidate_head` would return the first head fulfilling the predicate.

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns
+            the head collation hash, if it meets the windback condition. OR
+            None, if there's no qualified candidate head.
+        '''
+        start = time.time()
+
+        head_collation_hash = current_collation_hash = None
+        self.unfinished_verifying_tasks = []
+        candidate_heads = self.shard_tracker.fetch_candidate_heads_generator(
+            windback_length=self.vmc.config['WINDBACK_LENGTH']
```
jannikluhn: As mentioned above, I don't think `windback_length` is needed here.

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+        while True:
+            try:
+                head_header = next(candidate_heads)
+            except StopIteration:
+                return None
+            head_collation_hash = head_header.hash
+            if not self.chain_validity[head_collation_hash]:
+                continue
+            current_collation_hash = head_collation_hash
+            for _ in range(self.vmc.config['WINDBACK_LENGTH'] + 1):
```
jannikluhn: For the POC I implemented the iterators `iterate_ancestors` and `iterate_descendents` in the shard class. I think using something like this would make the code a little clearer, but I'm not sure at what place this could be implemented right now.

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+        self.chain_collations[head_collation_hash].append(current_collation_hash)
+
+    def get_chain_tasks(self, head_collation_hash):
+        # ignore tasks that are done
+        return [
+            self.collation_verifying_task[collation_hash]
+            for collation_hash in self.chain_collations[head_collation_hash]
+        ]
+
+    async def guess_head(self):
```
jannikluhn: I would separate `guess_head` from the windback. Windback could run indefinitely until it's cancelled, and `guess_head` could be called at any time (in the worst case it would return the genesis).

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                if self.is_time_up:
+                    return head_collation_hash
+                if current_collation_hash == GENESIS_COLLATION_HASH:
+                    break
+
+                # if a collation is an invalid head, set its descendants in the
+                # chain as invalid heads, and skip this chain
+                is_chain_of_current_collation_invalid = (
+                    (not self.chain_validity[current_collation_hash]) or
+                    (not self.collation_validity_cache.get(current_collation_hash, True))
```
jannikluhn: I don't really get why there's `chain_validity` and `collation_validity_cache`. Aren't they (or shouldn't they be) the same thing?

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                )
+                if is_chain_of_current_collation_invalid:
+                    self.chain_validity[current_collation_hash] = False
+                    for collation_hash in self.chain_collations[head_collation_hash]:
+                        self.chain_validity[collation_hash] = False
+                    break
+                # process current collation  ##################################
+                # if the collation is checked before, then skip it
+                if current_collation_hash not in self.collation_validity_cache:
+                    descendants = copy.deepcopy(
```
jannikluhn: this looks wrong. Why not set the descendents validity in here after `process_collation`?

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                        self.chain_collations[head_collation_hash]
+                    )
+                    coro = self.process_collation(
+                        head_collation_hash,
+                        current_collation_hash,
+                        descendants,
+                    )
+                    task = asyncio.ensure_future(coro)
+                    asyncio.wait(task, timeout=self.YIELDED_TIME)
```
jannikluhn: I'm still not very comfortable with asyncio, but I don't think `self.YIELDED_TIME` (and therefore `asyncio.wait`) is necessary here. `guess_head` should yield anytime a collation is actually downloaded. So I think `guess_head` should just wait for downloads to complete, not some arbitrary timeout.

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+                current_collation_hash = self.vmc.get_collation_parent_hash(
+                    self.get_shard_id(),
+                    current_collation_hash,
+                )
+
+            # when `WINDBACK_LENGTH` or GENESIS_COLLATION is reached
+            while self.chain_validity[head_collation_hash]:
+                # if time is up
+                if self.is_time_up:
+                    return head_collation_hash
```
jannikluhn: Tasks which are still running here should probably be canceled somehow.

pipermerriam: look into the `CancelToken` in the `p2p` module (we can extract it to it's own library if it's needed in another codebase.

evm/vm/forks/sharding/shard_tracker.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def clean_logs(self):
+        self.new_logs = []
+        self.unchecked_logs = []
+
+    def add_log(self, log_entry):
+        collation_height = log_entry['score']
+        while len(self.headers_by_height) <= collation_height:
+            self.headers_by_height.append([])
+        self.headers_by_height[collation_height].append(log_entry['header'])
```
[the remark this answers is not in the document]
mhchia: Sorry I didn't get your meaning. In `add_log`, a log is add to the end of the list of its height. So within the list of a specific height in `headers_by_height`, later logs are in the behind of the list. And every time `candidate_heads` is called, all logs happened since last time `candidate_heads` is called are handled by `add_logs`, in the order from older to newer, ordering by period, ascending? or am I understanding wrong?

evm/vm/forks/sharding/shard_tracker.py
```diff
@@ -0,0 +1,161 @@
...
+    new_logs = None
+    unchecked_logs = None
+    headers_by_height = None
+
+    def __init__(self, shard_id, log_handler, vmc_address):
+        # TODO: currently set one log_handler for each shard. Should see if there is a better way
+        #       to make one log_handler shared over all shards.
+        self.shard_id = shard_id
+        self.log_handler = log_handler
+        self.vmc_address = vmc_address
```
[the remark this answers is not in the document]
mhchia: Have changed it. Thank you:)

evm/vm/forks/sharding/shard_tracker.py
```diff
@@ -0,0 +1,161 @@
...
+        self.current_score = None
```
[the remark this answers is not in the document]
mhchia: yes you're right. I have removed it since it is used in the old `fetch_candidate_head`

evm/vm/forks/sharding/windback_worker.py
mhchia: Sorry I didn't get what you mean.
`descendants` is used by `process_collation` to propagate the invalidity when a collation is verified as invalid

evm/vm/forks/sharding/windback_worker.py
mhchia: You're right `asyncio.wait(task, timeout=self.YIELDED_TIME)` is not necessary here.  I'm removing it.
Thank you

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,210 @@
...
+                        head_collation_hash,
+                        current_collation_hash,
+                        descendants,
+                    )
+                    task = asyncio.ensure_future(coro)
+                    asyncio.wait(task, timeout=self.YIELDED_TIME)
+                    self.set_collation_task(head_collation_hash, current_collation_hash, task)
+
+                # clean up the finished tasks, yield CPU to tasks
+                if len(self.unfinished_verifying_tasks) != 0:
```
mhchia: @jannikluhn 
Should I just remove this `if`? Since anyway we have a `while self.chain_validity[head_collation_hash]:` after this for loop, to finish these jobs. Thank you!

evm/vm/forks/sharding/windback_worker.py
mhchia: @jannikluhn Originally I thought, since this instance will be reused when the next `guess_head` is called, we can finished that task at that time?
@pipermerriam Thank you for the direction. I will look into it.

mhchia:
Note: I think I just leave this PR not rebased to the latest commits, since that commits removed all the sharding code.

evm/vm/forks/sharding/windback_worker.py
mhchia: Thank you for the suggestion. I will see how I can apply them here

evm/vm/forks/sharding/windback_worker.py
```diff
@@ -0,0 +1,212 @@
...
+    # data = get_from_p2p_network(collation_hash.data_root)
+    # chunks = DATA_SIZE // 32
+    # mtree = [0] * chunks + [data[chunks*32: chunks*32+32] for i in range(chunks)]
+    # for i in range(chunks-1, 0, -1):
+    #     mtree[i] = sha3(mtree[i*2] + mtree[i*2+1])
+    # assert mtree[i] == collation.data_root
+    return True
+
+
+def clean_done_task(tasks):
```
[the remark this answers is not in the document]
mhchia: Have changed it. Thank you:)

mhchia:
Close by https://github.com/ethereum/py-evm/issues/539
This `windback` code will be kept in https://github.com/mhchia/py-evm/tree/guess_head

PR #417, opened 2018-02-28
p2p: Implement first cut of ChainSyncer

gsalgado:
For now it only fetches headers, but has hooks in place for triggering the download of block bodies and receipts as well.

Plenty of TODOs and FIXMEs as well, so not sure if we want to merge this into master or in a separate branch

### DIFF
--- a/p2p/cancel_token.py
+++ b/p2p/cancel_token.py
@@ -70,28 +70,30 @@ async def _wait_for_first(futures):
         return
 
 
-async def wait_with_token(future: Awaitable,
-                          cancel_token: CancelToken,
+async def wait_with_token(*futures: Awaitable,
+                          token: CancelToken,
                           timeout: float = None) -> Any:
-    """Wait for future to complete, unless we timeout or the cancel token is triggered.
+    """Wait for the first future to complete, unless we timeout or the cancel token is triggered.
 
-    Returns the future's result in case it completes.
+    Returns the result of the first future to complete.
 
     Raises TimeoutError if we timeout or OperationCancelled if the cancel token is triggered.
+
+    All pending futures are cancelled before returning.
     """
     done, pending = await asyncio.wait(
-        [future, cancel_token.wait()],
+        futures + (token.wait(),),
         timeout=timeout,
         return_when=asyncio.FIRST_COMPLETED)
     for task in pending:
         task.cancel()
     if not done:
         raise TimeoutError()
-    if cancel_token.triggered_token is not None:
+    if token.triggered_token is not None:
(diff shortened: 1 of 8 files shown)


p2p/chain.py
```diff
@@ -0,0 +1,242 @@
...
+                        "Too many consecutive timeouts waiting for header batch, aborting sync "
+                        "with %s", peer)
+                    break
+                continue
+
+            if peer.is_finished():
+                self.logger.debug("%s disconnected, stopping sync", peer)
+                break
+
+            consecutive_timeouts = 0
```
pipermerriam: I think this timeout logic could be wrapped up in a context manager to both allow it to be tested in isolation, reused, and to reduce some of the cognitive overhead needed to parse this `while` loop.

p2p/chain.py
```diff
@@ -0,0 +1,242 @@
...
+            await asyncio.sleep(0.1)
+        else:
+            self.logger.info("Waited too long for peers to finish, exiting anyway")
+
+    async def stop(self) -> None:
+        self.cancel_token.trigger()
+        self.peer_pool.unsubscribe(self)
+        await self.wait_until_finished()
+
+    def handle_msg(self, peer: ETHPeer, cmd: protocol.Command,
```
pipermerriam: I just learned about [`functools.single_dispatch`](https://docs.python.org/3/library/functools.html#functools.singledispatch) which would in *theory* allow you to split the handlers here up into many different single functions.

I'm not sure this is the right use case, and I suspect the fact it's a class method might make it a slightly more complex use case.

pipermerriam: Or maybe this library if you want multiple dispatch.  http://multiple-dispatch.readthedocs.io/en/latest/index.html though I'm not keen on adding the dependency

p2p/cancel_token.py
```diff
@@ -70,17 +70,19 @@ def __str__(self):
         return
 
 
-async def wait_with_token(future: Awaitable,
+async def wait_with_token(futures: List[Awaitable],
```
pipermerriam: Looking at the way this is used, I'm wondering if the UX for using this would be better if `futures` was done as `*futures`.

```
# current
wait_with_token([the_future], cancel_token)

# positional
wait_with_token(the_future, cancel_token=cancel_token)

# positional and change `cancel_token` to `token`
wait_with_token(the_future, token=cancel_token)
```

It's not shorter, but it does avoid the awkward need to wrap the first argument in an iterable.

gsalgado: Yeah, that's a nice improvement

PR #418, opened 2018-02-28
Extend paygas JSON tests

jannikluhn:
- [x] Test the value PAYGAS pushes on to the stack (denoting success or failure)
- [x] Test that the REVERT opcode does not revert gas payment

#### Cute Animal Picture

### DIFF
--- a/tests/fillers/vm_fillers/paygas_fillers.py
+++ b/tests/fillers/vm_fillers/paygas_fillers.py
@@ -1,10 +1,9 @@
-from collections import namedtuple
-
 from cytoolz import (
     pipe,
 )
 
 from evm.tools.test_builder.test_builder import (
+    Test,
     setup_sharding_filler,
     pre_state,
     expect,
@@ -15,13 +14,7 @@ from evm.tools.test_builder.builder_utils import (
 )
 
 
-Test = namedtuple("Test", ["filler", "fill_kwargs"])
-Test.__new__.__defaults__ = (None,)  # make `None` default for fill_kwargs
-
-
 address = generate_random_address()
-coinbase = generate_random_address()
-caller = generate_random_address()
 normal_contract_address = b"`\xa8\xdc~\xba\xd7\x03\x8c\x9a\x83}\x84\xd8\x13\xa8R\x92\xe5d\xbd"
 paygas_contract_address = b"[p\xb7s\x88\xb4e\xf7R@w\xd2e\xfa\x1c{j\x9acy"
 assert address != normal_contract_address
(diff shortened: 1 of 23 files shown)


jannikluhn:
I'm reasonably confident that the tests are correct, but it would be nice nevertheless to have another pair of eyes to look over them. Preferably, @NIC619 could review as he's written the implementation of the opcode.

Looking at the JSON files directly probably doesn't make a whole lot of sense, but https://github.com/jannikluhn/py-evm/blob/0812496bab24c91a9b9b1e12bdb814a080958809/tests/fillers/vm_fillers/paygas_fillers.py is supposed to be somewhat readable. To check correctness, one should look at the Vyper LLL code, the pre state and the post state changes. Usually the balance of the `address` account is reduced by a little as it had to pay for gas. The return code of paygas is usually stored in the storage, which is then checked with `paygas_successful` or `paygas_failed` in the post state. Also, the tests store something to check if the state is reverted (if it isn't supposed to be reverted, `storage_updated` is put into the post state).

pipermerriam:
I re-directed this PR to be towards `master` but it needs to be rebased now.

tests/fillers/vm_fillers/paygas_fillers.py
```diff
@@ -32,15 +25,18 @@
     1: 1,
 })
 
-broke_caller = (caller, "balance", 0)
-solvent_caller = (caller, "balance", 1000000)
+paygas_successful = (address, "storage", 2, 2)  # return code + 1
+paygas_failed = (address, "storage", 2, 1)  # return code + 1
```
NIC619: In `paygas_in_call_test` test case, there doesn't seem to be an `ADD` operation.

tests/fillers/vm_fillers/paygas_fillers.py
```diff
@@ -327,11 +337,41 @@
...
+    expect(
+        networks=["Sharding"],
+        transaction={
+            "to": address,
+            "accessList": [
+                [address, b""],
+            ],
+        },
+        post_state=[
+            (address, "balance", 18951),
```
NIC619: Should the balance in post state be `pre_state_balance - transaction_gas_limit * gas_price`? As all gas will be confiscated.

tests/fillers/vm_fillers/paygas_fillers.py
jannikluhn: It is, but it's hidden in the `paygas_contract`. See https://github.com/jannikluhn/py-evm/blob/0812496bab24c91a9b9b1e12bdb814a080958809/tests/fillers/vm_fillers/paygas_fillers.py#L37

tests/fillers/vm_fillers/paygas_fillers.py
jannikluhn: No, here state is reverted with the REVERT opcode which doesn't consume all gas. See https://github.com/ethereum/EIPs/blob/master/EIPS/eip-140.md

tests/fillers/vm_fillers/paygas_fillers.py
NIC619: But aren't both `paygas_successful` and `paygas_failed` the cases for `address`(which is generated by `generate_random_address()`)?
`paygas_successful ` is the case for `address` but shouldn't `paygas_failed` be the case for `paygas_contract_address`?

tests/fillers/vm_fillers/paygas_fillers.py
NIC619: Yep you are right. I mistake the `assert` here for the one in Solidity(`asset` there will take away all the gas).

tests/fillers/vm_fillers/paygas_fillers.py
jannikluhn: Ah, now I get what you mean. This is what's going on: The paygas contract returns the return code (+ 1), but the "main" contract (at `address`) stores it in its storage. So to check for success/failure, one has to check the storage at `address`, even if the actual call was done at `paygas_contract_address`. See https://github.com/jannikluhn/py-evm/blob/0812496bab24c91a9b9b1e12bdb814a080958809/tests/fillers/vm_fillers/paygas_fillers.py#L274

NIC619: Ah I see. I didn't see that in `paygas_contract`, result of `["ADD", 1, ["PAYGAS", 1]]` is store in memory and then returned.

PR #421, opened 2018-03-01
Eth utils v1 stable compatible

carver:
### What was wrong?

Some more methods were removed on the way to eth-utils v1 stable. These are the last ones till v2! (which should be quite a while)

### How was it fixed?

Removed all dependencies on `force_bytes`

#### Cute Animal Picture


p2p/kademlia.py
```diff
@@ -559,7 +559,7 @@ def refresh_idle_buckets(self):
             asyncio.ensure_future(self.lookup(rid))
 
     def _mkpingid(self, token: AnyStr, node: Node) -> bytes:
-        pid = force_bytes(token) + node.pubkey.to_bytes()
+        pid = to_bytes(text=token) + node.pubkey.to_bytes()
```
pipermerriam: This seems potentially broken since `token` can be either `bytes` or `str` (see [`typing.AnyStr`](https://docs.python.org/3/library/typing.html#typing.AnyStr)).

tests/p2p/test_discovery.py
```diff
@@ -196,6 +196,6 @@ def random_address():
 
 
 def random_node():
-    seed = force_bytes("".join(random.sample(string.ascii_lowercase, 10)))
+    seed = to_bytes(text="".join(random.sample(string.ascii_lowercase, 10)))
```
pipermerriam: this seems overly complicated.  Maybe change this to just `seed = os.urandom(10)` to seed it with 10 random bytes?

p2p/kademlia.py
carver: Ah thanks for pointing that out, I didn't look up the def'n and it surprised me.

I'll add a test to catch it, and fix it.

tests/p2p/test_discovery.py
carver: Do we know that the seed can contain any possible bytes? Maybe there is a reason we need or want it to be valid ascii lowercase letter bytes only.

pipermerriam: Gonna leave this as-is for now and merge.

Issue #370, opened 2018-02-13
Transaction origin and transaction salt in Account Abstraction


NIC619:
### What is wrong?
In account abstraction, now origin of the transaction(`transaction.origin`) is always the sender - `ENTRY_POINT`(`0xffff…ffff`) and it's meaningless in a contract's context. `transaction.origin` should probably be changed from `transaction.sender` to `transaction.target` or if origin of the transaction no longer holds significant meaning in account abstraction(e.g, it's just a contract that verifies signature, pays for gas and then forwards to another contract. A real account of a user may be composed of many contracts.) it probably can be removed.

Also I'm working on removing gas price from transaction format. As discussed in #326 , one option is to add `salt` to transaction format(perhaps adding `salt` for short term and transitioning to EIP232 transaction format for long term?). 

### How can it be fixed

- transaction origin
    - 1.1 change transaction origin from `transaction.sender` to `transaction.target`
    - 1.2 remove transaction origin
    - 1.3 ?
- salt
    - 2.1 add salt to transaction format
    - 2.2 EIP232 transaction format
    - 2.3 2.1 for short term and 2.2 for long term
    - 2.4 ?

lrettig:
Linking #262 and #366 -- this will impact those.

Issue #326, opened 2018-01-29
Contract creation transactions don't specify a salt


jannikluhn:
### What is wrong?

Contract creation transactions don't specify a salt (see https://github.com/ethereum/py-evm/blob/sharding/evm/vm/forks/sharding/vm_state.py#L60-L63). As a result, the address of the created contract is given by the bytecode only and thus it's impossible to deploy the same contract twice.

### How can it be fixed

Specify the salt somehow. Potential candidates could be

- the first/last 32 bytes of `transaction.code`
- the first/last 32 bytes of `transaction.data`
- `transaction.to` (which is meaningless for contract creation transactions right now)

NIC619:
One can append random bytes(which won't affect contract initialization process) to the bytecode as salt.

NIC619:
@pipermerriam @hwwhww @jannikluhn 
It seems cleaner if we just add `salt` to transaction format, as suggested by @vbuterin and to conform with suggested transaction format of main chain in this [EIP](https://github.com/ethereum/EIPs/issues/859).

NIC619:
I will make the PR that together removes `gas_price` and add `salt` to Sharding transaction format once #285 (PAYGAS) is merged.

jannikluhn:
What's wrong with reusing `transaction.to`/`target` for this? The salt actually defines the address (at least in part), so `target` isn't too forced. I think this would be better than adding another field which is empty for most transactions.

NIC619:
Since `transaction.to` is where the contract will be deployed(EVM will check if `transaction.to == generate_CREATE2_contract_address(*salt*, transaction.code)` ), if we reuse `transaction.to` as salt it will result in cyclic dependency, i.e, `transaction.to` is used as salt for calculating address but also has to be the calculated address.

If I understand correct, `transaction.target` is not part of the formal transaction format. Hence it seems to me that it's not a reliable source as a salt.

jannikluhn:
Yes, that check would have to be removed. But it's not necessary anyway, right now `transaction.to` is redundant information.

NIC619:
If I understand correctly, `transaction.to` is still needed especially if it's not a contract deploying transaction?

jannikluhn:
Yeah, sorry, `transaction.to` is only redundant for contract deployments. For normal transactions it's still required of course. But for those we don't need a salt.

hwwhww:
[...]

PR #422, opened 2018-03-01
Remove gas price from and add salt to Sharding transaction

NIC619:
### How was it fixed?
- [x] Abstract `gas_price` away. Gas price is now set via `PAYGAS` opcode.
- [x] Add `salt`. Address of contract deployed by transaction is now generated with `generate_CREATE2_contract_address(salt, code)` instead of `generate_CREATE2_contract_address(b'', code+salt)`
- [x] change `transaction_context .origin` from `ENTRY_POINT` to `transaction.to`

#### Cute Animal Picture


evm/auxiliary/user_account_contract/transaction.py
```diff
@@ -163,7 +165,7 @@ def as_signed_transaction(self, private_key):
             nonce=self.nonce,
             min_block=self.min_block,
             max_block=self.max_block,
-            gas_price=self.gas_price,
+            gas_price=self.int_gas_price,
```
pipermerriam: What does the `int_` prefix stand for?  integer?  internal?

evm/logic/context.py
```diff
@@ -106,6 +106,10 @@ def gasprice(computation):
     computation.stack.push(computation.transaction_context.gas_price)
 
 
+def sharding_gasprice(computation):
```
pipermerriam: What do you think about calling this `PAYGAS_gasprice` so that it's clealy linked to the `PAYGAS` opcode.  The `sharding_` prefix doesn't feel like the correct descriptor.

evm/auxiliary/user_account_contract/transaction.py
NIC619: I think it stands for internal before `gas_price` is removed from transaction format. Will update it to `gas_price`.

hwwhww:
Related issue: #326, #370

NIC619:
Re-direct to master branch and change `transaction.origin` in Sharding from `ENTRY_POINT` to `transaction.to`.
Will merge this today if the `transaction.origin` change is a go.

PR #423, opened 2018-03-01
Drop Python 3.5 support for Trinity

cburgdorf:
### What was wrong?

As a follow up to the discussion we had in #407, this PR aims to drop Python 3.5 support for Trinity at the benefit of being able to use stronger type hint support.

### How was it fixed?

1. Removed the test cases that built trinity with Python 3.5 from `travis.yaml`
2. Changed `envlist` in `tox.ini` (@pipermerriam unsure about that as you didn't explicitly mention it and I'm new to tox)
3. Splitted `flake8` job into `flake8py35` and `flake8py36` to exclude trinity + tests from being flake checked under Python 3.5 conditions
3. Added a change in trinity that would break under Python 3.5 just to confirm I haven't missed anything.

Notice that this PR is built on top of #407

#### Cute Animal Picture


cburgdorf:
One job timed out. Seems unrelated to the change https://travis-ci.org/ethereum/py-evm/jobs/347937335

PR #424, opened 2018-03-01
evm/db: Drop _to_db suffix of ChainDB's persist* methods

gsalgado:
Also refactor ChainDB.persist_block() to create a new method that
persist only uncles -- needed by ChainSyncer


pipermerriam:
ping @carver this is going to cause breakage in `eth-tester`.  We should be sure the `py-evm` backend is pinned to a specific version before making a new `py-evm` release with these changes.

carver:
> We should be sure the py-evm backend is pinned to a specific version before making a new py-evm release with these changes.

@pipermerriam argh, it used to be, for just that reason, but it got undone. Thanks for the heads-up.

pipermerriam:
- Option 1 - Start to establish *stable* APIs in py-evm, and accept the overhead of a deprecation process.
- Option 2 - Status-Quo, pin to exact alpha versions and deal with the issues as they arise.

I am in favor of 2 because we're still figuring things out in certain places in py-evm and these changes should become less and less frequent.

carver:
Yes, totally agree on option 2. Depending on an alpha means accepting API risk, which also means the safest thing to do is pin to a specific version :+1:

PR #425, opened 2018-03-01
p2p: ChainSyncer now fetches block bodies

gsalgado:
Depends on #424

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -1,10 +1,15 @@
 import asyncio
 import logging
 import operator
+import random
 import time
 from typing import Any, cast, Dict, List, Set, Tuple  # noqa: F401
 
+from cytoolz import partition_all
+
+from evm.constants import EMPTY_UNCLE_HASH
 from evm.db.chain import AsyncChainDB
+from evm.db.trie import make_trie_root_and_nodes
 from evm.rlp.headers import BlockHeader
 from p2p import protocol
 from p2p import eth
@@ -17,6 +22,7 @@ class ChainSyncer(PeerPoolSubscriber):
     logger = logging.getLogger("p2p.chain.ChainSyncer")
     # We'll only sync if we are connected to at least min_peers_to_sync.
     min_peers_to_sync = 2
+    _reply_timeout = 5
 
     def __init__(self, chaindb: AsyncChainDB, peer_pool: PeerPool) -> None:
         self.chaindb = chaindb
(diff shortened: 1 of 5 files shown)


p2p/chain.py
```diff
@@ -1,11 +1,14 @@
...
 import logging
 import operator
+import random
 import time
 from typing import Any, cast, Dict, List, Set, Tuple  # noqa: F401
 
+from evm.constants import EMPTY_UNCLE_HASH
 from evm.db.chain import AsyncChainDB
 from evm.rlp.headers import BlockHeader
+from evm.utils.state import make_trie_root_and_nodes
```
pipermerriam: I want to avoid cross-library imports from `utils`.  Possible solutions.

- copy this to a local `p2p.utils`
- create an alternate location for utils that are considered public API (this seems like a good idea).

p2p/chain.py
```diff
@@ -137,17 +146,66 @@ def register_peer(self, peer: BasePeer) -> None:
...
+                    header for header in headers
+                    if (header.transaction_root != self.chaindb.empty_root_hash or
+                        header.uncles_hash != EMPTY_UNCLE_HASH)
+                ]
+                self.request_bodies(headers_with_non_empty_bodies)
+
+            self.retry_timedout_bodies()
+
+    def request_bodies(self, headers: List[BlockHeader]) -> None:
+        while headers:
```
pipermerriam: This body can be cleaned up a bit by using [`cytoolz.partition_all`](http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition_all)

```python
for batch in partition_all(headers, eth.MAX_BODIES_FETCH):
    ...
```

Nice side effect of removing the mutation of the `headers` variable.

p2p/chain.py
```diff
@@ -137,17 +146,66 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            for header in batch:
+                key = (header.transaction_root, header.uncles_hash)
+                self._pending_bodies[key] = (header, now)
+
+    def retry_timedout_bodies(self) -> None:
+        timed_out = []  # type: List[BlockHeader]
+        now = time.time()
+        for header, req_time in self._pending_bodies.values():
+            if now - req_time > self._reply_timeout:
+                timed_out.append(header)
```
pipermerriam: Maybe better and even more readable as a list comprehension?

```python
timed_out = [
    header
    for header, req_time 
    in self._pending_bodies.values()
    if now - req_time > self._reply_timeout 
]
```

p2p/chain.py
```diff
@@ -137,17 +146,66 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            except OperationCancelled:
+                return
+
+            if headers:
+                headers_with_non_empty_bodies = [
+                    header for header in headers
+                    if (header.transaction_root != self.chaindb.empty_root_hash or
+                        header.uncles_hash != EMPTY_UNCLE_HASH)
+                ]
+                self.request_bodies(headers_with_non_empty_bodies)
```
pipermerriam: Is there a reason why this is synchronous?  It looks like to sends a network request on line 191.

p2p/chain.py
gsalgado: Whenever I do something like this I remember to look at itertools to see if there's something I could use, but I never remember to look at cytoolz as well. Thanks for the reminder!

p2p/chain.py
gsalgado: None of the p2p methods that send data are coroutines as they use [StreamReader.write()](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.WriteTransport.write), which is a regular function but nonblocking. IOW, it is not synchronous

p2p/chain.py
gsalgado: Yeah, makes sense. I've moved it to `evm.db.trie`, unless you have another place in mind?

p2p/chain.py
pipermerriam: FYI, this is the documentation page you should bookmark: http://toolz.readthedocs.io/en/latest/api.html

PR #426, opened 2018-03-01
[sharding] Compatibility with eth-utils v1 stable

carver:
### What was wrong?

eth-utils v1 stable dropped `force_*` and `coerce_*` methods.

### How was it fixed?

Cherry-picked from #421 merge, plus one more `force_text()` in build_utils, unique to sharding

* remove force_bytes from eth-utils v0

* Use tox for linting, to get all folders

* Upgrade to eth-utils v1.0.1

* test str & bytes response in pong + related bugfix

#### Cute Animal Picture


carver:
Yay, strict `xfails` reminded me to explicitly upgrade eth-tester to beta21. It's now pinned to the version, and should stay that way until it goes stable.

PR #431, opened 2018-03-02
Allow deep property configuration in Configurable classes

pipermerriam:
### What was wrong?

Currently, if you wanted to use the `MainnetChain` with a different set of opcodes, you would be required to duplicate a *lot* of code to properly reconfigure all the way down to the correct deeply burried objects to swap out the opcodes.

### How was it fixed?

This pull requests gets us a step closer to being able to do that easily, by allowing configuration of sub-properties.

```python
class Inner(Configurable):
    attr_a = 'value-a'

class Outer(Configurable):
    attr_b = Inner


NewClass = Outer.configure(**{'attr_b.attr_a': 'new-value-a'})
assert NewClaass.attr_b.attr_a == 'new-value-a'
```

This allows us to *reach* into sub-properties that are also of the `Configurable` type and do sub-configuration. 

#### Cute Animal Picture

flying_cat-630x438

### DIFF
--- a/docs/building_chains.rst
+++ b/docs/building_chains.rst
@@ -23,7 +23,7 @@ class:
   import evm.chains.mainnet as mainnet
 
   chain_class = Chain.configure(
-      name='Test Chain',
+      __name__='Test Chain',
       vm_configuration=(
           (constants.GENESIS_BLOCK_NUMBER, FrontierVM),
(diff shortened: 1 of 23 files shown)


pipermerriam:
cc @NIC619 as I think this is the *start* of implementing the API that would be needed for miners to do the `PAYGAS` pre-check on a computation.

PR #432, opened 2018-03-02
Fix gas estimation when a transaction is in the block

carver:
~Note, please wait for #397 to get merged, before merging this. It's a big one, and I don't want to be responsible for delaying it. :)~ _397 merged, and conflicts resolved via rebase_

## What was wrong?

`chain.estimate_gas(txn)` fails with a `ValidationError`, if there is a transaction already in the block.

`vm.state_in_temp_block()` is designed to make sure that a fresh block is available for estimating gas. However, the receipts created in `vm.get_state()` were always the receipts of the canonical head, rather than the receipts from the `header` passed to `get_state`. So some transactions were added to the temporary block. (and probably some other bad things were happening... it's a little surprising that no tests got triggered)

### How was it fixed?

- Fixed `vm.get_state()` to create the receipt list based on the supplied `header`.
- Added a test to catch the issue, plus an assertion about the state of a temporary block.
- `make lint` fixup, after the latest `flake8` split into trinity (py36) and everything else (py35)

#### Cute Animal Picture

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -210,8 +210,7 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         temp_block = self.generate_block_from_parent_header_and_coinbase(header, header.coinbase)
         prev_hashes = (header.hash, ) + self.previous_hashes
         state = self.get_state(self.chaindb, temp_block, prev_hashes)
-        if state.gas_used > 0:
-            raise Exception("There must not be any gas used in a fresh temporary block")
+        assert state.gas_used == 0, "There must not be any gas used in a fresh temporary block"
 
         snapshot = state.snapshot()
         yield state
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -209,7 +209,7 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         header = self.block.header
         temp_block = self.generate_block_from_parent_header_and_coinbase(header, header.coinbase)
         prev_hashes = (header.hash, ) + self.previous_hashes
-        state = self.get_state(block_header=temp_block.header, prev_hashes=prev_hashes)
+        state = self.get_state(self.chaindb, temp_block, prev_hashes)
         if state.gas_used > 0:
             raise Exception("There must not be any gas used in a fresh temporary block")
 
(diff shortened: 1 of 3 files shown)


evm/vm/base.py
```diff
@@ -210,6 +210,9 @@ def state_in_temp_block(self):
         temp_block = self.generate_block_from_parent_header_and_coinbase(header, header.coinbase)
         prev_hashes = (header.hash, ) + self.previous_hashes
         state = self.get_state(block_header=temp_block.header, prev_hashes=prev_hashes)
+        if state.gas_used > 0:
+            raise AssertionError("There must not be any gas used in a fresh temporary block")
```
pipermerriam: 👎 assertions in runtime code.  If this is supposed to be an unreachable code path you can just use `Exception`.

carver:
@pipermerriam this probably needs another look -- specifically that I changed `vm.get_state()` to accept a block instead of a header.

Without this change, several sharding tests were failing because they weren't able to look up the block from the header.

evm/vm/base.py
```diff
@@ -209,7 +209,10 @@ def state_in_temp_block(self):
         header = self.block.header
         temp_block = self.generate_block_from_parent_header_and_coinbase(header, header.coinbase)
         prev_hashes = (header.hash, ) + self.previous_hashes
-        state = self.get_state(block_header=temp_block.header, prev_hashes=prev_hashes)
+        state = self.get_state(self.chaindb, temp_block, prev_hashes)
+        if state.gas_used > 0:
+            raise Exception("There must not be any gas used in a fresh temporary block")
```
pipermerriam: Maybe `ValidationError`?

tests/core/helpers.py
```diff
@@ -62,3 +65,20 @@ def new_sharding_transaction(
...
+    amount = 100
+
+    assert chain.get_vm().state.gas_used == 0
+
+    while True:
+        vm = chain.get_vm()
+        tx = new_transaction(vm, from_, recipient, amount, key, gas=gas, data=data)
+        try:
+            chain.apply_transaction(tx)
+        except ValidationError:
```
pipermerriam: This bugs me because there are so many things that can throw validation errors....  Even though it's gross, since this is test code, what do you think about one extra step here checking something about the exception message to make sure it's the error you think it is....

evm/vm/base.py
pipermerriam: I've been thinking about adding an `Invariant` exception for these types of things where what we are really doing is assert.

PR #434, opened 2018-03-03
Fix misplaced access list and state db

NIC619:
### How was it fixed?
Remove `Message.access_list` and replace `Computation.state_db` with `Computation.vm_state.state_db`

#### Cute Animal Picture


hwwhww:
Just for helping me understand this PR, am I right that there exist three expressions:

1. **Type-original**: `with computation.vm_state.state_db(read_only=True) as state_db` means  `state_db = computation.vm_state._chaindb.get_state_db(state_root, read_only=True)`
2. **Type-access-list**: `with computation.state_db(read_only=True) as state_db` means `with computation.vm_state._chaindb.get_state_db(state_root, read_only=True, access_listcomputation.msg.access_list) as state_db`
3. **Type-property**: `state_db = computation.vm_state.read_only_state_db` means `state_db = computation.vm_state._chaindb.get_state_dbstate_root, read_only=True)`   # equals to **Type-original** with argument.

I recalled that you said you also would override `@contextmanager def state_db(...)` in `sharding/vm_state.py`, and it seems you changed your mind and decided to modify `evm/vm/forks/sharding/computation.py`?

If the two lines in `evm/vm/forks/sharding/computation.py` are the *only* place that we use **Type-access-list** expression, I think we can remove **Type-access-list** with **Type-original** because  **Type-access-list** doesn't explicitly indicate that it's with `access_list`, but that's just my opinion.

NIC619:
Sorry I forgot to update the description after pushing new commits.

After @hwwhww point out, I realize that `Computation.state_db` can not be removed(which is what I did before update - I replaced `Computation.state_db` with `Computation.vm_state.state_db`) because `VMState.state_db` in Sharding requires `Computation.state_db` to pass in `access_list` to it.
```python
## /vm/forks/sharding/computation.py
def state_db(self):
    return self.vm_state.state_db(self.msg.access_list)
```
So the branch name is kind of misleading as no access_list and state_db is really misplaced. Sorry about that.


I keep the state_db in `Computation`.(In ShardingVM version it passes an extra `access_list` into `vm_state.state_db`). And then split `state_db` into `read_only_state_db` and `mutable_state_db`

One problem is that `vm_state.read_only_state_db` in ShardingVM is different from the one in other VMs. It's not made a property since it requires passing in an additional `access_list` argument.
```python
## evm/vm_state.py
@property
    def read_only_state_db(self):
        return self._chaindb.get_state_db(self.state_root, read_only=True)

## usage
state_db = vm_state. read_only_state_db
```

```python
## /vm/forks/sharding/vm_state.py
def read_only_state_db(self, access_list=None):
        return self._chaindb.get_state_db(
            self.state_root,
            read_only=True,
            access_list=access_list,
        )

## usage
state_db = vm_state. read_only_state_db()
```

@pipermerriam @hwwhww Kindly asking for a review again. Sorry for the trouble!

NIC619:
Makes sense.

For `VMState.get_state_db`, I will also have to update eth-tester as it currently uses `vm_state.state_db`.
Or continue using `VMState.state_db` instead?

pipermerriam:
> Or continue using VMState.state_db instead?

No, we should change the API and then change `eth-tester` to use the new API.  I *think* we pinned `eth-tester` to an exact `py-evm` version so this shouldn't break things.

NIC619:
@pipermerriam sorry I might have missed something, but I still can't figure out how to get a `read_only_state_db` and a context managed `mutable_state_db` with only context managed `vm_state.get_state_db`


this is `vm_state.get_state_db(…)`
```python
    @contextmanager
    def get_state_db(self, read_only):
        state = self._chaindb.get_state_db(
            self.state_root,
            read_only,
        )
        yield state
        ...
```

this is the `computation.mutable_state_db(…)` which is context manageable
```python
    @property
    def mutable_state_db(self):
        return self.vm_state.get_state_db(read_only=False)
```

but for `computation.read_only_state_db(…)`
this will return a `_GeneratorContextManager` object just like mutable_state_db
```python
    @property
    def read_only_state_db(self):
        return self.vm_state._chaindb.get_state_db(
            self.vm_state.state_root,
            read_only=True,
        )
```

if we use `with` statement, it will return the real state_db object, however it will be cleaned up by the time the state db is used.
```python
    @property
    def read_only_state_db(self):
        with self.vm_state._chaindb.get_state_db(
            self.vm_state.state_root,
            read_only=True,
        )  as state_db:
            return state_db
```

pipermerriam:
@NIC619 I think you may be right that there may not be a good way to do this.  Can you explore what the minimal API we can get away with is?  Something that is ideally consistent with naming conventions and how it's used?  The `Computation` object could simply implement pass-through methods which delegate down to the `VMState` AIPs.

I have a feeling that a lot of the database access patterns are going to be refactored soon as I'm starting to see how they can be reworked in some ways to make all of this cleaner and less entangled.

NIC619:
@pipermerriam After re-evaluation, it seems to me that splitting `state_db` into `read_only_state_db` and `mutable_state_db` might not worth the effort.

To recap: the goal is to turn the read only `state_db` into a property so it does not need to be context managed. We do it by splitting `state_db` into `read_only_state_db`(a property) and `mutable_state_db`(context managed).

But this results in bloated `*_state_db` APIs.

(1) `Computation.read_only_state_db` and `Computation.mutable_state_db` are needed because `access_list` is added in Sharding and it is stored in `Computation.msg`. So unless we allow `VMState` to access `access_list` directly, we need `Computation.*_state_db` as an additional wrapper to pass in `access_list`.

(2) `VMState.get_state_db` is context managed and so `Computation.read_only_state_db` would also be context managed which fails our goal. Hence we still need two `state_db` APIs in `VMState` - `VMState.read_only_state_db`(a property) and `VMState.mutable_state_db`(context managed).

This will inevitably result in 8 `*_state_db` APIs in `evm/computation.py` , `evm/vm_state.py`, `evm/vm/forks/sharding/computation.py` and `evm/vm/forks/sharding/vm_state.py`.

Considering these trade-offs IMO it's not worth the effort to split the `state_db`.

pipermerriam:
That's fine.  I think we're going to be doing a significant refactor to most of this stuff in the near term future.  I'm fine leaving as-is, or doing whatever the minimal thing that's needed is.

Issue #435, opened 2018-03-03
Create Instructions on how to setup a Trinity node for development


cburgdorf:
### What is wrong?

The Trinity [Readme](https://github.com/ethereum/py-evm/blob/master/trinity/README.md) only lists a couple of environment variables but has no further information on how to setup and run a Trinity node.

This should be fixed to make it easier to onboard new contributors.

Here's what I tried to run the simple JSON-RPC server.

```
$ python3 trinity/rpc/server.py -db foo
Traceback (most recent call last):
  File "trinity/rpc/server.py", line 114, in <module>
    privkey=ecies.generate_privkey(),
  File "/home/cburgdorf/Documents/hacking/py-evm/evm/chains/base.py", line 292, in configure
    return super().configure(name, **overrides)
  File "/home/cburgdorf/Documents/hacking/py-evm/evm/utils/datatypes.py", line 18, in configure
    "not found on the base class `{2}`".format(cls.__name__, key, cls)
TypeError: The LightChain.configure cannot set attributes that are not already present on the base class. The attribute `privkey` was not found on the base class `<class 'p2p.lightchain.LightChain'>`
```

It's most likely related to the fact that I have not setup the environment variables. I actually have no idea how I would create a key to set as `TRINITY_NODEKEY` for instance.

The tinkerer in me tried to run `TRINITY_NODEKEY='foo' && python3 trinity/rpc/server.py -db foo` but that gave me the same error.

Long story short, the Readme should be newbie friendly and needs clear steps to get the node running.

### How can it be fixed

If someone can give me a couple of pointers I can happily compile that into a Readme update

cburgdorf:
Ok, I just noticed that there are actually docs in the repo and they can be found online at http://py-evm.readthedocs.io.

That said, even the [Quickstart](http://py-evm.readthedocs.io/en/latest/quickstart.html) doesn't get me unstuck.

Basically leaves me at the same place as before:

```
$ python3 -i -m trinity.rpc.server -db /tmp/testnet.db
Traceback (most recent call last):
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/cburgdorf/Documents/hacking/py-evm/trinity/rpc/server.py", line 114, in <module>
    privkey=ecies.generate_privkey(),
[...]

PR #437, opened 2018-03-04
WIP: Fixes broken trinity/rpc/server

cburgdorf:
### What was wrong?

Running `python3 trinity/rpc/server.py -db foo` currently leaves you with this:

```
  File "trinity/rpc/server.py", line 114, in <module>
    privkey=ecies.generate_privkey(),
  File "/home/cburgdorf/Documents/hacking/py-evm/evm/chains/base.py", line 292, in configure
    return super().configure(name, **overrides)
  File "/home/cburgdorf/Documents/hacking/py-evm/evm/utils/datatypes.py", line 18, in configure
    "not found on the base class `{2}`".format(cls.__name__, key, cls)
TypeError: The LightChain.configure cannot set attributes that are not already present on the base class. The attribute `privkey` was not found on the base class `<class 'p2p.lightchain.LightChain'>`
```

so, I poked here and there and also created this #435 and eventually created this patch.

### How was it fixed?

1. The chain gets a PeerPool passed as expected

2. The server uses `AsyncChainDB` instead of `ChainDB` because this is one that is intended to have all the coroutine APIs that the server is calling.
    
However, these methods aren't implemented yet so trying to retrieve a block currently leaves you with this.
```
INFO: Running LightChain...
======== Running on http://0.0.0.0:8080 ========
(Press CTRL+C to quit)
ERROR: Error handling request
  File "/home/cburgdorf/.local/lib/python3.6/site-packages/aiohttp/web_protocol.py", line 416, in start
    resp = yield from self._request_handler(request)
  File "/home/cburgdorf/.local/lib/python3.6/site-packages/aiohttp/web.py", line 325, in _handle
    resp = yield from handler(request)
  File "/home/cburgdorf/.local/lib/python3.6/site-packages/aiohttp/web_middlewares.py", line 93, in impl
    return (yield from handler(request))
  File "trinity/rpc/server.py", line 49, in handle
    block = await self.chain.get_canonical_block_by_number(number)
  File "/home/cburgdorf/Documents/hacking/py-evm/p2p/lightchain.py", line 256, in get_canonical_block_by_number
    block_hash = await self.chaindb.coro_lookup_block_hash(block_number)
  File "/home/cburgdorf/Documents/hacking/py-evm/evm/db/chain.py", line 593, in coro_lookup_block_hash
    raise NotImplementedError()
NotImplementedError
INFO: 127.0.0.1 [04/Mar/2018:15:33:25 +0000] "POST / HTTP/1.1" 500 331 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.186 Safari/537.36"
```
So, I noticed that there's a `ChainDBProxy` that is also used in `trinity/main.py` which basically wraps the sync APIs as async APIs. I was wondering if the `rpc/server` should use that proxy as well?

As mentioned in #435 I'm still not sure about the relationship of the different parts in Trinity so any guidance would be welcome :)

### DIFF
--- a/trinity/rpc/server.py
+++ b/trinity/rpc/server.py
@@ -19,6 +19,11 @@ from eth_utils import (
 
 from p2p.lightchain import LightChain
 
+from p2p.peer import (
+    LESPeer,
+    PeerPool,
+)
+
 
 class App(web.Application):
     allowed_methods = ['eth_getBlockByNumber', 'eth_getBlockByHash']
@@ -91,7 +96,7 @@ if __name__ == '__main__':
         MAINNET_GENESIS_HEADER, MAINNET_VM_CONFIGURATION, MAINNET_NETWORK_ID)
     from evm.chains.ropsten import ROPSTEN_GENESIS_HEADER, ROPSTEN_NETWORK_ID
     from evm.db.backends.level import LevelDB
-    from evm.db.chain import ChainDB
+    from evm.db.chain import AsyncChainDB
     from evm.exceptions import CanonicalHeadNotFound
     from p2p import ecies
     logging.basicConfig(level=logging.INFO, format='%(levelname)s: %(message)s')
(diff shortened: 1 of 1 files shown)


cburgdorf:
@pipermerriam  I just noticed that you just [recently moved](https://github.com/ethereum/py-evm/commit/8a2b468922d28d686f2a844afe935788324de636#diff-7e920679e5b9dd02b8ceff3bbba552a9) all the `/rpc/*` stuff `/evm/rpc/` to `trinity/rpc/`. 

With that in mind I guess the goal should probably be that `trinity/main.py` would be the one and only entry point for trinity and it would be responsible to kick off these JSON-RPC APIs (maybe based on a flag if serving those should be optional).

Because currently `trinity/rpc/server.py` has its own [bootstrapping](https://github.com/ethereum/py-evm/blob/3784a87d0850c92d5865d97ce4bb6f5075d47896/trinity/rpc/server.py#L88-L128) to be called directly. I *guess* that is just a leftover that needs to be refactored. Am I reading this right?

pipermerriam:
Correct on both counts.  Primary entry point should be moved to `trinity/main.py` as part of the default `trinity` process.

`trinity/rpc/server.py` will need to be refactored.

cburgdorf:
Alright, thank you. However, I'm still gonna close this PR for now because I gave it a second look and there's too much stuff going on that I'm not certain about how to handle so I rather sit back and wait for someone else to tackle it and learn from studying the changes.

pipermerriam:
This is on my radar.  Will try to get to it soon.

cburgdorf:
Yep, I came across some issues where you were leaving some insights about what is planned.

PR #441, opened 2018-03-06
p2p: ChainSyncer now fetches block receipts

gsalgado:


p2p/chain.py
```diff
@@ -152,61 +156,92 @@ def register_peer(self, peer: BasePeer) -> None:
...
                 return
 
             if headers:
-                headers_with_non_empty_bodies = [
-                    header for header in headers
-                    if (header.transaction_root != self.chaindb.empty_root_hash or
-                        header.uncles_hash != EMPTY_UNCLE_HASH)
-                ]
-                self.request_bodies(headers_with_non_empty_bodies)
+                await request_func(
```
pipermerriam: If this moved into an `else` block from the `try/except` above, you could remove the `headers = []` from inside of the `TimeoutError` block.

p2p/chain.py
```diff
@@ -242,8 +277,12 @@ def retry_timedout_bodies(self) -> None:
                 uncles_hash = await self.chaindb.coro_persist_uncles(body.uncles)
                 self._pending_bodies.pop((tx_root, uncles_hash), None)
         elif isinstance(cmd, eth.Receipts):
-            # TODO: Queue msg for processing by receipt downloader.
-            pass
+            msg = cast(List[List[eth.Receipt]], msg)
+            self.logger.debug("Got Receipts for %d blocks", len(msg))
+            for block_receipts in msg:
+                receipt_root, trie_dict_data = make_trie_root_and_nodes(block_receipts)
```
pipermerriam: Is there validation that needs to be done to verify that the `receipt_root` here matches the value from the `BlockHeader`?

gsalgado: You mean to prevent malicious peers from feeding us fake receipts? AIUI that cannot do any harm unless they also manage to feed us fake headers, right?

pipermerriam: The last line of this function is `self._pending_receipts.pop(receipt_root, None)` and `receipt_root` is computed from the data they sent us, so *if* a malicious node sent us bad receipts we would end up populating our database with extra junk.  The `pop` call would succeed without ever popping anything from the `self._pending_receipts` which *could* in theory allow them to repeatedly do this resulting in a bunch of database bloat?

gsalgado: That's right, but nothing other than DB bloat could be caused this way, and we'd be vulnerable only during the initial fast-sync, so may not be a big deal. Are you ok if I just add a TODO comment for now?

pipermerriam: And maybe an issue to re-visit validation during syncing?

pipermerriam: Eh, or maybe just re-order some of this so that it first checks that the `receipt_root` is in the pending list before persisting it to the database?

pipermerriam: Eh, or maybe just re-order some of this so that it first checks that the `receipt_root` is in the pending list before persisting it to the database?

Like this: https://github.com/ethereum/py-evm/pull/451

PR #442, opened 2018-03-06
p2p: Use cancel token in DiscoveryProtocol/KademliaProtocol

gsalgado:


p2p/discovery.py
```diff
@@ -84,6 +85,12 @@ def __init__(self, privkey: datatypes.PrivateKey, address: kademlia.Address,
         self.bootstrap_nodes = [kademlia.Node.from_uri(node) for node in bootstrap_nodes]
         self.this_node = kademlia.Node(self.pubkey, address)
         self.kademlia = kademlia.KademliaProtocol(self.this_node, wire=self)
+        self.cancel_token = CancelToken('DiscoveryProtocol')
+
+    async def lookup_random(self, cancel_token: CancelToken) -> List[kademlia.Node]:
```
pipermerriam: Am I correct that this is only used in the `_test` script below.  If it's only for testing can we get this out of the base class and into a subclass used only for testing?

p2p/kademlia.py
```diff
@@ -451,9 +454,10 @@ def process(response):
 
         self.neighbours_callbacks[remote] = process
         try:
-            await asyncio.wait_for(event.wait(), k_request_timeout)
+            await wait_with_token(
```
pipermerriam: Since *every* `wait_with_token` has a `CancelToken` as one of it's arguments, it seems like all of these call sites could be cleaned up were you to be able to call this directly from the `token` instance, rather than as a standalone function that has to be imported in every module.

```python
async def wait_neighbours(self, remote, cancel_token):
    ...
    try:
        cancel_token.wait(event.wait(), timeout=k_request_timeout)
    ...
```

This probably involves moving the `CancelToken.wait` to something like `_wait()` unless there is a reason for that API to be public.

p2p/discovery.py
gsalgado: It is also going to be used in production code (coming on my next PR). It will be called periodically to make sure the kademlia routing table is always fresh

p2p/kademlia.py
gsalgado: I thought about that before, but I'm not sure I like the idea of having a method on CancelToken that waits for and returns the first on a list of futures -- it just doesn't feel like the right place to have such an API

PR #443, opened 2018-03-07
extend transaction validation to allow EIP155 txs

dylanjw:
### What was wrong?

Attempting to execute EIP155 type transactions would cause validation errors like the following:

```
evm/vm/base.py:80: in apply_transaction
    self.block,
evm/vm_state.py:215: in apply_transaction
    computation = self.execute_transaction(transaction)
evm/vm/forks/spurious_dragon/vm_state.py:22: in execute_transaction
    computation = _execute_frontier_transaction(self, transaction)
evm/vm/forks/frontier/vm_state.py:54: in _execute_frontier_transaction
    transaction.validate()
evm/vm/forks/homestead/transactions.py:26: in validate
    super(HomesteadTransaction, self).validate()
evm/vm/forks/frontier/transactions.py:52: in validate
    validate_lte(self.v, maximum=28, title="Transaction.v")
evm.exceptions.ValidationError: Transaction.v 38 is not less than or equal to 28
```

### How was it fixed?

Add configuration min/max for v value validation.

#### Cute Animal Picture


evm/vm/forks/spurious_dragon/transactions.py
```diff
@@ -32,6 +32,13 @@ def get_message_for_signing(self):
                 data=self.data,
             ))
 
+    def validate(self):
+        if is_eip_155_signed_transaction(self):
+            chain_id = extract_chain_id(self.v)
+            self.v_min = 35 + (2 * chain_id)
+            self.v_max = 36 + (2 * chain_id)
```
pipermerriam: Instead of assigning these to self, what do you think about making them computed properties that are instead derived on-demand.  That would remove the mutation and side-effects from this `validate` call.

dylanjw: :+1:

PR #445, opened 2018-03-07
p2p: PeerPool now uses DiscoveryProtocol to find nodes

### LINKED ISSUE #430: Do not use hard-coded peers for ChainSyncer/StateDownloader
Those two services use the ETH sub-protocol, which is supported by most nodes in the network (unlike LES), so instead of using a `PeerPool` with hard-coded nodes[1] they should use one that employs `p2p.DiscoveryProtocol` to find and connect to as many peers as possible.

[1] That is a dirty hack we needed because the discovery v4 protocol is terrible at finding LES nodes (#429)

gsalgado:
The old PeerPool with hard-coded nodes was moved to a new class

Closes #430


p2p/discovery.py
```diff
@@ -88,9 +88,10 @@ def __init__(self, privkey: datatypes.PrivateKey, address: kademlia.Address,
         self.cancel_token = CancelToken('DiscoveryProtocol')
 
     async def lookup_random(self, cancel_token: CancelToken) -> List[kademlia.Node]:
-        node_id = random.randint(0, kademlia.k_max_node_id)
-        token_chain = self.cancel_token.chain(cancel_token)
-        return await self.kademlia.lookup(node_id, token_chain)
+        return await self.kademlia.lookup_random(self.cancel_token.chain(cancel_token))
```
pipermerriam: I'm trying to understand where `wait_with_token` needs to be used and doesn't.  I see that this passes in the chained tokens, however, since this is an `await` I would have expected this to need to be:

```python
return await wait_with_token(self.kademlia.lookup_random(self.cancel_token.chain(cancel_token)), token=cancel_token)
# or maybe?
return await wait_with_token(self.kademlia.lookup_random(self.cancel_token.chain(cancel_token)), token=self.cancel_token.chain(cancel_token))
```

This strikes me as something that needs to be written down somewhere in developer documentation (or maybe even just extracted into a stand-alone library and documented there.)

p2p/kademlia.py
```diff
@@ -220,11 +221,29 @@ def __lt__(self, other: 'KBucket') -> bool:
...
+                "Cannot get %d nodes as RoutingTable contains only %d nodes", count, len(self))
+            count = len(self)
+        seen = []  # type: List[Node]
+        # This is a rather inneficient way of randomizing nodes from all buckets, but even if we
+        # iterate over all nodes in the routing table, the time it takes would still be
+        # insignificant compared to the time it takes for the network roundtrips when connecting
+        # to nodes.
+        while len(seen) < count:
+            bucket = random.choice(self.buckets)
+            node = random.choice(bucket.nodes)
```
pipermerriam: I did some experiments and *if* we make `RountingTable` inherit from `collections.abc.Sequence` **and** we add a `__getitem__` implementation (required to inherit from `Sequence`), then we can replace this whole loop with `random.sample(self, count)`

tests/p2p/test_peer.py
```diff
@@ -47,6 +47,9 @@ def test_sub_protocol_selection():
     assert isinstance(proto, LESProtocolV2)
     assert proto.cmd_id_offset == peer.base_protocol.cmd_length
 
+    proto = peer.select_sub_protocol([('unknown', 1)])
```
pipermerriam: This looks like another API that I would be more comfortable with if it raised an exception to preven *accidental* mishandling of the return value in cases where it returns `None`.

p2p/kademlia.py
pipermerriam: Looking at the internals of `random.sample` we won't get any performance gains from this and are likely to take a performance hit (a little).  Easier to read though... I'm now leaning towards `-1` on using `random.sample` here.

However, I do think this implementation needs to be changed as this could take a **lot** of iterations if `len(all_bucket_nodes)` is large and `count` is also large as it will start to encounter more and more nodes it's already seen.

I came up with this solution.  Let me know what you think.

```python
def get_random_sample(self, count):
    all_node_idx_paths = [
        (bucket_idx, node_idx)
        for bucket_idx in range(len(self.buckets))
        for node_idx in range(len(self.buckets[bucket_idx].nodes))
    ]
    random.shuffle(all_node_idx_paths)
    selected_node_idx_paths = all_node_idx_paths[:count]
    return [
        self.buckets[bucket_idx].nodes[node_idx]
        for bucket_idx, node_idx
        in selected_node_idx_paths
    ]
```

p2p/discovery.py
gsalgado: That's a good point. Basically, CancelTokens are supposed to be passed on all the way through **our** async APIs until we end up calling any of python's asyncio APIs (i.e. as deep as possible in the call stack), at which point we have to use `wait_with_token()` so that the futures (created by `asyncio.wait()` and similar APIs) are cancelled when the token is triggered. IOW, whenever we use `asyncio.wait()` & co, we should use `wait_with_token()`. Does that make sense?

gsalgado: Created https://github.com/ethereum/py-evm/issues/449 for this

p2p/kademlia.py
gsalgado: Well, this method is a generator and is used in a loop where, on each iteration it will attempt to connect and perform two handshakes with the remote (which means several network round trips that may take up to 10s with the current timeouts we're using). Also, we break out of the loop that uses it as soon as we've connected to `PeerPool.min_peers` peers, so we're very unlikely to consume so many items as to cause it to slow down. Given that, I'd rather keep this code as simple as possible for now, but I'd be ok with using your solution if you disagree

pipermerriam: Somehow I missed that it was a generator and thus that it would exit early.  👍 on leaving it as-is.

PR #446, opened 2018-03-07
Add type hints for most of evm/utils

cburgdorf:
### What was wrong

`evm/utils` lacks type hints (see #132)

### How was it fixed

This PR adds most of the needed type hints for `evm/utils`. I left out some parts that I I want to follow up later. Also some type hints that were set to `Any` can probably be tighten up if one gives it more thought. I believe this will be an iterative process until we have 100 % coverage.

 #### Cute Animal Picture

### DIFF
--- a/evm/utils/datatypes.py
+++ b/evm/utils/datatypes.py
@@ -9,36 +9,40 @@ from eth_utils import (
 )
 
 
-def _is_local_prop(prop):
+from typing import Any, Dict, Tuple, Iterator, List
+
+
+def _is_local_prop(prop: str) -> bool:
     return len(prop.split('.')) == 1
 
 
-def _extract_top_level_key(prop):
+def _extract_top_level_key(prop: str) -> str:
     left, _, _ = prop.partition('.')
     return left
 
 
-def _extract_tail_key(prop):
+def _extract_tail_key(prop: str) -> str:
     _, _, right = prop.partition('.')
     return right
 
 
 @to_dict
-def _get_local_overrides(overrides):
+def _get_local_overrides(overrides: Dict[str, Any]) -> Iterator[Tuple[str, Any]]:
     for prop, value in overrides.items():
         if _is_local_prop(prop):
             yield prop, value
 
 
 @to_dict
-def _get_sub_overrides(overrides):
+def _get_sub_overrides(overrides: Dict[str, Any]) -> Iterator[Tuple[str, Any]]:
     for prop, value in overrides.items():
         if not _is_local_prop(prop):
             yield prop, value
 
 
 @to_dict
-def _get_sub_overrides_by_prop(overrides):
+def _get_sub_overrides_by_prop(
+        overrides: Dict[str, Any]) -> Iterator[Tuple[str, Dict[str, List[str]]]]:
     # we only want the overrides that are not top level.
     sub_overrides = _get_sub_overrides(overrides)
     key_groups = groupby(_extract_top_level_key, sub_overrides.keys())
@@ -47,7 +51,7 @@ def _get_sub_overrides_by_prop(overrides):
 
 
 @to_set
-def _get_top_level_keys(overrides):
+def _get_top_level_keys(overrides: Dict[str, Any]) -> Iterator[str]:
     for prop in overrides:
         yield _extract_top_level_key(prop)
 
@@ -83,6 +87,7 @@ class Configurable(object):
         sub_overrides_by_prop = _get_sub_overrides_by_prop(overrides)
 
         for key, sub_overrides in sub_overrides_by_prop.items():
+            sub_cls = None  # type: Configurable
             if key in local_overrides:
                 sub_cls = local_overrides[key]
             elif hasattr(cls, key):
--- a/evm/utils/db.py
+++ b/evm/utils/db.py
@@ -8,42 +8,49 @@ from evm.constants import (
     EMPTY_SHA3,
 )
 
+from evm.rlp.headers import BlockHeader
 
-def make_block_number_to_hash_lookup_key(block_number):
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from evm.db.chain import BaseChainDB  # noqa: F401
+
+
+def make_block_number_to_hash_lookup_key(block_number: int) -> bytes:
     number_to_hash_key = b'block-number-to-hash:%d' % block_number
     return number_to_hash_key
 
 
-def make_block_hash_to_score_lookup_key(block_hash):
+def make_block_hash_to_score_lookup_key(block_hash: bytes) -> bytes:
     return b'block-hash-to-score:%s' % block_hash
 
 
-def make_transaction_hash_to_data_lookup_key(transaction_hash):
+def make_transaction_hash_to_data_lookup_key(transaction_hash: bytes) -> bytes:
     '''
     Look up a transaction that is pending, after being issued locally
     '''
     return b'transaction-hash-to-data:%s' % transaction_hash
 
 
-def make_transaction_hash_to_block_lookup_key(transaction_hash):
+def make_transaction_hash_to_block_lookup_key(transaction_hash: bytes) -> bytes:
     return b'transaction-hash-to-block:%s' % transaction_hash
 
 
-def get_parent_header(block_header, db):
+def get_parent_header(block_header: BlockHeader, db: 'BaseChainDB') -> BlockHeader:
     """
     Returns the header for the parent block.
     """
     return db.get_block_header_by_hash(block_header.parent_hash)
 
 
-def get_block_header_by_hash(block_hash, db):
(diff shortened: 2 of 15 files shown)


evm/utils/db.py
```diff
@@ -8,50 +8,62 @@
     EMPTY_SHA3,
 )
 
+from evm.rlp.blocks import BaseBlock
+from evm.rlp.headers import BlockHeader
 
-def make_block_number_to_hash_lookup_key(block_number):
+from typing import Any
+
+# from evm.db.chain import BaseChainDB
```
cburgdorf: @gsalgado 

I was fighting really hard about circular dependencies with python here. Maybe you know a way out?
So the problem is that `evm.db.chain` imports this file (`evm.utils.db.py`) hence importing `BaseChainDB` here would create a circular dependency. 

Here's what I tried but failed:

- use forward references https://www.python.org/dev/peps/pep-0484/#forward-references
- use a `NewType` pointing to forward reference
- import just `evm` and then type it as `evm.db.chain.BaseChainDB` where it is needed.

Do you have any more ideas on how I could resolve this?

gsalgado: http://mypy.readthedocs.io/en/latest/common_issues.html#import-cycles

cburgdorf: :heart: that should work! Thank you

cburgdorf:
Ran into this while typing parts of the code https://github.com/python/mypy/issues/4709

cburgdorf:
Ah, still has a test failing. Will look into it soon and update the PR.

cburgdorf:
Green! I believe this is ready for review/merge!

evm/utils/datatypes.py
```diff
@@ -9,36 +9,40 @@
...
-def _get_local_overrides(overrides):
+def _get_local_overrides(overrides: Dict[str, Any]) -> Iterator[Tuple[str, int]]:
     for prop, value in overrides.items():
         if _is_local_prop(prop):
             yield prop, value
 
 
 @to_dict
-def _get_sub_overrides(overrides):
+def _get_sub_overrides(overrides: Dict[str, Any]) -> Iterator[Tuple[str, Any]]:
```
gsalgado: I don't think we could use something other than `Any` on the return value here, right? Usually we should avoid `Any` in return values as that makes it impossible for mypy to perform any checks on code that uses the values returned here

evm/utils/datatypes.py
```diff
@@ -105,7 +109,7 @@ def configure(cls,
                     )
                 )
 
-            configured_sub_cls = sub_cls.configure(**sub_overrides)
+            configured_sub_cls = sub_cls.configure(**sub_overrides)  # type: ignore
```
gsalgado: Why do we have to tell mypy to ignore this line? If there's no way around it, can you add a comment explaining why?

evm/utils/db.py
```diff
@@ -8,42 +8,50 @@
...
 )
 
+from evm.rlp.blocks import BaseBlock
+from evm.rlp.headers import BlockHeader
 
-def make_block_number_to_hash_lookup_key(block_number):
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from evm.db.chain import BaseChainDB  # noqa
```
gsalgado: Can you add the specific error we need to ignore to the `noqa` comment.

evm/utils/db.py
```diff
@@ -8,42 +8,50 @@
...
+
+
+def make_block_number_to_hash_lookup_key(block_number: int) -> bytes:
     number_to_hash_key = b'block-number-to-hash:%d' % block_number
     return number_to_hash_key
 
 
-def make_block_hash_to_score_lookup_key(block_hash):
+def make_block_hash_to_score_lookup_key(block_hash: str) -> bytes:
```
gsalgado: Why `str` and not `bytes`?

evm/utils/db.py
```diff
@@ -8,42 +8,50 @@
...
-def get_parent_header(block_header, db):
+def get_parent_header(block_header: BlockHeader, db: 'BaseChainDB') -> BaseBlock:
     """
     Returns the header for the parent block.
     """
     return db.get_block_header_by_hash(block_header.parent_hash)
 
 
-def get_block_header_by_hash(block_hash, db):
+def get_block_header_by_hash(block_hash: BlockHeader, db: 'BaseChainDB') -> BaseBlock:
```
gsalgado: The two methods above seem to return a block header, yet the return type is `BaseBlock`

evm/utils/env.py
```diff
@@ -215,4 +229,4 @@ def get(name, required=False, default=empty, type=None):
         'list': env_list,
         list: env_list,
     }.get(type, env_string)
-    return fn(name, default=default, required=required)
+    return fn(name, default=default, required=required)  # type: ignore
```
gsalgado: Again, can we have a comment stating why mypy must ignore this line?

evm/utils/headers.py
```diff
@@ -74,11 +76,11 @@ def compute_gas_limit(parent_header, gas_limit_floor):
 
 
 def generate_header_from_parent_header(
-        compute_difficulty_fn,
-        parent_header,
-        coinbase,
-        timestamp=None,
-        extra_data=b''):
+        compute_difficulty_fn: Callable,
```
gsalgado: Care to expand this to include the arguments it takes?

evm/utils/datatypes.py
cburgdorf: >Usually we should avoid Any in return values as that makes it impossible for mypy to perform any checks on code that uses the values returned here

Yes, fully agree here. We probably can tighten that up. I admit, This was me being too lazy to dig deep into how this code is being used. I'm gonna give it a shot ;)

evm/utils/datatypes.py
cburgdorf: `mypy` doesn't know that `sub_cls` is of (sub)type `Configurable`.

We could also initialize it to `None` before we first introduce it and add a comment type hint to set it to `Configureable`.

```
        for key, sub_overrides in sub_overrides_by_prop.items():
            sub_cls = None # type: Configurable
```

Do you have any other idea how to best resolve this?

evm/utils/db.py
cburgdorf: Good point. Will change.

evm/utils/db.py
cburgdorf: Yeah, good point. I typed that file before I figured out how to get monkeytype to help me (which I started to use later) and it seemed natural to have these lookup methods take `str`. If I had run this file through monkeytype (which I now did just to double check) I would have noticed that these methods take bytes.

I guess as a rule of thumb we are always dealing with bytes instead of strings (which makes a lot of sense so we don't have to worry about encodings)

evm/utils/db.py
cburgdorf: Yep, screwed that up. Fixed it.

evm/utils/env.py
cburgdorf: I took another try here. With the `mypy_extensions` it should be possible to type this out as `#type: Callable[[str, DefaultNamedArg(Union[Type[empty], T], 'default'), DefaultNamedArg(bool, 'required')], T]`

However this leaves me with:

```
evm/utils/env.py:216: error: Incompatible types in assignment (expression has type "function", variable has type "Callable[[str, DefaultNamedArg(Union[Type[empty], T], 'default'), DefaultNamedArg(bool, 'required')], T]")
```

I have no idea. I'll put a comment there as to explain the `# type: ignore` for now.

evm/utils/headers.py
cburgdorf: yep, let me try. I think this should be `Callable[[BlockHeader, int], int]`, right?

evm/utils/datatypes.py
cburgdorf: Gave it another try but failed to figure out how I could tighten the bounds of `Any` here. Basically because I'm not sure what to expect here but my intuition says this could really be anything ;)

In general, I believe we'll probably have to iterate over some things and fine tune over time as more and more of the rest of the code base gets typed. As the overall coverage grows it should become easier to find the right types.

cburgdorf:
@gsalgado Thank you for the review. I think I addressed all your comments. Mind giving it another look?

evm/utils/env.py
cburgdorf: @gsalgado I finally found something to improve that. The trick was to break down the initialization of the dict from the `get(...)` call so that I could add annotations for the `Dict`. (if we didn't have to use comment annotations this could be written without breaking it apart).

https://github.com/ethereum/py-evm/pull/459/commits/27a694dfdb8e3d1422d73dc678ab02e5c0eb20ce

It's still not perfect but at least it allows us to get rid of the mypy silencing. Anyway, no need to review/merge just yet. I'll again batch a couple of changes for the next PR.

PR #450, opened 2018-03-09
Blob utility functions

jannikluhn:
As I couldn't find anything more important to do, I went ahead and wrote some small utils for blobs which should hopefully be useful when the new spec is finalized.

- [x] chunk iterator: Iterates through a blob and yields chunks
- [x] chunk root calculation: Calculates the Merkle root of the chunks in a blob (which is quite slow, we'll probably have to optimize that in the future)

### DIFF
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -27,6 +27,11 @@ def chunk_iterator(blob):
         yield blob[chunk_start:chunk_start + CHUNK_SIZE]
 
 
+def hash_layer(layer):
+        for left, right in partition(2, layer):
+            yield keccak(left + right)
+
+
 def calc_merkle_root(leaves):
     leaves = list(leaves)  # convert potential iterator to list
     if len(leaves) == 0:
@@ -38,10 +43,6 @@ def calc_merkle_root(leaves):
     n_layers = int(n_layers)
 
     first_layer = (keccak(leaf) for leaf in leaves)
-
-    def hash_layer(layer):
-        for left, right in partition(2, layer):
-            yield keccak(left + right)
 
     root, *extras = pipe(first_layer, *itertools.repeat(hash_layer, n_layers))
     assert not extras, "Invariant: should only be a single value"
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -16,15 +16,14 @@ from cytoolz import (
 )
 
 
-def chunk_iterator(blob):
-    # TODO: proper blob serialization when specification is out
-    if len(blob) % CHUNK_SIZE != 0:
+def chunk_iterator(collation_body):
+    if len(collation_body) % CHUNK_SIZE != 0:
         raise ValueError("Blob size is {} which is not a multiple of chunk size ({})".format(
-            len(blob),
+            len(collation_body),
             CHUNK_SIZE,
         ))
-    for chunk_start in range(0, len(blob), CHUNK_SIZE):
-        yield blob[chunk_start:chunk_start + CHUNK_SIZE]
+    for chunk_start in range(0, len(collation_body), CHUNK_SIZE):
+        yield collation_body[chunk_start:chunk_start + CHUNK_SIZE]
 
 
 def hash_layer(layer):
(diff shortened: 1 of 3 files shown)


evm/utils/blobs.py
```diff
@@ -0,0 +1,68 @@
...
+            continue
+
+        try:
+            right = next(current_layer)
+        except StopIteration:
+            if len(next_layer) == 0:
+                # current_layer consists of a single element only, so we have reached the root
+                return left
+            else:
+                raise ValueError("Number of leaves must be a power of two")
```
pipermerriam: This validation step seems like it could also be done up-front on the `leaves` input variable.

evm/utils/blobs.py
```diff
@@ -0,0 +1,68 @@
...
+    if len(blob) % CHUNK_SIZE != 0:
+        raise ValueError("Blob size is {} which is not a multiple of chunk size ({})".format(
+            len(blob),
+            CHUNK_SIZE,
+        ))
+    for chunk_start in range(0, len(blob), CHUNK_SIZE):
+        yield blob[chunk_start:chunk_start + CHUNK_SIZE]
+
+
+def calc_merkle_root(leaves):
```
pipermerriam: I can't help myself when I see something like this which can be done functionally.  What do you think about the following?

1. do up-front validation that `is_power_of_2(len(leaves)) == True`.
2. calc `p` such that `2 ** p == len(leaves)`
3. run a non-loop based reduction using a fixed number of `cytoolz.partition` based steps.

```python
import math
import itertools
from cytoolz import pipe, compose

@to_tuple
def hash_layer(layer):
    for left, right in partition(2, layer):
        yield keccak(left + right)

def calc_merkle_root(leaves):
    p = math.log2(len(leaves))
    if p != int(p):
        # not a power of 2
    root, *extras = pipe(leaves, *itertools.repeat(hash_layer, int(p)))
    if extras:
        raise Exception("Invariant: should only be a single value")
    return root
```

evm/utils/blobs.py
```diff
@@ -0,0 +1,68 @@
...
+    keccak,
+)
+
+from evm.constants import (
+    CHUNK_SIZE,
+    COLLATION_SIZE,
+)
+
+
+def chunk_iterator(blob):
```
hwwhww: The **blob serialization** process would be more complicated than partitions. It requires some special encoding to serialize blobs to chunks. I guess Vitalik or Justin will post some preliminary ideas on ethresear.ch soon. Right now you can mark a `TODO: blob serialization` comment.

evm/utils/blobs.py
jannikluhn: I like this, the only issue is that it wouldn't work with iterators (because the number of leaves is not known in advance). I tried to make use of iterators for the sake of efficiency as calculating the Merkle root of a full-sized block already takes almost a second on my machine. But blobs are only a megabyte in size and iterators should only decrease memory usage, so that argument is flawed. So I'm gonna implement your suggestion and work with lists.

evm/utils/blobs.py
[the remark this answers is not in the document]
jannikluhn: Did not know about these functions, but they appear to convert the bytes to integers (`list(partition(2, b"asdf"))` becomes `[(97, 115), (100, 102)]`). So I'd rather leave it like this, also considering that we'll have to change it at some point anyway (see https://github.com/ethereum/py-evm/pull/450#discussion_r174029433)

jannikluhn: I think I've got the terminology wrong: What I called `blob` should actually be `collation_body`. My (updated) understanding right now:

- `collation_body`: `2**20` bytes of binary data
- `chunks`: 32 bytes piece of the body (so the body is made up of 32768 chunks, simply stringed together), needed to calculate the chunks root
- `blobs`: Another "interpretation" of the collation body. Eventually these will be transactions and we'll need some way to press blobs into the chunk structure, but for phase 1 we can don't need to consider them at all.

Is this correct?

evm/utils/blobs.py
pipermerriam: In *theory* this can be made to work with iterators, but you'll likely have to modify the implementation to be recursive which is *probably* fine, but could be problematic at high depths which should be easy enough to validate.

Either way is fine with me.

evm/utils/blobs.py
```diff
@@ -0,0 +1,59 @@
...
+        raise ValueError("No leaves given")
+
+    n_layers = math.log2(len(leaves))
+    if not n_layers.is_integer():
+        raise ValueError("Leave number is not a power of two")
+    n_layers = int(n_layers)
+
+    first_layer = (keccak(leaf) for leaf in leaves)
+
+    def hash_layer(layer):
```
pipermerriam: This function definition should probably be moved up to the module level.

Issue #453, opened 2018-03-09
estimate_gas should work with unsigned transactions


dylanjw:
* py-evm Version: v0.2.0-alpha.6-621-gb1967a8
* OS: linux
* Python Version (python --version): 3.6.3

### What is wrong?

Gas estimation should not require a signed transaction.  An example use case is getting a gas price estimate for transactions before local signing, where the private key is not available to the node.

### How can it be fixed

The instance of `SpoofTransaction` can be extended to provide the rest of the missing attributes from an unsigned transaction object.  The following stack trace + exceptions show each dependent attribute that would need to be added to SpoofTransaction when passing an unsigned transaction object.  Note, Im passing in a SpoofTransaction object to chain.Chain.estimate_gas(), adding override attributes one at a time to get all required attributes: 

`instrinsic_gas`

```python
eth_tester/main.py:445: in estimate_gas
    raw_gas_estimate = self.backend.estimate_gas(raw_transaction)
eth_tester/backends/pyevm/main.py:444: in estimate_gas
    return self.chain.estimate_gas(spoof_evm_transaction)
../py-evm/evm/chains/base.py:478: in estimate_gas
    return self.gas_estimator(state, transaction)
cytoolz/functoolz.pyx:232: in cytoolz.functoolz.curry.__call__
    ???
../py-evm/evm/estimators/gas.py:54: in binary_gas_search
    minimum_transaction = SpoofTransaction(transaction, gas=transaction.intrinsic_gas)
[...]
E           AttributeError: 'SpuriousDragonUnsignedTransaction' object has no attribute 'intrinsic_gas'
```

`get_sender()`

```python
    def __init__(self, transaction, **overrides):
        if 'get_sender' not in overrides:
>           current_sender = transaction.get_sender()
E           AttributeError: 'SpuriousDragonUnsignedTransaction' object has no attribute 'get_sender'
```

`s`, `r`, & `v` in `validate_transaction`

```
eth_tester/main.py:445: in estimate_gas
    raw_gas_estimate = self.backend.estimate_gas(raw_transaction)
eth_tester/backends/pyevm/main.py:444: in estimate_gas
    return self.chain.estimate_gas(spoof_evm_transaction)
../py-evm/evm/chains/base.py:478: in estimate_gas
    return self.gas_estimator(state, transaction)
cytoolz/functoolz.pyx:232: in cytoolz.functoolz.curry.__call__
    ???
../py-evm/evm/estimators/gas.py:55: in binary_gas_search
[...]

PR #454, opened 2018-03-09
Allow unsinged tx's with estimate_gas

dylanjw:
Built on #432

### What was wrong?

see #453 

#### Cute Animal Picture


dylanjw:
Using a spoofed transaction would still require providing a valid address with enough ether to cover the gas.

```
evm/chains/base.py:478: in estimate_gas
    return self.gas_estimator(state, transaction)
cytoolz/functoolz.pyx:232: in cytoolz.functoolz.curry.__call__
    ???
evm/estimators/gas.py:67: in binary_gas_search
    if _get_computation_error(state, minimum_transaction) is None:
evm/estimators/gas.py:35: in _get_computation_error
    computation = state.execute_transaction(transaction)
evm/vm/forks/frontier/vm_state.py:223: in execute_transaction
    computation = _execute_frontier_transaction(self, transaction)
evm/vm/forks/frontier/vm_state.py:56: in _execute_frontier_transaction
    vm_state.validate_transaction(transaction)
evm/vm/forks/frontier/vm_state.py:231: in validate_transaction
    validate_frontier_transaction(self, transaction)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

vm_state = <evm.vm.forks.frontier.vm_state.FrontierVMState object at 0x7fb770bc0c88>, transaction = <evm.utils.spoof.SpoofTransaction object at 0x7fb770bc0c50>

    def validate_frontier_transaction(vm_state, transaction):
        gas_cost = transaction.gas * transaction.gas_price
        sender_balance = vm_state.read_only_state_db.get_balance(transaction.sender)
    
        if sender_balance < gas_cost:
            raise ValidationError(
>               "Sender account balance cannot afford txn gas: `{0}`".format(transaction.sender)
            )
E           evm.exceptions.ValidationError: Sender account balance cannot afford txn gas: `b'00000000000000000000'`
```

pipermerriam:
I think the *first* option is what I suggested wherever else this was mentioned, which is to go see what go-ethereum is doing and whether or not we can just lift their gas estimation implementation.

Second is to look into performing an irregular state change prior to gas estimation to pre-fund the account (only when it doesn't have enough gas)?

dylanjw:
Im working on a hacky prototype to just see how things work.  Im probably not grasping some basic concepts.

Im stuck on the following.  After setting the balance to the max, subsequent state_db contexts don't show the change in balance.  My assumption was the mutable_state_db would have a lasting effect on the state, but maybe it is meant as a temporary state?

```
(Pdb) list
251  	                _transaction.get_sender = \
252  	                    lambda: to_bytes(hexstr='0x7760c6D8c0eA7A764F13fb5b034A574f59c1fdd4')
253  	                _transaction.sender = to_bytes(hexstr='0x7760c6D8c0eA7A764F13fb5b034A574f59c1fdd4')
254  	
255  	            state_db.set_balance(transaction.sender, UINT_256_MAX, validate=False)
256  ->	            computation = self.execute_transaction(_transaction)
257  	
258  	        self.revert(snapshot)
259  	
260  	        return computation
261  	
(Pdb) state_db.get_balance(transaction.sender)
115792089237316195423570985008687907853269984665640564039457584007913129639935
>>> with self.state_db() as db:
...   db.get_balance(transaction.sender)
... 
0
>>> 
now exiting InteractiveConsole...
```

pipermerriam:
I can't see your code so this is a guess.

https://github.com/ethereum/py-evm/blob/master/evm/vm_state.py#L117

I suspect you're still within the context manager context for the mutable statedb which doesn't commit it's state until the context exits.  This means that any *nested* retrieval of the state_db within that context won't be operating on the same state root (it would be operating on whatever the previous one was) so it would show the previous balance.

dylanjw:
Thanks!  I missed that detail of the db state context.

dylanjw:
@pipermerriam Im getting the following error in Travis, and I can't tell if this is from changes i have made: `trie.exceptions.InvalidNode: Unable to parse node`.  Has this come up before?  Could the irregular state change be having some unintended effect?

evm/vm_state.py
```diff
@@ -236,6 +241,33 @@ def get_transaction_context_class(cls):
...
+        _transaction.r = 1
+
+        snapshot = self.snapshot()
+
+        with self.mutable_state_db() as state_db:
+
+            if not hasattr(_transaction, "get_sender"):
+                # max out eth in random hardcoded address. what could go wrong?
+                _transaction.get_sender = \
+                    lambda: to_bytes(hexstr='0x7760c6D8c0eA7A764F13fb5b034A574f59c1fdd4')
```
pipermerriam: Might be good to store this as a constant

```python
DEFAULT_DO_CALL_SENDER = ...
```

evm/vm_state.py
```diff
@@ -236,6 +241,33 @@ def get_transaction_context_class(cls):
...
+                _transaction.sender = to_bytes(hexstr='0x7760c6D8c0eA7A764F13fb5b034A574f59c1fdd4')
+
+            state_db.set_balance(transaction.sender, UINT_256_MAX, validate=False)
+
+        computation = self.execute_transaction(_transaction)
+
+        self.revert(snapshot)
```
pipermerriam: This should probably go in a `finaly` block of a `try/finally` to ensure that the reversion happens even if an exception occurs.

tests/core/helpers.py
```diff
@@ -12,7 +18,8 @@ def new_transaction(
         private_key,
         gas_price=10,
         gas=100000,
-        data=b''):
+        data=b'',
+        signed=True):
```
pipermerriam: Instead of `signed=True` we could instead make `private_key` optional and only sign it if the `private_key is not None`.

jannikluhn:
> Is this potentially a bug in the binary trie?

Could be, but I don't think so. Either way, I don't think that's the reason why the build is failing, as this PR shouldn't affect what tests like `test_auxiliary/test_contract.py` write to the DB.

One thing that could (or could not) be happening is that a wrong value is used as a state root somewhere. If at that place some other value is stored then above error would be raised. No idea why this seems to happen with the binary trie only though.

@NIC619 who implemented the binary trie might have better insights.

evm/db/state.py
```diff
@@ -205,15 +205,17 @@ def delete_storage(self, address):
     #
     # Balance
     #
-    def get_balance(self, address):
-        validate_canonical_address(address, title="Storage Address")
+    def get_balance(self, address, validate=True):
```
pipermerriam: It is not clear to me why we need a flag here to disable validation.  First impression is that this is not a good idea as it allows invalid data to get into the state db.

evm/vm_state.py
```diff
@@ -236,6 +245,34 @@ def get_transaction_context_class(cls):
...
+        try:
+            with self.mutable_state_db() as state_db:
+
+                if not hasattr(_transaction, "get_sender"):
+                    # max out eth in random hardcoded address. what could go wrong?
+                    _transaction.get_sender = \
+                        lambda: to_bytes(hexstr=DEFAULT_DO_CALL_SENDER)
+                    _transaction.sender = to_bytes(hexstr=DEFAULT_DO_CALL_SENDER)
+
+                state_db.set_balance(transaction.sender, UINT_256_MAX, validate=False)
```
pipermerriam: This appears to be the only call site which uses the `validate=...` API.  It's not clear why it would be needed here.  Both of these values appear to be valid.

evm/db/state.py
dylanjw: We do not need it.  Initially I thought I would have to step around the validation in order for the set_balance to work with an arbitrary non-existent address. What was really happening, is the address I had been using was in fact not valid.  I'll take this out.

Does the arbitrary address that I hardcoded have the risk of colliding with a real address?  Are addresses generated (with eth_account) but never used in an on-chain transaction safely considered unique?

pipermerriam:
> Does the arbitrary address that I hardcoded have the risk of colliding with a real address? 

My recommendation would be to use an address with a private key like `b'10' * 16` so that:

1. it's clear that it's not a *real* private key.
2. if it collides with someone's actual address, it's a private key that nobody should be using for anything that matters.

evm/vm_state.py
```diff
@@ -236,6 +245,34 @@ def get_transaction_context_class(cls):
     #
     # Execution
     #
+
+    def do_call(self, transaction):
```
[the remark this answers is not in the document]
dylanjw: I was imagining it being used for "eth_call".

NIC619:
@dylanjw #511 should fix the node parsing error. Sorry for the trouble.

evm/vm_state.py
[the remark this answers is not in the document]
pipermerriam: My hesitance comes from the lack of congruence with the other APIs.  The concept of *calling* and *estimating gas* are JSON-RPC level concepts and while we have incorporated a gas estimation API in the `Chain` class, I would really like to see these get extracted out of the core API, maybe into something that lives in `evm.tools`.

Lets leave as-is for now and we can see about moving this stuff elsewhere in the codebase later.

evm/rlp/transactions.py
```diff
@@ -180,6 +180,30 @@ def as_signed_transaction(self, private_key: bytes) -> 'BaseTransaction':
...
+        """
+        raise NotImplementedError("Must be implemented by subclasses")
+
+    def gas_used_by(self, computation: BaseComputation) -> int:
+        """
+        Return the gas used by the given computation. In Frontier,
+        for example, this is sum of the intrinsic cost and the gas used
+        during computation.
+        """
+        return self.get_intrinsic_gas() + computation.get_gas_used()
```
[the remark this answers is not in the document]
dylanjw: I'll take a stab at this.

pipermerriam:
I pushed a rebase for this.  I think it's good to merge, looking it over one more time while CI runs.

Issue #160, opened 2017-11-16
Create a new BatchDB class for atomic database operations.


pipermerriam:
### What is wrong?

There are a few cases where we need to make multiple database updates and have them either all succeed or all fail.

### How can it be fixed

Implement a new class `BatchDB` in a new module `evm/db/batch.py`

This class should serve as a *wrapper* in much the same way as the current `JournalDB` found in the module `evm/db/journal.py`.

This class should be usable as a context manager similar to the following:

```python
with BatchDB(db) as batch_db:
    batch_db[b'a'] = b'arst'  # doesn't write the key to the database but rather to a local cache.
    assert batch_db[b'a'] == b'arst'  # intelligently returns the value from the cache
    batch_db[b'b'] = b'tsra'  
# upon exiting the context, it writes all of the keys from the cache into the underlying database.
```

Or via an alternate API that requires manually committing the batch.

```python
batch_db = BatchDB(db)
batch_db[b'a'] = b'arst'  # doesn't write the key to the database but rather to a local cache.
assert batch_db[b'a'] == b'arst'  # intelligently returns the value from the cache
batch_db[b'b'] = b'tsra'  
batch_db.commit()
```

If an exception is raised before committing (or leaving the context when used as a context manager) none of the changes should be committed to the database.

pipermerriam:
See: https://github.com/pipermerriam/py-evm/issues/135 for indication where this new database wrapper should be used.

carver:
:+1: for the context manager API

pipermerriam:
one of #455 or #322 will closes this.  I still think this is a useful API but I'm holding off on merging either PR until we have a direct use case which I forecast will show up after the launch of trinity.

PR #455, opened 2018-03-10
add BatchDB to allow commit DB update all fail or all success

sunxivincent:
### What was wrong?
[issue 160](https://github.com/ethereum/py-evm/issues/160)

### How was it fixed?
warps a cache with underlying db and use context manager to apply batch updates

#### Cute Animal Picture


pipermerriam:
Merged so we can experiment with alternate jounaling approach as described in #542

PR #458, opened 2018-03-12
p2p: body/receipt downloaders now batch items to reduce # of requests

gsalgado:


p2p/chain.py
```diff
@@ -312,6 +327,15 @@ def _should_skip_receipts_download(self, header: BlockHeader) -> bool:
             self.logger.warn("Got unexpected msg: %s (%s)", cmd, msg)
 
 
+def _make_data_tries(items: List[List[Union[BaseTransaction, Receipt]]]
+                     ) -> List[Tuple[bytes, Dict[Any, Any]]]:
+    trie_data_list = []
+    for item in items:
+        trie_root, trie_dict_data = make_trie_root_and_nodes(item)
+        trie_data_list.append((trie_root, trie_dict_data))
+    return trie_data_list
```
pipermerriam: *nitpick*

For consistency, this is a textbook candidate for `eth_utils.to_list` (or `to_tuple` if we don't need the returned valuable to be mutable).

However, better might be to just use [`itertools.starmap`](https://docs.python.org/3.5/library/itertools.html#itertools.starmap) to get rid of this function entirely since I think it does exactly this.

gsalgado: Actually, just a regular `map` does the trick

gsalgado:
```
iterator = map(make_trie_root_and_nodes, msg)
receipts_tries = await loop.run_in_executor(self.executor, list, iterator)
```

pipermerriam: Sure enough.  I saw the multiple return values and got cross-wired that there was something which needed to be called with star-args `fn(*args)`.

👍

PR #459, opened 2018-03-12
Add type hints for evm/rlp

cburgdorf:
### What was wrong?

evm/rlp lacked type hints

### How was it fixed?

Added type hints

#### Cute Animal Picture

### DIFF
--- a/evm/rlp/blocks.py
+++ b/evm/rlp/blocks.py
@@ -9,6 +9,11 @@ from evm.utils.datatypes import (
     Configurable,
 )
 
+from evm.db.chain import BaseChainDB
+
+from .transactions import BaseTransaction
+from .headers import BlockHeader
+
 
 class BaseBlock(rlp.Serializable, Configurable, metaclass=ABCMeta):
 
@@ -21,14 +26,14 @@ class BaseBlock(rlp.Serializable, Configurable, metaclass=ABCMeta):
     transaction_class = None
 
     @classmethod
-    def get_transaction_class(cls):
+    def get_transaction_class(cls) -> BaseTransaction:
         if cls.transaction_class is None:
             raise AttributeError("Block subclasses must declare a transaction_class")
         return cls.transaction_class
 
     @classmethod
     @abstractmethod
-    def from_header(cls, header, chaindb):
+    def from_header(cls, header: BlockHeader, chaindb: BaseChainDB) -> 'BaseBlock':
         """
         Returns the block denoted by the given block header.
         """
@@ -36,23 +41,23 @@ class BaseBlock(rlp.Serializable, Configurable, metaclass=ABCMeta):
 
     @property
     @abstractmethod
-    def hash(self):
+    def hash(self) -> bytes:
         raise NotImplementedError("Must be implemented by subclasses")
 
     @property
     @abstractmethod
-    def number(self):
+    def number(self) -> int:
         raise NotImplementedError("Must be implemented by subclasses")
 
     @property
-    def is_genesis(self):
+    def is_genesis(self) -> bool:
         return self.number == 0
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return '<{class_name}(#{b})>'.format(
             class_name=self.__class__.__name__,
             b=str(self),
         )
 
-    def __str__(self):
+    def __str__(self) -> str:
         return "Block #{b.number}".format(b=self)
--- a/evm/rlp/headers.py
+++ b/evm/rlp/headers.py
@@ -49,6 +49,14 @@ from evm.vm.execution_context import (
     ExecutionContext,
 )
 
+from typing import (
+    Tuple,
+    Union,
+    Iterator,
+    List,
+    Any
+)
+
 
 class BlockHeader(rlp.Serializable):
     fields = [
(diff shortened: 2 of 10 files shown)


pipermerriam:
Not to deter you from fixing this, but I think it's low priority/impact for that functionality to be typed.  For most *anything* there that could be done wrong it would show up very quickly in the test suites.

cburgdorf:
Totally agree. It was more like a personal vendetta I had with mypy there.

cburgdorf:
This should be ready for review. It went green on CI already pushed pushed again for a typo fix in a comment.

evm/rlp/blocks.py
```diff
@@ -16,35 +21,35 @@ def get_sedes(cls):
...
-    def from_header(cls, header, chaindb):
+    def from_header(cls, header: BlockHeader, chaindb: BaseChainDB) -> 'BaseBlock':
         """
         Returns the block denoted by the given block header.
         """
         raise NotImplementedError("Must be implemented by subclasses")
 
     @property
-    def hash(self):
+    def hash(self) -> bytes:
```
pipermerriam: @gsalgado what do you think about us creating a custom type for 32 byte hashes (as outlined here: http://mypy.readthedocs.io/en/latest/kinds_of_types.html#newtypes)

This would make things a bit stricter, ensuring that we are in-fact dealing with hashes rather than just any arbitrary `bytes` value.

evm/rlp/headers.py
```diff
@@ -70,21 +79,21 @@ class BlockHeader(rlp.Serializable):
...
-                 receipt_root=BLANK_ROOT_HASH,
-                 bloom=0,
-                 gas_used=0,
-                 extra_data=b'',
-                 mix_hash=ZERO_HASH32,
-                 nonce=GENESIS_NONCE):
+                 difficulty: int,
+                 block_number: int,
+                 gas_limit: int,
+                 timestamp: Optional[int] = None,
```
pipermerriam: I *think* that when `None` is the default value that `Optional[...]` is inferred if you just use `int`.

evm/rlp/transactions.py
```diff
@@ -36,22 +44,22 @@ class BaseTransaction(rlp.Serializable):
...
         return rlp.decode(rlp.encode(transaction), sedes=cls)
 
     @property
-    def hash(self):
+    def hash(self) -> bytes:
         return keccak(rlp.encode(self))
 
     @property
-    def sender(self):
+    def sender(self) -> bytes:
```
pipermerriam: Similarly, we could create a custom `address` type (like outlined here in mypy docs http://mypy.readthedocs.io/en/latest/kinds_of_types.html#newtypes)

This would prevent us from accepting just *any* bytes value as an address.

evm/rlp/blocks.py
cburgdorf: I'd be very much in favor of this!

gsalgado: Sounds good to me, but since this would likely be useful in other libraries as well (e.g. eth-utils and eth-keys?), I wonder if maybe it should be defined in one of those instead?

cburgdorf: Yes, especially because these types will move between layers. For instance the evm may return an `address` to trinity and if these are different NewTypes (one defined in evm, one in trinity) then mypy should complain about it. So moving those into a shared lib makes much sense.

pipermerriam: In that case, we should probably create a very thin `eth-typing` library with this stuff in it because I don't *think* `eth-utils` isn't a global dependency across all of the `eth` codebases.

pipermerriam: options:

1. do it locally now and migrate to something from `eth-typing` later.
2. keep it `bytes` now and migrate to something from `eth-typing` later.

I'm fine with either.

cburgdorf: I'd vote for 2.)

evm/rlp/headers.py
cburgdorf: Well, to be precise `mypy` infers that to `Union[builtins.int, builtins.None]` which is basically just a longer version of `Optional[int]`. However, I would not recommend to annotate it as `int` simply because I think the annotation should be aligned with the actual type (e.g. the one mypy would infer itself). So a developer looking at the code can directly see that passing a value is optional here. Your call!

pipermerriam: I'm going to go with just `int` under the reasoning that the `=None` portion of this is sufficient to signal that the value is option.

cburgdorf: No problem. I updated all of these cases from `Optional[T]` to just `T`

cburgdorf:
@pipermerriam updated and rebased!

evm/rlp/headers.py
```diff
@@ -126,15 +134,15 @@ def hex_hash(self):
...
-                    extra_data=None,
-                    transaction_root=None,
-                    receipt_root=None):
+                    parent: 'BlockHeader',
+                    gas_limit: int,
+                    difficulty: int,
+                    timestamp: int,
+                    coinbase: bytes = ZERO_ADDRESS,
+                    nonce: bytes = None,
+                    extra_data: bytes = None,
```
pipermerriam: I'm seeing inconsistent spacing around the `=` in these function/method signatures.  Can you update all of these to use the same style.  Preference is no spaces around the `=`

```python
def my_method(self,
              arg1: int=None) -> None:
    ...
```

cburgdorf: Sorry, for that again. Should be fixed now. I guess there should be some linter for that to have this blow up in CI directly. Will look into it.

cburgdorf: There's [pylint](https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements). Would you be open to use that if it turns out we could configure it to help eliminate such inconsistencies?

pipermerriam: Feels like something that our current linter `flake8` should have caught or should be able to catch.  Wondering if there's a setting we need to enable/disable.

cburgdorf: Will look into it

gsalgado: Another alternative is https://github.com/ambv/black/, which we could use (locally, when developing) to automatically format our code (with a CI check to ensure everyone is using it), although it's not configurable so we'd need to adopt its chosen style

pipermerriam: https://github.com/ethereum/py-evm/issues/476

PR #460, opened 2018-03-13
Update vmc to latest vyper

NIC619:
### How was it fixed?
Update VMC to latest Viper version

#### Cute Animal Picture


evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -1,20 +1,20 @@
 # Events
 CollationAdded: __log__({
-    shard_id: indexed(num),
-    expected_period_number: num,
+    shard_id: indexed(int128),
```
hwwhww: Sorry for that I haven't followed Vyper for a while, any reason that you're using signed integer 128 bit `int128` instead of `uint256` for this and following cases?

evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -182,78 +181,79 @@ def withdraw(validator_index: num) -> bool:
...
+    if parent_hash != convert(0, 'bytes32'):
         assert self.collation_headers[shard_id][parent_hash].score > 0
     # Check if only one collation in one period perd shard
     assert self.period_head[shard_id] < expected_period_number
 
     # Check the signature with validation_code_addr
-    validator_addr = self.get_eligible_proposer(shard_id, block.number / self.period_length)
-    assert validator_addr != zero_addr
+    validator_addr: address = self.get_eligible_proposer(shard_id, floor(block.number / self.period_length))
+    assert not not validator_addr
```
hwwhww: Is this a pattern of Vyper? (two `not`?)

evm/vm/forks/sharding/contracts/validator_manager.v.py
NIC619: Basically it's because `num` is renamed to `int128` and `num256` to `uint256`, but I'm fine with both. One advantage of going with `uint256` I can see is that converting `bytes32` into `int128` requires a few extra steps. 

first convert `bytes32` to `uint256`
(extra step)-> then mod 2^127
(extra step)-> then convert to `int128`

evm/vm/forks/sharding/contracts/validator_manager.v.py
NIC619: Yes, see https://github.com/ethereum/vyper/blob/5b70797a31ffcbdd90b1dfd98f7b38d306bca7a0/examples/voting/ballot.v.py#L30-L32

pipermerriam: Is the loosely equivalent to `bool(thing)` in python?

evm/vm/forks/sharding/contracts/validator_manager.v.py
hwwhww: I think it's reasonable to use `uint256` for the numbers that we already know that won't be negative.

NIC619: After I tried it out, it seems that the default type is `int128`(literal numbers and operators) . So if we turn all variables of type `int128` into type `uint256`, we will first need to convert literal number to `uint256` and use uint256-ish operators like `uint256_add`, `uint256_mul`, etc.

default:
```python
score: int128 = parent_score + 1
```

if set all variables to 'uint256' ==>

```python
score: uint256 =uint256_add(parent_score, convert(1, 'uint256'))
assert score > convert(1, 'uint256')
```

evm/vm/forks/sharding/contracts/validator_manager.v.py
NIC619: Yes I believe so.

evm/vm/forks/sharding/contracts/validator_manager.v.py
NIC619: Sorry I didn't realize that in spec `uint256` is used. I will update them to `uint256`.

Issue #420, opened 2018-02-28
Create the new Collation RLP object


hwwhww:
### What is wrong?
(Updated the issue, separated this ticket from VMC contract)

#### `CollationHeader`
```
shard_id           uint256 // pointer to shard
parent_hash        bytes32 // pointer to parent header
chunks_root        bytes32 // pointer to collation body
period             uint256 // timestamp
height             uint256 // height in collation tree
proposer_address   address
proposer_bid       uint256
proposer_signature bytes
```

#### `Collation`
```
header     CollationHeader
blobs        list of blob
```

### How can it be fixed

Since there are two other issues are related to VMC, it might be better to create a new `Collation` class to avoiding influencing VMC.

mhchia:
I think I can work on this.
Change the collation structure according to the new spec, and do VMC optimization

hwwhww:
@mhchia 
1. Some fields renamings are still in discussion.
2. Removing `state_root` of `Collation` may make some tests fails due to this line: https://github.com/ethereum/py-evm/blob/3784a87d0850c92d5865d97ce4bb6f5075d47896/evm/vm/forks/sharding/vm_state.py#L45
    Which means we *may* have to make another *temporary* `NewCollation` class **or** decouple state execution.

I suggest something you can do before this issue:
1. Move VMC out of `evm/vm` because (1) it's not part of VM and (2) help for decoupling state execution and other processes.
2. Ping @pipermerriam , what do you think of we move VMC contract to `ethereum/sharding` repository and we make sharding repository to be a submodule of py-evm? If it's okay, @mhchia would you help to move?
2. Write down basic VMC API doc. (don't need to be perfect)

mhchia:
@hwwhww Thank you for the suggestion. I would like to help. But I think I should finish `GUESS_HEAD` first, and then come to this issue after. So sorry for this.
If anyone would like to do this please feel free to take it.

jannikluhn:
This seems to overlap with #436

hwwhww:
@jannikluhn A little bit, but my plan is that  #420 defines the `CollationHeader` and the blobs in `CollationBody` are just binary data; #436 handles the blobs encoding and the binary chunk Merkle tree.

hwwhww:
Updated to ticket description.

mhchia:
@jannikluhn 
[...]

PR #462, opened 2018-03-13
New collation format

jannikluhn:
This implements the new collation format described in #420.

I deviated from the description slightly by removing the `period` field for now (see https://github.com/ethereum/py-evm/issues/420#issuecomment-372554339) and by replacing `blobs` with `body`. I think this makes sense at least for phase 1 because we don't care yet how the blobs will be encoded.

As expected this breaks a lot of tests (basically all that have anything to do with sharding). I'm not sure what the best way to deal with this is. My idea would simply be to remove everything that deals with transaction execution (this should include all failing tests), which isn't included in phase 1 anyway. Basically everything from `ShardingVM` and below. When we progress to stage 2 we can recover everything from the git history.


evm/rlp/headers.py
```diff
@@ -179,83 +179,135 @@ def create_execution_context(self, prev_hashes):
...
+            int_to_bytes32(self.shard_id),
+            self.parent_hash,
+            self.chunks_root,
+            int_to_bytes32(self.period),
+            int_to_bytes32(self.height),
+            pad32(self.proposer_address),
+            int_to_bytes32(self.proposer_bid),
+            self.proposer_signature,
+        ])
+        assert len(encoded) == self.smc_encoded_size
```
pipermerriam: We have tended to avoid assertions in runtime code as they are [easy to disable](https://stackoverflow.com/questions/1273211/disable-assertions-in-python).  Preference is to raise explicit exceptions like `ValueError` in cases like this.

evm/rlp/headers.py
```diff
@@ -179,83 +179,135 @@ def create_execution_context(self, prev_hashes):
...
+    assert [name for name, _ in smc_field_sizes] == [name for name, _ in fields]
+
+    def __repr__(self):
+        return "<CollationHeader #{0} {1} (shard #{2})>".format(
+            self.height,
+            encode_hex(self.hash)[2:10],
+            self.shard_id,
+        )
+
+    def to_bytes(self):
```
pipermerriam: I'm confused how `to_bytes` and `rlp.encode(obj)` are related? or how `to_bytes` is used. Looking at the implementation of the `hash` property I would have expected it to be `keccak(rlp.encode(self))`.

Can we get a comment explaining the intent behind `to_bytes` as well as maybe another one in the `hash` implementation which explains why it uses `to_bytes` over `rlp.encode`.

evm/rlp/headers.py
```diff
@@ -179,83 +179,135 @@ def create_execution_context(self, prev_hashes):
...
                     len(header_bytes),
                     encode_hex(header_bytes),
                 )
             )
-        for idx, field in enumerate(cls.fields):
-            field_name, field_type = field
-            start_index = idx * obj_size
-            field_bytes = header_bytes[start_index:(start_index + obj_size)]
+
+        start_indices = accumulate(lambda i, field: i + field[1], cls.smc_field_sizes, 0)
```
pipermerriam: I've never used `accumulate` and now I ❤️ it!

evm/rlp/headers.py
```diff
@@ -179,83 +179,135 @@ def create_execution_context(self, prev_hashes):
...
+        start_index = next(start_indices)
+        for field_name, field_type in cls.fields:
+            end_index = next(start_indices)
```
pipermerriam: I think this loop can be changed as follows.

```
from cytoolz import sliding_window
field_bounds = sliding_window(2, start_indices)
for ((start_idx, end_idx), (field_name, field_type)) in zip(cls.fields, field_bounts):
    ...
```

Removes mutability as well as extracting the start/end index logic out of the loop body which I think improves readability some.

evm/rlp/headers.py
jannikluhn: `to_bytes` encode by padding each field to a fixed length and concatenating them. So it's less space efficient than RLP encoding, but much easier to decode for the SMC (which also has to calculate the hash).

I'm not very happy with the names `to_bytes` and `from_bytes` as it doesn't distinguish from RLP encoding, but I wanted to stay close to the existing implementation (there used to be `from_bytes`, I've added `to_bytes`). Maybe a name such as `encode_for_SMC` would clear things up already (but `decode_from_SMC` sounds weird).

pipermerriam: 👍 

I like `encode/decode_for/from_SMC` more than `to/from_bytes`.  Alternate names that come to mind.

- `encode_with_padding`
- `decode_padded`

jannikluhn: @mhchia Which names would you prefer?

holgerd77:
Hi, I did some stub implementation of collations for ethereumjs today heavily inspired by what you guys are doing here: https://github.com/ethereumjs/ethereumjs-collation

I was especially unsure about the field lengths for the different collation headers, this is what a default header initialization now look like:

```javascript
{ raw:
   [ <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>,
     <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>,
     <Buffer 56 e8 1f 17 1b cc 55 a6 ff 83 45 e6 92 c0 f8 6e 5b 48 e0 1b 99 6c ad c0 01 62 2f b5 e3 63 b4 21>,
     <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>,
     <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>,
     <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>,
     <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00>,
     <Buffer 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 ... > ],
  _fields:
   [ 'shardId',
     'parentHash',
     'chunkRoot',
     'period',
     'height',
     'proposerAddress',
     'proposerBid',
     'proposerSignature' ]
```

Would be great if you could have a short look if you immediately spot something wrong.

Another question: do you already have some RLP-serialized testing data for headers and/or bodies?

jannikluhn:
@holgerd77 Opened an issue [here](https://github.com/ethereumjs/ethereumjs-collation/issues/2).

We don't have any "official" test data yet unfortunately, but it's definitely on the roadmap.

evm/rlp/headers.py
mhchia: My apology that I just noticed this comment. 
`encode_with_padding`, `decode_padded` sounds good to me. `encode/decode_for/from_SMC` is easier for people to realize that this is a SMC specific encoding. I'm fine with both:)

jannikluhn:
@cburgdorf mypy on Travis is failing for this PR while it's passing on my machine:

```
evm/rlp/headers.py:254: error: Invalid type "eth_keys.keys"
evm/rlp/headers.py:257: error: keys.PrivateKey? has no attribute "public_key"
evm/rlp/headers.py:260: error: keys.PrivateKey? has no attribute "sign_msg_hash"
```

I think the reason for this is that the `eth_keys` package is not installed for the mypy run. Is that correct and if so what's the right place to add it? `dev_requirements.txt`?

cburgdorf:
@jannikluhn 

>I think the reason for this is that the eth_keys package is not installed for the mypy run

Nope, the dependency is installed (see the selected text on this screenshot from your travis build)

image

I haven't really looked into `eth_keys` but it appears that `PrivateKey` isn't exported on that level. If you change the import to this

```
from eth_keys.datatypes import (
    PrivateKey
)
```

And then on line `254` you replace `keys.PrivateKey` with `PrivateKey` the error should disappear.

jannikluhn:
Thanks a lot, @cburgdorf, importing `PrivateKey` directly fixed it.

Should be ready for review now.

jannikluhn:
I fully agree that if we add type hints that we should do so consistently. `evm/vm/forks/sharding/collations.py` was not supposed to be there. I moved it to `evm/rlp/collations.py` but missed to clean up afterwards. I deleted it now, thanks for spotting this!

Issue #464, opened 2018-03-14
Need to validate receipts received during fast-sync


gsalgado:
We need to validate the receipts we receive during fast-sync before inserting them in the DB. That was introduced in https://github.com/ethereum/py-evm/pull/451 but we're getting lots of warnings about unexpected receipt roots when syncing from a single geth node, so I'm going to revert that for now, investigate what's going on and either fix our code or the validation

gsalgado:
Closing since I didn't actually revert the change that introduce the checks. See #470 as for why those warnings were being triggered

PR #465, opened 2018-03-14
p2p: Drop receipt root validation during fast sync

gsalgado:
We were getting too many unexpected receipt root warnings when syncing
from a single geth node, so I believe we're doing something wrong.
Created https://github.com/ethereum/py-evm/issues/464 to track this

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -6,10 +6,6 @@ import time
 import traceback
 from typing import Any, Awaitable, Callable, cast, Dict, List, Set, Tuple, Union  # noqa: F401
 
-from eth_utils import (
-    encode_hex,
-)
-
 from evm.constants import EMPTY_UNCLE_HASH
 from evm.db.chain import AsyncChainDB
 from evm.db.trie import make_trie_root_and_nodes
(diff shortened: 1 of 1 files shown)


pipermerriam:
Is this really the right thing to do.  Seems like this points to something systemic that needs to be addressed.  If the receipt roots don't match then it seems like one of the following is happening:

- we're incorrectly calculating the `receipt_root`
- we're getting receipts we didn't ask for.
- we're getting mixed receipts (multiple receipts across different blocks) or maybe extra receipts.

I'm sure there are other possibilities. Any of these though seems like something that needs to be fixed in the *now* term as it will result in a failed sync.  Let me know what I'm missing if anything that makes this something we can kick down the road.

gsalgado:
You're right, but I didn't intend to kick this down the road -- well, maybe a little bit but not too far down the road. I disabled it because it was getting on the way of me debugging the slow down and frequent timeouts I'm seeing, and my plan was to revisit it asap, but I think I've found the cause of the slow down so it's probably better to keep the check and #464 open

Issue #466, opened 2018-03-14
Make use of the `abc` module for base classes


pipermerriam:
### What is wrong?

There are a number of base classes in both the `evm` and `trinity` codebases which implement methods which **must** be implemented when subclassed.

Python provides mechanisms for this pattern which enforce these requirements in the [`abc`](https://docs.python.org/3.5/library/abc.html) module.

### How can it be fixed

Across the various base classes, use the [`abc.abstractmethod`](https://docs.python.org/3.5/library/abc.html#abc.abstractmethod) API on any of the methods which raise `NotImplementedError`.

cburgdorf:
Coming https://github.com/ethereum/py-evm/pull/467

PR #467, opened 2018-03-14
Implement abc.abstractmethod pattern

cburgdorf:
### What was wrong?

Abstract methods did not use the `abc.abstractmethod` decorator (see #466)

### How was it fixed?

Added decorator

#### Cute Animal Picture

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -1,5 +1,11 @@
 import itertools
 
+from abc import (
+    ABCMeta,
+    abstractmethod
+)
+
+
 import rlp
 
 from trie import (
@@ -63,7 +69,7 @@ class TransactionKey(rlp.Serializable):
     ]
 
 
-class BaseChainDB:
+class BaseChainDB(metaclass=ABCMeta):
     trie_class = None
     empty_root_hash = None
 
@@ -93,12 +99,14 @@ class BaseChainDB:
     #
     # Canonical chain API
     #
+    @abstractmethod
     def get_canonical_head(self):
         """
         Returns the current block header at the head of the chain.
         """
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
     def get_canonical_block_header_by_number(self, block_number):
         """
         Returns the block header with the given number in the canonical chain.
@@ -111,6 +119,7 @@ class BaseChainDB:
     #
     # Block Header API
     #
+    @abstractmethod
     def get_block_header_by_hash(self, block_hash):
         """
         Returns the requested block header as specified by block hash.
@@ -119,12 +128,14 @@ class BaseChainDB:
         """
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
     def header_exists(self, block_hash):
         """
         Returns True if the header with the given block hash is in our DB.
         """
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
     def persist_header(self, header):
         """
         :returns: iterable of headers newly on the canonical chain
@@ -133,18 +144,22 @@ class BaseChainDB:
 
     #
     # Block API
+    @abstractmethod
     def lookup_block_hash(self, block_number):
         """
         Return the block hash for the given block number.
         """
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
     def get_block_uncles(self, uncles_hash):
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
     def get_score(self, block_hash):
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
     def persist_block(self, block):
         """
         Chain must do follow-up work to persist transactions to db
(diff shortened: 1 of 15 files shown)


cburgdorf:
Ok, hold off this isn't finished yet. I had to partially roll back some changes and will reintroduce them properly soon.

cburgdorf:
@pipermerriam It's green and ready for another review now. I had to create some Dummy classes for tests that previously instantiated abstract classes that are now forbidden to instantiate.

There's one thing though. I don't fully understand why the `metaclass=ABCMeta` is needed because it also seems to work without that: https://stackoverflow.com/questions/49297254/abstractmethod-works-just-fine-without-metaclass-abcmeta-when-it-shouldnt

evm/db/chain.py
```diff
@@ -93,12 +99,14 @@ def __init__(self, db, account_state_class=MainAccountStateDB, trie_class=Hexary
...
     # Canonical chain API
     #
+    @abstractmethod
     def get_canonical_head(self):
         """
         Returns the current block header at the head of the chain.
         """
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
```
pipermerriam: Note to self/others

[This section of our style guide](https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#avoid-from-x-import-y-for-most-standard-library-imports) recommends that this should be done as

```
import abc

class Thing:
    @abc.abstractmethod
    def some_method(self):
        ...
```

However, I think this specific use case is a good explicit exception to that style as the extra `abc` prefix makes this less readable in my opinion.  Going to add that to the style guide but I thought I'd highlight it here.

cburgdorf:
@pipermerriam so it turns out that **without** `metaclass=ABCMeta` only `mypy` will catch the error whereas with `metaclass=ABCMeta` both `mypy` and `python` will catch the error.

See:

```python

from abc import abstractmethod

class C():
    @abstractmethod
    def my_abstract_method(self):
        print('foo')

class D(C):
    pass

x = C()
y = D()
```

```shell
$ mypy test.py 
test.py:13: error: Cannot instantiate abstract class 'C' with abstract attribute 'my_abstract_method'
test.py:15: error: Cannot instantiate abstract class 'D' with abstract attribute 'my_abstract_method'
```
but 

```shell
$ python3 test.py
$
```

Wheras with that, `python` will catch the error as well.

```python
from abc import ABCMeta, abstractmethod

class C(metaclass=ABCMeta):
    @abstractmethod
    def my_abstract_method(self):
        print('foo')

class D(C):
    pass

x = C()

y = D()
```

```shell
$ python3 test.py 
  File "test.py", line 11, in <module>
    x = C()
TypeError: Can't instantiate abstract class C with abstract methods my_abstract_method
```

evm/db/chain.py
cburgdorf: Thank you for bringing the style guide to my attention!

gsalgado:
This broke `ChainSyncer`. I'm now getting this when I try to run it:
```
19:07:40 ERROR: Unexpected error when handling remote msg: Traceback (most recent call last):                                                                                   
  File "/home/salgado/src/py-evm/p2p/peer.py", line 256, in run                                                                                                                 
    await self.read_loop()                                                                                                                                                      
  File "/home/salgado/src/py-evm/p2p/peer.py", line 300, in read_loop                                                                                                           
  File "/home/salgado/src/py-evm/p2p/peer.py", line 318, in read_msg                                                                                                            
    decoded_msg = cmd.decode(msg)                                                                                                                                               
  File "/home/salgado/src/py-evm/p2p/protocol.py", line 83, in decode                                                                                                           
    return self.decode_payload(data[1:])    
  File "/home/salgado/src/py-evm/p2p/protocol.py", line 70, in decode_payload           
    data = rlp.decode(rlp_data, sedes=decoder)                                          
  File "/home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/rlp-0.6.0-py3.6.egg/rlp/codec.py", line 226, in decode                                                      
    obj = sedes.deserialize(item, **kwargs) 
  File "/home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/rlp-0.6.0-py3.6.egg/rlp/sedes/lists.py", line 126, in deserialize                                           
    result.append(self.element_sedes.deserialize(element))                              
  File "/home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/rlp-0.6.0-py3.6.egg/rlp/sedes/lists.py", line 255, in deserialize                                           
    obj = cls(**dict(list(params.items()) + list(kwargs.items())))                      
TypeError: Can't instantiate abstract class BaseTransaction with abstract methods check_signature_validity, create_unsigned_transaction, get_intrinsic_gas, get_message_for_signing, get_sender        
```

Any idea how I'd go about fixing it? You can see it by running the following:
> python -m p2p.chain -db /tmp/foo

cburgdorf:
Sorry, I'm on mobile now but basically that means that the code tried to
instantiate BaseTransaction which is prohibited now as it is considered an
abstract class that needs to be derived from and have the mentioned methods
overwritten. Is that wrong for BaseTransaction?

pipermerriam:
> Any idea how I'd go about fixing it?

- Quick fix is to remove `metaclass=ABCMeta` from the `BaseTransaction`
- Intermediate fix is to make a `P2PTransaction` class which stubs out all of the un-implemented methods.
- *Correct* fix is to use the proper transaction classes for each `VM` based on the block number the transaction is in.

I'm fine with *any* of these, with the understanding that we probably need to eventually get to where we're using the proper transaction class for each VM.

cburgdorf:
I'd volunteer to make the *correct* fix in case you want to go for a
quicker one meanwhile. However, chances are that I won't be able to work on
it for the next one or two weeks. If that would be too long to wait don't
worry about me.

gsalgado:
@pipermerriam the _correct fix_ is not really possible since the P2P messages that include transactions do not include any info about the block to which they belong. We could look up that information but for that we'd have to make the chaindb available to `p2p.Command`, and even then we'd still need to first deserialize the tx into a non-vm-specific transaction type so that we could generate the trie data and hash it in order to lookup the block to which the transactions belong

pipermerriam:
@gsalgado can't any `rlp.Serializable` also be decoded as a simple list of values?  Would it be possible to replace all of the usages of `BaseTransaction` with `sedes.List` or `sedes.CountableList` (i don't now which is the right one).?

gsalgado:
That should be possible, but using `BaseTransaction` helps ensuring whatever we store in the DB can later be deserialized into the vm-specific transaction types. Another alternative would be to have `P2PTransaction` inherit from `rlp.Serializable` and just copy the fields from `BaseTransaction`, like this:
```
 class P2PTransaction(rlp.Serializable):
    fields = BaseTransaction.fields.copy()
```

PR #468, opened 2018-03-15
p2p: Revert from eth-hash to pysha3 to fix performance issues

gsalgado:
See https://github.com/ethereum/eth-hash/issues/10 for more info

This is what was causing fast-sync to slow to a crawl after running for only a few minutes. With this change we should be able to actually complete a full-node sync

This is intended as a temporary solution while we work on a fix for eth-hash

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,12 +39,12 @@ setup(
         'trinity': [
             "leveldb>=0.194,<1.0.0",
             "coincurve>=7.0.0,<8.0.0",
-            "eth-hash[pycryptodome]>=0.1.0a4,<1.0.0",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -28,6 +28,7 @@
         "trie>=1.3.0,<2.0.0",
         "eth-tester==0.1.0b21",
         "web3>=4.0.0b6,<5.0.0",
+        "pysha3>=1.0.0,<2.0.0",
```
pipermerriam: since all of the usages of this are in the `p2p` code, can this get moved under the `p2p` extras section?

PR #471, opened 2018-03-15
Add support for mainnet syncing to Trinity

pipermerriam:
### What was wrong?

No support for syncing with the mainnet.

### How was it fixed?

Added ability to sync with mainnet via `trinity` cli.  Also modified some internals to use `network_id` as the primary identifier for choosing what network you want to use.

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -25,7 +25,6 @@ from trinity.console import (
     console,
 )
 from trinity.constants import (
-    ROPSTEN,
     SYNC_LIGHT,
 )
 from trinity.db.chain import ChainDBProxy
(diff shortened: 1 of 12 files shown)


trinity/main.py
```diff
@@ -58,22 +57,20 @@
...
-        # TODO: actually use args.sync_mode (--sync-mode)
-        sync_mode = SYNC_LIGHT
-
-    chain_config = ChainConfig.from_parser_args(
-        chain_identifier,
-        args,
-    )
+    logger, log_queue, listener = setup_trinity_logging(args.log_level.upper())
+
+    if args.network_id not in {1, 3}:
```
gsalgado: Any reason not to use the _NETWORK_ID constants here?

gsalgado:
Shall we merge this?

pipermerriam:
> Notes for self

Up next: 

- some form of tests that provide some guarantees that running `trinity` works
- evaluate what RPC endpoints work, which ones don't, which ones won't and ensure they behave in sane ways (raising proper exceptions, returning good error messages, etc).
- evaluate the CLI options compared to `geth/parity` and see which things are clearly missing.
- support for custom genesis parameters (probably integral for testing)
- rework console output to output less information and more useful information.

PR #473, opened 2018-03-16
memory_read and memory_write

murduk:
#### What is wrong?
Currently, the Computation object houses lots of information and sub-APIs for things like maniputlating the stack, consuming gas, etc. This creates a lot of public API surface area since we are effectively support two public APIs for anything needed from a sub-property of the Computation object.

Computation.memory is the API for getting at the Memory
Computation.memory.read is the API for reading from memory.

#### How can it be fixed
Computation.memory -> Computation._memory
Computation.memory.read() -> Computation.memory_read()
Computation.memory.write() -> Computation.memory_write()

#### Cute Animal Picture

### DIFF
--- a/evm/computation.py
+++ b/evm/computation.py
@@ -66,7 +66,7 @@ class BaseComputation(Configurable, metaclass=ABCMeta):
     msg = None
     transaction_context = None
 
-    memory = None
+    _memory = None
     stack = None
     gas_meter = None
 
@@ -92,7 +92,7 @@ class BaseComputation(Configurable, metaclass=ABCMeta):
         self.msg = message
         self.transaction_context = transaction_context
 
-        self.memory = Memory()
+        self._memory = Memory()
         self.stack = Stack()
         self.gas_meter = GasMeter(message.gas)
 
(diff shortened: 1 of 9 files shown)


evm/computation.py
```diff
@@ -133,6 +133,10 @@ def should_return_gas(self):
     def should_erase_return_data(self):
         return self.is_error and self._error.erases_return_data
 
+    @property
+    def memory_bytes(self):
```
pipermerriam: We've tended to stay away from APIs that are only used in testing.  What do you think about just reaching down into `computation._memory.bytes` in the tests and removing this property?

murduk: Sure, I wasn't very comfortable to modify all the tests .

murduk:
@pipermerriam hello, I am not super familiar with github. Do I have to do anything now? Thank you

murduk:
@pipermerriam sorry, I get it now, builds failed. I will fix right away.

pipermerriam:
> I am not super familiar with github. Do I have to do anything now?

Not a problem.  Everyone is new at some point.

Once you open a pull request, all you need to do is continue pushing your changes to your branch and the PR will update itself, trigger new test runs, etc.  When your tests are green/passing please ping me again and I'll give this another review.

Thanks for your contribution!

murduk:
@pipermerriam so sorry for bothering, could you help on fixing the first step of the build? The rest builds without an error. Thank you

pipermerriam:
@murduk see: https://travis-ci.org/ethereum/py-evm/jobs/354548922#L506-L507

You have two minor linting errors that need to be cleaned up.  You should be able to see these locally by running `flake8 evm`.

PR #474, opened 2018-03-16
p2p: Use a separate Transaction class in Commands

gsalgado:
Needed because BaseTransaction now uses metaclass=ABCMeta and thus
cannot be instantiated.

### DIFF
--- a/p2p/rlp.py
+++ b/p2p/rlp.py
@@ -1,0 +1,18 @@
+import rlp
+from rlp import sedes
+
+from evm.rlp.headers import BlockHeader
+from evm.rlp.transactions import BaseTransaction
+
+
+# This is needed because BaseTransaction has several @abstractmethods, which means it can't be
+# instantiated.
+class P2PTransaction(rlp.Serializable):
+    fields = BaseTransaction.fields.copy()
+
+
+class BlockBody(rlp.Serializable):
+    fields = [
+        ('transactions', sedes.CountableList(P2PTransaction)),
+        ('uncles', sedes.CountableList(BlockHeader))
+    ]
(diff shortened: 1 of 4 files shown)


p2p/rlp.py
```diff
@@ -0,0 +1,33 @@
...
+
+from evm.rlp.headers import BlockHeader
+from evm.rlp.transactions import BaseTransaction
+
+
+# This is needed because BaseTransaction has several @abstractmethods, which means it can't be
+# instantiated.
+class P2PTransaction(BaseTransaction):
+
+    def check_signature_validity(self):
```
cburgdorf: Event though these methods doesn't do much I think we should annotate them as `-> None`. I hope that once the entire code base is annotated we will enable checks to require all code to be annotated hence I think we should already try our best to land new code always with annotations.

gsalgado: We'll never implement those methods here, so I don't see the point. Type hints are specially useful when you have non-trivial/non-obvious things -- there's even plenty of people who argue there's no need for type hints when an argument's type is obvious.

But yeah, we'll definitely want to enable as many mypy checks as possible, eventually.

cburgdorf: >We'll never implement those methods here, so I don't see the point

It's about consistency. At some point we could enable `--disallow-untyped-defs` in `mypy` which would basically reject any function to land that isn't typed (with the exception of cases that may need a `type: ignore` because they are too dynamic to type out).

>there's even plenty of people who argue there's no need for type hints when an argument's type is obvious

I see the point, I guess I just value consistency more. Something may be obvious to someone but isn't obvious to someone else so I'd rather stay consistent and just require all function definitions to be typed.

gsalgado:
I've updated this with the suggestion from https://github.com/ethereum/py-evm/pull/467#issuecomment-373805278; please let me know if you're ok with it and I'll merge

PR #477, opened 2018-03-18
Formalizeapiforcomputationobject

murduk:
Added consume_gas function to the computation class


murduk:
I will sync with the master branch then make a pull request

pipermerriam:
If you're not already familiar with it, `git rebase` is the standard workflow for when you need to catch your work up with what's in master.  If you run `git rebase master` in most cases it will *just work*.  You will need to force push after the rebase, `git push -f ....`

PR #478, opened 2018-03-18
Formalizeapiforcomputationobject

murduk:
Added consume_gas method to the computation class


murduk:
Yes, I was going to do it that way. gas_meter is everywhere :)

PR #483, opened 2018-03-18
gas_meter to _gas_meter in computation class

murduk:


evm/logic/system.py
```diff
@@ -125,7 +125,7 @@ def __call__(self, computation):
         call_data = computation.memory_read(start_position, size)
 
         create_msg_gas = self.max_child_gas_modifier(
-            computation.gas_meter.gas_remaining
+            computation._gas_meter.gas_remaining
```
pipermerriam: Shouldn't this be using `computation.get_gas_remaining()` since the intent of this PR is to make the `Computation` object the sole API that is used by all of the opcodes (and to prevent them from reaching into these internal APIs and properties.

murduk: Yes, you are right it should use the getter, but I thought it was left intentionally, I have just changed from gas_meter to_gas_meter.

pipermerriam: Yes, the `computation._gas_meter` should not be accessed anywhere other than from within the `Computation` class and tests.

murduk: Thank you, I have committed the fix

murduk:
Ok, I have changed test to use public api get_gas_remaning() and also refund_gas method of computation class on vm_state files

murduk:
Should I continue with stack? Thank you.

Issue #476, opened 2018-03-17
Look into using black for automated code formatting


pipermerriam:
### What is wrong?

Enforcing code style is tedious.

### How can it be fixed

Look into using https://github.com/ambv/black/

cburgdorf:
I looked into this. I like the idea of enforcing a common style to completely eliminate any debate about formatting. That said `black` is still alpha software so at this point we need to expect that certain formatting rules will still change over time. May not be a huge problem after all, just something to keep in mind. Even though it is alpha state software they do run a safety check by default that ensures that the AST from before the reformatting matches the AST from after the formatting. So at least, it shouldn't break code.

However, the only really configurable rule in `black` is the max-line width (having no options is the main feature after all) which means it is extremely opinionated and may hurt some peoples feelings. I'm absolutely fine to go with `black` if everyone else is fine as well. Just to give you a feel for the changes, here is what running black would do to the code.

https://github.com/cburgdorf/py-evm/commit/d8484e48e15d86e3bf62d2fac4d8ce2af75f34e4

If we do decide that we want to introduce black then it may be a good idea to do a bit of coordination among the team. There probably exist some bigger PRs that would have to go through many merge conflicts if they are forced to rebase on top of the reformatted source code.

So, in case we decide for using `black` I'd recommend we create a PR that prepares the changes to CI and coordinate to make the actual formatting changes after some other changes landed (if needed).

I also looked into the spacing rules that [triggered](https://github.com/ethereum/py-evm/pull/459#discussion_r175168638) this discussion. It turns out that the official style guide [favors to use the spaces around the `=` symbol](https://www.python.org/dev/peps/pep-0008/#variable-annotations) in this case. While I haven't found a way to configure `flake8` to enforce this rule `pylint` can enforce to have the spaces around the `=` symbol (but not the opposite!). Also `black` enforces this so that the code in question would be formatted as.

```python

    def __init__(
        self,
        shard_id: int,
        expected_period_number: int,
        period_start_prevhash: bytes,
        parent_hash: bytes,
        number: int,
        transaction_root: bytes = EMPTY_SHA3,
        coinbase: bytes = ZERO_ADDRESS,
        state_root: bytes = EMPTY_SHA3,
        receipt_root: bytes = EMPTY_SHA3,
        sig: bytes = b'',
    ) -> None:
        super(CollationHeader, self).__init__(
            shard_id=shard_id,
            expected_period_number=expected_period_number,
[...]

PR #484, opened 2018-03-19
Enable mypy checks for entire evm module

cburgdorf:
### What was wrong?

So far, the `evm` modules was only partially checked with mypy since there were still many errors when running `mypy` over the entire `evm` module.

### How was it fixed?

This PR adds type hints only **selectively** for all the cases that are currently preventing us to run `mypy` on the entire `evm` module. It does **not add** all missing type annotations per file yet.

This in contrast to my previous PR which tried to add (more or less) all type hints per subdirectory in `evm`.

I will continue to send subsequent PRs to eventually have the entire code base typed but I believe it is a big when to be able to activate the `mypy` check for the entire `evm` module so that we can ensure new code already conforms to these rules.

#### Cute Animal Picture


cburgdorf:
@pipermerriam @gsalgado This should be ready for review

evm/utils/datatypes.py
```diff
@@ -55,6 +55,11 @@ def _get_top_level_keys(overrides: Dict[str, Any]) -> Iterator[str]:
     for prop in overrides:
         yield _extract_top_level_key(prop)
 
+# Dynamic subclassing is not supported by mypy
+# https://github.com/python/mypy/wiki/Unsupported-Python-Features
+# Most of the cases where we silence mypy boil down to cases where
+# dynamic subclasses where generated through this method
+
```
gsalgado: I'm not sure I like the idea of having a blanket comment like this; I'd prefer if we had a separate, specific, comment for every place we need to silence mypy

evm/rlp/collations.py
```diff
@@ -14,6 +14,8 @@
 from .transactions import BaseShardingTransaction
 from .headers import CollationHeader
 
+from typing import Type
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/rlp/blocks.py
```diff
@@ -14,6 +14,8 @@
 from .transactions import BaseTransaction
 from .headers import BlockHeader
 
+from typing import Type  # noqa: F401
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/logic/logging.py
```diff
@@ -4,6 +4,8 @@
 
 from evm import constants
 
+from typing import List  # noqa: F401
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/db/journal.py
```diff
@@ -7,6 +7,8 @@
 from evm.db.backends.base import BaseDB
 from evm.exceptions import ValidationError
 
+from typing import Dict  # noqa: F401
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/computation.py
```diff
@@ -48,6 +48,18 @@
     Stack,
 )
 
+from evm.opcode import (  # noqa: F401
+    Opcode
+)
+
+from typing import (  # noqa: F401
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/chains/tester/__init__.py
```diff
@@ -19,33 +19,46 @@
     generate_vms_by_range,
 )
 
+from evm.rlp.headers import (
+    BlockHeader
+)
+
+from typing import (
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/__init__.py
```diff
@@ -21,8 +21,8 @@
 #  Setup TRACE level logging.
 #
 logging.addLevelName(TRACE_LEVEL_NUM, 'TRACE')
-logging.TRACE = TRACE_LEVEL_NUM
-logging.Logger.trace = trace
+logging.TRACE = TRACE_LEVEL_NUM  # type: ignore
+logging.Logger.trace = trace  # type: ignore
```
gsalgado: Why do we have to ignore those?

evm/vm/code_stream.py
```diff
@@ -7,43 +7,48 @@
     validate_is_bytes,
 )
 
+from typing import (  # noqa: F401
+    Iterator,
+    Set
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/vm/forks/frontier/blocks.py
```diff
@@ -28,6 +28,10 @@
     FrontierTransaction,
 )
 
+from typing import (  # noqa: F401
+    List
+)
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/vm/forks/frontier/vm_state.py
```diff
@@ -40,9 +40,14 @@
...
 from .computation import FrontierComputation
 from .constants import REFUND_SELFDESTRUCT
-from .transaction_context import FrontierTransactionContext
+from .transaction_context import (  # noqa: F401
+    BaseTransactionContext,
+    FrontierTransactionContext
+)
 from .validation import validate_frontier_transaction
 
+from typing import Type  # noqa: F401
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -32,6 +32,10 @@
     get_sharding_config,
 )
 
+from typing import (  # noqa: F401
+    Dict
+)
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/vm_state.py
```diff
@@ -22,6 +22,19 @@
...
+from evm.computation import (  # noqa: F401
+    BaseComputation,
+)
+from evm.transaction_context import (  # noqa: F401
+    BaseTransactionContext,
+)
+
+from typing import (  # noqa: F401
+    Type
+)
```
gsalgado: This should be at the top of the file together with other stdlib imports

evm/__init__.py
cburgdorf: Because these properties do not exist and we are making them up out of thin air. I'll put a comment on it.

evm/chains/tester/__init__.py
cburgdorf: Alright, will move all of them to the top

evm/utils/datatypes.py
cburgdorf: I've put a short comment on the places where we had to add a `ignore` because of this and then referenced to here to lookup this longer explanation. I thought that's a good trade off so that I don't have to put a lengthy explanation on each `ignore`

cburgdorf:
@gsalgado Thank you for the review. I moved all the typing imports to the top and added a comment at the place where we are making up those two properties on the logging module.

The other `ignore` places already had comments attached and I also left the longer explainer comment on the `Configurable` class. Is that alright?

evm/vm/forks/sharding/vmc_handler.py
```diff
@@ -1,3 +1,6 @@
...
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/vm_state.py
```diff
@@ -1,3 +1,7 @@
...
+
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/vm/code_stream.py
```diff
@@ -1,3 +1,7 @@
...
+)
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/tools/test_builder/test_builder.py
```diff
@@ -1,3 +1,8 @@
+from typing import (  # noqa: F401
+    Dict,
+    List,
+    Any
+)
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/rlp/blocks.py
```diff
@@ -1,3 +1,4 @@
...
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/logic/logging.py
```diff
@@ -1,5 +1,5 @@
 from __future__ import absolute_import
-
+from typing import List  # noqa: F401
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/db/journal.py
```diff
@@ -1,3 +1,5 @@
...
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

evm/computation.py
```diff
@@ -7,6 +7,14 @@
     abstractmethod
 )
 
+from typing import (  # noqa: F401
```
gsalgado: There should be no blank lines between stdlib imports, and they should be ordered alphabetically

cburgdorf:
@gsalgado updated! Now this makes me even more supportive of switching to `black` ;)

evm/db/journal.py
pipermerriam: @carver implemented automated import sorting using [`isort`](https://github.com/timothycrosley/isort).  That's like a nice half-step towards what `black` does.

carver: Yup, I've been happy with `isort`. Here are the configs we put in `tox.ini`, which should be close enough to the py-evm standards to drop in place with only a few edits:

https://github.com/ethereum/web3.py/blob/307dc879f18bec9cc0071fc2a6205c1d54b985f5/tox.ini#L9-L19

```
[isort]
combine_as_imports=True
force_sort_within_sections=True
include_trailing_comma=True
known_standard_library=pytest
known_third_party=lru,eth_tester  # customize
known_first_party=web3,ens  # customize
line_length=21
multi_line_output=3
skip=web3/main.py,web3/utils/windows.py  # customize
use_parentheses=True
```

cburgdorf: Thanks for sharing @carver! I could give that a try. We also have a [tracking issue](https://github.com/ethereum/py-evm/issues/476) regarding black. I think moving to black would be my favorite option but it would be great to get more opinions on it :)

Issue #362, opened 2018-02-08
Formalize API for Computation object.


pipermerriam:
### What is wrong?

Currently, the `Computation` object houses *lots* of information and sub-APIs for things like maniputlating the stack, consuming gas, etc.  This creates a lot of public API surface area since we are effectively support *two* public APIs for anything needed from a sub-property of the `Computation` object.

- `Computation.memory` is the API for getting at the `Memory`
- `Computation.memory.read` is the API for reading from memory.

### How can it be fixed

I propose moving **all** sub-APIs to private properties and only exposing them through a single public API on the `Computation` object itself.

For example.

- `Computation.memory -> Computation._memory`
- `Computation.memory.read() -> Computation.memory_read()`

In the process we should organize the body of the `Computation` object by subject/sub-API meaning that we co-locate all of the *memory* APIs together, and so on for each category.

#### Maybe use abbreviated suffixes

Some of these accessors can be quite long:

- `Computation.gas_meter.consume_gas`

Maybe it's ok for us to shorten these using a standard abbreviated prefix.  This trades one type of readability for another and since these are reasonably *low level* APIs that are only consumed from within EVM execution, I think that I'm comfortable with this trade-off.

- `Computation.gm_consume_gas()`

#### Maybe create a common base class

It might also be good to create a `BaseComputation` class which contains stubs for all of the public APIs which raise `NotImplementedError` exceptions.  This serves as a sort of documentation from within the code for the API for the computation object.

A possible argument against this is that the `sharding` changes may introduce new APIs which are not used pre-sharding, and in some cases, potentially removing APIs which are no longer part of the sharding EVM.

carver:
👍  Occasionally it's nice to be able to `dir()` at the repl, so `dir(computation)` would be much nicer this way.

> `Computation.gm_consume_gas()`

Maybe this could even just be `Computation.consume_gas()`. It seems unambiguously named. The fact that it's called gas meter seems like an unnecessary implementation detail.

The other main modules seem to work follow the memory example nicely:
- stack
- memory
- code

Like: `Computation.stack_pop()`

> A possible argument against this is that the sharding changes may introduce new APIs which are not used pre-sharding, and in some cases, potentially removing APIs which are no longer part of the sharding EVM.

Seems like BaseComputation should just shrink, then, to the intersection of the APIs. I imagine there would still be a useful subset of methods to document.

vs77bb:
[...]

PR #486, opened 2018-03-19
making stack of computation class private

murduk:
What was wrong?
Computation class was exposing all private members. Other classes were accessing private members and then calling methods from these members.  #362 

How was it fixed?
Methods implemented on computation class to hide private members. Now other classes are calling these methods instead of accessing the members.

Cute Animal Picture
23ac461a-211e-42a4-bcc1-1bc56e5dc447

### DIFF
--- a/evm/computation.py
+++ b/evm/computation.py
@@ -201,6 +201,12 @@ class BaseComputation(Configurable, metaclass=ABCMeta):
     def consume_gas(self, amount, reason):
         return self.gas_meter.consume_gas(amount, reason)
 
+    def return_gas(self, amount):
+        return self.gas_meter.return_gas(amount)
+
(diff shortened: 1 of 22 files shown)


pipermerriam:
Even though it's pretty obvious to us two what these pull requests are part of, can you please still fill them in with a full description and link to the original master issue.

Also, where is my [cute animal picture](https://github.com/ethereum/py-evm/blob/master/.github/PULL_REQUEST_TEMPLATE.md#cute-animal-picture) 😄

murduk:
:))) Sorry about that. I have filled the documentation, and took the opportunity to upload my cute cat's photo :)

pipermerriam:
Excellent cat-in-a-bag!

PR #488, opened 2018-03-21
Add more type hints

cburgdorf:
### What was wrong?

Missing type hints

### How was it fixed?

This adds type hints for `evm/computation` and some related classes. 

However, apart from just adding type hints, this introduces some supportive changes.

1. Instead of mutating `Logger` to add a `trace` method it introduces a `TraceLogger` and calls the `logging.setLoggerClass(...)` API to register it.

2. At places where we request a logger via `logging.getLogger('some.logger')` we do a `cast(TraceLogger, logging.getLogger('some.logger')` to give `mypy` a hint about the type (a `cast` doesn't do anything apart from hinting `mypy` what to assume about the type).

3. All the wiring of the trace logging was moved to a `setup_trace_logging` API in `evm.utils.logging`

4. I still left the `logging.TRACE = TRACE_LEVEL_NUM` but I'm actually tempted to get rid of that as well and just import `TRACE_LEVEL_NUM` where we currently use `logging.TRACE` to get rid of that `type: ignore` as well. Thoughts?

5. `mypy` forbids the usage of the variable name `bytes` in `Memory` so I changed it to `_bytes` because I also believe treating it as *private* is aligned with our intentions.

Btw, totally unrelated but as someone who is new to the code base I must say that these type hints make it so much easier for me to understand the code and see what is going on in these APIs.

#### Cute Animal Picture

### DIFF
--- a/evm/computation.py
+++ b/evm/computation.py
@@ -8,7 +8,9 @@ import logging
 from typing import (  # noqa: F401
     Any,
     Callable,
+    cast,
     Dict,
+    Iterator,
     List,
     Tuple,
 )
@@ -33,6 +35,9 @@ from evm.utils.hexadecimal import (
 from evm.utils.numeric import (
     ceil32,
 )
+from evm.utils.logging import (
+    TraceLogger
+)
 from evm.validation import (
     validate_canonical_address,
     validate_is_bytes,
@@ -53,13 +58,21 @@ from evm.vm.message import (
 from evm.vm.stack import (
     Stack,
 )
-
+from evm.vm_state import (
+    BaseVMState,
+)
 from evm.opcode import (  # noqa: F401
     Opcode
 )
+from evm.db.state import (
+    BaseAccountStateDB
+)
+from evm.transaction_context import (
+    BaseTransactionContext
+)
 
 
-def memory_gas_cost(size_in_bytes):
+def memory_gas_cost(size_in_bytes: int) -> int:
     size_in_words = ceil32(size_in_bytes) // 32
     linear_cost = size_in_words * GAS_MEMORY
     quadratic_cost = size_in_words ** 2 // GAS_MEMORY_QUADRATIC_DENOMINATOR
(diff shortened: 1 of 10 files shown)


cburgdorf:
@pipermerriam @gsalgado This should be ready for review. 

Personally, I'd like to get rid of that `logging.TRACE` mutation (see point 4 above) so that we can get rid of yet another `type: ignore` here but I like to hear your thoughts about it before.

cburgdorf:
@pipermerriam Alright, updated!

evm/computation.py
```diff
@@ -448,7 +473,7 @@ def apply_computation(cls, vm_state, message, transaction_context):
     # Opcode API
     #
     @property
-    def precompiles(self):
+    def precompiles(self) -> Dict[bytes, Callable]:
```
pipermerriam: I think this can be made more specific as `Callable[[Computation], None]` or `Callable[[Computation], Any]`.  I think the later with `Any` is fine to **allow** opcodes to return values but not require any kind of return type or value.

cburgdorf: Good point. Updated the PR. There are actually two other annotations in here that are annotated as just `Tuple` that could probably be more specific. I haven't had time to follow up on them yet so if you know the types from the top of your head I'm happy to update them as well.

If not we can either:

- keep this PR open until I updated them
- merge now and update them with a subsequent PR

The cool thing is that as more and more type annotations get added it becomes easier and easier for `mypy` to assist you. I already updated a couple of type annotations that were previously wrong because as you put in more parts of the puzzle `mypy` becomes stronger and stronger to infer the correct types and eventually things converge to the correct types.

cburgdorf: Ok, I updated the two APIs that where previously typed to return just `Tuple` to now be.

```
def get_accounts_for_deletion(self) -> Tuple[Tuple[bytes, bytes], ...]:
```

and 

```
def get_log_entries(self) -> Tuple[Tuple[bytes, List[int], bytes], ...]:
```

pipermerriam:
just in case you didn't notice it.  These'r a mypy failure in CI: https://travis-ci.org/ethereum/py-evm/jobs/358017797

cburgdorf:
>just in case you didn't notice it. These'r a mypy failure in CI: https://travis-ci.org/ethereum/py-evm/jobs/358017797

just returned home after a 6 hour drive.

So the reason this fails is because it turns out that our CI already runs on `mypy 0.580` (even though we didn't explicitly update it yet)

image

Our current CI jobs can't deal with `0.580` yet. The good news is that there's already https://github.com/ethereum/py-evm/pull/498 to fix that.

So, if you merge #498 I can rebase this one and it should go green again.

pipermerriam:
> So, if you merge #498 I can rebase this one and it should go green again.

Done

cburgdorf:
>Done

Rebased & pushed ;)

PR #493, opened 2018-03-22
p2p: Redesign handling of timeouts in ChainSyncer

gsalgado:
We were seeing timeouts too frequently and because of that I'd introduced
an eager retry mechanism, but it turns out those timeouts were because we
had too many cpu-intensive tasks that were blocking the main thread and
thus preventing the asyncio main loop from processing received msgs fast
enough. Now that those cpu-intensive tasks are performed in sub-processes,
we rarely see timeouts and in those cases it generally means the peer has
disconnected so instead of retrying we simply abort the sync with that
peer, which causes it to be resumed with a different peer.

Depends on #492

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -3,10 +3,13 @@ import logging
 import operator
 import time
 import traceback
-from typing import Any, Awaitable, Callable, cast, Dict, List, Set, Tuple, Union  # noqa: F401
+from typing import Any, Awaitable, Callable, cast, Dict, Generator, List, Set, Tuple  # noqa: F401
+
+from cytoolz.itertoolz import unique
 
 from eth_utils import (
     encode_hex,
+    to_list,
 )
 
 from evm.constants import EMPTY_UNCLE_HASH
@@ -170,7 +173,7 @@ class ChainSyncer(PeerPoolSubscriber):
 
     async def _downloader(self,
                           queue: 'asyncio.Queue[List[BlockHeader]]',
-                          should_skip: Callable[[BlockHeader], bool],
+                          filter_func: Callable[[List[BlockHeader]], List[BlockHeader]],
                           request_func: Callable[[List[BlockHeader]], Awaitable[None]],
                           pending: Dict[Any, Tuple[BlockHeader, float]],
                           batch_size: int) -> None:
@@ -181,6 +184,7 @@ class ChainSyncer(PeerPoolSubscriber):
                     queue.get(),
                     token=self.cancel_token,
                     timeout=self._reply_timeout)
+                batch.extend(headers)
             except TimeoutError:
                 # We use a timeout above to make sure we periodically retry timedout items
                 # even when there's no new items coming through.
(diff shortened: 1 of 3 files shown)


p2p/chain.py
```diff
@@ -209,29 +215,36 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            'receipts')
 
-    def _should_skip_body_download(self, header: BlockHeader) -> bool:
-        return (header.transaction_root == self.chaindb.empty_root_hash and
-                header.uncles_hash == EMPTY_UNCLE_HASH)
+    def _skip_empty_bodies(self, headers: List[BlockHeader]) -> List[BlockHeader]:
+        return [
+            header
+            for header in headers
+            if (header.transaction_root != self.chaindb.empty_root_hash or
```
pipermerriam: Probably improved readability if you moved this out of a comprehension and instead used the `to_tuple` decorator to do this in a proper for loop.  It would allow this logic to be broken across multiple lines in an easier to read manner.

pipermerriam: Probably improved readability if you moved this out of a comprehension and instead used the `to_tuple` decorator to do this in a proper for loop.  It would allow this logic to be broken across multiple lines in an easier to read manner.

p2p/chain.py
```diff
@@ -242,8 +255,16 @@ def _should_skip_body_download(self, header: BlockHeader) -> bool:
...
             self._pending_bodies[key] = (header, now)
 
-    def _should_skip_receipts_download(self, header: BlockHeader) -> bool:
-        return header.receipt_root == self.chaindb.empty_root_hash
+    def _skip_empty_and_duplicated_receipts(self, headers: List[BlockHeader]) -> List[BlockHeader]:
+        # Post-Byzantium blocks may have identical receipt roots (e.g. when they have the same
+        # number of transactions and all succeed/failed: ropsten blocks 2503212 and 2503284), so
+        # we have an extra check here to avoid requesting those receipts multiple times.
+        headers = list(unique(headers, key=operator.attrgetter('receipt_root')))
+        return [
```
pipermerriam: Same here about moving this out of a list comprehension and using `to_tuple` to allow expansion of this logic across multiple lines within a `for` loop.

p2p/chain.py
```diff
@@ -331,53 +352,43 @@ def _test() -> None:
...
-        listen_port = 30303
-        addr = kademlia.Address(listen_host, listen_port, listen_port)
-        discovery = DiscoveryProtocol(privkey, addr, ROPSTEN_BOOTNODES)
-        loop.run_until_complete(discovery.create_endpoint())
-        print("Bootstrapping discovery service...")
-        loop.run_until_complete(discovery.bootstrap())
-        peer_pool = PeerPool(ETHPeer, chaindb, RopstenChain.network_id, privkey, discovery)
+        from p2p.peer import HardCodedNodesPeerPool
+        peer_pool = HardCodedNodesPeerPool(ETHPeer, chaindb, RopstenChain.network_id, privkey)
+        peer_pool.min_peers = 5
```
pipermerriam: Given this monkeypatch (and the known need for this to be configurable) can this go ahead and get moved to something we can set in the `__init__`?

p2p/chain.py
gsalgado: Makes sense. Doing that on #492 though as that's where it was introduced

p2p/chain.py
gsalgado: Makes sense. Doing that on #492 though as that's where it was introduced

PR #494, opened 2018-03-22
p2p: When fast syncing, don't send multiple simultaneous body/receipt requests to a single peer

gsalgado:


p2p/chain.py
```diff
@@ -246,13 +249,28 @@ def _should_skip_receipts_download(self, header: BlockHeader) -> bool:
...
+    async def get_idle_peer(self) -> ETHPeer:
+        """Return a random peer which we're not already expecting a response from."""
+        while True:
+            idle_peers = [
+                peer
+                for peer in self.peer_pool.peers
+                if peer not in self._peers_with_pending_requests
+            ]
+            if idle_peers:
+                return cast(ETHPeer, random.choice(idle_peers))
```
pipermerriam: This is a good candidate for `secrets.choice`

gsalgado: I thought about that too, but couldn't think of any reasons why it'd be preferred over plain `random`. Can you think of any?

pipermerriam: Yes, the `random` module provides no guarantees about providing good sources of randomness and has explicit warnings at the top of the module that the functions should not be used for "security purposes".  The `secrets` module uses secure sources of randomness by default which I don't believe is the case for `random`.

gsalgado: Right, but I don't think we need true randomness here -- this is just a way to pick, arbitrarily, one of our connected peers. I believe where we'd need true randomness is where we chose which peers we connect to.

pipermerriam: I would *strongly* prefer we use strong sources of randomness.  I know there's not much of an attack vector here, but it's cheap to use the *better* version of `choice`.

gsalgado: Fair enough

PR #496, opened 2018-03-23
Start writing p2p.Server

njgheorghita:
@gsalgado would love some feedback on this. My first time using the asyncio library - so all suggestions on how to improve this pr would be great!!

### What was wrong?
`p2p.Server` is needed to listen to incoming connections

### How was it fixed?
Started writing the `Server` class

#### Cute Animal Picture
image


tests/p2p/test_server.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def send_sub_proto_handshake(self):
+        pass
+
+@pytest.fixture
+def server():
+    privkey = keys.PrivateKey(test_values['receiver_private_key'])
+    return Server(privkey)
+
+# @pytest.mark.asyncio
```
njgheorghita: what's the best way to unit test `receive_handshake`? I got things working initially by setting some variables in the function - but that's obviously not a longterm solution

p2p/server.py
```diff
@@ -0,0 +1,125 @@
...
+        
+        while not self.cancel_token.triggered:
+            try:
+                loop.run_forever()
+            except:
+                self.logger.error("Unexpected error.")
+            await awyncio.wait([self.cancel_token.wait()], timeout=_connect_loop_sleep)
+
+    async def receive_handshake(self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        # Use reader to read the auth_init msg !!!CONSTANT???!!!
```
njgheorghita: what constant should I be using here? I got `437` from the length of `auth_init_ciphertext` - but none of the existing constants matched that length

p2p/server.py
```diff
@@ -0,0 +1,125 @@
...
+        # Get remote's address 
+        ip, udp, _, _ = writer.get_extra_info("peername")
+        remote_address = Address(ip, udp)
+
+        # Create a `HandshakeResponder(remote: kademlia.Node, privkey: datatypes.PrivateKey)` instance
+        initiator_remote = Node(initiator_pubkey, remote_address)
+        responder = HandshakeResponder(initiator_remote, self.privkey)
+
+        # Call `HandshakeResponder.create_auth_ack_message(nonce: bytes)` to create the reply
+        responder_nonce = keccak(os.urandom(HASH_LEN))
```
pipermerriam: we recently realized we can use the new [`secrets.token_bytes`](https://docs.python.org/3/library/secrets.html#secrets.token_bytes) functions for things like this.

p2p/server.py
```diff
@@ -0,0 +1,125 @@
...
+class Server:
+    """Server listening for incoming connections"""
+    logger = logging.getLogger("p2p.server.Server")
+    _server_address = ('localhost', 10000)
+    _connect_loop_sleep = 2
+    
+    def __init__(self, privkey: datatypes.PrivateKey) -> None:
+        self.privkey = privkey
+        self.cancel_token = CancelToken('Server')
+        self.incoming_connections = []
```
pipermerriam: I think that instead of this `list` based storage we'll want the `Server` to be given a `PeerPool` that it can push new peers into.

tests/p2p/test_server.py
```diff
@@ -0,0 +1,130 @@
...
+    "initiator_ephemeral_private_key": "869d6ecf5211f1cc60418a13b9d870b22959d0c16f02bec714c960dd2298a32d",
+    "receiver_ephemeral_private_key": "e238eb8e04fee6511ab04c6dd3c89ce097b11f25d584863ac2b6d5b35b1847e4",
+    "initiator_nonce": "7e968bba13b6c50e2c4cd7f241cc0d64d1ac25c7f5952df231ac6a2bda8ee5d6",
+    "receiver_nonce": "559aead08264d5795d3909718cdd05abd49572e84fe55590eef31a88a08fdffd",
+    "mac_secret": "2ea74ec5dae199227dff1af715362700e989d889d7a493cb0639691efb8e5f98",
+    "aes_secret": "80e8632c05fed6fc2a13b0f8d31a3cf645366239170ea067065aba8e28bac487",
+}
+
+for k, v in test_values.items():
+    test_values[k] = decode_hex(v)
```
pipermerriam: This is odd to me.  Can we just embed the `decode_hex` calls in the declaration above?

tests/p2p/test_server.py
```diff
@@ -0,0 +1,130 @@
...
+        # (object,),
+        # {"write": responder_reader.feed_data}
+    # )
+
+    # responder_reader.feed_data(AUTH_INIT_CIPHERTEXT)
+    # aes, mac = await asyncio.ensure_future(server.receive_handshake(responder_reader, responder_writer))
+    # assert aes == test_values['aes_secret']
+
+
+SERVER_ADDRESS = ('localhost', 10000)
```
pipermerriam: In tests, it's good to use a random port to ensure that concurrent test runs don't fail due to the port being taken.  If you do some grepping you'll find some code somewhere here or maybe in populus for finding a random port which is open.

tests/p2p/test_server.py
```diff
@@ -0,0 +1,130 @@
...
+
+def test_responder_server(server):
+    # Start server
+    loop = asyncio.get_event_loop()
```
pipermerriam: There is an `event_loop` fixture` that should be usable instead of manual instantiation.

p2p/server.py
```diff
@@ -0,0 +1,125 @@
...
+        loop = asyncio.get_event_loop()
+        factory = asyncio.start_server(self.receive_handshake, *self._server_address)
+        asyncio.ensure_future(factory)
+        
+        while not self.cancel_token.triggered:
+            try:
+                loop.run_forever()
+            except:
+                self.logger.error("Unexpected error.")
+            await awyncio.wait([self.cancel_token.wait()], timeout=_connect_loop_sleep)
```
gsalgado: This loop doesn't feel right to me; can you maybe describe what you were trying to achieve?

p2p/server.py
```diff
@@ -0,0 +1,125 @@
...
+    
+    def __init__(self, privkey: datatypes.PrivateKey) -> None:
+        self.privkey = privkey
+        self.cancel_token = CancelToken('Server')
+        self.incoming_connections = []
+
+    async def stop(self) -> None:
+        self.logger.info("Closing server...")
+        self.cancel_token.trigger()
+        await asyncio.sleep(1)
```
gsalgado: Why the sleep here?

p2p/server.py
gsalgado: I believe that's because your test is dealing with a eip8 auth msg, in which case the length is encoded in plaintext on the first two bytes. See `decode_auth_eip8()` and https://github.com/ethereum/EIPs/blob/master/EIPS/eip-8.md

tests/p2p/test_server.py
gsalgado: Did you have a look at `test_handshake()`? It uses known-good values to ensure that our implementation matches the spec, and then uses the derived secrets to create two peer instances and talk to each other. I believe you could do something similar, no?

p2p/server.py
```diff
@@ -0,0 +1,125 @@
...
+            auth_init_ciphertext=msg,
+            auth_ack_ciphertext=auth_ack_ciphertext
+        )
+
+        # Store peer creation data in incoming_connections[]
+        peer = (aes_secret, mac_secret)
+        self.incoming_connections.append(peer)
+
+
+def _decode_authentication(ciphertext: bytes,
```
gsalgado: This was just converted from a `HandshakeResponder()` method into a standalone function, right? In that case, we should delete the method

p2p/server.py
njgheorghita: @pipermerriam The list is a temp hack - It looks like implementing `registerPeer` is not insignificant i thought it would be better off in a following pr - to avoid this pr growing too large? I've added a peer_pool to the `Server` object. If you think the `registerPeer` logic would be better implemented in this pr - i can go ahead and do that

p2p/server.py
njgheorghita: It's not exactly the same - this version has an extra `privkey` arg, since it doesn't have a `HandshakeResponder` object to grab that private key from -- Does that warrant it being left here?

gsalgado: Even if we want to keep the method there, it should be changed to call this function instead of us maintaining two nearly identical copies of that code. But I believe we can just change the tests that use the `HandshakeResponder` method to use this function instead -- I think that's the only place this is used if it is used at all

p2p/server.py
```diff
@@ -0,0 +1,137 @@
...
+        self._server_address = server_addr
+
+    def register_peer(self, peer: ETHPeer) -> None:
+        pass
+
+    async def stop(self) -> None:
+        self.logger.info("Closing server...")
+        self.cancel_token.trigger()
+        self.peer_pool.unsubscribe(self)
+        await asyncio.sleep(1)
```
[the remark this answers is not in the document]
gsalgado: What do you mean by "once PeerPool gets implemented"?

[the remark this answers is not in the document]
njgheorghita: Sorry - i think that context got buried in a comment somewhere. Storing the peer info in the `incoming_connections` list was a temporary hack. I was thinking of splitting this pr into 2. First, to set up the server and store peer info in the list. A following pr would actually implement `register_peer` and `handle_peer` logic. The goal is to keep this pr from getting too unwieldy - unless you think it's a better idea to implement `PeerPool` and it's logic all at once?

p2p/server.py
```diff
@@ -0,0 +1,156 @@
...
+        self._running_peers = set()
+        self.privkey = privkey
+        self._server_address = server_addr
+
+    def register_peer(self, peer: BasePeer) -> None:
+        asyncio.ensure_future(self.handle_peer(cast(ETHPeer, peer)))
+
+    async def handle_peer(self, peer: ETHPeer) -> None:
+        """Handle the lifecycle of the given peer."""
+        self._running_peers.add(peer)
```
njgheorghita: @gsalgado Am I missing some steps here that are needed to add a peer to the peer_pool?

p2p/server.py
```diff
@@ -0,0 +1,156 @@
...
+        while True:
+            try:
+                cmd, msg = await peer.read_sub_proto_msg(self.cancel_token)
+            except OperationCancelled:
+                break
+
+    async def stop(self) -> None:
+        self.logger.info("Closing server...")
+        self.cancel_token.trigger()
+        self.peer_pool.unsubscribe(self)
```
njgheorghita: Is it necessary to stop the peer_pool here as well? Something like `peer_pool.stop()`?

p2p/server.py
```diff
@@ -0,0 +1,156 @@
...
+            responder_nonce=responder_nonce,
+            remote_ephemeral_pubkey=ephem_pubkey,
+            auth_init_ciphertext=msg,
+            auth_ack_ciphertext=auth_ack_ciphertext
+        )
+
+        # Create and register peer in peer_pool
+        peer = (aes_secret, mac_secret, initiator_nonce, ephem_pubkey)
+        eth_peer = ETHPeer(
+                remote=initiator_remote, privkey=None, reader=None,
```
njgheorghita: @gsalgado I'm confused as to how we instantiate this peer to pass to the peer_Pool. How do we know the peer's privkey?

tests/p2p/test_server.py
```diff
@@ -0,0 +1,126 @@
...
+@pytest.fixture
+def server():
+    chaindb = ChainDB(MemoryDB())
+    privkey = keys.PrivateKey(test_values['receiver_private_key'])
+    disc = discovery.DiscoveryProtocol(privkey, random_address(), bootstrap_nodes=[])
+    peer_pool = PeerPool(ETHPeer, chaindb, 1, test_values['receiver_private_key'], disc)
+
+    return Server(privkey, SERVER_ADDRESS, peer_pool)
+
+def test_responder_server(server, event_loop):
```
njgheorghita: @gsalgado this test is still a mess, but I'll tighten it up once I get a better idea of how the peer_pool and server are working together

p2p/server.py
pipermerriam: @njgheorghita I *think* that the `PeerPool` class is going to need a new API for this.  It looks like the peer needs to be added to `PeerPool.connected_nodes`, however, we should not directly push the peer into that data structure.  You can probably do something as simple as:

```python
class PeerPool:
    ...
    def add_peer(self, peer):
        for subscriber in self._subscribers:
            subscriber.register_peer(peer)
```

Then, from within your `receive_handshake` method, instead of calling `self.register_peer(...)` you would just instead call `peer_pool.add_peer(peer)` with the knowledge that it will *subscribe* you to the peer.  Similarly, you'll need to implement `Server.remove` to handle the other call from the peerpool when the peer is removed.

p2p/server.py
pipermerriam: I don't believe so, I think the assumption is that the peer pool runs at a higher level than this process.

p2p/server.py
pipermerriam: We don't.  It looks like we're supposed to use our own `privkey` here.  cc @gsalgado to confirm.

p2p/server.py
```diff
@@ -104,13 +117,18 @@ def register_peer(self, peer: ETHPeer) -> None:
             auth_ack_ciphertext=auth_ack_ciphertext
         )
 
-        # Store peer creation data in incoming_connections[]
-        # TODO register peer in PeerPool
+        # Create and register peer in peer_pool
         peer = (aes_secret, mac_secret, initiator_nonce, ephem_pubkey)
+        eth_peer = ETHPeer(
+                remote=initiator_remote, privkey=None, reader=None,
+                writer=None, aes_secret=aes_secret, mac_secret=mac_secret,
```
pipermerriam: Is there a reason `reader` and `writer` are set to `None`?  Seems like the local `reader` and `writer` instances should be used here.

p2p/server.py
gsalgado: Actually, my current thinking is that `Server` represents the node's main process and manages all services needed to run a full node (e.g. finding/connecting-to remote peers, accepting incoming connections, syncing the chain, etc), so yes it would be responsible for starting/stopping `PeerPool`

p2p/server.py
gsalgado: ISTM that ETHPeer doesn't use the `privkey` anymore, so we should just remove it

p2p/server.py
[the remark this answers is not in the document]
gsalgado: I think that makes sense, yeah

p2p/server.py
njgheorghita: @gsalgado Looks that way to me, should I go ahead and remove the `privkey` arg from the BasePeer class?

njgheorghita: Col, I'll just use the available `privkey` here for now and then in a following pr remove the `privkey` arg from BasePeer class, if that sounds good?

gsalgado: It is actually used in `P2PProtocol.send_handshake()`, so yeah, just pass the pool's privkey here

njgheorghita:
@pipermerriam I'm having some trouble getting travis to pass. 1. the `secrets` module was introduced in 3.6 and `tox -e py35-p2p` is failing because it can't find the module. 2. i'm not sure why `tox -e py36-p2p` is failing, I can't reproduce the failure locally, any idea what might be going wrong?

pipermerriam:
@njgheorghita So we can drop `tox -e py35-p2p` as our p2p code only needs to be compatible with python 3.6.  Does that help?

njgheorghita:
@pipermerriam sounds good - i made the change to `tox.ini` that i thought would drop `py35-p2p` - now when i run `tox -l` locally `py35-p2p` no longer shows up as an option - but travis is still evaluating it, i'm having trouble figuring out what i'm missing in `tox.ini`?

tests/p2p/test_server.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.fixture
+def server():
+    privkey = keys.PrivateKey(test_values['receiver_private_key'])
+    chaindb = ChainDB(MemoryDB())
+    server = Server(privkey, SERVER_ADDRESS, chaindb)
+    return server
+
+
+def test_server_authenticates_incoming_connections(server, event_loop):
```
njgheorghita: @gsalgado @pipermerriam Thoughts on improving this test? It passes locally, but fails on Travis. I haven't been able to figure out a better way to test this (from what i can tell we can't exactly follow `test_handshake()`'s strategy here because it doesn't seem possible to set the responder_nonce in `receive_handshake()` as a known value).

njgheorghita: Yeah, and also when I run the tests with pytest.

hwwhww:
Hello! I wonder if it's because Travis env doesn't accept some IPv6 address, so I experimented it with this commit: https://github.com/hwwhww/py-evm/commit/b201f034c302e700b97f9ab979872c8c2cdbd46a

And it can pass the CI: https://travis-ci.org/hwwhww/py-evm/jobs/365489490
p.s. I'm not sure if we should use `127.0.0.1` or `0.0.0.0` for test.

njgheorghita:
@hwwhww Great catch, thanks!

njgheorghita:
@pipermerriam @gsalgado @carver Thanks for the help in getting this cleaned up, I've implemented your suggested changes, let me know if there's anything else that needs changing

pipermerriam:
Haven't looked this over but.. I think this is something that people need *soon* to start building off of, so I'm good if we think we can merge a *mildly* incomplete version of this and then clean it up iteratively.

p2p/server.py
```diff
@@ -0,0 +1,136 @@
...
+    def __init__(self,
+                 privkey: datatypes.PrivateKey,
+                 server_address: Address,
+                 chaindb: AsyncChainDB
+                 ) -> None:
+        self.cancel_token = CancelToken('Server')
+        self.chaindb = chaindb
+        self.privkey = privkey
+        self.server_address = server_address
+        discovery = DiscoveryProtocol(self.privkey, self.server_address, bootstrap_nodes=[])
```
gsalgado: Why no bootstrap nodes?

p2p/server.py
```diff
@@ -0,0 +1,136 @@
...
+        self.peer_pool = PeerPool(ETHPeer, self.chaindb, 1, self.privkey, discovery)
```
gsalgado: The network_id should be passed as an argument to Server

p2p/server.py
```diff
@@ -0,0 +1,136 @@
...
+        self.logger.info("Running server...")
+        loop = asyncio.get_event_loop()
+        factory = asyncio.start_server(
+            self.receive_handshake, host=self.server_address.ip, port=self.server_address.udp_port)
+        asyncio.ensure_future(factory)
+
+        while not self.cancel_token.triggered:
+            try:
+                loop.run_forever()
+            except OperationCancelled:
```
gsalgado: This will never raise OperationCancelled. You can just `await self.cancel_token.wait()` if all you want is to yield and only resume (although you'd immediately return) once the token is triggered.

p2p/server.py
```diff
@@ -0,0 +1,136 @@
...
+        # Use reader to read the auth_init msg until EOF
+        msg = await reader.read()
+
+        # Use HandshakeResponder.decode_authentication(auth_init_message) on auth init msg
+        ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(msg, self.privkey)
+
+        # Get remote's address: IPv4 or IPv6
+        peername = writer.get_extra_info("peername")
+        ip = peername[0]
+        port = peername[1]
```
gsalgado: Can you use sequence unpacking here?

p2p/server.py
```diff
@@ -0,0 +1,136 @@
...
+            auth_init_ciphertext=msg,
+            auth_ack_ciphertext=auth_ack_ciphertext
+        )
+
+        # Create and register peer in peer_pool
+        eth_peer = ETHPeer(
+            remote=initiator_remote, privkey=self.privkey, reader=reader,
+            writer=writer, aes_secret=aes_secret, mac_secret=mac_secret,
+            egress_mac=egress_mac, ingress_mac=ingress_mac, chaindb=self.chaindb,
+            network_id=1
```
gsalgado: Again, do not hardcode network_id

tests/p2p/test_server.py
```diff
@@ -0,0 +1,129 @@
...
+    Node,
+    Address,
+)
+from p2p.peer import (
+    ETHPeer,
+    PeerPool,
+)
+from p2p.server import Server
+
+# constants copied from tests/p2p/test_auth.py : `test_handshake_eip8()`
```
gsalgado: Can't you move them to a common place instead of duplicating them here?

p2p/server.py
```diff
@@ -0,0 +1,136 @@
...
+
+    async def stop(self) -> None:
+        self.logger.info("Closing server...")
+        self.cancel_token.trigger()
+        asyncio.ensure_future(self.peer_pool.stop())
+
+    async def receive_handshake(
+            self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        # Use reader to read the auth_init msg until EOF
+        msg = await reader.read()
```
gsalgado: Do peers really send an EOF immediately after the auth msg?

p2p/server.py
njgheorghita: If the address is IPv4 it returns 2 elements, but if it's IPv6 it returns 4 elements, which is why I went with this setup. Is it safe to assume only IPv4 addresses?

pipermerriam: @njgheorghita you can do this in python 3

```python
In [7]: a, b, *c = [1, 2, 3, 4]

In [8]: print(a, b, c)
1 2 [3, 4]

In [9]: a, b, *c = [1, 2]

In [10]: print(a, b, c)
1 2 []
```

So you can just add an exta `*_` to the unpacking to catch the extra that you don't plan to use (`_`  is the convention for assigning to a variable that you don't plan to use).

```python
ip, port, *_ = peername
```

p2p/server.py
```diff
@@ -0,0 +1,150 @@
...
+        self.privkey = privkey
+        self.server_address = server_address
+        self.network_id = network_id
+        discovery = DiscoveryProtocol(
+            self.privkey, self.server_address, bootstrap_nodes=bootstrap_nodes)
+        self.peer_pool = PeerPool(ETHPeer, self.chaindb, self.network_id, self.privkey, discovery)
+
+    async def run(self) -> None:
+        self.logger.info("Running server...")
+        # loop = asyncio.get_event_loop()
```
pipermerriam: leftover?

tests/p2p/test_server.py
```diff
@@ -0,0 +1,93 @@
...
+
+    writer.write(messages)
+    if writer.can_write_eof():
+        writer.write_eof()
+    await writer.drain()
+
+    while True:
+        data = await reader.read(len(eip8_values['auth_ack_ciphertext']))
+        if data:
+            print(data)
```
pipermerriam: leftover?  Normally we try to avoid print statements in our testing code.

njgheorghita: I just got rid of the whole while statement, since there isn't much use for it

njgheorghita:
@pipermerriam Everything's failing on Travis all of a sudden, any clue what's going on?

pipermerriam:
@gsalgado you've been reviewing thus far.  Thoughts on where this is at and what needs to be done before merge?

p2p/server.py
```diff
@@ -0,0 +1,147 @@
...
+                 bootstrap_nodes: List[str],
+                 network_id: int
+                 ) -> None:
+        self.cancel_token = CancelToken('Server')
+        self.chaindb = chaindb
+        self.privkey = privkey
+        self.server_address = server_address
+        self.network_id = network_id
+        discovery = DiscoveryProtocol(
+            self.privkey, self.server_address, bootstrap_nodes=bootstrap_nodes)
```
gsalgado: Ideally we should lookup the bootstrap nodes based on the network_id. Care to add a TODO comment for that here?

p2p/server.py
```diff
@@ -0,0 +1,147 @@
...
+        self.peer_pool = PeerPool(ETHPeer, self.chaindb, self.network_id, self.privkey, discovery)
+
+    async def run(self) -> None:
+        self.logger.info("Running server...")
+        factory = asyncio.start_server(
+            self.receive_handshake, host=self.server_address.ip, port=self.server_address.udp_port)
+        asyncio.ensure_future(factory)
```
gsalgado: The `start_server()` call returns a `asyncio.Server` instance that we must store as an instance attribute in `p2p.Server` so that we can close it before we exit. Something like:

```
async def run(self) -> None:
    self.server = await asyncio.start_server(...)
    await self.stop()
```

p2p/server.py
```diff
@@ -0,0 +1,147 @@
...
+        await self.cancel_token.wait()
+
+    async def stop(self) -> None:
+        self.logger.info("Closing server...")
+        self.cancel_token.trigger()
+        asyncio.ensure_future(self.peer_pool.stop())
```
gsalgado: We should `await self.peer_pool.stop()` here instead. And we must also call `await self.server.close(); await self.server.wait_closed()`

p2p/server.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Use HandshakeResponder.decode_authentication(auth_init_message) on auth init msg
+        try:
+            ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
+                msg, self.privkey)
+        # Try to decode as EIP8
+        except DecryptionError:
+            msg_size = big_endian_to_int(msg[:2])
+            remaining_bytes = msg_size - ENCRYPTED_AUTH_MSG_LEN + 2
+            msg += await reader.read(remaining_bytes)
```
hwwhww: If it's a *wrong* pre-eip8 auth, this line will be executed but there is no more byte to read, and the coroutine will hang at this line. Maybe add a timeout?

pipermerriam: https://github.com/ethereum/py-evm/issues/574

pipermerriam:
@gsalgado I *think* it may have been caught by tests but `mypy` found this bug for me 😄 

647e6ee

p2p/server.py
```diff
@@ -0,0 +1,153 @@
...
+        discovery = DiscoveryProtocol(
+            self.privkey, self.server_address, bootstrap_nodes=bootstrap_nodes)
+        self.peer_pool = PeerPool(ETHPeer, self.chaindb, self.network_id, self.privkey, discovery)
+
+    async def run(self) -> None:
+        self.logger.info("Running server...")
+        self._server = await asyncio.start_server(
+            self.receive_handshake,
+            host=self.server_address.ip,
+            port=self.server_address.udp_port,
```
jannikluhn: This should probably be `self.server_address.tcp_port`.

pipermerriam: That I don't know.  cc @gsalgado ?

gsalgado: It should be `tcp_port`, yeah.

tests/p2p/test_server.py
```diff
@@ -0,0 +1,84 @@
...
+    assert server.cancel_token.triggered is False
+    assert server.peer_pool.cancel_token.triggered is False
+    # The sole connected node is our initiator
+    assert len(server.peer_pool.connected_nodes) is 1
+    initiator_peer = server.peer_pool.connected_nodes[INITIATOR_REMOTE]
+    assert isinstance(initiator_peer, ETHPeer)
+    assert initiator_peer.remote == INITIATOR_REMOTE
+    assert initiator_peer.privkey == keys.PrivateKey(eip8_values['receiver_private_key'])
+    # Stop server
+    event_loop.run_until_complete(server.stop())
```
hwwhww: Fixed CI via this change: https://github.com/hwwhww/py-evm/commit/8b06ae9cb21fdf5e40f586b8acfe630d49924e5f

```python
-    event_loop.run_until_complete(server.stop())
+    asyncio.ensure_future(server.stop())
+    event_loop.run_until_complete(stall(1))
```

gsalgado: @hwwhww I was going to just push your changes directly, but ended up reworking the whole test

p2p/server.py
```diff
@@ -0,0 +1,157 @@
...
+        self._server.close()
+        await self._server.wait_closed()
+        await self.peer_pool.stop()
+
+    async def receive_handshake(
+            self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        # Use reader to read the auth_init msg until EOF
+        msg = await reader.read(ENCRYPTED_AUTH_MSG_LEN)
+
+        # Use HandshakeResponder.decode_authentication(auth_init_message) on auth init msg
```
hwwhww: This comment reminds that it's a little weird to move `decode_authentication` to `server.py`. To me, it's more organized to leave this kind of functions in `auth.py`.

p2p/server.py
```diff
@@ -0,0 +1,157 @@
...
+        )
+
+        # Create and register peer in peer_pool
+        eth_peer = ETHPeer(
+            remote=initiator_remote, privkey=self.privkey, reader=reader,
+            writer=writer, aes_secret=aes_secret, mac_secret=mac_secret,
+            egress_mac=egress_mac, ingress_mac=ingress_mac, chaindb=self.chaindb,
+            network_id=self.network_id
+        )
+        self.peer_pool.add_peer(eth_peer)
```
hwwhww: I'm writing a test to make two client connection, maybe we should add a line after here?
```python
await eth_peer.do_p2p_handshake()
```

p2p/server.py
gsalgado: Agreed

p2p/server.py
gsalgado: Yeah, and you must catch `HandshakeFailure` and skip adding the peer to the pool in that case

hwwhww: Could we merge this PR now, and then I create another PR for the patch with tests?

pipermerriam: @hwwhww :shipit:

PR #502, opened 2018-03-26
p2p: ChainSyncer._handle_peer() no longer waits for handle_msg()

gsalgado:
This way the main loop can keep running and consuming peer messages while
sub-processes are running the cpu-intensive tasks.

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -3,7 +3,6 @@ import logging
 import operator
 import secrets
 import time
-import traceback
 from typing import Any, Awaitable, Callable, cast, Dict, Generator, List, Set, Tuple  # noqa: F401
 
 from cytoolz.itertoolz import unique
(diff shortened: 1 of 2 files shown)


p2p/chain.py
```diff
@@ -82,12 +82,23 @@ def register_peer(self, peer: BasePeer) -> None:
...
+                         msg: protocol._DecodedMsgType) -> None:
+        try:
+            await self._handle_msg(peer, cmd, msg)
+        except OperationCancelled:
+            # Silently swallow OperationCancelled exceptions because we run unsupervised (i.e.
+            # with ensure_future()). Our caller will also get an OperationCancelled anyway, and
+            # there it will be handled.
+            pass
+        except Exception:
+            self.logger.error("Unexpected error when processing msg from %s: %s",
```
pipermerriam: I think that you can use `self.logger.exception("only the message")` and it will include a stacktrace in the logging output, allowing us to skip the `traceback.formac_exc()` logic.

PR #505, opened 2018-03-26
evm/db: New non-journaled version of ChainDB

### LINKED ISSUE #495: JournalDB makes ChainDB.persist_trie_data_dict() up to 20x slower on a fully populated DB
`ChainDB.persist_trie_data_dict()` is yet another bottleneck I've found when running a full-node sync: it takes between 3 to 9 seconds to persist a batch of (128) bodies or (256) receipts, and most of that was spent on `JournalDB.set()`, which does a `wrapped_db.get()` before setting the new value.

This is specially problematic when performing a chain sync, but at the same a chain sync doesn't need a journaled DB, so maybe in that case we can just use a custom, non-journaled, DB class? Can anybody think of possible problems with that? Also, does that mean we'd need to extend trinity's DB-serving process to also offer a non-journaled chaindb?

gsalgado:
Also use it in p2p's FakeAsyncChainDB

Closes #495

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -631,3 +631,23 @@ class AsyncChainDB(ChainDB):
 
     async def coro_persist_trie_data_dict(self, *args, **kwargs):
         raise NotImplementedError()
+
+
+class NonJournaledAsyncChainDB(AsyncChainDB):
+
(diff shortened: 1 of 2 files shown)


pipermerriam:
After https://github.com/ethereum/py-evm/issues/507 has been done we should re-visit this issue and see if this mechanism can be improved.

PR #506, opened 2018-03-26
WIP: Christoph/chore/evm types 5

cburgdorf:
### What was wrong?

Missing type hints

### How was it fixed?

Adds type hints

#### Cute Animal Picture


evm/rlp/transactions.py
```diff
@@ -235,3 +236,6 @@ def get_intrinsic_gas(self) -> int:
         for computation).
         """
         raise NotImplementedError("Must be implemented by subclasses")
+
+
+AnyBaseTransaction = Union[Type[BaseShardingTransaction], Type[BaseTransaction]]
```
cburgdorf: The only reason I did that is because we have many APIs that otherwise need to accept or return `Union[Type[BaseShardingTransaction], Type[BaseTransaction]]` which feels a little noisy. 

I'm not quite happy with the naming though because it doesn't include the `BaseUnsignedTransaction` whereas the name suggests it could be any kind of base transaction

pipermerriam: Generally good though #508 is likely to make this obsolete.  No change needed on your part, just pointing it out so you're aware (also, don't start typing the sharding stuff)

cburgdorf:
Is there anything blocking this from getting merged?

cburgdorf:
Oh, just noticed that this needs a rebase. Will do that on Monday.

PR #510, opened 2018-03-26
Piper/add support for custom genesis data in trinity

pipermerriam:
### What was wrong?

Needed support for running *custom* chains with custom genesis data.

### How was it fixed?

Added new cli flag `--genesis` which can be used to supply the genesis block parameters. This value is used during chain initialization.

TODO:

- [ ] decide whether this flag should be allowed in normal top level `trinity` command or if it should only be allowed in `trinity init` (which doesn't currently exist).
- [ ] add dynamic selection of `PeerPool` class since `HardCodedNodesPeerPool` doesn't support custom network ids.

#### Cute Animal Picture

flying-cats

### DIFF
--- a/evm/chains/mainnet/__init__.py
+++ b/evm/chains/mainnet/__init__.py
@@ -38,7 +38,7 @@ MainnetChain = Chain.configure(
 )
 
 
-MAINNET_GENESIS_HEADER = BlockHeader(
+MAINNET_GENESIS_PARAMS = dict(
     difficulty=17179869184,
     extra_data=decode_hex("0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa"),
     gas_limit=5000,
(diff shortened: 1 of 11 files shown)


pipermerriam:
I think the `PeerPool` issue is low priority and it raises clear errors when you do try to run any chain with a network id that isn't mainnet or ropsten.

@carver I need some help sorting out how the `--genesis` flag should be used.

- in a new `init` command `trinity init --genesis "path/to/genesis.json`  --data-dir /custom/data-dir-location
- in the primary cli `trinity --genesis "path/to/genesis.json" --network-id 42 --data-dir /custom/data-dir-location

 We can do both and just *punt* on implementing `init` since we don't really have the right things in place to support running custom local chains at the moment.

carver:
Can we reframe the problem as: we want to choose which of several preconfigured networks to run?

If so, :+1: to punting on the full genesis implementation until all the under-the-hood stuff is set up.

I like this syntax:

```sh
trinitiy --network ropsten --data-dir ...
# or
trinitiy --network mainnet --data-dir ...
```

But the more geth-consistent one would be `trinity --rinkeby`. I don't like that it pollutes the global argument namespace. Who knows how many names of networks we would want to support?

pipermerriam:
Closing this for now.  Since we're light client *only* right now **and** we have a hard-coded list of peers for the peerpool (for now), network IDs outside of mainnet and ropsten will cause a hard failure. Since I don't think fixing that is high priority right now (since you can't really run a private chain with trinity yet), I'm closing this until it's time to implement the full feature set.

PR #511, opened 2018-03-27
Fix wrong trie class in generating transaction trie in Sharding

NIC619:
### What was wrong?
Didn't set Binary Trie when generating transaction trie in Sharding(default to Hexary Trie) and result in node parsing error when trying to fetch value with Hexary Trie data in a Binary Trie.

### How was it fixed?
Specify Binary Trie as trie_class in `make_trie_root_and_nodes`.

#### Cute Animal Picture

### DIFF
--- a/evm/vm/forks/sharding/vm_state.py
+++ b/evm/vm/forks/sharding/vm_state.py
@@ -240,7 +240,10 @@ class ShardingVMState(ShardingTransactionExecutor, ByzantiumVMState):
         block.transaction_fee_sum += transaction_fee
 
         # Get trie roots and changed key-values.
-        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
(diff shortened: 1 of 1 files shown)


evm/vm/forks/sharding/vm_state.py
```diff
@@ -240,7 +240,10 @@ def add_transaction(self, transaction, computation, block):
         block.transaction_fee_sum += transaction_fee
 
         # Get trie roots and changed key-values.
-        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
+        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(
```
pipermerriam: It strikes me that `make_trie_root_and_nodes` should require the trie class in order to prevent these types of mistakes/errors.

pipermerriam:
I'm still inclined to make the `trie_class` parameter required, it can just happen in a different PR

PR #512, opened 2018-03-27
Code reorganization within evm.vm module

pipermerriam:
### What was wrong?

- Many classes and modules which are specific to VM logic and execution were located outside of `evm.vm` module.
- The `VMState` object is easier to reference as just the `State`.  This fits well with the rename of what we currently refer to as the `AccountStateDB` to just be the `AccountDB` which removes the name collision.

### How was it fixed?

Relocated these modules to be within `evm.vm`.

- `evm.computation`
- `evm.vm_state` -> `evm.vm.state`
- `evm.opcode`
- `evm.opcode_values`
- `evm.mnemonics`

And renamed all `vm_state` -> `state` and all `VMState` -> `State` (including suffixes like `FrontierVMState` -> `FrontierState`.

#### Cute Animal Picture

6858a19067461464d21f3ed1c53f9e72


docs/api.rst
```diff
@@ -0,0 +1,4 @@
+API Docs
+========
+
+TODO
```
carver: Maybe this wasn't meant to be included in this PR?

PR #515, opened 2018-03-27
Handle latest web3 beta, enabling accounts features

pipermerriam:
### What was wrong?

Latest web3 beta requires explicit opt-in of account features.

### How was it fixed?

Added conditional opt-in.

#### Cute Animal Picture

farm-animal-buddies-ftr

### DIFF
--- a/tests/sharding/fixtures.py
+++ b/tests/sharding/fixtures.py
@@ -66,6 +66,8 @@ def vmc():
     )
     provider = EthereumTesterProvider(eth_tester)
     w3 = Web3(provider)
+    if hasattr(w3.eth, "enable_unaudited_features"):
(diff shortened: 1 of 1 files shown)


tests/sharding/fixtures.py
```diff
@@ -66,6 +66,8 @@ def vmc():
     )
     provider = EthereumTesterProvider(eth_tester)
     w3 = Web3(provider)
+    if hasattr(w3.eth, "enable_unaudited_features"):
+        w3.eth.enable_unaudited_features()
```
carver: I'll probably leave this method in, so you don't *have* to do an attribute test here. After the audit is complete it would be a noop. But no harm to check the attribute. v5 forward compatibility :P

PR #516, opened 2018-03-28
Remove block_class from VM

### LINKED ISSUE #513: Both `VM` and `State` classes define a `block_class`
### What is wrong?

Both the `VM` and `State` (aka `VMState`) objects implement a `block_class` or (`_block_class` from the `VM`) property and API.  There should only be one `block_class` defined for a given VM ruleset.

### How can it be fixed

- Remove the `block_class` API from the `VM` class and change all usage to delegate to the `State` class API.
- Change the `State` class to have a `get_block_class()` which is a `classmethod` for retrieval of the class instead of directly accessing it via the attribute.

cburgdorf:
### What was wrong?

Both the `VM` and the `VMState` had the `block_class` defined.

### How was it fixed?

This removes the `block_class` from `VM` and delegates to the `get_block_class` API of the `VMState` instead. Fixes #513 

#### Cute Animal Picture

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -52,7 +52,6 @@ class BaseVM(Configurable, metaclass=ABCMeta):
     network.
     """
     chaindb = None
-    _block_class = None
     _state_class = None
 
     def __init__(self, header, chaindb):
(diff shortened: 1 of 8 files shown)


evm/vm/base.py
```diff
@@ -344,13 +343,7 @@ def create_unsigned_transaction(self, *args, **kwargs):
     #
     @classmethod
     def get_block_class(cls):
-        """
-        Return the class that this VM uses for blocks.
```
cburgdorf: Should I put the comment back in? Not sure if this is meant to have a comment now that it only delegates to `get_block_class()` on the state?

pipermerriam: I think the comment should be changed to indicate that it's a passthrough to to the underlying API.

cburgdorf: :+1: I updated it

cburgdorf:
@pipermerriam I put the comment back in and adjusted it and also rebased this on latest master.

Issue #475, opened 2018-03-17
SMC(VMC) optimize collation header to 32 bytes


NIC619:
### How can it be fixed

Following the new spec, compressing the size of collation header from 2*32 bytes to 32 bytes

hwwhww:
close via #517

PR #517, opened 2018-03-29
SMC collation header optimization

NIC619:
### What was wrong?
#475 
Compress collation header info stored in contract to 32 bytes: 24 bytes parent header hash + 8 bytes header score.

#### Cute Animal Picture


evm/rlp/headers.py
```diff
@@ -247,7 +247,7 @@ def hash(self) -> bytes:
                 int_to_bytes32(self.number),
             ))
         )
-        return header_hash
+        return pad32(header_hash[8:])
```
mhchia: Will we keep using the padded `bytes32` from `bytes24` in the future? or is it just temporary

NIC619: Dealing with `bytes24` as header hash(e.g., `collation_headers: public(bytes32[bytes <= 24][int128])`) in Vyper takes extra ~30000 gas in `addHeader`.
Perhaps pad it to 32 bytes in contract while remain 24 bytes outside out contract? Though it could be confusing for other developers since this is a Vyper specific issue.

evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -28,11 +28,10 @@
...
 num_validators: public(int128)
 
-# Collation headers
-collation_headers: public({
-    parent_hash: bytes32,
-    score: int128,
-}[bytes32][int128])
+# Collation headers: (parent_hash || score)
+# parent_hash: 24 bytes
+# score: 8 bytes
```
hwwhww: Last time (March 5th) I heard from V it's 26 bytes + 6 bytes for JMRS, so did it changed to 24 bytes + 8 bytes?

evm/vm/forks/sharding/contracts/validator_manager.v.py
```diff
@@ -237,37 +251,68 @@ def add_header(
...
+    entire_header_hash: bytes32 = convert(
+        uint256_mod(
+            convert(sha3(header_bytes), 'uint256'),
+            # Mod 2^192, i.e., extract right most 24 bytes
+            convert(6277101735386680763835789423207666416102355444464034512896, 'uint256')
+        ),
+        'bytes32'
+    )
+    # Check that this header has not yet been added.
+    assert not self.collation_headers[shard_id][entire_header_hash]
```
hwwhww: Just saying, if you move `assert self.period_head[shard_id] < expected_period_number` in front, then you don't need to check `assert not self.collation_headers[shard_id][entire_header_hash]`.... unless SHA-3 collision(!)

evm/vm/forks/sharding/contracts/validator_manager.v.py
NIC619: I remember the 24+8 I heard was before March 5th so I guess 26+6 is the latest one.

evm/vm/forks/sharding/contracts/validator_manager.v.py
NIC619: Good point. Though if we implement slashing condition for adding multiple header at same shard during same period then `self.period_head` can be removed.

evm/rlp/headers.py
mhchia: yes it makes sense.

evm/rlp/headers.py
```diff
@@ -247,7 +247,7 @@ def hash(self) -> bytes:
                 int_to_bytes32(self.number),
             ))
         )
-        return header_hash
+        return pad32(header_hash[6:])
```
mhchia: nitpick: can we leave a comment or use a constant to store the hash length?

Issue #508, opened 2018-03-26
Delete most of the sharding code


pipermerriam:
### What is wrong?

Most of the code that's been implemented for sharding is no longer part of the phase 1 roadmap.  Given that the future of this code is uncertain and we have no direct use for it now....

### How can it be fixed

Lets delete as much of it as seems reasonable.

hwwhww:
1. The things that I suppose would be implemented in eWASM or the Python interpreter?
	1. `ShardingAccountStateDB`
	2. account list restrictions
	3. part of account abstraction?
2. #452 : move SMC and SMCHandler to sharding repo 
3. Remove all:
	1. `Shard` (`evm/shard.py`)
	2. `ShardVM` (`evm/vm/shard_vm.py`)
	3. `ShardingVM` (`evm/vm/forks/sharding`)
	4. `TrackedDB` and the the usage
4. Considering:
	1. `ShardingAccountStateDB` (@pipermerriam  your call)
	2. `evm/auxiliary` (seems no harm?)

pipermerriam:
Less about harm of leaving things in and more about whether they are used.  I know *most* of that stuff will be used eventually, but given the ease of resurrecting it, I'd prefer to take the route of delete and then revive later when we need it again.

carver:
> I know most of that stuff will be used eventually, but given the ease of resurrecting it, I'd prefer to take the route of delete and then revive later when we need it again.

I think we are better off finding a different way to indicate that the sharding code is dormant for the next phase:

We are setting ourselves up to do a lot of work when reviving the code, because the internal APIs are so fluid. I expect that it's going to be like the recent merge of the sharding branch, all over again. Instead, if we leave the code in, the tests will catch it when internal API changes break the dormant sharding code.

jannikluhn:
I guess it depends on what would take more effort:

- keeping the sharding code running at all times through the API changes
- resurrecting it once

Intuitively I'd say resurrecting once is easier, especially given that both the py-EVM and the sharding state execution model (-> alternative execution engines) will change. Also I would be surprised if there are no additional EVM changes coming. But I wasn't involved in the recent merges, so this is just my gut feeling, with no practical experience to justify it.

hwwhww:
closed via #518

PR #518, opened 2018-03-29
Remove sharding code

jannikluhn:
### What was wrong?

#508 

### How was it fixed?

Deleted most of the sharding code. Only collation and collation header remain. I tried (semi-successfully) to split it up into small commits so that reviving becomes easier.

This should be merged only after the SMC and SMC handler have been transferred to the sharding repository.

#### Cute Animal Picture


evm/exceptions.py
```diff
@@ -151,14 +151,6 @@ class OutOfBoundsRead(VMError):
...
 
-class UnannouncedStateAccess(VMError):
-    """
-    Error raised if a storage slot or account field is accessed that has not
-    been specified in the transaction's read or write list, respectively.
-    """
-    pass
-
-
 class GasPriceAlreadySet(Exception):
```
pipermerriam: Should this get remove since I believe it is `PAYGAS` specific.

hwwhww:
And remove `eth-tester` from setup.py?

evm/exceptions.py
jannikluhn: :+1: missed this (same for `NotTopLevelCall`)

hwwhww:
Also, remove `TrackedDB(BaseDB)` which was for the stateless client? And it requires removing the all `access_logs` stuff.

tests/core/helpers.py
```diff
@@ -1,7 +1,3 @@
-from eth_utils import decode_hex
-
-from evm.utils.padding import pad32
-from evm.vm.forks.sharding.transactions import ShardingTransaction
```
hwwhww: Remove two empty lines.

setup.py
```diff
@@ -26,7 +26,6 @@
         "rlp>=0.4.7,<1.0.0",
         "eth-keys>=0.2.0b3,<1.0.0",
         "trie>=1.3.0,<2.0.0",
-        "eth-tester==0.1.0b21",
         "web3>=4.0.0b6,<5.0.0",
```
hwwhww: oh sorry, also remove `web3`?

tests/core/helpers.py
jannikluhn: Flake8 fails on my machine if I do this (`tests/core/helpers.py:1:1: E302 expected 2 blank lines, found 0`). Pushed this anyway, let's see what Travis thinks about this.

jannikluhn: Apparently it's a known bug in pep8/pycodestyle: https://github.com/PyCQA/pycodestyle/issues/473

I've added a noqa for now.

jannikluhn:
Rebased and removed recently added sharding code (shard tracker and some SMC stuff). Will merge when the tests have passed.

carver:
> I had to add eth-hash as a dependency

eth-hash is already loaded by eth-utils, but web3.py loads a specific eth-hash backend (pycryptodome). In keeping with the eventual idea of py-evm becoming a library/utility, and trinity being the full-featured node, py-evm should probably _not_ be opinionated about the hashing backend used.

Instead, we can add a backend only during testing, like how eth-utils does it:
https://github.com/ethereum/eth-utils/blob/master/tox.ini#L14-L15

jannikluhn:
Good point, I've added it to `dev-requirements.txt`. Can I merge this now?

NIC619:
There seems to be `access_list` left uncleaned in `Message`(`vm/message.py`) and `state_db()`(`vm/state.py`)?

PR #519, opened 2018-03-29
p2p: Redesign the fast-sync mechanism

gsalgado:
We now fetch a batch of headers and then all the bodies/receipts for those
headers before persisting them. This makes the code a lot easier to follow
but also guarantees our ChainDB always has full blocks for everything up to
the current chain head.

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -319,45 +319,15 @@ class ChainSyncer(PeerPoolSubscriber):
 
     async def _handle_msg(self, peer: ETHPeer, cmd: protocol.Command,
                           msg: protocol._DecodedMsgType) -> None:
-        loop = asyncio.get_event_loop()
         if isinstance(cmd, eth.BlockHeaders):
             msg = cast(List[BlockHeader], msg)
             self.logger.debug(
                 "Got BlockHeaders from %d to %d", msg[0].block_number, msg[-1].block_number)
             self._new_headers.put_nowait(msg)
         elif isinstance(cmd, eth.BlockBodies):
-            self._peers_with_pending_requests.remove(peer)
-            msg = cast(List[eth.BlockBody], msg)
-            self.logger.debug("Got %d BlockBodies from %s", len(msg), peer)
-            iterator = map(make_trie_root_and_nodes, [body.transactions for body in msg])
-            transactions_tries = await wait_with_token(
-                loop.run_in_executor(None, list, iterator),
-                token=self.cancel_token)
-            for i in range(len(msg)):
-                body = msg[i]
-                tx_root, trie_dict_data = transactions_tries[i]
-                await self.chaindb.coro_persist_trie_data_dict(trie_dict_data)
-                # TODO: Add transactions to canonical chain; blocked by
-                # https://github.com/ethereum/py-evm/issues/337
-                uncles_hash = await self.chaindb.coro_persist_uncles(body.uncles)
-                self._pending_bodies.pop((tx_root, uncles_hash), None)
+            await self._handle_block_bodies(peer, cast(List[eth.BlockBody], msg))
         elif isinstance(cmd, eth.Receipts):
-            self._peers_with_pending_requests.remove(peer)
-            msg = cast(List[List[eth.Receipt]], msg)
-            self.logger.debug("Got Receipts for %d blocks from %s", len(msg), peer)
-            iterator = map(make_trie_root_and_nodes, msg)
-            receipts_tries = await wait_with_token(
-                loop.run_in_executor(None, list, iterator),
-                token=self.cancel_token)
-            for receipt_root, trie_dict_data in receipts_tries:
-                if receipt_root not in self._pending_receipts:
-                    self.logger.warning(
-                        "Got unexpected receipt root: %s",
-                        encode_hex(receipt_root),
-                    )
-                    continue
-                await self.chaindb.coro_persist_trie_data_dict(trie_dict_data)
-                self._pending_receipts.pop(receipt_root)
+            await self._handle_block_receipts(peer, cast(List[List[eth.Receipt]], msg))
         elif isinstance(cmd, eth.NewBlock):
             msg = cast(Dict[str, Any], msg)
             header = msg['block'][0]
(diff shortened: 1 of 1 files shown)


p2p/chain.py
```diff
@@ -164,142 +149,134 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            missing = [
+                header for header in missing
+                if key_func(header) not in downloaded]
+
+    def _request_block_parts(
+            self,
+            headers: List[BlockHeader],
+            request_func: Callable[[ETHPeer, List[BlockHeader]], None]) -> int:
+        length = math.ceil(len(headers) / len(self.peer_pool.peers))
+        batches = list(partition_all(length, headers))
```
pipermerriam: I assume the `list` casting here has to do with the following `zip` not supporting whatever `partition_all` natively returns?

p2p/chain.py
```diff
@@ -164,142 +149,134 @@ def register_peer(self, peer: BasePeer) -> None:
...
-                peer
-                for peer in self.peer_pool.peers
-                if peer not in self._peers_with_pending_requests
-            ]
-            if idle_peers:
-                return cast(ETHPeer, secrets.choice(idle_peers))
-            else:
-                self.logger.debug("No idle peers availabe, sleeping a bit")
-                await asyncio.sleep(0.2)
+    def _body_key(self, header: BlockHeader) -> Tuple[bytes, bytes]:
```
pipermerriam: Maybe elevate these to top level functions to get them out of the class namespace?

p2p/chain.py
```diff
@@ -164,142 +149,134 @@ def register_peer(self, peer: BasePeer) -> None:
...
+        """Return the unique key of the body for the given header.
+
+        i.e. a two-tuple with the transaction root and uncles hash.
+        """
+        return cast(Tuple[bytes, bytes], (header.transaction_root, header.uncles_hash))
+
+    def _is_body_empty(self, header: BlockHeader) -> bool:
```
pipermerriam: Same (elevate to top level module function)

p2p/chain.py
```diff
@@ -164,142 +149,134 @@ def register_peer(self, peer: BasePeer) -> None:
...
+        return False
+
+    def request_bodies(self, headers: List[BlockHeader]) -> int:
+        """Ask our peers for bodies for the given headers.
+
+        See request_receipts() for details of how this is done.
+        """
+        return self._request_block_parts(headers, self._send_get_block_bodies)
+
+    def _receipts_key(self, header: BlockHeader) -> bytes:
```
pipermerriam: And so on about elevating to top level module function, not going to comment on the others.

p2p/chain.py
```diff
@@ -378,6 +325,35 @@ def _skip_empty_and_duplicated_receipts(
...
+        self.logger.debug("Got Receipts for %d blocks from %s", len(receipts), peer)
+        loop = asyncio.get_event_loop()
+        iterator = map(make_trie_root_and_nodes, receipts)
+        receipts_tries = await wait_with_token(
+            loop.run_in_executor(None, list, iterator),
+            token=self.cancel_token)
+        receipt_roots = []
+        for receipt_root, trie_dict_data in receipts_tries:
+            await self.chaindb.coro_persist_trie_data_dict(trie_dict_data)
+            receipt_roots.append(receipt_root)
```
pipermerriam: What do you think about changing this to the following (after the exit of the loop)

```python
receipt_roots = tuple(zip(*receipt_tries))[0]
```

p2p/chain.py
```diff
@@ -378,6 +325,35 @@ def _skip_empty_and_duplicated_receipts(
...
+
+    async def _handle_block_bodies(self, peer: ETHPeer, bodies: List[eth.BlockBody]) -> None:
+        self.logger.debug("Got Bodies for %d blocks from %s", len(bodies), peer)
+        loop = asyncio.get_event_loop()
+        iterator = map(make_trie_root_and_nodes, [body.transactions for body in bodies])
+        transactions_tries = await wait_with_token(
+            loop.run_in_executor(None, list, iterator),
+            token=self.cancel_token)
+        body_keys = []
+        for i, body in enumerate(bodies):
```
pipermerriam: I think you can change this to be:

```python
for (body, (tx_root, trie_data_dict)) in zip(bodies, transaction_tries):
    ...
```

p2p/chain.py
gsalgado: Yep, otherwise we can't get its length

p2p/chain.py
gsalgado: In order to more this one I'd need to add an extra argument (for the empty_root_hash, which it currently takes from the chaindb), so I'd rather keep it here unless you feel strongly about it

pipermerriam: You can actually hard code that value as the alternate trie implementation which required the empty root hash to be specified is being removed as part of #518

p2p/chain.py
gsalgado: Moved as well

p2p/chain.py
gsalgado: I'd prefer the current version as it doesn't have any cognitive overhead and when debugging one can immediately see the list will contain all receipt roots (and just them) at the end of the loop, whereas the latter requires me to remember/lookup how zip works, and in fact I wasn't even aware it is a variadic function

p2p/chain.py
gsalgado: Indeed!

p2p/chain.py
pipermerriam: Ok, how about:

```python
receipt_roots = [receipt_root for receipt_root, _ in receipt_tries]
```

Simply removes the loop based accumulation in favor of a comprehension which maintains readability.

pipermerriam: Just an FYI so you know it exists, I just found [`cytoolz.pluck`](http://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.pluck) 

```python
receipt_roots = tuple(pluck(0, receipt_tries))  # cast to tuple to force evaluation
```

PR #521, opened 2018-03-30
Adjust how MainnetTesterChain configures VMs to not require updates for every fork rule

pipermerriam:
link: https://github.com/ethereum/web3.py/issues/729

### What was wrong?

The way that `evm.chains.tester.MainnetTesterChain` was done, it would require an update to the fork block number configuration for every new set of fork rules.  It also didn't allow extension with custom vms.

### How was it fixed?

- Modified the `BaseVM` class to have a record of the *name* of the fork.
- Modified the helper function which generates the VM fork rules to derive it's available VMs from the `MainnetChain`
- Simplified the logic some so that it expects very explicitly defined start blocks and doesn't try to be *smart*.

### DIFF
--- a/evm/chains/tester/__init__.py
+++ b/evm/chains/tester/__init__.py
@@ -1,31 +1,37 @@
+import collections
+import operator
 from typing import (
-    Any
+    Any,
+    Generator,
+    Sequence,
+    Tuple,
+    Type,
+    Union,
 )
 
 from cytoolz import (
     assoc,
+    last,
 )
 
 from eth_utils import (
-    reversed_return,
+    to_tuple,
 )
 
 from evm.chains.base import Chain
-
-from evm.vm.forks import (
-    FrontierVM as BaseFrontierVM,
-    HomesteadVM as BaseHomesteadVM,
-    TangerineWhistleVM as BaseTangerineWhistleVM,
-    SpuriousDragonVM as BaseSpuriousDragonVM,
-)
-
-from evm.utils.chain import (
-    generate_vms_by_range,
-)
-
+from evm.chains.mainnet import MainnetChain
+from evm.exceptions import ValidationError
 from evm.rlp.headers import (
     BlockHeader
 )
+from evm.utils.chain import (
+    generate_vms_by_range,
+)
+from evm.validation import (
+    validate_gte,
+)
+from evm.vm.base import BaseVM
+from evm.vm.forks.homestead import HomesteadVM
 
 
 class MaintainGasLimitMixin(object):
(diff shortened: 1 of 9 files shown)


pipermerriam:
This is step one, step two is going to be in the `eth-tester` repo to adjust to this change in API.

pipermerriam:
Probably monday before I can get the eth-tester component in.  Once I got into things it looks like the API for configuring fork blocks needs a little love.

evm/vm/forks/spurious_dragon/__init__.py
```diff
@@ -3,7 +3,10 @@
 from .state import SpuriousDragonState
 
 SpuriousDragonVM = HomesteadVM.configure(
+    # class name
     __name__='SpuriousDragonVM',
+    # fork name
+    fork='spurious-dragon',
```
carver: Do these fork names follow any pre-existing standard?

It is nice that the formatting is internally consistent, but it would be even better to make them compatible with something external out of the gate, either code or documentation. My first instinct is that we could name them the same way that ethereum/tests names the forks.

tests/core/tester/test_generate_vm_configuration.py
```diff
@@ -39,37 +55,53 @@ class Forks(enum.Enum):
...
         (
-            dict(homestead_start_block=1, dao_start_block=2),
+            ((0, 'frontier'), (1, 'homestead')),
+            {'dao_start_block': 2},
             ((0, Forks.Frontier), (1, Forks.Homestead)),
         ),
         (
-            dict(homestead_start_block=1, dao_start_block=False),
+            ((0, 'frontier'), (1, 'homestead')),
+            {'dao_start_block': False},
```
carver: I had a hard time guessing what `False` and `None` did... but I see that you added docs, and I don't have a better suggestion for the API, so I guess nothing to do here.

evm/chains/tester/__init__.py
```diff
@@ -42,91 +48,76 @@ def create_header_from_parent(cls,
...
+            "The `dao_start_block` parameter is only valid for the 'homestead' "
+            "fork rules.  The 'homestead' VM was not included in the provided "
+            "fork configuration"
+        )
 
-        remaining_blocks_not_declared = (
-            homestead_start_block is None and
-            tangerine_whistle_start_block is None
+    # Validate that there is at least one start block for block 0
+    start_blocks = sorted(set(start_block for start_block, _ in fork_start_blocks))
```
carver: why `sorted`?

evm/chains/tester/__init__.py
```diff
@@ -42,91 +48,76 @@ def create_header_from_parent(cls,
...
-            yield (0, FrontierTesterVM)
+    ordered_fork_start_blocks = sorted(fork_start_blocks, key=operator.itemgetter(0))
+
+    for start_block, fork in ordered_fork_start_blocks:
+        if isinstance(fork, type) and issubclass(fork, BaseVM):
+            vm_class = fork
+        elif isinstance(fork, str):
+            vm_class = MAINNET_VMS[fork]
+        else:
+            raise Exception("Invariant: unreachable code path")
```
carver: Do the type definitions make this literally unreachable? If it's possible for someone to supply a bad fork value here, then a better error message would be nice.

evm/vm/forks/spurious_dragon/__init__.py
pipermerriam: Currently the fork names have all used their *fancy* name.  see the `evm.vm.forks` module.

evm/chains/tester/__init__.py
pipermerriam: misplaced.  moving it to within the `if` block into the error message format.

evm/chains/tester/__init__.py
pipermerriam: I believe that the typing makes this unreachable (with code that has passed the `mypy` checks) but I still like having these in place to basically document the remaining logic branch.

PR #528, opened 2018-04-04
p2p: Performance/misc improvements to StateDownloader

gsalgado:
- Run cpu-intensive tasks using asyncio.run_in_executor()
- _handle_peer() uses ensure_future() to handle msgs so it can keep
  consuming other msgs from the same peer while the cpu-intensive tasks
  are executed in other processes
- Redesign coroutine that retries timed out items so it wakes up only
  when necessary
- Redesign progress reporting coroutine as above


p2p/state.py
```diff
@@ -101,74 +101,95 @@ def register_peer(self, peer: BasePeer) -> None:
...
-        if not timed_out:
-            return
-        self.logger.debug("Re-requesting %d trie nodes", len(timed_out))
-        await self.request_nodes(timed_out)
+    async def _periodically_retry_timedout(self):
+        while True:
+            now = time.time()
+            oldest_request_time = now
+            timed_out = []
+            for node_key, req_time in list(self._pending_nodes.items()):
```
pipermerriam: reason for the `list` casting here?

p2p/state.py
```diff
@@ -101,74 +101,95 @@ def register_peer(self, peer: BasePeer) -> None:
...
+                    timed_out.append(node_key)
+                elif req_time < oldest_request_time:
+                    oldest_request_time = req_time
+            if timed_out:
+                self.logger.debug("Re-requesting %d trie nodes", len(timed_out))
+                self._total_timeouts += len(timed_out)
+                await self.request_nodes(timed_out)
+
+            now = time.time()
+            sleep_until = (oldest_request_time + self._reply_timeout) - now
```
pipermerriam: I think `sleep_for` or `sleep_seconds` or `sleep_duration` would be a more appropriate name as I *think* this is an offset, rather than a timestamp.

p2p/state.py
```diff
@@ -101,74 +101,95 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            try:
+                await wait_with_token(asyncio.sleep(self._report_interval), token=self.cancel_token)
+            except OperationCancelled:
+                break
 
 
 class StateSync(HexaryTrieSync):
 
+    def __init__(self, root_hash, db):
+        super().__init__(root_hash, db, logging.getLogger("p2p.state.StateSync"))
```
pipermerriam: Note to self:

Might be nice to drop the `logger` from `__init__` at some point and just have it as a class property.  Subclasses could easily override with their own logger.

p2p/state.py
gsalgado: Not really. Removed

p2p/state.py
gsalgado: Indeed, that's an offset. Went with `sleep_duration`

PR #532, opened 2018-04-05
Override supplied transaction nonce in do_call

dylanjw:
### What was wrong?

If the user supplies a nonce other than 0, do_call with fail.

### How was it fixed?

Add nonce to the do_call override parameters.

#### Cute Animal Picture


dylanjw:
While writing tests, Im grasping why you (@pipermerriam ) didn't think `do_call` shoudnt be a vm state method.  Now, trying to include a json-rpc level concept in the core api is biting me with a circular reference.  I will move the functionality added with do_call into a helper function for use by the gas estimation.

dylanjw:
The existing gas estimation test now covers the original problem with the transaction nonce: mixing the actual sender nonce with the dummy address.

These lines in the original do_call method made the non-zero nonce valid:

```
                if not hasattr(_transaction, "get_sender"):
                    _transaction.get_sender = \ 
                        lambda: to_bytes(hexstr=DEFAULT_DO_CALL_SENDER)
                    _transaction.sender = to_bytes(hexstr=DEFAULT_DO_CALL_SENDER)
```

The conditional has been removed so the sender is always replaced with the dummy address. Now the case is caught, validation will fail with the dummy sender mixed with the original sender nonce.

pipermerriam:
@dylanjw does this mean that `msg.sender` will always be `0x7760c6D8c0eA7A764F13fb5b034A574f59c1fdd4` during transaction execution?

dylanjw:
@pipermerriam It would be the hardcoded address just during gas estimation.  its not ideal, because contract code that is conditional to the transaction sender will get wrong gas estimation results.

Saying that out loud, I realize thats a problem.  The dummy address and nonce should only be set in the case of unsigned transactions, where there is no sender address.  I'll make that change.

carver:
> Saying that out loud, I realize thats a problem. The dummy address and nonce should only be set in the case of unsigned transactions, where there is no sender address. I'll make that change.

I think at the `py-evm` level, this should accept only signed transactions, or an explicit `from` override. At the `trinity` or `eth-tester` level, we can pick a default from address. Calls/gas-estimations etc should never run from a surprise sender.

pipermerriam:
Yeah, we need a different mechanism.  The `sender` needs to remain the same whether it's signed, or not.  Like @carver said, we can move the requirement to set a `from` address up to the trinity/eth-tester level.

dylanjw:
At the py-evm level, an unsigned transaction has no defined sender.

This is the problem that im trying to overcome, which I dont think is solvable with the current py-evm api without overriding the sender at some point in the stack:

1. with a locally managed account you want to do a gas estimation prior to making a signed transaction in web3.
2. eth_estimateGas call is made to eth-tester.
3. eth-test builds a transaction to send to py-evm, on which to run gas estimation.  At this point, creating a signed transaction is impossible, because eth-tester does not have the key.  So the options are to override the sender, or to send the transaction on as an unsigned transaction (no sender).  There currently exists no way to pass the sender to py-evm except by signing a transaction.

The sender override is not ideal, in either py-evm or eth-tester.  I think ideally there would be a mechanism to run computations on a set of transaction parameters the can happen outside the normal signed transaction pathway.  I believe this is what geth does.  It takes the transaction parameters, creates a temporary state pre-funding the transaction senders address, and then runs the computation circumventing some of the normal transaction computation pre-steps.

What do you thin about bringing the spoof transaction object forward where is can totally replace the signed or unsigned transaction early enough to allow the real sender to be passed through to the gas estimation without the need for the key.  Im not sure how this would look.  Could SpoofTransaction be part of the api?

carver:
> Could SpoofTransaction be part of the api?

Yes, but not explicitly. I think it's reasonable to use `SpoofTransaction` to build the transaction that's sent to `estimate_gas` here: https://github.com/ethereum/py-evm/blob/master/evm/chains/base.py#L502

It's supposed to be a drop-in replacement for a transaction, so the API shouldn't actually have to change.

evm/estimators/gas.py
```diff
@@ -25,8 +37,52 @@ def execute_plus_buffer(multiplier, state, transaction):
...
+
+                setattr(
+                    _transaction,
+                    'nonce',
+                    DEFAULT_SPOOF_NONCE)
+
+            # set the account balance of the sender to an arbitrary large
+            # amount to ensure they have the necessary funds to pay for the
+            # transaction.
+            state_db.set_balance(transaction.sender, UINT_256_MAX // 2)
```
carver: Also, I think this might hide failures we care about. Maybe we could spoof gas_price to 0 instead? If the transaction is trying to send funds in `value`, and the account doesn't have enough funds, I think that estimate gas shouldn't pretend that it succeeds.

dylanjw: Thats a good idea.  Pre-funding is what geth does, but just making the actual cost 0 seems better.

PR #534, opened 2018-04-09
Preserve order or tx logs

gsalgado:


evm/vm/transaction_context.py
```diff
@@ -11,12 +13,18 @@ class BaseTransactionContext:
...
+    _log_counter = None
 
     def __init__(self, gas_price, origin):
         validate_uint256(gas_price, title="TransactionContext.gas_price")
         self._gas_price = gas_price
         validate_canonical_address(origin, title="TransactionContext.origin")
         self._origin = origin
+        self._log_counter = itertools.count()
+
+    @property
```
pipermerriam: The side-effect nature of this property doesn't sit well with me.  Can we just make this a method `get_next_log_idx` or `get_next_log_counter`

pipermerriam:
Also looks like there's a test in the core EVM tests which needs to be updated.

pipermerriam:
Hrm, looking at the test which is failing, it *might* be good to move the `computation.log_entries` to a private property `computation._log_entries` to maintain a good distinction between public and private APIs since the property exposed the index counter which probably shouldn't be exposed externally.

gsalgado:
Makes sense. I've also added a test showing that the order of the logs are preserved when one of them comes from a child computation

Issue #538, opened 2018-04-09
Missing dependencies after fresh install


cburgdorf:
* py-evm Version: `0.2.0-alpha.14`
* OS: linux
* Python Version (python --version): `3.6.4`
* Environment (output of `pip freeze`): 

```
aiohttp==2.3.10
apipkg==1.4
asn1crypto==0.24.0
async-lru==0.1.0
async-timeout==2.0.1
attrs==17.4.0
backcall==0.1.0
bumpversion==0.5.3
cffi==1.11.5
chardet==3.0.4
coverage==4.5.1
cryptography==2.2.2
cytoolz==0.9.0.1
decorator==4.2.1
eth-bloom==1.0.0
eth-hash==0.1.2
eth-keys==0.2.0b3
eth-utils==1.0.2
execnet==1.5.0
flake8==3.5.0
hypothesis==3.44.26
idna==2.6
idna-ssl==1.0.1
ipython==6.3.1
ipython-genutils==0.2.0
jedi==0.11.1
mccabe==0.6.1
multidict==4.1.0
mypy==0.580
parso==0.1.1
pexpect==4.4.0
pickleshare==0.7.4
pluggy==0.6.0
prompt-toolkit==1.0.15
ptyprocess==0.5.2
py==1.5.3
py-ecc==1.4.2
-e git+https://github.com/ethereum/py-evm.git@3cadfc739b469bb4c24324a32851eb5a7b28a9d2#egg=py_evm
pycodestyle==2.3.1
pycparser==2.18
pycryptodome==3.6.0
pyethash==0.1.27
pyflakes==1.6.0
Pygments==2.2.0
pytest==3.1.2
pytest-asyncio==0.8.0
pytest-cov==2.5.1
pytest-logging==2015.11.4
pytest-xdist==1.18.1
rlp==0.6.0
simplegeneric==0.8.1
six==1.11.0
toolz==0.9.0
tox==2.7.0
traitlets==4.3.2
trie==1.3.2
typed-ast==1.1.0
virtualenv==15.2.0
wcwidth==0.1.7
yarl==1.1.1
```

### What is wrong?

I'm planning to improve the developer guide for py-evm / trinity. I noticed that when I do a fresh clone, create a fresh `virtualenv` and run `pip install -e . -r requirements-dev.txt` I'm left with a system that doesn't have all dependencies installed to run the entire test suite without failures.

[...]

PR #548, opened 2018-04-11
Refactor dependency handling

cburgdorf:
### What was wrong?

If one follows the current development instructions, one is left with a system that doesn't have all needed dependencies installed (See #538 )

### How was it fixed?

This PR is based on #546 because it also updates the documentation which I couldn't get to work without #546.

Basically this PR follows the pattern that @carver already uses for `eth-account`. Maybe I implemented it with a tiny twist but it seemed most logical to me.

The gist of it goes like this.

- `requirements-dev.txt` and `requirements-doc.txt` are no more
- in `setup.py` there's a dictionary of dependencies with dependencies grouped as `evm`, `p2p`, `trinity`, `tests`, `docs` and `dev`.
- the `dev` group automatically includes all others which means there is nothing else to run apart from `pip install -e .[dev]` to get all dependencies that are needed to run, develop, test and release anything related to this repository.

Let me finish with my usual disclaimer that I'm still quite new to Python and have no idea what I'm doing so please be nice in case this turns out as utter nonsense. :sweat_smile: 

#### Cute Animal Picture


docs/contributing.rst
```diff
@@ -1,20 +1,52 @@
...
+    pip install -e .[dev]
+
 
 Running the tests
 ~~~~~~~~~~~~~~~~~
 
-You can run the tests with:
+A great way to explore the code base is to run the tests.
+
+We can run all tests with:
```
pipermerriam: Probably worth an admonition here that naively running the full test suite is likely to take multiple hours to run.

setup.py
```diff
@@ -3,48 +3,86 @@
...
-    author_email='piper@pipermerriam.com',
-    url='https://github.com/ethereum/py-evm',
-    include_package_data=True,
-    py_modules=['evm', 'trinity', 'p2p'],
-    install_requires=[
+deps={
+    'evm': [
         "aiohttp>=2.3.1,<3.0.0",
         "async_lru>=0.1.0,<1.0.0",
+        "coincurve>=7.0.0,<8.0.0",
```
pipermerriam: this should not be part of the default `py-evm` install.  It requires a system dependency and we are working to avoid as many system dependencies for the base library.

setup.py
```diff
@@ -3,48 +3,86 @@
...
         "cryptography>=2.0.3,<3.0.0",
         "cytoolz>=0.9.0,<1.0.0",
         "eth-bloom>=1.0.0,<2.0.0",
+        "eth-hash[pycryptodome]",
```
pipermerriam: same, should not be part of the base `py-evm` install.

setup.py
```diff
@@ -3,48 +3,86 @@
...
         "eth-utils>=1.0.1,<2.0.0",
+        "plyvel==1.0.4",
```
pipermerriam: same, should not be part of the base `py-evm` install.

setup.py
```diff
@@ -3,48 +3,86 @@
...
+        "flake8==3.5.0",
+        "mypy>=0.580",
+        "hypothesis==3.44.26",
+        "pytest~=3.2",
+        "pytest-asyncio==0.8.0",
+        "pytest-cov==2.5.1",
+        "pytest-logging>=2015.11.4",
+        "pytest-xdist==1.18.1",
+        "pytest-watch>=4.1.0,<5",
+        "tox==2.7.0",
```
pipermerriam: This should probably be part of the `dev` requirements as the `test` requirements would be what `tox` installs.

setup.py
```diff
@@ -3,48 +3,86 @@
...
-    },
+    ],
+    'trinity': [
+            "plyvel==1.0.4",
+            "coincurve>=7.0.0,<8.0.0",
+            "web3>=4.0.0b11,<5.0.0",
+    ],
+    'test': [
+        "flake8==3.5.0",
+        "mypy>=0.580",
```
pipermerriam: this should probably be under a separate `lint` section so that in tox we can reduce the installation size for the linting and mypy runs

setup.py
```diff
@@ -3,48 +3,86 @@
...
-        ],
-    },
+    ],
+    'trinity': [
+            "plyvel==1.0.4",
+            "coincurve>=7.0.0,<8.0.0",
+            "web3>=4.0.0b11,<5.0.0",
+    ],
+    'test': [
+        "flake8==3.5.0",
```
pipermerriam: this should probably be under a separate `lint` section so that in tox we can reduce the installation size for the linting and mypy runs

tox.ini
```diff
@@ -42,13 +42,8 @@ commands=
...
     native-state-metropolis: py.test {posargs:tests/json-fixtures/test_state.py -k Metropolis}
-extras =
-    coincurve
-    database: leveldb
-    trinity: trinity, p2p
-    rpc: trinity, p2p
-    p2p: p2p
-deps = -r{toxinidir}/requirements-dev.txt
+
+deps = .[dev]
```
pipermerriam: Yous can just change this to `extras=dev` which is the slightly more semantically correct way to do this.

setup.py
cburgdorf: We can do that but considering that we cache all packages between runs anyway I wouldn't expect it to have a big impact on CI performance.

pipermerriam: Less about performance and more about isolation.  Not a big deal but it matches the convention we've used elsewhere and I think I like the distinction.

setup.py
cburgdorf: Ah, so that explains the existing grouping that I couldn't make sense of :+1: 

However, I think I haven't fully grokked that yet.

> It requires a system dependency and we are working to avoid as many system dependencies for the base library

So, if I get you right you want that the `install_requires` (the ones taken into account for anyone running `pip install py-evm`) dependencies to not include `coincurve`, `eth-hash` and `plyvel` but still want these to be included for everyone running a `pip install -e .[dev]`. Is that correct?

Do you mind elaborating a bit on that? I understand that we want to avoid system dependencies for the base library but if **not installing** these dependencies means that the evm doesn't work how would that help?

setup.py
cburgdorf: :+1:

tox.ini
cburgdorf: :+1:

setup.py
pipermerriam: > but still want these to be included for everyone running a `pip install -e .[dev]`. Is that correct?

Yes.

The reasoning for having `pip install py-evm` install an absolute minimum number of requirements (and absolute minimal system requirements) is to make the base library **very** easy to install and use as a *library* with minimal assumptions about underlying things like what crypto implementations they want to use, what database backend they want to use, etc.

pipermerriam: >  if not installing these dependencies means that the evm doesn't work how would that help?

This is not the case.  `eth-keys` ships with a pure python backend for the ecc operations.  `coincurve` is just faster.  Same with `plyvel`.  py-evm doesn't care about what database backend it uses and it does just fine with a minimal `MemoryDB`.

cburgdorf: Ah `py-evm` has existing fallbacks for these libraries in case these are missing. That clears it up! :+1:

tox.ini
carver: What was wrong with the extras as the were? They had a clean isolation of which dependencies to install for which tests.

carver: If we do it right, we shouldn't need to install the "developer" extra inside the test suite at all: https://github.com/ethereum/eth-account/blob/master/tox.ini#L32-L35

carver: One quirk that I never found an answer for: I'd *like* to specify `eth-hash` as a primary dependency in a library like `py-evm` and then specify an extra, such that `py-evm[test]` depends on `eth-hash[pycryptodome]`. The first few ways I've tried that didn't work as I expected, so I was stuck doing it as an explicit dependency outside the test extra, like:

https://github.com/ethereum/eth-utils/blob/03eabede3d55e2d00686d2b14bda0cb2148c63af/tox.ini#L14-L16

This is a tradeoff that seems like the better option, but does have a downside:
- Explicit `eth-hash[pycryptodome]` in `tox.ini`
  - Bad: We have to specify the backend of choice twice: once in the tox file and once in setup.py. Changing the preferred backend means changing it twice (or more realistically: changing it once by accident, and hunting down a confusing bug from two slightly different environments)
  - Good: We get to specify which major versions of pycryptodome that eth-hash supports **inside** eth-hash, so any changes in supported versions auto-propagate to all libraries that use eth-hash
- Specify `pycryptodome` version explicitly in `py-evm[test]`
  - Bad: if we need to change the supported versions of pycryptodome (say, they make a non-semver-compliant release, and we have to explicitly exclude some minor version), then all libraries that depend on eth-hash will be broken until we go tweak the lib.
  - Good: the backend is only specified once in `py-evm` so we eliminate the class of bugs from the first approach

Since a breakage in the first one only affects this one library (py-evm), and the second one affects *all* libraries depending on eth-hash, the first one seems the clear winner. (unless we have a strong reason to believe that one kind of breakage is orders of magnitude less likely than the other)

docs/contributing.rst
cburgdorf: Added:

>However, running the entire test suite does take a very long time so often we just want to run a subset instead, like:

tox.ini
cburgdorf: >What was wrong with the extras as the were? They had a clean isolation of which dependencies to install for which tests.

I haven't had a good mental model of the dependencies before @pipermerriam explained it further. I'm still probably lacking some important details, but here comes my second take nevertheless.

Let's start with this: I changed the `install_requires` to 

```
install_requires = deps['evm'] + deps['p2p'] + deps['trinity']
```

Because I believe anyone who runs `pip install py-evm` does expect this to install `evm`, `p2p` and `trinity` fully to the point that no further manual installs are needed. At least for now that `p2p` and `trinity` are part of this package.

That's why extras such as `p2p` or `trinity` are gone because they are simply installed as part of the regular dependencies.

However, based on the feedback that I received I separated out a new section `evm-extras`

```
    'evm-extras': [
        "coincurve>=7.0.0,<8.0.0",
        "plyvel==1.0.4",
        "eth-hash[pycryptodome]",
    ]
``` 
Previously these were separate sections (`coincurve` and `leveldb`) and  `"eth-hash[pycryptodome]"` was living under `requirements-dev.txt` (which is now gone). It made sense to me to group these under one section but maybe there are good reasons to have them separately? I think it's more convenient to have them grouped and yet people can still install separately whatever they choose.

>In the ideal setup, we shouldn't need to install the "developer" extra inside the test suite at all

Yes, I agree and in fact, the tests now only need `evm-extras` and `test` as extras whereas the linting jobs just need `lint` as extras.
   
>One quirk that I never found an answer for: I'd like to specify eth-hash as a primary dependency in a library like py-evm and then specify an extra, such that py-evm[test] depends on eth-hash[pycryptodome]

Ok, that sounds odd indeed. So, if I understand you correctly you are saying that if you specify `eth-hash` as `install_requires` you can not overwrite it in an extra with `eth-hash[pycryptodome]`?

@pipermerriam 

>Yous can just change this to `extras=dev` which is the slightly more semantically correct way to do this.

This didn't seem to work for me. The `extras =...` option seems to only work in combination with `deps=...`. As soon as you don't have `deps` anymore the `extras` didn't seem to have an effect. It worked when I did `deps = .` in combination with `extras = ...` but that seems more verbose than just going for `deps = .[x, y, z]` straight away.

Alright, hope this update makes more sense :grimacing:

setup.py
cburgdorf: Yep, it's now separated and the linting jobs also only require `lint` as extras.

setup.py
cburgdorf: Yep, moved it to `dev`. :+1:

setup.py
cburgdorf: Moved it into new `evm-extras` section (see longer comment below)

setup.py
cburgdorf: Moved it into new `evm-extras` section (see longer comment below)

tox.ini
cburgdorf: @pipermerriam now that I refactored the CI runs to not depend on the `dev` dependencies anymore the trinity tests are missing `ipython`. Is trinity supposed to have a hard dependency against ipython by default (e.g. putting it into the `trinity` section directly) or should that go into something like `trinity-extras` (somewhat reasonable I believe since one can also run trinity with `--vanilla-shell`)?

cburgdorf:
Alright, I just updated this again and made iPython a base dependency of trinity. I also added one more commit to this PR and combined `flake8` and `mypy` jobs to a common `lint` job as suggested in #551

tox.ini
carver: > I believe anyone who runs pip install py-evm does expect this to install evm, p2p and trinity fully to the point that no further manual installs are needed. At least for now that p2p and trinity are part of this package.

Fair enough. This mostly just makes me eager to split trinity out into its own repo for real (but we're not there right now).

> So, if I understand you correctly you are saying that if you specify eth-hash as install_requires you can not overwrite it in an extra with eth-hash[pycryptodome]?

That was my experience, but I don't know for sure. It was one of those things that didn't *just work* and I didn't have time to investigate if it was travis or tox or pip.

> As soon as you don't have deps anymore the extras didn't seem to have an effect.

That's super-weird and feels broken. Heck, that might have even been the cause of the previous problem. If that's trivially reproducible, that sounds like a fairly straightforward thing to file a bug against tox for.

cburgdorf: >That's super-weird and feels broken

Yep, in fact it really felt broken because it sometimes seemed to work for one run and then failed for another whereas `.deps[x,y,z]` worked reliably every single time.

>Heck, that might have even been the cause of the previous problem

Yes, that might be it.

>If that's trivially reproducible, that sounds like a fairly straightforward thing to file a bug against tox for

I can investigate that further *another time*  :grimacing:

Issue #542, opened 2018-04-10
Performance Optimizations


pipermerriam:
### What is wrong?

There is rudamentary full sync code in place and with that, we're finding performance issues.

- https://drive.google.com/open?id=1OwLcvS-ONGmgucL4Tccf5nqP2r_QCglv
- [github-mirror](https://github.com/ethereum/py-evm/files/1895559/block-import-stats.txt)

The link above is the `cProfile` output which can be loaded using the [`pstats`](https://docs.python.org/3.5/library/profile.html) standard library module to see where time is spent.

### Known "low hanging fruit"

The following things are where we're seeing the *most* time spent which can likely be optimized.

- Remove support for python 2 from `pyrlp`
- Further `pyrlp` optimizations
- Caching of the `VMState.prev_hashes` generation.

### Ways to make things faster

[Cython](http://cython.org/) looks to be a very promising way for us to do certain optimizations.  We should spend a little time doing *discovery* to see how this well this works and what kind of performance gains we can get.

gsalgado:
Another thing I've just noticed is that after applying a transaction we seem to throw away objects representing the state and then re-read (and decode) them from the DB in order to apply the next block's tx. One example is the block's receipts: ISTM that for every tx we append the generated receipts to `BaseState.receipts`, but then for the next transaction we re-read all those receipts from the DB. Would it be possible to preserve those `Receipt` instances (and maybe other things?) across transaction boundaries?

pipermerriam:
> Would it be possible to preserve those Receipt instances (and maybe other things?) across transaction boundaries?

Yes, though I'd like to try and accomplish this with caching as I like the way we currently have things with minimal statefullness.

gsalgado:
So, with all the changes we made to py-evm/py-rlp/py-trie things got significantly better, but `JournalDB` is still the biggest offender and makes block importing just plain unfeasible (because it does an extra `LevelDB.get()` every time we update/delete a key) on a fully-synced ropsten DB. Here's a 
[new stats file](https://github.com/ethereum/py-evm/files/1908308/block-import-stats.txt) that aggregates the profiling stats of only `Chain.import_block()` for 8 consecutive blocks (unlike the previous one, which was profiling the networking code as well).

[...]

PR #549, opened 2018-04-11
Use a lru cache for MainAccountStateDB._get_account()

gsalgado:
Part of the fixes for #542

### DIFF
--- a/evm/db/state.py
+++ b/evm/db/state.py
@@ -3,6 +3,8 @@ from abc import (
     abstractmethod
 )
 
+from lru import LRU
+
 import rlp
 
 from trie import (
@@ -20,6 +22,7 @@ from evm.constants import (
 from evm.db.immutable import (
     ImmutableDB,
 )
+from evm.exceptions import DecommissionedStateDB
 from evm.rlp.accounts import (
     Account,
 )
@@ -37,6 +40,12 @@ from evm.utils.padding import (
 )
 
 from .hash_trie import HashTrie
+
+
+# Use lru-dict instead of functools.lru_cache because the latter doesn't let us invalidate a single
+# entry, so we'd have to invalidate the whole cache in _set_account() and that turns out to be too
+# expensive.
+account_cache = LRU(2048)
 
 
 class BaseAccountStateDB(metaclass=ABCMeta):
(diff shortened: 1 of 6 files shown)


setup.py
```diff
@@ -26,6 +26,7 @@
         "rlp>=0.4.7,<1.0.0",
         "eth-keys>=0.2.0b3,<1.0.0",
         "trie>=1.3.2,<2.0.0",
+        "pylru>=1.0.9",
```
pipermerriam: It looks like https://pypi.python.org/pypi/lru-dict might be worth evaluating as it's implemented in C.

evm/db/state.py
```diff
@@ -294,7 +316,13 @@ def account_is_empty(self, address):
     # Internal
     #
     def _get_account(self, address):
-        rlp_account = self._trie[address]
+        cache_key = (self._unwrapped_db, self.root_hash, address)
+        if cache_key in account_cache:
+            rlp_account = account_cache[cache_key]
+        else:
+            rlp_account = self._trie[address]
+            account_cache[cache_key] = rlp_account
```
pipermerriam: I've been a fan of the following pattern for cached functions.  I find it easier to parse.

```python
def cached_fn():
    if key not in cache:
        cache[key] = get_value()

    value = cache[key]
```

It's a minor improvement but it's a smaller conditional block to parse and the *value* always comes from the same place.

setup.py
gsalgado: For this limited one-block-import test I'm running it doesn't seem to provide any benefits. I guess it depends how large this cache can get. Maybe just leave a comment there to evaluate it properly if `_get_account()` shows up again as one of the worst offenders in cProfile stats?

evm/db/state.py
gsalgado: I like that!

setup.py
carver: :-1: `pylru` is GPL https://github.com/jlhutch/pylru/blob/master/LICENSE.txt

gsalgado: Should we go with lru-dict, then?

pipermerriam: @gsalgado I think yes, (that's what web3.py is using at least).

PR #550, opened 2018-04-11
Add an lru cache to BaseVM.get_prev_hashes()

gsalgado:
Not sure what maxsize to use, but assuming that method is used most of the time
to lookup the state of the current head, I think a large maxsize won't provide any benefit. The main motivation for the cache was actually to reduce the number of times we perform this during the importing of a single block, and it reduced the total time by 30% in that case.

Part of the fixes for #542

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -8,6 +8,8 @@ import contextlib
 import logging
 import rlp
 
+import functools
+
 from eth_utils import (
     keccak,
     to_tuple,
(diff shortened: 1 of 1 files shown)


evm/vm/base.py
```diff
@@ -364,6 +366,7 @@ def get_block_by_header(cls, block_header, db):
 
     @classmethod
     @to_tuple
+    @functools.lru_cache(maxsize=32)
```
pipermerriam: I think you need to elevate this to above the `to_tuple` call.

Issue #499, opened 2018-03-26
Create new eth_typing lib


cburgdorf:
This is a follow up of [this discussion](https://github.com/ethereum/py-evm/pull/459#discussion_r174853525)

### What is wrong?

We currently have many API that take/return `bytes` [1] where we could really benefit from stricter type checking by introducing [`NewType`](http://mypy.readthedocs.io/en/latest/kinds_of_types.html#newtypes)s. In short a `NewType` lets us create a substitute for a type (e.g. `miles` for `int`) so that an API that takes `miles` as inputs can not be called with any arbitrary `int` instead. These kind of errors already killed NASA missions when `km` and `miles` where [accidentally twisted](http://edition.cnn.com/TECH/space/9909/30/mars.metric.02/)

[1] also applies to other types (e.g. `int`) but is mostly problematic with `bytes` in our case

### How can it be fixed

1. Create NewTypes for types that we currently define to loosely (mostly `bytes` but we should check also our usage of `int`)

2. Put them in a new library `eth_typings` that can be distributed on `pypi` so that it can be shared across different projects

pipermerriam:
I think we should create this locally here in the py-evm repo and then migrate it to something separate once we have something semi-stable.  Feel free to do this as a new top-level module `eth_typing` like `evm/p2p/trinity`.

PR #554, opened 2018-04-12
Convert addresses and hashes to NewTypes

cburgdorf:
### What was wrong?

As discussed in #499 for certain usages of `bytes` (and potentially others) we can benefit from introducing a NewType so that the type system forbids calling an API that expects an `Address` with just any other arbitrary bytes.

### How was it fixed?

Implemented `Address` and `Hash32` NewType and used where appropriate. 

#### Cute Animal Picture

### DIFF
--- a/evm/rlp/headers.py
+++ b/evm/rlp/headers.py
@@ -12,6 +12,10 @@ from cytoolz import (
     first,
     sliding_window,
 )
+from eth_typing import (
+    Address,
+    Hash32
+)
 from eth_utils import (
     keccak,
     to_dict,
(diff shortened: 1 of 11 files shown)


carver:
I don't feel strongly, but it might be more consistent to call it `Hash256` for the number of bits instead of 32 for bytes. My reference point is `keccak256`.

cburgdorf:
>it might be more consistent to call it `Hash256` for the number of bits instead of `32`

I don't have any preference about that. Other opinions anyone?

pipermerriam:
I 👍 what @carver said, but I'm not sure it's actually right.  I'm more in favor of the `32` suffix now that I've thought about it for more than 2 seconds.  It's more inline with our other naming conventions.

evm/rlp/headers.py
```diff
@@ -83,15 +86,15 @@ def __init__(self,
                  gas_limit: int,
                  timestamp: int=None,
                  coinbase: bytes=ZERO_ADDRESS,
```
hwwhww: Should it be `coinbase: Address=ZERO_ADDRESS,` ?

cburgdorf: Thank you for spotting this. Fixed it!

cburgdorf:
Alright, keeping the name `Hash32` then.

carver:
> more inline with our other naming conventions.

Which conventions? (curiosity more than disagreement)

PR #555, opened 2018-04-13
Blob serialization

jannikluhn:
Implement blob serialization as described here: https://ethresear.ch/t/blob-serialisation/1705

Flags are not yet implemented (they are set to 0 when writing and ignored when reading) because I can't see right now what the interface would look like ideally. We should revisit this when state execution is implemented.

#### Cute Animal Picture

### DIFF
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -21,7 +21,7 @@ from cytoolz import (
 )
 
 
-def chunk_iterator(collation_body: bytes) -> Iterator[bytes]:
+def iterate_chunks(collation_body: bytes) -> Iterator[bytes]:
     if len(collation_body) % CHUNK_SIZE != 0:
         raise ValueError("Blob size is {} which is not a multiple of chunk size ({})".format(
             len(collation_body),
@@ -60,5 +60,5 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
             COLLATION_SIZE
         ))
 
-    chunks = chunk_iterator(collation_body)
+    chunks = iterate_chunks(collation_body)
     return calc_merkle_root(chunks)
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -49,11 +49,11 @@ def hash_layer(layer: Iterable[bytes]) -> Iterator[bytes]:
 def calc_merkle_root(leaves: Iterable[bytes]) -> bytes:
     leaves = list(leaves)  # convert potential iterator to list
     if len(leaves) == 0:
-        raise ValueError("No leaves given")
+        raise ValidationError("No leaves given")
 
     n_layers = math.log2(len(leaves))
     if not n_layers.is_integer():
-        raise ValueError("Leave number is not a power of two")
+        raise ValidationError("Leave number is not a power of two")
     n_layers = int(n_layers)
 
     first_layer = (keccak(leaf) for leaf in leaves)
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -106,7 +106,7 @@ def serialize_blobs(blobs: Iterable[bytes]) -> Iterator[bytes]:
                 length_bits = 0
 
             flag_bits = 0  # TODO: second parameter? blobs as tuple `(flag, blob)`?
-            indicator_byte = int_to_big_endian(length_bits | flag_bits * 0b00100000)
+            indicator_byte = int_to_big_endian(length_bits | (flag_bits << 5))
             assert len(indicator_byte) == 1
 
             yield indicator_byte
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -78,7 +78,7 @@ def check_body_size(body):
 @apply_to_return_value(check_body_size)
 @apply_to_return_value(lambda v: zpad_right(v, COLLATION_SIZE))
 @apply_to_return_value(b"".join)
-def serialize_blobs(blobs: Iterable[bytes]) -> bytes:
+def serialize_blobs(blobs: Iterable[bytes]) -> Iterator[bytes]:
     """Serialize a sequence of blobs and return a collation body."""
     for blob in blobs:
         if len(blob) == 0:
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -36,11 +36,7 @@ from cytoolz import (
 
 
 def iterate_chunks(collation_body: bytes) -> Iterator[bytes]:
-    if len(collation_body) % CHUNK_SIZE != 0:
-        raise ValueError("Blob size is {} which is not a multiple of chunk size ({})".format(
-            len(collation_body),
-            CHUNK_SIZE,
-        ))
+    check_body_size(collation_body)
     for chunk_start in range(0, len(collation_body), CHUNK_SIZE):
         yield collation_body[chunk_start:chunk_start + CHUNK_SIZE]
 
@@ -68,18 +64,13 @@ def calc_merkle_root(leaves: Iterable[bytes]) -> bytes:
 
 
 def calc_chunk_root(collation_body: bytes) -> bytes:
-    if len(collation_body) != COLLATION_SIZE:
-        raise ValueError("Blob is {} instead of {} bytes in size".format(
-            len(collation_body),
-            COLLATION_SIZE
-        ))
-
+    check_body_size(collation_body)
     chunks = iterate_chunks(collation_body)
     return calc_merkle_root(chunks)
 
 
 def check_body_size(body):
-    if len(body) > COLLATION_SIZE:
+    if len(body) != COLLATION_SIZE:
         raise ValidationError("{} byte collation body exceeds maximum allowed size".format(
             len(body)
         ))
--- a/evm/utils/blobs.py
+++ b/evm/utils/blobs.py
@@ -7,7 +7,12 @@ from typing import (
 )
 
 from eth_utils import (
+    apply_to_return_value,
+    int_to_big_endian,
     keccak,
+)
+from evm.utils.padding import (
+    zpad_right,
 )
 
 from evm.constants import (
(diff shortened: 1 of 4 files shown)


evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+def check_body_size(body):
+    if len(body) > COLLATION_SIZE:
+        raise ValueError("Collation body exceeds maximum allowed size")
+    return body
+
+
+@apply_to_return_value(check_body_size)
+@apply_to_return_value(lambda v: zpad_right(v, COLLATION_SIZE))
+@apply_to_return_value(b"".join)
+def serialize_blobs(blobs: Iterable[bytes]) -> bytes:
```
cburgdorf: @jannikluhn You probably annotated the return value as `bytes` because that is the type that the caller of `serialize_blobs` sees. At least, this confused me a lot when I first came across something like this because something inside me naturally wanted to annotate the return type as the type it is **after all decorators are applied**. However, you do need to make the return type `Iterator[bytes]` and still on the call site `mypy` will infer the correct type (e.g. `bytes` in this case) based on the applied decorators.

jannikluhn: It confused me indeed, only realized this when mypy was failing, but in hindsight it makes sense.

> and still on the call site mypy will infer the correct type (e.g. bytes in this case) based on the applied decorators.

Is this true even with the unannotated lambda function in the decorator or should I replace it with a `def` function (or probably better `partial(zpad_right, to_size=COLLATION_SIZE)`)?

cburgdorf: >It confused me indeed, only realized this when mypy was failing, but in hindsight it makes sense

You have no idea how much better I feel now, knowing I wasn't the only one to think like that ;)

>Is this true even with the unannotated lambda function in the decorator

In general, I do believe that the type inference would work out the lamda just fine. However, in our case the type system will still fall back to `Any` (yeah, bummer!) which I *believe* can only be fixed through a stronger [`apply_to_return_value`](https://github.com/ethereum/eth-utils/blob/a1a2ee20988f56f6f99e31e2eff4b98e10c35814/eth_utils/functional.py#L18-L25) decorator.

I think with the current decorator as it stands, whatever you apply it on will make the type system blind to figure out the proper return type.

cburgdorf: Maybe we can do something about it https://github.com/ethereum/eth-utils/pull/88

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
         ))
 
-    chunks = chunk_iterator(collation_body)
+    chunks = iterate_chunks(collation_body)
     return calc_merkle_root(chunks)
+
+
+def check_body_size(body):
+    if len(body) > COLLATION_SIZE:
+        raise ValueError("Collation body exceeds maximum allowed size")
```
pipermerriam: This looks like a `ValidationError` might be more appropriate?

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+    return body
+
+
+@apply_to_return_value(check_body_size)
```
pipermerriam: ooh, fancy use of `apply_to_return_value`

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+@apply_to_return_value(lambda v: zpad_right(v, COLLATION_SIZE))
+@apply_to_return_value(b"".join)
+def serialize_blobs(blobs: Iterable[bytes]) -> Iterator[bytes]:
+    """Serialize a sequence of blobs and return a collation body."""
+    for blob in blobs:
+        if len(blob) == 0:
+            raise ValueError("Cannot serialize blob of length 0")
```
pipermerriam: Same here with raising a validation error.  Also, it might be nice to indicate the index of the invalid blob.

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+        if len(blob) > 31 * 2**((COLLATION_SIZE - 1).bit_length() - 5):
```
pipermerriam: This inlined *magic* number stands out as something that belongs in a constant and with a comment indicating where it comes from.

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+            raise ValueError("Cannot serialize blob of size {}".format(len(blob)))
+
+        blob_index = 0
```
pipermerriam: I assume initializing this value to zero here is for the case where `len(blob) < 31` which causes the loop below to not run, thus `blob_index` never gets initialized. Maybe we could handle this case explicitely instead?

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+                length_bits = 0
+
+            flag_bits = 0  # TODO: second parameter? blobs as tuple `(flags, blobs)`?
+            indicator_byte = int_to_big_endian(length_bits | flag_bits * 0b00100000)
+            assert len(indicator_byte) == 1
+
+            yield indicator_byte
+            yield blob[blob_index:blob_index + 31]
+
+        chunk_filler = b"\x00" * (31 - (len(blob) - blob_index))
```
pipermerriam: This *implicit* use of whatever the last `blob_index` value was from the loop above seems both hard to parse, as well as a fragile coupling between the loop logic and this logic.  Maybe you could compute this value independent of the loop so that refactors to how the above loop works don't accidentally break this code?

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+        length_bits = indicator_byte & 0b00011111
+
+        if length_bits == 0:
+            length = 31
+            terminal = False
+        else:
+            length = length_bits
+            terminal = True
+
+        blob += chunk[1:length + 1]
```
pipermerriam: Instead of using `+=` here, you might consider using an `io.BytesIO` for the `blob` variable.  That will allow you to do `blob.write(chunk[1:lenght+1])` and then use `blob.getvalue()` later when you want to `yield` it's value.

tests/core/test_blob_utils.py
```diff
@@ -22,25 +27,38 @@
...
 
+BLOB_SERIALIZATION_TEST_DATA = [  # [(blobs, unpadded_body), ...]
+    ([], b""),
+    ([b"\x00"], b"\x01\x00"),
+    ([b"\x01"], b"\x01\x01"),
+    ([b"\xaa" * 31], b"\x1f" + b"\xaa" * 31),
+    ([b"\xaa" * 32], b"\x00" + b"\xaa" * 31 + b"\x01" + b"\xaa"),
+    (
+        [b"\xaa" * 20, b"\xbb" * 15],
+        b"\x14" + b"\xaa" * 20 + b"\x00" * 11 + b"\x0f" + b"\xbb" * 15
```
pipermerriam: These test cases are really opaque.  Maybe you could split these up across multiple lines like this.

```python
(
    [
        b"\xaa" * 20,  # first blob
        b"\xbb" * 15,  # second blob
    ],
    (
        b"\x14"  # whatever this value is
        b"\xaa * 20"  # the first blob body
        b"\x00 * 11"  # padding
        b"\x0f"  # whatever this value is
        b"\xbb" * 15  # the second blob body
    )
)
```

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+        for blob_index in range(0, len(blob), 31):
+            remaining_blob_bytes = len(blob) - blob_index
+
+            if remaining_blob_bytes <= 31:
+                length_bits = remaining_blob_bytes
+            else:
+                length_bits = 0
+
+            flag_bits = 0  # TODO: second parameter? blobs as tuple `(flags, blobs)`?
+            indicator_byte = int_to_big_endian(length_bits | flag_bits * 0b00100000)
```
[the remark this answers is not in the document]
jannikluhn: I multiply by `0b00100000` to move the flags to the beginning of the byte. `flag_bits << 5` is probably more clear, I'll change it.

evm/utils/blobs.py
```diff
@@ -60,5 +65,63 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+            assert len(indicator_byte) == 1
+
+            yield indicator_byte
+            yield blob[blob_index:blob_index + 31]
+
+        chunk_filler = b"\x00" * (31 - (len(blob) - blob_index))
+        yield chunk_filler
+
+
+def iterate_blobs(body: bytes) -> Iterator[bytes]:
```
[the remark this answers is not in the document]
jannikluhn: What's wrong with `iterate_blobs`? It expresses that one should use it to get blobs and that it returns an iterator instead of a list. `deserialized_blobs` looks to me more like a variable name (e.g., `deserialized_blobs = list(iterate_blobs(body))`.

evm/utils/blobs.py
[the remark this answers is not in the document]
hwwhww: oh-oh sorry I somehow read it as `&`, thanks for clarifying!

evm/utils/blobs.py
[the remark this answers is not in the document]
hwwhww: Sorry, I should say `deserialize_blobs` for pairing with `serialize_blobs`.

evm/utils/blobs.py
```diff
@@ -60,5 +74,68 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+            indicator_byte = int_to_big_endian(length_bits | (flag_bits << 5))
+            assert len(indicator_byte) == 1
+
+            yield indicator_byte
+            yield blob[blob_index:blob_index + CHUNK_DATA_SIZE]
+
+        # end of range(0, N, k) is given by the largest multiple of k smaller than N, i.e.,
+        # (ceil(N / k) - 1) * k where ceil(N / k) == -(-N // k)
+        last_blob_index = (-(-len(blob) // CHUNK_DATA_SIZE) - 1) * CHUNK_DATA_SIZE
+        assert last_blob_index == blob_index
```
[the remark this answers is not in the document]
mhchia: And why do we need to check for `assert last_blob_index == blob_index`, it seems definite?

evm/utils/blobs.py
```diff
@@ -60,5 +74,68 @@ def calc_chunk_root(collation_body: bytes) -> bytes:
...
+            remaining_blob_bytes = len(blob) - blob_index
+
+            if remaining_blob_bytes <= CHUNK_DATA_SIZE:
+                length_bits = remaining_blob_bytes
+            else:
+                length_bits = 0
+
+            flag_bits = 0  # TODO: second parameter? blobs as tuple `(flag, blob)`?
+            indicator_byte = int_to_big_endian(length_bits | (flag_bits << 5))
+            assert len(indicator_byte) == 1
```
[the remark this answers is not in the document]
jannikluhn: I like to use `assert`s to check for invariants (if the code is correct the assert will never fail). So it's basically not much more than a comment making the function easier to understand. I guess an exception would work too, but they make the code a little harder to read (requires an `if` statement) and it's unclear which exception type to use.

evm/utils/blobs.py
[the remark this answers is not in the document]
jannikluhn: See above comment.

jamesray1:
I'm thinking about importing this into Rust rather than rewriting this in Rust. Would you say that the code is memory safe?

cburgdorf:
>I'm thinking about importing this into Rust rather than rewriting this in Rust

Notice that this code has dependencies against `eth_utils` and `cytoolz` so I'm not sure if importing it in Rust will be as straight forward as you might have assumed. But I have never imported Python code into Rust before so I honestly don't know.

Anyway I think the bulk of the code looks small enough to just reimplement in Rust plus you'll most likely see the best performance and safety guarantees when you write it in Rust.

>Would you say that the code is memory safe?

I honestly can't tell.

jamesray1:
Yeah I realized that you were importing stuff, I'm not sure how that would work with importing into Rust.

I've already started implementing it in Rust, but did want to consider importing it as we do not have enough developers and a lot to do.

PR #556, opened 2018-04-14
Remove do_call method, extend SpoofTransaction to work with unsigned transactions

dylanjw:
### What was wrong?
`do_call` method used a surprise sender to accomplish calls and gas estimation when the signing key is not available.

### How was it fixed?
SpoofTransaction will inject a `sender` attribute when used with an unsigned transaction.  This behaviour requires the parameter `from_` to be supplied.

~~`intrinsic_gas` and `get_intrinsic_gas` where moved from the transaction class to the transaction_context class.~~

EDIT: I also made the transaction executor classes stand-alone rather than a mix-in for the state class.  This made it a little simpler to manage the `transaction_context`, `transaction`, `message` and `computation` objects during the transaction execution steps as instance attributes.

#### Cute Animal Picture


evm/vm/state.py
```diff
@@ -407,24 +369,32 @@ def get_uncle_reward(block_number, uncle):
...
+            data=self.transaction.data,
+        )
+        self.vm_state = vm_state
+
+        self.message = self.build_transaction_message()
+        self.computation = self.finalize_computation(self.build_computation())
 
     @abstractmethod
-    def run_pre_computation(self, transaction):
+    def build_transaction_message(self):
```
pipermerriam: This name feels slightly off.  The `message` object is not really the *transaction* message, but the *evm* message so this might be better named `build_evm_message` or even just `build_message.`

tests/core/chain-object/conftest.py
```diff
@@ -7,3 +11,14 @@
     chain as valid_chain,
     chain_without_block_validation as chain,
 )
+from eth_keys import KeyAPI
+
+
+@pytest.fixture
+def unfunded_address():
```
pipermerriam: Assuming this is the address of the private key from `unfunded_address_private_key` how about changing this to return `unfunded_address_private_key.public_key.to_canonical_address()`

dylanjw:
I made the assumption that having `computation.get_gas_used()` include the intrinsic gas would be ok, since the only `get_gas_used` reference I found was to calculate the total transaction gas used.  Is this an ok approach, or am I mixing together separate conceptual domains?

pipermerriam:
@dylanjw I think inclusion of the intrinsic gas concept within the `TransactionContext` is appropriate and does not violate any separation of concerns that come to mind 👍

evm/estimators/gas.py
```diff
@@ -50,20 +57,36 @@ def binary_gas_search(state, transaction, tolerance=1):
...
+        raise TypeError(
+            "Transaction is missing attribute sender.",
+            "If sending an unsigned transaction, use SpoofTransaction and provide the",
+            "sender using the 'from' parameter")
+
+    transaction_context = state.get_transaction_context_class()(
+        gas_price=0,
+        origin=transaction.sender,
+        data=transaction.data,
+    )
```
carver: It seems like removing `intrinsic_gas()` from the transaction means we have to do a lot of setup here for the transaction_context, which is only used to get the intrinsic_gas back. What was the payoff for moving it?

I think that the intrinsic cost of the transaction should not be semantically part of the computation. It seems like the Transaction itself is the most natural place to define the intrinsic gas in a new fork. Also, what if it's a transaction with no computation at all? Would we have to build an empty computation or TransactionContext just to get the gas cost of an empty transaction?

evm/vm/forks/frontier/state.py
```diff
@@ -49,96 +50,97 @@
 
 
 class FrontierTransactionExecutor(BaseTransactionExecutor):
-    def run_pre_computation(self, transaction):
+    def build_evm_message(self):
```
carver: What's the value here of making this rely on the internal state of `self.transaction` instead of a passed in version?

evm/vm/forks/frontier/state.py
```diff
@@ -214,11 +217,12 @@ def _make_frontier_receipt(state, transaction, computation):
     return receipt
 
 
-class FrontierState(BaseState, FrontierTransactionExecutor):
+class FrontierState(BaseState):
```
carver: Separating these seems very reasonable :+1:

evm/vm/transaction_context.py
```diff
@@ -15,12 +19,13 @@ class BaseTransactionContext:
...
     _log_counter = None
 
-    def __init__(self, gas_price, origin):
+    def __init__(self, gas_price, origin, data):
         validate_uint256(gas_price, title="TransactionContext.gas_price")
         self._gas_price = gas_price
         validate_canonical_address(origin, title="TransactionContext.origin")
         self._origin = origin
         self._log_counter = itertools.count()
+        self.data = data
```
carver: Seems odd to me to add this data here. I guess I don't totally get what TransactionContext is, because it seems like it's just adding more and more things until it eventually just becomes a copy of Transaction. Why not use the original Transaction directly?

dylanjw: @carver
The `data` field was required to make the intrinsic gas calculation.  Adding the intrinsic gas to the transaction context does create an awkward api, as you pointed out.

The move of the intrinsic gas to the transaction context began with @pipermerriam suggestion after I duplicated some BaseTransaction attributes in BaseUnsignedTransaction.  Maybe it would make sense to have a shared abstract class between BaseUnsignedTransaction and BaseTransaction containing the required attributes to make the gas estimation with unsigned transactions.  Or come up with a clever way to add the intrinsic gas to SpoofTransaction.

pipermerriam: I think I'm going to flip-flop my opinion on this.

The purpose of `TransactionContext` is to provide the transaction level data necessary for vm execution, i.e. the `gas_price` and the `origin` address.

@dylanjw I think a common base for the transaction classes which implements the intrinsic gas calculations should be fine.  I don't think there is anything in the signed vs unsigned that restricts either from being able to compute the intrinsic gas costs.

evm/estimators/gas.py
```diff
@@ -9,12 +9,18 @@
...
-    computation = state.do_call(transaction)
 
-    if computation.is_error:
-        return computation._error
-    else:
-        return None
+    snapshot = state.snapshot()
+
+    try:
+        computation = state.execute_transaction(transaction)
```
pipermerriam: I like this quite a bit.

evm/vm/forks/frontier/state.py
```diff
@@ -143,42 +154,44 @@ def run_post_computation(self, transaction, computation):
...
-            self.account_db.set_balance(account, 0)
-            self.account_db.delete_account(account)
+            self.vm_state.account_db.set_balance(account, 0)
+            self.vm_state.account_db.delete_account(account)
 
         return computation
 
 
-class FrontierState(BaseState, FrontierTransactionExecutor):
+class FrontierState(BaseState):
```
pipermerriam: I like this a lot more.  The inheritance piece there felt odd, adding extra methods to the `State`.

evm/vm/state.py
```diff
@@ -195,34 +187,6 @@ def get_block_class(cls) -> Type['BaseBlock']:
...
-        try:
-            if not hasattr(_transaction, "get_sender"):
-                _transaction.get_sender = \
-                    lambda: to_bytes(hexstr=DEFAULT_DO_CALL_SENDER)
-                _transaction.sender = to_bytes(hexstr=DEFAULT_DO_CALL_SENDER)
-
-            # set the account balance of the sender to an arbitrary large
-            # amount to ensure they have the necessary funds to pay for the
-            # transaction.
-            self.account_db.set_balance(transaction.sender, UINT_256_MAX // 2)
```
pipermerriam: This looks like it was removed and not added back in to ensure gas estimations work from unfunded accounts.  Did I miss something?

evm/vm/state.py
```diff
@@ -287,24 +251,44 @@ def get_uncle_reward(block_number, uncle):
     def get_nephew_reward(cls):
         raise NotImplementedError("Must be implemented by subclasses")
 
+    def get_transaction_executor(self, transaction):
+        vm_state = self
```
pipermerriam: This is odd to me.  can we just use `self` on the next line?

dylanjw: Sure. I added it when there were more arguments passed in, which made it look weird to me:

```
return self.transaction_executor(self, transaction)
```

evm/vm/state.py
dylanjw: Rather than pre-funding the address, Im setting the gasPrice to 0, per @carver suggestion.  It will have different behavior than other implementations.

For example go-ethereum will let you check the gas estimation on a transaction that could never occur, like a value transfer from an account with insufficient funds.
This implementation will throw an error in such a case, but in other cases will still let you estimate the gas without having sufficient funds to pay the gas.

PR #558, opened 2018-04-16
Store collations and headers by hashes

jannikluhn:
### What was wrong?

Headers were stored in the database using shard id and period as key. This prevents storing multiple collation(-header)s for the same period and shard, which is required at least temporarily during sync as pointed out by @hwwhww.

### How was it fixed?

Headers and their availabilities are stored using their hash, bodies using their chunk root as key. For each shard and period a canonical hash is stored, identifying the header actually included in the chain.


evm/db/shard.py
```diff
@@ -35,75 +36,102 @@ def __init__(self, db):
...
+        key = make_collation_availability_lookup_key(chunk_root)
         try:
             available = bool(rlp.decode(self.db.get(key), big_endian_int))
         except KeyError:
             return False
         return available
 
-    def is_unavailable(self, shard_id: int, period: int) -> bool:
-        key = make_collation_availability_lookup_key(shard_id, period)
+    def is_unavailable(self, chunk_root: bytes) -> bool:
```
hwwhww: Just use `not self.is_available(chunk_root)`:
```python
def is_unavailable(self, chunk_root: bytes) -> bool:
    return not self.is_available(chunk_root)
```

jannikluhn: There's a slight difference: A collation can both be `not available` and `not unavailable`, but `availability_unknown`. I think this is important because we need to be able to distinguish between "I tried downloading the body but no one knew about it" and "I just got that header and haven't downloaded the body yet".

But I get that the names are confusing. Would it be better to have a single function `get_availability` that returns an enum `Availability` with values `available`, `unavailable` and `unknown`?

hwwhww: Oh, I see. Agree with enum Availability. 👍

evm/db/shard.py
```diff
@@ -35,75 +36,102 @@ def __init__(self, db):
         self.db = db
 
     #
-    # Collation and header API
+    # Collation Getters by Hash
     #
-    def get_header(self, shard_id: int, period: int) -> CollationHeader:
-        key = make_collation_header_lookup_key(shard_id, period)
+    def get_header_by_hash(self, collation_hash: bytes) -> CollationHeader:
```
pipermerriam: link to #554 since I think that would introduce the new `Hash32` type for this (depending on what gets merged first.)

evm/db/shard.py
```diff
@@ -35,75 +36,102 @@ def __init__(self, db):
...
-        body = self.get_body(shard_id, period)
+    def get_collation_by_hash(self, collation_hash: bytes) -> Collation:
+        header = self.get_header_by_hash(collation_hash)
+        body = self.get_body_by_chunk_root(header.chunk_root)
         return Collation(header, body)
 
+    #
+    # Canonical Collations
+    #
+    def mark_canonical(self, header: CollationHeader) -> None:
```
pipermerriam: maybe use `set_canonical_collation` to stay more inline with the `get/set` naming convention.

evm/db/shard.py
```diff
@@ -35,75 +36,102 @@ def __init__(self, db):
...
+        key = make_canonical_hash_lookup_key(header.shard_id, header.period)
+        self.db.set(key, header.hash)
+
+    def get_canonical_hash(self, shard_id: int, period: int) -> bytes:
```
pipermerriam: maybe `get_canonical_collation_hash` since it's somewhat ambiguous what `hash` refers to here?

evm/db/shard.py
pipermerriam: You *could* get by with `True/False/None` here if you wan't to avoid the minor mental overhead of the enum, but I think the enum likely carries more semantic meaning so 👍

Issue #561, opened 2018-04-16
Refactor JournalDB


pipermerriam:
### What is wrong?

The current implementation of `JournalDB` works by keeping track of the  *original* values in the database so that when a *revert* occurs, it can restore all of the original values.

This causes a lot of read overhead since every *set* requires reading the existing value.

### How can it be fixed

Lets invert this model.  Instead of tracking existing values, lets instead, avoid writing to the database until a commit occurs.  This means that the journal will now effectively be a cache as well as allowing batch committing of values (this probably also means removing the `BatchDB` implementation since `JournalDB` will be able to do everything that `BatchDB` can currently do.

- On `set` and `delete` operations, the journal should now store the value that was written.
- When committing a checkpoint, all of it's values should be merged into the previous checkpoint.
- When reverting a checkpoint, we can just throw away all of the changes from that checkpoint and all subsequent checkpoints.
- If the final checkpoint is committed, all changes should be written to the underlying database.

pipermerriam:
link #542

PR #563, opened 2018-04-17
Refactor JournalDB for effiency

cburgdorf:
### What was wrong?

See #561 

### How was it fixed?

This is an incomplete first draft. Just means: I'm on it.

#### Cute Animal Picture

### DIFF
--- a/evm/db/cached_journal.py
+++ b/evm/db/cached_journal.py
@@ -1,0 +1,231 @@
+from typing import Dict, List  # noqa: F401
+import uuid
+
+from cytoolz import (
+    merge,
+)
+
+from evm.db.backends.base import BaseDB
+from evm.exceptions import ValidationError
+
+
+class Journal(object):
+    """
+    A Journal is an ordered list of checkpoints.  A checkpoint is a dictionary
+    of database keys and values.  The values are the "original" value of that
+    key at the time the checkpoint was created.
+
+    Checkpoints are referenced by a random uuid4.
+    """
+    checkpoints = None  # type: List[uuid.UUID]
+
+    def __init__(self):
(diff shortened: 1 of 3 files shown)


cburgdorf:
@pipermerriam I started working on #561 but I think I hit a bit of a conceptional road block.

In the old model of the `JournalDB` we could do the following:

1. Do changes
2. Take `Snapshot 1`
3. Do changes
4. Take `Snapshot 2`
3. Do changes
4. Take `Snapshot 3`
5. Commit `Snapshot 3` (merged `Snapshot 3` into `Snapshot 2`
6. Revert to `Snapshot 1`

Now the underlying DB would be rolled back to the exact state of `Snapshot 1`. I added a [test case](https://github.com/ethereum/py-evm/pull/563/commits/bbe63061994844e6a62444aa79893b00e8302655) for that to demonstrate things with code.

In the new proposed model, if I'm not mistaken, we can revert back and forth between different snapshots but at the point where we commit `Snapshot 3` there's no turning back to `Snapshot 1` anymore.

This is because Snapshots (as I understood your proposal) do not track the original values anymore but rather track the *changes* (which we write to the underlying db only when commited).

Of course we could say that when we commit `Snapshot 3` we are basically throwing away all previous snapshots and start over but while that keeps the system sound it still means we are losing functionality that was previously perfectly fine.

Notice that the PR is still WIP and the `CachedJournalDB`  does not even write back to the underlying db yet. I just stumbled over this and wanted to reassure if I'm on the right track.

cburgdorf:
Ah wait, actually, just thought through it again and noticed I might be wrong here. Need to give it more thought later. Currently on mobile.

PR #565, opened 2018-04-17
Redesign receipt/tx trie generation on block imports

gsalgado:
We used to generate the tries for all block's transactions/receipts
after applying every transaction, but that is way too expensive. Now the
trie data is only generated once, after we've applied all transactions

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -3,12 +3,12 @@ from abc import (
     ABCMeta,
     abstractmethod
 )
-
 import contextlib
-import logging
-import rlp
-
 import functools
+import logging
+from typing import List  # noqa: F401
+
+import rlp
 
 from eth_utils import (
     keccak,
@@ -28,6 +28,7 @@ from evm.exceptions import (
 from evm.rlp.headers import (
     BlockHeader,
 )
+from evm.rlp.receipts import Receipt  # noqa: F401
 from evm.utils.datatypes import (
     Configurable,
 )
@@ -62,7 +63,7 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         self.chaindb = chaindb
         block_class = self.get_block_class()
         self.block = block_class.from_header(header=header, chaindb=self.chaindb)
-        self.receipts = []
+        self.receipts = []  # type: List[Receipt]
 
     #
     # Logging
@@ -218,7 +219,10 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         header = self.block.header
         temp_block = self.generate_block_from_parent_header_and_coinbase(header, header.coinbase)
         prev_hashes = (header.hash, ) + self.previous_hashes
-        state = self.get_state(self.chaindb, temp_block, prev_hashes)
+        gas_used = 0
+        if self.receipts:
+            gas_used = self.receipts[-1].gas_used
+        state = self.get_state(self.chaindb, temp_block, prev_hashes, gas_used)
         assert state.gas_used == 0, "There must not be any gas used in a fresh temporary block"
 
         snapshot = state.snapshot()
@@ -444,7 +448,7 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         return cls._state_class
 
     @classmethod
-    def get_state(cls, chaindb, block, prev_hashes, receipts=None):
+    def get_state(cls, chaindb, block, prev_hashes, gas_used):
         """
         Return state object
 
@@ -454,23 +458,24 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         :return: state with root defined in block
         """
         execution_context = block.header.create_execution_context(prev_hashes)
-        if receipts is None:
-            receipts = block.get_receipts(chaindb)
 
         return cls.get_state_class()(
             chaindb,
             execution_context=execution_context,
             state_root=block.header.state_root,
-            receipts=receipts,
+            gas_used=gas_used,
         )
 
     @property
     def state(self):
         """Return current state property
         """
+        gas_used = 0
+        if self.receipts:
+            gas_used = self.receipts[-1].gas_used
         return self.get_state(
             chaindb=self.chaindb,
             block=self.block,
             prev_hashes=self.previous_hashes,
-            receipts=list(self.receipts),
+            gas_used=gas_used,
         )
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -62,6 +62,7 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         self.chaindb = chaindb
         block_class = self.get_block_class()
         self.block = block_class.from_header(header=header, chaindb=self.chaindb)
+        self.receipts = []
 
     #
     # Logging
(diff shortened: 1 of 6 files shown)


evm/vm/base.py
```diff
@@ -132,6 +133,7 @@ def execute_bytecode(self,
     # Mining
     #
     def import_block(self, block):
+        self.receipts = []
```
pipermerriam: This stood out to me.  What if instead of tracking this state we instead returned the receipt from `apply_transaction` and accumulated the receipts in the loop [here](https://github.com/ethereum/py-evm/pull/565/files#diff-7a66be18d19df3f35fdebb216454a46fR149)

It looks like this change goes well with the changes within the `VMState` class.  It looks like we could remove the need for that class to be passed the receipts since I *think* this changes things so that the only thing it needs to know is the `gas_used` value from the latest receipt.

evm/vm/base.py
```diff
@@ -266,6 +275,8 @@ def validate_block(self, block):
                     )
                 )
 
+        # FIXME: make_trie_root_and_nodes() is rather expensive, and we already run that once in
+        # import_block(), so should refactor some of this code to avoid re-generating it here.
```
pipermerriam: This seems like we could fix this with some caching on the `make_trie_root_and_nodes` function (or maybe a wrapper around that function which implements a cache).

evm/vm/base.py
gsalgado: Yeah, but should we then pass `gas_used` as an argument to `__init__()` and keep updating it in `add_transaction()`? Or did you have something else in mind?

pipermerriam: Honestly not sure.  Nothing specific in mind for exact implementation but starting with `gas_used` being passed into `__init__` seems like a good start.  It'd be nice if it didn't need to be mutated but i'm more ok with mutating that value than the receipt storage on the `VM` instance.

evm/vm/base.py
```diff
@@ -459,7 +459,7 @@ def get_state_class(cls):
         return cls._state_class
 
     @classmethod
-    def get_state(cls, chaindb, block, prev_hashes, receipts=None):
+    def get_state(cls, chaindb, block, prev_hashes, gas_used=0):
```
pipermerriam: It seems like `gas_used` shouldn't have a default value.

evm/vm/base.py
```diff
@@ -77,14 +80,12 @@ def apply_transaction(self, transaction):
         """
         Apply the transaction to the vm in the current block.
         """
-        computation, block, trie_data_dict = self.state.apply_transaction(
+        computation, block, receipt = self.state.apply_transaction(
```
hwwhww: Reminder: the reason why `apply_transaction` returned `trie_data_dict` is for the stateless client to generate witness, which was discussed here: https://github.com/ethereum/py-evm/pull/247#issuecomment-356180921

But since the sharding roadmap was (largely) changed, now the stateless client implementation won't happen in couple months, so I don't have a strong opinion about the change. Hope that we will find a solution to solve both stateless/non-stateless client by then. :)

evm/vm/base.py
gsalgado: So, we need `gas_used` in `BaseState` so I'm passing it via `__init__()` and updating it on `BaseState.add_transaction()`, but in `BaseVM` we also need that value in the `.state` @property, which is currently looking that up from `BaseVM.receipts` but we could instead have a `BaseVM.gas_used` and update that on `BaseVM.apply_transaction()`. Would you prefer that?

evm/vm/base.py
gsalgado: We could have the VM incrementally generate the trie data, I suppose? The problem was that it was re-generating it from scratch for every transaction applied

gsalgado: Except it wouldn't make the VM class stateless, so I guess that wouldn't really buy us anything

hwwhww:
The second thoughts of this PR:

**[Previously]** 
* The transaction and receipt roots are calculated in `State.add_transaction()`.
* Each time `VM.apply_transaction()` is called, the `vm.block.header` **is updated with the latest `transaction_root` and `receipt_root`**.

**[Now]**
* `State.add_transaction()` returns `receipt`.
* The transaction and receipt roots are calculated in `VM.import_block(self, block)`. 
* Each time `VM.apply_transaction()` is called, the `vm.block.header` **isn't** updated with the latest `transaction_root` and `receipt_root`.

---

I took a look at `vm.mine_block()` which hasn't been changed in this PR:

https://github.com/gsalgado/py-evm/blob/bceb324d7f33b4d41aa1cead97229414021906b7/evm/vm/base.py#L164-L176

I think now the `vm.mine_block` and `chain.mine_block()` don't work as what it used to be: **return a mined block**,  because the current block (`vm.block`) hasn't been updated with correct `transaction_root` and `receipt_root`.

The fastest way to solve it may be calculating the roots in `vm.mine_block()` instead of in `vm.import_block()`, and we lost the propoerty of "having `vm.block` always has the latest roots info".

evm/vm/base.py
pipermerriam: That sounds viable.  We can continue to iterate on this to I'd recommend just picking something that seems good enough.

evm/vm/base.py
pipermerriam: @hwwhww we'll continue to move towards stateless clients but for now going to prioritze full sync and then we'll figure out how to make it work with stateless clients also.

gsalgado:
#573 has an optimization to `make_trie_root_and_nodes()` that we could do instead of what's suggested here. I'd need to profile it more thoroughly to make sure it's good enough but I'd be happy to go with that if you guys think that's preferable

hwwhww:
Oooooh, correct my own words: for the witness, we don't need `trie_data_dict`, only need the state trie branches, the main difference between is do we expect `state.apply_transaction()` returns the block with latest header or not.

If #573 doesn't work well, may be we could redesign the APIs to be "for importing a block (only generating the root once per block)" and "for creating a block (generating the root per transaction?)". Maybe divide `state.apply_transaction` into multiple functions.

evm/vm/base.py
gsalgado: Yeah, I guess #573 might still be useful!

evm/vm/base.py
gsalgado: We actually need to keep the receipts somewhere in order to generate the receipt trie

evm/vm/base.py
```diff
@@ -156,6 +159,13 @@ def mine_block(self, *args, **kwargs):
         Mine the current block. Proxies to self.pack_block method.
         """
         block = self.block
+        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
+        receipt_root_hash, receipt_kv_nodes = make_trie_root_and_nodes(self.receipts)
+        trie_data = merge(tx_kv_nodes, receipt_kv_nodes)
```
hwwhww: Probably unimportant: for performance optimization, this line could be removed, and just add them separately:
```python
self.chaindb.persist_trie_data_dict(tx_kv_nodes)
self.chaindb.persist_trie_data_dict(receipt_kv_nodes)
```

Issue #507, opened 2018-03-26
Database architecture refactor: Adapters and Wrappers


pipermerriam:
Replaces https://github.com/ethereum/py-evm/issues/410

### What is wrong?

The current architecture uses a single global `ChainDB` for each chain.

This is proving problematic because:

- The `ChainDB` is not aware of different VM rules and thus cannot have it's underlying business logic modified.
- Some of the network syncing code needs to interact with the database in ways that aren't currently supported.  See #337 
- It has a lot of API surface area.

### How can it be fixed

First, we establish the following new terminology.

- *Connection*: The base database class used to connect to the underlying db.  e.g. `MemoryDB` or the `LevelDB`.  This object is treated as a key/value store.
- *Wrapper*: A class which *wraps* a *connection*, adding some extra functionality.  e.g. `JournalDB`, `HashTrie`
- *Adapter*: A class which implements a high level API over top of an underlying database.

We also establish a new convention/rule regarding the APIs exposed by *wrappers*.  Wrapper APIs may **only** be used at the same level that the wrapper is applied.

For Example:

Suppose the `VM` class applies the `JournalDB` wrapper to the database connection, and then passes the resulting object down to the `VMState`.  The `VMState` is not allowed to interact with any of the APIs exposed by the `JournalDB`.

Now, we establish the following new Adapter classes.

- `MetadataDB`: 
  - Only deals with chain metadata (e.g. canonical head, mapping block numbers to canonical block hashes, mapping transaction hashes to canonical block hash in which it was mined).
  - Used by the `Chain` class.
  - Lets look into using SQLAlchemy + sqlite3 for storage of this data.
- `ChainDB`:
  - Deals with storage of chain data (e.g. blocks, transactions, receipts)
  - Lets look into using SQLAlchemy + sqlite3 for storage of this data.
  - Each `VM` may specify their own `ChainDB` class.
- `AccountDB`:
  - Replaces `evm.db.state.AccountStateDB`
  - Deals with VM state (e.g. account balances, nonces, storage, code)
  - each `VMState` may specify their own `AccountDB` class.

So, responsibility for databases is as follows.

- `Chain` -> `MetadataDB` (replaces what we currently refer to as `ChainDB`
- `VM` -> `ChainDB` (not to be confused with the existing `ChainDB`
- `VMState` -> `AccountDB` (replaces the current `AccountStateDB`

[...]

PR #569, opened 2018-04-18
Refactor JournalDB for efficiency

cburgdorf:
### What was wrong?

As discussed in #561 the current implementation of the `JournalDB` imposes efficiency problems on the system as every `set` and `delete` operation causes an extra `get` operation on the underlying db.

### How was it fixed?

The new `JounalDB` works as follows.

`.record()` - starts recording and returns a `changeset_id` where all changes are recorded in. Nothing is written to the db. Any `set` and `delete` actions that are recorded into the active changeset. No extra `get` calls are made on the underlying database.

`.discard(changeset_id)` - throws away all changes that were recorded in the given changeset. Again, the underlying db isn't hit for any of this.

`commit(changeset_id)` - merges the given changeset into the previous changeset. If the previous changeset is the root changeset, the changes are written back to the underlying db.

`persist()` - collapses all changesets, writes them to the db and starts a new recording.

#### Cute Animal Picture

### Initial discussion moved below

@pipermerriam I gave this journaldb stuff another look this morning.

I have a better understanding now how the `JournalDB` is used. For instance, before a transaction is applied we take a snapshot and if the transaction errors we can revert to the snapshot and in case everything goes through we commit the changes.

With this in mind it sure makes sense to design this similar to how files are used e.g.

```python
with db.record():
  db.set(...)
  db.delete(...)

  # write changes at the end of the block or discard in case of error
```

In such a scenario tho, working with snapshot ids (or maybe record ids?!) would be completely abstracted away from us.

Looking at all call sites I can see that working out for places like `apply_message` in the computations but there seem to be other places that seem more tightly coupled to the concept of working with bare snapshot/record ids.

E.g. there are places where we combine the `state_root` with the snapshot to get a *full snapshot*

```python
(self.state_root, self._chaindb.snapshot())
```

So, am I right to assume that we would still keep the functionality of taking bare ids + commiting and have the contextmanager API sit on top as a higher level API to be used in places that can benefit from it?

I'm holding back to follow down further that path for now because I first want to make sure I got the basics right.

So, what I did now is that I took your work and try to approach things from a test-driven side. I added tests to prove when and how things are written back to the underlying database.

I added a couple more assertions in these tests to prove the inner workings e.g. that commiting a snapshot will merge and remove it etc. The tests reach into some supposed to be private APIs but I *think* I once read you rather have tests reaching into private APIs than exposing public APIs that are only used for tests. This is all WIP stuff anyway.

**Here's comes the most important part**

I actually *think* that we may need to change the whole snapshot terminology for the new model.

In the old model a snapshot was truly a snapshot of a state that we could revert to. That's actually not really the case with the new model. When we say `db.revert(snapshot)` we don't really revert everything *to the point* of `snapshot`. 

What we really do is that we throw away the *changes* that were recorded in `snapshot` (including subsequent points). So, essentially this just means that we are back at the point we were *before* we took the `snapshot` (aka recording) simply because we throw away a *change set*.

So, wouldn't it make more sense to call the API something like `db.forget(changeset)`?

I think this would go well together with renaming `snapshot()` to `record()`. Because, isn't that essentially what we do in the new model? We start recording changes that we either throw away (`forget`) or commit.

I followed down that path a bit and started renaming APIs just to see what that would look like. I also added an API `commit_all` which is basically the part that collapses all changesets to the point where they get written to the underlying database.

I also made it so that the `JournalDB` starts recording implicitly and maybe that would go well together with an API to retrieve the root changeset.

>I think the core concept here is that committing doesn't commit to the underlying database, it just commits to no longer reverting a given checkpoint (or any of the later checkpoints)

Conceptionally, I don't understand why commiting checkpoint `C1` out of `C1`, `C2`, `C3` means we are also commiting to not revert the later checkpoints `C2`, `C3`. Why would it be wrong if commiting `C1` would not make assumptions about what may happen to `C2` and `C3` just yet?

Alright that's it for now. Waiting for your feedback now. Hope it's not totally off.

### DIFF
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -244,7 +244,7 @@ class JournalDB(BaseDB):
 
             # Ensure the journal automatically restarts recording after
             # it has been persisted to the underlying db
-            self.record()
+            self.reset()
 
     def persist(self) -> None:
         """
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -1,5 +1,5 @@
 import collections
-from typing import Dict  # noqa: F401
+from typing import Dict, Union  # noqa: F401
 import uuid
 
 from cytoolz import (
@@ -10,6 +10,13 @@ from cytoolz import (
 
 from evm.db.backends.base import BaseDB
 from evm.exceptions import ValidationError
+
+
+class DeletedEntry:
+    pass
+
+
+DELETED_ENTRY = DeletedEntry()
 
 
 class Journal(BaseDB):
@@ -25,7 +32,7 @@ class Journal(BaseDB):
         # contains a mapping from all of the `uuid4` changeset_ids
         # to a dictionary of key:value pairs with the recorded changes
         # that belong to the changeset
-        self.journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, bytes]]  # noqa E501
+        self.journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, Union[bytes, DeletedEntry]]]  # noqa E501
 
     @property
     def root_changeset_id(self) -> uuid.UUID:
@@ -42,14 +49,14 @@ class Journal(BaseDB):
         return last(self.journal_data.keys())
 
     @property
-    def latest(self) -> Dict[bytes, bytes]:
+    def latest(self) -> Dict[bytes, Union[bytes, DeletedEntry]]:
         """
         Returns the dictionary of db keys and values for the latest changeset.
         """
         return self.journal_data[self.latest_id]
 
     @latest.setter
-    def latest(self, value: Dict[bytes, bytes]) -> None:
+    def latest(self, value: Dict[bytes, Union[bytes, DeletedEntry]]) -> None:
         """
         Setter for updating the *latest* changeset.
         """
@@ -119,31 +126,22 @@ class Journal(BaseDB):
         """
         # Ignored from mypy because of https://github.com/python/typeshed/issues/2078
         for changeset_data in reversed(self.journal_data.values()):  # type: ignore
-            try:
-                value = changeset_data[key]
-            except KeyError:
-                continue
+            if key in changeset_data:
+                return changeset_data[key]
             else:
-                if value is None:
-                    raise KeyError(key)
-                else:
-                    return value
-        else:
-            raise KeyError(key)
+                continue
+
+        return None
 
     def set(self, key: bytes, value: bytes) -> None:
         self.latest[key] = value
 
     def exists(self, key: bytes) -> bool:
-        try:
-            self.get(key)
-        except KeyError:
-            return False
-        else:
-            return True
+        val = self.get(key)
+        return val is not None and val is not DELETED_ENTRY
 
     def delete(self, key: bytes) -> None:
-        self.latest[key] = None
+        self.latest[key] = DELETED_ENTRY
 
 
 class JournalDB(BaseDB):
@@ -172,12 +170,13 @@ class JournalDB(BaseDB):
         self.wrapped_db = wrapped_db
         self.reset()
 
-    # TODO: Measure perf impact of exception ping pong
     def get(self, key: bytes) -> bytes:
-        try:
-            return self.journal[key]
-        except KeyError:
+
+        val = self.journal[key]
+        if val is None:
             return self.wrapped_db[key]
+        else:
+            return val
 
     def set(self, key: bytes, value: bytes) -> None:
         """
@@ -233,7 +232,7 @@ class JournalDB(BaseDB):
 
         if self.journal.is_empty():
             for key, value in journal_data.items():
-                if value is not None:
+                if value is not DELETED_ENTRY:
                     self.wrapped_db[key] = value
                 else:
                     try:
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -14,112 +14,112 @@ from evm.exceptions import ValidationError
 
 class Journal(BaseDB):
     """
-    A Journal is an ordered list of checkpoints.  A checkpoint is a dictionary
-    of database keys and values.  The values are the "original" value of that
-    key at the time the checkpoint was created.
+    A Journal is an ordered list of changesets.  A changeset is a dictionary
+    of database keys and values.  The values are tracked changes that were
+    written after the changeset was created
 
-    Checkpoints are referenced by a random uuid4.
+    Changesets are referenced by a random uuid4.
     """
 
     def __init__(self):
-        # contains a mapping from all of the `uuid4` in the `checkpoints` array
-        # to a dictionary of key:value pairs wher the `value` is the original
-        # value for the given key at the moment this checkpoint was created.
+        # contains a mapping from all of the `uuid4` changeset_ids
+        # to a dictionary of key:value pairs with the recorded changes
+        # that belong to the changeset
         self.journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, bytes]]  # noqa E501
 
     @property
-    def root_checkpoint_id(self):
+    def root_changeset_id(self):
         """
-        Returns the checkpoint_id of the latest checkpoint
+        Returns the id of the root changeset
         """
         return first(self.journal_data.keys())
 
     @property
     def latest_id(self):
         """
-        Returns the checkpoint_id of the latest checkpoint
+        Returns the id of the latest changeset
         """
         return last(self.journal_data.keys())
 
     @property
     def latest(self):
         """
-        Returns the dictionary of db keys and values for the latest checkpoint.
+        Returns the dictionary of db keys and values for the latest changeset.
         """
         return self.journal_data[self.latest_id]
 
     @latest.setter
     def latest(self, value):
         """
-        Setter for updating the *latest* checkpoint.
+        Setter for updating the *latest* changeset.
         """
         self.journal_data[self.latest_id] = value
 
     def is_empty(self):
         return len(self.journal_data) == 0
 
-    def has_checkpoint(self, checkpoint_id):
-        return checkpoint_id in self.journal_data
+    def has_changeset(self, changeset_id):
+        return changeset_id in self.journal_data
 
-    def create_checkpoint(self):
+    def record_changeset(self):
         """
-        Creates a new checkpoint. Checkpoints are referenced by a random uuid4
-        to prevent collisions between multiple checkpoints.
+        Creates a new changeset. Changesets are referenced by a random uuid4
+        to prevent collisions between multiple changesets.
         """
-        checkpoint_id = uuid.uuid4()
-        self.journal_data[checkpoint_id] = {}
-        return checkpoint_id
+        changeset_id = uuid.uuid4()
+        self.journal_data[changeset_id] = {}
+        return changeset_id
 
-    def pop_checkpoint(self, checkpoint_id):
+    def pop_changeset(self, changeset_id):
         """
-        Returns all changes from the given checkpoint.  This includes all of
-        the changes from any subsequent checkpoints, giving precidence to
-        later checkpoints.
+        Returns all changes from the given changeset.  This includes all of
+        the changes from any subsequent changeset, giving precidence to
+        later changesets.
         """
-        if checkpoint_id not in self.journal_data:
-            raise KeyError("Unknown checkpoint: {0}".format(checkpoint_id))
+        if changeset_id not in self.journal_data:
+            raise KeyError("Unknown changeset: {0}".format(changeset_id))
 
         all_ids = tuple(self.journal_data.keys())
-        checkpoint_idx = all_ids.index(checkpoint_id)
-        checkpoints_to_pop = all_ids[checkpoint_idx:]
+        changeset_idx = all_ids.index(changeset_id)
+        changesets_to_pop = all_ids[changeset_idx:]
 
-        # we pull all of the checkpoints *after* the checkpoint we are
+        # we pull all of the changesets *after* the changeset we are
         # reverting to and collapse them to a single set of keys (giving
-        # precedence to later checkpoints)
-        checkpoint_data = merge(*(
+        # precedence to later changesets)
+        changeset_data = merge(*(
             self.journal_data.pop(c_id)
             for c_id
-            in checkpoints_to_pop
+            in changesets_to_pop
         ))
 
-        return checkpoint_data
+        return changeset_data
 
-    def commit_checkpoint(self, checkpoint_id):
+    def commit_changeset(self, changeset_id):
         """
-        Collapses all changes for the given checkpoint into the previous
-        checkpoint if it exists.
+        Collapses all changes for the given changeset into the previous
+        changesets if it exists.
         """
-        checkpoint_data = self.pop_checkpoint(checkpoint_id)
+        changeset_data = self.pop_changeset(changeset_id)
         if not self.is_empty():
-            # we only have to merge the changes into the latest checkpoint if
+            # we only have to merge the changes into the latest changeset if
             # there is one.
             self.latest = merge(
                 self.latest,
-                checkpoint_data,
+                changeset_data,
             )
-        return checkpoint_data
+        return changeset_data
 
     #
     # Database API
     #
     def get(self, key):
         """
-        For key lookups we need to iterate through the checkpoints in reverse
+        For key lookups we need to iterate through the changesets in reverse
         order, returning from the first one in which the key is present.
         """
-        for checkpoint_data in reversed(self.journal_data.values()):
+        for changeset_data in reversed(self.journal_data.values()):
             try:
-                value = checkpoint_data[key]
+                value = changeset_data[key]
             except KeyError:
                 continue
             else:
@@ -134,9 +134,9 @@ class Journal(BaseDB):
         self.latest[key] = value
 
     def exists(self, key):
-        for checkpoint_data in reversed(self.journal_data.values()):
+        for changeset_data in reversed(self.journal_data.values()):
             try:
-                value = checkpoint_data[key]
+                value = changeset_data[key]
             except KeyError:
                 continue
             else:
@@ -168,14 +168,14 @@ class JournalDB(BaseDB):
     The added memory footprint for a JournalDB is one key/value stored per
     database key which is changed.  Subsequent changes to the same key within
     the same changeset will not increase the journal size since we only need
-    to track latest value for any given key within any given checkpoint.
+    to track latest value for any given key within any given changeset.
     """
     wrapped_db = None
     journal = None  # type: Journal
 
     def __init__(self, wrapped_db: BaseDB) -> None:
         self.wrapped_db = wrapped_db
-        self.clear()
+        self.reset()
 
     # TODO: Measure perf impact of exception ping pong
     def get(self, key: bytes) -> bytes:
@@ -202,37 +202,39 @@ class JournalDB(BaseDB):
     #
     # Snapshot API
     #
-    def _validate_checkpoint(self, checkpoint):
+    def _validate_changeset(self, changeset_id):
         """
-        Checks to be sure the checkpoint is known by the journal
+        Checks to be sure the changeset is known by the journal
         """
-        if not self.journal.has_checkpoint(checkpoint):
-            raise ValidationError("Checkpoint not found in journal: {0}".format(
-                str(checkpoint)
+        if not self.journal.has_changeset(changeset_id):
+            raise ValidationError("Changeset not found in journal: {0}".format(
+                str(changeset_id)
             ))
 
     def record(self):
         """
         Starts a new recording and returns an id for the associated changeset
         """
-        return self.journal.create_checkpoint()
+        return self.journal.record_changeset()
 
-    def discard(self, changeset):
+    def discard(self, changeset_id):
         """
         Throws away all journaled data starting at the given changeset
         """
-        self._validate_checkpoint(changeset)
-        self.journal.pop_checkpoint(changeset)
+        self._validate_changeset(changeset_id)
+        self.journal.pop_changeset(changeset_id)
 
-    def commit(self, checkpoint):
+    def commit(self, changeset_id):
         """
-        Commits a given checkpoint. This involves committing the journal up to
-        the given checkpoint which returns all of the journal changes.  *if*
-        this is the base checkpoint then all journal data should be committed
-        to the underlying database.
+        Commits a given changeset. This merges the given changeset and all
+        subsequent changesets into the previous changeset giving precidence
+        to later changesets in case of any conflicting keys.
+
+        If this is the base changeset then all changes will be written to
+        the underlying database and the Journal starts a new recording.
         """
-        self._validate_checkpoint(checkpoint)
-        journal_data = self.journal.commit_checkpoint(checkpoint)
+        self._validate_changeset(changeset_id)
+        journal_data = self.journal.commit_changeset(changeset_id)
 
         if self.journal.is_empty():
             for key, value in journal_data.items():
@@ -244,26 +246,19 @@ class JournalDB(BaseDB):
                     except KeyError:
                         pass
 
-            # ensure new root checkpoint
-            self.journal.create_checkpoint()
+            # Ensure the journal automatically restarts recording after
+            # it has been persisted to the underlying db
+            self.record()
 
     def persist(self):
         """
         Persist all changes in underlying db
         """
-        self.commit(self.journal.root_checkpoint_id)
+        self.commit(self.journal.root_changeset_id)
 
-    # TODO: rename to reset
-    def clear(self):
+    def reset(self):
         """
-        Cleare the entire journal.
+        Reset the entire journal.
         """
         self.journal = Journal()
         self.record()
-
-    # temporary aliases to assist refactoring
-    def snapshot(self):
-        return self.record()
-
-    def revert(self, changeset):
-        return self.discard(changeset)
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -247,9 +247,9 @@ class JournalDB(BaseDB):
             # ensure new root checkpoint
             self.journal.create_checkpoint()
 
-    def commit_all(self):
+    def persist(self):
         """
-        Collapses all outstanding changes
+        Persist all changes in underlying db
         """
         self.commit(self.journal.root_checkpoint_id)
 
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -253,3 +253,10 @@ class JournalDB(BaseDB):
         Cleare the entire journal.
         """
         self.journal = Journal()
+
+    # temporary aliases to assist refactoring
+    def snapshot(self):
+        return self.record()
+
+    def revert(self, changeset):
+        return self.discard(changeset)
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -163,6 +163,8 @@ class JournalDB(BaseDB):
     subsequent changesets into the previous changeset giving precidence to later
     changesets in case of conflicting keys.
 
+    Nothing is written to the underlying db until `persist()` is called.
+
     The added memory footprint for a JournalDB is one key/value stored per
     database key which is changed.  Subsequent changes to the same key within
     the same changeset will not increase the journal size since we only need
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -211,7 +211,7 @@ class JournalDB(BaseDB):
         """
         return self.journal.create_checkpoint()
 
-    def forget(self, changeset):
+    def discard(self, changeset):
         """
         Throws away all journaled data starting at the given changeset
         """
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -173,7 +173,9 @@ class JournalDB(BaseDB):
     def get(self, key: bytes) -> bytes:
 
         val = self.journal[key]
-        if val is None:
+        if val is DELETED_ENTRY:
+            raise KeyError(key)
+        elif val is None:
             return self.wrapped_db[key]
         else:
             return val
--- a/evm/db/journal.py
+++ b/evm/db/journal.py
@@ -3,6 +3,7 @@ from typing import Dict  # noqa: F401
 import uuid
 
 from cytoolz import (
+    first,
     merge,
     last,
 )
@@ -25,6 +26,13 @@ class Journal(BaseDB):
         # to a dictionary of key:value pairs wher the `value` is the original
         # value for the given key at the moment this checkpoint was created.
         self.journal_data = collections.OrderedDict()
+
+    @property
+    def root_checkpoint_id(self):
+        """
+        Returns the checkpoint_id of the latest checkpoint
+        """
+        return first(self.journal_data.keys())
 
     @property
     def latest_id(self):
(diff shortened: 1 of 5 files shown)


evm/db/journal.py
```diff
@@ -130,42 +164,34 @@ class JournalDB(BaseDB):
...
         self.wrapped_db = wrapped_db
         self.journal = Journal()
+        # Not sure if that's the right thing to do but for now assume
+        # we just start recording right away.
+        self.record()
 
-    def get(self, key):
-        return self.wrapped_db.get(key)
+    #TODO: Discuss potential inefficiency issue
+    def get(self, key: bytes) -> bytes:
```
cburgdorf: @pipermerriam unrelated to the bigger picture but as a little side discussion. I *think* not finding a key in the journal and further having to look it up in the underlying database won't be an exceptional case but rather be something quite normal. Lookups in the journal a quite costly in itself because we have to loop through the checkpoints so I wonder if throwing a `KeyError` and then catching it to lookup the underlying db maybe inefficient?

tests/database/test_journal_db.py
```diff
@@ -47,30 +50,113 @@ def test_revert_clears_reverted_journal_entries(journal_db):
...
+
+    journal_db.set(b'1', b'test-b')
+    changeset_b = journal_db.record()
+    assert len(journal_db.journal.journal_data) == 3
+
+    journal_db.set(b'1', b'test-c')
+    changeset_c = journal_db.record()
+    assert len(journal_db.journal.journal_data) == 4
+
+    # TODO: Clarify
```
cburgdorf: I added this test to prove I understood how the system is supposed to handle the case that we commit `C1` where `C2` and `C3` also exist. In fact, committing `C1` merges in `C2` and `C3` but I'm not sure I understand the **why** of that. What would be wrong if committing `C1` would not make assumptions about what happens to `C2` or `C3` just yet?

evm/db/journal.py
pipermerriam: Not sure, try it out?

pipermerriam:
TIL about [`collections.ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap) which you almost definitely can leverage in the journal for lookups across all of the underlying journal data.

evm/db/journal.py
pipermerriam: > Lookups in the journal a quite costly in itself because we have to loop through the checkpoints

I think this cost is orders of magnitude lower than querying the underlying leveldb.

evm/db/journal.py
```diff
@@ -174,53 +200,56 @@ def _validate_checkpoint(self, checkpoint):
...
+                        del self.wrapped_db[key]
+                    except KeyError:
+                        pass
+
+
+    def commit_all(self):
+        """
+        Collapses all outstanding changes
+        """
+        while not self.journal.is_empty():
```
pipermerriam: I think that exposing a `journal.root_checkpoint_id` would be appropriate here so that you could do a single call to `self.commit(self.journal.root_checkpoint_id)`

evm/db/journal.py
cburgdorf: >I think this cost is orders of magnitude lower than querying the underlying leveldb.

Definitely faster than that. But I *think* the lookups could be even faster without the exception ping pong (because stack traces etc) But again, that's really just a side discussion and I can measure it once I got the basics right.

pipermerriam: Ah, I mis-understood.  Yes, this performance gain is desirable.

evm/db/journal.py
cburgdorf: Good point. I wrote that API before I understood commits always merge in subsequent commits. :thumbsup:

cburgdorf:
@pipermerriam this has reached the "make it run" phase of the implementation in the sense that it at least passes all tests. :tada:  Commit history is still a mess and there's still other work left to get done.

I'm reaching out to get your opinion about a couple of things.

**1. Regarding the discussed context manager API**

I'm not yet 100 % sure we want it but since I'm still very new to this, there's a high chance that I'm wrong, hence I'm reaching out to discuss it.

Ok, so why would we want a context manager API in the first place? Because we assumed otherwise it would be easy to use the API wrong (e.g. forget to start recording changes, forget to write them to the underlying db etc).

This doesn't seem to be the case though. The way this works with the current implementation is  that the `JournalDB` is always recording changes by default. If we call `persist()` the changes will be written back to the underlying db and a new recording starts automatically.

But there's another reason. I noticed that there exists an integration test that uses a `ChainDB` that does not use `JournalDB` at all. In fact this thing is called [`NonJournaledAsyncChainDB`](https://github.com/ethereum/py-evm/blob/7122c02ef6dde72c9921a3c6d06d2b0cc8635196/evm/db/chain.py#L680-L697) and basically for the test that uses this every call inside `ChainDB` that boils down to `self.db.<something>` will hit the `MemoryDB`. I'm not sure if that is a thing that is only relevant for the tests and could easily be changed on the testing side or if it is actually a hard requirement that things are supposed to work in such a scenario.

However, what that basically means is that if the `JournalDB` gets a context manager API and uses it within `ChainDB` then `MemoryDB` would need that same API even though it makes probably less sense for `MemoryDB` to have this kind of API.

In order to work around that, what I did for now, is that I expose the raw passed `base_db` as a property on the `ChainDB` next to the usual `db` property (which is supposed to be wrapped as `JournalDB` but sometimes ends up not being one).

This allows that certain actions within the `ChainDB` can simply call `self.base_db.set(...)` and `self.base_db.delete(...)` to bypass the journaling/caching altogether. This is based on the assumption that certain actions simply don't need the journaling anyway.

If that would be a viable way forward then I'd propose we go further and rename the properties to `non_journaled_db` and `journaled_db` or something similar to make it more clear that there are two different ways to access the database.

Btw, I just moved enough calls from `self.db.<method>` to `self.base_db.<method>` to get the test suite passing but I guess there may be even more that could be moved to always hit the non-journaled db.

And my third and last aspect of this would be that there really just is a single call to `persist()` in [`apply_transaction`](https://github.com/cburgdorf/py-evm/blob/d346bd2b059d91c11b436cebbae1a556db70eec1/evm/vm/base.py#L89) anway. This is again, based on the assumption that the transaction processing is the only part of the system that actually needs the functionality of undoing changes.

**Regarding usage of [`ChainMap`](https://docs.python.org/3/library/collections.html#collections.ChainMap)

I haven't looked much into that yet because I first want to get the basics right but from the docs it says

>A ChainMap groups multiple dicts or other mappings together to create a single, updateable view. If no maps are specified, a single empty dictionary is provided so that a new chain always has at least one mapping.

>The underlying mappings are stored in a list. That list is public and can be accessed or updated using the maps attribute. There is no other state.

If the underlying mechanics work as a `List[Dict[TKey, TVal]` isn't that a deal breaker because it would mean we could only reference to the changesets (the inner dict) by there index in the list whereas we really need the functionality of referencing them through a `UUID` hence we currently use a `Dict[UUID, Dict[TKey, TVal]]`?

**Regarding terminology**

I changed the terminology to reflect the new model of the `JournalDB` as I mentioned before. I personally think the old terminology just doesn't apply anymore because we previously had it as follows

`.snapshot()` - snapshot the **original values** and return a checkpoint id that allows **going back** to the original values (because changes were immediately applied to the underlying db).

`revert(snapshot)` - which really meant *revert to `<snapshot-id>`* as in: get all the original values from the `snapshot` and reset the underlying database to it.

That's just not how the new model works so I changed the terminology to 

`.record()` - starts recording and returns a `changeset_id` where all **changes** are recorded in. Nothing is written to the db.

`.discard(changeset_id)` - This simply throws away all **changes** that were recorded in the given `changeset`. Again, the underlyng db wasn't hit at all.

Alright, that's it from me so far. Sorry for the long blogging :grimacing:

pipermerriam:
> So, am I right to assume that we would still keep the functionality of taking bare ids + commiting and have the contextmanager API sit on top as a higher level API to be used in places that can benefit from it?

Yes, I think there is need for both of these APIs

> I actually think that we may need to change the whole snapshot terminology for the new model.

New names seem appropriate.

> Ok, so why would we want a context manager API in the first place? Because we assumed otherwise it would be easy to use the API wrong (e.g. forget to start recording changes, forget to write them to the underlying db etc).

In the previous `JournalDB` model, forgetting to commit changes is an *ok* action since changes would still be persisted to the underlying database.  In this new model, if you forget to commit, changes would be reverted implicitly when the journal is garbage collected.

> This is again, based on the assumption that the transaction processing is the only part of the system that actually needs the functionality of undoing changes.

I think this is an ok assumption.

pipermerriam:
BTW, if a context manager API starts cascading into widespread API changes then we should punt on it and see about doing it later down the road.

cburgdorf:
:+1: Thank you for the feedback. I'll do some more changes and clean this up further but I believe we could start profiling the performance impact of this. 

@gsalgado How would I best measure the impact of this? I guess I could use the tests for the block importing and swap out the `MemoryDB` against the `LevelDB` to get something close to the real world impact? Any better way to do that?

cburgdorf:
Ah, btw, regarding

>In the previous JournalDB model, forgetting to commit changes is an ok action since changes would still be persisted to the underlying database. In this new model, if you forget to commit, changes would be reverted implicitly when the journal is garbage collected.

Currently, in this model the `Journal` instances aren't thrown away and garbage collected. I need to take a deeper look into this but my current reasoning is that the `clear()` / `reset()` is just not needed anymore and not recreating the `Journal` all the time should also save a bit of GC pressure. Again, this part is still super vague to me. There may be good reasons why this assumption is false.

gsalgado:
>  @gsalgado How would I best measure the impact of this? I guess I could use the tests for the block importing and swap out the MemoryDB against the LevelDB to get something close to the real world impact? Any better way to do that?

The current JournalDB becomes a real problem as the DB gets big, so maybe if you'd fill a DB with random data until it's >10GB in size and then write a script to make lots of arbitrary reads/writes you could get that. We'd still want to see how it performs when importing recent blocks on ropsten, though. I can do that next week.

evm/db/journal.py
```diff
@@ -1,226 +1,264 @@
...
+                else:
+                    return value
+        else:
+            raise KeyError(key)
+
+    def set(self, key, value):
+        self.latest[key] = value
+
+    def exists(self, key):
+        for changeset_data in reversed(self.journal_data.values()):
```
pipermerriam: Looking at this we can de-dup some of this code by changing this implementation to be:

```python
def exists(self, key):
    try:
        self.get(key)
    except KeyError:
        return False
    else:
        return True
```

pipermerriam:
@carver I added you as a reviewer.  I think you two should start talking about direction since:

1. The work that you (@carver) are taking on is related to this
2. The changes in this PR are likely just a first step towards the larger changes incoming with #507 (so we may be just looking for a *good enough* solution in the interum)

evm/vm/base.py
```diff
@@ -86,7 +86,7 @@ def apply_transaction(self, transaction):
         self.block = block
         self.receipts.append(receipt)
 
-        self.clear_journal()
+        self.chaindb.persist()
```
carver: If this `persist` is run anywhere between a pair of `id = record()` and `commit(id)` calls, it will break the commit call. What if some caller higher up the stack decides it wants to revert? It just lost the option. `persist()` has the semantics of "commit all in-progress transactions/recordings", which seems problematic anywhere we use it. I don't think we should even have the option to do that.

The more I think about it, the more I don't see a reason to allow committing multiple levels of transactions in one call:
```
record
  record
    record
commit <- I don't think this should be possible. Child transactions are in an unknown state
    commit
  commit
commit
```

Maybe we want to keep `persist` explicit, so we know when we are really writing to the wrapped database. That could make sense. Since the JournalDB implicitly starts with a record, that special outermost transaction can have a special commit: `persist`. So:

```
journal init
  record
    record
    commit
  commit
persist
```

That implies we should make the public `commit()` call fail if there's only one changeset.

tests/database/test_journal_db.py
carver: This is quite surprising to me. If I start a transaction `Turkey` and inside it start transaction `Duck`, but never commit transaction `Duck`, then I expect that committing transaction `Turkey` should either:

1) Fail _<- my preference_
2) Ignore the uncommitted Duck transactions

evm/vm/base.py
cburgdorf: >`persist()` has the semantics of "commit all in-progress transactions/recordings"

That's true but my understanding was that transaction processing is sequential and self-contained. I didn't see the case of "child transactions" to be honest.

If the processing goes like this, then there shouldn't be a problem I think

```
transaction 1
  commit()
    commit()
    revert()
  commit()
  persist()
transaction 2
  commit()
    commit()
  persist()
transaction 3
 ...
  persist()
```

However, your comment mentioning "child transactions" indicates that my assumption of this sequential processing maybe wrong and that `apply_transaction` may be run stacked (which I actually didn't come across while debugging but I also didn't explicitly pay attention)

evm/db/journal.py
```diff
@@ -1,226 +1,264 @@
...
+        # to a dictionary of key:value pairs with the recorded changes
+        # that belong to the changeset
+        self.journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, bytes]]  # noqa E501
+
+    @property
+    def root_changeset_id(self):
+        """
+        Returns the id of the root changeset
+        """
+        return first(self.journal_data.keys())
```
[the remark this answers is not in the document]
cburgdorf: Ah you mean when creating a `Journal` directly. I didn't test for that tbh as I considered `Journal` to be something somewhat internal to the `JournalDB` and the `JournalDB` starts recording automatically. Can certainly smooth the edges here though.

evm/db/journal.py
```diff
@@ -1,226 +1,264 @@
...
+        self.reset()
 
-    def get(self, key):
-        return self.wrapped_db.get(key)
+    # TODO: Measure perf impact of exception ping pong
+    def get(self, key: bytes) -> bytes:
+        try:
+            return self.journal[key]
+        except KeyError:
+            return self.wrapped_db[key]
```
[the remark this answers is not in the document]
cburgdorf: I think if we would want to optimize this for performance than we would make the `get(key`) access to return `None` (or a special `Empty` to support `None` as legal value) so that we don't have to loop through all the changesets twice (once for `exists()`, again for `get` if it exists). And then if the returned value is `None` / `Empty` we would look in the underlying db. Downside is that this handling of non-existing key access would be a bit non-standard.

evm/vm/base.py
carver: I think it could be totally reasonable to say that we don't want to support nested transactions. If so, we could simplify several things:

- `JournalDB` writes through all changes until `record()` is called
- once `record()` is called, it cannot be called again until `commit()`
- `commit()` writes through all pending changes
- remove `persist()` as redundant

As I'm typing it out, I realize that's starting to sound a lot like BatchDB. So I'm realizing that I still don't totally get what is the primary goal here and/or what we want different from BatchDB.

evm/db/journal.py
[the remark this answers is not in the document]
carver: Sounds right. Our delete mechanism already assumes that `None` is not an allowed value, it seems.

evm/vm/base.py
cburgdorf: Judging by your last comment on the main thread you may have already realized what I'm about to write but just for completeness let me elaborate further on this.

The refactored `JournalDB` works very much like the `BatchDB` in the sense that it does not write to the underlying db until some condition is met. This condition could be leaving a `with` statement or calling an explicit API. In the current design, you have to explicitly call `persist()` in order to write to the underlying database.

Before you call `persist` all changes (`set` / `delete`) go into the active changeset. Notice that there is *always* an active changeset where changes end up in.

However, you can build a stack of changesets if you explicitly call `record`. E.g.

```
changeset_1 = db.record()
db.set(a, 1)
changeset_2 = db.record()
db.set(a, 2)
changeset_3 = db.record()
db.delete(b)
db.set(c, 1)
```
At this point you'll end up with a stack of changesets that looks like this:

```
root_changeset
changeset_1
  sets a to 1
changeset_2
  sets a to 2
changset_3
  deletes b
  sets c to 1
```

Calling `commit()` or `discard(changeset_id)` will simply modify how this stack looks like but again, won't do any changes to the underlying db.

For instance, we could call `discard(changeset_3)` and would end up with this:

```
root_changeset
changeset_1
  sets a to 1
changeset_2
  sets a to 2
```

Or we could call `commit(changeset_3)` and would end up with this

```
root_changeset
changeset_1
  sets a to 1
changeset_2
  sets a to 2
  deletes b
  sets c to 1
```

Notice that `commit()` and `discard(changeset_id)` are called as part of the `Computation` handling. E.g. if a computation fails for some reason we discard the changes.

The `persist()` part on the other hand is called on the very end of the transaction processing (when all the computations are done) that's why there is only a single call to `persist()` and I *think* this should be sound.

cburgdorf:
>and I see now that these aren't really intended to be transactions but batching some subset of changes

Exactly. I elaborated further on this [here](https://github.com/ethereum/py-evm/pull/569#discussion_r183639713)

So, let me elaborate a bit further on the part that I find to be the most unsound about this change. Especially as it is very subtle and may have been overlooked. 

https://github.com/ethereum/py-evm/pull/569/files#diff-d71f9bfe38d29edaf17d6231470873ccR118

Notice that this PR changes `self.db` on the `ChainDB` to be the non-wrapped underlying db and moves the wrapped `JournalDB` to a new property `self.journal_db`. 

So why is that? It's because the new `JournalDB` does not write anything back to the underlying db by default (until one calls `persist()`). However, there are certain places in `ChainDB` that do some changes to the db and also do not need journaling or batching. In order for these places to work one would have to do something like this:

```python
db.set()
db.persist()
```

But here's the problem. We have tests where the chaindb uses a `NonJournaledAsyncChainDB` that entirely bypasses the `JournalDB` so that there is no `persist()` available. Where the `NonJournaledAsyncChainDB` is used, there also is no transaction processing so it seemed *safe* to simply change all these places to bypass the `JournalDB` by default and only use the `JournalDB`  for the places that really need the journaling functionality.

While that works to get the test suite to pass I'm not sure if it is really really sound or this actually introduces undesired behavior that our test suite simply does not catch yet.

In addition to that, I feel like this change lowered the cohesion of the `ChainDB` and I was wondering about possible alternatives. However, in the light of #507 and as you already said, this may just be good enough for now?

gsalgado:
> But here's the problem. We have tests where the chaindb uses a NonJournaledAsyncChainDB that entirely bypasses the JournalDB so that there is no persist() available

The main reason I created this non-journaled `ChainDB` was because the overhead caused by the journal made the fast sync too slow. If that overhead is gone, we can probably get rid of this class and use the journaled chaindb for fast sync as well

cburgdorf:
>The main reason I created this non-journaled `ChainDB` was because the overhead caused by the journal made the fast sync too slow

Ah, I see! I was missing that perspective. Thanks for chiming in to explain. That makes sense.

>If that overhead is gone, we can probably get rid of this class and use the journaled chaindb for fast sync as well

And I guess at this point I'd start seeing value in the contextmanager API so that all these trivial `.set(..)` cases can be written as.

```python
with self.db.record():
  self.db.set(...)
```

instead of 

```
self.db.set(...)
self.db.persist()
```

On the other hand, we may also consider something simpler such as

```
self.db.set(foo, bar, True)
```

where the last parameter basically indicates an implicit `persist()` call for these cases that do not need journaling / batching.

I think we should take a look at the performance impact of this PR as it is then decide about further actions.

gsalgado:
Just did a quick test running fast-sync (on a >30GB db) with these changes and the cProfile stats are pretty much the same as when using a non-journaled `ChainDB`

gsalgado:
> Notice that this PR changes self.db on the ChainDB to be the non-wrapped underlying db and moves the wrapped JournalDB to a new property self.journal_db

One thing that surprised me is that now we have some `ChainDB` methods using `self.db` and others using `self.journal_db`, but it's not clear to me when to use one or the other.

cburgdorf:
The separation was really just to get the tests for the `NonJournaledAsyncChainDB` passing. We can move everything back to the `JournalDB` if that turns out fast enough. However, since the `set` / `delete` API does not write to the underlying db until `.persist()` is called we need to decide how we want to handle places like this in `ChainDB`

```python
def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
        """
        :returns: iterable of headers newly on the canonical chain
        """
        if header.parent_hash != GENESIS_PARENT_HASH and not self.header_exists(header.parent_hash):
            raise ParentNotFound(
                "Cannot persist block header ({}) with unknown parent ({})".format(
                    encode_hex(header.hash), encode_hex(header.parent_hash)))

        self.db.set(
            header.hash,
            rlp.encode(header),
        )

        if header.parent_hash == GENESIS_PARENT_HASH:
            score = header.difficulty
        else:
            score = self.get_score(header.parent_hash) + header.difficulty
        self.db.set(
            make_block_hash_to_score_lookup_key(header.hash),
            rlp.encode(score, sedes=rlp.sedes.big_endian_int))

        try:
        except CanonicalHeadNotFound:
            new_headers = self._set_as_canonical_chain_head(header)
        else:
            if score > head_score:
                new_headers = self._set_as_canonical_chain_head(header)
            else:
                new_headers = tuple()

        return new_headers
```

The calls to `self.db.set(...)` won't write to the underlying db directly with the new `JournalDB`.

Here are some options:

1. Use a context manager API

```python
with db.record():
  db.set(...)
```

That will implicitly call `persist()` when the context is left.

2. Change `set` and `delete` to the following

```
def set(key: bytes, value:bytes, force_persist: boolean):
   ....
def delete(key: bytes, force_persist: boolean):
   ....
```

Then these places would call `self.db.set(x,y, True)`

3. Continue to use the raw database for the places where the `JournalDB` isn't needed

Personally, I think I'd just go with the second option for now.

cburgdorf:
>Just did a quick test running fast-sync (on a >30GB db) with these changes and the cProfile stats are pretty much the same as when using a non-journaled ChainDB

But the `fast-sync` doesn't process the transactions right? So in this case, it should be the same performance as with the non-journaled ChainDB basically just because as of the current state of the PR the (refactored) `JournalDB` is only used for transaction processing that actually needs the journaling functionality.

cburgdorf:
Yep, that would be another possible way forward. The downside of that is that it is less explicit. Consider you are hunting a bug that in the end turns out was caused because for some reason there was an active `changeset` being recorded so that all further `set`/`delete` calls were waiting for an `persist()` but that active changeset shouldn't have been there in the first place (because somewhere else before we forgot to call `persist()`)

This again sounds as if we could solve that by using a context manager API but in reality I think there are several reasons why a context manager API doesn't really solve that.

evm/db/journal.py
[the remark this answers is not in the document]
cburgdorf: @pipermerriam @carver let me admit that this benchmark is very biased as we are going to look up a key 1 million times in the `JournalDB` that only exists in the underlying db and each time we are going through 10 changesets.

But it shows that not doing the exception ping pong is 5.5 times faster

```python
    memory_db.set(b'x', b'x')
    for _ in range(10):
        journal_db.record()

    then = time.time()
    for _ in range(1000000):
        journal_db.get(b'x')
    now = time.time()

    duration = now - then
```

With the current version this takes about 8.5 seconds to run vs 1.5 seconds in the optimized not-exception based version that I haven't pushed yet.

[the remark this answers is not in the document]
cburgdorf: I have to back off I'm afraid. My initial test implementation was flawed and the proper implementation that passes all tests is only 1.5 x faster. Not sure if it's worth it. I'll push it for discussion.

cburgdorf:
I added missing type hints and also implemented the [performance improvement](https://github.com/ethereum/py-evm/pull/569/commits/7a6d2e4b0a6a5349f09aaf693f1e94dd6bafe7ec) where the looks rely less on exception throwing / catching.

Here's a quick, heavily biased test that poves a 1.5x performance improvement but I'm honestly not sure if it is worth the API change.

```python
 memory_db.set(b'x', b'x')
    for _ in range(10):
        journal_db.record()

    then = time.time()
    for _ in range(1000000):
        journal_db.get(b'x')
    now = time.time()

    duration = now - then
```

evm/db/journal.py
```diff
@@ -1,226 +1,258 @@
...
-                changes_to_merge,
                 self.latest,
+                changeset_data,
             )
+        return changeset_data
+
+    #
+    # Database API
+    #
+    def get(self, key: bytes) -> bytes:
```
pipermerriam: This should be:

```python
def get(self, key: bytes) -> Union[bytes, None]
    ...
```

right?

evm/db/journal.py
```diff
@@ -1,226 +1,258 @@
...
         self.wrapped_db = wrapped_db
-        self.journal = Journal()
+        self.reset()
 
-    def get(self, key):
-        return self.wrapped_db.get(key)
+    def get(self, key: bytes) -> bytes:
 
-    def set(self, key, value):
+        val = self.journal[key]
```
pipermerriam: It *appears* that `val` *could* be `DELETED_ENTRY` which would result in returning `DELETED_ENTRY` below.  Am I missing something?

evm/db/journal.py
```diff
@@ -1,226 +1,258 @@
...
+                    self.wrapped_db[key] = value
+                else:
+                    try:
+                        del self.wrapped_db[key]
+                    except KeyError:
+                        pass
+
+            # Ensure the journal automatically restarts recording after
+            # it has been persisted to the underlying db
+            self.record()
```
pipermerriam: We can use `self.record()` here because we *implicitely* know that the journal is empty.  However, do you think it'd be better to use `reset()`?  That'd remove any possibility of previous journal state hanging around.

Suppose a future refactor the the `Journal` adds some additional container where state is stored... Then this might not do what we want.

evm/vm/base.py
pipermerriam: I *think* I'm in support of having this separate mechanism.  We can keep it for now and sort it out in #507 if it proves problematic.

tests/database/test_journal_db.py
```diff
@@ -4,22 +4,38 @@
...
+    return JournalDB(memory_db)
+
+
+def test_delete_removes_data_from_underlying_db_after_persist(journal_db, memory_db):
+    memory_db.set(b'1', b'test-a')
+
+    assert memory_db.exists(b'1') is True
+
+    journal_db.delete(b'1')
+    journal_db.persist()
```
pipermerriam: Maybe add a sanity check that the value is still in the memory db after the `delete` and before the `persist` call.

tests/database/test_journal_db.py
```diff
@@ -47,30 +63,137 @@ def test_revert_clears_reverted_journal_entries(journal_db):
...
+    assert len(journal_db.journal.journal_data) == 3
+
+    journal_db.set(b'1', b'test-c')
+    changeset_c = journal_db.record()
+    assert len(journal_db.journal.journal_data) == 4
+
+    # TODO: Clarify
+    # This seems counterintuitive to me. Why does commiting to changeset_b
+    # Mean we are implicitly commiting to subsequent changesets?
+    # Why don't we just merge b into a but keep c seperate?
```
pipermerriam: This comment should either be rewritten or removed assuming you now understand why this happens.  Probably worth including such an explanitory comment in the actual `commit()` or underlying `Journal` method since it's not intuitive functionality unless you know how the journal is intended to be used.

evm/db/journal.py
cburgdorf: If I'm not mistaken then every API in Python that accepts `T` implicitly accepts `Union[T, None]` and every API in Python that returns `T` implicitly possibly returns `Union[T, None]` hence making this extra type annotation would be superfluous. I'm think it would be non-standard to explicitly add this.

evm/db/journal.py
cburgdorf: Eagle eyes! You're right. But honestly, do we really want to go for this perf improvement? I liked it when the perf improvement seemed to be 5.5x but now that it is down to 1.5 in a *very very* biased benchmark I'm not so crazy about it anymore. Personally I rather prefer the consistency of raising a `KeyError` for missing keys.

evm/db/journal.py
cburgdorf: Personally, I'd rather not call `reset()` to not create lots of instances that need to be garbage collected. Your call.

tests/database/test_journal_db.py
cburgdorf: Yes, sounds good.

tests/database/test_journal_db.py
cburgdorf: I removed the (left over) comment here. The doc comment of `commit()` already cover that to some extend. I think further elaboration there would make that API comment rather wordy. Maybe there's a better place in the upcoming docs for a deep dive into the internals of the `JournalDB`

evm/db/journal.py
pipermerriam: Since this should only occur once per transaction I doubt the perf hit will be measurable.

evm/db/journal.py
```diff
@@ -1,226 +1,258 @@
...
+    #
+    def get(self, key: bytes) -> bytes:
+        """
+        For key lookups we need to iterate through the changesets in reverse
+        order, returning from the first one in which the key is present.
+        """
+        # Ignored from mypy because of https://github.com/python/typeshed/issues/2078
+        for changeset_data in reversed(self.journal_data.values()):  # type: ignore
+            if key in changeset_data:
+                return changeset_data[key]
```
pipermerriam: Shouldn't this check that it's not returning `DELETED_ENTRY`?  Same issue I'm trying to understand here: https://github.com/ethereum/py-evm/pull/569/files/7a6d2e4b0a6a5349f09aaf693f1e94dd6bafe7ec#r184168693

evm/db/journal.py
cburgdorf: Yes, that *is* a bug. My comment was about if we maybe want to throw away that entire performance improvement altogether and go back to the API as it was before (the exception ping pong). However, now that I thought about it again I actually think we *should* keep the performance improvement and just fix this bug because in my previous comment I assumed we gave up on the `KeyError` exceptions for this perf improvement but that is actually not the case. The `JournalDB` continues to raise `KeyError` exceptions even with this perf improvement.

Sorry, for the total confusion. I'll just fix the bug and move on haha

evm/db/journal.py
cburgdorf: Well in case of the `Journal` (not the `JournalDB`) we may actually *want* to return the `DELETED_ENTRY` to give the higher order API the proper view to introspect. E.g. the `JournalDB` needs to know about the `DELETED_ENTRY` here in order to delete those entries from the underlying db. https://github.com/cburgdorf/py-evm/blob/7a6d2e4b0a6a5349f09aaf693f1e94dd6bafe7ec/evm/db/journal.py#L235

pipermerriam: This I don't mind.  Just wanted to make sure it was caught and handled in the `JournalDB`.

cburgdorf:
@pipermerriam I *think* we are good to squash and merge but I'm waiting for a final thumbs up just to be extra cautious.

evm/db/journal.py
```diff
@@ -1,226 +1,260 @@
...
 from evm.db.backends.base import BaseDB
 from evm.exceptions import ValidationError
 
 
-class Journal(object):
+class DeletedEntry:
+    pass
+
+
+DELETED_ENTRY = DeletedEntry()
```
carver: Purely a personal style thing, but you may like `DELETED_ENTRY = object()` in order to skip the class def'n.

cburgdorf: I like it. Give me that lightweight JavaScript feeling. I'm on mobile. Will make that change later and then squash and merge.

cburgdorf: Actually, I think I'm gonna keep the `DeletedEntry` because that's a bit more strict. Can't just push an arbitrary `object` into a `Dict[bytes, Union[bytes, DeletedEntry]]`. One could achieve the same strictness with a `NewType` but at this point I think the extra class is just fine.

carver: Yup, my style on that was developed in a dynamic type context. Makes total sense. :+1:

PR #570, opened 2018-04-18
Shard chain

jannikluhn:
### What was wrong?

Shard chain class was missing.

### How was it fixed?

A basic shard chain class has been added, along with some minor changes in the shard db.

Right now it's a very thin wrapper around the shard db. Originally I had implemented some additional functionality, but I was unsure what we'll actually end up using so I removed all but the most basic stuff. I think extending it on the go as needed is the better approach.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/shard.py
+++ b/evm/chains/shard.py
@@ -1,0 +1,69 @@
+from typing import (
+    Union,
+)
+from eth_typing import (
+    Hash32,
+)
+
+from evm.db.shard import (
+    Availability,
+    ShardDB,
+)
+
+from evm.rlp.headers import (
+    CollationHeader,
+)
+from evm.rlp.collations import (
+    Collation,
+)
+
+
+class Shard:
+
+    def __init__(self, shard_db: ShardDB, shard_id: int) -> None:
+        self.shard_id = shard_id
+        self.shard_db = shard_db
+
+    def _check_shard_id(self, header_or_collation: Union[CollationHeader, Collation]) -> None:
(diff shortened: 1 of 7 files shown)


evm/chains/shard.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Shard:
+
+    def __init__(self, shard_db: ShardDB, shard_id: int) -> None:
+        self.shard_id = shard_id
+        self.shard_db = shard_db
+
+    def _check_shard_id(self, header_or_collation: Union[CollationHeader, Collation]) -> None:
+        if header_or_collation.shard_id != self.shard_id:
+            raise ValueError("Header or collation belongs to shard {} instead of shard {}".format(
```
pipermerriam: `ValidationError` would be more inline with other library conventions.  Also, this could be done as a stand-alone validator function to reduce class size.

evm/db/shard.py
```diff
@@ -73,6 +73,12 @@ def get_collation_by_hash(self, collation_hash: Hash32) -> Collation:
     # Canonical Collations
     #
     def set_canonical(self, header: CollationHeader) -> None:
+        try:
+            self.get_header_by_hash(header.hash)
+        except CollationHeaderNotFound:
+            raise ValueError("Cannot set unknown header as canonical: {}".format(
```
pipermerriam: Maybe re-raise `CollactionHeaderNotFound` with this message instead of using `ValueError`?

PR #572, opened 2018-04-18
Shorten test ids for blob serialization tests

jannikluhn:
### What was wrong?

As pointed out by @gsalgado Travis crashed sometimes (?) when executing the blob serialization tests, due to a very long reported test name. The test name was auto generated by `@pytest.mark.parametrize` using a 1MB collation body as parameter.

### How was it fixed?

Pass an explicit test id to `@pytest.mark.parametrize` at the relevant tests.

#### Cute Animal Picture

### DIFF
--- a/tests/core/test_blob_utils.py
+++ b/tests/core/test_blob_utils.py
@@ -127,6 +127,10 @@ BLOB_SERIALIZATION_TEST_DATA = [  # [(blobs, unpadded_body), ...]
         ])
     )
 ]
+
+# By default, tests parametrized with above values have test ids which are too long to print (as
+# they contain the test data). Therefore, the following ids should be specified explicitly
+# instead:
 BLOB_SERIALIZATION_TEST_IDS = [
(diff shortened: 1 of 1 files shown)


tests/core/test_blob_utils.py
```diff
@@ -201,12 +205,20 @@ def test_chunk_root_calculation():
     assert calc_chunk_root(body) == calc_merkle_root(chunks)
 
 
-@pytest.mark.parametrize("blobs,unpadded_body", BLOB_SERIALIZATION_TEST_DATA)
+@pytest.mark.parametrize(
+    "blobs,unpadded_body",
+    BLOB_SERIALIZATION_TEST_DATA,
+    ids=BLOB_SERIALIZATION_TEST_IDS,
```
gsalgado: I didn't know about the `ids` arg to `parametrize()`, nice!  It'd be good to have a comment here explaining why we want to specify the ids ourselves, though

PR #573, opened 2018-04-18
Add an LRU cache to make_trie_root_and_nodes()

gsalgado:


gsalgado:
Unfortunately this won't be enough by itself because we'd still be making one `block.get_receipts()` call for every tx applied, and that's too expensive to run for every tx.

gsalgado:
And caching `ChainDB.get_receipts()` doesn't help because every time it is called (as part of `VM.apply_transaction()`) it is passed a different `receipt_root` for the trie containing only the receipts of the transactions that have already been applied

gsalgado:
So, this won't replace #565 but it might still be beneficial, specially since Byzantium blocks may have duplicate receipt_roots. Will benchmark to find out

gsalgado:
Now that #565 has been merged and we're no longer generating/storing partial tx/receipt tries, it makes more sense to cache the whole thing here.

evm/db/trie.py
```diff
@@ -1,19 +1,33 @@
...
-        index_key = rlp.encode(index, sedes=rlp.sedes.big_endian_int)
-        transaction_db[index_key] = rlp.encode(transaction)
 
-    return transaction_db.root_hash, transaction_db.db.wrapped_db.kv_store
+# This cache is expected to be useful when importing blocks as we call this once when importing
+# and again when validating the imported block. But it should also help for post-Byzantium blocks
+# as it's common for them to have duplicate receipt_roots. Given that, it probably makes sense to
+# use a relatively small cache size here.
+@functools.lru_cache(128)
+def _make_trie_root_and_nodes(items: Tuple[bytes, ...]) -> Tuple[bytes, Dict]:
```
pipermerriam: Return type should be `Tuple[bytes, Dict[bytes, bytes]]` right?

gsalgado: Indeed. I guess mypy considers a bare `Dict` to be `Dict[Any, Any]`

PR #575, opened 2018-04-18
WIPPiper/use rlp 1.0.0

pipermerriam:
#### Cute Animal Picture


pipermerriam:
@carver @gsalgado this is a **REALLY** dirty branch showing the changes that are needed to be compatible with the current `rlp==1.0.0b2`.

One thing I've identified that I want to fix is a prevailing assumption that RLP objects are mutable.  I'd like to default all of our RLP objects to immutable by default and fix up all of these changes to use a non-mutation based field updating pattern (currently there isn't one in the rlp library so it'll need to be added).

evm/vm/forks/frontier/headers.py
```diff
@@ -87,15 +87,16 @@ def create_frontier_header_from_parent(parent_header, **header_params):
...
-        vm.block.header.difficulty = compute_frontier_difficulty(
+    if 'timestamp' in header_params and header.block_number > 0:
+        parent_header = get_parent_header(header, vm.chaindb)
+        header.difficulty = compute_frontier_difficulty(
             parent_header,
             header_params['timestamp'],
         )
 
-    return vm.block.header
+    return header
```
carver: This file's changes seem like a great example of the payoff to the approach: not risking a partial update to the header, in fact not even changing the current block header unless the caller decides to.

p2p/peer.py
```diff
@@ -610,20 +610,20 @@ class ETHPeer(BasePeer):
...
+        msg = cast(Dict[bytes, Any], msg)
+        if msg[b'network_id'] != self.network_id:
             self.disconnect(DisconnectReason.other)
             raise HandshakeFailure(
                 "{} network ({}) does not match ours ({}), disconnecting".format(
-                    self, msg['network_id'], self.network_id))
+                    self, msg[b'network_id'], self.network_id))
         genesis = await self.genesis
-        if msg['genesis_hash'] != genesis.hash:
+        if msg[b'genesis_hash'] != genesis.hash:
```
gsalgado: I'm slightly worried about using `bytes` as dict keys here since we're kinda wired to writing dict keys as string literals, and although mistakenly trying to access a dict item here using something other than `bytes` **could** be caught by mypy, it could lead to hard-to-debug issues when not caught by mypy

pipermerriam: I agree, I'm not satisfied with this solution.  I'll rework this to do the conversion between bytes/strings at the proper location.

PR #576, opened 2018-04-19
ChainDB: cache block header RLP decoding

gsalgado:
This is a very easy/safe thing to cache and provides a big performance
improvement for fast sync as that involves frequently looking up recent
block headers.

Part of #542 as well

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -1,3 +1,4 @@
+import functools
 import itertools
 
 from abc import (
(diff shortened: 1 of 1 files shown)


evm/db/chain.py
```diff
@@ -312,11 +313,19 @@ def get_block_header_by_hash(self, block_hash: bytes) -> BlockHeader:
...
                 encode_hex(block_hash)))
-        return rlp.decode(block, sedes=BlockHeader)
+        return self._decode_block_header(header_rlp)
+
+    # When performing a chain sync (either fast or regular modes), we'll very often need to look
+    # up recent block headers to validate the chain, and decoding their RLP representation is
+    # relatively expensive so we cache that here, but use a small cache because we *should* only
+    # be looking up recent blocks.
+    @functools.lru_cache(128)
+    def _decode_block_header(self, header_rlp: bytes) -> BlockHeader:
```
pipermerriam: Probably good to go ahead and wrap this in `staticmethod` *or* elevate it to a top level module function.

gsalgado: Makes sense, went with a top-level function

Issue #557, opened 2018-04-15
Create a complete, user-friendly and up-to-date documentation


cburgdorf:
This is a fresh attempt to organize efforts to release a complete, user friendly and up to date documentation. I know that we already have two existing (https://github.com/ethereum/py-evm/issues/19, https://github.com/ethereum/py-evm/issues/177) github issues regarding documentation and the existing content was (and further will be) incorporated in this issue.

Since it is planned that I'll be dedicating a significant amount of my time to the documentation I wanted to set up a new issue that I can maintain and keep up to date as I go. In that sense, this ticket should supersede the existing two tickets and I propose that we close them.

### References

We include a collection of links to other ethereum client documentations here as a reference. There may be further links to existing docs directly included in the table of contents.

- [Geth Documentation](https://github.com/ethereum/go-ethereum/wiki)
- [Parity Documentation](https://wiki.parity.io/Setup)
- [Ethereum Harmony Documentation](https://github.com/ether-camp/ethereum-harmony/wiki/Setting-up-a-private-network)
- [cpp-ethereum Documentation](http://www.ethdocs.org/en/latest/ethereum-clients/cpp-ethereum/)

#### Other inspirations

- [Trio Docs](http://trio.readthedocs.io) (Nice rtd gems)
- [Solidity docs](http://solidity.readthedocs.io/en/v0.4.23/index.html)
- [gotdotengine](https://github.com/godotengine/godot-docs/blob/3.0/index.rst) (nice rtd gems)
- Would be nice to have these copy-n-past'able [terminal demos](https://openzeppelin.org/) in the guide

### Style / Prose / Grammar

The documentation should follow a common style in regard to grammar, wording and structure. It's planned to flesh out some basic rules but for now, a lot of great rules can be derived from this [style guide by Adobe](https://www.adobe.io/about/contributing/doc-style_master.html).

### Audience

We assume the audience of readers will range from regular blockchain enthusiasts to dApp developers as well as protocol engineers and scientists.

### Outline

The following outline assumes (which is debatable as everything) that we will clearly communicate the different roles of Trinity and Py-EVM but will still maintain a common end point to host the documentation.

#### Index

High level description of the library. Links to github repository, gitter chat channel, table of contents for the remainder of the documentation.

- #### Introduction
  - Overview
  - Project goals
  - Role of Trinity and Py-EVM
- #### Installation ([Geth Example](https://ethereum.github.io/go-ethereum/install/))
  - Using pip
  - From source
  - …
[...]

PR #581, opened 2018-04-20
WIP: new rlp library with only immutable objects

pipermerriam:
Continuation of: https://github.com/ethereum/py-evm/pull/575

### What was wrong?

WIP

#### Cute Animal Picture


evm/vm/base.py
```diff
@@ -133,45 +133,48 @@ def execute_bytecode(self,
...
         for transaction in block.transactions:
             self.apply_transaction(transaction)
 
         # transfer the list of uncles.
-        self.block.uncles = block.uncles
+        self.block = self.block.copy(
+            uncles=block.uncles,
+            transactions=block.transactions,
+            header=self.block.header,
+        )
```
carver: This repeated assignment of `self.block` suggests to me that this might be another place for the context manager.

```py
        block_builder = self.block.build_copy(
              transactions=block.transactions,
              uncles=block.uncles,
        )
        with block_builder as next_block:
            next_block.header = self.configure_header(**block.header.as_dict())
 
            # run all of the transactions.
            for transaction in next_block.transactions:
                self.apply_transaction(transaction, to_block=next_block)

         self.block = next_block
```

So either we:
1. embrace the mutable block within this context, and have `apply_transaction()` modify the block in place (shown above, but still not assign to `self.block`), or...
2. return a copy of the new block after applying the transaction

evm/vm/base.py
```diff
@@ -133,45 +133,48 @@ def execute_bytecode(self,
...
 
         # run all of the transactions.
         for transaction in block.transactions:
             self.apply_transaction(transaction)
 
         # transfer the list of uncles.
-        self.block.uncles = block.uncles
+        self.block = self.block.copy(
+            uncles=block.uncles,
+            transactions=block.transactions,
```
gsalgado: It feels wrong to overwrite `transactions` immediately after we've applied those transactions to the block.

evm/vm/base.py
```diff
@@ -133,45 +133,48 @@ def execute_bytecode(self,
...
+            header=self.block.header,
```
gsalgado: This will not change anything, right, since we're copying `self.block`?

evm/vm/base.py
```diff
@@ -133,45 +133,48 @@ def execute_bytecode(self,
...
         self.chaindb.persist_trie_data_dict(receipt_kv_nodes)
-        self.block.header.transaction_root = tx_root_hash
-        self.block.header.receipt_root = receipt_root_hash
 
-        self.pack_block(block, *args, **kwargs)
+        with self.block.header.build_copy() as header:
+            header.transaction_root = tx_root_hash
+            header.receipt_root = receipt_root_hash
+
+        block = self.pack_block(self.block.copy(header=header), *args, **kwargs)
```
gsalgado: I don't quite like the fact that `header` is used after we leave the contextmanager... Not sure if that's an indication a contextmanager is not the right solution here or if maybe we should just move the `pack_block()` call inside the contextmanager. Thoughts?

gsalgado: Ok, that feeling came from me not understanding the purpose of the context manager... it was not immediately clear to me that what it does is make the object mutable inside the context, so I guess it makes sense to leave `pack_block()` outside of it as it shouldn't mutate `header`. Still, it doesn't feel right to use `header` outside of the contextmanager, specially since it gives no indication that it's no longer mutable after you exit the context. Will try to think of alternatives, but just thought I'd write this down here in case you guys have any ideas

pipermerriam: Linking to https://github.com/ethereum/pyrlp/pull/77 where this API is being created.

I generally agree.  see [this comment](https://github.com/ethereum/pyrlp/pull/77#pullrequestreview-114145636) for another reason why this API is probably not quite right.

pipermerriam: @carver @gsalgado have a look at https://github.com/ethereum/pyrlp/pull/78

tests/database/test_chaindb.py
```diff
@@ -53,9 +53,11 @@
...
 def set_empty_root(chaindb, header):
     root_hash = get_empty_root_hash(chaindb)
-    header.transaction_root = root_hash
-    header.receipt_root = root_hash
-    header.state_root = root_hash
+    return header.copy(
+        transaction_root=root_hash,
+        receipt_root=root_hash,
+        state_root=root_hash,
+    )
```
carver: This is super clear, I'm enjoying reading the `copy()` and `build_changeset()` updates.

evm/vm/forks/byzantium/headers.py
```diff
@@ -70,12 +70,13 @@ def create_byzantium_header_from_parent(parent_header, **header_params):
 def configure_byzantium_header(vm, **header_params):
     validate_header_params_for_configuration(header_params)
 
-    with vm.block.header.build_copy(**header_params) as header:
-        if 'timestamp' in header_params and vm.block.header.block_number > 0:
-            parent_header = get_parent_header(header, vm.chaindb)
-            header.difficulty = compute_byzantium_difficulty(
+    with vm.block.header.build_changeset(**header_params) as changeset:
+        if 'timestamp' in header_params and changeset.block_number > 0:
+            parent_header = get_parent_header(changeset.commit(), vm.chaindb)
```
carver: Ah, interesting, it surprised me to call `commit()` and keep changing `changeset`. Maybe it's the name `commit()` which implied some finality to me. Some alternatives to consider:

- `changeset.frozen()`
- `changeset.rlp_obj()`
- `changeset.copy()`
- `changeset.emit()`
- `changeset.produce()`
- `changeset.hatch()`
- `changeset.stamp()`
- `changeset.sealed()`

_I suppose this comment really goes in pyrlp_

pipermerriam: - `changeset.freeze()` would work for me.  I'll modify `commit()` to be equivalent to `freeze() -> close()`

evm/vm/base.py
pipermerriam: Probably a mistake, I'll revisit.

tests/p2p/conftest.py
```diff
@@ -1,2 +1,27 @@
...
+    level = logging.DEBUG
+    level = logging.INFO
+
+    logger.setLevel(level)
+    handler.setLevel(level)
+
+    logger.addHandler(handler)
+
+    return logger
+"""
```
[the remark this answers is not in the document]
cburgdorf: I linked this here https://github.com/ethereum/py-evm/issues/557#issuecomment-384255221 Will take care. Whenever something about the docs comes to your mind, feel free to drop it there and I'll make sure to pick it up somehow.

tests/p2p/conftest.py
```diff
@@ -1,2 +1,27 @@
...
+
+"""
+import pytest
+
+@pytest.fixture(autouse=True, scope="session")
+def vm_logger():
+    import logging
+    import sys
+
+    logger = logging.getLogger('p2p')
```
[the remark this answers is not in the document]
gsalgado: Oh, now that I see which file this is in, I guess that's what you want, although the function's name doesn't match it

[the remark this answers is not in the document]
pipermerriam: copy/paste error.  I'll clean it up.

PR #583, opened 2018-04-23
p2p: Testing P2P handshake

hwwhww:
### What was wrong?
After auth handshake, the server should apply P2P handshake.

### How was it fixed?
1. Add `Server.apply_p2p_handshake(peer)` function to apply p2p handshake.
2. Add `tests/p2p/test_server.py::test_two_servers` for simulating two servers. `server_1` is the bootstrap node for `server_2`, `server_2` will try to connect to `server_1`.
3. Use `DumbPeer` instead of `ETHPeer` for testing. (**No sub-protocol handshake yet**)
4. Add a testing daemon in `server.py` (basically, referred to `chain.py`).
    * Run `server_1` (the bootstrap node):
        ```bash
        python -m p2p.server -debug -server-address 127.0.0.1:33333 -privkey b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291
        ```
    * Run `server_2`:
        ```bash
        python -m p2p.server -debug -server-address 127.0.0.1:30303 -bootstrap enode://ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31387574077f301b421bc84df7266c44e9e6d569fc56be00812904767bf5ccd1fc7f@127.0.0.1:33333 -privkey 49a7b37aa6f6645917e7b807e9d1c00d4fa71f18343b0d4122a4d2df64dd6fee
        ```

    * Result of `server_1`:
        ```bash
        ➜ python -m p2p.server -debug -server-address 127.0.0.1:33333 -privkey b71c71a67e1177ad4e901695e1b4b9ee17ae16c6668d313eac2f96dbcda3f291
        19:00:59 INFO: Running PeerPool...
        19:00:59 WARNING: Cannot get 5 nodes as RoutingTable contains only 0 nodes
        19:00:59 DEBUG: no peer connected
        19:00:59 INFO: lookup finished for 85658218438996085151406112006981061491734845553074397258703399980243123788092: []
        19:00:59 INFO: Running server...
        19:00:59 INFO: enode://ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31387574077f301b421bc84df7266c44e9e6d569fc56be00812904767bf5ccd1fc7f@127.0.0.1:33333
        19:01:00 DEBUG: no peer connected
        19:01:01 WARNING: Cannot get 5 nodes as RoutingTable contains only 0 nodes
        19:01:01 DEBUG: no peer connected
        19:01:02 DEBUG: no peer connected
        19:01:03 WARNING: Cannot get 5 nodes as RoutingTable contains only 0 nodes
        19:01:03 DEBUG: no peer connected
        19:01:04 DEBUG: Receiving handshake...
        19:01:04 DEBUG: peers: 1
        ```
    * Result of `server_2`:
        ```bash
        ➜ python -m p2p.server -debug -server-address 127.0.0.1:30303 -bootstrap enode://ca634cae0d49acb401d8a4c6b6fe8c55b70d115bf400769cc1400f3258cd31387574077f301b421bc84df7266c44e9e6d569fc56be00812904767bf5ccd1fc7f@127.0.0.1:33333 -privkey 49a7b37aa6f6645917e7b807e9d1c00d4fa71f18343b0d4122a4d2df64dd6fee
        19:01:04 INFO: Running PeerPool...
        19:01:04 WARNING: Cannot get 5 nodes as RoutingTable contains only 0 nodes
        19:01:04 DEBUG: no peer connected
        19:01:04 INFO: lookup finished for 6337412118514148720559954161251210272405612403602241289921825684304857839603: []
        19:01:04 INFO: Running server...
        19:01:04 INFO: enode://fda1cff674c90c9a197539fe3dfb53086ace64f83ed7c6eabec741f7f381cc803e52ab2cd55d5569bce4347107a310dfd5f88a010cd2ffd1005ca406f1842877@127.0.0.1:30303
        19:01:04 INFO: Successfully connected to DumbPeer <Node(0xca63@127.0.0.1)>
        19:01:05 DEBUG: peers: 1
        ````

#### Cute Animal Picture

raccoon

### DIFF
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -60,3 +60,7 @@ ROPSTEN_BOOTNODES = [
     # 'enode://6332792c4a00e3e4ee0926ed89e0d27ef985424d97b6a45bf0f23e51f0dcb5e66b875777506458aea7af6f9e4ffb69f43f3778ee73c81ed9d34c51c4b16b0b0f@52.232.243.152:30303',  # noqa: E501
     # 'enode://94c15d1b9e2fe7ce56e458b9a3b672ef11894ddedd0c6f247e0f1d3487f52b66208fb4aeb8179fce6e3a749ea93ed147c37976d67af557508d199d9594c35f09@192.81.208.223:30303',  # noqa: E501
 ]
+
+# Default configuration
+# Minimum peers number, we'll try to keep open connections to at least this number of peers
+DEFAULT_MIN_PEERS = 5
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -78,6 +78,7 @@ from p2p.p2p_proto import (
 
 from .constants import (
     CONN_IDLE_TIMEOUT,
+    DEFAULT_MIN_PEERS,
     HANDSHAKE_TIMEOUT,
     HEADER_LEN,
     MAC_LEN,
@@ -649,7 +650,7 @@ class PeerPool:
                  network_id: int,
                  privkey: datatypes.PrivateKey,
                  discovery: DiscoveryProtocol,
-                 min_peers: int = 10,
+                 min_peers: int = DEFAULT_MIN_PEERS,
                  ) -> None:
         self.peer_class = peer_class
         self.chaindb = chaindb
@@ -673,8 +674,14 @@ class PeerPool:
         if subscriber in self._subscribers:
             self._subscribers.remove(subscriber)
 
+    def start_peer(self, peer):
+        asyncio.ensure_future(peer.run(finished_callback=self._peer_finished))
+        self.add_peer(peer)
+
     def add_peer(self, peer):
+        self.logger.debug('Adding peer (%s) ...', str(peer))
         self.connected_nodes[peer.remote] = peer
+        self.logger.debug('Number of peers: %d', len(self.connected_nodes))
         for subscriber in self._subscribers:
             subscriber.register_peer(peer)
 
@@ -757,7 +764,10 @@ class PeerPool:
             await self._connect_to_nodes(self._get_random_bootnode())
 
     def _get_random_bootnode(self) -> Generator[Node, None, None]:
-        yield random.choice(self.discovery.bootstrap_nodes)
+        if self.discovery.bootstrap_nodes:
+            yield random.choice(self.discovery.bootstrap_nodes)
+        else:
+            self.logger.warning('No bootstrap_nodes')
 
     async def _connect_to_nodes(self, nodes: Generator[Node, None, None]) -> None:
         for node in nodes:
@@ -767,8 +777,7 @@ class PeerPool:
             peer = await self.connect(node)
             if peer is not None:
                 self.logger.info("Successfully connected to %s", peer)
-                asyncio.ensure_future(peer.run(finished_callback=self._peer_finished))
-                self.add_peer(peer)
+                self.start_peer(peer)
                 if len(self.connected_nodes) >= self.min_peers:
                     return
 
--- a/p2p/server.py
+++ b/p2p/server.py
@@ -4,7 +4,7 @@ import secrets
 
 from typing import (
     List,
-    Tuple,
+    Type,
 )
 
 from eth_keys import datatypes
@@ -13,27 +13,33 @@ from eth_utils import big_endian_to_int
 
 from evm.db.chain import AsyncChainDB
 from p2p.auth import (
-    decode_auth_plain,
-    decode_auth_eip8,
+    decode_authentication,
     HandshakeResponder,
 )
-from p2p.cancel_token import CancelToken
+from p2p.cancel_token import (
+    CancelToken,
+    wait_with_token,
+)
 from p2p.constants import (
     ENCRYPTED_AUTH_MSG_LEN,
+    DEFAULT_MIN_PEERS,
+    HANDSHAKE_TIMEOUT,
     HASH_LEN,
 )
 from p2p.discovery import DiscoveryProtocol
-from p2p.ecies import ecdh_agree
-from p2p.exceptions import DecryptionError
+from p2p.exceptions import (
+    DecryptionError,
+    HandshakeFailure,
+)
 from p2p.kademlia import (
     Address,
     Node,
 )
 from p2p.peer import (
+    BasePeer,
     ETHPeer,
     PeerPool,
 )
-from p2p.utils import sxor
 
 
 class Server:
@@ -46,17 +52,27 @@ class Server:
                  server_address: Address,
                  chaindb: AsyncChainDB,
                  bootstrap_nodes: List[str],
-                 network_id: int
+                 network_id: int,
+                 min_peers: int = DEFAULT_MIN_PEERS,
+                 peer_class: Type[BasePeer] = ETHPeer,
                  ) -> None:
         self.cancel_token = CancelToken('Server')
         self.chaindb = chaindb
         self.privkey = privkey
         self.server_address = server_address
         self.network_id = network_id
+        self.peer_class = peer_class
         # TODO: bootstrap_nodes should be looked up by network_id.
         discovery = DiscoveryProtocol(
             self.privkey, self.server_address, bootstrap_nodes=bootstrap_nodes)
-        self.peer_pool = PeerPool(ETHPeer, self.chaindb, self.network_id, self.privkey, discovery)
+        self.peer_pool = PeerPool(
+            peer_class,
+            self.chaindb,
+            self.network_id,
+            self.privkey,
+            discovery,
+            min_peers=min_peers,
+        )
 
     async def start(self) -> None:
         self._server = await asyncio.start_server(
@@ -68,6 +84,12 @@ class Server:
     async def run(self) -> None:
         await self.start()
         self.logger.info("Running server...")
+        self.logger.info(
+            "enode://%s@%s:%s",
+            self.privkey.public_key.to_hex()[2:],
+            self.server_address.ip,
+            self.server_address.tcp_port,
+        )
         await self.cancel_token.wait()
         await self.stop()
 
@@ -80,10 +102,22 @@ class Server:
 
     async def receive_handshake(
             self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
-        # Use reader to read the auth_init msg until EOF
-        msg = await reader.read(ENCRYPTED_AUTH_MSG_LEN)
+        await wait_with_token(
+            self._receive_handshake(reader, writer),
+            token=self.cancel_token,
+            timeout=HANDSHAKE_TIMEOUT,
+        )
 
-        # Use HandshakeResponder.decode_authentication(auth_init_message) on auth init msg
+    async def _receive_handshake(
+            self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        self.logger.debug("Receiving handshake...")
+        # Use reader to read the auth_init msg until EOF
+        msg = await wait_with_token(
+            reader.read(ENCRYPTED_AUTH_MSG_LEN),
+            token=self.cancel_token,
+        )
+
+        # Use decode_authentication(auth_init_message) on auth init msg
         try:
             ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
                 msg, self.privkey)
@@ -91,13 +125,16 @@ class Server:
         except DecryptionError:
             msg_size = big_endian_to_int(msg[:2])
             remaining_bytes = msg_size - ENCRYPTED_AUTH_MSG_LEN + 2
-            msg += await reader.read(remaining_bytes)
+            msg += await wait_with_token(
+                reader.read(remaining_bytes),
+                token=self.cancel_token,
+            )
             ephem_pubkey, initiator_nonce, initiator_pubkey = decode_authentication(
                 msg, self.privkey)
 
         # Get remote's address: IPv4 or IPv6
-        ip, port, *_ = writer.get_extra_info("peername")
-        remote_address = Address(ip, port)
+        ip, socket, *_ = writer.get_extra_info("peername")
+        remote_address = Address(ip, socket)
 
         # Create `HandshakeResponder(remote: kademlia.Node, privkey: datatypes.PrivateKey)` instance
         initiator_remote = Node(initiator_pubkey, remote_address)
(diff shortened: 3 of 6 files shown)


p2p/server.py
```diff
@@ -46,17 +51,27 @@ def __init__(self,
                  server_address: Address,
                  chaindb: AsyncChainDB,
                  bootstrap_nodes: List[str],
-                 network_id: int
+                 network_id: int,
+                 min_peers: int = 5,
```
cburgdorf: should this rather go into `p2p/constants.py` maybe?

p2p/peer.py
```diff
@@ -754,7 +760,8 @@ def add_peer(self, peer):
             await self._connect_to_nodes(self._get_random_bootnode())
 
     def _get_random_bootnode(self) -> Generator[Node, None, None]:
-        yield random.choice(self.discovery.bootstrap_nodes)
+        if self.discovery.bootstrap_nodes:
+            yield random.choice(self.discovery.bootstrap_nodes)
```
gsalgado: Maybe log a warning when there are no bootstrap nodes, as that will probably lead to us not finding any nodes to connect to. Or maybe we should make `bootstrap_nodes` required

p2p/peer.py
```diff
@@ -670,8 +670,14 @@ def unsubscribe(self, subscriber: PeerPoolSubscriber) -> None:
         if subscriber in self._subscribers:
             self._subscribers.remove(subscriber)
 
+    def start_peer(self, peer):
+        asyncio.ensure_future(peer.run(finished_callback=self._peer_finished))
+        self.add_peer(peer)
+
     def add_peer(self, peer):
+        self.logger.debug('Adding peer...')
```
gsalgado: Care to include the peer in the msg as well?

p2p/server.py
```diff
@@ -122,36 +143,121 @@ def __init__(self,
...
-    """
-    if len(ciphertext) < ENCRYPTED_AUTH_MSG_LEN:
-        raise ValueError("Auth msg too short: {}".format(len(ciphertext)))
-    elif len(ciphertext) == ENCRYPTED_AUTH_MSG_LEN:
-        sig, initiator_pubkey, initiator_nonce, _ = decode_auth_plain(
-            ciphertext, privkey)
+
+        await self.apply_p2p_handshake(peer)
+
+    async def apply_p2p_handshake(self, peer: BasePeer) -> None:
```
gsalgado: Nitpick: it sounds weird to _apply_ a handshake. Also, `do_p2p_handshake()` would make it consistent with the name used in other classes

p2p/server.py
```diff
@@ -122,36 +143,121 @@ def __init__(self,
...
+        finish_p2p_handshake = False
+        try:
+            # P2P Handshake.
+            await wait_with_token(
+                peer.do_p2p_handshake(),
+                token=peer.cancel_token,
```
gsalgado: This should use the server's cancel token. Basically, any `Server` methods that use `wait_with_token()` should use the server's token

p2p/server.py
```diff
@@ -122,36 +143,121 @@ def __init__(self,
...
+                timeout=HANDSHAKE_TIMEOUT,
```
gsalgado: This is a rather long timeout as it is intended to be used as the overall timeout for both the auth and p2p handshakes.  I think the best way forward is to not use `wait_with_token()` here, rename the current `receive_handshake()` to `_receive_handshake()` and create a new `receive_handshake()` like this:
```
async def receive_handshake(...):
    return wait_with_token(self._receive_handshake(...), token=self.cancel_token, timeout=HANDSHAKE_TIMEOUT)
```

Also, while you're at it, the `reader.read()` calls in `receive_handshake()` need to be wrapped with `wait_with_token()` as well

p2p/server.py
```diff
@@ -122,36 +143,121 @@ def __init__(self,
...
+            )
+            finish_p2p_handshake = True
+        except (HandshakeFailure, asyncio.TimeoutError) as e:
+            self.logger.debug('Unable to finish P2P handshake: {}'.format(e))
+
+        # Run peer and add peer.
+        if finish_p2p_handshake:
+            self.peer_pool.start_peer(peer)
```
gsalgado: You could move this into an else block and get rid of `finish_p2p_handshake`

p2p/server.py
```diff
@@ -122,36 +143,121 @@ def __init__(self,
...
+        level=logging.INFO, format='%(asctime)s %(levelname)s: %(message)s', datefmt='%H:%M:%S')
+    log_level = logging.INFO
+    if args.debug:
+        log_level = logging.DEBUG
+    logging.getLogger('p2p.server.Server').setLevel(log_level)
+
+    loop = asyncio.get_event_loop()
+    # Use a ProcessPoolExecutor as the default because the tasks we want to offload from the main
+    # thread are cpu intensive.
+    loop.set_default_executor(ProcessPoolExecutor())
```
gsalgado: No need for a `ProcessPoolExecutor` here -- that is only used when we are doing a chain sync

tests/p2p/dumb_peer.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+from p2p import protocol
+
+
+class DumbPeer(BasePeer):
+    _supported_sub_protocols = [eth.ETHProtocol]
+    sub_proto = None  # type: eth.ETHProtocol
+    head_td = None  # type: int
+    head_hash = None  # type: bytes
```
gsalgado: Why not subclass `ETHPeer`?

tests/p2p/test_server.py
```diff
@@ -29,25 +31,58 @@ def get_open_port():
...
+
 # FIXME: Instead of starting an actual server and send data over the wire, this test should create
 # StreamReaders and StreamWriters and connect them directly, like in test_auth.py. That way we
 # don't need those "strategically" placed sleep() calls.
 @pytest.mark.asyncio
 async def test_server_authenticates_incoming_connections(server, event_loop):
     await server.start()
+
+    # Only test the authentication in this test.
+    ETHPeer.do_p2p_handshake = mock_do_p2p_handshake
```
gsalgado: This is not a good idea as any test that runs after this will use the mock method implementation as well. You should use https://docs.pytest.org/en/latest/monkeypatch.html instead

p2p/server.py
hwwhww: I guess the default config will be configurated with `evm.utils.env`? Right now I can add a `DEFAULT_MIN_PEERS`. Referred to [geth](https://github.com/ethereum/go-ethereum/blob/77da203547b83c70d12552a86bb7beee52f6383e/eth/sync.go#L148), maybe set to `5`?
cc @gsalgado

p2p/peer.py
hwwhww: I'd like to add warning message because sometimes some nodes really don't have bootstrap nodes for testing and testnet.

tests/p2p/dumb_peer.py
hwwhww: I thought both works, but `ETHPeer` does make sense.   👍 Will update it.

p2p/server.py
hwwhww: @gsalgado  Thank you so much! For `reader.read()`, I added `wait_with_token` with `timeout=REPLY_TIMEOUT` (p.s. `HANDSHAKE_TIMEOUT = 5`, `REPLY_TIMEOUT = 3`), does that make sense?

p2p/constants.py
```diff
@@ -60,3 +60,6 @@
     # 'enode://6332792c4a00e3e4ee0926ed89e0d27ef985424d97b6a45bf0f23e51f0dcb5e66b875777506458aea7af6f9e4ffb69f43f3778ee73c81ed9d34c51c4b16b0b0f@52.232.243.152:30303',  # noqa: E501
     # 'enode://94c15d1b9e2fe7ce56e458b9a3b672ef11894ddedd0c6f247e0f1d3487f52b66208fb4aeb8179fce6e3a749ea93ed147c37976d67af557508d199d9594c35f09@192.81.208.223:30303',  # noqa: E501
 ]
+
+# Default configuration
```
gsalgado: That comment doesn't help much. Maybe
```
# We'll try to keep open connections to at least this number of peers
```

p2p/peer.py
```diff
@@ -670,8 +671,14 @@ def unsubscribe(self, subscriber: PeerPoolSubscriber) -> None:
...
             self._subscribers.remove(subscriber)
 
+    def start_peer(self, peer):
+        asyncio.ensure_future(peer.run(finished_callback=self._peer_finished))
+        self.add_peer(peer)
+
     def add_peer(self, peer):
+        self.logger.debug('Adding peer ({}) ...'.format(peer))
         self.connected_nodes[peer.remote] = peer
+        self.logger.debug('Number of peers: {}'.format(len(self.connected_nodes)))
```
gsalgado: When logging we always use string interpolation -- .format() is only used for exception messages

p2p/server.py
```diff
@@ -68,6 +85,11 @@ def __init__(self,
     async def run(self) -> None:
         await self.start()
         self.logger.info("Running server...")
+        self.logger.info("enode://{}@{}:{}".format(
+            self.privkey.public_key.to_hex()[2:],
+            self.server_address.ip,
+            self.server_address.tcp_port,
+        ))
```
gsalgado: should use string interpolation here too

p2p/server.py
```diff
@@ -80,24 +102,41 @@ def __init__(self,
...
+
+    async def _receive_handshake(
+            self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        self.logger.debug("Receiving handshake...")
         # Use reader to read the auth_init msg until EOF
-        msg = await reader.read(ENCRYPTED_AUTH_MSG_LEN)
+        msg = await wait_with_token(
+            reader.read(ENCRYPTED_AUTH_MSG_LEN),
+            token=self.cancel_token,
+            timeout=REPLY_TIMEOUT,
```
gsalgado: Maybe it doesn't make sense to enforce a timeout on every reply here. It should be fine if a peer takes 90% of the overall handshake timeout to send the auth handshake and then less than 10% to send the p2p handshake

p2p/server.py
```diff
@@ -122,36 +161,109 @@ def __init__(self,
...
-            ciphertext, privkey)
+
+        await self.do_p2p_handshake(peer)
+
+    async def do_p2p_handshake(self, peer: BasePeer) -> None:
+        try:
+            # P2P Handshake.
+            await peer.do_p2p_handshake(),
+        except (HandshakeFailure, asyncio.TimeoutError) as e:
+            self.logger.debug('Unable to finish P2P handshake: {}'.format(e))
```
gsalgado: String interpolation here too

p2p/server.py
```diff
@@ -122,36 +161,109 @@ def __init__(self,
...
+        privkey = keys.PrivateKey(decode_hex(args.privkey))
     else:
-        sig, initiator_pubkey, initiator_nonce, _ = decode_auth_eip8(
-            ciphertext, privkey)
+        privkey = ecies.generate_privkey()
+
+    ip, port = args.server_address.split(':')
+    server_address = Address(ip, port)
+    if args.list:
+        bootstrap_nodes = bootstrap_nodes = args.list.split(',')
```
gsalgado: Is there a spare `bootstrap_nodes = ` here?

tests/p2p/test_server.py
```diff
@@ -60,13 +93,49 @@ def server():
...
+    await server_2.start()
+
+    # server_2 will try to connect to server_1.
+    asyncio.ensure_future(server_2.peer_pool.run())
 
-async def send_auth_msg_to_server(address, messages):
-    reader, writer = await asyncio.open_connection(address.ip, address.udp_port)
-    writer.write(messages)
-    await writer.drain()
+    await asyncio.sleep(0.2)
```
gsalgado: Can we await for an event or something else that would indicate the peers have connected? Ideally we should have no sleep()s in tests as those make them very fragile

tests/p2p/test_server.py
```diff
@@ -60,13 +93,49 @@ def server():
...
+
+    assert len(server_1.peer_pool.connected_nodes) == 1
+    assert len(server_2.peer_pool.connected_nodes) == 1
+
+    # Stop the servers.
+    asyncio.ensure_future(server_1.stop())
+    await asyncio.sleep(0.2)
+    asyncio.ensure_future(server_2.stop())
+    await asyncio.sleep(0.2)
```
gsalgado: Why can't you `await server_{1,2}.stop()` here?

p2p/peer.py
pipermerriam: FYI, the reason for that is that logging aggregators can group logging messages if the message is done with `logger.info("message with %s placeholders", thing_for_placeholder)` even when `thing_for_placeholder` is different for each message.

It can't do that aggregation when the string has already been formatted.

tests/p2p/test_server.py
hwwhww: @gsalgado In order to remove the sleep()s, I changed to test the connecting via calling `peer_pool._connect_to_nodes(nodes)` directly instead of setting bootstrap nodes.

tests/p2p/dumb_peer.py
```diff
@@ -0,0 +1,23 @@
...
+)
+
+from p2p import protocol
+
+
+class DumbPeer(ETHPeer):
+    _supported_sub_protocols = [eth.ETHProtocol]
+    sub_proto = None  # type: eth.ETHProtocol
+    head_td = None  # type: int
+    head_hash = None  # type: bytes
```
gsalgado: The reason I suggest you inherit from `ETHPeer` is just so that you don't have to duplicate those attributes here. You should be able to remove them now

tests/p2p/test_server.py
```diff
@@ -29,44 +31,88 @@ def get_open_port():
...
+    monkeypatch.setattr(ETHPeer, 'do_p2p_handshake', mock_do_p2p_handshake)
+
     await server.start()
-    await send_auth_msg_to_server(SERVER_ADDRESS, eip8_values['auth_init_ciphertext'])
+    reader, writer = await asyncio.open_connection(SERVER_ADDRESS.ip, SERVER_ADDRESS.tcp_port)
+    writer.write(eip8_values['auth_init_ciphertext'])
+    await writer.drain()
+
     # Yield control just so that the server has a chance to process the auth msg and respond to
     # the remote.
```
gsalgado: This comment doesn't apply anymore, right?

p2p/server.py
```diff
@@ -122,36 +159,109 @@ def __init__(self,
...
+            # P2P Handshake.
+            await peer.do_p2p_handshake(),
+        except (HandshakeFailure, asyncio.TimeoutError) as e:
+            self.logger.debug('Unable to finish P2P handshake: %s', str(e))
+        else:
+            # Run peer and add peer.
+            self.peer_pool.start_peer(peer)
+
+
+def _test() -> None:
```
pipermerriam: It'd be nice to actually have these run in our CI environment.  Thoughts on adding that as an extra CI run?  Alternatively, it seems like we should move away from this pattern since this code ends up rotting since it's not covered by CI.

@gsalgado you have a better handle on how these are used and how critical they are.  Do you think we can phase them out?  Roll them into CI?  Move them to a `scripts` directory of some sort with some basic tests to demonstrate they *probabalistically* work?

hwwhww: No objection to running in CI.
And also, I think it makes sense if we have some scripts for `trinity` daemon?

gsalgado: I understand your concern but most of those (chain.py, state.py, sync.py) need to run against a fully-synced geth instance and it'd be really tricky to try and somehow automate checks to ensure they work. Also, and I know this is not a good argument, this is what I use (several times a day, actually) to manually test all the p2p code, so it is unlikely to bitrot in the near future

As for moving them to a `scripts` dir, I think that may be a good idea as it'd force us to refactor some of this into helpers to reduce duplication, but at the same time I like the fact that now they live right next to the code they exercise, so I can just run, say, `python -m p2p.chain` to test the chain syncer

pipermerriam: @gsalgado since you are the primary consumer of these I'm going to shelve this idea for now since they are clearly working for you and I don't want to fix what isn't broken.  Once we're in public release mode we can try and address this.

hwwhww:
@pipermerriam @gsalgado 
Is it okay to merge this PR?

gsalgado:
Yeah, maybe just squash all those commits?

PR #585, opened 2018-04-23
Set fixed pytest cache dir

cburgdorf:
What was wrong?

The directory was renamed to `.pytest_cache` in Pytest 3.4.0 so the directory was coming up in version control when using a newer Pytest version

### How was it fixed?

Configured it to be set to `.cache` as it was before.

#### Cute Animal Picture

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,6 +43,7 @@ docs/modules.rst
 
 # pytest
 .cache/
+.pytest_cache/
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
I think it'd be better to just add the new directory to the `.gitignore`

cburgdorf:
Sure, no problem. Updated.

PR #586, opened 2018-04-23
p2p: StateDownloader no longer sends requests to peers with pending requests

gsalgado:
This way we make better use of peers with higher throughput and achieve
much faster syncs

### DIFF
--- a/p2p/state.py
+++ b/p2p/state.py
@@ -1,5 +1,6 @@
 import asyncio
 import logging
+import secrets
 import time
 from typing import (  # noqa: F401
     Any,
@@ -8,6 +9,8 @@ from typing import (  # noqa: F401
     List,
     Set,
 )
+
+from cytoolz.itertoolz import partition_all
 
 import rlp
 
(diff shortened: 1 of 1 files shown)


p2p/state.py
```diff
@@ -52,12 +55,28 @@ def __init__(self,
...
         self.cancel_token = CancelToken('StateDownloader')
         if token is not None:
             self.cancel_token = self.cancel_token.chain(token)
 
     def register_peer(self, peer: BasePeer) -> None:
-        asyncio.ensure_future(self.handle_peer(cast(ETHPeer, peer)))
+        peer = cast(ETHPeer, peer)
+        asyncio.ensure_future(self.handle_peer(peer))
+        # A peer may get disconnected while we're waiting for a reply, so when it comes back we
+        # clear this.
```
pipermerriam: I don't understand this comment, specifically the "when it comes back".  I assume *it* is the peer, but what is it coming back from?

gsalgado: Yes, _it_ is the peer, and it is coming back from the `PeerPool` -- this is the `.register_peer()` method, which is the callback used by `PeerPool` when it connects to a new peer. This is actually not needed, though, as when the `PeerPool` calls this we'll be passed a different `BasePeer` instance even if it's for a peer we were connected before. I thought it was needed because the `Node` class defines an `__eq__()` method that compares only the node's public ID, but that's not the case for `BasePeer`. I'm removing this, and it shouldn't be a problem that we're not removing old peers from that dict here because the `_periodically_retry_timedout()` coroutine does that

PR #588, opened 2018-04-24
Improve various parts of the documentation

cburgdorf:
### What was wrong?

There are many aspects of the documentation that need work and this PR doesn't aim to "fix" any of them fully but rather provides a first step in an iterative processes. With that in mind, let's focus on what it does.

### How was it fixed?

1. the TOC was restructured and spiced up

2. Elaborated on the index file. It's worth pointing out that the index file includes contents from the introduction files (via include directive). In the future, introduction and index file may diverge but for now it seemed most practical to just go this way to have a "Introduction" link in the toc that shows the same content as the landing page.

3. **Most importantly** all the existing api docs were merged back with the docstrings directly in the source files and the API docs are directly generated from these source files using the autodoc directives. This makes the docstrings the **single source of truth** for the API docs so we don't have to maintain two versions. This is well aligned with how other projects handle this (shout out to [trio](http://trio.readthedocs.io))

4. It adds minimalist API docs for Trinity (really only what `--help` outputs but better than nothing)

5. The contribution guide now includes guidance regarding type hints and documentation. Please review carefully if there's anything in there you disagree or would like to discuss further.

6. A code of conduct was added

7. API docs on `Chain` were added (there's tons more API docs left to cover! Just getting started!)

In the spirit of a-gif-says-more-than-thousand-words:

doc_love_1

Source code generated API docs (with working links to other types)

doc_love_2

#### Cute Animal Picture


docs/api/evm/api.opcode.rst
```diff
@@ -23,7 +23,7 @@ Here are some simple examples.
 
 
 The `:func:~evm.vm.opcode.as_opcode` helper
```
cburgdorf: @pipermerriam is it correct that all the `api.*.rst` files have been hand written so far? I think our goal would be to extract the API docs directly from the source code to not maintain them redundant or is there something else I'm missing?

pipermerriam: I'm fine with you defining the right path forward on this.  Historically they are probably a mix of generated and hand written.

carver: Yup, handwritten guides plus auto-generated API-level docs sounds good to me. :+1:

cburgdorf: Alright. I took a deeper look into how all this sphinx stuff works and found out about this nice `autodoc` featre where you can write directives such as `automodule`, `autoclass` etc. to *selectively* pull things out of the doc comments that are spread across the code base.

That's actually a super nice mechanism as it gives one much freedom to surround the bare generated content with some additional text or move things around to have fine grained control about how things fit into a toc etc.

It's also what trio [uses heavily](https://raw.githubusercontent.com/python-trio/trio/master/docs/source/reference-io.rst) for their documentation.

You probably know all this already but I felt a bit like Jake from Avatar where he suddenly finds himself surrounded by all these glowing stuff fying around him.

I think I'll be able to finish up a PR this week that will already improve a few for the docs :)

docs/introduction.rst
```diff
@@ -0,0 +1,40 @@
...
+* Is highly flexible to support:
+
+  * Public chains
+  * Private chains
+  * Consortium chains
+  * Advanced research
+
+.. note::
+
+  This project is still in an alpha state and not yet meant to be used by regular consumers to connect to the main Ethereum network. We are quickly progressing forward and shoot for an MVP in a matter of weeks rather than months.
```
cburgdorf: @pipermerriam please review if this aligns with what you think the current state and timeline is.

cburgdorf:
@pipermerriam this one job needs a manual retry (should be fixed once #626 lands)

docs/api/evm/index.rst
```diff
@@ -0,0 +1,14 @@
...
+
+This section aims to provide a detailed description of all APIs. If you are looking for something more hands-on or higher-level check out the existing `EVM guides </guides/evm>`_.
+
+.. toctree::
+   :maxdepth: 2
+   :name: toc-api-evm
+   :caption: API
+
+   api.chain
+   api.computation
```
pipermerriam: I think these are currently ordered alphabetically.  **if** hierarchical ordering makes sense it would be `chain -> vm -> state -> computation` but I'm not sure what the best practice here is.

docs/api/trinity/index.rst
```diff
@@ -0,0 +1,53 @@
+Trinity
+=======
+
+This section aims to provide a detailed description of all APIs. If you are looking for something more hands-on or higher-level check out the existing `Trinity guides </guides/trinity>`_..
```
pipermerriam: Probably worth adding a `.. warning::` to this page indicating that the trinity API is actively being developed and likely to change regularly.

docs/api/evm/index.rst
carver: IMO, guides should have hierarchical ordering, and API docs should be alphabetical.

docs/api/trinity/index.rst
carver: We can be pretty aggressive here to drive the point home: "We expect each alpha release to have breaking changes to the API."

docs/contributing.rst
```diff
@@ -47,6 +52,67 @@ We can also install ``tox`` to run the full test suite which also covers things
...
+    def __init__(self, wrapped_db: BaseDB) -> None:
+        self.wrapped_db = wrapped_db
+        self.reset()
+
+Documentation
+~~~~~~~~~~~~~
+
+Public APIs are expected to be annotated with docstrings as seen in the following example.
+
+.. code:: python
```
carver: Cool, if this highlights everything correctly!

I thought it had to be `.. code-block:: python` based on my reading of sphinx docs: http://www.sphinx-doc.org/en/stable/markup/code.html

docs/contributing.rst
```diff
@@ -1,5 +1,10 @@
-Contributing to py-evm
-----------------------
+Contributing
+------------
+
+Thank you for your interest in contributing! We welcome all contributions no matter their size. Please read along to learn how to get started. If you get stuck, feel free to reach for help in our `Gitter channel <https://gitter.im/ethereum/py-evm>`_.
+
+Setting the stage
+~~~~~~~~~~~~~~~~~
```
carver: Maybe a link to https://github.com/pipermerriam/ethereum-dev-tactical-manual/ is appropriate in this section?

docs/introduction.rst
carver: I'm a little pessimistic about weeks vs months, but... that's just my nature. :) I think we can be optimistic here, with little ramifications.

docs/api/evm/index.rst
cburgdorf: >IMO, guides should have hierarchical ordering, and API docs should be alphabetical.

I think that makes perfect sense :+1:

docs/contributing.rst
carver: Sure, I agree. Some pages are more relevant than others to external contributors. This part seems the most relevant to them: https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md

cburgdorf: Thanks for bringing up the style guide. In fact, I re-read it again carefully and noticed a couple of things I did wrong on past PRs :+1: 

I'll make sure to link to it.

cburgdorf:
Thank you both for your review! Did minor adjustments according to your feedback and landed it :tada:

cburgdorf:
Aaand it's live http://py-evm.readthedocs.io

cburgdorf:
Oh snap! The generated API docs don't work on the live site. There's probably some configuration I'm missing because trio uses the same mechanism with readthedocs. Investigating.

cburgdorf:
Fix is out. It's small enough that I'm *tempted* to merge to just fix the live site will you are still in your well deserved sleep but I rather wait patiently ;)

https://github.com/ethereum/py-evm/pull/631

PR #589, opened 2018-04-24
Fix broken ReadTheDocs build

cburgdorf:
### What was wrong?

ReadTheDocs can't build the docs anymore hence the last online version dates 4 months back.

### How was it fixed?

Add `.readthedocs.yml` with the right settings (hopefully).
Ideally I'd test this on a separate branch directly here in the repo but I don't have write access yet so I hope these settings should just cut it.

#### ~~Cute~~ Majestic Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ deps = {
         "flake8==3.5.0",
         "mypy>=0.580",
     ],
-    'doc': [
+    'docs': [
         "py-evm>=0.2.0-alpha.14",
         "pytest~=3.2",
         "Sphinx>=1.5.5,<2.0.0",
@@ -72,7 +72,7 @@ deps['dev'] = (
     deps['p2p'] +
     deps['trinity'] +
     deps['test'] +
-    deps['doc']
+    deps['docs']
 )
 
 # As long as evm, p2p and trinity are managed together in the py-evm
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,6 @@
+# https://docs.readthedocs.io/en/latest/yaml-config.html
+python:
+  version: 3.5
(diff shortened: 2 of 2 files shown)


cburgdorf:
@pipermerriam just letting you know I'm gonna use my new super powers (write access :tada: ) to push a branch `doctest` into this repo to test out if this PR would resolve the failing RTD builds so we don't have to merge in advance.

I'll take care to remove the branch afterwards again.

cburgdorf:
Alright, this is now tested and it works. Already removed the `doctest` branch again.

setup.py
```diff
@@ -72,7 +72,7 @@
     deps['p2p'] +
     deps['trinity'] +
     deps['test'] +
-    deps['doc']
```
carver: I don't totally disagree that `docs` is more natural than `doc`, but the other extras are all singular (like `test`). More importantly, all the other python ethereum projects that follow this extras pattern use `doc`. Inconsistency between these projects adds unnecessary friction. If you think it's important to change the name to `docs`, then let's go have a cleanup party on all the other repos to change them. If it's not important to you, can we go back to `doc`?

pipermerriam: +1 for consistency, don't care which one we use.

cburgdorf: True. Reasoning was that `docs` feels more natural and seems to be used more widespred (e.g. [ReadTheDocs](https://docs.readthedocs.io/en/latest/yaml-config.html)) and when we had the deps in an `txt` file it also was `requirements-docs.txt` as it is still for e.g [eth-account(]https://github.com/ethereum/eth-account/blob/master/requirements-docs.txt)

But I agree that is inconsistent in regard to the points you mention. Let me switch it back.

PR #590, opened 2018-04-24
p2p: Cleanly break out of read loop when remote disconnects

gsalgado:
Peer used to simply close its network transport upon receiving a
Disconnect msg, which causes us to exit the read loop because our
next readexactly() call will then raise an IncompleteReadError, but
then it looks like the remote simply sent an EOF, which is not actually
the case.

Now when we receive a Disconnect msg we raise an exception, which gets
caught in Peer.read_loop(), causing it to return and the peer to finish
cleanly.

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -52,6 +52,7 @@ from p2p.exceptions import (
     NoMatchingPeerCapabilities,
     OperationCancelled,
     PeerConnectionLost,
+    RemoteDisconnected,
     UnexpectedMessage,
     UnknownProtocolCommand,
     UnreachablePeer,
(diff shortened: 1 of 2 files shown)


p2p/peer.py
```diff
@@ -325,9 +330,7 @@ def handle_p2p_msg(self, cmd: protocol.Command, msg: protocol._DecodedMsgType) -
         """Handle the base protocol (P2P) messages."""
         if isinstance(cmd, Disconnect):
             msg = cast(Dict[str, Any], msg)
-            self.logger.debug(
-                "%s disconnected; reason given: %s", self, msg['reason_name'])
-            self.close()
```
pipermerriam: So this now happens implicitly?

gsalgado: Oops, no, we should still `close()` before returning from `read_loop()`, although that's not a big deal as once the read_loop() returns the peer's `run()` method returns as well and everything (including the transport) should get garbage collected.

Issue #591, opened 2018-04-24
Add common base exception for all `p2p.exception` classes.


pipermerriam:
### What is wrong?

All of the exceptions in `p2p.exceptions` inherit from the base `Exception` class.  Our convention in pyevm has been to have a single base exception that all library exceptions inherit from.  This makes it easy to catch *any* library exception by just catching this one base exception class.

### How can it be fixed

Add a new `BaseP2PError` class which all of the other exception classes from `p2p.exception` inherit from.

PR #602, opened 2018-04-26
Add common base exception to all p2p exceptions

cburgdorf:
### What was wrong?

As described in #591 the exceptions in p2p did not have a common project base exception yet.

### How was it fixed?

Added base exceptions and also added docstrings.

#### Cute Animal Picture

### DIFF
--- a/p2p/exceptions.py
+++ b/p2p/exceptions.py
@@ -1,54 +1,96 @@
-class AuthenticationError(Exception):
+class BaseP2PError(Exception):
+    """
+    The base class for all p2p errors.
+    """
     pass
 
 
-class DecryptionError(Exception):
+class AuthenticationError(BaseP2PError):
+    """
+    Raised when a message could not be decrypted.
+    """
     pass
 
 
-class PeerConnectionLost(Exception):
+class DecryptionError(BaseP2PError):
+    """
+    Raised when a message could not be decrypted.
+    """
     pass
 
 
-class HandshakeFailure(Exception):
+class PeerConnectionLost(BaseP2PError):
+    """
+    Raised when the connection to a peer was lost.
+    """
     pass
 
 
-class UnknownProtocolCommand(Exception):
+class HandshakeFailure(BaseP2PError):
+    """
+    Raised when the protocol handshake was unsuccessful.
+    """
     pass
 
 
-class UnexpectedMessage(Exception):
+class UnknownProtocolCommand(BaseP2PError):
+    """
+    Raised when the received protocal command isn't known.
+    """
     pass
 
 
-class UnreachablePeer(Exception):
+class UnexpectedMessage(BaseP2PError):
+    """
+    Raised when the received message was unexpected.
+    """
     pass
 
 
-class EmptyGetBlockHeadersReply(Exception):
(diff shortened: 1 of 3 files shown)


p2p/exceptions.py
```diff
@@ -1,54 +1,95 @@
-class AuthenticationError(Exception):
+class BaseP2PError(Exception):
+    """
+    Base error class for all p2p errors.
+    """
     pass
 
-
-class DecryptionError(Exception):
+class AuthenticationError(BaseP2PError):
```
cburgdorf: Judging from the callsites it seemed to me as if this and `DecryptionError` could get merged into one. Thoughts?

p2p/exceptions.py
```diff
@@ -1,54 +1,95 @@
...
 
-class TooManyTimeouts(Exception):
+class LESAnnouncementProcessingError(BaseP2PError):
+    """
+    Error signaling that an LESAnnouncement could not be processed.
+    """
     pass
 
 
-class PeerFinished(Exception):
```
cburgdorf: That one wasn't used anywhere

p2p/exceptions.py
gsalgado: Seems like it

cburgdorf: Alright, will update the PR.

p2p/exceptions.py
```diff
@@ -1,54 +1,89 @@
...
+    """
+    Base error class for all p2p errors.
+    """
     pass
 
 
-class DecryptionError(Exception):
+class DecryptionError(BaseP2PError):
+    """
+    Error signaling that a message could not be decrypted.
```
gsalgado: I'd get rid of the "Error signaling that" part from all docstrings here.

p2p/peer.py
```diff
@@ -396,7 +396,7 @@ def decrypt_header(self, data: bytes) -> bytes:
         self.ingress_mac.update(sxor(aes, header_ciphertext))
         expected_header_mac = self.ingress_mac.digest()[:HEADER_LEN]
         if not bytes_eq(expected_header_mac, header_mac):
-            raise AuthenticationError('Invalid header mac')
+            raise DecryptionError('Invalid header mac')
```
pipermerriam: Curious why the exception class here was changed?

cburgdorf: https://github.com/ethereum/py-evm/pull/602#discussion_r184365395

Both seemed to mean the same thing so I got rid of one.

cburgdorf:
In regard to [the discussion](https://github.com/ethereum/py-evm/pull/602#discussion_r184398587) around the docstrings, I looked [how the standard lib handles this](https://docs.python.org/3/library/exceptions.html) and changed all docstrings to be aligned with that.

So, the pythonic way seems to be:

1. Base exceptions start with

>The base class for...

2. Other exceptions start with

>Raised when...

pipermerriam:
👍 I love borrowing conventions from elsewhere.

cburgdorf:
@pipermerriam This failed CI run seems like...not legit. Is that a bug in the Circle setup or just a hiccup and we need to retrigger (I can't)

Issue #606, opened 2018-04-27
New JournalDB doesn't seem to persist everything when importing a block


gsalgado:
With the new `JournalDB`, when I import ropsten block 3072902 (via `Chain.import_block()`), the block's state_root doesn't end up in the underlying DB, although it is present in the journal. That doesn't happen when I use the old `JournalDB`

gsalgado:
Here's something else that seems suspicious: https://github.com/gsalgado/py-evm/blob/journal-messup/evm/tools/fixture_tests.py#L633-L635

cburgdorf:
Oops. So, from the top of my head this is what I think:

- either this means there is a code path that should not use through the `JournalDB` but rather the underlying db directly or, 

- we need a another `persist()` call at some later point to also persist this missed data

- we need to move the existing `persist()` call to some later point in time to include all changes. Currently at is called at the end of every single transaction processing

cburgdorf:
Would it possible to add a test or change an existing test to fail that would have succeeded before? Then I could sprinkle a couple of conditional breakpoints across the code base to identify the part that needs to be fixed.

gsalgado:
https://github.com/ethereum/py-evm/issues/606#issuecomment-384948658

cburgdorf:
Oh, I meant a lookup of a key in the MemoryDB that should exist but doesn't. But yeah, the pending changes there are most likely unintended. But we need to identify the place where they should have been persisted. So far I *assumed* we only hit the `JournalDB` for the transaction processing, hence calling `persist()` at the end of every transaction processing should cut it but this bug seems to indicate there are other changes that either need to be offloaded from the `JournalDB` or need another or moved `persist()` call.

gsalgado:
In that test, any keys from `chain.chaindb.journal_db.journal.latest.keys()` should exist in `chain.chaindb.db`, no?

cburgdorf:
Maybe. But at least I could verify that but looking up one of the keys there and then running that test again against the previous JournalDB and see if it ends up in the MemoryDB. Then I can set a conditional breakpoint to figure out at which code path it would have been written. Unfortunately I'm in a train though and have to head out for my stop.

gsalgado:
So, I've been thinking about the new `JournalDB` and how some `ChainDB` methods use it whereas others use the underlying db directly, and I believe this is not ideal and will probably lead to more issues like this one. How do you guys feel about that? Is there a plan to improve that in the short term?

[...]

PR #607, opened 2018-04-27
Ensure important journal data is persisted

cburgdorf:
### What was wrong?

https://github.com/ethereum/py-evm/commit/f6ffc5e5f78557133f348074fb17be2f2bcc383d introduced a regression in which important data wasn't persisted anymore. (See #606)

### How was it fixed?

This is a hacky way to resolve the issue in a short term manner by sprinkling two more `persist()` calls over the code paths that use the journal while we investigate a more robust way forward.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -470,6 +470,7 @@ class Chain(BaseChain):
         genesis_header = BlockHeader(**genesis_params)
         genesis_chain = cls(chaindb, genesis_header)
         chaindb.persist_block(genesis_chain.get_block())
+        chaindb.persist()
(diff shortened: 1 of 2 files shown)


cburgdorf:
Let's wait a bit with this. I may have a less hacky solution brewing here: https://github.com/ethereum/py-evm/pull/610

cburgdorf:
I'm gonna close this for now as I personally would like to promote #610 (ready for review) as a better step into the right direction. If something with #610 turns out to be wrong I'll do the requested changes on this PR and reopen.

PR #608, opened 2018-04-27
Get rid of APIs related to pending transactions

gsalgado:
When working on something else I came across those APIs that were only used in a couple tests

Removed them!

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -162,20 +162,6 @@ class BaseChain(Configurable, metaclass=ABCMeta):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def add_pending_transaction(self, transaction):
-        """
-        Adds a transaction to the set of pending transactions.
-        """
-        raise NotImplementedError("Chain classes must implement this method")
-
-    @abstractmethod
-    def get_pending_transaction(self, transaction_hash):
-        """
-        Retrieves a transaction from the set of pending transactions.
-        """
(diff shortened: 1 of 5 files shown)


carver:
> A pending transaction was one that is in a block which has not been mined/finalized yes?

Partially: in some cases they might not be in any block at all. Two kinds of relevant pending transactions that follow slightly different rules:

- _peer-received_: This populates the txpool. Can be dropped across restarts, or by failing arbitrary (trinity-chosen) tests about gas price, etc.
- _locally-issued_: A user requested to broadcast this transaction (eg~ via `eth_sendRawTransaction'). We likely want to persist these across trinity restarts, and want to keep these as pending regardless of the txpool garbage collection rules related to gas price, etc.

> I think if so then I support removing these specialized getters assuming we can do what we need without them.

Because they might not be in any block, I think we need to keep them.

> Maybe they were for looking up transactions by hash when they had not yet been included in a canonical block?

Right, it's valid to make a json-rpc call requesting the body of a pending transaction. Even if we punt for a while on doing this from the transaction pool, doing this for locally issued transactions early seems important.

gsalgado:
Right, I understand this might be needed at some point, but right now there was nothing using `add_pending_transaction()` to store those transactions, and I believe it's better to re-add that whenever we need it than to maintain unused code. Not a big deal, though, so feel free to either merge or close this

pipermerriam:
I believe we'll deal with all of the use cases outlined by @carver with trinity specific code.

Issue #599, opened 2018-04-25
Thoughts on API refactors for Chain, VM, and State


pipermerriam:
# Architecture Refactor for Py-EVM

## What is wrong?

The APIs on `Chain`, `VM` and `State` have all evolved organically.  Now that
we have a better picture of how they are used, we should be able to make some
improvements to make things more intuitive, less stateful, and with reduced
mutation.

#### Blocks

- Finalized:
    - A block which has been *sealed*.  
        - In proof of work this means setting the `mix_hash` and `nonce`
- HEAD: 
    - The highest finalized block in the chain
- TIP: 
    - The unsealed block who's parent is the HEAD block

### `Chain`

The `Chain` is now the only stateful object with non-pure APIs

- rename `Chain.header` to `Chain.tip`.
- add `Chain.head` for access to the canonical chain head.

#### `Chain.apply_transaction(transaction) -> computation`

Adds the transaction to the current *TIP* block.

- error if `chain.tip` is empty.
- executes the `transaction` on top of the *TIP* block.
- mutates `chain.header` with the post execution state.
- persists the `transaction` and resulting `receipt`.

#### `Chain.import_block(*args, **kwargs) -> block`

- error if `chain.tip` is empty.
- delegates to `VM` to do block import.
- validates that the imported block is equivalent to the import result.
- persists newly imported block to the database
- mutates `chain.header` with new *tip block* header.
- leaves `chain.tip` empty.

#### `Chain.finalize_block(*args, **kwargs) -> block`

- error if `chain.tip` is empty.
- delegates to `VM.mine_block`
- persists newly mined block to the database
- mutates `chain.header` with new *tip block* header.
- leaves `chain.tip` empty.

#### `Chain.initialize_tip(parent=None, **context)

- error if `chain.tip` is not empty.
- initializes `chain.tip` with a new fresh header on top of `chain.head`

### `VM`

The `VM` object is now immutable, only exposing *pure* functions.

- remove `configure_header` in favor of:
    - new `initialize_block` which initializes the vm with a new *empty* header with some configured params like `coinbase`, `difficulty`, etc.

#### `VM.execute_transaction(transaction, block) -> state_root, receipt, computation`

- **pure**
- executes the transaction on top of the `block`
- returns:
    - the updated state root
    - the receipt for the transaction
    - the computation

#### `VM.import_block(block) -> block`

- **pure**
[...]

PR #611, opened 2018-04-27
WIP: Refactory how account db is used.

pipermerriam:
link: #599 

### What was wrong?

This is an attempt at an incremental first step towards #599.  It also ended up doing some bits of what I think relate to #507 

- During the initial implementaiton of the sharding code, support for multiple trie types was added.  Changes to the sharding roadmap removed the need for supporting multiple tries.
- The name `AccountStateDB` is more concisely named `AccountDB`
- The name `account_db` is more appropriate than `state_db`
- The `State` class is a more appropriate home for the the `account_db`
- The context manager APIs aren't working out in my opinion and are low value add.

### How was it fixed?

This is a work in progress.  I'm trying to simplify the API through which we access and modify the account state as well as figure out a better mechanism for updating the `state_root`.

- TODO

#### Cute Animal Picture


carver:
> first step towards #599. It also ended up doing some bits of what I think relate to #507

_called it_ :P

PR #617, opened 2018-04-30
Piper/remove multi trie support

pipermerriam:
link: #599 and #615 and #611 

### What was wrong?

During the sharding work, support for different types of underlying `trie` classes was introduced as part of the work to support lighter witnesses.

### How was it fixed?

Dropped the support since this is no longer part of the immediate roadmap.

#### Cute Animal Picture

cute-goats-and-camel-pals

### DIFF
--- a/evm/db/chain.py
+++ b/evm/db/chain.py
@@ -12,13 +12,11 @@ from typing import (
     Tuple,
     Type,
     TYPE_CHECKING,
-    Union
 )
 
 import rlp
 
 from trie import (
-    BinaryTrie,
     HexaryTrie,
 )
 
@@ -31,8 +29,6 @@ from eth_hash.auto import keccak
 
 from evm.constants import (
     GENESIS_PARENT_HASH,
-    BLANK_ROOT_HASH,
-    EMPTY_SHA3,
 )
 from evm.exceptions import (
     BlockNotFound,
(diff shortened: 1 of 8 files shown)


evm/db/chain.py
```diff
@@ -88,36 +84,14 @@ class TransactionKey(rlp.Serializable):
...
-        if trie_class is HexaryTrie:
-            empty_root_hash = BLANK_ROOT_HASH
-        elif trie_class is BinaryTrie:
-            empty_root_hash = EMPTY_SHA3
-        else:
-            raise NotImplementedError(
-                "trie_class {} is not supported.".format(trie_class)
-            )
-        self.trie_class = trie_class
-        self.empty_root_hash = empty_root_hash
```
carver: The root hash code feels like unnecessary inspection on the trie class details, so I'm happy to see that go. Maybe we can push the blank root hash responsibility into the trie class. IIRC, the class already keeps track of what a blank hash is anyway.

Anyway, I guess that's only relevant when we bring back configurable trie classes.

pipermerriam: I think this stuff is implemented in the wrong place so....

> the cost of keeping the configurable tries around might be higher than the cost of adding them back in later

I think this is accurate.

cburgdorf:
>Hm, this feels a bit wrong, since it seems likely that we will want configurable trie classes again later.

Judging from the diff it seems easy enough to readd *if* it's needed at some later point. I'm definitely a fan of removing things when they aren't needed and keeping things as simple as possible.

PR #618, opened 2018-04-30
Remove apply_state_dict API

pipermerriam:
### What was wrong?

The `AccoutDB.apply_state_dict` API was just a convenience API used mostly in testing.  The only place it was used outside of testing was during setup of genesis state.

### How was it fixed?

Removed it from the `AccountDB` in favor of a stand-alone utility function.

### DIFF
--- a/evm/utils/db.py
+++ b/evm/utils/db.py
@@ -58,3 +58,15 @@ def get_empty_root_hash(db: 'BaseChainDB') -> bytes:
             "db.trie_class {} is not supported.".format(db.trie_class)
         )
     return root_hash
+
+
+def apply_state_dict(account_db, state_dict):
+    for account, account_data in state_dict.items():
+        account_db.set_balance(account, account_data["balance"])
+        account_db.set_nonce(account, account_data["nonce"])
+        account_db.set_code(account, account_data["code"])
+
+        for slot, value in account_data["storage"].items():
+            account_db.set_storage(account, slot, value)
+
+    return account_db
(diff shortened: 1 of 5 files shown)


evm/utils/db.py
```diff
@@ -58,3 +58,15 @@ def get_empty_root_hash(db: 'BaseChainDB') -> bytes:
...
+
+def apply_state_dict(account_db, state_dict):
+    for account, account_data in state_dict.items():
+        account_db.set_balance(account, account_data["balance"])
+        account_db.set_nonce(account, account_data["nonce"])
+        account_db.set_code(account, account_data["code"])
+
+        for slot, value in account_data["storage"].items():
+            account_db.set_storage(account, slot, value)
+
```
cburgdorf: we *could* sneak in a `account_db.persist()` because the way it stands now it really relies on some other code path calling `account_db.persist()` at some point. That works today, because we call `persist()` as part of the block finalization but it's a very fragile thing to rely on.

That said, I think we all agreed we eventually want to push the `JournalDB` further down to be only used for the transaction processing so that **this code** wouldn't end up using the `JournalDB` underneath at all. We may just accept the fragility for now. Just wanted to raise awareness again.

pipermerriam: Hrm, good catch.  I'll add a persist call.

pipermerriam: Actually, with everything passing and the uncertainty around these APIs I'm going to leave as-is and try and address this as part of the broader #507

PR #620, opened 2018-04-30
Piper/rename account db.root hash to account db.state root

pipermerriam:
link: #611 #507 #599 

### What was wrong?

The `AccountDB` class used the property `AccountDB.root_hash`.  This value represents the `Header.state_root` and is more appropriately named `state_root`.

### How was it fixed?

Renamed it.

#### Cute Animal Picture

mama_moose_licking_calf-2974-6


pipermerriam:
@jannikluhn clicked your name on accident but you're welcome to review 😄

PR #625, opened 2018-05-02
Return header instead of block from vm.apply_txn

carver:
### What was wrong?

Related to #599 & #507 - VM returns partially-completed block objects as a result of `apply_transaction()`

### How was it fixed?

Return the header instead, and let `chain` deal with building the block (at the appropriate time). As a bonus, we build fewer block objects during `vm.import_block()`.

*TODO in follow-up PR: decide how to deal with the exact some problem with `header` which is only partially formed (no state/receipt root) -- maybe this is where a `TransactionDiff` comes in`*

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -487,20 +487,29 @@ class Chain(BaseChain):
         heavy and incurs significant perferomance overhead.
         """
         vm = self.get_vm()
-        block, receipt, computation = vm.apply_transaction(transaction)
-        receipts = block.get_receipts(self.chaindb) + [receipt]
+        old_block = vm.block
 
-        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
-        receipt_root_hash, receipt_kv_nodes = make_trie_root_and_nodes(receipts)
+        new_header, receipt, computation = vm.apply_transaction(transaction)
+
+        transactions = old_block.transactions + (transaction, )
+        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(transactions)
         self.chaindb.persist_trie_data_dict(tx_kv_nodes)
+
+        receipts = old_block.get_receipts(self.chaindb) + (receipt, )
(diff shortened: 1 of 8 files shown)


evm/chains/base.py
```diff
@@ -487,20 +487,29 @@ def apply_transaction(self, transaction):
         heavy and incurs significant perferomance overhead.
         """
         vm = self.get_vm()
-        block, receipt, computation = vm.apply_transaction(transaction)
-        receipts = block.get_receipts(self.chaindb) + [receipt]
+        old_block = vm.block
```
pipermerriam: - `base_block`
- `previous_block`

If you like either better.

evm/vm/base.py
```diff
@@ -178,30 +178,32 @@ def import_block(self, block):
...
+            header_with_txns = self.block.header
 
-        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
-        receipt_root_hash, receipt_kv_nodes = make_trie_root_and_nodes(receipts)
 
+        self.block = self.seal_block(self.block, header_with_txns, block.transactions, receipts)
+
+        return self.mine_block()
+
+    def seal_block(self, base_block, new_header, transactions, receipts):
```
pipermerriam: I don't like the term `seal` as for me at least, it implies mining like setting the `mix_hash` and `nonce`.

What about just `add_transaction_to_block` or something?

carver: How about `set_transactions_in_block`?
- pro: transactions is plural, old transactions in block are replaced, and -- for now -- the trie data is persisted
- con: implies block mutation, when it's really building a block copy

Other alternatives:
- `place_transactions_in_block`
- `commit_transactions_to_block`
- `get_block_with_transactions` (problem: get implies no side-effects)
- `add_transactions_to_block` (problem: add implies that previous transactions are left in place)

pipermerriam: - `set_block_transactions`  (keeps with `get_`, `set_` convention...)
- `update_block_transactions`  just an alternate prefix.

any of of those work for you?

PR #626, opened 2018-05-02
Retry flaky CI job up to three times

cburgdorf:
### What was wrong?

The CI job that fetches the PRs base branch via the GitHub API sometimes fails but works after retry.

image

### How was it fixed?

Extracted the command into a standalone script and configured two more commands that only run if the previous command failed.

This may not be the most elegant way to solve this but my Google fu didn't yield anything else and it seems to get the job done.

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,20 +12,15 @@ common: &common
         command: git submodule update --init --recursive
     - run:
         name: merge pull request base
-        command: |
-          if [[ -n "${CIRCLE_PR_NUMBER}" ]]; then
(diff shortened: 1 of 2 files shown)


cburgdorf:
Ah, actually `curl` has a `--retry` build in. That's probably what @carver meant what we should use. Let me give that a try.

cburgdorf:
Ah no, we can't simply do a `--retry` on curl because *technically* it's not the `curl` that fails. The GitHub API returns just fine, it's just that the API doesn't know that PR yet (a simple race condition) so our parsing logic fails.

An alternative would be to build the retry logic right into the shell script but maybe not worth the effort.

PR #630, opened 2018-05-03
Delete gas used from state

carver:
### What was wrong?

Related to #599 & #507 - there was state mutation inside `VM.apply_transaction()`, to update `gas_used`, and `get_block_by_header` lived in VM (we're generally looking to move block manipulation out of VM, into Chain).

### How was it fixed?

There wasn't really any need for `gas_used` to be in the state, we can verify the transaction against the header in the VM, before even calling `state.apply_transaction`.

- add a new transaction verification, explicitly against the header
- make_receipt new uses the header to generate the cumulative gas used
- apply_transaction now builds on top of a passed-in header (rather than using a local mutation of `self.block.header`)
- `get_block_by_header` really needed to belong in Chain instead of VM.
- use only `account_db` instead of the full state to validate transaction (after separately running the transaction->header validation)

Yay for more use of (immutable) rlp object, and less state mutation.

#### Cute Animal Picture

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -75,7 +75,6 @@ class BaseVM(Configurable, metaclass=ABCMeta):
             db=self.chaindb.db,
             execution_context=self.block.header.create_execution_context(self.previous_hashes),
             state_root=self.block.header.state_root,
-            gas_used=self.block.header.gas_used,
         )
 
     #
@@ -131,29 +130,63 @@ class BaseVM(Configurable, metaclass=ABCMeta):
         )
 
     @abstractmethod
-    def make_receipt(self, transaction, computation, state):
+    def make_receipt(self, base_header, transaction, computation, state):
         """
-        Make receipt.
+        Generate the receipt resulting from applying the transaction.
+
+        :param base_header: the header of the block before the transaction was applied.
+        :param transaction: the transaction used to generate the receipt
+        :param computation: the result of running the transaction computation
+        :param state: the resulting state, after executing the computation
+
+        :return: receipt
         """
         raise NotImplementedError("Must be implemented by subclasses")
 
-    def apply_transaction(self, transaction):
+    @abstractmethod
+    def validate_transaction_against_header(self, base_header, transaction):
+        """
+        Validate that the given transaction is valid to apply to the given header.
+
+        :param base_header: header before applying the transaction
+        :param transaction: the transaction to validate
+
+        :raises: ValidationError if the transaction is not valid to apply
+        """
+        raise NotImplementedError("Must be implemented by subclasses")
+
+    def apply_transaction(self, header, transaction):
         """
         Apply the transaction to the current block. This is a wrapper around
         :func:`~evm.vm.state.State.apply_transaction` with some extra orchestration logic.
-        """
-        state_root, computation = self.state.apply_transaction(transaction)
-        receipt = self.make_receipt(transaction, computation, self.state)
-        # TODO: remove this mutation.
-        self.state.gas_used = receipt.gas_used
 
-        new_header = self.block.header.copy(
-            bloom=int(BloomFilter(self.block.header.bloom) | receipt.bloom),
+        :param header: header of the block before application
+        :param transaction: to apply
+        """
+        self.validate_transaction_against_header(header, transaction)
+        state_root, computation = self.state.apply_transaction(transaction)
+        receipt = self.make_receipt(header, transaction, computation, self.state)
+
+        new_header = header.copy(
+            bloom=int(BloomFilter(header.bloom) | receipt.bloom),
             gas_used=receipt.gas_used,
             state_root=state_root,
         )
 
         return new_header, receipt, computation
+
+    def _apply_all_transactions(self, transactions, base_header):
+        receipts = []
+        previous_header = base_header
+        result_header = base_header
+
+        for transaction in transactions:
+            result_header, receipt, _ = self.apply_transaction(previous_header, transaction)
+
+            previous_header = result_header
+            receipts.append(receipt)
+
+        return result_header, receipts
 
     #
     # Mining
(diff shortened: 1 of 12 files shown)


evm/vm/forks/byzantium/__init__.py
```diff
@@ -16,20 +13,15 @@
...
-    old_receipt = make_frontier_receipt(transaction, computation, state)
+def make_byzantium_receipt(base_header, transaction, computation, state):
+    frontier_receipt = make_frontier_receipt(base_header, transaction, computation, state)
 
     if computation.is_error:
-        state_root = EIP658_TRANSACTION_STATUS_CODE_FAILURE
+        success_code = EIP658_TRANSACTION_STATUS_CODE_FAILURE
     else:
-        state_root = EIP658_TRANSACTION_STATUS_CODE_SUCCESS
+        success_code = EIP658_TRANSACTION_STATUS_CODE_SUCCESS
```
carver: I'm wrapping for the day, but on another review, this should probably be `status_code` instead of `success_code`

evm/vm/base.py
```diff
@@ -123,29 +122,63 @@ def execute_bytecode(self,
...
+        previous_header = base_header
+        result_header = base_header
+
+        for transaction in transactions:
+            result_header, receipt, _ = self.apply_transaction(previous_header, transaction)
+
+            previous_header = result_header
+            receipts.append(receipt)
+
+        return result_header, receipts
```
carver: It was nice to pull this chunk of logic out, but it was awkward to both:

- chain the result of one application into the next, and
- accumulate the results

A small version involved a lot of variable clobbering, like:

```py
    def _apply_all_transactions(self, transactions, header):
        receipts = []

        for transaction in transactions:
            header, receipt, _ = self.apply_transaction(header, transaction)
            receipts.append(receipt)

        return header, receipts
```

I'm happy to hear suggestions about a way to do chaining & accumulation at the same time, in a more elegant way.

evm/vm/base.py
```diff
@@ -123,29 +122,63 @@ def execute_bytecode(self,
...
         """
         raise NotImplementedError("Must be implemented by subclasses")
 
-    def apply_transaction(self, transaction):
+    def apply_transaction(self, header, transaction):
         """
         Apply the transaction to the vm in the current block.
+
+        :param transaction: to apply
+        :param header: header of the block before application
```
carver: Oops, I'll swap the order of these params in the docs. I went back and forth a few times locally. In the end a semi-chronological order made sense to me: the header is in the state before the transaction is applied, so comes first.

evm/vm/base.py
```diff
@@ -300,6 +321,13 @@ def generate_block_from_parent_header_and_coinbase(cls, parent_header, coinbase)
     # Validate
     #
     def validate_block(self, block):
+        if not isinstance(block, self.get_block_class()):
+            raise ValidationError(
+                "This vm ({0!r}) is not equipped to validate a block of type {1!r}".format(
+                    self,
+                    block,
+                )
+            )
```
carver: I guess this check will be moot, when this method moves up to Chain. But we are a couple days from moving this one, I think.

evm/vm/base.py
```diff
@@ -123,29 +122,63 @@ def execute_bytecode(self,
...
         """
         Apply the transaction to the vm in the current block.
+
+        :param transaction: to apply
+        :param header: header of the block before application
         """
+        self.validate_transaction_against_header(header, transaction)
         state_root, computation = self.state.apply_transaction(transaction)
-        receipt = self.make_receipt(transaction, computation, self.state)
-        # TODO: remove this mutation.
```
carver: This was the inspiration for most of the PR.

evm/vm/base.py
```diff
@@ -123,29 +122,63 @@ def execute_bytecode(self,
...
-    def make_receipt(self, transaction, computation, state):
+    def make_receipt(self, base_header, transaction, computation, state):
         """
-        Make receipt.
+        Generate the receipt resulting from applying the transaction.
+
+        :param base_header: the header of the block before the transaction was applied.
+        :param transaction: the transaction used to generate the receipt
+        :param computation: the result of running the transaction computation
+        :param state: the resulting state, after executing the computation
```
cburgdorf: Great to see you're taking the chance to directly improve the documentation :heart:

evm/vm/base.py
```diff
@@ -123,29 +122,63 @@ def execute_bytecode(self,
...
-            bloom=int(BloomFilter(self.block.header.bloom) | receipt.bloom),
+        new_header = header.copy(
+            bloom=int(BloomFilter(header.bloom) | receipt.bloom),
             gas_used=receipt.gas_used,
             state_root=state_root,
         )
 
         return new_header, receipt, computation
 
+    def _apply_all_transactions(self, transactions, base_header):
```
cburgdorf: How do you feel about sneaking in type definitions for new / changed functions? I do realize it reduces conformity for that file in question when only few defs are typed while others are not but on the other hand it helps to work towards our goal to have full coverage and eventually enable stricter type checks to disallow untyped defs altogether.

carver: :balance_scale: I'm a little on the fence. These APIs will continue to change a lot. I guess I'm slightly :-1:  at the moment, but when I feel like things start to settle, I'll be a lot more keen.

evm/vm/base.py
carver: Thanks! It got me thinking of some other ways to look at it.

Right now, to my eyes, that particular implementation is a drop in readability, compared to having some variable reuse in a short function.

I have an idea for an improvement to get rid of it, but I don't want to throw this PR off track, so I'll put it in the next one.

pipermerriam: I think in implementation it will look like this.

```python
# top level helper function.
def _accumulate_apply_txn_fn(prev_result, apply_transaction_fn):
    return apply_transaction_fn(prev_result[0])

def import_block(self, block):
    ...
    if block.transactions:
        apply_transaction_fns = tuple(
            functools.partial(self.apply_transaction, transaction=transaction)
            for transaction in block.transactions
        )
        headers, receipts = zip(*tuple(accumulate(
            _accumulate_apply_txn_fn, 
            apply_transaction_fns,
            (self.block.header, None),
        ))[1:])
        last_header = headers[-1]
    else:
        last_header = self.block.header
        receipts = tuple()
     ...
```

I think I generally agree about the readability, but I'm also inclined to cut the extra API from the `VM` class.  We could still use the logic above to encapulate the act of applying all of the transactions and getting back the resulting header and list of receipts but still keep it encapsulated in a utility function to keep the `import_block` method easier to read.

Dunno, your call.  Just wanted to see it all written out mostly.

PR #631, opened 2018-05-03
Use Python 3.6 to generate docs

cburgdorf:
### What was wrong?

Our live doc site is currently broken not showing any of the API docs as seen here.

image

This PR fixes it as seen on the following screenshot (notice the URL. This is live on a special `doctest` branch that I pushed into this repo but will delete again now)

image

### How was it fixed?

Apparently readthedocs had issues building our API docs with Python 3.5 (which should work in general, trio uses Python 3.5 for rtd as well and has no issues with the autodoc feature). Instructing readthedocs to build with Python 3.6 instead fixes the issue.

#### Cute Animal Picture

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,6 +1,6 @@
 # https://docs.readthedocs.io/en/latest/yaml-config.html
 build:
-    image: latest
+  image: latest
 python:
   version: 3.6
   pip_install: True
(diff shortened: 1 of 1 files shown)


.readthedocs.yml
```diff
@@ -1,6 +1,8 @@
 # https://docs.readthedocs.io/en/latest/yaml-config.html
+build:
+    image: latest
```
cburgdorf: Note to self: indentation is inconsistent

pipermerriam:
Maybe python 3.6 required because of trinity/p2p code being python3.6 only due to type annotations?

cburgdorf:
@pipermerriam nope, the errors where in evm but probably not worth following up.

Issue #568, opened 2018-04-17
Cleanup `evm.utils.numeric`


pipermerriam:
### What is wrong?

The `evm.utils.numeric` module could be cleaned up.

### How can it be fixed

- switch to use `eth_utils.int_to_big_endian` and `eth_utils.big_endian_to_int`
- change `int_to_bytes32` to use `value.to_bytes(32, 'big')` but preserve the validation checks.
- remove `int_to_byte` as it doesn't appear to be used anywhere.
- remove `byte_to_int` as it isn't used.
- remove `safe_ord` as it should not be needed anymore.

hiteshramani:
I'll be taking this up and submitting a PR soon.

hiteshramani:
I was going through the methods to be double sure if they're not used anywhere and came across one use of `safe_ord` in `p2p.discovery` which is as follows:

`cmd_id = safe_ord(message[HEAD_SIZE])`

Wanted to be sure if this is not needed anymore?

gsalgado:
> cmd_id = safe_ord(message[HEAD_SIZE])
> Wanted to be sure if this is not needed anymore?

It shouldn't as message should be of type `bytes`, but in order to get rid of safe_ord() there you'll also need to update other methods there to change the type of arguments from `AnyStr` to `bytes`: https://paste.ubuntu.com/p/B9qBPPQNmw/

hiteshramani:
@pipermerriam 
> - switch to use eth_utils.int_to_big_endian and eth_utils.big_endian_to_int

When you say "switch to use" you mean the entire repo needs to use the above rather than the numeric methods, right?

hiteshramani:
I am creating a PR with the rest of  the changes, I will commit the one left as soon as the doubt is clarified.

pipermerriam:
Thanks for the contribution @hiteshramani

PR #633, opened 2018-05-03
Cleaning up evm.utils.numeric

hiteshramani:
### What was wrong?
The `evm.utils.numeric` required cleanup. Issue -> #568 

### How was it fixed?
I modified and removed certain methods from the module which were not being used. `safe_ord` which was being used was safely modified.

#### Cute Animal Picture


cburgdorf:
Thanks for the PR @hiteshramani ! Can you rebase on top of latest master to get the CI to pass?

hiteshramani:
@cburgdorf I suggest the flake8 linter should be configured on the developer's side, and the README should include how to configure the linter as a pre-commit hook to avoid unnecessary failures due to lint errors. What do you think?

cburgdorf:
>I suggest the flake8 linter should be configured on the developer's side

You should definitely run `flake8` locally to get early feedback but we still need to run it as part of CI as a *last line of defence* so to speak 

>and the README should include how to configure the linter as a pre-commit

I'm *personally* not a friend of pre-commit hooks for this because my workflow does lots of work in progress commit which I cleanup later. I don't want git to yell at me about anything at this stage. Not even about totally broken code.

But you are correct that our README falls short on any guidance regarding code style. We do have a [style guide](https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md) that also mentions `flake8` and we also have a general [contribution guide](http://py-evm.readthedocs.io/en/latest/contributing.html#code-style) that highlights a couple of other things but none of this is mentioned in the README and I agree we should at least have some pointers there.

pipermerriam:
I'd vote for just a dedicated section about how to run the linting tests, maybe as a *checklist* for things that developers should do prior to pushing commits. (though I'm not sure what else would go in this section).

Issue #632, opened 2018-05-03
Use separate database names for light/full sync


cburgdorf:
### What is wrong?

Trinity currently doesn't use different databases for `full` and `light` mode. It should as toggling between these modes will yield unexpected results with a shared database

### How can it be fixed

Add a suffix to the `database-dir` part of the path.

- `~/.local/share/trinity/mainnet/light-chain/`
- `~/.local/share/trinity/mainnet/full-chain/`

This will need to be done in the `trinity.utils.chain.ChainConfig`

gsalgado:
Do we need separate `data_dir`s? I think it'd be preferable to have just one `data_dir` per network_id, with separate DBs depending on the sync mode.

cburgdorf:
Oh yes, makes sense!

cburgdorf:
Updated text to reflect we are just talking about the name of the db, not the data directory.

Issue #580, opened 2018-04-20
Need an async method for importing a block


gsalgado:
Regardless of how much time we spend optimizing the process of importing a block, it will always be an expensive (CPU-heavy) operation, and when importing blocks as part of a chain sync, we cannot run the block import as a synchronous (i.e. blocking) method because the process running that is also responsible for processing network messages (via the asyncio event loop).

I believe that what we want is a Chain-like service running in a separate process, exposed via a `multiprocessing` proxy (like trinity's `ChainDBProxy`). That new service would have to access the database via `ChainDBProxy` as well, though, as we can only have one process accessing our DB. Does this make sense? Can you guys think of any other possible issues?

Issue #622, opened 2018-05-01
Explore alternative for current configuration approach


cburgdorf:
### What is wrong?

The way I understand it, we currently use `Configurable` as a way to dynamically generate classes (or maybe better: class like objects?). We use that all over the place e.g.

```python
FrontierVM = VM.configure(
    # class name
    __name__='FrontierVM',
    # fork name
    fork='frontier',
    # classes
    _state_class=FrontierState,
    # helpers
    create_header_from_parent=staticmethod(create_frontier_header_from_parent),
    compute_difficulty=staticmethod(compute_frontier_difficulty),
    configure_header=configure_frontier_header,
)
```

The way I understand the mechanism, it basically lets you take a class as a blueprint and generate a new one with several properties and sub properties altered.

I'm not entirely sure if all of its functionality could be achieved with regular style inheritance or if it contains things that would not work with classical inheritance. In other words, I'm not sure if it is just syntactic sugar to have a more concise mechanism or if contains stuff that would be hard to achieve the classical way. I'll assume that **it does** have some special power for now.

While there is no big problem with this approach per se I noticed a couple of minor things that I considered worth enough to write down.

- the current mechanism doesn't play nice with `mypy` or types in general. In the above example `FrontierVM` really acts as a class but isn't recognized as such from the type system

- as an extension to the previous point, the type system doesn't know that `FrontierVM` inherits from `VM`

- not having proper `class` syntax makes it harder to document as one can not simple write e.g.

```python

class FrontierVM
    """
    Frontier anyone?
    """
    ...
```

- just from looking at the code it is less obvious that this thing really is meant to be treated as a class

### How can it be fixed

1. If we assume that there is nothing `Configurable` achieves that the classical inheritance style could not achieve than obviously we could simply *consider* using the classical style if we think it would be worthwhile.

2. Alternatively, I guess we could come up with some kind of decorator based approach that might translate the example above into something roughly equivalent to the following.

```python

@configure(
    # fork name
    fork='frontier',
    # classes
    _state_class=FrontierState,
    # helpers
    create_header_from_parent=staticmethod(create_frontier_header_from_parent),
    compute_difficulty=staticmethod(compute_frontier_difficulty),
    configure_header=configure_frontier_header,
)
class FrontierVM(VM):
    """
    Definitely Frontier! 
    """
    pass
```

This isn't a real proposal for something I think we *should do*. I don't know much about the reasoning behind the current mechanism so I'm just putting this out as something to discuss.

[...]

PR #638, opened 2018-05-03
trinity: Add support for full node syncing and run that by default

gsalgado:
Trinity used to support only light sync; now it supports both light and full sync

Also:
* Changed the default from light to full sync, as light is very unreliable since we can't find peers
* Includes fixes for #580 and #632

### DIFF
--- a/tests/p2p/integration_test_helpers.py
+++ b/tests/p2p/integration_test_helpers.py
@@ -4,7 +4,7 @@ from typing import Type
 from eth_utils import decode_hex
 from eth_keys import datatypes, keys
 
-from evm import RopstenChain
+from evm import MainnetChain, RopstenChain
 from evm.db.chain import AsyncChainDB
 
 from p2p import kademlia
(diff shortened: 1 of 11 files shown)


evm/chains/base.py
```diff
@@ -384,6 +384,15 @@ def get_vm(self, header=None):
         vm_class = self.get_vm_class_for_block_number(header.block_number)
         return vm_class(header=header, chaindb=self.chaindb)
 
+    def get_block_class_for_block_number(self, block_number):
+        """
+        Returns the Block class for the given block number.
+        """
+        # This method is needed because our VM classes cannot be pickled, so we can't use
```
pipermerriam: It looks like if we change our VM classes to use standard inheritance they become pickleable.  cc @cburgdorf and link to #622 

```python
class FrontierVM(VM):
    # fork name
    fork='frontier'
    # classes
    _state_class = FrontierState
    # helpers
    create_header_from_parent = staticmethod(create_frontier_header_from_parent)
    compute_difficulty = staticmethod(compute_frontier_difficulty)
    configure_header = configure_frontier_header
    make_receipt = staticmethod(make_frontier_receipt)
```

trinity/utils/chains.py
```diff
@@ -52,14 +56,8 @@ def get_data_dir_for_network_id(network_id: int) -> str:
...
 
-DATABASE_DIR_NAME = 'chain'
-
-
-def get_database_dir(data_dir: str) -> str:
-    """
-    Returns the directory path where chain data will be stored.
-    """
-    return os.path.join(data_dir, DATABASE_DIR_NAME)
+FULL_DATABASE_DIR_NAME = 'chain'
```
pipermerriam: I'd prefix this one too as `fullchain` so there is no ambiguity.  Or maybe even add an extra directory level and do `./chain/full/` and `./chain/light`

tests/p2p/integration_test_helpers.py
```diff
@@ -56,10 +56,16 @@ class FakeAsyncChainDB(AsyncChainDB):
         return self.persist_trie_data_dict(*args, **kwargs)
 
 
+async def coro_import_block(chain, block, perform_validation=True):
+    # Be nice and yield control to give other coroutines a chance to run before us as
+    # importing a block is a very expensive operation.
+    await asyncio.sleep(0)
+    return chain.import_block(block, perform_validation=perform_validation)
```
pipermerriam: Am I correct that from the networking process's perspective this will be run in an executor (via the proxy, via the `async_method` util) and then within the db process, this will block during execution?

gsalgado: Well, that actually happens because of https://github.com/ethereum/py-evm/pull/638/files#diff-763cbc43185be6a4e5239e86e9b71e1dR141. Here we're looking at the `FakeAsyncChain` implementations, whose async methods are actually blocking and hence the `await sleep(0)` at the beginning

pipermerriam: Oh, lol, didn't even notice what file I was in.  👍

PR #639, opened 2018-05-03
WIP: Split Chain out of LightChain class.

pipermerriam:
### What was wrong?

WIP

#### Cute Animal Picture

### DIFF
--- a/evm/chains/__init__.py
+++ b/evm/chains/__init__.py
@@ -1,5 +1,6 @@
 from .base import (  # noqa: F401
     AsyncChain,
+    BaseChain,
     Chain,
 )
 from .mainnet import (  # noqa: F401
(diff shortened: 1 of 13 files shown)


pipermerriam:
@carver @gsalgado I think this PR is going to morph a bit into me implementing a `LightChain` class that is like `evm.chain.base.Chain` but only has methods for doing header retrieval (basically can only deal with the header chain).  It seems like a better approach since in a *light* context, pretty much every API on the `Chain` class won't work.

Then, I'm going to rename the current `p2p.lightchain.LightChain` to something like `LightNetwork` or `LightClient`.

pipermerriam:
Replaced by https://github.com/ethereum/py-evm/pull/664

PR #641, opened 2018-05-04
Add understanding mining guide

cburgdorf:
### What was wrong?

The docs don't cover the basics of the mining process yet.

### How was it fixed?

Started writing on a guide.

understanding_mining

#### Cute Animal Picture

### DIFF
--- a/docs/guides/evm/understanding_the_mining_process.rst
+++ b/docs/guides/evm/understanding_the_mining_process.rst
@@ -1,0 +1,213 @@
+Understanding the mining process
+================================
+
+In the `Building Chains Guide <building_chains>`_ we already learned how to use the
+:class:`~evm.chains.base.Chain` class to create a single blockchain as a combination of different
+virtual machines for different spans of blocks.
+
+In this guide we want to build up on that knowledge and look into the actual mining process.
+
+
+.. note::
+
+  Mining is an overloaded term and in fact the names of the mentioned APIs are subject to change.
+
+
+Mining
+------
+
+The term *mining* can refer to different things depending on our point of view. Most of the time
+when we read about *mining*, we talk about the process where several parties are *competing* to be
+the first to create a new valid block and pass it on to the network.
+
+In this guide, when we talk about the :func:`~evm.chains.base.Chain.mine_block` API, we are only
+referring to the part that creates, validates and sets a block as the new canonical head of the
+chain but not necessarily as part of the mentioned competition to be the first. In fact, the
+:func:`~evm.chains.base.Chain.mine_block` API is internally also called when we import existing
+blocks that others created.
+
+Mining an empty block
+---------------------
+
+Usually when we think about creating blocks we naturally think about adding transactions to the
+block first because, after all, one primary use case for the Ethereum blockchain is to process
+*transactions* which are wrapped in blocks.
+
+For the sake of simplicity though, we'll mine an empty block as a first example (meaning the block
+will not contain any transactions)
+
+As a refresher, he's where we left of as part of the `Building Chains Guide <building_chains>`_.
+
+::
+
+  from evm.db.backends.memory import MemoryDB
+  from evm.chains.mainnet import MAINNET_GENESIS_HEADER
+
+  # initialize a fresh chain
+  chain = chain_class.from_genesis_header(MemoryDB(), MAINNET_GENESIS_HEADER)
+
+Since we decided to not add any transactions to our block let's just call
+:func:`~~evm.chains.base.Chain.mine_block` and see what happens.
+
+::
+
+  # initialize a fresh chain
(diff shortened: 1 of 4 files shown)


docs/guides/evm/understanding_the_mining_process.rst
```diff
@@ -0,0 +1,181 @@
...
+      validate_length(mining_hash, 32, title="Mining Hash")
+      validate_length(nonce, 8, title="POW Nonce")
+      cache = get_cache(block_number)
+      mining_output = hashimoto_light(
+          block_number, cache, mining_hash, big_endian_to_int(nonce))
+      if mining_output[b'mix digest'] != mix_hash:
+          raise ValidationError("mix hash mismatch; {0} != {1}".format(
+              encode_hex(mining_output[b'mix digest']), encode_hex(mix_hash)))
+      result = big_endian_to_int(mining_output[b'result'])
+      validate_lte(result, 2**256 // difficulty, title="POW Difficulty")
```
pipermerriam: Sphinx has a [*literal include* directive](https://build-me-the-docs-please.readthedocs.io/en/latest/Using_Sphinx/ShowingCodeExamplesInSphinx.html#literalinclude-directive) that can be used to inline a section of an external file.  Might be a good way to include this source code without actualy copy/pasting it.

docs/guides/evm/understanding_the_mining_process.rst
```diff
@@ -0,0 +1,181 @@
...
+    File "/py-evm/evm/consensus/pow.py", line 72, in check_pow
+      validate_lte(result, 2**256 // difficulty, title="POW Difficulty")
+    File "/py-evm/evm/validation.py", line 92, in validate_lte
+      title=title,
+  evm.exceptions.ValidationError: POW Difficulty
+  50084831392019586920263880526078408578250402437811116223241331859679343306571
+  is not less than or equal to
+  882992383764307249142653314182893391999679604880738805815775866336575232
+
+Let's make another adjustment and also set the ``difficulty`` to the value we were told that was expected and try again.
```
pipermerriam: cc @carver 

@cburgdorf I know this is the path that you went down, but I'm skeptical of the value it provides to fully explore this in the docs.  It might be more valuable to provide a brief example of how someone *can* CPU mine a block if they so desire, and just to document what values are required to *properly* mine a POW block and potentially link out to external resources for details of the actual mining algorithm.

cburgdorf: Yep, I've been wondering the same thing as I wrote it and I agree that this is more an tutorial, "let's tinker with this" kind of approach.  Maybe we better follow the "happy path" and put in an info box that quickly explains why we can't just try to make up values like this.

cburgdorf:
@pipermerriam I went over this again. This still has plenty of stuff missing so it is not complete in that sense. However, it explains how to use `mine_block` with actual PoW. Notice that this is based on @gsalgado other PR and that I slightly amended the commit to rename the API.

I'm aware that the code in this guide is actually a hacky solution to do PoW mining and the guide explicitly mentions that this is for demo purpose only.

But maybe it is *good enough for now* as this is really just meant as a small byproduct that evolved from when I first started to tinker with the benchmark suite.

cburgdorf:
>How do you feel about wrapping lines at 100 characters?

Not a fan of wrapping lines in markdown or rst files but can do if you feel strongly about it ;)

docs/guides/evm/understanding_the_mining_process.rst
```diff
@@ -0,0 +1,183 @@
...
+    # the actual Proof of Work (PoW) mechanism to mine the correct
+    # nonce and mix_hash for this block
+    nonce, mix_hash = mine_nonce_and_mixhash(
+        block.number,
+        block.header.mining_hash,
+        block.header.difficulty)
+
+    block = chain.mine_block(mix_hash=mix_hash, nonce=nonce)
+
+    print(block)
```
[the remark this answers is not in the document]
carver: Also, it is a step toward doctests, which will run the code after the `>>>` and compare the result with the line following it.

carver:
My editor screams angry red about lines longer than 100 chars, so I do it sort of out of habit. Most of our other projects do wrap these lines. The only notable benefit I can think of is making it easier to read diffs. But it does add some overhead to re-wrap the lines when modifying the docs.

pipermerriam:
> But it does add some overhead to re-wrap the lines when modifying the docs.

This is something your editor should automate for you.  I know vim does this natively.  I'm sure there are ways to do it in most editors.

cburgdorf:
@pipermerriam @carver I think I addressed all open points including wrapping lines at 100 chars. I also amended @gsalgado commit to rename the mining API to `mine_pow_nonce`. Amended the commit directly to avoid *history noise*.

PR #642, opened 2018-05-04
Docs: Cover database APIs

cburgdorf:
### What was wrong?

Database APIs aren't covered in docs yet

### How was it fixed?

Adds docs for database APIs

#### Cute Animal Picture

### DIFF
--- a/docs/api/evm/api.db.rst
+++ b/docs/api/evm/api.db.rst
@@ -1,0 +1,13 @@
+DataBase
+========
+
+
+.. toctree::
+   :maxdepth: 4
+   :name: toc-evm-api-db
+   :caption: Database
+
+   db/api.db.backends
+   db/api.db.account
+   db/api.db.journal
+   db/api.db.chain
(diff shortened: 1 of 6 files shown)


pipermerriam:
cc @carver 

@cburgdorf this API is likely to change a lot very soon... so maybe punt on this for a bit to remove the overhead of rewriting this.

cburgdorf:
@pipermerriam good point. I actually wanted to add and tune the docstrings in the source files but I'll postpone. That said, we *could* take the PR as is then to just have *some* sort of database coverage and since we are just pulling in the docs from the source files it will automatically pull in any updates at least for the types that keep their place and name.

I'll just add the cutest animal to the PR and have you decide to either close or merge.

cburgdorf:
@pipermerriam what's your current thinking here? I actually would like to quickly bring up all API docs that already exist in the code. The effort seems pretty manageable since it's mostly just pointers using `autoclass` directives. So, even if things get dropped it's not that we waste a lot of time.

Once the API docs are online in *some way* we can start to improve the actual docstrings in the code which is automatically reflected in the online docs.

PR #643, opened 2018-05-04
Pin mypy to versions before 0.600

gsalgado:
That release seems to be doing a lot more checks by default, and we're not ready for that yet. More info at http://mypy-lang.blogspot.pt/2018/05/mypy-0600-released.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ deps = {
     ],
     'lint': [
         "flake8==3.5.0",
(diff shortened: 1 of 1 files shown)


carver:
cc @cburgdorf so it seems like a good idea to start using `Optional[...]` on incremental changes (when a value might be `None`), in preparation for eventually doing strict type checking.

PR #644, opened 2018-05-04
p2p: Add NAT port-mapping support to p2p.Server

### LINKED ISSUE #636: Need to add NAT port mapping support to p2p.Server
Most of the times people will run py-evm behind a NAT so other peers in the network won't be able to connect unless we have NAT port mapping configured. However, we should not require users to mess with their routers to set it up manually, so we must do that automatically for them. One way to achieve that is via upnp, and there are python libraries for that: https://github.com/flyte/upnpclient

gsalgado:
Closes #636

### DIFF
--- a/p2p/server.py
+++ b/p2p/server.py
@@ -7,6 +7,8 @@ from typing import (
     Type,
 )
 
+import upnpclient
+
 from eth_keys import datatypes
 
 from eth_utils import big_endian_to_int
@@ -30,6 +32,7 @@ from p2p.discovery import DiscoveryProtocol
 from p2p.exceptions import (
     DecryptionError,
     HandshakeFailure,
+    OperationCancelled,
 )
 from p2p.kademlia import (
     Address,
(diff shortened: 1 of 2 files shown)


p2p/server.py
```diff
@@ -74,6 +77,56 @@ def __init__(self,
...
+        # Use loop.run_in_executor() because upnpclient.discover() is blocking and may take a
+        # while to complete.
+        devices = await wait_with_token(
+            loop.run_in_executor(None, upnpclient.discover),
+            token=self.cancel_token)
+        if not devices:
+            self.logger.info("No UPNP-enabled devices found")
+            return
+        device = devices[0]
+        device.WANIPConn1.AddPortMapping(
```
pipermerriam: No need to deal with this now, but I *think* that in my tinkering yesterday I found this property was not reliably present on all device objects which showed up in my network.

setup.py
```diff
@@ -37,6 +37,7 @@
         "plyvel==1.0.4",
         "coincurve>=7.0.0,<8.0.0",
         "web3>=4.1.0,<5.0.0",
+        "upnpclient>=0.0.8",
```
pipermerriam: Can you change to the pin below `1` as well `upnpclient>=0.0.8,<1`

carver: Maybe this belongs in the `p2p` extra, instead?

gsalgado: Yeah, it does!

p2p/server.py
gsalgado: Yeah, I'm sure this is going to crash in all sorts of ways, but I guess we'll have to rely on users reporting those crashes back to us so that we can handle all variants of this property

PR #645, opened 2018-05-04
Attempt at a very simple sharding protocol

jannikluhn:
Goal:
- [x] simple sharding protocol with two message types: `Status` (for handshake) and `Collations` (to send collations)
- [x] peers that listen to collations, add them to their local DB, and relay them to their peers
- [x] peers that propose collations in random time intervals
- [x] tests
- [x] implement remaining suggestions in #570

For the tests I've tried to setup two local nodes that are connected to each other (https://github.com/ethereum/py-evm/compare/master...jannikluhn:sharding-protocol?expand=1#diff-f7c78d280d6b0f400c831d798adf0781R152), but I didn't manage to do that (the `assert` checking if the nodes are connected fails). Since I've been stuck for quite some time on that and haven't made much progress, could someone who's familiar with the networking code could have a quick look?

#### Cute Animal Picture

### DIFF
--- a/p2p/sharding.py
+++ b/p2p/sharding.py
@@ -149,7 +149,6 @@ class ShardSyncer(BaseService, PeerPoolSubscriber):
         self.incoming_collation_queue: asyncio.Queue[Collation] = asyncio.Queue()
 
         self.collations_received_event = asyncio.Event()
-        self.collations_proposed_event = asyncio.Event()
 
         self.cancel_token = CancelToken("ShardSyncer")
         if token is not None:
@@ -199,9 +198,6 @@ class ShardSyncer(BaseService, PeerPoolSubscriber):
         # broadcast collation
         for peer in self.peer_pool.peers:
             cast(ShardingPeer, peer).send_collations([collation])
-
-        self.collations_proposed_event.set()
-        self.collations_proposed_event.clear()
 
         return collation
 
--- a/p2p/sharding.py
+++ b/p2p/sharding.py
@@ -16,6 +16,13 @@ from evm.db.shard import (
     Availability,
 )
 
+from evm.constants import (
+    COLLATION_SIZE,
+)
+from evm.exceptions import (
+    CanonicalCollationNotFound,
+)
+
 from p2p.cancel_token import (
     CancelToken,
     wait_with_token,
@@ -108,14 +115,13 @@ class ShardingPeer(BasePeer):
     def _handle_collations_msg(self, msg: List[Collation]) -> None:
         for collation in msg:
             self.known_collation_hashes.add(collation.hash)
-            print(self.known_collation_hashes)
             if self.incoming_collation_queue is not None:
                 self.incoming_collation_queue.put_nowait(collation)
 
     def send_collations(self, collations: List[Collation]) -> None:
         for collation in collations:
             if collation.hash not in self.known_collation_hashes:
-                self.known_collation_hashes.put(collation.hash)
+                self.known_collation_hashes.add(collation.hash)
                 self.sub_proto.send_collations(collations)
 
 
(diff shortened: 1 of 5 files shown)


jannikluhn:
Seems to work now, so this would be ready for a preliminary review.

The tests pass at least locally. There might be timeout issues on CircleCI though because calculating the chunk root takes a lot of time.

Stuff I still on my todo list:

- there's some logic in `ShardingPeer` which should probably be done in `ShardSyncer`, if only for consistency with the rest of the p2p codebase.
- add some more elaborate tests (more peers, not directly connected peers, more shards)
- make tests a little clearer by looking at the received/proposed collations

p2p/sharding.py
```diff
@@ -0,0 +1,209 @@
...
+    # Handshake
+    #
+    async def send_sub_proto_handshake(self) -> None:
+        self.sub_proto.send_handshake()
+
+    async def process_sub_proto_handshake(
+        self,
+        cmd: Command,
+        msg: protocol._DecodedMsgType
+    ) -> None:
```
pipermerriam: indentation *nitpick*.

I'd put all of this indented out at the level of the end of the function name.  This is a bit confusing to visually parse imho.

p2p/sharding.py
```diff
@@ -0,0 +1,209 @@
...
+    def handle_sub_proto_msg(self, cmd: Command, msg: protocol._DecodedMsgType) -> None:
+        if isinstance(cmd, Collations):
+            self._handle_collations_msg(msg)
+        else:
+            super().handle_sub_proto_msg(cmd, msg)
+
+    def _handle_collations_msg(self, msg: List[Collation]) -> None:
+        for collation in msg:
+            self.known_collation_hashes.add(collation.hash)
+            if self.incoming_collation_queue is not None:
```
pipermerriam: This warrants a comment explanation.  Why is this *sometimes* `None` and how does it get set.

p2p/sharding.py
```diff
@@ -0,0 +1,209 @@
...
+
+class ShardSyncer(PeerPoolSubscriber):
+
+    def __init__(
+        self,
+        shard: Shard,
+        mean_proposing_period: int,
+        peer_pool: PeerPool,
+        token: CancelToken = None,
+    ) -> None:
```
pipermerriam: Same beef with this indentation style.  Hard to visually parse imho.  Consider either:

```python
def my_method(self
              arg_2,
              arg_3) -> None:
# or
def my_method(
        self,
        arg_2,
        arg_3) -> None:
    thing = ...
```

p2p/sharding.py
```diff
@@ -0,0 +1,215 @@
...
+                self.sub_proto.send_collations(collations)
+
+
+class ShardSyncer(PeerPoolSubscriber):
+    logger = logging.getLogger("p2p.sharding.ShardSyncer")
+
+    def __init__(self, shard: Shard, peer_pool: PeerPool, token: CancelToken = None) -> None:
+        self.shard = shard
+        self.peer_pool = peer_pool
+        self.peer_pool.subscribe(self)
```
gsalgado: You should only `subscribe()` in your `run()` method because if you get a PeerPool which already has connected peers here, your `register_peer()` method will be called before your `__init__()` has finished and you may not have everything in place to handle a peer

gsalgado: I've realized that today while testing `FullNodeSyncer`, which also `subscribe()`d in its `__init__()`

p2p/sharding.py
```diff
@@ -0,0 +1,215 @@
...
+
+        self.incoming_collation_queue = asyncio.Queue()
+
+        self.collations_received_event = asyncio.Event()
+        self.collations_proposed_event = asyncio.Event()
```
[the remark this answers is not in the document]
jannikluhn: Nowhere, I removed them. I've added them initially because peers proposed randomly on their own, but proposing by simply calling a method turned out to be easier to test..

tests/p2p/test_sharding.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+
+    await asyncio.gather(
+        peer1.do_p2p_handshake(),
+        peer2.do_p2p_handshake(),
+    )
+    await asyncio.gather(
+        peer1.do_sub_proto_handshake(),
+        peer2.do_sub_proto_handshake(),
+    )
```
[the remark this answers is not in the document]
jannikluhn: Added the assertions you suggested, but that's pretty much all we can do. The `do_sub_proto_handshake()` methods raise exceptions when the handshake fails, so just calling them should be sufficient.

PR #649, opened 2018-05-05
WIP - squashing trie data during block import

carver:
### What was wrong?

We are storing state roots into the database for every step in the transaction. We only need to store the state roots from the end of the block.

### How was it fixed?

Use the `HexaryTrie`s ability to squash changes, and join all the state changes in a block to one. We don't need to index into state roots from the middle of a block (or the middle of a transaction!), so we can drop them all, except the final one.

It seems natural to do a follow-up where we don't bother calculating the hashes along the way

Also, moved `BaseDB` closer to implementing the `collections.abc.MutableMapping` interface, by adding an optional default parameter. (`HexaryTrie` expects the db to meet that interface)

There are a few hacks in place to make this work without the PR getting a lot bigger. Probably the biggest follow-up todo item is to rethink the trie and account db, and how they fit together, finally getting to #507 

Also, we're technically storing the state root at the end of the last transaction, and then again after the block finalization (eg~ paying the miner), but we'll clean that up later.

#### Cute Animal Picture

### DIFF
--- a/evm/db/backends/base.py
+++ b/evm/db/backends/base.py
@@ -2,6 +2,13 @@ from abc import (
     ABCMeta,
     abstractmethod
 )
+from typing import Union
+
+
+class Unset:
+    pass
+
+unset = Unset()
 
 
 class BaseDB(metaclass=ABCMeta):
@@ -12,15 +19,24 @@ class BaseDB(metaclass=ABCMeta):
             "The `init` method must be implemented by subclasses of BaseDB"
         )
 
-    @abstractmethod
-    def get(self, key: bytes) -> bytes:
-        """Return the value for the given key.
-
-        Raises KeyError if key doesn't exist.
+    def get(self, key: bytes, default: Union[bytes, Unset] = unset) -> bytes:
         """
-        raise NotImplementedError(
-            "The `get` method must be implemented by subclasses of BaseDB"
-        )
+        Return the value for the given key.
+
+        If the key doesn't exist, and a default is provided, return the default value.
+        If the key doesn't exist, and a default is not provided, raise a KeyError
+
+        :return: the value with the associated key
+        :raise: KeyError if key is missing
+        """
+        try:
+            return self[key]
+        except KeyError as exc:
+            if default is unset:
+                raise exc
+            else:
+                return default
+
 
     @abstractmethod
     def set(self, key: bytes, value: bytes) -> None:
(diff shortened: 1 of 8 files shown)


evm/db/backends/base.py
```diff
@@ -12,15 +19,24 @@ def __init__(self) -> None:
...
+        :raise: KeyError if key is missing
         """
-        raise NotImplementedError(
-            "The `get` method must be implemented by subclasses of BaseDB"
-        )
+        try:
+            return self[key]
+        except KeyError as exc:
+            if default is unset:
+                raise exc
```
pipermerriam: Anyone have a preference for re-raising.  I've always used a bare `raise` like this (which looks like it'd allow you to remove the `as exc` part.

```python
try:
    thing()
except SomeError:
    raise
```

evm/vm/state.py
```diff
@@ -69,6 +69,12 @@ def __init__(self, db, execution_context, state_root):
         self.execution_context = execution_context
         self.account_db = self.get_account_db_class()(self._db, state_root)
 
+    def copy(self, **kwargs):
+        kwargs.setdefault('db', self._db)
+        kwargs.setdefault('execution_context', self.execution_context)
+        kwargs.setdefault('state_root', self.state_root)
```
pipermerriam: I think your implementation is cleaner, but I ended up writing this out to see.

```python
def copy(self, **overrides):
    defaults = {
        'db': self.db, 
        'execution_context': self.execution_context, 
        'state_root': self.state_root,
    }
    kwargs = merge(defaults, overrides)
    return type(self)(**kwargs)
```

gsalgado:
Are there any cases where we wouldn't want to use a squashing trie? If not, would it make sense to use it by default everywhere?

carver:
> Are there any cases where we wouldn't want to use a squashing trie?

AFAICT, it all comes down to whether we want to access historical state. We don't need to access historical state in the middle of a transaction or block, but we do need to access historical state at the end of previous blocks.

> If not, would it make sense to use it by default everywhere?

If you mean to use it without a context, I think this would be problematic. Squashing in a context means we can be sure to delete all the intermediate state without accidentally deleting any state we need for historical access. But in general, most places that we update the trie should probably be squashed.

carver:
:+1: opened #657

carver:
Closing in favor of #704

carver:
> If not, would it make sense to use it by default everywhere?

I figured a different solution to this, in #704 -- prune the trie always, but add a layer that ignores deletes on the wrapped database at `persist()` time.

All my favorite parts of this PR have now been isolated into others. Closing

Issue #592, opened 2018-04-24
Add benchmarking test suite.


pipermerriam:
Currently we don't have good benchmarking tools.

Lets write a script that does the following.

- ~~For each set of mainnet VM rules, build a chain of X empty blocks.~~ ✅
  - ~~one using `mine_block`~~ ✅
  - ~~another using `import_block`~~ ✅
- ~~For each set of mainnet VM rules, build a chain of X block, each full of simple value transfer transactions.~~ ✅
  - ~~`sender` should be static across all tests~~ ✅
  - ~~one run using empty `to` accounts.~~ ✅
  - ~~one run using existing `to` accounts.~~ ✅
- For each set of mainnet VM rules, benchmark using the open zeppelin ERC20 token contract.
  - one run with X calls to `transfer`
  - one run with X calls to `approve`
  - one run with X calls to `transferFrom`
- ~For each set of mainnet VM rules, benchmark deployment of the open zeppelin ERC20 token contract.~ ✅
- For each set of mainnet VM rules, benchmark DOS style transaction spam.
  - Make X calls to a contract which uses all available gas to create empty contracts.
  - Make X calls to a contract which uses all available gas on SSTORE operations.
  - Two more same as above but which revert at the end of contract execution.
- ~For each set of mainnet VM rules, benchmark gas per second for block import.~ ✅
  - ~Measure for pure value transfer transactions.~ ✅
  - ~Measure for representative EVM execution (which will have to be an approximation.  probably worth starting with just ERC20 contracts and transfers and then expanding from there.~ ✅
- ensure blocks use pre  generated fixtures with real blocks that qualify the PoW so that we don't miss to benchmark the POW validation/verification overhead
- have at least one benchmark to use `LevelDB` rather than `MemoryDB` (ideally all of them but may be too big of an performance hit)

These benchmark scripts should produce detailed output to the console and be added as a stand-alone test run in our CI environment.  Any failing benchmark scripts should cause CI to fail.

Stats worth tracking:

- time to run
- total DB size

carver:
> Any failing benchmark scripts should cause CI to fail.

Where passing presumably means both correctness and within some tolerance of a performance target.

pipermerriam:
> Where passing presumably means both correctness and within some tolerance of a performance target.

[...]

PR #650, opened 2018-05-05
Add benchmarking suite

cburgdorf:
### What was wrong?

Missing benchmark suite (See #592 )

### How was it fixed?

This PR adds a first rudimentary version of a benchmarking suite under `evm.tools.benchmark`. It defines some utilities as well as `BaseBenchmark` class that benchmarks are expected to derive from and tries to generalize as much as possible from the heavy lifting so that benchmark implementations should be rather slim.

Here's a gif of it in action.

benchmark_latest

We should still keep #592 open and I intend to continue to work on this but I feel this is good enough to be given another review and potentially getting merged as groundwork.

#### Cute Animal Picture

### DIFF
--- a/scripts/benchmark/checks/simple_value_transfers.py
+++ b/scripts/benchmark/checks/simple_value_transfers.py
@@ -1,0 +1,137 @@
+import logging
+
+from typing import (
+    NamedTuple,
+    Tuple,
+)
+
+from eth_typing import (
+    Address,
+)
+from evm.chains.base import (
+    Chain,
+)
+from evm.rlp.blocks import (
+    BaseBlock,
+)
+
+from .base_benchmark import (
+    BaseBenchmark,
+)
+from utils.chain_plumbing import (
+    FUNDED_ADDRESS,
+    FUNDED_ADDRESS_PRIVATE_KEY,
+    get_all_chains,
+    SECOND_EXISTING_ADDRESS,
+)
+from utils.address import (
+    generate_random_address,
+)
+from utils.reporting import (
+    DefaultStat,
+)
+from utils.shellart import (
+    bold_yellow,
+)
+from utils.tx import (
+    new_transaction,
+)
+
+
+class SimpleValueTransferBenchmarkConfig(NamedTuple):
+    to_address: Address
+    greeter_info: str
+    num_blocks: int = 1
+
+
+TO_EXISTING_ADDRESS_CONFIG = SimpleValueTransferBenchmarkConfig(
+    to_address=SECOND_EXISTING_ADDRESS,
+    greeter_info='Sending to existing address\n'
+)
+
+
+TO_NON_EXISTING_ADDRESS_CONFIG = SimpleValueTransferBenchmarkConfig(
+    to_address=None,
+    greeter_info='Sending to non-existing address\n'
+)
+
+# TODO: Investigate why 21000 doesn't work
+SIMPLE_VALUE_TRANSFER_GAS_COST = 22000
+
+
+class SimpleValueTransferBenchmark(BaseBenchmark):
+
+    def __init__(self, config: SimpleValueTransferBenchmarkConfig) -> None:
+        self.config = config
+
+    @property
+    def name(self) -> str:
+        return 'Simple value transfer'
+
+    def print_result_header(self) -> None:
+        logging.info(bold_yellow(self.config.greeter_info))
+        super().print_result_header()
+
+    def execute(self) -> DefaultStat:
+        total_stat = DefaultStat()
+        num_blocks = self.config.num_blocks
+
+        for chain in get_all_chains():
+
+            value = self.as_timed_result(lambda: self.mine_blocks(chain, num_blocks))
+
+            total_gas_used, total_num_tx = value.wrapped_value
+
+            stat = DefaultStat(
+                caption=chain.get_vm().fork,
+                total_blocks=num_blocks,
+                total_tx=total_num_tx,
+                total_seconds=value.duration,
+                total_gas=total_gas_used,
+            )
+            total_stat = total_stat.cumulate(stat)
+            self.print_stat_line(stat)
+
+        return total_stat
+
+    def mine_blocks(self, chain: Chain, num_blocks: int) -> Tuple[int, int]:
+        total_gas_used = 0
+        total_num_tx = 0
+
+        for i in range(1, num_blocks + 1):
+            num_tx = chain.get_block().header.gas_limit // SIMPLE_VALUE_TRANSFER_GAS_COST
+            block = self.mine_block(chain, i, num_tx)
+            total_num_tx = total_num_tx + len(block.transactions)
+            total_gas_used = total_gas_used + block.header.gas_used
+
+        return total_gas_used, total_num_tx
+
+    def mine_block(self, chain: Chain, block_number: int, num_tx: int) -> BaseBlock:
+        for i in range(1, num_tx + 1):
+            self.apply_transaction(chain)
+
+        return chain.mine_block()
+
+    def apply_transaction(self, chain: Chain) -> None:
+
+        if self.config.to_address is None:
+            to_address = generate_random_address()
+        else:
+            to_address = self.config.to_address
+
+        tx = new_transaction(
+            vm=chain.get_vm(),
+            private_key=FUNDED_ADDRESS_PRIVATE_KEY,
+            from_=FUNDED_ADDRESS,
+            to=to_address,
+            amount=100,
+            data=b''
+        )
+
+        logging.debug('Applying Transaction {}'.format(tx))
+
+        block, receipt, computation = chain.apply_transaction(tx)
+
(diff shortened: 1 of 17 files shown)


evm/vm/forks/__init__.py
```diff
@@ -13,3 +13,11 @@
 from .byzantium import (  # noqa: F401
     ByzantiumVM,
 )
+
+ALL_VM = [
+    TangerineWhistleVM,
+    FrontierVM,
+    HomesteadVM,
+    SpuriousDragonVM,
+    ByzantiumVM
```
pipermerriam: A potentially better way to get this list would be to extract them from the`MainnetChain`

scripts/benchmark/utils/decorator.py
```diff
@@ -0,0 +1,75 @@
...
+        return decorator
+
+    # @timecall syntax -- we are a decorator.
+    fp = FuncTimer(fn, log_level=log_level, formatter=formatter)
+
+    # We cannot return fp or fp.__call__ directly as that would break method
+    # definitions, instead we need to return a plain function.
+    def new_fn(*args: Any, **kw: Any) -> Any:
+        return fp(*args, **kw)
+    new_fn.__doc__ = fn.__doc__
```
pipermerriam: Instead of all of this I think using `functools.wraps` is the proper utility to decorate the `new_fn`

cburgdorf: Thanks for the pointer. I had the feeling this could be improved. This file is largely copied from somewhere :)

evm/vm/forks/__init__.py
cburgdorf: :+1: as in `ALL_VM = [vm for _, vm in MainnetChain.vm_configuration]` I guess

cburgdorf:
@pipermerriam I moved the benchmark under `evm.tools` but unfortunately that means the code needs to be Python 3.5 complaint. There doesn't seem an elegant way to just have `evm.tools.benchmark` be checked against Python 3.6. The last commit deals entirely with making the code Python 3.5 complaint which totally ruins the fun (mainly workarounds for less good `NamedTuple` support).

This got me thinking if we should just move `evm.tools` out of `evm` into a standalone `tools` or `evm-tools` package which could use Python 3.6.

This also comes at the benefit that the evm doesn't implicitly inherit any added dependencies from `evm.tools` anymore (remember we had to workaround `Pytest` and the benchmarking suite also adds another dependency on `termcolor` which we may have to treat the same way otherwise)

Would there be anything wrong with moving `tools` out of the `evm` package?

pipermerriam:
> Would there be anything wrong with moving `tools` out of the `evm` package?

I don't think this will fix your problem.  The `fixture_tests` and `test_builder` both need to support python 3.5.

pipermerriam:
Regarding the benefits of removing un-necessary dependencies, 👍 but maybe we can just do this by having `evm[tools]` and doing explicit dependency checks in `evm/tools/__init__.py` for the necessary dependencies and raising a helpful error when they aren't found.

cburgdorf:
>I don't think this will fix your problem. The fixture_tests and test_builder both need to support python 3.5.

Ah, yep, there was something. Well, tbh then, I'd vote for having the benchmark living somewhere *outside* of the evm because frankly, I don't see why we should restrict ourselves to writing older Python if it's not absolutely necessary.

Would you be ok with that?

cburgdorf:
For clarify, this is the link to this CI's benchmark run https://circleci.com/gh/ethereum/py-evm/24024

pipermerriam:
I think the number of rounds needs to be bumped up some.  Something like 500-1000 blocks.

scripts/benchmark/checks/simple_value_transfers.py
```diff
@@ -0,0 +1,121 @@
...
+    def name(self) -> str:
+        return 'Simple value transfer'
+
+    def print_result_header(self) -> None:
+        logging.info(bold_yellow(self.config.greeter_info))
+        super().print_result_header()
+
+    def execute(self) -> DefaultStat:
+        number_blocks = 10
+        number_tx = 10
```
pipermerriam: It looks like you're using a static number of transactions per block.  Can you change this to use `block.gas_limit // 21000` which should be the number of value transfer transactions that can be sent within a single block.

scripts/benchmark/checks/simple_value_transfers.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+TO_EXISTING_ADDRESS_CONFIG = SimpleValueTransferBenchmarkConfig(
+    to_address=SECOND_EXISTING_ADDRESS,
+    greeter_info='Sending to existing address\n'
+)
+
+
+TO_NON_EXISTING_ADDRESS_CONFIG = SimpleValueTransferBenchmarkConfig(
+    to_address=NON_EXISTING_ADDRESS,
```
pipermerriam: This won't actually work.  You need to generate a new address for each transaction to benchmark the thing we're actually trying to test, which is the additional overhead we incur for transactions which create new accounts in the state database.  If we re-use the same address then only the first transaction triggers this and the rest are just simple value transfers.

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -0,0 +1,131 @@
...
+    'block_number': constants.GENESIS_BLOCK_NUMBER,
+    'gas_limit': constants.GENESIS_GAS_LIMIT,
+    'extra_data': constants.GENESIS_EXTRA_DATA,
+    'nonce': constants.GENESIS_NONCE
+}
+
+
+def genesis_state(setup: Iterable[AddressSetup]) -> Any:
+
+    state = {}
```
pipermerriam: Not a big deal but we have a nice decorator for doing this in a non-state mutating way.

```python
from eth_utils import to_dict

@to_dict
def make_a_dict():
    yield 'key-a', 1
    yield 'key-b', 2

>>> make_a_dict()
{'key-a': 1, 'key-b': 2}
```

scripts/benchmark/utils/format.py
```diff
@@ -0,0 +1,14 @@
...
+    encode_hex,
+)
+
+
+def format_block(block: BaseBlock) -> str:
+    return (
+        "\n\n"
+        "------------------------Block------------------------------------\n"
+        "Number #{b.number}\n"
+        "Hash {hash}\n"
```
pipermerriam: nitpick, but these could probably be put on the same line if you put the number first and print it with a fixed-width of about 10-12 characters depending on how future proof you want to be.

cburgdorf:
>I think the number of rounds needs to be bumped up some. Something like 500-1000 blocks.

Sure, no problem. Otherwise ok to merge?

scripts/benchmark/utils/reporting.py
```diff
@@ -0,0 +1,110 @@
...
+        '|{stat.gas_per_second:^16,.3f}'
+        '|'.format(
+            stat=stat
+        ))
+
+
+def print_default_benchmark_total_line(stat: DefaultStat) -> None:
+    logging.info(SINGLE_UNDERLINE)
+    logging.info(bold_white(
+        '|{caption:^19}'
```
pipermerriam: Not a big deal at all, and not something we should try to do in this PR, but we could potentially:

1. Use a special class of [`LogRecord`](https://docs.python.org/3/library/logging.html#logrecord-objects)
2. Use the contextual data APIs: https://docs.python.org/3/howto/logging-cookbook.html#context-info

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -0,0 +1,131 @@
...
+            "balance": value.balance,
+            "nonce": 0,
+            "code": value.code,
+            "storage": {}
+        }
+
+    return state
+
+
+def chain_without_pow(
```
pipermerriam: Can you open up an issue to migrate these to use pre-generated fixtures so that we can be sure to capture the overhead caused by POW validation/verification.

scripts/benchmark/checks/simple_value_transfers.py
cburgdorf: Oops, of course! 🙈

scripts/benchmark/utils/chain_plumbing.py
cburgdorf: I updated the original ticket and propose we keep that open as tracking ticket until we have worked through the entire list of things that we want to get done regarding the benchmarking.

https://github.com/ethereum/py-evm/issues/592

cburgdorf:
>I think the number of rounds needs to be bumped up some. Something like 500-1000 blocks

I bumped it up to 500 per benchmark but that puts us at 21 minutes for the entire suite to run and we still want to  add a couple more benchmarks. What are your thoughts on that?

scripts/benchmark/utils/chain_plumbing.py
cburgdorf: I like that :+1:

PR #653, opened 2018-05-07
p2p: Pass CancelTokens to handshake() functions

gsalgado:
CancelTokens should be passed down the the lowest level where they are
then used to cancel asyncio tasks. When performing handshakes, we were
using them to cancel the handshake task wrapping the actual asyncio
tasks, which could leave those asyncio tasks behind and cause python to
emit warnings upon exit (because it doesn't like to close an event loop
when there are pending tasks)


tests/p2p/test_server.py
```diff
@@ -71,49 +88,31 @@ def server():
...
-    )
-    server_2 = get_server(
-        INITIATOR_PRIVKEY,
-        INITIATOR_ADDRESS,
-    )
-
-    await server_1.start()
-    await server_2.start()
+async def test_two_servers(
+        event_loop, receiver_server_with_dumb_peer, initiator_server_with_dumb_peer):
```
pipermerriam: *nitpick*

style guide has us dropping each of these onto its own line.

PR #655, opened 2018-05-07
Perform sub proto handshake in server

jannikluhn:
### What was wrong?

When receiving a connection, `p2p.Server` did only perform the p2p handshake but not the sub protocol handshake.

### How was it fixed?

Do the sub protocol handshake after a successful p2p handshake.

#### Cute Animal Picture

### DIFF
--- a/p2p/server.py
+++ b/p2p/server.py
@@ -238,17 +238,24 @@ class Server:
             network_id=self.network_id
         )
 
-        await self.do_p2p_handshake(peer)
+        await self.do_handshake(peer)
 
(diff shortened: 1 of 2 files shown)


pipermerriam:
Generally looks good.  Is there a test we can add to cover this?

gsalgado:
> Generally looks good. Is there a test we can add to cover this?

Currently the test reads but discards the auth ack from the server. It should decode that (like in `auth._handshake()`) and then we could use the secrets derived from that to create a `Peer` instance and send p2p/sub-proto handshake msgs to the server

gsalgado:
Filed #656 for that. Otherwise looks good

PR #658, opened 2018-05-07
A new DBDiff and DBDiffTracker, to track db changes

carver:
### What was wrong?

Many classes share access to a bunch of database state, and different layers all necessarily share the same view of the state. This leads to hard-to-reason-about code.

### How was it fixed?

One step on the way there is to explicitly track changes to a database, and share those changes. Because the changes can include deletes, it's not *quite* as easy as a simple `dict`, but almost. A new `DBDiff` (read-only) and `DBDiffTracker` (read/write) were added to unify any key/value diffs that we pass around between classes.

Instead of modifying a database in place, this enables classes to return diffs that were a result of their actions. The calling method can keep the diff in memory, drop it, or apply it, at the leisure of the caller.

#### Cute Animal Picture

### DIFF
--- a/evm/db/diff.py
+++ b/evm/db/diff.py
@@ -1,0 +1,113 @@
+from collections.abc import (
+    Mapping,
+    MutableMapping,
+)
+from typing import (  # noqa: F401
+    Dict,
+    Iterable,
+    Union,
+)
+
+from cytoolz import (
+    merge,
+)
+
+
+class MissingReason:
+    def __init__(self, reason):
+        self.reason = reason
+
+    def __str__(self, reason):
+        return "Key is missing because it was {}".format(self.reason)
+
+
+NEVER_INSERTED = MissingReason("never inserted")
+DELETED = MissingReason("deleted")
+
+
+class DiffMissingError(KeyError):
+    def __init__(self, missing_key: bytes, reason: MissingReason) -> None:
+        self.reason = reason
+        super().__init__(missing_key, reason)
+
+    def is_deleted(self):
+        return self.reason == DELETED
+
+
+class DBDiffTracker(MutableMapping):
+    """
+    Recorded changes to a :class:`~evm.db.BaseDB`
+    """
+    def __init__(self):
+        self._changes = {}  # type: Dict[bytes, Union[bytes, DiffMissingError]]
+
+    def __contains__(self, key):
+        result = self._changes.get(key, NEVER_INSERTED)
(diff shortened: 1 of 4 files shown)


evm/db/diff.py
```diff
@@ -0,0 +1,113 @@
...
+                try:
+                    del db[key]
+                except KeyError:
+                    pass
+            else:
+                db[key] = value
+
+    @classmethod
+    def join(cls, diffs: Iterable['DBDiff']) -> 'DBDiff':
+        new_diff = cls()
```
pipermerriam: This probably deserves a docstring indicating precedence ordering for overlapping diffs.

evm/db/diff.py
```diff
@@ -0,0 +1,113 @@
...
+        new_diff._changes = merge(diff._changes for diff in diffs)
```
pipermerriam: This private attribute access makes me thing that the class should be able to be instantiated with an existing dictionary of diff changes.

tests/database/test_db_diff.py
```diff
@@ -0,0 +1,141 @@
...
+    assert b'key-2' not in db
+
+
+def test_database_api_missing_key_retrieval(db):
+    assert db.get(b'does-not-exist') is None
+
+    try:
+        val = db[b'does-not-exist']
+    except DiffMissingError as exc:
+        assert not exc.is_deleted()
```
pipermerriam: I was surprised that this was a function and not a property.  I think functions named like this are foot guns because were this line to be `if exc.is_deleted` it would always evaluate to `True` because `bool(fn)` returns `True`.

evm/db/diff.py
carver: I wanted to avoid a public way to initialize with a `dict` of changes, because it would be easy to do it wrong (like assume that a `None`) value would delete the underlying. These two classes inherently share this special type of dict.

So I guess a third option is to create a third class whose job is essentially to document the deletion mechanism, and do a little type checking with mypy, like:

```py
class DBDiffChanges(dict):
    """
    Changes recorded by `DBDiffTracker` and later used by `DBDiff`. It
    has a special internal mechanism for tracking deletes. Best not to try
    to build this manually, and to use the above two classes, instead.
    """
    pass
```

pipermerriam: It seems like we should be able to get pretty far with mypy checks and maybe even explicit validation of key/value types if the overhead is acceptable.

pipermerriam: do what you think is best

PR #659, opened 2018-05-07
Add Hash32 type hinting

pipermerriam:
### What was wrong?

We have a `Hash32` type to differentiate between just any old `bytes` value and a 32 byte hash.  It wasn't being used in many places.

### How was it fixed?

Added type hints for `block_hash` type values.

#### Cute Animal Picture

80cbf21372410372a88921f7a2fc628f

### DIFF
--- a/eth_typing/__init__.py
+++ b/eth_typing/__init__.py
@@ -6,5 +6,4 @@ from .misc import (  # noqa: F401
     Address,
     BlockNumber,
     Hash32,
-    HeaderParams,
 )


evm/chains/base.py
```diff
@@ -71,23 +88,27 @@ class BaseChain(Configurable, metaclass=ABCMeta):
...
     #
     @classmethod
     @abstractmethod
     def from_genesis(cls,
-                     chaindb,
-                     genesis_params,
-                     genesis_state=None):
+                     chaindb: BaseChainDB,
+                     genesis_params: Dict,
+                     genesis_state: Dict[Address, Dict]=None) -> 'BaseChain':
```
cburgdorf: Can we specify the type of the inner Dict or if not make it explicitly `Dict[Any, Any] `. *Implicit* Any is not allowed once we migrate to stricter checks (for good reasons)

evm/chains/base.py
```diff
@@ -115,7 +136,9 @@ def get_block_header_by_hash(self, block_hash):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def create_header_from_parent(self, parent_header, **header_params):
+    def create_header_from_parent(self,
+                                  parent_header: BlockHeader,
+                                  **header_params: Dict) -> BlockHeader:
```
cburgdorf: Same here, can we get rid of the *implicit* `Any` for the `Dict` by *explicitly* setting the types.

evm/chains/base.py
```diff
@@ -142,29 +165,29 @@ def get_canonical_block_by_number(self, block_number):
...
         return vm.get_block_class().from_header(block_header, self.chaindb)
 
-    @to_tuple
-    def get_ancestors(self, limit):
-        lower_limit = max(self.header.block_number - limit, 0)
-        for n in reversed(range(lower_limit, self.header.block_number)):
-            yield self.get_canonical_block_by_number(n)
+    def get_ancestors(self, limit: int) -> Generator[BaseBlock, None, None]:
+        """
+        Returns `limit` number of ancestor blocks from the current canonical head.
```
cburgdorf: Nitpick, we want imperative present tense. `Return` not `Returns`

evm/chains/base.py
```diff
@@ -427,13 +456,13 @@ def get_block_by_hash(self, block_hash):
     #
     @classmethod
     def from_genesis(cls,
-                     chaindb,
-                     genesis_params,
-                     genesis_state=None):
+                     chaindb: BaseChainDB,
+                     genesis_params: Dict,
```
cburgdorf: Same here, can we get rid of the *implicit* `Any` for the `Dict` by *explicitly* setting the types.

evm/chains/base.py
```diff
@@ -427,13 +456,13 @@ def get_block_by_hash(self, block_hash):
...
+                     genesis_state: Dict[Address, Dict]=None) -> 'BaseChain':
```
cburgdorf: Same here, can we get rid of the *implicit* `Any` for the `Dict` by *explicitly* setting the types.

evm/chains/base.py
```diff
@@ -427,13 +456,13 @@ def get_block_by_hash(self, block_hash):
...
         """
         Initializes the Chain from a genesis state.
         """
-        genesis_vm_class = cls.get_vm_class_for_block_number(0)
+        genesis_vm_class = cls.get_vm_class_for_block_number(cast(BlockNumber, 0))
```
cburgdorf: Don't cast here `Blocknumber(0)` should do it

evm/chains/base.py
```diff
@@ -560,15 +591,15 @@ def get_chain_at_block_parent(self, block):
         return type(self)(self.chaindb, init_header)
 
     @to_tuple
-    def get_ancestors(self, limit):
+    def get_ancestors(self, limit: int) -> Generator[BaseBlock, None, None]:
         lower_limit = max(self.header.block_number - limit, 0)
         for n in reversed(range(lower_limit, self.header.block_number)):
-            yield self.get_canonical_block_by_number(n)
+            yield self.get_canonical_block_by_number(cast(BlockNumber, n))
```
cburgdorf: Don't cast here `Blocknumber(n) ` should do it

evm/chains/base.py
```diff
@@ -142,29 +165,29 @@ def get_canonical_block_by_number(self, block_number):
...
+    def get_block_by_header(self, block_header: BlockHeader) -> BaseBlock:
         vm = self.get_vm(block_header)
         return vm.get_block_class().from_header(block_header, self.chaindb)
 
-    @to_tuple
-    def get_ancestors(self, limit):
-        lower_limit = max(self.header.block_number - limit, 0)
-        for n in reversed(range(lower_limit, self.header.block_number)):
-            yield self.get_canonical_block_by_number(n)
+    def get_ancestors(self, limit: int) -> Generator[BaseBlock, None, None]:
```
cburgdorf: Hard to follow up on mobile but are you sure we t don't want `Iterator[Baseblock] `?

p2p/peer.py
```diff
@@ -216,7 +218,9 @@ def __init__(self,
 
     @property
     async def genesis(self) -> BlockHeader:
-        genesis_hash = await self.chaindb.coro_lookup_block_hash(GENESIS_BLOCK_NUMBER)
+        genesis_hash = await self.chaindb.coro_lookup_block_hash(
+            cast(BlockNumber, GENESIS_BLOCK_NUMBER),
```
cburgdorf: Don't cast here `Blocknumber(GENESIS_BLOCK_NUMBER) ` should do it

cburgdorf:
>I am a little edgy about having to cast to block number everywhere

We can get rid of the casts. I addressed that in an inline comment.

gsalgado:
I'm not super comfortable with all the casting (either via mypy's `cast()` or by creating actual instances of the new `Hash32`/`BlockNumber` types as Christoph suggests) either. I know I'll forget it most of the time and also I'm pretty sure we'll end up with a mix of both `cast()` and object-instantiation. I'm not going to push back as I don't think those are a big deal, but thought it's worth bringing it up

pipermerriam:
> I know I'll forget it most of the time

Well, if you forget to do either it'll complain at your in the type check...  If we end up with a mix of `cast` and `TheType(value)` style code it doesn't seem like too big a deal.  More about whether the added `NewType` types add value (which I don't know if they do).

Moving forward with this and we can fully remove the `Hash32`, `Address` types quite easily at some point if we decide they aren't value adds.

cburgdorf:
>Well, if you forget to do either it'll complain at your in the type check

Correct! Except if one side is `Any` then it would not complain. Notice that this is a bit dangerous and one reason why we want to move away from implicit `Any` and enable a stricter mypy check.

Regarding casting: Casting is usually less type safe. In general, casting means to overrule type checking so I would *assume* the following to work and cause trouble at runtime.

`cast('foo', Blocknumber)`

In contrast the following should (rightfully) not work. 

`Blocknumber('foo')`

pipermerriam:
@cburgdorf it looks like neither `cast` or `BlockNumber(v)` provide much protection.

```python
In [1]: from typing import NewType, cast

In [2]: B = NewType('B', int)

In [3]: cast(B, 'arst')
Out[3]: 'arst'

In [4]: B('arst')
Out[4]: 'arst'
```

cburgdorf:
Odd. Worth following up when I get back.

carver:
But would a mypy run catch it?

pipermerriam:
> But would a mypy run catch it?

I'm pretty sure it wouldn't but not sure about that.

cburgdorf:
Ah wait, that wasn't mypy output of course. My comment was meant about mypy.

I assume mypy lets you cast anything to `Blocknumber` but enforces instantiation of `Blocknumber` to be `int`.

carver:
file:casting.py
```py
from typing import NewType, cast
B = NewType('B', int)
cast(B, 'arst')
B('arst')
```

shell:
```sh
$ mypy casting.py
casting.py:4: error: Argument 1 to "B" has incompatible type "str"; expected "int"`
```

Yup, mypy only catches the instantiation.

pipermerriam:
image

PR #661, opened 2018-05-07
Piper/remove validate block number

pipermerriam:
Builds on https://github.com/ethereum/py-evm/pull/660

### What was wrong?

The `validate_block_number` function doesn't actually do all of the proper validation for block numbers.

### How was it fixed?

Changed all usage of `validate_block_number` to instead use `validate_uint256`

#### Cute Animal Picture

il_570xn 511004122_hw7n

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -52,7 +52,6 @@ from evm.exceptions import (
     VMNotFound,
 )
 from evm.validation import (
-    validate_block_number,
     validate_uint256,
     validate_word,
     validate_vm_configuration,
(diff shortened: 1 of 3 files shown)


carver:
I'm finding it confusing to review a series of accumulated PRs. I think I already saw all these changes?

It would be nice to get the PRs isolated. But if they really have to depend on each other, maybe you could open a PR from one branch into another branch so we can see just the diff between each PR and focus on the relevant changes one PR at a time?  Then if they all get :+1: , you could merge them in reverse chronological order.

pipermerriam:
Fair, the dependency chain came from them touching the same sections of code and there being merge conflicts that I was trying to avoid.  I realize now that I'm having to do the same work rebasing them to clean up the diffs so... 

I've rebased this directly onto master.  Have a look?

pipermerriam:
So, upon further review of the yellow paper, I realize I am  wrong to remove/change this validation, as block numbers are not  constrained by any upper bound.

PR #662, opened 2018-05-07
rename lookup_block_hash

pipermerriam:
Builds on https://github.com/ethereum/py-evm/pull/661

### What was wrong?

The `ChainDB.lookup_block_hash` API was not named in a way that is consistent with the other APIs.

### How was it fixed?

Renamed this function to `get_canonical_block_hash`

#### Cute Animal Picture

il_570xn 511009912_gmub


evm/validation.py
```diff
@@ -197,16 +197,11 @@ def validate_unique(values, title="Value"):
...
-    validate_is_integer(block_number, title="Block Number")
-    validate_gte(block_number, 0, title="Block Number")
-
-
 def validate_vm_block_numbers(vm_block_numbers):
     validate_unique(vm_block_numbers, title="Block Number set")
 
     for block_number in vm_block_numbers:
-        validate_block_number(block_number)
+        validate_uint256(block_number)
```
carver: Isn't this going the opposite direction of the other type check that says block numbers are more special than just a uint? I'm not sure which one is better, but it seems like consistency is best.

evm/chains/base.py
```diff
@@ -338,19 +362,23 @@ def get_canonical_transaction(self, transaction_hash):
...
+                                    **kwargs: Any) -> BaseUnsignedTransaction:
         """
         Passthrough helper to the current VM class.
         """
         return self.get_vm().create_unsigned_transaction(*args, **kwargs)
 
-    def create_header_from_parent(self, parent_header, **header_params):
+    def create_header_from_parent(self,
+                                  parent_header: BlockHeader,
+                                  **header_params: Any) -> BlockHeader:
```
carver: *style nitpick* -- indentation

pipermerriam:
So, upon further review of the yellow paper, I realize I am wrong to remove/change this validation, as block numbers are not constrained by any upper bound.

evm/validation.py
pipermerriam: I think these are two separate issues that don't actually overlap by much.

- type hints let us say "this integer is a block number, this integer is not a block number, don't allow regular integers in places where we should be using block numbers"
- validation is for "a block number *should* conform to these properties"

Either way this PR is being closed without merging due to figuring out that block numbers are not constrained by any upper bound.

pipermerriam:
Oops, wrong PR.

PR #663, opened 2018-05-07
Cleanup to the Chain, ChainDB and VM classes and their base classes.

pipermerriam:
### What was wrong?

The `Chain`, `ChainDB` and `VM` API and classes were not as well organized as they could be.  

- Methods were not consistently grouped
- Methods were not consistently ordered
- Docstrings duplicated across base and implementation classes.
- No implementation class for `VM`

### How was it fixed?

- All `BaseXXX` classes are now pure abstract classes with no implementation details.
- Docstrings have been removed from `BaseXXX` classes with the exception of a few `BaseVM` methods which do not have default implementations in the main `VM` implementation class.
- Method ordering is now done by hierarchically sections (Header -> Block -> Transaction -> ...)
- Method ordering within a section is now alphabetical.
- Docstrings have been created for all methods on implementation classes (`Chain`, `ChaindB`, `VM`)
- Removed `evm.vm.vm` module in favor of locating `VM` class in `evm.vm.base` which is consistent with `evm.chain` and `evm.db.chain` organization.

I also added a `HeaderNotFound` exception for more precise exception handling.

#### Cute Animal Picture

cute-baby-chickens-wearing-knit-hats4

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -46,7 +46,7 @@ from evm.estimators import (
     get_gas_estimator,
 )
 from evm.exceptions import (
-    BlockNotFound,
+    HeaderNotFound,
     TransactionNotFound,
     ValidationError,
     VMNotFound,
@@ -102,7 +102,7 @@ class BaseChain(Configurable, metaclass=ABCMeta):
     chaindb = None  # type: BaseChainDB
 
     #
-    # Chain Initialization API
+    # Chain API
     #
     @classmethod
     @abstractmethod
@@ -110,9 +110,6 @@ class BaseChain(Configurable, metaclass=ABCMeta):
                      chaindb: BaseChainDB,
                      genesis_params: Dict[str, HeaderParams],
                      genesis_state: AccountState=None) -> 'BaseChain':
-        """
-        Initializes the Chain from a genesis state.
-        """
         raise NotImplementedError("Chain classes must implement this method")
 
     @classmethod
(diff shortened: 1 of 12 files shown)


evm/chains/base.py
```diff
@@ -387,23 +420,6 @@ def get_vm(self, header=None):
     #
     # Header/Block Retrieval
```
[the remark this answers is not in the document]
pipermerriam: I'm probably going to give this another full overhaul once #660 #661 and #662 are all merged.

pipermerriam:
@carver this is ready for round 2

evm/db/chain.py
```diff
@@ -250,15 +255,28 @@ def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
...
+        Returns an iterable of uncle headers specified by the given uncles_hash
+        """
+        validate_word(uncles_hash, title="Uncles Hash")
+        return rlp.decode(self.db.get(uncles_hash), sedes=rlp.sedes.CountableList(BlockHeader))
+
+    def get_score(self, block_hash: Hash32) -> int:
+        """
+        Returns the score for the header with the given hash.
+
+        Raises HeaderNotFound if no header with the given has is found in the database.
```
carver: Looks like this is missing the `HeaderNotFound` raise, which I suppose means that it's not tested.

evm/db/chain.py
```diff
@@ -250,15 +255,28 @@ def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
...
+            self.db.get(SchemaV1.make_block_hash_to_score_lookup_key(block_hash)),
+            sedes=rlp.sedes.big_endian_int)
 
-    # TODO: This method sould take a chain of headers as that's the most common use case
+    # TODO: This method should take a chain of headers as that's the most common use case
     # and it'd be much faster than inserting each header individually.
     def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
         """
-        :returns: iterable of headers newly on the canonical chain
+        Returns iterable of headers newly on the canonical chain
```
carver: Why drop the Sphinx `:return:` decorator?

evm/db/chain.py
```diff
@@ -250,15 +255,28 @@ def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
                 encode_hex(block_hash)))
```
carver: This should also be a `HeaderNotFound` right?

evm/db/chain.py
pipermerriam: Consistency.  A majority of the other methods weren't using this format so I decided to make them all the same.  I figure we can go through and convert these to *all* use sphinx formatting in a subsequent PR if that is how we're documenting things.

PR #666, opened 2018-05-08
p2p: Ensure correct and consistent use of CancelTokens

gsalgado:
In some places we were not using CancelTokens at the deepest possible level
of the call stack, and as a consequence we were cancelling tasks that had
created sub-tasks and when those sub-tasks were destroyed by the event loop
(because their parents had been cancelled), we were getting ERROR messages
in the logs
(https://docs.python.org/3/library/asyncio-dev.html#pending-task-destroyed)

This makes sure we use CancelTokens correctly and also adds some docs on
how to do so.

### DIFF
--- a/p2p/DEVELOPMENT.md
+++ b/p2p/DEVELOPMENT.md
@@ -1,0 +1,14 @@
+# Some guidelines about development in the p2p module
+
+The plan is for this to eventually become a comprehensive guide to aid when developing in the p2p
+package, but for now it's just a collection of random notes/recommendations.
+
+
+## Task cancellation
+
+In order to make sure we stop all pending asyncio tasks upon exit (or when a service terminates),
+we use `CancelToken`s, which are heavily inspired by https://vorpus.org/blog/timeouts-and-cancellation-for-humans/
+
+- A `CancelToken` must be available to all our async APIs. Either as an instance attribute or as an explicit argument.
+- When one of our async APIs `await` for stdlib/third-party coroutines, it must use `wait_with_token()` to ensure the scheduled task is cancelled when the token is triggered.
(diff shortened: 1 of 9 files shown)


jannikluhn:
I like the p2p development guidelines :+1:

Clarifying question about cancel tokens: Is there ever a good reason to use `wait_with_token` for non-stdlib/third party coroutines? My thinking is that every coroutine has a stdlib coroutine at the bottom (or is completely synchronous anyway), `wait_with_token` should be unnecessary at higher levels.

gsalgado:
That is correct, it should not be necessary at higher levels, but we still have some APIs that are not CancelToken-aware (e.g. the `AsyncChainDB.coro_*` family) and in those cases I think our best option (at this point) is to use `wait_with_token()` so that at least the outer tasks are cancelled when we exit

gsalgado:
> Regarding consistent use of cancel tokens: ChainSyncer seems to be special in that it takes a cancel token as initialization argument, whereas Peer, PeerPool and Server just create their own one. Should that be removed?

The tokens passed to `ChainSyncer` and `StateDownloader` are optional, and when they are passed they are chained with their internal token. We do this because they are used by `FullNodeSyncer`, which is a bit special as it doesn't run anything else -- it just waits for those to finish, in order. So this is to cause them to quit whenever the `FullNodeSyncer` is asked to stop.

As for the `stop()` method, that is supposed to be the public API for stopping a service; the cancel token is meant to be private. The only reason we use the cancel token in those `_test()` functions is because we need a sync callback to register as the signal handler. I'm working on a base class for those services to abstract some of this stuff and make those APIs more consistent; I'll ask for your feedback on it later today

PR #671, opened 2018-05-09
dirty

pipermerriam:
### What was wrong?

The light chain in trinity if using the `Chain` and `ChainDB` API for chain interactions.  This is problematic because light *mode* makes certain assumptions that may be fundamentally different from a full chain.

### How was it fixed?

Still a WIP

- Implementing a new `LightChain` and `LightDB` API which is looking to be a true subset of the main full chain APIs.
- Looks like these may turn into being called `HeaderChain` and `HeaderDB`.
- Looks like `Chain` and `ChainDB` may either:
  - inherit from `HeaderChain` and `HeaderDB` respectively 
  - use their own internal instances of `HeaderChain` and `HeaderDB` to delegate header related API calls to these instances (**preferred approach**)

#### Cute Animal Picture


evm/chains/header.py
```diff
@@ -0,0 +1,156 @@
...
+    @abstractmethod
+    def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
+        raise NotImplementedError("Chain classes must implement this method")
+
+    @abstractmethod
+    def import_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
+        raise NotImplementedError("Chain classes must implement this method")
+
+
+class HeaderChain(BaseHeaderChain):
```
carver: I don't understand why this class is reimplementing all these headerdb methods as pass-throughs. Can it just have the couple of extra methods it needs, and expose the headerdb? Just seeing if we can cut down on total code weight.

pipermerriam: I'm not sold this is the right way to go either, however, I was trying to not expose the database directly.  Going to think on this a bit and see if I can talk myself in-to/out-of it.  

I'm thinking about this in the same way that the `Chain` proxies some of the `VM` APIs.  I forsee (with casper) having the `VM` specify it's own class of `HeaderDB` at some point when the format of individual headers changes.

pipermerriam:
Merging so I can move forward from master, still want to discuss what the right approach is for exposing APIs on the databases.

evm/chains/header.py
carver: Still seems like we can cover that case by exposing the header db as a `BaseHeaderDB` from both `HeaderChain` implementations.

`Chain` & `VM` seems a bit more reasonable to duplicate the methods (though I'm not sure that's the right solution either), because the VM is different depending on the block. But `HeaderDB` is always the same for the given `HeaderChain`.

...unless we want to do `HeaderChain` -> `VM` by block number -> `HeaderDB`. But I don't think we need to store headers differently by fork, even if the header format changes, so that's probably unnecessary complexity.

PR #672, opened 2018-05-10
p2p: Correctly handle expected excpetions on received handshakes

gsalgado:
The Server.receive_handshake() handler was treating all exceptions other
than TimeoutError as unexpected, but that's not what happens in practice
-- there are several exceptions that we can get (from all different APIs
used in Server._receive_handshake())


p2p/server.py
```diff
@@ -166,10 +167,13 @@ def __init__(self,
 
     async def receive_handshake(
             self, reader: asyncio.StreamReader, writer: asyncio.StreamWriter) -> None:
+        expected_exceptions = (
```
pipermerriam: Maybe move these to a top level constant?

p2p/server.py
```diff
@@ -166,10 +167,13 @@ def __init__(self,
...
+            TimeoutError, PeerConnectionLost, HandshakeFailure, asyncio.IncompleteReadError,
+            ConnectionResetError, BrokenPipeError)
         try:
             await self._receive_handshake(reader, writer)
-        except TimeoutError:
-            self.logger.debug("Timeout waiting for handshake")
+        except expected_exceptions as e:
+            self.logger.debug("Could not complete handshake: %s", e)
```
pipermerriam: recently I learned that you can use `self.logger.debug("msg", exc_info=True)` and it will output a fully formatted stacktrace with the log output (if you are within an `except` block)

gsalgado: Nice, I'll use that

p2p/server.py
gsalgado: I'd rather keep them here as they're only expected when calling `_receive_handshake()`, and are not used anywhere else. Happy to move if you feel strong about that, though

pipermerriam: Motivation is:

- cutting out the overhead of building this tuple every time a handshake occurs.
- maybe increasing readability a bit by shortening function body.

I don't feel strongly as both of these are minor improvements.

PR #674, opened 2018-05-10
p2p: Implement a BaseService class and use it everywhere

gsalgado:
We had several classes for our different services and they all had
slightly different APIs for running/stopping them. BaseService aims
to provide a standard API for that.

Includes #673 , so better to review that one first

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -1,0 +1,73 @@
+from abc import ABC, abstractmethod
+import asyncio
+import logging  # noqa: F401
+from typing import Callable, Optional
+
+from p2p.cancel_token import CancelToken
+from p2p.exceptions import OperationCancelled
+
+
+class BaseService(ABC):
+    # Subclasses must define their own logger.
+    logger: logging.Logger = None
+    # Number of seconds cancel() will wait for run() to finish.
+    _wait_until_finished_timeout = 5
+
+    def __init__(self, token: CancelToken) -> None:
+        self.finished = asyncio.Event()
+        self.cancel_token = token
+
+    async def run(
+            self,
+            finished_callback: Optional[Callable[['BaseService'], None]] = None) -> None:
+        """Await for the service's _run() coroutine.
+
+        Once _run() returns, call cleanup(), set the finished event and call
+        finished_callback (if one was passed).
+        """
+        try:
+            await self._run()
+        except OperationCancelled as e:
+            self.logger.info("%s finished: %s", self, e)
+        except Exception:
+            self.logger.exception("Unexpected error in %s, exiting", self)
+        else:
+            self.logger.info("%s finished cleanly", self)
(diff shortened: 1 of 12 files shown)


p2p/service.py
```diff
@@ -0,0 +1,53 @@
...
+        try:
+            await self._run()
+        except OperationCancelled as e:
+            self.logger.info("%s finished: %s", self, e)
+        except Exception:
+            self.logger.exception("Unexpected error in %s, exiting", self)
+        finally:
+            self._finished.set()
+
+    async def stop(self) -> None:
```
jannikluhn: Should this raise an exception (or log a warning) when trying to stop a service that hasn't been started yet?

p2p/service.py
```diff
@@ -0,0 +1,53 @@
...
+        except asyncio.futures.TimeoutError:
+            self.logger.info("Timed out waiting for %s to finish, exiting anyway", self)
+
+    #
+    # These methods should be implemented by subclasses
+    #
+    async def _run(self) -> None:
+        """Runs the service's loop.
+
+        Should return when the CancelToken is triggered.
```
jannikluhn: I interpret this comment as `_run` should catch `OperationCancelled` exceptions and return cleanly in those cases. But would this be necessary considering that `run` catches them itself and logs a nice message as well?

gsalgado: The idea is that `_run()` can either return or let `OperationCancelled` bubble up, so I'll amend the docstring to make that clear

p2p/service.py
gsalgado: I don't think that adds any value at this point, and it'd require another event/attribute to track whether or not run() has been called, so I'd rather not do that now

p2p/service.py
```diff
@@ -0,0 +1,73 @@
...
+import logging  # noqa: F401
+from typing import Callable, Optional
+
+from p2p.cancel_token import CancelToken
+from p2p.exceptions import OperationCancelled
+
+
+class BaseService(ABC):
+    # Subclasses must define their own logger.
+    logger = None  # type: logging.Logger
```
pipermerriam: We can use python 3.6 syntax for these types.

```python
logger: logging.Logger = None
```

pipermerriam: I'm with you but I don't think it's fair to do it anytime in the near term future.  We *just* forced everyone to upgrade to 3.x and while `3.5 -> 3.6` is not a significant as `2.x -> 3.x` it's still not something we can just throw at the community imho.  I'm sure that there are people for whom the upgrade would require some non-trivial amount of infrastructure upgrades to get all of their ops infrastructure in a 3.6 environment.

cburgdorf: That would also allow ethpm to start with 3.6. But I can't say much about the community backing of such a move.

PR #676, opened 2018-05-10
Journal before trie

carver:
### What was wrong?

Right now, when an account changes multiple times within a transaction, then we:
- waste time calculating keccak on intermediate state
- waste time doing trie manipulation on intermediate state
- store unnecessary data for intermediate account values

### How was it fixed?

- Journal the raw account values, before entering them into the trie. 
- Replaced values get dropped in the journal, in between calling `persist`, so they never hit the trie

Right now, this will save state at every `persist` call. We can do even better, saving it only at the end of every block, but this PR was getting full.

Side note: The account cache started returning stale results, because the state_root is no longer updated at every change. I tried to fix it in place, with a sprinkling of cache invalidations, but it was taking too long to hunt down all the bugs, so I wrote a simpler `CacheDB` instead. It works, but I'm not sure of the performance impact yet.

Plus a little unrelated VM refactor for creating state more succinctly.

~_Edit: probably need some more persists in there somewhere... Will come back to it after some reviews_~

#### Cute Animal Picture

### DIFF
--- a/evm/db/account.py
+++ b/evm/db/account.py
@@ -5,7 +5,7 @@ from abc import (
 from uuid import UUID
 import logging
 from lru import LRU
-from typing import Tuple
+from typing import Set, Tuple  # noqa: F401
 
 import rlp
 
@@ -354,8 +354,10 @@ class AccountDB(BaseAccountDB):
     def persist(self) -> None:
         self._journaldb.persist()
         self.logger.debug("Persisting account values...")
-        accounts_displayed = set()
-        for checkpoint, accounts in reversed(self._journaltrie.journal.journal_data.items()):
+        accounts_displayed = set()  # type: Set[bytes]
+        queued_changes = self._journaltrie.journal.journal_data.items()
+        # mypy bug for ordered dict reversibility: https://github.com/python/typeshed/issues/2078
+        for checkpoint, accounts in reversed(queued_changes):  # type: ignore
             for address in accounts:
                 if address in accounts_displayed:
                     continue
--- a/evm/db/account.py
+++ b/evm/db/account.py
@@ -352,8 +352,11 @@ class AccountDB(BaseAccountDB):
         self._journaltrie.commit(trie_changeset)
 
     def persist(self) -> None:
+        self.logger.debug("Persisting AccountDB...")
         self._journaldb.persist()
-        self.logger.debug("Persisting account values...")
+        self._journaltrie.persist()
+
+    def _log_pending_accounts(self) -> None:
         accounts_displayed = set()  # type: Set[bytes]
         queued_changes = self._journaltrie.journal.journal_data.items()
         # mypy bug for ordered dict reversibility: https://github.com/python/typeshed/issues/2078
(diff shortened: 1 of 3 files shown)


evm/db/account.py
```diff
@@ -115,37 +118,33 @@ def account_is_empty(self, address):
...
     def __init__(self, db, state_root=BLANK_ROOT_HASH):
-        # Keep a reference to the original db instance to use it as part of _get_account()'s cache
-        # key.
-        self._unwrapped_db = db
-        self.db = JournalDB(db)
-        self._trie = HashTrie(HexaryTrie(self.db, state_root))
+        self._journaldb = JournalDB(db)
+        self._trie = HashTrie(HexaryTrie(db, state_root))
+        self._trie_cache = CacheDB(self._trie)
+        self._journaltrie = JournalDB(self._trie_cache)
```
pipermerriam: I could use comments/docstring/something here that explains what each of these is for.

carver: This is where I'm at for the docstring so far:

```
        r"""
        Internal implementation details (subject to rapid change): 
        Database entries go through several pipes, like so... 

        .. code:: 

                                                     -> hash-trie -> storage lookups
                                                    /
            db -----------------------> _journaldb ----------------> code lookups
             \
              > _trie -> _trie_cache -> _journaltrie --------------> account lookups

        Journaling sequesters writes here ^, until persist is called.

        _journaldb is a journaling of the keys and values used to store
        code and account storage.

        _trie is a hash-trie, used to generate the state root

        _trie_cache is a cache tied to the state root of the trie. It
        is important that this cache is checked *after* looking for
        the key in _journaltrie, because the cache is only invalidated
        after a state root change.

        _journaltrie is a journaling of the accounts (an address->rlp mapping,
        rather than the nodes stored by the trie). This enables
        a squashing of all account changes before pushing them into the trie.

        .. NOTE:: There is an opportunity to do something similar for storage

        AccountDB synchronizes the snapshot/revert/persist of both of the
        journals. 

        """
```

carver:
I wanted to be really sure that it wasn't the little refactor that broke it... and it was: https://github.com/ethereum/py-evm/pull/680

evm/db/account.py
```diff
@@ -301,24 +333,44 @@ def _get_account(self, address):
...
+                    accounts_displayed.add(address)
+                    account = self._get_account(address)
+                    self.logger.debug(
+                        "Saving Account %s: balance %d, nonce %d, storage root %s, code hash %s",
+                        encode_hex(address),
+                        account.balance,
+                        account.nonce,
+                        encode_hex(account.storage_root),
+                        encode_hex(account.code_hash),
+                    )
```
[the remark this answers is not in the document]
pipermerriam: If it was pure logging, then I'd be good with it, but the loop and whatnot doesn't sit well with me.  Maybe if you moved this outside the function body into a stand-alone and left it inline but commented out so that it's easy to enable as a one-liner but it doesn't clutter up the actual function body.

tests/p2p/test_state_sync.py
```diff
@@ -22,22 +23,23 @@ def make_random_state(n):
         code = b'not-real-code'
         account_db.set_code(addr, code)
         contents[addr] = (balance, nonce, storage, code)
-    return account_db, contents
+    account_db.persist()
+    return raw_db, account_db.state_root, contents
 
 
 def test_state_sync():
```
pipermerriam: This seems like a good candidate for using `hypothesis.strategies.random_module`

https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.random_module

You'll also need to convert the `make_random_state` function to use the `random` module for randomness generation.

gsalgado: I think using hypothesis won't buy us anything in those tests: https://github.com/ethereum/py-evm/pull/279#issuecomment-358521463

pipermerriam: Hrm, I think you are correct, though I also don't think it will cost us anything, and in the event that there is some *strange* randomness based test failure, it will be really nice to be able to reproduce it, which now, we wouldn't be able to do.

PR #681, opened 2018-05-11
WIP: HeaderChain and HeaderDB integration with p2p/trinity

pipermerriam:
### What was wrong?

WIP

#### Cute Animal Picture

### DIFF
--- a/evm/chains/header.py
+++ b/evm/chains/header.py
@@ -51,7 +51,7 @@ class BaseHeaderChain(Configurable, metaclass=ABCMeta):
         raise NotImplementedError("Chain classes must implement this method")
 
     #
-    # Header API
+    # Canonical Chain API
     #
     @abstractmethod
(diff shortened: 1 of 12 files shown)


pipermerriam:
I'm still in discovery mode on this it seems.  Probably a few pull requests will get extracted from this, primarily the conversion of `Peer` and other `p2p` classes to use `HeaderDB` instead of the full `ChainDB` as they only access the header data for the chain.

I also think I've got a handle on the inheritance structure for the main classes and then async versions of those classes used by trinity.

1. `BaseThing` (abstract, defines API)
2. `Thing` (full implementation)
3. `BaseAsyncThing` (abstract, defines `coro_` prefixed versions of `BaseThing` API
4. `AsyncThing` (currently largely `NotImplemented` errors in all `coro_` methods)  
5. `AsyncThingProxy` (multiprocessing proxy object which has full async and sync API).

pipermerriam:
closing as I'm off in a slightly different direction now.

Issue #667, opened 2018-05-08
Move `validate_seal` down to the VM class


pipermerriam:
### What is wrong?

The `validate_seal` method exists on the `Chain` class, however seal validation can and should be a `VM` level operation.

### How can it be fixed

Move the `validate_seal` method to the `VM` class.  The `VM` class should call out to this within it's `validate_block` method.

pipermerriam:
We can also consider moving the body of `validate_block` into individual methods to make `VM.validate_block` purely an orchestration layer with each piece of validation happening in their own individual methods.

glaksmono:
@pipermerriam would like to take a stab on this

cburgdorf:
Landed with #707 (thank you @darxtrix for the hint)

PR #682, opened 2018-05-12
[WIP] Move the validate_seal method to the VM class.

glaksmono:
### What was wrong?
The `validate_seal` method exists on the `Chain` class, however seal validation can and should be a `VM` level operation.

Issue Reference: https://github.com/ethereum/py-evm/issues/667

### How was it fixed?
Move the `validate_seal` method to the `VM` class. The `VM` class should call out to this within it's `validate_block` method.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -35,9 +35,6 @@ from eth_utils import (
 )
 
 from evm.db.chain import BaseChainDB
-from evm.consensus.pow import (
-    check_pow,
-)
 from evm.constants import (
     BLANK_ROOT_HASH,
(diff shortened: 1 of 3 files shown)


pipermerriam:
Test failure is due to this code here:

https://github.com/ethereum/py-evm/blob/master/tests/core/conftest.py#L105

It needs to be updated so that not only is the `Chain` class modified, but the actual `VM` class it uses needs to be modified as well.

```
SpuriousDragonVMForTesting = SpuriousDragonVM.configure(validate_seal=lambda self, block: None)
```

Something like that should work to bypass seal validation properly.

Issue #651, opened 2018-05-07
Use p2p.Server in trinity processes


gsalgado:
### What is wrong?

We're currently running the sync processes (i.e. `FullNodeSyncer` and `LightChain`) directly in trinity, together with the `PeerPool` required for them to work.

### How can it be fixed

We should teach `p2p.Server` how to select the appropriate sync service and then use just `p2p.Server` in trinity, as it takes care of setting up and running the `PeerPool` and the server that accepts incoming peer connections

PR #684, opened 2018-05-14
trinity: Use p2p.Server instead of running FullNodeSyncer directly

gsalgado:
With this, when we run trinity we'll also be listening for incoming peer
connections

Part of #651


pipermerriam:
#685 caused some merge conflicts.

PR #685, opened 2018-05-14
Piper/update headerdb api

pipermerriam:
### What was wrong?

Some updates to the API for working with the `HeaderChain` and `HeaderDB`.

### How was it fixed?

- The test suite for the `HeaderChain` class has been slimmed down to reduce duplication in what we're testing.  Now the passthrough methods simply do a sanity check that the API isn't fundamentally broken.
- Added `header_exists` to both `HeaderChain` and `HeaderDB`.  This is needed for upcoming conversion of other parts of py-evm and trinity to use these APIs
- Now the `header` argument is optional when creating a `HeaderChain` instance.  When not provided, it's looked up from the database.

#### Cute Animal Picture

cute-baby-chickens-wearing-knit-hats5

### DIFF
--- a/tests/database/test_header_db.py
+++ b/tests/database/test_header_db.py
@@ -242,3 +242,45 @@ def test_headerdb_canonical_header_retrieval_by_number(headerdb, genesis_header)
     for header in headers:
         actual = headerdb.get_canonical_block_header_by_number(header.block_number)
         assert_headers_eq(actual, header)
+
+
+def test_headerdb_header_exists(headerdb, genesis_header):
+    assert headerdb.header_exists(genesis_header.hash) is False
+    headerdb.persist_header(genesis_header)
+    assert headerdb.header_exists(genesis_header.hash) is True
+
+    chain_a = mk_header_chain(genesis_header, 3)
+    chain_b = mk_header_chain(genesis_header, 5)
+
+    assert not any(headerdb.header_exists(h.hash) for h in chain_a)
+    assert not any(headerdb.header_exists(h.hash) for h in chain_b)
+
+    for idx, header in enumerate(chain_a):
+        assert all(headerdb.header_exists(h.hash) for h in chain_a[:idx])
+        assert not any(headerdb.header_exists(h.hash) for h in chain_a[idx:])
+
+        # sanity pre-check
+        assert not headerdb.header_exists(header.hash)
+        headerdb.persist_header(header)
+        assert headerdb.header_exists(header.hash)
+
+    # `chain_a` should now all exist
+    assert all(headerdb.header_exists(h.hash) for h in chain_a)
+    # `chain_b` should not be in the database.
+    assert not any(headerdb.header_exists(h.hash) for h in chain_b)
+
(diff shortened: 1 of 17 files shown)


tests/database/test_header_db.py
```diff
@@ -242,3 +242,45 @@ def test_headerdb_canonical_header_retrieval_by_number(headerdb, genesis_header)
...
+
+    assert not any(headerdb.header_exists(h.hash) for h in chain_a)
+    assert not any(headerdb.header_exists(h.hash) for h in chain_b)
+
+    for idx, header in enumerate(chain_a):
+        assert all(headerdb.header_exists(h.hash) for h in chain_a[:idx])
+        assert not any(headerdb.header_exists(h.hash) for h in chain_a[idx:])
+
+        # sanity pre-check
+        assert not headerdb.header_exists(header.hash)
```
carver: This seems over-cautious, since it is tested in 260 as well. Could be dropped.

tests/database/test_header_db.py
```diff
@@ -242,3 +242,45 @@ def test_headerdb_canonical_header_retrieval_by_number(headerdb, genesis_header)
...
+        headerdb.persist_header(header)
+        assert headerdb.header_exists(header.hash)
```
carver: Could also be dropped.

tests/database/test_header_db.py
```diff
@@ -242,3 +242,45 @@ def test_headerdb_canonical_header_retrieval_by_number(headerdb, genesis_header)
...
+        # `chain_a` should remain accessible
+        assert all(headerdb.header_exists(h.hash) for h in chain_a)
+
+        assert all(headerdb.header_exists(h.hash) for h in chain_b[:idx])
+        assert not any(headerdb.header_exists(h.hash) for h in chain_b[idx:])
+
+        # sanity pre-check
+        assert not headerdb.header_exists(header.hash)
+        headerdb.persist_header(header)
+        assert headerdb.header_exists(header.hash)
```
carver: Same with these two asserts.

tests/database/test_header_db.py
pipermerriam: I'm inclined to leave it in because while they overlap, they test two distinctly different things.

- line 260 and 268) are testing properties about the chain of headers leading up to and following this header (and they happen to include this header in those checks)
- this line is checking that the specific header we are processing was correctly persisted.

Overhead should be mild so if you don't object....

tests/database/test_header_db.py
pipermerriam: I'm wishing to leave them in for the same reason as the previous section.  Veto if you feel strongly.

tests/database/test_header_db.py
carver: Eh, I guess not. When reading it, the duplication made me concerned that I misunderstood the earlier assertion and go back to re-read it. So it adds overhead to the read-through.

I guess a comment like `# rerun the existence test for this specific header` would have saved me the re-read.

PR #690, opened 2018-05-14
Pypy support

carver:
### What was wrong?

pypy was not being tested in CI. There were a few little incompatibilities left.

### How was it fixed?

- add pypy3 tests
- remove pysha3 from p2p
- use eth-hash's `new()` to get an object that can be updated sequentially
- upgrade eth-hash so that we can call `update()` after `digest()`

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,7 +29,7 @@ common: &common
         command: pip install --user tox
     - run:
         name: install leveldb build tools for pypy
-        command: if [[ "$CIRCLE_JOB" == *"pypy3"* ]]; then apt install libleveldb-dev; fi
+        command: if [[ "$CIRCLE_JOB" == *"pypy3"* ]]; then apt update && apt install libleveldb-dev; fi
     - run:
         name: run tox
         command: ~/.local/bin/tox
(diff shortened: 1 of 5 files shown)


pipermerriam:
Can you add the conditional `toolz/cytoolz` install pattern too?

pipermerriam:
For now, what do you think about only testing `p2p/trinity` with pypy?

carver:
Pro testing more tox envs:
- find out faster if a pypy3 breakage slips in
Pro testing fewer tox envs:
- PRs can be merged faster, and generally give faster feedback

I think -- *at least for now* -- that the tradeoff should go the direction of quicker iteration at the cost of maybe missing a pypy breakage. Maybe that should change as trinity gets more stable.

---

Update: Leveldb is causing headaches. Looks like we would have to build a custom docker container, because the default one doesn't support a recent enough version of `libleveldb-dev` (1.20 according to [this issue](https://github.com/wbolster/plyvel/issues/62)). This may be a good idea, as would be building a fully compiled version of trinity in docker. But I've already spent more time on this than I intended. So I'll pause and return later.

ninjaahhh:
since pypy3 only supports python3.5, how about some language features currently used in this repo like variable annotations?

carver:
This PR is super old. We will probably drop py3.5 support soon, sooner than pypy will add py3.6 support. So I'll close this until it makes sense to resurrect.

PR #699, opened 2018-05-15
Keymap db

carver:
### What was wrong?

A few other places use a key mapping DB (like Storage). Don't want to rewrite a new one like `HashTrie` each time.

### How was it fixed?

Added a generic `KeyMapDB`, plus a `HashTrie` alias that uses `keccak` as the key mapper.

#### Cute Animal Picture

### DIFF
--- a/evm/db/hash_trie.py
+++ b/evm/db/hash_trie.py
@@ -1,28 +1,10 @@
+from functools import partial
+
 from eth_hash.auto import keccak
 
+from evm.db.keymap import (
+    KeyMapDB,
+)
 
-class HashTrie(object):
-    _trie = None
 
-    def __init__(self, trie):
-        self._trie = trie
-
-    def __setitem__(self, key, value):
-        self._trie[keccak(key)] = value
-
-    def __getitem__(self, key):
-        return self._trie[keccak(key)]
-
-    def __delitem__(self, key):
-        del self._trie[keccak(key)]
-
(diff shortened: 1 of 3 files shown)


evm/db/hash_trie.py
```diff
@@ -1,28 +1,10 @@
...
+)
 
-    @property
-    def root_hash(self):
-        return self._trie.root_hash
 
-    @root_hash.setter
-    def root_hash(self, value):
-        self._trie.root_hash = value
+HashTrie = partial(KeyMapDB, keccak)
```
pipermerriam: This is probably going to be problematic when it comes to applying type hints.  Should probably figure out a way to do this with proper subclassing.

carver: I guess this should do it:

```py
class HashTrie(KeyMapDB):
    def __init__(self, db):
        super().__init__(keccak, db)
```

pipermerriam: I'd do it as a class attribute.

```python
class KeyMapDB:
    @staticmethod
    @abstractmethod
    def key_fn(value: bytes) -> bytes:
        raise NotImplementedError

class HashTrie(KeyMapDB):
    key_fn = staticmethod(keccak)
```

carver: Huh, mypy complained about wrapping `keccak` in `staticmethod()`. Works fine without it.

Issue #353, opened 2018-02-05
Require parameters to be passed as keyword arguments for long funciton signatures.


pipermerriam:
### What is wrong?

There are lots of places in the codebase with either classes or functions which take 6+ parameters.  In these cases, calling these using positional arguments can be both error prone as well as hard to read.

- Errors arise when arguments are accidentally called in the wrong order.
- Difficult readability comes when it's not clear what each parameter actually means since sometimes multiple parameters might be called with the same constant value `f(b'', b'')`.

### How can it be fixed

https://www.python.org/dev/peps/pep-3102/

PEP3102 introduces a new syntax for requiring certain arguments to be called as keyword arguments.  I propose changing as many of the functions as seems appropriate to this format.

thedadams:
A quick `grep` search shows that there are about 7 places where functions are defined or called with 6+ parameters. Is this what was expected here?

palash25:
This is what I got from a `git grep`. I didn't find 7 but only six functions with 6 or more arguments

```
evm/vm/logic/call.py
272:def compute_eip150_msg_gas(computation, gas, extra_gas, value, mnemonic, callstipend):

tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py
131:def validate_account_attribute(fixture_key, rpc_method, rpc, state, addr, at_block):

evm/vm/forks/byzantium/transactions.py
13:    def create_unsigned_transaction(cls, nonce, gas_price, gas, to, value, data):

evm/vm/forks/frontier/transactions.py
79:    def create_unsigned_transaction(cls, nonce, gas_price, gas, to, value, data):

evm/vm/forks/homestead/transactions.py
43:    def create_unsigned_transaction(cls, nonce, gas_price, gas, to, value, data):

evm/vm/forks/spurious_dragon/transactions.py
35:    def create_unsigned_transaction(cls, nonce, gas_price, gas, to, value, data):

```

Are these the functions that you had in mind @pipermerriam ? Can I work on this?

pipermerriam:
Yes, these types of call sites which have long parameter lists.  👍

Bhargavasomu:
@pipermerriam this issue ideally needs the function signature to be changed and also the corrections of wherever the functions are called right?

pipermerriam:
Yes, it would involve changing the signature to be something like:

```python
def create_unsigned_transaction(cls, *, nonce, gas_price, gas, to, value, data):
    ...
```

[...]

PR #700, opened 2018-05-15
switched long function signatures to keyword arguments

elihanover:


pipermerriam:
@elihanover please read the issue here: https://github.com/ethereum/py-evm/issues/353

This PR should not change argument defaults or introduce new argument defaults.  Rather it should use the syntax from PEP3102 to require arguments to be passed in as keyword args.

pipermerriam:
Like this:

```python
def my_fn((*, arg_a, arg_b):
    print(arg_a, arg_b)
```

This will require `my_fn` to be called like `my_fn(arg_a=1, arg_b=2)` and will disallow calling using those as positional arguments.

elihanover:
@pipermerriam understood, thanks for the clarification

PR #704, opened 2018-05-16
Squash all trie changes, persisting minimally

carver:
### What was wrong?

We are storing state roots into the database for every step in the transaction. We only need to store the trie at the end of the block.

### How was it fixed?

Added an extra batch db in AccountDB, so that we can calculate the state root without persisting, and only persist the new state root when absolutely necessary.

Bonus bugfix: `BatchDB` had a bug when it tried to skip deletes.

#### Cute Animal Picture

### DIFF
--- a/evm/db/account.py
+++ b/evm/db/account.py
@@ -71,7 +71,7 @@ class BaseAccountDB(metaclass=ABCMeta):
         raise NotImplementedError("Must be implemented by subclasses")
 
     @abstractmethod
-    def has_root(self):
+    def has_root(self, state_root: bytes) -> bool:
         raise NotImplementedError("Must be implemented by subclasses")
 
     #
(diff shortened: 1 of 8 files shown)


carver:
@gsalgado I'm hoping to see a smaller db, with this change

evm/db/account.py
```diff
@@ -65,6 +70,10 @@ def __init__(self) -> None:
     def state_root(self):
         raise NotImplementedError("Must be implemented by subclasses")
 
+    @abstractmethod
+    def has_root(self):
```
pipermerriam: This method signature doesn't match the implementation here: https://github.com/ethereum/py-evm/pull/704/files#diff-05a5232d6f5def53d9b58b9811d78282R218

Shouldn't this be: `def has_root(self, state_root: bytes) -> True:`?

> note I added type hints.

pipermerriam:
> I'm hoping to see a smaller db, with this change

@carver I wonder if we could add a benchmark for this in #592

PR #706, opened 2018-05-16
p2p: Server now listens on all network interfaces

### LINKED ISSUE #683: p2p.Server should listen on all interfaces
Like geth does, we should listen on all interfaces, so users only have to specify the port (defaulting to 30303) on which to listen. When doing that we'll also need to detect the internal IP address to use when setting up the NAT port mapping. Something like https://gist.github.com/gsalgado/821bec709ec42d545a87999341f85f31

### LINKED ISSUE #703: p2p tests have unclean finish
### What is wrong?

p2p tests end with a bunch of repeats of this:

```
Error 'HTTPConnectionPool(host='192.168.1.250', port=49152): Max retries exceeded with url: /description.xml (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f256f014b00>: Failed to establish a new connection: [Errno 113] No route to host',))' for http://192.168.1.250:49152/description.xml
Error '404 Client Error: Not Found for url: http://192.168.2.44:8008/ssdp/notfound' for http://192.168.2.44:8008/ssdp/device-desc.xml
exception calling callback for <Future at 0x7f256e86dac8 state=finished returned list>
  File "/usr/lib/python3.6/concurrent/futures/_base.py", line 324, in _invoke_callbacks
    callback(self)
  File "/usr/lib/python3.6/asyncio/futures.py", line 419, in _call_set_state
    dest_loop.call_soon_threadsafe(_set_state, destination, source)
  File "/usr/lib/python3.6/asyncio/base_events.py", line 620, in call_soon_threadsafe
    self._check_closed()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 357, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
```

### How can it be fixed

:man_shrugging:

gsalgado:
Closes #683
Closes #703

### DIFF
--- a/p2p/server.py
+++ b/p2p/server.py
@@ -1,12 +1,16 @@
 import asyncio
 from concurrent.futures import ThreadPoolExecutor
+import ipaddress
 import logging
 import secrets
+import socket
 from typing import (
+    cast,
     List,
     Type,
     TYPE_CHECKING,
 )
+from urllib.parse import urlparse
 
 import netifaces
 import upnpclient
@@ -30,7 +34,6 @@ from p2p.auth import (
     HandshakeResponder,
 )
 from p2p.cancel_token import (
-    CancelToken,
     wait_with_token,
 )
 from p2p.constants import (
@@ -67,11 +70,13 @@ if TYPE_CHECKING:
 class Server(BaseService):
     """Server listening for incoming connections"""
     logger = logging.getLogger("p2p.server.Server")
-    _server = None
+    _tcp_listener = None
+    _udp_listener = None
+    _nat_portmap_lifetime = 30 * 60
 
     def __init__(self,
                  privkey: datatypes.PrivateKey,
-                 address: Address,
+                 port: int,
                  chain: AsyncChain,
                  chaindb: AsyncChainDB,
                  headerdb: 'BaseAsyncHeaderDB',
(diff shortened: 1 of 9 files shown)


p2p/server.py
```diff
@@ -82,136 +85,187 @@ def __init__(self,
...
+                self.logger.exception("Failed to setup NAT portmap")
+
+    def _find_internal_ip_on_device_network(self, upnp_dev: upnpclient.upnp.Device) -> str:
+        for iface in netifaces.interfaces():
+            for family, addresses in netifaces.ifaddresses(iface).items():
+                # TODO: Support IPv6 addresses as well.
+                if family != netifaces.AF_INET:
+                    continue
+                for item in addresses:
+                    network = item['addr'].rsplit('.', 1)[0]
```
pipermerriam: This line is really cryptic to me.  Mostly just curious about what format this string is in, what we're pulling out of it, and whether we can use existing parsing tools to do it.

p2p/server.py
```diff
@@ -82,136 +85,187 @@ def __init__(self,
...
+    async def add_nat_portmap(self) -> None:
+        self.logger.info("Setting up NAT portmap...")
+        # This is experimental and it's OK if it fails, hence the bare except.
+        try:
+            upnp_dev = await self._discover_upnp_device()
+            if upnp_dev is None:
+                return
+            await self._add_nat_portmap(upnp_dev)
+        except Exception as e:
+            if (isinstance(e, upnpclient.soap.SOAPError) and
```
pipermerriam: This seems like it'd be cleaner as it's own except block even if it involves having the `self.logger.exception` section in both `except` blocks.

p2p/server.py
```diff
@@ -82,136 +85,187 @@ def __init__(self,
...
+    async def _start_udp_listener(self, discovery: DiscoveryProtocol) -> None:
+        loop = asyncio.get_event_loop()
+        # TODO: Support IPv6 addresses as well.
+        self._udp_transport, _ = await loop.create_datagram_endpoint(
+            lambda: discovery,
+            local_addr=('0.0.0.0', self.port),
+            family=socket.AF_INET)
+
+    async def _close_udp_listener(self) -> None:
+        cast(asyncio.DatagramTransport, self._udp_transport).abort()
```
pipermerriam: Why is the `cast` needed here?  Do you think it'd be a good practice for us to document all uses of `cast` so it's clear why they are needed?

p2p/server.py
```diff
@@ -82,136 +85,187 @@ def __init__(self,
...
+
     async def _close(self) -> None:
-        self._server.close()
-        await self._server.wait_closed()
+        await self._close_tcp_listener()
+        await self._close_udp_listener()
```
pipermerriam: Could these two lines be run in parallel with [`asyncio.gather`](https://docs.python.org/3.5/library/asyncio-task.html#asyncio.gather)

p2p/server.py
```diff
@@ -305,6 +359,10 @@ def __init__(self,
     async def do_handshake(self, peer: BasePeer) -> None:
         await peer.do_p2p_handshake(),
         await peer.do_sub_proto_handshake()
+        self._start_peer(peer)
+
+    def _start_peer(self, peer: BasePeer) -> None:
```
pipermerriam: It seems like we *should* be able to get rid of this by instead using a `PeerPool` class in the tests who's `start_peer` method was overridden.

p2p/server.py
gsalgado: Yep, I don't see why not!

p2p/server.py
gsalgado: Indeed, but for that I'd need to change `get_server()` and `receiver_server_with_dumb_peer()` to use that mock `PeerPool`. I'd prefer to keep it as it is since I intend to rewrite the tests in `test_server.py`, and then I should be able to get rid of this

p2p/server.py
gsalgado: Because `self._udp_transport` gets the return value of `asyncio.create_datagram_endpoint()`, which is a `BaseTransport`. I don't like the casts because they make the code less readable, but not sure what we would gain by documenting them...

p2p/server.py
gsalgado: Sounds good to me

p2p/server.py
pipermerriam: Exactly this.  Documenting why the type system isn't working for us in these situations.

PR #707, opened 2018-05-16
Move the validate_seal method to the VM class.

glaksmono:
### What was wrong?
The `validate_seal` method exists on the `Chain` class, however seal validation can and should be a `VM` level operation.

Issue Reference: https://github.com/ethereum/py-evm/issues/667

### How was it fixed?
Move the `validate_seal` method to the `VM` class. The `VM` class should call out to this within it's `validate_block` method.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -39,9 +39,6 @@ from evm.db.chain import (
     BaseChainDB,
     ChainDB,
 )
-from evm.consensus.pow import (
-    check_pow,
-)
 from evm.constants import (
     BLANK_ROOT_HASH,
     MAX_UNCLE_DEPTH,
@@ -233,10 +230,6 @@ class BaseChain(Configurable, metaclass=ABCMeta):
 
     @abstractmethod
     def validate_gaslimit(self, header: BlockHeader) -> None:
-        raise NotImplementedError("Chain classes must implement this method")
-
-    @abstractmethod
-    def validate_seal(self, header: BlockHeader) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
@@ -589,10 +582,7 @@ class Chain(BaseChain):
         Since block validation (specifically the uncle validation must have
         access to the ancestor blocks, this validation must occur at the Chain
         level.
-
-        TODO: move the `seal` validation down into the vm.
         """
-        self.validate_seal(block.header)
         self.validate_uncles(block)
         self.validate_gaslimit(block.header)
 
@@ -610,14 +600,6 @@ class Chain(BaseChain):
             raise ValidationError(
                 "The gas limit on block {0} is too high: {1}. It must be at most {2}".format(
                     encode_hex(header.hash), header.gas_limit, high_bound))
-
-    def validate_seal(self, header: BlockHeader) -> None:
-        """
-        Validate the seal on the given header.
-        """
-        check_pow(
-            header.block_number, header.mining_hash,
-            header.mix_hash, header.nonce, header.difficulty)
 
     def validate_uncles(self, block: BaseBlock) -> None:
         """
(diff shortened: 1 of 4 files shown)


evm/vm/base.py
```diff
@@ -674,10 +681,20 @@ def validate_block(self, block):
...
+        """
+        check_pow(
+            header.block_number, header.mining_hash,
+            header.mix_hash, header.nonce, header.difficulty)
+
     def validate_uncle(self, block, uncle):
         """
         Validate the given uncle in the context of the given block.
         """
+        self.validate_seal(block.header)
```
carver: I pushed an update to the test so we can see the stack trace that failed the block. It points to this line. Maybe there are some paths that call `validate_uncle` before this block's mixhash has been finalized?

glaksmono: Tried to execute the `pytest` individually per your suggestion and I'm getting this: https://gist.github.com/glaksmono/8c273fe4bdd01d2396f95ae36df29794 - thoughts?

glaksmono:
```
...
  /Users/glaksmono/Documents/py-evm/venv/lib/python3.6/site-packages/pluggy/callers.py(180)_multicall()
-> res = hook_impl.function(*args)
  /Users/glaksmono/Documents/py-evm/venv/lib/python3.6/site-packages/_pytest/python.py(156)pytest_pyfunc_call()
-> testfunction(**testargs)
  /Users/glaksmono/Documents/py-evm/tests/json-fixtures/test_blockchain.py(103)test_blockchain_fixtures()
-> (block, mined_block, block_rlp) = apply_fixture_block_to_chain(block_fixture, chain)
  /Users/glaksmono/Documents/py-evm/evm/tools/fixture_tests.py(632)apply_fixture_block_to_chain()
-> mined_block = chain.import_block(block)
  /Users/glaksmono/Documents/py-evm/evm/chains/base.py(546)import_block()
-> imported_block = parent_chain.get_vm().import_block(block)
  /Users/glaksmono/Documents/py-evm/evm/vm/base.py(428)import_block()
-> return self.mine_block()
  /Users/glaksmono/Documents/py-evm/evm/vm/base.py(434)mine_block()
-> block = self.pack_block(self.block, *args, **kwargs)
  /Users/glaksmono/Documents/py-evm/evm/vm/base.py(528)pack_block()
-> self.validate_block(packed_block)
  /Users/glaksmono/Documents/py-evm/evm/vm/base.py(661)validate_block()
-> self.validate_uncle(block, uncle)
  /Users/glaksmono/Documents/py-evm/evm/vm/base.py(695)validate_uncle()
-> self.validate_seal(block.header)
  /Users/glaksmono/Documents/py-evm/evm/vm/base.py(689)validate_seal()
-> header.mix_hash, header.nonce, header.difficulty)
> /Users/glaksmono/Documents/py-evm/evm/consensus/pow.py(65)check_pow()
-> encode_hex(mining_output[b'mix digest']), encode_hex(mix_hash)))
```
So the reason why `validate_uncle` is called prior to the `validate_seal` here is because I put `validate_seal` within `validate_uncle`

But that is the [first](https://github.com/glaksmono/py-evm/blob/021d244d86a6c695720425e3b7d83ff953fca947/evm/vm/base.py#L695) thing to be executed in `validate_uncle`

Prior to this is all "python related stuff"

evm/chains/base.py
```diff
@@ -589,10 +582,7 @@ def validate_block(self, block: BaseBlock) -> None:
         Since block validation (specifically the uncle validation must have
         access to the ancestor blocks, this validation must occur at the Chain
         level.
-
-        TODO: move the `seal` validation down into the vm.
         """
-        self.validate_seal(block.header)
```
carver: Maybe do a `vm.validate_seal()` here instead of trying to merge it into the uncle check?

evm/vm/base.py
carver: >  I put validate_seal(block.header) within validate_uncle

This change repeatedly checks the block header for each uncle. There should be a separate check anyway, in case there are 0 uncles. What's the benefit of putting it there?

evm/chains/base.py
glaksmono: Done

PR #714, opened 2018-05-17
p2p: Improve sharding tests

gsalgado:
- No longer use/run a p2p.Server
- No longer run a PeerPool nor access its private bits
- Use peer with directly connected reader/writers, avoiding going
  through the network

### DIFF
--- a/tests/p2p/test_sharding.py
+++ b/tests/p2p/test_sharding.py
@@ -1,29 +1,16 @@
+import asyncio
+
 import pytest
 
-import asyncio
-import collections
-
-from eth_keys import keys
-from eth_utils import (
-    int_to_big_endian,
-)
-from evm.utils.padding import (
-    pad32,
-)
-
-from p2p.cancel_token import (
-    CancelToken,
-)
-from p2p.kademlia import (
-    Address,
-    Node,
-)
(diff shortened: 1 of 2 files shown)


tests/p2p/test_sharding.py
```diff
@@ -145,75 +130,47 @@ class InvalidShardingPeer(ShardingPeer):
...
+            if syncer != proposer
+        ]), timeout=2)
+        for syncer in syncers:
+            assert syncer.shard.get_collation_by_hash(collation.hash) == collation
 
     # stop everything
-    cancel_token.trigger()
-    await asyncio.gather(*[peer_tuple.server.cancel() for peer_tuple in peer_tuples])
-    await asyncio.gather(*[peer_tuple.syncer.cancel() for peer_tuple in peer_tuples])
+    all_peers = itertools.chain(
```
pipermerriam: should this be in a `finally` block to ensure it runs?

gsalgado: Actually, the one for peers is unnecessary as `get_directly_linked_peers()` already sets up a pytest finalizer that does that. I'll change the one that cancels syncers to use a finalizer as well

PR #717, opened 2018-05-18
Enable ipc queries to light chain

carver:
### What was wrong?

Want to use IPC to query the light chain, starting from trinity command line.

### How was it fixed?

Still very much in progress, but this shows the number of the latest downloaded header:

```sh
$ trinity --ropsten --light --data-dir /tmp/trinity-light-ropsten/
... <logs> ...
```
```py
from web3 import Web3
w3t = Web3(Web3.IPCProvider('/tmp/trinity-light-ropsten/jsonrpc.ipc'))
w3t.eth.blockNumber
```
I've been hacking away to get `w3t.eth.getBlock('latest')` working, then I'll pinch this PR off.

Some outstanding questions:
- What is the best way to interface with `LightPeerChain` (formerly `p2p.lightchain.LightChain`) from a synchronous interface. I tried to avoid asyncio spreading virally through the code base. But still getting errors about futures on different threads, etc, and I'm new enough to asyncio that a *good* solution isn't obvious to me.
- Which things should be running in their own threads?
- Some modules are split by network (mainnet/ropsten/etc), others split by node type (full/light/sharding?/stateless?). Which makes more sense? (all the ropsten variants in one file, or all the full nodes in one file?)
- Surely a lot more.

#### Cute Animal Picture

### DIFF
--- a/evm/db/header.py
+++ b/evm/db/header.py
@@ -132,10 +132,12 @@ class HeaderDB(BaseHeaderDB):
         return rlp.decode(header_rlp, BlockHeader)
 
     def get_score(self, block_hash: Hash32) -> int:
-        return rlp.decode(
-            self.db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)],
-            sedes=rlp.sedes.big_endian_int,
-        )
+        try:
+            score = self.db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)]
+        except KeyError:
+            raise HeaderNotFound("No header with hash {0} found".format(
+                encode_hex(block_hash)))
+        return rlp.decode(score, sedes=rlp.sedes.big_endian_int)
 
     def header_exists(self, block_hash: Hash32) -> bool:
         validate_word(block_hash, title="Block Hash")
--- a/evm/db/header.py
+++ b/evm/db/header.py
@@ -133,11 +133,11 @@ class HeaderDB(BaseHeaderDB):
 
     def get_score(self, block_hash: Hash32) -> int:
         try:
-            score = self.db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)]
+            encoded_score = self.db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)]
         except KeyError:
             raise HeaderNotFound("No header with hash {0} found".format(
                 encode_hex(block_hash)))
-        return rlp.decode(score, sedes=rlp.sedes.big_endian_int)
+        return rlp.decode(encoded_score, sedes=rlp.sedes.big_endian_int)
 
     def header_exists(self, block_hash: Hash32) -> bool:
         validate_word(block_hash, title="Block Hash")
--- a/p2p/cancel_token.py
+++ b/p2p/cancel_token.py
@@ -6,10 +6,11 @@ from p2p.exceptions import OperationCancelled
 
 class CancelToken:
 
-    def __init__(self, name: str) -> None:
+    def __init__(self, name: str, loop: asyncio.AbstractEventLoop = None) -> None:
         self.name = name
         self._chain = []  # type: List['CancelToken']
-        self._triggered = asyncio.Event()
+        self._triggered = asyncio.Event(loop=loop)
+        self._loop = loop
 
     def chain(self, token: 'CancelToken') -> 'CancelToken':
         """Return a new CancelToken chaining this and the given token.
@@ -19,7 +20,7 @@ class CancelToken:
         has no effect on either of the chained tokens.
         """
         chain_name = ":".join([self.name, token.name])
-        chain = CancelToken(chain_name)
+        chain = CancelToken(chain_name, loop=self._loop)
         chain._chain.extend([self, token])
         return chain
 
@@ -52,16 +53,16 @@ class CancelToken:
         if self.triggered_token is not None:
             return
 
-        futures = [asyncio.ensure_future(self._triggered.wait())]
+        futures = [asyncio.ensure_future(self._triggered.wait(), loop=self._loop)]
         for token in self._chain:
-            futures.append(asyncio.ensure_future(token.wait()))
+            futures.append(asyncio.ensure_future(token.wait(), loop=self._loop))
 
         def cancel_pending(f):
             for future in futures:
                 if not future.done():
                     future.cancel()
 
-        fut = asyncio.ensure_future(_wait_for_first(futures))
+        fut = asyncio.ensure_future(_wait_for_first(futures), loop=self._loop)
         fut.add_done_callback(cancel_pending)
         await fut
 
--- a/p2p/cancel_token.py
+++ b/p2p/cancel_token.py
@@ -1,7 +1,10 @@
 import asyncio
 from typing import Any, Awaitable, cast, List
 
-from p2p.exceptions import OperationCancelled
+from p2p.exceptions import (
+    EventLoopMismatch,
+    OperationCancelled,
+)
 
 
 class CancelToken:
@@ -19,6 +22,8 @@ class CancelToken:
         called on either of the chained tokens, but calling trigger() on the new token
         has no effect on either of the chained tokens.
         """
+        if self._loop != token._loop:
+            raise EventLoopMismatch("Chained CancelToken objects must be on the same event loop")
         chain_name = ":".join([self.name, token.name])
         chain = CancelToken(chain_name, loop=self._loop)
         chain._chain.extend([self, token])
--- a/p2p/cancel_token.py
+++ b/p2p/cancel_token.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import Any, Awaitable, cast, List  # noqa: F401
+from typing import Any, Awaitable, cast, List
 
 from p2p.exceptions import OperationCancelled
 
@@ -8,7 +8,7 @@ class CancelToken:
 
     def __init__(self, name: str, loop: asyncio.AbstractEventLoop = None) -> None:
         self.name = name
-        self._chain = []  # type: List['CancelToken']
+        self._chain: List['CancelToken'] = []
         self._triggered = asyncio.Event(loop=loop)
         self._loop = loop
 
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -10,14 +10,13 @@ from abc import (
     abstractmethod
 )
 
-from typing import (  # noqa: F401
+from typing import (
     Any,
     Callable,
     Dict,
     Generator,
     Iterator,
     List,
-    Optional,
     Tuple,
     Type,
     TYPE_CHECKING,
@@ -38,7 +37,7 @@ from eth_utils import (
     encode_hex,
 )
 
-from eth_typing import BlockNumber, Hash32  # noqa: F401
+from eth_typing import BlockNumber, Hash32
 
 from eth_keys import (
     datatypes,
@@ -146,11 +145,11 @@ class BasePeer(BaseService):
     reply_timeout = REPLY_TIMEOUT
     # Must be defined in subclasses. All items here must be Protocol classes representing
     # different versions of the same P2P sub-protocol (e.g. ETH, LES, etc).
-    _supported_sub_protocols = []  # type: List[Type[protocol.Protocol]]
+    _supported_sub_protocols: List[Type[protocol.Protocol]] = []
     # FIXME: Must be configurable.
     listen_port = 30303
     # Will be set upon the successful completion of a P2P handshake.
-    sub_proto = None  # type: protocol.Protocol
+    sub_proto: protocol.Protocol = None
 
     def __init__(self,
                  remote: Node,
@@ -172,7 +171,7 @@ class BasePeer(BaseService):
         self.base_protocol = P2PProtocol(self)
         self.headerdb = headerdb
         self.network_id = network_id
-        self.sub_proto_msg_queue = asyncio.Queue()  # type: asyncio.Queue[Tuple[protocol.Command, protocol._DecodedMsgType]]  # noqa: E501
+        self.sub_proto_msg_queue: asyncio.Queue[Tuple[protocol.Command, protocol._DecodedMsgType]] = asyncio.Queue()  # noqa: E501
 
         self.egress_mac = egress_mac
         self.ingress_mac = ingress_mac
@@ -462,12 +461,12 @@ class BasePeer(BaseService):
 class LESPeer(BasePeer):
     max_headers_fetch = les.MAX_HEADERS_FETCH
     _supported_sub_protocols = [les.LESProtocol, les.LESProtocolV2]
-    sub_proto = None  # type: les.LESProtocol
-    head_info = None  # type: les.HeadInfo
+    sub_proto: les.LESProtocol = None
+    head_info: les.HeadInfo = None
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._pending_replies = {}  # type: Dict[int, Callable[[protocol._DecodedMsgType], None]]
+        self._pending_replies: Dict[int, Callable[[protocol._DecodedMsgType], None]] = {}
 
     async def send_sub_proto_handshake(self):
         self.sub_proto.send_handshake(await self._local_chain_info)
@@ -594,9 +593,9 @@ class LESPeer(BasePeer):
 
 class ETHPeer(BasePeer):
     _supported_sub_protocols = [eth.ETHProtocol]
-    sub_proto = None  # type: eth.ETHProtocol
-    head_td = None  # type: int
-    head_hash = None  # type: Hash32
+    sub_proto: eth.ETHProtocol = None
+    head_td: int = None
+    head_hash: Hash32 = None
 
     async def send_sub_proto_handshake(self):
         self.sub_proto.send_handshake(await self._local_chain_info)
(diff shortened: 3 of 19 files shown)


pipermerriam:
> Some modules are split by network (mainnet/ropsten/etc), others split by node type (full/light/sharding?/stateless?). Which makes more sense? (all the ropsten variants in one file, or all the full nodes in one file?)

Network and Mode are different axis so it's not necessarily one or the other, but I think Mode takes precedent in most cases.

pipermerriam:
> Which things should be running in their own threads?

Processes, not threads.

I think every *long* running thing which can be well isolated from the rest deserves it's own process.  We have to be conservative at first because there is overhead to IPC in a multithreading environment, both in terms of performance as well as complexity.  Categories I see us shooting for are.

- Database
- JSON-RPC (currently bundled with networking)
- Networking (likely subdividable)
- The Syncers (currently bundled with networking)

pipermerriam:
> What is the best way to interface with LightPeerChain (formerly p2p.lightchain.LightChain) from a synchronous interface. I tried to avoid asyncio spreading virally through the code base. But still getting errors about futures on different threads, etc, and I'm new enough to asyncio that a good solution isn't obvious to me.

The easiest way you can make it work.  If heavy RPC load breaks networking (which it will) we'll deal with it then, but this is an alpha I'm ok fixing it when it actually manifests as a problem.

trinity/chains/light.py
```diff
@@ -0,0 +1,230 @@
...
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
+
+    @classmethod
+    def from_genesis_header(cls,
+                            base_db: BaseDB,
+                            genesis_header: BlockHeader) -> 'BaseChain':
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
+
+    def get_chain_at_block_parent(self, block: BaseBlock) -> 'BaseChain':
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
```
pipermerriam: Shouldn't the method name be available as part of the stacktrace?

trinity/chains/light.py
```diff
@@ -0,0 +1,230 @@
...
+
+    def get_block(self) -> BaseBlock:
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
+
+    def get_block_by_hash(self, block_hash: Hash32) -> BaseBlock:
+        try:
+            header = self._headerdb.get_block_header_by_hash(block_hash)
+        except HeaderNotFound:
+            header = self._run_async(
+                self._peer_chain.get_block_header_by_hash(block_hash)
```
pipermerriam: This looks fishy to me.  Should we allow header retrieval for headers that aren't in the headerdb, knowing that the light client does have a full db of the header chain?

trinity/nodes/base.py
```diff
@@ -0,0 +1,76 @@
+from abc import abstractmethod
+import asyncio
+from pathlib import PurePath
```
pipermerriam: TIL this exists.

trinity/nodes/base.py
```diff
@@ -0,0 +1,76 @@
...
+)
+
+
+class Node(BaseService):
+    """
+    Create usable nodes by adding subclasses that define the following
+    unset attributes...
+    """
+
+    chain_class = None  # type: BaseChain
```
pipermerriam: These can be written as:

```python
chain_class: BaseChain = None
```

trinity/nodes/base.py
```diff
@@ -0,0 +1,76 @@
...
+    def make_ipc_server(self) -> IPCServer:
+        if self._jsonrpc_ipc_path:
+            rpc = RPCServer(self.get_chain())
+            return IPCServer(rpc, self._jsonrpc_ipc_path)
+        else:
+            return EmptyService()
+
+    async def _run(self):
+        ipc_server = self.make_ipc_server()
+        ipc_loop = self._make_new_loop_thread()
```
pipermerriam: Feels like these should *maybe* be set as properties (maybe private ones) so that they can be inspected if necessary rather than having them burried in this closure.

trinity/nodes/base.py
```diff
@@ -0,0 +1,76 @@
...
+            asyncio.run_coroutine_threadsafe(ipc_server.run(), loop=ipc_loop)
+            await self._peer_chain.run()
+        finally:
+            await ipc_server.stop()
+            await self._peer_pool.cancel()
+
+    async def _cleanup(self):
+        await self._peer_chain.stop()
+
+    def _make_new_loop_thread(self):
```
pipermerriam: Can we document/explain why a new loop is being used here.

trinity/nodes/light.py
```diff
@@ -0,0 +1,23 @@
...
+
+from trinity.chains.light import (
+    LightDispatchChain,
+)
+
+from trinity.nodes.base import Node
+
+
+class LightNode(Node):
+    chain_class = None  # type: LightDispatchChain
```
pipermerriam: use python3.6 style type definition instead of comment

trinity/nodes/light.py
```diff
@@ -0,0 +1,23 @@
...
+    peer_chain_class = LightPeerChain
+
+    _chain = None
+
+    def get_chain(self) -> LightDispatchChain:
+        if self._chain is None:
+            if self.chain_class is None:
+                raise NotImplementedError("LightNode subclass must set chain_class")
```
pipermerriam: `AttributeError` is the exception we've been using elsewhere for this.

trinity/rpc/ipc.py
```diff
@@ -129,6 +129,7 @@ def __init__(self, rpc, ipc_path):
             loop=loop,
             limit=MAXIMUM_REQUEST_BYTES,
         )
+        logging.getLogger('trinity.rpc.ipc.IPCServer').info('IPC Server started')
```
pipermerriam: `self.logger`? (add it if it isn't already present)

trinity/chains/light.py
```diff
@@ -0,0 +1,230 @@
...
+                     base_db: BaseDB,
+                     genesis_params: Dict[str, HeaderParams],
+                     genesis_state: AccountState=None) -> 'BaseChain':
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
+
+    @classmethod
+    def from_genesis_header(cls,
+                            base_db: BaseDB,
+                            genesis_header: BlockHeader) -> 'BaseChain':
+        raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
```
pipermerriam: This pattern of implement exception raising methods on required methods from our ABCs is showing up in a few places.  Can we macro this?

```python
class Thing(BaseThing):
    method_a = stub_abstract_method()
```

Should let us drop a lot of code overhead and we don't really care about the type signature for these methods when we do this.

trinity/chains/light.py
carver: When RPC fails, it only returns the exception message. It was easier to see it in the web3py failure message than go hunting for the stack trace in the trinity logs.

trinity/chains/light.py
carver: The only win I could think of is that it allows us to access headers that we haven't synced to yet, but I don't think it's a critical feature (at least right now). I'll drop it.

trinity/chains/light.py
carver: So I just realized that doing it this way interferes with the inspection. There are some other options, but I'm not clear how common a case this is. It's really common in `BaseThing`, but we care about the method signature there.

The intention here in `LightDispatchChain` is to implement these methods now, or delete the signatures altogether, and let the parent class raise the warning. I could do something for the superclass, like:

```py
class BaseThing(abc):
    @abstract_runtime_raise
    def method(self, arg):
        pass
```

or

```py
class BaseThing(abc):
    @abstractmethod
    def method(self, arg):
        raise UnimplementedException(self)
```

trinity/nodes/base.py
carver: I don't like the idea of weird things happening like having `ipc_server` get reassigned between run and stop, but we can add a copy of these objects on self, and use the local instance everywhere in the method to avoid the problem.

carver:
> Network and Mode are different axis so it's not necessarily one or the other, but I think Mode takes precedent in most cases.

So just to play this out:

trinity/chains/mainnet.py
```
class MainnetLightDispatchChain:
...
class MainnetFullChain:
...
```

trinity/chains/ropsten.py
```
class RopstenLightDispatchChain:
...
class RopstenFullChain:
...
```

trinity/chains/rinkeby.py
```
class RinkebyLightDispatchChain:
...
class RinkebyFullChain:
...
```

OR

trinity/chains/light.py
```
class MainnetLightDispatchChain:
...
class RopstenLightDispatchChain:
...
class RinkebyLightDispatchChain:
...
```

trinity/chains/full.py
```
class RopstenFullChain:
...
class MainnetFullChain:
...
class RinkebyFullChain:
...
```

I *think* I like having the module be the mode, and each of the networks go inside it.

carver:
I was originally trying to get `w3.eth.getBlock('latest')` to work before removing the `[WIP]` tag. But we *could* just merge with `w3.eth.blockNumber` working. I'm happy to go either way. This is still my top priority, so I will keep driving toward `getBlock` on lightchain on Monday.

trinity/chains/light.py
pipermerriam: I don't love the decorator approach. As-is is fine for now.  The overhead isn't too great and if we don't have a good solution I'd rather not waste time architecting.

carver:
Boom:

```sh
$ trinity --ropsten --light --data-dir /tmp/trinity-light-ropsten/
```

```py
>>> from web3 import Web3
>>> w3 = Web3(Web3.IPCProvider('/tmp/trinity-light-ropsten/jsonrpc.ipc'))

>>> w3.eth.blockNumber
1568046

>>> w3t.eth.getBlock('latest') 
AttributeDict({'difficulty': 375839595,
 'extraData': HexBytes('0xd783010607846765746887676f312e382e31856c696e7578'),
 'gasLimit': 4712388,
 'gasUsed': 224035,
 'hash': HexBytes('0xd8bfa67f3bb9736b0aa2e41a3342b641287158bf395d4100023cb6ffa74e72e4'),
 'logsBloom': HexBytes('0x04002000000000008040000000000000000500000000000000000000000000000000000010000000000000002000200000000000000000000000100000000000000000000000000000000008000000000020000000000000000000010040000000000000020400000000000000000800400000000000000000800010000000200000008800000000002000000000000000000000000020000000010000000088000000001000000000000100000000000000000000000002000000000400000010000002000000200000000000000000000000000000000000000020000020000000000000000000000000000000000008000000008000000000000000000000'),
 'miner': '0x3b873A919aA0512D5A0F09E6dCCEaA4a6727faFE',
 'mixHash': HexBytes('0xed839518ce8bb02d68da9f4df14cffc63c806bdea53cd583425e48653e3dd523'),
 'nonce': HexBytes('0x0248f7eba4e71909'),
 'number': 1568046,
 'parentHash': HexBytes('0xa491a47abcbdfdc722cbaaee274065b86178ce8708e8b7469ffe0ee4c4636a34'),
 'receiptsRoot': HexBytes('0x44a1ab474eb98a47744c2e757c941ce8fb61d29df1d5e76c274b0d22b685a4a7'),
 'sha3Uncles': HexBytes('0xa39e00118a1a316451ab07ba3da635032086fc0e99ad272b68725802e494678a'),
 'size': 1845,
 'stateRoot': HexBytes('0xd7df90f777fa2930daa386a52ad19ff0259b0324f51c99c369440662fc8bf60f'),
 'timestamp': 1503925128,
 'totalDifficulty': 1694672428080420,
 'transactions': [HexBytes('0xbff80e712cef9dd58d8858b37eb862da15d358e17d3f6a116d12f7b90009b1aa'),
  HexBytes('0x306bcdd06ab860c6ab7977abe6bac50a1fcb55e541eb35ec545104da558678a5'),
  HexBytes('0xca1509bfd8fe862a97db0382efba8fd35fa41507549a2a771d5b0da35e4ca6b0'),
  HexBytes('0xe10f19b74703ee588c892636b8732c5da0f5396f77516ee44029553a765c055a'),
  HexBytes('0x003e8d87de40628c14ff2c72b2946ba88b8a92e7c69e240a00c51eb6d82e6fb1')],
 'transactionsRoot': HexBytes('0xdd3d2fadfbf179961df4cddd07f34705ae92fa51f99ee462e5f598ff77139fc2'),
 'uncles': [HexBytes('0xc085b0a48c3da24c3e32e4fec8c6d463fbb112ebd582f9624cd74f0c79962fa0')]})
```

carver:
I think this is ready to go. @gsalgado any thoughts on the `CancelToken` changes?

evm/db/header.py
```diff
@@ -132,10 +132,12 @@ def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
         return rlp.decode(header_rlp, BlockHeader)
 
     def get_score(self, block_hash: Hash32) -> int:
-        return rlp.decode(
-            self.db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)],
-            sedes=rlp.sedes.big_endian_int,
-        )
+        try:
+            score = self.db[SchemaV1.make_block_hash_to_score_lookup_key(block_hash)]
```
pipermerriam: blah blah naming nitpick.  `encoded_score`?

p2p/peer.py
```diff
@@ -710,7 +710,14 @@ def add_peer(self, peer):
         if remote in self.connected_nodes:
             self.logger.debug("Skipping %s; already connected to it", remote)
             return None
-        expected_exceptions = (UnreachablePeer, TimeoutError, PeerConnectionLost, HandshakeFailure)
+        expected_exceptions = (
+            UnreachablePeer,
+            TimeoutError,
+            PeerConnectionLost,
+            HandshakeFailure,
```
pipermerriam: nitpick, sort alphabetically.

trinity/chains/light.py
```diff
@@ -0,0 +1,208 @@
+import asyncio
+import inspect
+from typing import (  # noqa: F401
```
pipermerriam: We shouldn't have to `noqa: F401` very often anymore in the `p2p` and `trinity` namespace.  Can you check if dropping this is doable?

trinity/nodes/base.py
```diff
@@ -0,0 +1,84 @@
...
+        new_loop = asyncio.new_event_loop()
+
+        def start_loop(loop):
+            asyncio.set_event_loop(loop)
+            loop.run_forever()
+
+        thread = Thread(target=start_loop, args=(new_loop, ))
+        thread.start()
```
pipermerriam: Maybe worth hiding this away in `self._thread` for the same debugging reasons as `self._ipc_server`.

trinity/chains/light.py
carver: I didn't want to do only one file, but I also didn't want to obscure this PR. So I opened #743.

trinity/nodes/base.py
```diff
@@ -0,0 +1,84 @@
...
+        # The RPC server needs its own thread, because it provides a synchcronous
+        # API which might call into p2p async methods. These sync->async calls
+        # deadlock if they are run in the same Thread and loop.
+        ipc_loop = self._make_new_loop_thread()
+
+        # keep a copy on self, for debugging
+        self._ipc_server, self._ipc_loop = ipc_server, ipc_loop
+
+        try:
+            asyncio.ensure_future(self._peer_pool.run())
```
gsalgado: So, in general the running/stopping of `BaseService` instances should be the responsibility of the services which create them, as we may want them to outlive the services they're being passed to (e.g. the `PeerPool` being passed to the various syncer services in `FullNodeSyncer`), so we should probably move this back to `run_lightnode_process()`

trinity/nodes/base.py
```diff
@@ -0,0 +1,84 @@
...
+)
+from trinity.rpc.main import (
+    RPCServer,
+)
+from trinity.rpc.ipc import (
+    IPCServer,
+)
+
+
+class Node(BaseService):
```
gsalgado: There seems to be some overlap between this and `p2p.Server`, and we were planning to extend the latter to take a `ChainConfig` instance and then instantiate the appropriate syncer, etc. Do you think we'd be able to combine them somehow?

p2p/cancel_token.py
```diff
@@ -19,7 +20,7 @@ def chain(self, token: 'CancelToken') -> 'CancelToken':
         has no effect on either of the chained tokens.
         """
         chain_name = ":".join([self.name, token.name])
-        chain = CancelToken(chain_name)
+        chain = CancelToken(chain_name, loop=self._loop)
```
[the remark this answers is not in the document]
carver: Will do on `chain()`. I didn't see a straightforward API to get the event loop from an `Awaitable` or `asyncio.Future` so I'm not sure how to check that the loop matches on those. (Just get a runtime error, instead)

trinity/nodes/base.py
carver: > in general the running/stopping of BaseService instances should be the responsibility of the services which create them

Agree

> we should probably move this back to run_lightnode_process()

I'm thinking the opposite: moving the PeerPool init into the FullNode setup. Is there any reason we need `PeerPool` to last longer than the whole node? I can't think of any.

That change is pending in a separate branch right now. Ultimately, I am aiming to completely unify the `run_lightnode_process` and `run_fullnode_process` methods. Will post as soon as this PR is merged.

trinity/nodes/base.py
carver: I think we're headed in the same general direction. In general, `p2p.Server` will only be one of many services that the node needs to manage (like the json-rpc server, etc). So my instinct is that this Node service should take the `ChainConfig` and start up `p2p.Server` with that info, to handle all the syncing needs.

carver:
I have some changes queued up behind this, so I'm going to be a bit aggressive on this merge. I addressed all the outstanding stuff that had a clear answer, and I'll do a followup pull for anything that comes out of further conversation here.

PR #724, opened 2018-05-18
Implement a limited version of PoW mining

gsalgado:
So, I'm trying to generate a full blockchain DB programatically, from scratch,
by creating arbitrary transactions and then mining the blocks. I intend to use
that to test the chain syncers

However, I first realized we currently cannot do that as BaseVM.mine_block() doesn't
set the block's mix_hash, so I changed it to do that, but now I get a validation
error on the POW difficulty check, and I've no idea why, so it'd be great if you
guys could have a look at this and maybe shed some light.

Also, is the change to BaseVM.finalise_block() something we'd want to merge?

Finally, can you guys think of a better way to programmatically generate a
blockchain DB with a dozen or so blocks? Or is this approach ok?

The failing test: https://circleci.com/gh/ethereum/py-evm/11905?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

### DIFF
--- a/evm/consensus/pow.py
+++ b/evm/consensus/pow.py
@@ -65,3 +65,18 @@ def check_pow(block_number: int,
             encode_hex(mining_output[b'mix digest']), encode_hex(mix_hash)))
     result = big_endian_to_int(mining_output[b'result'])
     validate_lte(result, 2**256 // difficulty, title="POW Difficulty")
+
+
+MAX_TEST_MINE_ATTEMPTS = 1000
+
+
+def test_mine_pow_nonce(block_number, mining_hash, difficulty):
+    cache = get_cache(block_number)
+    for nonce in range(MAX_TEST_MINE_ATTEMPTS):
+        mining_output = hashimoto_light(block_number, cache, mining_hash, nonce)
+        result = big_endian_to_int(mining_output[b'result'])
+        result_cap = 2**256 // difficulty
+        if result <= result_cap:
+            return nonce.to_bytes(8, 'big'), mining_output[b'mix digest']
+
+    raise Exception("Too many attempts at POW mining, giving up")
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -212,7 +212,7 @@ class BaseVM(Configurable, metaclass=ABCMeta):
 
     @classmethod
     @abstractmethod
-    def get_prev_hashes(cls, last_block_hash, db):
+    def get_prev_hashes(cls, last_block_hash, chaindb):
         raise NotImplementedError("VM classes must implement this method")
 
     @staticmethod
@@ -565,16 +565,16 @@ class VM(BaseVM):
     @classmethod
     @functools.lru_cache(maxsize=32)
     @to_tuple
-    def get_prev_hashes(cls, last_block_hash, db):
+    def get_prev_hashes(cls, last_block_hash, chaindb):
         if last_block_hash == GENESIS_PARENT_HASH:
             return
 
-        block_header = get_block_header_by_hash(last_block_hash, db)
+        block_header = get_block_header_by_hash(last_block_hash, chaindb)
 
         for _ in range(MAX_PREV_HEADER_DEPTH):
             yield block_header.hash
             try:
-                block_header = get_parent_header(block_header, db)
+                block_header = get_parent_header(block_header, chaindb)
             except (IndexError, HeaderNotFound):
                 break
 
--- a/evm/vm/forks/frontier/__init__.py
+++ b/evm/vm/forks/frontier/__init__.py
@@ -81,3 +81,15 @@ class FrontierVM(VM):
     @classmethod
     def get_nephew_reward(cls):
         return cls.get_block_reward() // 32
+
+
+# A VM that does POW mining as well. Should be used only in tests, when we need to programatically
+# populate a ChainDB.
+class _PoWMiningVM(FrontierVM):
+
+    def finalize_block(self, block):
+        from evm.consensus import pow
+        block = super().finalize_block(block)
+        nonce, mix_hash = pow.test_mine_pow_nonce(
+            block.number, block.header.mining_hash, block.header.difficulty)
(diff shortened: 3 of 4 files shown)


pipermerriam:
Hrm, my understanding of actually mining blocks was that you'd need to enumerate `nonce` values until the resulting `mix_hash` satisfied the POW requirements based on the `difficulty`.  It looks like this code just uses `header.nonce` which is likely set to `0`.  I suspect you need a loop here to enumerate `nonce` values until the resulting `mix_hash` qualifies.

gsalgado:
Oh, of course. And my test only passes the mix_hash check because BaseVM.finalise_block() now poisons the cache used by `check_pow()`

pipermerriam:
I just pushed a commit fixing this, though on my machine with the genesis difficulty it takes a long time to mine blocks, consider setting difficulty to `1` if that works for the tests your running.

pipermerriam:
I think these changes to finalize block should be updated with the following conditionals.

1. block mining **must** be explicitely enabled.
2. mining the `mix_hash/nonce` should **only** occur if the `mix_hash` and `nonce` are set to their default values.

pipermerriam:
> it takes a long time to mine blocks, consider setting difficulty to 1 if that works for the tests your running.

As in, been running for a few minutes now with no luck.

pipermerriam:
More thoughts on how to include actual POW mining here.

We probably shouldn't hard code the implementation but rather use a swappable backend system.  This would allow use of both `hashimoto_light`, the version that uses the full DAG, or some other version that moved the computation to a GPU.

gsalgado:
Oh, damn, only saw your comments after I'd pushed an amended commit. Should've suspected when git told me I needed to force. Anyway, it's pretty much the same as yours but it uses a random nonce on every iteration. Is there any reason why sequentially incrementing it would be preferable?

gsalgado:
In the new version here I used 99 as the difficulty and with that it finds a solution pretty quickly (less than 100 attempts, usually)

pipermerriam:
> Is there any reason why sequentially incrementing it would be preferable?

cheaper to produce?

pipermerriam:
And even more thoughts on how to include actual POW mining here.

With a backend system, we don't need a flag to enable POW mining, but can instead just throw an error if no backend has been explicitly set.  To support chains that have disabled POW validation, this should probably leverage `VM.validate_seal` (not yet merged from #667 ) as the check for whether it should generate the seal.

evm/vm/base.py
```diff
@@ -484,7 +491,31 @@ def finalize_block(self, block):
...
+            else:
+                print("==============================")
+                print("Found mix_hash after %d attempts" % i)
+                print("==============================")
+                break
+        else:
+            raise Exception("Too many attempts, giving up")
+        header = pre_mining_header.copy(
+            nonce=nonce,
+            mix_hash=mining_output[b'mix digest'])
```
carver: Looks like the original question is mostly resolved. The first thing I noticed is that this code block can be extracted with a pretty nice interface of `(block.number. mining_hash) -> (nonce, mix_hash)`.  That API only makes sense for "test mining" though since real mining is going to be a simultaneous mix of constantly maximizing gas in the block while searching for a nonce. Which means that managing "real mining" probably belongs out in `trinity` anyway.

evm/consensus/pow.py
```diff
@@ -65,3 +65,18 @@ def check_pow(block_number: int,
             encode_hex(mining_output[b'mix digest']), encode_hex(mix_hash)))
     result = big_endian_to_int(mining_output[b'result'])
     validate_lte(result, 2**256 // difficulty, title="POW Difficulty")
+
+
+MAX_TEST_MINE_ATTEMPTS = 1000
+
+
+def test_mine(block_number, mining_hash, difficulty):
```
cburgdorf: Coincidentally I implemented the same method when I started writing the "Understanding the mining process" guide. I think it's beneficial for have this method even if only being useful for tests and docs.

 I would however propose that we rename it to something like `mine_nonce` or maybe even `mine_nonce_and_mixhash`. Thoughts?

cburgdorf:
@carver @pipermerriam Just read your comments **after** I submitted a review. Having something pluggable makes sense but the PR as it stands right now does not change anything but rather only adds an additional API to be used in tests/docs. With that in mind I think this would be safe to merge (maybe after renaming the `test_mine` API to something like `mine_nonce_and_mixhash`

carver:
Yup, in current state I'm :+1: to merge. My comments are stale, from an old version.

evm/consensus/pow.py
pipermerriam: I'd suggest adding `pow` or `ethash` to the name to make it clear that it's proof-of-work mining.

pipermerriam: I like `mine_pow_nonce`.

cburgdorf: sgtm :+1:

gsalgado: Currently the callsites import the `pow` package, so this is always called like `pow.test_mine()`, which is why I didn't want to include `pow` in the function's name. I can rename if you guys feel strong about it

cburgdorf: It's still prefixed with `test` though.

gsalgado: Correct, because it is not meant to be used in production code as it performs only a small number of attempts and hence will never actually work on anything other than dummy blocks with a ridiculously low difficulty.

cburgdorf: Sure. Might be just me but the `test` prefix suggests to me that this method performs a test which it doesn't.

pipermerriam: It will also cause `pytest` to pick up and try to run the function (if it's not run against the `tests` directory).

cburgdorf: I renamed it as part of the other PR that is about to land https://github.com/ethereum/py-evm/pull/641/commits/9a3cfdac75b842378f108f75283c5ecaf2d69dae

PR #727, opened 2018-05-18
handle division by zero when peer pool is empty

pipermerriam:
#### Cute Animal Picture

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -3,8 +3,8 @@ import logging
 import math
 import operator
 import time
-from typing import (  # noqa: F401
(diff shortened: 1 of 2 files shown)


gsalgado:
I'm afraid this is not the right approach. If you check `_sync()`'s main loop, you'll see it aborts if the peer that triggered the sync gets dropped, so these coroutines must either not be scheduled or must also check if the originating peer has been dropped and abort themselves. I can have a proper look at it on Monday -- it's quite late here already

pipermerriam:
I feel like I'm almost there in grasping what is going on here but.... if you have a handle on what needs to be done, go for it.

PR #732, opened 2018-05-18
Make vm_configuration a standard across all Chains

carver:
### What was wrong?

It would be *weird* if a Chain subclass configured the VMs differently than the `vm_configuration` standard. So weird that it would feel broken. So implementing `get_vm_class_by_block_number()` in every subclass is redundant.

### How was it fixed?

Moved `vm_configuration` and `get_vm_class_by_block_number()` to `BaseChain`.

Inspired by #717

#### Cute Animal Picture


pipermerriam:
Merged despite what appears to be a clear violation of PR rules.

Issue #725, opened 2018-05-18
Convert path types to use `PurePath`


pipermerriam:
### What is wrong?

In the `trinity` module we have a number of path objects that are typed as `str`.  

The `pathlib` standard library module has a [`PurePath`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath) class which is more appropriate.

### How can it be fixed

Update the `trinity` module to use the `PurePath` type for all filesystem paths.  If it seems appropriate, lets also update the code to actually use and return `PurePath` objects.

glaksmono:
@pipermerriam would like to take a stab on this 🗡

pipermerriam:
⚔️ stab away

PR #716, opened 2018-05-17
Add more info to terminal output for trinity

pipermerriam:
Fixes #709 

### What was wrong?

The console output when running the trinity node needed some more information.

### How was it fixed?

- added an ascii header
- added `enode` info for light node
- added path to database IPC socket.
- added network id.

```bash
$ trinity
INFO trinity.db 2018-05-17 12:42:08,787 - database: /Users/piper/.local/share/trinity/mainnet/db.ipc
INFO trinity 2018-05-17 12:42:09,792 -
  ______     _       _ __
 /_  __/____(_)___  (_) /___  __
  / / / ___/ / __ \/ / __/ / / /
 / / / /  / / / / / / /_/ /_/ /
/_/ /_/  /_/_/ /_/_/\__/\__, /
                       /____/
INFO trinity 2018-05-17 12:42:09,792 - Trinity/0.2.0a17/darwin/cpython3.6.3
INFO p2p.server.Server 2018-05-17 12:42:09,822 - Running server...
INFO p2p.server.Server 2018-05-17 12:42:14,827 - No UPNP-enabled devices found
INFO p2p.server.Server 2018-05-17 12:42:14,829 - enode://7b8fb3e32de9aa5d7930fe0dd0e2365361b85c97daa1eb042ac37005f527bb80a450e9ec9012d53d8fcb94d015b115630f4556153fcf4e746993a29b14e13734@0.0.0.0:30303
INFO p2p.server.Server 2018-05-17 12:42:14,829 - network: 1
```

#### Cute Animal Picture

tumblr_maqmw0fl8u1qzamioo6_1280

PR #734, opened 2018-05-19
Convert path types to use `PurePath`

glaksmono:
### What was wrong?

In the `trinity` module we have a number of path objects that are typed as `str`.

The `pathlib` standard library module has a [PurePath](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath) class which is more appropriate.

Issue Reference: https://github.com/ethereum/py-evm/issues/725, https://github.com/ethereum/py-evm/pull/716#discussion_r189361151

### How was it fixed?

Update the `trinity` module to use the `PurePath` type for all filesystem paths. If it seems appropriate, lets also update the code to actually use and return `PurePath` objects.

#### Cute Animal Picture

### DIFF
--- a/trinity/chains/__init__.py
+++ b/trinity/chains/__init__.py
@@ -79,7 +79,7 @@ def is_database_initialized(chaindb: AsyncChainDB) -> bool:
 
 
 def initialize_data_dir(chain_config: ChainConfig) -> None:
-    if is_under_xdg_trinity_root(chain_config.data_dir):
+    if is_under_xdg_trinity_root(str(chain_config.data_dir)):
         os.makedirs(chain_config.data_dir, exist_ok=True)
     elif not os.path.exists(chain_config.data_dir):
         # we don't lazily create the base dir for non-default base directories.
--- a/trinity/utils/chains.py
+++ b/trinity/utils/chains.py
@@ -1,5 +1,7 @@
 import os
 
+from pathlib import PurePath
+
 from eth_utils import (
     decode_hex,
     to_dict,
@@ -36,17 +38,17 @@ DATABASE_DIR_NAME = 'chain'
 #
 # Filesystem path utils
 #
-def get_local_data_dir(chain_name: str) -> str:
+def get_local_data_dir(chain_name: str) -> PurePath:
     """
     Returns the base directory path where data for a given chain will be stored.
     """
-    return os.environ.get(
+    return PurePath(os.environ.get(
         'TRINITY_DATA_DIR',
         os.path.join(get_xdg_trinity_root(), chain_name),
-    )
+    ))
 
 
-def get_data_dir_for_network_id(network_id: int) -> str:
+def get_data_dir_for_network_id(network_id: int) -> PurePath:
     """
     Returns the data directory for the chain associated with the given network
     id.  If the network id is unknown, raises a KeyError.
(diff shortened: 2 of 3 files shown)


pipermerriam:
Looks like some of these may need to be downcast back to strings.

https://circleci.com/gh/ethereum/py-evm/12997?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I'd be good with that downcasting happening in the `ChainConfig` object (basically, reverting some of the return values back to `str` for things like the IPC paths.  If you can leave a comment in there as well indicating why `str` is used instead of `PurePath` that would be great.  It looks primarily like an issue with `multiprocessing` not being able to interpret `PurePath` objects correctly.

glaksmono:
Done @pipermerriam 👍

trinity/utils/chains.py
```diff
@@ -151,14 +159,14 @@ def data_dir(self) -> str:
...
     @data_dir.setter
     def data_dir(self, value: str) -> None:
-        self._data_dir = os.path.abspath(value)
+        self._data_dir = PurePath(os.path.abspath(value))
 
     @property
     def database_dir(self) -> str:
         if self.sync_mode == SYNC_FULL:
-            return os.path.join(self.data_dir, DATABASE_DIR_NAME, "full")
+            return str(self.data_dir / DATABASE_DIR_NAME / "full")
```
pipermerriam: TIL `PurePath` objects use the division operator in place of `os.path.join`.  👍 

This syntax is going to take some time to get used to.

trinity/chains/__init__.py
```diff
@@ -79,7 +79,7 @@ def is_database_initialized(chaindb: AsyncChainDB) -> bool:
 
 
 def initialize_data_dir(chain_config: ChainConfig) -> None:
-    if is_under_xdg_trinity_root(chain_config.data_dir):
+    if is_under_xdg_trinity_root(str(chain_config.data_dir)):
```
carver: Can't we change this util method to accept `PurePath`s?

pipermerriam:
Thanks for the contribution.

trinity/utils/chains.py
```diff
@@ -60,55 +62,61 @@ def get_data_dir_for_network_id(network_id: int) -> str:
...
     """
     Returns the path to the private key used for devp2p connections.
+
+    We're still returning 'str' here on ipc-related path because an issue with
+    multi-processing not being able to interpret 'PurePath' objects correctly.
     """
     return os.environ.get(
         'TRINITY_DATABASE_IPC',
-        os.path.join(data_dir, DATABASE_SOCKET_FILENAME),
+        os.path.join(str(data_dir), DATABASE_SOCKET_FILENAME),
```
carver: I think the approach you did above method is more of a standard `PurePath` approach (but I mostly care about internal consistency).

```py
str(data_dir / DATABASE_SOCKET_FILENAME)
```

trinity/utils/chains.py
```diff
@@ -60,55 +62,61 @@ def get_data_dir_for_network_id(network_id: int) -> str:
...
     """
     Returns the path to the ipc socket for the JSON-RPC server.
+
+    We're still returning 'str' here on ipc-related path because an issue with
+    multi-processing not being able to interpret 'PurePath' objects correctly.
     """
     return os.environ.get(
         'TRINITY_JSONRPC_IPC',
-        os.path.join(data_dir, JSONRPC_SOCKET_FILENAME),
+        os.path.join(str(data_dir), JSONRPC_SOCKET_FILENAME),
```
carver: `str(data_dir / JSONRPC_SOCKET_FILENAME)`

trinity/utils/chains.py
```diff
@@ -60,55 +62,61 @@ def get_data_dir_for_network_id(network_id: int) -> str:
...
     )
 
 
 #
 # Nodekey loading
 #
-def load_nodekey(nodekey_path: str) -> PrivateKey:
-    with open(nodekey_path, 'rb') as nodekey_file:
+def load_nodekey(nodekey_path: PurePath) -> PrivateKey:
+    with open(str(nodekey_path), 'rb') as nodekey_file:
```
carver:
```py
with nodekey_path.open('rb') as nodekey_file:
```
[according to docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.open)

trinity/utils/chains.py
```diff
@@ -151,14 +159,14 @@ def data_dir(self) -> str:
 
     @data_dir.setter
     def data_dir(self, value: str) -> None:
-        self._data_dir = os.path.abspath(value)
+        self._data_dir = PurePath(os.path.abspath(value))
```
carver: Try:

`self._data_dir = PurePath(value).resolve()`

[`resolve()`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve)

trinity/utils/chains.py
```diff
@@ -182,7 +190,7 @@ def nodekey_path(self) -> str:
 
     @nodekey_path.setter
     def nodekey_path(self, value: str) -> None:
-        self._nodekey_path = os.path.abspath(value)
+        self._nodekey_path = PurePath(os.path.abspath(value))
```
carver: `PurePath(value).resolve()`

glaksmono:
Will submit a follow up PR for the clean ups @pipermerriam @carver 👍

PR #735, opened 2018-05-20
Add subscribe context manager API to PeerPoolSubscriber

pipermerriam:
### What was wrong?

The various classes which are `PeerPoolSubscriber` subclasses were not consistently unsubscribing themselves from the peer pool.

### How was it fixed?

Added a new context manager API so that unsubscribe-ing is automatic upon exiting the context. 

#### Cute Animal Picture

chicks_in_hats7

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -102,20 +102,20 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             self.logger.exception("Unexpected error when processing msg from %s", peer)
 
     async def _run(self) -> None:
-        self.peer_pool.subscribe(self)
-        while True:
-            peer_or_finished = await wait_with_token(
-                self._sync_requests.get(), self._sync_complete.wait(),
-                token=self.cancel_token)
+        with self.subscribe(self.peer_pool):
+            while True:
+                peer_or_finished = await wait_with_token(
+                    self._sync_requests.get(), self._sync_complete.wait(),
+                    token=self.cancel_token)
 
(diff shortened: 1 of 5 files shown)


p2p/sharding.py
```diff
@@ -156,27 +156,27 @@ def __init__(self, shard: Shard, peer_pool: PeerPool, token: CancelToken=None) -
...
+                self.logger.debug("Adding collation {} to shard".format(collation))
+                self.shard.add_collation(collation)
+                for peer in self.peer_pool.peers:
+                    cast(ShardingPeer, peer).send_collations([collation])
+
+                self.collations_received_event.set()
+                self.collations_received_event.clear()
 
     async def _cleanup(self) -> None:
         self.peer_pool.unsubscribe(self)
```
gsalgado: This should be removed as well, right?

p2p/chain.py
```diff
@@ -102,20 +102,20 @@ def register_peer(self, peer: BasePeer) -> None:
...
-
-            # In the case of a fast sync, we return once the sync is completed, and our caller
-            # must then run the StateDownloader.
-            if self._sync_complete.is_set():
-                return
-
-            # Since self._sync_complete is not set, peer_or_finished can only be a ETHPeer
-            # instance.
-            asyncio.ensure_future(self.sync(peer_or_finished))
+        with self.subscribe(self.peer_pool):
```
gsalgado: We should remove the call to `unsubscribe()` from `_cleanup()` here, right?

Issue #726, opened 2018-05-18
division by zero in `FastChainSyncer`


pipermerriam:
### What is wrong?

When the `PeerPool` has no peers we get a division by zero error.

```
Traceback (most recent call last):
  File "/Users/piper/sites/py-evm/p2p/chain.py", line 133, in sync
    await self._sync(peer)
  File "/Users/piper/sites/py-evm/p2p/chain.py", line 172, in _sync
    head_number = await self._process_headers(peer, headers)
  File "/Users/piper/sites/py-evm/p2p/chain.py", line 194, in _process_headers
    'receipt')
  File "/Users/piper/sites/py-evm/p2p/chain.py", line 229, in _download_block_parts
    pending_replies = request_func(missing)
  File "/Users/piper/sites/py-evm/p2p/chain.py", line 292, in request_receipts
    return self._request_block_parts(headers, self._send_get_receipts)
  File "/Users/piper/sites/py-evm/p2p/chain.py", line 260, in _request_block_parts
    length = math.ceil(len(headers) / len(self.peer_pool.peers))
ZeroDivisionError: division by zero
```

### How can it be fixed

TODO: how do we fix this

pipermerriam:
@gsalgado should we do some sort of `await` for the peer pool till it has peers?

gsalgado:
This is a race condition, I'm afraid. The chain syncer runs `sync()` in the background when it hears about a new peer/block, and it looks like by the time `sync()` got scheduled the `PeerPool` had lost all its peers. I guess we need to check for that at the beginning of `sync()` and abort if it's the case

pipermerriam:
@gsalgado what do you think about https://github.com/ethereum/py-evm/pull/727

PR #736, opened 2018-05-20
Cancel sync when peer pool is found to be empty

pipermerriam:
### What was wrong?

The `p2p.chain.FastChainSyncer` ends up throwing `ZeroDivisionErrors` during sync when the peer pool ends up empty.

### How was it fixed?

Catch the `ZeroDivisionError` and instead raise an `OperationCancelled`.

#### Cute Animal Picture

c2a00b7aae67e2da033a199445acdef5--easter-chick-baby-chickens

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -257,7 +257,13 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             self,
             headers: List[BlockHeader],
             request_func: Callable[[ETHPeer, List[BlockHeader]], None]) -> int:
-        length = math.ceil(len(headers) / len(self.peer_pool.peers))
+
+        try:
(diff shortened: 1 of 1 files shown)


pipermerriam:
Another stab at this.  Seems too simple and that I may be mis-using `OperationCancelled`.

link: https://github.com/ethereum/py-evm/issues/726

gsalgado:
Indeed, OperationCancelled should be triggered only by a CancelToken, and it is meant to bubble up and cause all pending tasks (including the ChainSyncer's main loop) to exit, which is not what we want here. Here what we want is to exit only up to the current `sync()` task, but continue running the syncer's main loop as it will fire another `sync()` task when a new peer shows up

gsalgado:
#739 cleanly aborts a sync when there are no peers, but leaves the syncer running so that we can resume syncing when peers show up

PR #739, opened 2018-05-21
p2p: Cleanly abort sync if there are no connected peers

### LINKED ISSUE #726: division by zero in `FastChainSyncer`

gsalgado:
Closes #726

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -19,7 +19,7 @@ from evm.rlp.transactions import BaseTransaction  # noqa: F401
 from p2p import protocol
 from p2p import eth
 from p2p.cancel_token import CancelToken, wait_with_token
-from p2p.exceptions import OperationCancelled
+from p2p.exceptions import NoConnectedPeers, OperationCancelled
 from p2p.peer import BasePeer, ETHPeer, PeerPool, PeerPoolSubscriber
 from p2p.service import BaseService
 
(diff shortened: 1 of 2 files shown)


p2p/chain.py
```diff
@@ -257,6 +272,8 @@ def _request_block_parts(
             self,
             headers: List[BlockHeader],
             request_func: Callable[[ETHPeer, List[BlockHeader]], None]) -> int:
+        if not self.peer_pool.peers:
+            raise NoConnectedPeers()
         length = math.ceil(len(headers) / len(self.peer_pool.peers))
```
pipermerriam: Do you think changing this to use a `try/except` with `ZeroDivisionError` so that we are guaranteed to not have a race condition here would be good (peer pool empties between the check on 275 and this line).

gsalgado: That can't happen because nothing yields control between line 275 and here

pipermerriam: I was more advocating that approach as a future proof approach since it's possible that future refactors change the structure in some minor way that it becomes possible.  `try/except` pattern provides a stronger link.  Not a big deal.  I may throw it in here in some of my other cleanup. 👍

gsalgado: Oh, sorry, I misunderstood you. I thought you meant to catch `ZeroDivisionError` in addition to checking, but now I think you suggested I do that instead of the check? I will include that change in another PR, unless you beat me to it

PR #740, opened 2018-05-21
Fixes for uncle validation

pipermerriam:
Follow up work from #707 

### What was wrong?

The uncle validation was embeded within the `VM.validate_block` which is problematic since uncles are for *other* block numbers, meaning that in situations where we are immediately after the boundary of two fork rules, we'll be running the validation for the new fork on blocks which were mined in the old fork.

### How was it fixed?

Move the orchestration and execution of this up to the `Chain` class.

#### Cute Animal Picture

b3e525d1d4dc839849c78028cc76f7e8

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -4,6 +4,7 @@ from abc import (
     ABCMeta,
     abstractmethod
 )
+import operator
 from typing import (  # noqa: F401
     Any,
     Optional,
@@ -22,6 +23,7 @@ import logging
 
 from cytoolz import (
     assoc,
+    groupby,
 )
 
 from eth_typing import (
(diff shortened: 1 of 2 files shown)


pipermerriam:
@gsalgado sanity check requested on this.

evm/chains/base.py
```diff
@@ -610,33 +612,65 @@ def validate_uncles(self, block: BaseBlock) -> None:
...
+
+        recent_ancestors = {
+            ancestor.hash: ancestor
+            for ancestor in self.get_ancestors(MAX_UNCLE_DEPTH + 1)
+        }
+        recent_uncles = {
+            uncle.hash
+            for ancestor in recent_ancestors.values()
+            for uncle in ancestor.uncles
+        }
```
gsalgado: Hmm, this confused the hell out of me... it looks like you're specifying just the dict keys but no values? Maybe just something I'm not used to, or do you think it'd make sense to rewrite this without the nested comprehensions?

pipermerriam: It's constructing a `set` instead of a dictionary since we don't actually need the uncle here, just the hash and a check for inclusion.

I'm inclined to leave the double comprehension unless you feel strongly about it.

gsalgado: Oh, set comprehensions! I'd never seen that before, and I guess that's what really confused me as I thought you were constructing a dict and I'd no idea how that was working. So yeah, just leave them nested, and thanks for the clarification!

evm/vm/base.py
```diff
@@ -691,27 +687,23 @@ def validate_seal(self, header: BlockHeader) -> None:
             header.block_number, header.mining_hash,
             header.mix_hash, header.nonce, header.difficulty)
 
-    def validate_uncle(self, block, uncle):
+    def validate_uncle(self, block, uncle, uncle_parent):
```
carver: Maybe `niece` instead of `uncle_parent`? I had a hard time parsing `uncle_parent`, which seems to mean `grandparent` without digging further.

pipermerriam: I think uncles can come from even a few blocks back so it's not something we can anchor to the `block` (and even the term `uncle` is a misnomer as what we call uncles can actually be great-uncles (block -> parent -> great-parent -> uncle (aka grand-uncle)).

So this is meant to be the *parent of the uncle*.  No better names come to mind.

carver: Got it, thanks. I misread the validation. I also cannot think of a better name.

Issue #741, opened 2018-05-21
Extend `HardCodedNodesPeerPool` class to also do discovery.


pipermerriam:
### What is wrong?

We currently use a stop-gap solution for working around v4 discovery, `p2p.peer.HardCodedNodesPeerPool`.  It contains known nodes to connect to, specifically for LES connections.

This approach is problematic because these nodes are not guaranteed to be online or to have capacity so.  If they are, then Trinity will not be able to connect until a slot becomes available.  Worse, there's no way without code modification to add new nodes to this list.

### How can it be fixed

Extend how we use this class to to *also* do discovery.  For the first iteration it should treat the hard coded list the same as any nodes that have been connected to via discovery.

Also, add a new CLI flag which can be used to add nodes to this list.

```bash
$ trinity --static-node "enode://deadbeef...@1.2.3.4:30303"
```

This flag should be accepted multiple times to allow specifying multiple nodes.

gsalgado:
>  For the first iteration it should treat the hard coded list the same as any nodes that have been connected to via discovery.

Not sure what you mean by that?

pipermerriam:
Alright, how about this logic which should be simple to implement.

1. For each static node, maintain a timestamp of the last attempted time to connect.
2. When we request a *random* node, we first exhaust our static node list, ignoring any for which we recently (maybe 300 seconds?).
3. If there are no nodes in our static nodes list which we can try (all of them have been tried in the last 300 seconds), we pull random nodes from the routing table.

PR #744, opened 2018-05-21
Add PreferredNodePeerPool

pipermerriam:
link: https://github.com/ethereum/py-evm/issues/741

### What was wrong?

- Running `trinity --light` with a niave peer pool and discovery will rarely find an LES peer to connect to.
- Using `HardCodedNodesPeerPool` eliminates discovery, but also locks Trinity into only being able to connect to a few nodes.
- We will be implementing APIs for Trinity to support reserving slots for certain nodes and trusted nodes.

### How was it fixed?

Implement a new `PeerPool` class which implements list of *preferred* nodes.  

- Anytime the peer pool is looking for a new node to connect to it will try to use one of the preferred nodes.  
- Each preferred node may only be used once every 300 seconds.
- If there are no preferred nodes which have not been used recently then it falls back to use the discovery protocol

#### Cute Animal Picture

86037878e80b3023632ab6b9907a750b

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -912,7 +912,7 @@ class PreferredNodePeerPool(PeerPool):
     @to_tuple
     def _get_eligible_preferred_nodes(self) -> Generator[Node, None, None]:
         """
-        Returns nodes from the preferred_nodes which have not been used within
+        Return nodes from the preferred_nodes which have not been used within
         the last preferred_node_recycle_time
         """
         for node in self.preferred_nodes:
@@ -922,7 +922,7 @@ class PreferredNodePeerPool(PeerPool):
 
     def _get_random_preferred_node(self) -> Node:
         """
-        Returns a random node from the preferred list.
+        Return a random node from the preferred list.
         """
         eligible_nodes = self._get_eligible_preferred_nodes()
         if not eligible_nodes:
@@ -932,7 +932,7 @@ class PreferredNodePeerPool(PeerPool):
 
     def _get_random_bootnode(self) -> Generator[Node, None, None]:
         """
-        Returns a single node to bootstrap, preferring nodes from the preferred list.
+        Return a single node to bootstrap, preferring nodes from the preferred list.
         """
         try:
             node = self._get_random_preferred_node()
@@ -943,7 +943,7 @@ class PreferredNodePeerPool(PeerPool):
 
     def get_nodes_to_connect(self) -> Generator[Node, None, None]:
         """
-        Returns up to `min_peers` nodes, preferring nodes from the preferred list.
+        Return up to `min_peers` nodes, preferring nodes from the preferred list.
         """
         preferred_nodes = self._get_eligible_preferred_nodes()[:self.min_peers]
         for node in preferred_nodes:
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -1,4 +1,5 @@
 import asyncio
+import collections
 import contextlib
 import logging
 import operator
@@ -17,9 +18,10 @@ from typing import (
     Generator,
     Iterator,
     List,
+    Sequence,
+    TYPE_CHECKING,
     Tuple,
     Type,
-    TYPE_CHECKING,
     cast,
 )
 
@@ -35,6 +37,7 @@ from cryptography.hazmat.primitives.constant_time import bytes_eq
 from eth_utils import (
     decode_hex,
     encode_hex,
+    to_tuple,
 )
 
 from eth_typing import BlockNumber, Hash32
@@ -48,6 +51,8 @@ from eth_hash.auto import keccak
 
 from trie import HexaryTrie
 
+from evm.chains.mainnet import MAINNET_NETWORK_ID
+from evm.chains.ropsten import ROPSTEN_NETWORK_ID
 from evm.constants import GENESIS_BLOCK_NUMBER
 from evm.exceptions import BlockNotFound
 from evm.rlp.accounts import Account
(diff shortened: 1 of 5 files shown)


p2p/peer.py
```diff
@@ -796,61 +802,68 @@ def peers(self) -> List[BasePeer]:
...
 
     def _get_random_bootnode(self) -> Generator[Node, None, None]:
         # We don't have a DiscoveryProtocol with bootnodes, so just return one of our regular
         # hardcoded nodes.
-        yield random.choice(list(self.get_nodes_to_connect()))
+        options = list(self.get_nodes_to_connect())
+        if options:
+            yield random.choice(options)
+        else:
+            yield None
```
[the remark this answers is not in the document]
pipermerriam: No, this is a mistake, it should just not have the `else` clause and return an empty generator.

p2p/peer.py
```diff
@@ -859,6 +872,82 @@ def get_nodes_to_connect(self) -> Generator[Node, None, None]:
...
+            self._preferred_node_tracker[node] = time.time()
+            yield node
+        except NoEligibleNodes:
+            yield from super()._get_random_bootnode()
+
+    def get_nodes_to_connect(self) -> Generator[Node, None, None]:
+        """
+        Returns up to `min_peers` nodes, preferring nodes from the preferred list.
+        """
+        preferred_nodes = self._get_eligible_preferred_nodes()[:self.min_peers]
```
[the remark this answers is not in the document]
pipermerriam: I think there is a lot of room for improvement to this class, including something like you suggest here, but for now I'd like to call this *good enough*.  

The bootnodes will not allow additional connections if all of their peer slots are full and trinity will then fallback to discovery.  It should only take a moment to exhaust the pre-defined list of preferred nodes, after which discovery will take over.

pipermerriam:
@carver Can you double check the body of `run_lightnode_process`.  I had handle a merge conflict and I want to be sure everything still looks right.

p2p/peer.py
```diff
@@ -17,9 +18,10 @@
     Generator,
     Iterator,
     List,
+    Sequence,
+    TYPE_CHECKING,
```
cburgdorf: nitpick, this seems to screw up alphabetical ordering

p2p/peer.py
```diff
@@ -865,6 +878,82 @@ def get_nodes_to_connect(self) -> Generator[Node, None, None]:
...
+        else:
+            self.logger.debug('PreferredNodePeerPool operating with no preferred nodes')
+            self.preferred_nodes = tuple()
+
+        self._preferred_node_tracker = collections.defaultdict(lambda: 0)
+
+    @to_tuple
+    def _get_eligible_preferred_nodes(self) -> Generator[Node, None, None]:
+        """
+        Returns nodes from the preferred_nodes which have not been used within
```
cburgdorf: nitpick: Style guide prefers imperative, present tense.

>Return nodes from the preferred_nodes which have not been used within the last preferred_node_recycle_time

p2p/peer.py
```diff
@@ -865,6 +878,82 @@ def get_nodes_to_connect(self) -> Generator[Node, None, None]:
...
+        the last preferred_node_recycle_time
+        """
+        for node in self.preferred_nodes:
+            last_used = self._preferred_node_tracker[node]
+            if time.time() - last_used > self.preferred_node_recycle_time:
+                yield node
+
+    def _get_random_preferred_node(self) -> Node:
+        """
+        Returns a random node from the preferred list.
```
cburgdorf: see above

p2p/peer.py
```diff
@@ -865,6 +878,82 @@ def get_nodes_to_connect(self) -> Generator[Node, None, None]:
...
+        """
+        eligible_nodes = self._get_eligible_preferred_nodes()
+        if not eligible_nodes:
+            raise NoEligibleNodes("No eligible preferred nodes available")
+        node = random.choice(eligible_nodes)
+        return node
+
+    def _get_random_bootnode(self) -> Generator[Node, None, None]:
+        """
+        Returns a single node to bootstrap, preferring nodes from the preferred list.
```
cburgdorf: see above

p2p/peer.py
```diff
@@ -865,6 +878,82 @@ def get_nodes_to_connect(self) -> Generator[Node, None, None]:
...
+        try:
+            node = self._get_random_preferred_node()
+            self._preferred_node_tracker[node] = time.time()
+            yield node
+        except NoEligibleNodes:
+            yield from super()._get_random_bootnode()
+
+    def get_nodes_to_connect(self) -> Generator[Node, None, None]:
+        """
+        Returns up to `min_peers` nodes, preferring nodes from the preferred list.
```
cburgdorf: see above

p2p/peer.py
```diff
@@ -865,6 +878,82 @@ def get_nodes_to_connect(self) -> Generator[Node, None, None]:
...
 
+class PreferredNodePeerPool(PeerPool):
+    """
+    A PeerPool which has a list of preferred nodes which it will prioritize
+    using before going to the discovery protocol to find nodes.  Each preferred
+    node can only be used once every preferred_node_recycle_time seconds.
+    """
+    preferred_nodes: Sequence[Node] = None
+    preferred_node_recycle_time: int = 300
+    _preferred_node_tracker: Dict[Node, float] = None
```
cburgdorf: Not as part of this PR but would it make sense to maintain a list of preferred nodes in the db or in some updated configuration file so that the list of preferred nodes becomes something well maintained (with newly discovered peers added and stale peers removed) to not fall back to the hardcoded nodes whenever there's a restart?

cburgdorf: :+1: https://github.com/ethereum/py-evm/issues/757

p2p/peer.py
pipermerriam: I think `isort` which is implemented in one of our repos (maybe `eth-account`?) is the programatic solution for this.

PR #748, opened 2018-05-22
Move bootnode config to `ChainConfig` object

pipermerriam:
### What was wrong?

Various places where we are juggling a list of bootnodes.

### How was it fixed?

Moved this into the `ChainConfig` as a property which handles the logic for defaulting based on `network_id`

#### Cute Animal Picture

a3aea23f9f469c47c2517be5751a37f8

### DIFF
--- a/trinity/utils/chains.py
+++ b/trinity/utils/chains.py
@@ -1,4 +1,8 @@
 import os
+from typing import (
+    Tuple,
+    Union,
+)
 
 from eth_utils import (
     decode_hex,
@@ -15,6 +19,12 @@ from evm.chains.ropsten import (
     ROPSTEN_NETWORK_ID,
 )
 
+from p2p.kademlia import Node
+from p2p.constants import (
+    MAINNET_BOOTNODES,
+    ROPSTEN_BOOTNODES,
+)
+
 from trinity.constants import (
     SYNC_FULL,
     SYNC_LIGHT,
@@ -22,8 +32,6 @@ from trinity.constants import (
 from .xdg import (
     get_xdg_trinity_root,
 )
-
-from typing import Union
 
 
 DEFAULT_DATA_DIRS = {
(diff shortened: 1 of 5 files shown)


trinity/utils/chains.py
```diff
@@ -114,17 +122,32 @@ class ChainConfig:
...
+    bootstrap_nodes: Tuple[Node, ...] = None
+
     def __init__(self,
                  network_id: int,
                  data_dir: str=None,
                  nodekey_path: str=None,
                  nodekey: PrivateKey=None,
                  sync_mode: str=SYNC_FULL,
-                 port: int=30303) -> None:
+                 port: int=30303,
```
carver: Total sidetrack: this line change makes me wonder if we can improve the style guide for long function signatures. Something like:

```py
def a(
    b,
    c,
    d,
    e,
) -> ReturnType:
    do_something()
```

pipermerriam: I generally agree.  I've not found a solution I like for this and type hints are still something that I'm getting used to (and that we need to add to the style guide).

trinity/utils/chains.py
```diff
@@ -114,17 +122,32 @@ class ChainConfig:
...
 
+        if bootstrap_nodes is None:
+            if self.network_id == MAINNET_NETWORK_ID:
+                self.bootstrap_nodes = tuple(
+                    Node.from_uri(enode) for enode in MAINNET_BOOTNODES
+                )
+            elif self.network_id == ROPSTEN_NETWORK_ID:
+                self.bootstrap_nodes = tuple(
+                    Node.from_uri(enode) for enode in ROPSTEN_BOOTNODES
+                )
```
carver: Especially since we'll see this pattern, maybe a utility method is worth it. Or maybe just:
```py
               self.bootstrap_nodes = tuple(map(Node.from_uri, ROPSTEN_BOOTNODES))
```

pipermerriam: I'm actually looking to add support back into at least the high level interface classes to accept either format, but I wanted to move this direction first to make things strict before loosening them back up.

pipermerriam: `normalize_kademlia_nodes(*node_or_enodes)`

PR #751, opened 2018-05-22
Personal RPC module baseline/structure

mateuszk87:
### What was wrong?
Personal RPC module has not been implemented yet and the error messages make it not explicit to users. 

### How was it fixed?
The base structure has been added without any implementation yet.

### DIFF
--- a/trinity/rpc/main.py
+++ b/trinity/rpc/main.py
@@ -9,6 +9,7 @@ from evm.exceptions import (
 from trinity.rpc.modules import (
     Eth,
     EVM,
+    Personal,
     RPCModule,
 )
 
@@ -55,6 +56,7 @@ class RPCServer:
     module_classes = (
         Eth,
         EVM,
+        Personal,
     )
 
     def __init__(self, chain):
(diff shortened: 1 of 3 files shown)


carver:
I don't think the `personal` module will be implemented for quite a while (if at all). The intention in the near term is to not do *any* account management inside trinity.  So I don't think the `Personal` module stubbing is necessary.

> the error messages make it not explicit to users.

Which error were you seeing? I would have expected you to see:

> Module unavailable: personal

Any suggestions on how to improve that message to make it clear that RPC does not support that module?

mateuszk87:
Even if the modules should not be implemented, trinity should provide some hints for the users.
Currently, if I use http://web3py.readthedocs.io/en/stable/web3.personal.html only some cryptic exception is thrown. This applies for the other Management APIs, too.

pipermerriam:
Closing until we have some better sense for what we're doing in this realm. 

> Even if the modules should not be implemented, trinity should provide some hints for the users.

@mateuszk87 I would agree with you except that we're in pre-alpha so while we care about UX a lot, we can't address everything at the same time.

mateuszk87:
keep me in loop if you are ready. I am already implementing various managements api methods, maybe it will help: 
e.g. https://github.com/mateuszk87/py-evm/commit/a53eed7c7cf116733dea3ab172953c50ac8ff684

Issue #728, opened 2018-05-18
Pull `trinity` and `p2p` module out into dedicated `trinity` repository.


pipermerriam:
### What is wrong?

We want to do a public release of the Trinity client under the `trinity` package.  I wasn't able to find a way to do this without leaving a mess in this repository

### How can it be fixed?

I'd like to go ahead and extract the `trinity` and `p2p` modules to **both** migrate to https://github.com/ethereum/trinity

I would plan on copying the full git history to the new repository and then introducing commits to removed the `py-evm` code.

The new `trinity` package would remain pinned to an exact py-evm version to prevent breaking changes with new alpha releases on the `py-evm` side.

PR #753, opened 2018-05-22
Try separating trinity and evm packages

pipermerriam:
Fixes: https://github.com/ethereum/py-evm/issues/728

### What was wrong?

Trinity and Py-EVM are not the same python package but they live in the same repo.

### How was it fixed?

Added a `setup_trinity.py` file which can be used to release **only** the trinity code (maybe)

#### Cute Animal Picture

376740 xcitefun-baby-donkey-7

### DIFF
--- a/README-trinity.md
+++ b/README-trinity.md
@@ -1,0 +1,8 @@
+# Trinity
+
+[Documentation hosted by ReadTheDocs](http://py-evm.readthedocs.io/en/latest/)
+
+
+## Introducing Py-EVM
+
+Trinity is a client for the Ethereum blockchain written in Python.
--- a/README-trinity.md
+++ b/README-trinity.md
@@ -1,8 +1,0 @@
-# Trinity
-
-[Documentation hosted by ReadTheDocs](http://py-evm.readthedocs.io/en/latest/)
-
-
-## Introducing Py-EVM
-
-Trinity is a client for the Ethereum blockchain written in Python.
--- a/setup_trinity.py
+++ b/setup_trinity.py
@@ -9,7 +9,7 @@ setup(
     # NOT CURRENTLY APPLICABLE. VERSION BUMPS MANUAL FOR NOW
     version='0.1.0-alpha.2',
     description='The Trinity Ethereum Client',
-    author='Piper Merriam',
+    author='Ethereum Foundation',
     author_email='piper@pipermerriam.com',
     url='https://github.com/ethereum/py-evm',
     include_package_data=True,
--- a/setup_trinity.py
+++ b/setup_trinity.py
@@ -1,0 +1,35 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+from setuptools import setup
+
+
+setup(
+    name='trinity',
+    # *IMPORTANT*: Don't manually change the version here. Use the 'bumpversion' utility.
+    # NOT CURRENTLY APPLICABLE. VERSION BUMPS MANUAL FOR NOW
+    version='0.1.0-alpha.2',
+    description='The Trinity Ethereum Client',
+    author='Piper Merriam',
+    author_email='piper@pipermerriam.com',
+    url='https://github.com/ethereum/py-evm',
+    include_package_data=True,
+    py_modules=[],
+    install_requires=[
+        'py-evm[trinity,p2p]==0.2.0a18',
+    ],
+    license='MIT',
+    zip_safe=False,
+    keywords='ethereum blockchain evm trinity',
+    packages=[],
+    classifiers=[
+        'Development Status :: 2 - Pre-Alpha',
+        'Intended Audience :: Developers',
+        'License :: OSI Approved :: MIT License',
+        'Natural Language :: English',
+        'Programming Language :: Python :: 3.6',
+    ],
+    # trinity
+    entry_points={
+        'console_scripts': ['trinity=trinity:main'],
+    },
+)
--- a/trinity/__init__.py
+++ b/trinity/__init__.py
@@ -4,7 +4,8 @@ import pkg_resources
 try:
     __version__: str = pkg_resources.get_distribution("trinity").version
 except pkg_resources.DistributionNotFound:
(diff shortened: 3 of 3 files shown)


README-trinity.md
```diff
@@ -0,0 +1,8 @@
+# Trinity
```
cburgdorf: Not really sure if we need a `README-trinity.md` on the root level (unless it serves a special purpose that I might be overseeing). I think the `README.md` on the root level serves as explainer for the entire repo so even in a mono repository scenario there would only be one top level `README.md`.

setup_trinity.py
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env python
```
cburgdorf: I would feel better if each package had it's own `setup.py` (without any suffix) directly in a package sublevel directory. Makes packages more self-contained. Essentially something like this.

```
evm/
├── src/
│   ├── something.py
│   ├── ...
├── setup.py
├── README.md
p2p/
├── src/
│   ├── something.py
│   ├── ...
├── setup.py
├── README.md
trinity/
├── src/
│   ├── something.py
│   ├── ...
├── setup.py
├── README.md
README.md
tox.ini
...
```

This is the defacto standard in the JS/TS camp but I'm not sure if there is anything that makes this impractical in the Python ecosystem?

setup_trinity.py
```diff
@@ -0,0 +1,35 @@
...
+    # *IMPORTANT*: Don't manually change the version here. Use the 'bumpversion' utility.
+    # NOT CURRENTLY APPLICABLE. VERSION BUMPS MANUAL FOR NOW
+    version='0.1.0-alpha.2',
+    description='The Trinity Ethereum Client',
+    author='Piper Merriam',
+    author_email='piper@pipermerriam.com',
+    url='https://github.com/ethereum/py-evm',
+    include_package_data=True,
+    py_modules=[],
+    install_requires=[
```
cburgdorf: Related to my comment above, if this was a proper `setup.py` in a subdirectory I assume it would be recognized as such during a `pip install` so that all the install_requires` of trinity could really be moved here rather than delegating to `py-evm[trinity,p2p]`

setup_trinity.py
```diff
@@ -0,0 +1,35 @@
...
+from setuptools import setup
+
+
+setup(
+    name='trinity',
+    # *IMPORTANT*: Don't manually change the version here. Use the 'bumpversion' utility.
+    # NOT CURRENTLY APPLICABLE. VERSION BUMPS MANUAL FOR NOW
+    version='0.1.0-alpha.2',
+    description='The Trinity Ethereum Client',
+    author='Piper Merriam',
```
cburgdorf: Not trying to belittle your role but I think this should be something like "The Trinity developers" or maybe "Ethereum Foundation" but not sure about any legal implications. There are several other places that are hardcoded to you and @carver that I think should be resolved in a similar way.

trinity/__init__.py
```diff
@@ -1,7 +1,11 @@
 import pkg_resources
 
 # TODO: update this to use the `trinity` version once extracted from py-evm
-__version__: str = pkg_resources.get_distribution("py-evm").version
+try:
+    __version__: str = pkg_resources.get_distribution("trinity").version
+except pkg_resources.DistributionNotFound:
+    # mypy doesn't like that `__version__` is defined twice
+    __version__: str = pkg_resources.get_distribution("py-evm").version  # type: ignore
```
cburgdorf: How about defining `__version__` *before* the `try` instead of silencing mypy?

README-trinity.md
carver: I assume it's just an intermediate step toward what would be the `README.md` file in the new `trinity` repo.

trinity/__init__.py
carver: Do mean a bare type def'n, like:

```py
__version__: str
try:
    __version__ = pkg_resources.get_distribution("trinity").version
    # mypy doesn't like that `__version__` is defined twice
    __version__ = pkg_resources.get_distribution("py-evm").version
```

README-trinity.md
pipermerriam: This was a leftover from a previous design.  Removing for now.

setup_trinity.py
pipermerriam: I'm avoiding doing this, largely because I don't want to move the `evm` stuff, then move it back later, so I'm thinking of this as a short term (few months maybe) stop gap solution.

setup_trinity.py
pipermerriam: I don't forsee using this file for anything other than releases, hence the stubbed out nature of it and still relying on dependencies declared in the main `setup.py` file.

trinity/__init__.py
pipermerriam: All of these are equally ugly in my opinion so unless someone feel strongly leaving as is.  Stop-gap solution until this lives by itself in it's own repo.

setup_trinity.py
cburgdorf: *I know, I know*, while I'm looking at it from a *monorepos are so much easier to maintain* kind of perspective :wink: 

But I realize I may be alone in this camp so I'm fine to go with the majority vote.

trinity/__init__.py
cburgdorf: No strong feeling at all.

setup_trinity.py
pipermerriam: Your camp may very well be lonely 😄 

I've never tried the approach but I don't think I want to at this time.  Also, with github it seems like a CI and permissions quagmire.

cburgdorf: >Also, with github it seems like a CI and permissions quagmire

Yep, point taken. It's just much easier to handle PRs that touch different parts of the system, handle labels, manage releases and versions etc

Anyway, there are pros and cons for both approaches and I also have to admit that I have zero experience regarding Python and monorepos and what kind of special problems there may be.

I can happily party with the other camp, too ;)

Issue #665, opened 2018-05-08
Need timeouts and proper task-cancellation when dealing with trinity's proxied-services


gsalgado:
Currently our chain syncer implementation do things like:

```
await wait_with_token(chaindb.coro_import_block(block), token=self.cancel_token)
```

Where chaindb is a `multiprocessing` proxy that runs the actual block importing in a separate process. Since we're dealing with a separate process that may die halfway through importing the block, or hang completely and never respond, I believe we should always specify timeouts when dealing with them -- in the same way we do for network operations.

Also, by wrapping those with wait_with_token() only cancels the outer task and not the one that is actually doing the work, which is created inside coro_import_block() by the await loop.run_in_executor(...) call. Maybe we just need to pass the CancelToken to coro_import_block() and have it use that with wait_with_token() instead of us.

carver:
> I believe we should always specify timeouts when dealing with them -- in the same way we do for network operations.

Thoughts on blocking the shutdown of the db process until all of the other processes are closed out? Otherwise, we're likely to get a spread of `ConnectionResetError`, `BrokenPipeError`, etc. in addition to timeout errors.

gsalgado:
> Thoughts on blocking the shutdown of the db process until all of the other processes are closed out? Otherwise, we're likely to get a spread of ConnectionResetError, BrokenPipeError, etc. in addition to timeout errors.

Isn't that how we do it, currently? trinity.main() first kills the networking process and only after that process is terminated it moves on to kill the database process

carver:
Ah yes, on closer look of the logs, the problem seems more likely to be that asking the networking process to shut down is the problem itself. Here is an example of the connection getting reset:

(first line: networking process asked to shut down)
```
    INFO  06-02 12:15:23        main  Keyboard Interrupt: Stopping
   DEBUG  06-02 12:15:23     service  Cancelling <trinity.nodes.mainnet.MainnetFullNode object at 0x7fd6067672b0>
    INFO  06-02 12:15:23     service  <p2p.chain.FastChainSyncer object at 0x7fd606f41470> finished: Cancellation requested by MainnetFullNode token
    INFO  06-02 12:15:23       chain  Sync with ETHPeer <Node(0x3688@213.168.13.151)> aborted: Cancellation requested by FastChainSyncer:FullNodeSyncer:Server:MainnetFullNode token
   ERROR  06-02 12:15:23        sync  Unclean exit from FullNodeSyncer
Traceback (most recent call last):
[...]

PR #754, opened 2018-05-23
json-rpc support for full node

carver:
### What was wrong?

light and full nodes had some duplicate implementations, and the full node did not support requests over RPC

### How was it fixed?

Not done, but:

- unify the light and full node launch code some more
- launch the RPC server with the full node (reusing the same logic from the base node implementation)

Some open issues:

- shutdown is still ugly on the full node
- had to stop using process pool as the default executor, because I couldn't get it to respect the signal handler (also GvR said it was probably a bad idea)
- Directly initialized the Chain for use with the RPC server. Is that a bad idea? The `ChainProxy` didn't really support anything I needed.
- probably linting and testing, I was only testing at the shell

#### Cute Animal Picture

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -37,6 +37,7 @@ from p2p.exceptions import NoConnectedPeers, OperationCancelled
 from p2p.peer import BasePeer, ETHPeer, PeerPool, PeerPoolSubscriber
 from p2p.rlp import BlockBody, P2PTransaction
 from p2p.service import BaseService
+from p2p.utils import unclean_close_exceptions
 
 
 class FastChainSyncer(BaseService, PeerPoolSubscriber):
@@ -112,10 +113,6 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
 
     async def handle_msg(self, peer: ETHPeer, cmd: protocol.Command,
                          msg: protocol._DecodedMsgType) -> None:
-        known_exceptions = (
-            BrokenPipeError,
-            ConnectionResetError,
-        )
         try:
             await self._handle_msg(peer, cmd, msg)
         except OperationCancelled:
@@ -123,8 +120,8 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             # with ensure_future()). Our caller will also get an OperationCancelled anyway, and
             # there it will be handled.
             pass
-        except known_exceptions as exc:
-            self.logger.warn("Halted while handling message from %s with %r", peer, exc)
+        except unclean_close_exceptions:
+            self.logger.exception("Unclean exit while handling message from %s", peer)
         except Exception:
             self.logger.exception("Unexpected error when processing msg from %s", peer)
 
@@ -160,8 +157,8 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             await self._sync(peer)
         except OperationCancelled:
             pass
-        except EOFError as exc:
-            logging.warn("Chain syncer halt, probably because of a dropped db connection, %r", exc)
+        except unclean_close_exceptions:
+            self.logger.exception("Unclean exit while syncing")
         finally:
             self._syncing = False
 
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -112,6 +112,10 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
 
     async def handle_msg(self, peer: ETHPeer, cmd: protocol.Command,
                          msg: protocol._DecodedMsgType) -> None:
+        known_exceptions = (
+            BrokenPipeError,
+            ConnectionResetError,
+        )
         try:
             await self._handle_msg(peer, cmd, msg)
         except OperationCancelled:
@@ -119,6 +123,8 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             # with ensure_future()). Our caller will also get an OperationCancelled anyway, and
             # there it will be handled.
             pass
+        except known_exceptions:
+            self.logger.warn("Halted while handling message from %s", peer)
         except Exception:
             self.logger.exception("Unexpected error when processing msg from %s", peer)
 
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -123,8 +123,8 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             # with ensure_future()). Our caller will also get an OperationCancelled anyway, and
             # there it will be handled.
             pass
-        except known_exceptions:
-            self.logger.warn("Halted while handling message from %s", peer)
+        except known_exceptions as exc:
+            self.logger.warn("Halted while handling message from %s with %r", peer, exc)
         except Exception:
             self.logger.exception("Unexpected error when processing msg from %s", peer)
 
@@ -160,6 +160,8 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             await self._sync(peer)
         except OperationCancelled:
             pass
+        except EOFError as exc:
+            logging.warn("Chain syncer halt, probably because of a dropped db connection, %r", exc)
         finally:
             self._syncing = False
 
@@ -340,6 +342,7 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
         return self._request_block_parts(headers, self._send_get_receipts)
 
     async def wait_until_finished(self) -> None:
+        self.logger.info("Shutting down FastChainSyncer")
         start_at = time.time()
         # Wait at most 1 second for pending peers to finish.
         self.logger.info("Waiting for %d running peers to finish", len(self._running_peers))
--- a/p2p/lightchain.py
+++ b/p2p/lightchain.py
@@ -1,6 +1,5 @@
 import asyncio
 import logging
-import time
 from typing import (
     cast,
     Dict,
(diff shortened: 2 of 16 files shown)


trinity/chains/mainnet.py
```diff
@@ -1,5 +1,6 @@
-from evm.chains.mainnet import (
+from evm.chains.mainnet import (  # noqa: F401
     BaseMainnetChain,
+    MainnetChain,
```
pipermerriam: I assume so that you can import it from here?  That feels weird if it's the case.

trinity/main.py
```diff
@@ -205,7 +169,16 @@ class DBManager(BaseManager):
 
 
 @with_queued_logging
-def run_lightnode_process(chain_config: ChainConfig) -> None:
+def launch_node(chain_config: ChainConfig) -> None:
+    display_launch_logs(chain_config)
+
+    node_class = choose_node_class(chain_config)
```
pipermerriam: Maybe condense to `chain_config.get_node()`.  The `choose_node_class` looks a lot like something that the `ChainConfig` class should handle (though it may need to leave the `utils` namespace soon since `utils` is not supposed to import from outside of `utils`.)

trinity/nodes/base.py
```diff
@@ -61,15 +96,15 @@ def make_ipc_server(self) -> IPCServer:
         self._ipc_server, self._ipc_loop = ipc_server, ipc_loop
 
         try:
-            asyncio.ensure_future(self._peer_pool.run())
             asyncio.run_coroutine_threadsafe(ipc_server.run(loop=ipc_loop), loop=ipc_loop)
-            await self._peer_chain.run()
+
+            async with RunningServices(self._auxiliary_services):
```
pipermerriam: It feels like the `self._auxiliary_services` list should get locked down once things begin running.  Otherwise we could end up in a weird situation where a new service is added after things are running.

trinity/nodes/base.py
```diff
@@ -82,3 +117,20 @@ def start_loop(loop):
...
+    # Typeshed definitions for multiprocessing.managers is incomplete, so ignore them for now:
+    # https://github.com/python/typeshed/blob/85a788dbcaa5e9e9a62e55f15d44530cd28ba830/stdlib/3/multiprocessing/managers.pyi#L3
+    DBManager.register('get_db', proxytype=DBProxy)  # type: ignore
+    DBManager.register('get_chaindb', proxytype=ChainDBProxy)  # type: ignore
+    DBManager.register('get_chain', proxytype=ChainProxy)  # type: ignore
+    DBManager.register('get_headerdb', proxytype=AsyncHeaderDBProxy)  # type: ignore
+    DBManager.register('get_header_chain', proxytype=AsyncHeaderChainProxy)  # type: ignore
+
+    manager = DBManager(address=ipc_path)  # type: ignore
+    manager.connect()  # type: ignore
```
pipermerriam: Maybe the `connect()` should be moved out of this function, or rename to `create_and_connect_db_manager()` or `get_db_manager` or `construct_db_manager`.  Something about `create` and the class being named `DBManager` makes me think it's just going to return the class.

trinity/nodes/full.py
```diff
@@ -0,0 +1,45 @@
...
+        self._node_key = chain_config.nodekey
+        self._node_port = chain_config.port
+
+    def get_chain(self):
+        if self._chain is None:
+            self._chain = self.chain_class(self.db_manager.get_db())
+
+        return self._chain
+
+    def get_peer_service(self) -> BaseService:
```
pipermerriam: The term `peer_service` seems a little off here.  Maybe something like `get_p2p_server` or `get_networking_server`?  Feels like the name `server` should be present.

p2p/utils.py
```diff
@@ -30,3 +34,16 @@ def get_devp2p_cmd_id(msg: bytes) -> int:
     as an integer.
     """
     return rlp.decode(msg[:1], sedes=rlp.sedes.big_endian_int)
+
+
+class RunningServices:
```
pipermerriam: Maybe this should be a subclass of [`collections.UserList`](https://docs.python.org/3/library/collections.html#collections.UserList) so that you can use it on the `Node` as the actual `_auxiliary_services` property as well as locking down the list items while the context manager is active.

trinity/chains/mainnet.py
carver: Yeah, it seemed more natural for the nodes to import both the `MainnetLightDispatchChain` and the `MainnetChain` from the same trinity module. But I don't feel strongly about it, so I'll roll it back.

trinity/main.py
carver: #759 - moving ChainConfig and adding the node class lookup

I'll also see if it makes sense to separate out another PR with the Node refactor that lets you initialize with only the chain config. I think that might be too tied into the rest of this PR to make sense, though.

trinity/nodes/base.py
carver: `launch_db_manager()`?

trinity/chains/mainnet.py
pipermerriam: I think my preference would be to:

1. Try to keep this practice constrained to `__init__` files.
2. Explicitly comment anytime we do it outside of an `__init__` file (but I'm kind of inclined to just veto this as something we do)

trinity/nodes/base.py
pipermerriam: Historically I feel like I always *wish* I'd separated `construct -> run` into individual units of functionality.

carver: Yeah, there are a number of things that I would like to wrap up into a `BaseService`. DBManager might be one of them. Then it should have a clear: init/run/stop cycle.

carver:
Ok, I think this is ready for another look @pipermerriam . The logs stayed relatively clean, and now both full and light mode supports RPC requests (well, at least `eth_getBlock*`). I would normally have split this into a few smaller pieces, but I am starting to run out of time before I sign off for a while, and wanted to get this in.

trinity/nodes/base.py
```diff
@@ -9,39 +12,71 @@
...
 
     @abstractmethod
     def get_chain(self) -> BaseChain:
         raise NotImplementedError("Node classes must implement this method")
 
+    @abstractmethod
+    def get_p2p_server(self) -> BaseService:
+        """
+        This is the main service that will be run, when calling :meth:`run`.
+        It's typically responsible for syncing the chain, with peer connections.
```
gsalgado: Thoughts on eventually combining `p2p.Server` with `trinity.Node`?

p2p/sync.py
```diff
@@ -39,6 +40,14 @@ def __init__(self,
         self.peer_pool = peer_pool
 
     async def _run(self) -> None:
+        try:
+            await self._sync()
+        except OperationCancelled:
```
gsalgado: You shouldn't catch `OperationCancelled` here as `p2p.BaseService` will do that and log it

p2p/service.py
```diff
@@ -32,28 +35,47 @@ def __init__(self, token: CancelToken=None) -> None:
...
+                raise
+
             if finished_callback is not None:
                 finished_callback(self)
 
     async def cleanup(self) -> None:
         """Run the service's _cleanup() coroutine."""
+        # Set self.finished before anything else so that other coroutines started by this
+        # service exit while we wait for cleanup().
+        self.finished.set()
```
gsalgado: I think it'd make more sense to leave this at the end of `run()` as it is not really part of any cleanup

p2p/chain.py
```diff
@@ -112,13 +112,19 @@ def register_peer(self, peer: BasePeer) -> None:
...
+        )
         try:
             await self._handle_msg(peer, cmd, msg)
         except OperationCancelled:
             # Silently swallow OperationCancelled exceptions because we run unsupervised (i.e.
             # with ensure_future()). Our caller will also get an OperationCancelled anyway, and
             # there it will be handled.
             pass
+        except known_exceptions as exc:
+            self.logger.warn("Halted while handling message from %s with %r", peer, exc)
```
gsalgado: I'd really prefer if those were caught closer to the code that originated them, as these should definitely not be bubbling up from `Peer` methods, so I suspect they're coming from trinity's proxy services when they're shut down and this would not be needed if we were able to shut everything down cleanly ([which is currently not the case](https://github.com/ethereum/py-evm/issues/665))

p2p/sync.py
```diff
@@ -39,6 +40,14 @@ def __init__(self,
...
+            pass
+        except BrokenPipeError as exc:
+            self.logger.warning("FullNodeSyncer exited because of a dropped connection %r", exc)
```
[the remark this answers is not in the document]
carver: I think the DB got closed in the middle of a `persist_trie_data_dict`.

trinity/nodes/base.py
carver: I think it makes sense for there to be an orchestration service in p2p (like `Server`) that manages the several services it needs to run. But I also think it makes sense to have an umbrella orchestration service in trinity (like `Node`), which manages several services (including `Server`). My instinct is that some services should not be managed inside p2p, but I think we already have one example: the RPC server.

But... there are some things handled in `Node` right now that could/should be pushed into `Server` (especially the light node, which explicitly manages the peer pool, at the moment).

p2p/chain.py
carver: Fair enough. Unfortunately one crash is causing a snowball of all the other cleanups to fail. 

I'll log out the full exception, which will have almost the same effect as not catching it (but will allow the cleanups to complete).

p2p/lightchain.py
```diff
@@ -45,22 +44,25 @@
...
+from p2p.service import (
+    BaseService,
+)
 
 if TYPE_CHECKING:
     from trinity.db.header import BaseAsyncHeaderDB  # noqa: F401
 
 
-class LightPeerChain(PeerPoolSubscriber):
+class LightPeerChain(BaseService, PeerPoolSubscriber):
```
pipermerriam: *nitpick*

The class ordering on this is backwards from what I expected.  `PeerPoolSubscriber` is a mixin interface and I've normally put these in the front.  Do you have the opposite convention?

p2p/service.py
```diff
@@ -88,3 +114,40 @@ class EmptyService(BaseService):
...
+
+    def __init__(self, services: List[BaseService] = None) -> None:
+        if services is None:
+            super().__init__()
+        else:
+            super().__init__(services)
+        self.started_services: List[BaseService] = []
+        self._run_lock = asyncio.Lock()
+
+    async def __aenter__(self):
```
pipermerriam: TIL `__aenter__` and `__aexit__` are things.

trinity/nodes/base.py
```diff
@@ -82,3 +114,23 @@ def start_loop(loop):
         thread.start()
 
         return new_loop
+
+
+def create_db_manager(ipc_path: str) -> BaseManager:
```
pipermerriam: We should use a `PurePath` here and downcast to a `str` before passing into the `DBManager`

PR #755, opened 2018-05-23
Cleaning up consistency of PurePath utilizations

glaksmono:
### What was wrong?
Clean up the internal consistency and utilizations of `PurePath`

Issue Reference: https://github.com/ethereum/py-evm/pull/734

#### Cute Animal Picture

### DIFF
--- a/trinity/utils/chains.py
+++ b/trinity/utils/chains.py
@@ -6,8 +6,6 @@ from typing import (
 
 from pathlib import PurePath
 
-from pathlib import PurePath
-
 from eth_utils import (
     decode_hex,
     to_dict,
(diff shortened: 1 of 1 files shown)


glaksmono:
@pipermerriam @carver 

```
trinity/utils/chains.py:119: error: "PurePath" has no attribute "open"
trinity/utils/chains.py:185: error: "PurePath" has no attribute "resolve"
trinity/utils/chains.py:216: error: "PurePath" has no attribute "resolve"
```

So it seems like `open` and `resolve` are the methods of the `PurePath`'s child namely `Path`: [PurePath Inheritance](https://docs.python.org/3/_images/pathlib-inheritance.png)

PR #756, opened 2018-05-23
Convert PurePath to Path. Cleaning up consistency of Path utilizations.

glaksmono:
### What was wrong?
Convert `PurePath` to `Path`. Cleaning up consistency of `Path` utilizations.

Issue Reference: https://github.com/ethereum/py-evm/pull/734

#### Cute Animal Picture

### DIFF
--- a/trinity/utils/chains.py
+++ b/trinity/utils/chains.py
@@ -1,5 +1,5 @@
 import os
-from pathlib import PurePath
+from pathlib import Path
 
 from eth_utils import (
     decode_hex,
@@ -30,17 +30,17 @@ DEFAULT_DATA_DIRS = {
 #
 # Filesystem path utils
 #
-def get_local_data_dir(chain_name: str) -> PurePath:
+def get_local_data_dir(chain_name: str) -> Path:
     """
     Returns the base directory path where data for a given chain will be stored.
     """
-    return PurePath(os.environ.get(
+    return Path(os.environ.get(
         'TRINITY_DATA_DIR',
         os.path.join(get_xdg_trinity_root(), chain_name),
     ))
 
 
-def get_data_dir_for_network_id(network_id: int) -> PurePath:
+def get_data_dir_for_network_id(network_id: int) -> Path:
     """
     Returns the data directory for the chain associated with the given network
     id.  If the network id is unknown, raises a KeyError.
@@ -54,11 +54,11 @@ def get_data_dir_for_network_id(network_id: int) -> PurePath:
 NODEKEY_FILENAME = 'nodekey'
 
 
-def get_nodekey_path(data_dir: PurePath) -> PurePath:
+def get_nodekey_path(data_dir: Path) -> Path:
     """
     Returns the path to the private key used for devp2p connections.
     """
-    return PurePath(os.environ.get(
+    return Path(os.environ.get(
         'TRINITY_NODEKEY',
         str(data_dir / NODEKEY_FILENAME),
     ))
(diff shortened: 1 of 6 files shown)


glaksmono:
@pipermerriam @carver 

```
trinity/utils/chains.py:119: error: "PurePath" has no attribute "open"
trinity/utils/chains.py:185: error: "PurePath" has no attribute "resolve"
trinity/utils/chains.py:216: error: "PurePath" has no attribute "resolve"
```

So it seems like `open` and `resolve` are the methods of the `PurePath`'s child namely `Path`: [PurePath Inheritance](https://docs.python.org/3/library/pathlib.html#module-pathlib)

This makes me think if we should use `Path` instead of `PurePath`, thoughts?

cburgdorf:
>This makes me think if we should use `Path` instead of `PurePath`, thoughts

:+1: I think that's correct.

Citing from the docs,

>If you’ve never used this module before or just aren’t sure which class is right for your task, `Path` is most likely what you need. It instantiates a concrete path for the platform the code is running on.

It goes on about `PurePath`

>Pure paths are useful in some special cases; for example:

>1. If you want to manipulate Windows paths on a Unix machine (or vice versa). You cannot instantiate a WindowsPath when running on Unix, but you can instantiate PureWindowsPath.

>2. You want to make sure that your code only manipulates paths without actually accessing the OS. In this case, instantiating one of the pure classes may be useful since those simply don’t have any OS-accessing operations.

glaksmono:
Makes sense. Decided to convert `PurePath` to `Path`.

@pipermerriam @carver @cburgdorf

trinity/utils/chains.py
```diff
@@ -83,48 +83,48 @@ def get_nodekey_path(data_dir: PurePath) -> PurePath:
...
     Returns the path to the ipc socket for the JSON-RPC server.
 
     We're still returning 'str' here on ipc-related path because an issue with
-    multi-processing not being able to interpret 'PurePath' objects correctly.
+    multi-processing not being able to interpret 'Path' objects correctly.
     """
     return os.environ.get(
         'TRINITY_JSONRPC_IPC',
-        os.path.join(str(data_dir), JSONRPC_SOCKET_FILENAME),
+        os.path.join(str(data_dir / JSONRPC_SOCKET_FILENAME)),
```
cburgdorf: This doesn't look correct to me

trinity/utils/chains.py
```diff
@@ -83,48 +83,48 @@ def get_nodekey_path(data_dir: PurePath) -> PurePath:
...
     Returns the path to the private key used for devp2p connections.
 
     We're still returning 'str' here on ipc-related path because an issue with
-    multi-processing not being able to interpret 'PurePath' objects correctly.
+    multi-processing not being able to interpret 'Path' objects correctly.
     """
     return os.environ.get(
         'TRINITY_DATABASE_IPC',
-        os.path.join(str(data_dir), DATABASE_SOCKET_FILENAME),
+        os.path.join(str(data_dir / DATABASE_SOCKET_FILENAME)),
```
cburgdorf: This looks wrong

trinity/utils/chains.py
glaksmono: My bad.

trinity/utils/chains.py
```diff
@@ -83,48 +83,48 @@ def get_nodekey_path(data_dir: PurePath) -> PurePath:
...
     Returns the path to the private key used for devp2p connections.
 
     We're still returning 'str' here on ipc-related path because an issue with
-    multi-processing not being able to interpret 'PurePath' objects correctly.
+    multi-processing not being able to interpret 'Path' objects correctly.
     """
     return os.environ.get(
         'TRINITY_DATABASE_IPC',
-        os.path.join(str(data_dir), DATABASE_SOCKET_FILENAME),
+        str(data_dir / DATABASE_SOCKET_FILENAME),
```
cburgdorf: What I meant to say [here](https://github.com/ethereum/py-evm/pull/756#discussion_r190263967) was that the `/` looks wrong to me. To be honest, I didn't even know this syntax was possible because for simple strings this would result in an error:

```python
str('foo' / 'bar')
TypeError: unsupported operand type(s) for /: 'str' and 'str'
```

So, I just learned that with `Path` that syntax is in fact possible

```Python
str(Path('foo') / 'bar')
'foo/bar'
```

However, wouldn't something like `os.path.join(data_dir, 'foo')` still be clearer? If that's an accepted Python thing, I'm happy to keep it the way it is.

@pipermerriam @carver

trinity/utils/chains.py
```diff
@@ -83,48 +83,48 @@ def get_nodekey_path(data_dir: PurePath) -> PurePath:
...
     Returns the path to the ipc socket for the JSON-RPC server.
 
     We're still returning 'str' here on ipc-related path because an issue with
-    multi-processing not being able to interpret 'PurePath' objects correctly.
+    multi-processing not being able to interpret 'Path' objects correctly.
     """
     return os.environ.get(
         'TRINITY_JSONRPC_IPC',
-        os.path.join(str(data_dir), JSONRPC_SOCKET_FILENAME),
+        str(data_dir / JSONRPC_SOCKET_FILENAME),
```
cburgdorf: See my comment above

glaksmono:
Done rebased @pipermerriam @cburgdorf

trinity/utils/chains.py
pipermerriam: Yeah, I visually rejected it the first time I saw it as well.  A full read through the pathlib docs has me convinced this is the right way to go across the board with path manipulation.

pipermerriam:
Looks like #759 caused another merge conflict (kind of a big one).  I'll clean it up tomorrow if it's still pending.

glaksmono:
Fixed it @pipermerriam

cburgdorf:
I'll go ahead and merge this based on @pipermerriam 's :+1: and to save you from more rebases while America is still sleeping.

Issue #762, opened 2018-05-23
`AlreadyWaiting` exception during initial discovery.


pipermerriam:
### What is wrong?

```python
   ERROR  05-23 16:16:10  base_events  Task exception was never retrieved
future: <Task finished coro=<PeerPool.lookup_random_node() done, defined at /Users/piper/sites/py-evm/p2p/peer.py:741> exception=AlreadyWaiting("There's another coroutine waiting for a neighbours packet from <Node(0x9124@148.251.19.50)>",)>
Traceback (most recent call last):
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 745, in lookup_random_node
    await self.discovery.lookup_random(self.cancel_token)
  File "/Users/piper/sites/py-evm/p2p/discovery.py", line 93, in lookup_random
    return await self.kademlia.lookup_random(self.cancel_token.chain(cancel_token))
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 575, in lookup_random
    return await self.lookup(random.randint(0, k_max_node_id), cancel_token)
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 565, in lookup
    *[_find_node(node_id, n) for n in nodes_to_ask])
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 541, in _find_node
    candidates = await self.wait_neighbours(remote, cancel_token)
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 461, in wait_neighbours
    "There's another coroutine waiting for a neighbours packet from {}".format(remote))
p2p.kademlia.AlreadyWaiting: There's another coroutine waiting for a neighbours packet from <Node(0x9124@148.251.19.50)>
```

looks like a race condition of some kind which results in stacktrace being dumped to terminal.

pipermerriam:
Fixed in #765

Issue #737, opened 2018-05-20
Trinity raises `AlreadyWaiting` error semi-regularly


pipermerriam:
### What is wrong?

Encountered when running in full/fast sync mode.

```bash
   ERROR  05-20 15:16:13  base_events  Task exception was never retrieved
future: <Task finished coro=<PeerPool.lookup_random_node() done, defined at /Users/piper/sites/py-evm/p2p/peer.py:719> exception=AlreadyWaiting("There's another coroutine waiting for a ping packet from <Node(0x47ca@165.227.61.121)>",)>
Traceback (most recent call last):
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 723, in lookup_random_node
    await self.discovery.lookup_random(self.cancel_token)
  File "/Users/piper/sites/py-evm/p2p/discovery.py", line 90, in lookup_random
    return await self.kademlia.lookup_random(self.cancel_token.chain(cancel_token))
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 575, in lookup_random
    return await self.lookup(random.randint(0, k_max_node_id), cancel_token)
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 565, in lookup
    *[_find_node(node_id, n) for n in nodes_to_ask])
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 550, in _find_node
    bonded = await asyncio.gather(*[self.bond(c, cancel_token) for c in candidates])
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 514, in bond
    await self.wait_ping(node, cancel_token)
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 413, in wait_ping
    "There's another coroutine waiting for a ping packet from {}".format(remote))
p2p.kademlia.AlreadyWaiting: There's another coroutine waiting for a ping packet from <Node(0x47ca@165.227.61.121)>
```

This exception happens semi-regularly when syncing.

gsalgado:
This is something that needs fixing, for sure, but it should only happen if we run a standard `PeerPool`, which uses discovery to find nodes, but that's probably not a good idea as in my tests the discovery v4 never seems to work at finding LES peers. And actually, trinity seems to use a `HardcodedNodesPeerPool` for both light/full modes, so I'm guessing you're running a locally-modified version?

pipermerriam:
Yes, I am running with `PeerPool`

pipermerriam:
[...]

PR #765, opened 2018-05-24
Handle errors during sync

### LINKED ISSUE #760: Handle empty node list in `PreferredPeerPool`
### What is wrong?

This happens

```python
   ERROR  05-23 16:09:45     service  Unexpected error in <p2p.peer.PreferredNodePeerPool object at 0x10f68bc88>, exiting
  File "/Users/piper/sites/py-evm/p2p/service.py", line 36, in run
    await self._run()
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 697, in _run
    await self.maybe_connect_to_more_peers()
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 763, in maybe_connect_to_more_peers
    await self._connect_to_nodes(self.get_nodes_to_connect())
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 777, in _connect_to_nodes
    for node in nodes:
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 954, in get_nodes_to_connect
    yield from self.discovery.get_random_nodes(num_nodes_needed)
  File "/Users/piper/sites/py-evm/p2p/kademlia.py", line 240, in get_random_nodes
    node = random.choice(bucket.nodes)
  File "/Users/piper/python-environments/trinity/lib/python3.6/random.py", line 257, in choice
    raise IndexError('Cannot choose from an empty sequence') from None
IndexError: Cannot choose from an empty sequence
```

### How can it be fixed

Gracefully handle this situation.

pipermerriam:
fixes #760 #762 #737
replaces #764 #763 #761

### What was wrong?

Errors that occur during sync

### How was it fixed?

Add condition checks.

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -466,10 +466,6 @@ class KademliaProtocol:
         called or a timeout (k_request_timeout) occurs. At that point it returns whether or not
         a ping was received from the given node.
         """
-        if remote in self.ping_callbacks:
-            raise AlreadyWaiting(
-                "There's another coroutine waiting for a ping packet from {}".format(remote))
-
         event = asyncio.Event()
 
         with self.ping_callbacks.acquire(remote, event.set):
@@ -491,12 +487,6 @@ class KademliaProtocol:
         a pong was received with the given pingid.
         """
         pingid = self._mkpingid(token, remote)
-        if pingid in self.pong_callbacks:
-            raise AlreadyWaiting(
-                "There's another coroutine waiting for a pong packet with id "
-                "{}".format(pingid)
-            )
-
         event = asyncio.Event()
 
         with self.pong_callbacks.acquire(pingid, event.set):
@@ -519,10 +509,6 @@ class KademliaProtocol:
 
         Returns the list of neighbours received.
         """
-        if remote in self.neighbours_callbacks:
-            raise AlreadyWaiting(
-                "There's another coroutine waiting for a neighbours packet from {}".format(remote))
-
         event = asyncio.Event()
         neighbours: List[Node] = []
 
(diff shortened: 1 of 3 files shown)


p2p/kademlia.py
```diff
@@ -516,7 +517,12 @@ def ping(self, node: Node) -> bytes:
         return True
 
     async def bootstrap(self, bootstrap_nodes: List[Node], cancel_token: CancelToken) -> None:
-        bonded = await asyncio.gather(*[self.bond(n, cancel_token) for n in bootstrap_nodes])
+        bonded = await asyncio.gather(*(
+            self.bond(n, cancel_token)
+            for n
+            in bootstrap_nodes
+            if (n not in self.ping_callbacks and n not in self.pong_callbacks)
```
gsalgado: I'm slightly concerned about these changes because if there's ever an API (and we might have one already) that does not clean up those callback lists as it should, we'll have a memory leak that will be hard to track. More importantly, though, we cold end up in a situation where a node with which we fail to bond once will keep failing forever because we think there's a bond in progress with it. 

Instead, I'd like to suggest we change `PeerPool` so that it never runs multiple discovery lookups in parallel, as that, together with the fact that `KademliaProtocol.lookup()` does not try to bond with peers more than once, should stop those `AlreadyWaiting` errors. I've just created #767 for that

Would you be ok with shelving just the changes that check whether a node is already present in a callback list and trying #767 for a while to see if it works as I expect?

p2p/kademlia.py
```diff
@@ -320,16 +325,64 @@ def binary_get_bucket_for_node(buckets: List[KBucket], node: Node) -> KBucket:
         raise ValueError("No bucket found for node with id {}".format(node.id))
 
 
+class CallbackLock:
+    def __init__(self,
+                 callback: Callable,
+                 timeout: int=300) -> None:
```
gsalgado: How about making the default here something like `2 * k_request_timeout` as by then we can be sure nobody should be waiting for the callback anyway?

pipermerriam:
@gsalgado if you have follow up review on the extra commits I'll address it in another PR

PR #766, opened 2018-05-24
Sync eth_keys stub with actual code

cburgdorf:
### What was wrong?

While working on the upcoming benchmarking suite I ran into the problem that the `eth_keys` stubs were out of date and incomplete. This hasn't pop up in the rest of the code base yet because their the type checks are less strict compared to the type checks that ran against the benchmarking suite.

### How was it fixed?

Synced type stubs with actual code

#### Cute Animal Picture


pipermerriam:
We need to spin up an `eth-typeshed` repo that we can use everywhere as a submodule.  That should allow us to catch things going out of sync in CI with some reliability.

PR #767, opened 2018-05-24
p2p: A couple fixes to KademliaProtocol/PeerPool

gsalgado:
- Ensure PeerPool never runs multiple node discovery lookups in parallel
- KademliaProtocol.bond() no longer triggers
  self.populate_not_full_buckets() when bondig fails as that can
  potentially create a never-ending bonding spiral

The `AlreadyWaiting` errors (#737) seem to be gone with these changes

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -647,8 +647,9 @@ class PeerPool(BaseService):
     """PeerPool attempts to keep connections to at least min_peers on the given network."""
     logger = logging.getLogger("p2p.peer.PeerPool")
     _connect_loop_sleep = 2
-    _last_lookup: float = 0
-    _lookup_interval: int = 20
+    _discovery_lookup_running = asyncio.Lock()
+    _discovery_last_lookup: float = 0
+    _discovery_lookup_interval: int = 30
 
     def __init__(self,
                  peer_class: Type[BasePeer],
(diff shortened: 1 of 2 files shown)


p2p/peer.py
```diff
@@ -738,14 +739,22 @@ def add_peer(self, peer):
             self.logger.exception("Unexpected error during auth/p2p handshake with %s", remote)
         return None
 
-    async def lookup_random_node(self) -> None:
+    async def maybe_lookup_random_node(self) -> None:
+        if self._discovery_last_lookup + self._discovery_lookup_interval > time.time():
+            return
+        if self._discovery_lookup_running:
```
pipermerriam: nitpick

Can you make this an `elif`

p2p/peer.py
```diff
@@ -738,14 +739,22 @@ def add_peer(self, peer):
...
+            self.logger.debug("Node discovery lookup already in progress, not running another")
+            return
+        self._discovery_lookup_running = True
```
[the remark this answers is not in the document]
gsalgado: Definitely! Should've thought of that myself

p2p/peer.py
gsalgado: Sure thing

p2p/peer.py
```diff
@@ -738,14 +739,21 @@ def add_peer(self, peer):
...
-        except OperationCancelled:
-            pass
-        self._last_lookup = time.time()
+    async def maybe_lookup_random_node(self) -> None:
+        if self._discovery_last_lookup + self._discovery_lookup_interval > time.time():
+            return
+        elif self._discovery_lookup_running.locked():
+            self.logger.debug("Node discovery lookup already in progress, not running another")
+            return
+        with (await self._discovery_lookup_running):
```
pipermerriam: is this equvalent to `async with self.discovery_lookup_running`?

That's what I found here: https://www.pythonsheets.com/notes/python-asyncio.html

gsalgado: Looks like it: https://www.python.org/dev/peps/pep-0492/#asynchronous-context-managers-and-async-with

I'll switch to `async with`; I think that reads better

PR #769, opened 2018-05-24
p2p: RegularChainSyncer now serves peer requests as well

gsalgado:
Up until now we'd only fetch chain/state data, ignoring any peer requests for data

We now handle chain/state data requests from other peers.

Also adds tests for chain syncing and state downloading

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -16,6 +16,7 @@ from typing import (
 
 from cytoolz.itertoolz import partition_all, unique
 
+from eth_typing import BlockNumber, Hash32
 from eth_utils import (
     encode_hex,
 )
@@ -34,6 +35,7 @@ from p2p import eth
 from p2p.cancel_token import CancelToken, wait_with_token
 from p2p.exceptions import NoConnectedPeers, OperationCancelled
 from p2p.peer import BasePeer, ETHPeer, PeerPool, PeerPoolSubscriber
+from p2p.rlp import BlockBody, P2PTransaction
 from p2p.service import BaseService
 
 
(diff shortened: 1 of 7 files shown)


p2p/chain.py
```diff
@@ -421,9 +429,56 @@ def __init__(self,
...
+        elif isinstance(cmd, eth.GetNodeData):
+            self._handle_get_node_data(peer, cast(List[Hash32], msg))
         else:
             self.logger.debug("%s msg not handled yet, need to be implemented", cmd)
 
+    def _handle_get_block_headers(self, peer: ETHPeer, msg: Dict[str, Any]) -> None:
+        n = msg['block_number_or_hash']
+        if isinstance(n, bytes):
+            header = self.chaindb.get_block_header_by_hash(cast(Hash32, n))
+            n = header.block_number
```
pipermerriam: This pattern of setting `n` above and then overriding in here is hard for me to parse.  I initially thought there was a bug until I realized what was going on.

Can we do something like `raw_block_number = msg['block_number_or_hash']` and then have:

```python
if isinstance(raw_block_number, bytes):
    block_number = raw_block_number
elif isinstance(raw_block_number, int):
    ...
else:
    raise TypeError("...")
```

That way we explicitly handle both allowed types.  (Also gets rid of the single letter variable name).

p2p/chain.py
```diff
@@ -421,9 +429,56 @@ def __init__(self,
...
+        limit = msg['max_headers']
+        if msg['reverse']:
+            block_numbers = list(reversed(range(max(0, n - limit), n + 1)))
+        else:
+            head_number = self.chaindb.get_canonical_head().block_number
+            block_numbers = list(range(n, min(head_number + 1, n + limit)))
+        headers = []
+        for i in block_numbers:
+            headers.append(self.chaindb.get_canonical_block_header_by_number(cast(BlockNumber, i)))
```
pipermerriam: Can we do this as a list comprehension instead of using `append`?

```python
headers = [
    self.chaindb.get_canonical_block_header_by_number(cast(BlockNumber, i))
    for i in block_numbers
]
```

p2p/chain.py
```diff
@@ -421,9 +429,56 @@ def __init__(self,
...
+            self._handle_get_node_data(peer, cast(List[Hash32], msg))
         else:
             self.logger.debug("%s msg not handled yet, need to be implemented", cmd)
 
+    def _handle_get_block_headers(self, peer: ETHPeer, msg: Dict[str, Any]) -> None:
+        n = msg['block_number_or_hash']
+        if isinstance(n, bytes):
+            header = self.chaindb.get_block_header_by_hash(cast(Hash32, n))
+            n = header.block_number
+        limit = msg['max_headers']
```
pipermerriam: Should we have an upper limit on the number that we allow fetching in a single query?

gsalgado: Yep, we actually have the constant already, which is what we use when fetching and is the same limit imposed by geth. I'll use it here as well

Issue #773, opened 2018-05-24
Trinity warning message about unexpected parts dominates terminal


pipermerriam:
### What is wrong?

```bash
[...]

PR #774, opened 2018-05-24
quiet terminal output on unexpected parts

pipermerriam:
### What was wrong?

#773 

### How was it fixed?

- Use the full `headers` set to ensure unexpected parts are **really** unexpected.
- Separately log duplicate parts at a lower warning level.

#### Cute Animal Picture

030042507a2b4fb720f9c31de81b271b

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -389,21 +389,24 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             peer.head_td = actual_td
             self._sync_requests.put_nowait(peer)
 
-    async def _handle_block_receipts(
-            self, peer: ETHPeer, receipts: List[List[eth.Receipt]]) -> None:
-        self.logger.debug("Got Receipts for %d blocks from %s", len(receipts), peer)
+    async def _handle_block_receipts(self,
+                                     peer: ETHPeer,
+                                     receipts_by_block: List[List[eth.Receipt]]) -> None:
+        self.logger.debug("Got Receipts for %d blocks from %s", len(receipts_by_block), peer)
         loop = asyncio.get_event_loop()
-        iterator = map(make_trie_root_and_nodes, receipts)
+        iterator = map(make_trie_root_and_nodes, receipts_by_block)
         receipts_tries = await wait_with_token(
             loop.run_in_executor(None, list, iterator),
             token=self.cancel_token)
         downloaded: List[DownloadedBlockPart] = []
-        for (receipt, (receipt_root, trie_dict_data)) in zip(receipts, receipts_tries):
+        for (receipts, (receipt_root, trie_dict_data)) in zip(receipts_by_block, receipts_tries):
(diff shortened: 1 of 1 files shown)


pipermerriam:
Still running this locally for a bit to see if it sufficiently quiets the output or whether we need to drop the unexepected logging level down to a warning.

p2p/chain.py
```diff
@@ -284,13 +284,20 @@ def register_peer(self, peer: BasePeer) -> None:
             parts.extend(received)
             received_keys = set([part.unique_key for part in received])
             pending_replies -= 1
-            unexpected = received_keys.difference(
-                [key_func(header) for header in missing])
+
+            unexpected = received_keys.difference(key_func(header) for header in headers)
+            duplicates = received_keys.intersection(part.unique_key for part in parts)
```
carver: It looks weird to calculate duplicates here *after* adding `received` to `parts`. Seems like it will always include all the `received` keys.

pipermerriam:
Need to investigate some more.  Not sure if A) we just get a **lot** of unrequested data or B) something malicious is happening or C) we're not interpreting responses correctly.

pipermerriam:
So I ran this for 5-6 hours today (still getting lots of output from unexpected part).  I piped all of that out to a logfile.  Logfile is now 3.8gb.  Lots of output.

gsalgado:
This is very interesting... I have never got those warnings, and I've been running this for a long time. I'm now running it with the changes from https://github.com/ethereum/py-evm/pull/779 and a pdb breakpoint but is still hasn't happened, so I suspect you've connected to a malicious node!

gsalgado:
Another thing that suggests it was a malicious peer is that we currently use a 60s timeout when waiting for body parts and IME that pretty much eliminates the possibility of us receiving duplicate body parts on retries since after 30s either the peer will have dropped the connection (if our main loop is too busy and it doesn't hear from us) or we'll have done so if we don't hear from the peer

p2p/chain.py
```diff
@@ -281,16 +287,31 @@ def register_peer(self, peer: BasePeer) -> None:
                 pending_replies = request_func(missing)
                 continue
 
-            parts.extend(received)
+            received_keys_by_peer = groupby(operator.attrgetter('peer'), received)
```
gsalgado: All items in `received` will have the same 'peer' here. That's why in https://github.com/ethereum/py-evm/pull/779 I used a tuple (`(peer, items)`) instead of adding a peer attribute to `DownloadedBlockPart`

gsalgado: I think this code would be significantly simpler if you did the same

pipermerriam: 👍 I'll rebase on top of that.

pipermerriam:
@gsalgado I'll cleanup any issue you have with the latest changes post-merge.

gsalgado:
Nothing from me, looks great!

Issue #742, opened 2018-05-21
Enable logging to file by default for alpha releases


pipermerriam:
### What is wrong?

Users of the alpha releases are very likely going to encounter errors for which we need debug level logging output to diagnose.  They however are unlikely to be both logging at the `DEBUG` level **and** retaining log output.

### How can it be fixed

Lets enable file based logging in a `$XDG_TRINITY_ROOT/<chain-dir>/logs` at the `DEBUG` level.  This way when a user has an issue we can get them to dig out the relevant logfiles for us.

pipermerriam:
@glaksmono sorry for the confusion, updated to correctly reference `$XDG_TRINITY_ROOT`.

https://github.com/ethereum/py-evm/blob/master/trinity/utils/xdg.py

glaksmono:
@pipermerriam Would like to take a stab on this 🔪 

Do we want this file logging behavior to be the `default`?

pipermerriam:
@glaksmono just in case there was still any question, yes, I think this logging should be on by default.

- make use of the rotating file handler so that logfiles don't become too large.
- if it's possible to place an upper bound on the total size of the logs 👍 
- lets display in the terminal when trinity starts up where the logfile is located.

cburgdorf:
This seems to be solved. Please reopen if I you feel there's something unaddressed.

PR #777, opened 2018-05-25
Enable logging to file by default for alpha releases

glaksmono:
### What was wrong?

Users of the alpha releases are very likely going to encounter errors for which we need debug level logging output to diagnose. They however are unlikely to be both logging at the `DEBUG` level and retaining log output.

Issue Reference: https://github.com/ethereum/py-evm/issues/742

### How was it fixed?

Lets enable file based logging in a `$XDG_TRINITY_ROOT/<chain-dir>/logs` at the `DEBUG` level. This way when a user has an issue we can get them to dig out the relevant logfiles for us.

#### Cute Animal Picture

### DIFF
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -37,6 +37,7 @@ from trinity.utils.chains import (
     get_data_dir_for_network_id,
     get_database_socket_path,
     get_jsonrpc_socket_path,
+    get_logfile_path,
     get_nodekey_path,
     load_nodekey,
 )
@@ -47,6 +48,7 @@ DATABASE_DIR_NAME = 'chain'
 class ChainConfig:
     _data_dir: Path = None
     _nodekey_path: Path = None
+    _logfile_path: Path = None
     _nodekey = None
     _network_id: int = None
 
@@ -59,6 +61,7 @@ class ChainConfig:
                  network_id: int,
                  data_dir: str=None,
                  nodekey_path: str=None,
+                 logfile_path: str=None,
                  nodekey: PrivateKey=None,
                  sync_mode: str=SYNC_FULL,
                  port: int=30303,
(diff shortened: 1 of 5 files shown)


glaksmono:
Before fixing the `lint`, would like to check on the structure of the implementation. The current implementation write the log of `debug` to the file and `stdout` @pipermerriam @carver

trinity/utils/logging.py
```diff
@@ -8,30 +9,47 @@
...
-    log_queue = ctx.Queue()
+    log_queue_stream = ctx.Queue()
+    log_queue_file = ctx.Queue()
 
     logging.basicConfig(level=level)
     logger = logging.getLogger('trinity')
 
-    handler = logging.StreamHandler(sys.stdout)
+    log_dir = chain_config.data_dir / "logs"
+    log_file = log_dir / "trinity.log"
```
pipermerriam: Oh, and this logic should probably move into the `ChainConfig` class itself so that you can just do `ChainConfig.logfile_path` to access the appropriate path.

trinity/utils/logging.py
```diff
@@ -8,30 +8,44 @@
...
 
     log_queue = ctx.Queue()
 
     logging.basicConfig(level=level)
     logger = logging.getLogger('trinity')
 
-    handler = logging.StreamHandler(sys.stdout)
+    handler_stream = logging.StreamHandler(sys.stdout)
+    handler_file = logging.handlers.RotatingFileHandler(
+        str(chain_config.logfile_path), maxBytes=(1000000 * 5), backupCount=7
```
glaksmono: @pipermerriam what do you feel about this number?

trinity/config.py
```diff
@@ -96,6 +98,20 @@ def __init__(self,
         elif nodekey is not None:
             self.nodekey = nodekey
 
+        self.logfile_path = get_logfile_path(self.data_dir)
```
pipermerriam: It'd be nice if users could specify the path in the `__init__` and if it isn't provided then it falls back to this value similar to how `data_dir` gets set.

trinity/utils/logging.py
pipermerriam: I'd bump it up to 50mb filesize and 10 backups (max space 500mb).  Non trivial amount of storage but debug logs are going to be pretty verbose.

pipermerriam:
closing in favor of #802

PR #785, opened 2018-05-25
Return early in Kademlia.recv_ping if remote is this_node

pipermerriam:
### What was wrong?

Occasionally receiving a `ping` from `this_node`.

### How was it fixed?

Avoid trying to add `remote` to routing table if it is `this_node` and return early from `recv_ping` when `remote == this_node`.

#### Cute Animal Picture

163e48dbb409544f5332da5060194d52

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -427,7 +427,7 @@ class KademliaProtocol:
         """
         self.logger.debug('<<< ping from %s', remote)
         if remote == self.this_node:
-            self.logger.debug('Received ping from this_node: %s', remote)
+            self.logger.info('Invariant: received ping from this_node: %s', remote)
             return
         else:
             self.update_routing_table(remote)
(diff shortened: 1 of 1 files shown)


p2p/kademlia.py
```diff
@@ -426,7 +426,11 @@ def recv_ping(self, remote: Node, hash_: bytes) -> None:
         reply with a pong, whereas in the latter we'll also fire a callback from ping_callbacks.
         """
         self.logger.debug('<<< ping from %s', remote)
-        self.update_routing_table(remote)
+        if remote == self.this_node:
+            self.logger.debug('Received ping from this_node: %s', remote)
```
gsalgado: I wonder how this could happen given that `ping()` refuses to send a ping msg to ourselves. Maybe leave a comment here saying that and that we should investigate this further?

PR #790, opened 2018-05-26
Raise HandshakeFailure instead of ValueError when Ack is wrong size

### LINKED ISSUE #776: Trinity `Auth ack msg too short`
### What is wrong?

```
   ERROR  05-24 15:52:55        peer  Unexpected error during auth/p2p handshake with <Node(0x4382@104.149.231.208)>
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 739, in connect
    self.cancel_token)
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 136, in handshake
    ) = await auth.handshake(remote, privkey, token)
  File "/Users/piper/sites/py-evm/p2p/auth.py", line 55, in handshake
    initiator, reader, writer, token)
  File "/Users/piper/sites/py-evm/p2p/auth.py", line 77, in _handshake
    ephemeral_pubkey, responder_nonce = initiator.decode_auth_ack_message(auth_ack)
  File "/Users/piper/sites/py-evm/p2p/auth.py", line 179, in decode_auth_ack_message
    raise ValueError("Auth ack msg too short: {}".format(len(ciphertext)))
ValueError: Auth ack msg too short: 0
```

This error occurs semi regularly during fast sync (and in general during running the trinity node.)

### How can it be fixed

Explicit error handling.

pipermerriam:
Fixes #776 

### What was wrong?

`ValueError` was escaping during `Peer.connect`

### How was it fixed?

Changed `ValueError` exceptions to be `HandshakeFailure` exceptions which are caught and handled in `Peer.connect`.

#### Cute Animal Picture

d4b7240f2b7ac8965221eb4c2f2ec59a

### DIFF
--- a/p2p/auth.py
+++ b/p2p/auth.py
@@ -23,7 +23,7 @@ from p2p.cancel_token import CancelToken, wait_with_token
 from p2p.constants import REPLY_TIMEOUT
 from p2p.exceptions import (
     DecryptionError,
-    HandshakeFailure,
+    BadAckMessage,
 )
 from p2p.utils import (
     sxor,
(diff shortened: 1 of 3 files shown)


gsalgado:
I think we should find out if those peers are actually replying with an empty/incomplete auth ack, or if it is a bug in our code (e.g. we're not reading the whole reply, or maybe it's an EIP8 ack), but if we use a `HandshakeFailure` here, these will now be relegated to the debug log together with the *expected* handshake errors (e.g. different network/genesis) and we'll forget about them. Given that, I'd prefer if we keep using a separate exception here and catch it in `peer.handshake()`, with a TODO comment saying we need to investigate

pipermerriam:
@gsalgado take a look at latest changes.  How does that look?

I ended up putting the handling in `connect` because `p2p.peer.handshake` only has a happy-path or exception raising where-as `Peer.connect` has a failure path by returning `None`.

PR #794, opened 2018-05-28
p2p: Explicitly use a ProcessPoolExecutor where needed

### LINKED ISSUE #778: Make sure we use a ProcessPoolExecutor for cpu-intensive tasks
### What is wrong?

@carver [run into some issues when using a ProcessPoolExecutor](https://github.com/ethereum/py-evm/pull/754) as the default executor, and apparently [GvR thinks that's not a good idea anyway](https://groups.google.com/d/msg/python-tulip/91NCCqV4SFs/9McZnfea_VcJ) so we're back to using a ThreadPoolExecutor as the default

### How can it be fixed

Places where we use loop.run_in_executor to offload cpu-intensive tasks must explicitly use a ProcessPoolExecutor

gsalgado:
Without this the fast sync is insanely slow

Closes #778

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -1,7 +1,9 @@
 import asyncio
+from concurrent.futures import ProcessPoolExecutor
 import logging
 import math
 import operator
+import os
 import time
 from typing import (
     Any,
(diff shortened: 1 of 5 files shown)


gsalgado:
Failing because of https://github.com/ethereum/eth-hash/issues/18

gsalgado:
I think it's important that we include this in the alpha, so maybe for now we can just pin eth-hash to `<=1.0.1`

pipermerriam:
I think I have `eth-hash` fixed, lets see what tests say.

Issue #689, opened 2018-05-14
Consider using p2p.BaseService in trinity.rpc.ipc.IPCServer


carver:
### What is wrong?

There is (probably) a lot of duplicated logic across services, which is why `p2p.BaseService` was written.

### How can it be fixed

Consider whether anything in `trinity.rpc.ipc` would benefit from using `BaseService`, especially `IPCServer`.

carver:
Also, remove the ipc file on server shutdown.

PR #798, opened 2018-05-28
Ensure trinity.nodes is a proper python module

cburgdorf:
### What was wrong?

When trying to run Trinity on a fresh Ubuntu I ran into the following error.

```
    INFO  05-28 12:52:55        main  
  ______     _       _ __       
                       /____/   
    INFO  05-28 12:52:55        main  Trinity/0.2.0a18/linux/cpython3.6.5
    INFO  05-28 12:52:55        main  enode://781245c14c5885cf79df99e233733ec7a8fcdf8a9e3bfeef50aedd43b3e42d03@[:]:30303
    INFO  05-28 12:52:55        main  network: 1
Process SpawnProcess-2:
  File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/py-evm/venv/lib/python3.6/site-packages/trinity/utils/logging.py", line 63, in inner
    return fn(*args, **inner_kwargs)
  File "/py-evm/venv/lib/python3.6/site-packages/trinity/main.py", line 156, in launch_node
    NodeClass = chain_config.node_class
  File "/py-evm/venv/lib/python3.6/site-packages/trinity/config.py", line 173, in node_class
    from trinity.nodes.mainnet import (
ModuleNotFoundError: No module named 'trinity.nodes'
^CError in atexit._run_exitfuncs:
  File "/usr/lib/python3.6/multiprocessing/popen_fork.py", line 28, in poll
    pid, sts = os.waitpid(self.pid, flag)
```

To highlight the important part

```
ModuleNotFoundError: No module named 'trinity.nodes'
```

### How was it fixed?

Added `__init__.py` to ensure `trinity.nodes` is an actual python module. 

However, after I made that a proper Python module, it started to get type checked and there were a bunch of errors. Most of them boiled down to APIs expecting are more concrete type than what was passed into (e.g. an API expecting a `SportsCar` isn't satisfied when getting a `Car` passed, as a `Car` might be missing `go_real_fast()` API)

There were also cases where we were expecting to deal with a `Path` but where passed an `str`. I first tried to get rid of the `str` return types in the `ChainConfig` altogether but that turned out to be problematic so for now I just fixed that by wrapping the `str` as `Path`.

#### Cute Animal Picture

### DIFF
--- a/trinity/nodes/base.py
+++ b/trinity/nodes/base.py
@@ -5,10 +5,13 @@ from multiprocessing.managers import (
     BaseManager,
 )
 from threading import Thread
-from typing import Type
+from typing import (
+    Type,
+    Union
+)
 
 from evm.chains.base import BaseChain
-from evm.db.header import BaseHeaderDB
+
 from p2p.service import (
     BaseService,
     EmptyService,
@@ -22,7 +25,10 @@ from trinity.chains.header import (
 )
 from trinity.db.chain import ChainDBProxy
 from trinity.db.base import DBProxy
-from trinity.db.header import AsyncHeaderDBProxy
+from trinity.db.header import (
+    BaseAsyncHeaderDB,
+    AsyncHeaderDBProxy
+)
 from trinity.rpc.main import (
     RPCServer,
 )
@@ -68,7 +74,7 @@ class Node(BaseService):
         return self._db_manager
 
     @property
-    def headerdb(self) -> BaseHeaderDB:
+    def headerdb(self) -> BaseAsyncHeaderDB:
         return self._headerdb
 
     def add_service(self, service: BaseService) -> None:
@@ -77,7 +83,7 @@ class Node(BaseService):
         else:
             self._auxiliary_services.append(service)
 
-    def make_ipc_server(self) -> IPCServer:
+    def make_ipc_server(self) -> Union[IPCServer, BaseService]:
         if self._jsonrpc_ipc_path:
             rpc = RPCServer(self.get_chain())
             return IPCServer(rpc, self._jsonrpc_ipc_path)
--- a/trinity/nodes/full.py
+++ b/trinity/nodes/full.py
@@ -35,10 +35,10 @@ class FullNode(Node):
             self._p2p_server = Server(
                 self._node_key,
                 self._node_port,
-                manager.get_chain(),
-                manager.get_chaindb(),
+                manager.get_chain(),  # type: ignore
+                manager.get_chaindb(),  # type: ignore
                 self.headerdb,
-                manager.get_db(),
+                manager.get_db(),  # type: ignore
                 self._network_id,
                 peer_pool_class=PreferredNodePeerPool,
                 bootstrap_nodes=self._bootstrap_nodes,
--- a/trinity/nodes/light.py
+++ b/trinity/nodes/light.py
@@ -9,9 +9,6 @@ from p2p.peer import (
     PeerPool,
     PreferredNodePeerPool,
 )
-from p2p.service import (
-    BaseService,
-)
 
 from trinity.chains.light import (
     LightDispatchChain,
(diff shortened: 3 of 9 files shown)


trinity/nodes/base.py
```diff
@@ -77,7 +83,7 @@ def add_service(self, service: BaseService) -> None:
         else:
             self._auxiliary_services.append(service)
 
-    def make_ipc_server(self) -> IPCServer:
+    def make_ipc_server(self) -> Union[IPCServer, BaseService]:
```
cburgdorf: These two don't share a common base class at all, so `Union` is the only way to go if that API wants to either return `IPCServer` or `BaseService`. This is related to https://github.com/ethereum/py-evm/issues/689

trinity/nodes/full.py
```diff
@@ -35,10 +35,10 @@ def get_p2p_server(self) -> BaseService:
             self._p2p_server = Server(
                 self._node_key,
                 self._node_port,
-                manager.get_chain(),
-                manager.get_chaindb(),
+                manager.get_chain(),  # type: ignore
```
cburgdorf: Doesn't seem like we have a proper class for that. Was unsure how to resolve that and ignored for now.

trinity/nodes/base.py
```diff
@@ -44,11 +50,11 @@ class Node(BaseService):
     def __init__(self, chain_config: ChainConfig) -> None:
         super().__init__()
 
-        self._db_manager = create_db_manager(chain_config.database_ipc_path)
+        self._db_manager = create_db_manager(Path(chain_config.database_ipc_path))
```
carver: Maybe it's better for `chain_config` to return a `Path`.

trinity/nodes/base.py
```diff
@@ -44,11 +50,11 @@ class Node(BaseService):
...
         self._db_manager.connect()  # type: ignore
         self._headerdb = self._db_manager.get_headerdb()  # type: ignore
 
-        self._jsonrpc_ipc_path: Path = chain_config.jsonrpc_ipc_path
+        self._jsonrpc_ipc_path: Path = Path(chain_config.jsonrpc_ipc_path)
```
carver: same here

cburgdorf:
Yep, was my first take but then levelb (and I think other places, too) were
expecting a str and I rolled back.

trinity/nodes/base.py
cburgdorf: That was my first take but I abandoned it after looking through these errors https://github.com/ethereum/py-evm/pull/800

Can work to resolve them though.

trinity/nodes/light.py
```diff
@@ -60,7 +57,7 @@ def get_chain(self) -> LightDispatchChain:
 
         return self._chain
 
-    def get_p2p_server(self) -> BaseService:
+    def get_p2p_server(self) -> LightPeerChain:
```
carver: What is the rationale for specifying `LightPeerChain` here? I think it's only important that this method returns a `BaseService` (like the superclass's abstract method). The implementation choice shouldn't matter.

cburgdorf: The return value of this method ends up as `peer_chain` in [`LightDispatchChain`](https://github.com/cburgdorf/py-evm/blob/222e53d79da62e5f48214cb6d56178ca61ce350e/trinity/chains/light.py#L63) which is expecting a `LightPeerChain` not a `BaseService`. In other words, we can not pass a base class somewhere where a more concrete class is expected. (e.g. an API expecting `SportsCar` isn't satisfied with `Car` because `Car` may lack `go_real_fast()` API.

So, if you make this `BaseService` mypy will rightfully yell at you.

```
trinity/nodes/light.py:60: error: Argument "peer_chain" to "LightDispatchChain" has incompatible type "BaseService"; expected "LightPeerChain"
```

cburgdorf:
@carver if https://github.com/ethereum/py-evm/pull/800 goes green, we may wanna swap as it has the proper "ensure everything is `Path`" solution

cburgdorf:
@carver @pipermerriam This now includes the proper `Path` handling

PR #800, opened 2018-05-28
Christoph/fix/trinity nodes alternative

cburgdorf:
Temporary alternative to #798


cburgdorf:
Closing as this is now incorporated in #798

PR #804, opened 2018-05-29
Fix HeaderNotFound error

pipermerriam:
#### Cute Animal Picture


pipermerriam:
Pretty sure this is not the correct fix, trying other things.

gsalgado:
#805 has a fix, with a modified test that exercises the fix as well

Issue #771, opened 2018-05-24
Need a way to limit number of connected peers


gsalgado:
### What is wrong?

We accept incoming peer connections but there's no way to limit the number of simultaneous connections.

### How can it be fixed

Add a CLI flag to specify max number of peers, and have `p2p.Server` reject new connections once we reach that

glaksmono:
Would like to take on this 🔪

pipermerriam:
Fixed with #824

PR #824, opened 2018-05-30
[WIP] Add CLI argument to Trinity to limit number of connected peers

glaksmono:
### What was wrong?

We accept incoming peer connections but there's no way to limit the number of simultaneous connections.

Issue Reference: https://github.com/ethereum/py-evm/issues/771

### How was it fixed?

Add a CLI flag to specify max number of peers, and have `p2p.Server` reject new connections once we reach that.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -103,6 +103,7 @@ from p2p.p2p_proto import (
 from .constants import (
     CONN_IDLE_TIMEOUT,
     DEFAULT_MIN_PEERS,
+    DEFAULT_MAX_PEERS,
     HEADER_LEN,
     MAC_LEN,
     REPLY_TIMEOUT,
@@ -670,6 +671,7 @@ class PeerPool(BaseService):
                  privkey: datatypes.PrivateKey,
                  discovery: DiscoveryProtocol,
                  min_peers: int = DEFAULT_MIN_PEERS,
+                 max_peers: int = DEFAULT_MAX_PEERS
                  ) -> None:
         super().__init__()
         self.peer_class = peer_class
@@ -678,6 +680,7 @@ class PeerPool(BaseService):
         self.privkey = privkey
         self.discovery = discovery
         self.min_peers = min_peers
+        self.max_peers = max_peers
         self.connected_nodes: Dict[Node, BasePeer] = {}
         self._subscribers: List[PeerPoolSubscriber] = []
 
@@ -693,6 +696,7 @@ class PeerPool(BaseService):
             raise Exception(
                 "We cannot handle multiple subscribers yet; "
                 "see https://github.com/ethereum/py-evm/issues/768 for details")
+
         self._subscribers.append(subscriber)
         for peer in self.connected_nodes.values():
             subscriber.register_peer(peer)
@@ -749,6 +753,15 @@ class PeerPool(BaseService):
             peer = await handshake(
                 remote, self.privkey, self.peer_class, self.headerdb, self.network_id,
                 self.cancel_token)
+
+            """
+            There is max_peers defined from CLI.
+            We drop the subscriber if it's already maxed out.
+            """
+            if len(self.connected_nodes) >= self.max_peers:
+                peer.disconnect(DisconnectReason.too_many_peers)
+                return None
+
             return peer
         except OperationCancelled:
             # Pass it on to instruct our main loop to stop.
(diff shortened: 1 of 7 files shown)


glaksmono:
@gsalgado thought about the code structure? I'm also still looking if https://github.com/glaksmono/py-evm/blob/727a6766db47eeda9cf647300e6bc23f3379ab75/p2p/peer.py#L363 is the right place to make decision for the peer `disconnect` if this peer is the `max_peer + 1`th

trinity/utils/chains.py
```diff
@@ -137,6 +137,9 @@ def construct_chain_config_params(args):
     if args.sync_mode is not None:
         yield 'sync_mode', args.sync_mode
 
+    if args.max_peers is not None:
+        yield 'max_peers', int(args.max_peers)
```
pipermerriam: This `int` cast shouldn't be necessary as it should be done as part of the argument parsing.

p2p/server.py
```diff
@@ -75,6 +75,7 @@ def __init__(self,
                  headerdb: 'BaseAsyncHeaderDB',
                  base_db: BaseDB,
                  network_id: int,
+                 max_peers: int = None,
```
pipermerriam: I think we can create a `DEFAULT_MAX_PEERS` for this set at `25`

p2p/peer.py
```diff
@@ -359,6 +359,11 @@ def process_p2p_handshake(self, cmd: protocol.Command, msg: protocol._DecodedMsg
             self.disconnect(DisconnectReason.other)
             raise HandshakeFailure("Expected a Hello msg, got {}, disconnecting".format(cmd))
         remote_capabilities = msg['capabilities']
+
+        # if current_peers
+        #     self.disconnect(DisconnectReason.too_many_peers)
```
pipermerriam: For right now, I'm ok with making this a TODO and opening up an issue to enforce the max at a later date.

pipermerriam:
I think you also need to update the logic in `PeerPool.maybe_connect_to_more_peers` to use the max peers number.

pipermerriam:
Oh, and lets validate that the provided `max_peers` argument is `>=` whatever `min_peers` is set to.

glaksmono:
@pipermerriam @gsalgado updated the logic to raise `HandshakeFailure` at the `PeerPool` level. I thought it's kind of weird if we do the `HandshakeFailure` at the `peer` level because I presumed `peer` is the subset of `PeerPool`, thus it probably wouldn't know what are the number of peers connected at any given time.

Thoughts about this?

trinity/utils/chains.py
glaksmono: If I don't cast this, then I will be getting the following:

```
{'max_peers': '10',
 'network_id': 1,
 'port': 30303,
 'preferred_nodes': None,
 'sync_mode': 'full'}

  File "/Users/glaksmono/Documents/py-evm/p2p/server.py", line 101, in __init__
    if self.max_peers < self.min_peers:
TypeError: '<' not supported between instances of 'str' and 'int'
```

p2p/peer.py
```diff
@@ -693,6 +696,16 @@ def subscribe(self, subscriber: PeerPoolSubscriber) -> None:
...
                 "see https://github.com/ethereum/py-evm/issues/768 for details")
+
+        """
+        There is max_peers defined from CLI.
+        We drop the subscriber if it's already maxed out.
+        """
+        if len(self.connected_nodes) >= self.max_peers:
+            raise HandshakeFailure(
+                "Not processing the handshake because too many peers. "
+                "Max peers %d", self.max_peers)
```
gsalgado: This is not the correct place to raise `HandshakeFailure` as there's no handshake going on here; this is the `PeerPool.subscribe()` method. Or am I missing something?

p2p/peer.py
```diff
@@ -787,11 +800,12 @@ def add_peer(self, peer):
                 self._discovery_last_lookup = time.time()
 
     async def maybe_connect_to_more_peers(self) -> None:
-        """Connect to more peers if we're not yet connected to at least self.min_peers."""
-        if len(self.connected_nodes) >= self.min_peers:
```
gsalgado: We should get rid of `min_peers` now as we don't use it anymore

pipermerriam:
I went ahead and merged, we can drop `min_peers` independently.

PR #829, opened 2018-05-30
Handle missing $HOME and report better info to user

### LINKED ISSUE #784: Trinity fails if `$HOME` environment variable is not set.
### What is wrong?

```bash
  File "/home/ubuntu/py-evm/venv/bin/trinity", line 11, in <module>
    load_entry_point('py-evm', 'console_scripts', 'trinity')()
  File "/home/ubuntu/py-evm/trinity/main.py", line 101, in main
    initialize_data_dir(chain_config)
  File "/home/ubuntu/py-evm/trinity/chains/__init__.py", line 88, in initialize_data_dir
    chain_config.data_dir,
ValueError: The base chain directory provided does not exist: `//$HOME/.local/share/trinity/mainnet`
```

If the `$HOME` environment variable isn't set Trinity crashes.

### How can it be fixed

Gracefully handle this not being set and provide a helpful error message informing the user that they can:

- set `$HOME` to the desired path.
- set `$XDG_TRINITY_HOME` to the desired path.
- set `$TRINITY_DATA_DIR` to the desired path.
- run trinity with the `--data-dir /some/path` where `/some/path` is the desired path.

All of these should be tested to work as solutions.

cburgdorf:
### What was wrong?

This fixes #784 where Trinity would not boot correctly if the `$HOME` variable isn't set.

### How was it fixed?

Work through all fallbacks (e.g. `XDG_TRINITY_ROOT`, `TRINITY_DATA_DIR`) until we hit `$HOME` and if `$HOME` is also missing give a clear explanation to the user.

If the user decides to provide the `TRINITY_DATA_DIR` or passes `--data-dir` but that is outside of the `XDG_TRINITY_ROOT` all paths must be pre-initialized as trinity won't attempt to create any directories outside of the `XDG_TRINITY_ROOT`.

#### Cute Animal Picture

### DIFF
--- a/trinity/chains/__init__.py
+++ b/trinity/chains/__init__.py
@@ -85,11 +85,9 @@ def is_database_initialized(chaindb: AsyncChainDB) -> bool:
 
 
 def initialize_data_dir(chain_config: ChainConfig) -> None:
-    if (not os.path.exists(chain_config.data_dir) and
-            is_under_xdg_trinity_root(chain_config.data_dir)):
-
+    if not chain_config.data_dir.exists() and is_under_xdg_trinity_root(chain_config.data_dir):
         chain_config.data_dir.mkdir(parents=True, exist_ok=True)
-    elif not os.path.exists(chain_config.data_dir):
+    elif not chain_config.data_dir.exists():
         # we don't lazily create the base dir for non-default base directories.
         raise AmbigiousFileSystem(
             "The base chain directory provided does not exist: `{0}`".format(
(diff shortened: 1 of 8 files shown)


trinity/chains/__init__.py
```diff
@@ -82,30 +85,32 @@ def is_database_initialized(chaindb: AsyncChainDB) -> bool:
...
-        raise ValueError(
+        raise AmbigiousFileSystem(
             "The base chain directory provided does not exist: `{0}`".format(
                 chain_config.data_dir,
             )
         )
 
     # Logfile
-    if is_under_xdg_trinity_root(chain_config.logfile_path):
+    if (not os.path.exists(chain_config.logfile_path) and
```
pipermerriam: the `Path.exists()` method works here too, we can drop some of the `os.path` calls now that we're using `Path` objects most places.

trinity/main.py
```diff
@@ -70,15 +87,29 @@ def main() -> None:
             "networks are supported.".format(args.network_id)
         )
 
-    chain_config = ChainConfig.from_parser_args(args)
+    logger, formatter, handler_stream = setup_trinity_stdout_logging(log_level)
+
+    try:
+        chain_config = ChainConfig.from_parser_args(args)
+    except AmbigiousFileSystem:
+            exit_because_ambigious_filesystem(logger)
```
pipermerriam: Indentation looks off here.

cburgdorf:
@pipermerriam mind giving that another look.

The tests are all condensed into one parametrized one and I added one more test case for each method. I also added another exception class to give a more specialized message to the user when a data directory is set that is outside of the root which is ok but not if the user did not pre-init the paths themselves.

PR #832, opened 2018-05-30
Piper/handle header not found during difficulty calculation

pipermerriam:
### What was wrong?

```bash
   ERROR  05-30 17:56:58  base_events  Task exception was never retrieved
future: <Task finished coro=<FastChainSyncer.sync() done, defined at /home/ubuntu/py-evm/venv/lib/python3.6/site-packages/p2p/chain.py:161> exception=HeaderNotFound('No header with hash 0xe3b100bfd7ab6e5c28bda5e469fb7c3c9082245efda142d26a12220e8e9cb153 found',)>
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/p2p/chain.py", line 174, in sync
    await self._sync(peer)
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/p2p/chain.py", line 220, in _sync
    head_number = await self._process_headers(peer, headers)
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/p2p/chain.py", line 244, in _process_headers
    target_td = await self._calculate_td(headers)
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/p2p/chain.py", line 237, in _calculate_td
    td = await self.chaindb.coro_get_score(headers[0].parent_hash)
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/trinity/utils/mp.py", line 23, in method
    args,
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.6/multiprocessing/managers.py", line 772, in _callmethod
    raise convert_to_error(kind, result)
evm.exceptions.HeaderNotFound: No header with hash 0xe3b100bfd7ab6e5c28bda5e469fb7c3c9082245efda142d26a12220e8e9cb153 found
```

A `HeaderNotFound` exception is not being handled.

### How was it fixed?

When the `HeaderNotFound` occurs while we're trying to calculate the header difficult, we now catch the exception in the main `_sync` loop.  When this occurs, we step backwards to try and find a common ancestor in our chains.  If we reach the genesis block

#### Cute Animal Picture

4358238267_2ffe7a1bff_o

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -218,10 +218,17 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             # TODO: Process headers for consistency.
             try:
                 head_number = await self._process_headers(peer, headers)
+            except HeaderNotFound:
+                if start_at == 0:
+                    self.logger.warn("No common ancestry found from %s, aborting sync", peer)
+                    break
+                start_at = max(0, start_at - eth.MAX_HEADERS_FETCH)
+                continue
             except NoEligiblePeers:
                 self.logger.info("No peers have the blocks we want, aborting sync")
                 break
-            start_at = head_number + 1
+            else:
+                start_at = head_number + 1
 
     async def _calculate_td(self, headers: List[BlockHeader]) -> int:
         """Return the score (total difficulty) of the last header in the given list.
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -223,6 +223,13 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
                     self.logger.warn("No common ancestry found from %s, aborting sync", peer)
                     break
                 start_at = max(0, start_at - eth.MAX_HEADERS_FETCH)
+                if start_at < head.block_number - 10000:
+                    self.logger.warn(
+                        "No common ancestor found with %s within 10,000 blocks, "
+                        "aborting sync",
+                        peer,
+                    )
+                    break
                 continue
             except NoEligiblePeers:
                 self.logger.info("No peers have the blocks we want, aborting sync")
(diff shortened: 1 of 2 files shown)


p2p/chain.py
```diff
@@ -218,10 +218,29 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            except HeaderNotFound:
+                # This is comes from when we attempt to compute the difficulty
+                # of a header whos parent is not in our database.
+                if start_at == 0:
+                    self.logger.warn("No common ancestry found from %s, aborting sync", peer)
+                    break
+                start_at = max(0, start_at - eth.MAX_HEADERS_FETCH)
+                if start_at < head.block_number - 10000:
+                    self.logger.warn(
+                        "No common ancestor found with %s within 10,000 blocks, "
```
cburgdorf: Nitpick: number should better come from variable (at least, if not even outside constant)

p2p/chain.py
```diff
@@ -218,10 +218,29 @@ def register_peer(self, peer: BasePeer) -> None:
             # TODO: Process headers for consistency.
             try:
                 head_number = await self._process_headers(peer, headers)
+            except HeaderNotFound:
+                # This is comes from when we attempt to compute the difficulty
+                # of a header whos parent is not in our database.
```
gsalgado: I'd still like to understand what exactly is going on in #837 before merging a fix, but if you need this fix badly my only suggestion would be to instead doing an explicit check and rewind before calling `_process_headers()`, so we don't have to catch `HeaderNotFound` here, as that might actually hide other bugs in our code. 

Something like
```
    if not await chaindb.coro_header_exists(headers[0].parent_hash):
        start_at = max(0, start_at - eth.MAX_HEADERS_FETCH)
        ...
        continue
    try:
        self._process_headers(...)
        ...
```

pipermerriam: This is not needed badly and I'm not convinced it is the right solution.  Going to close and wait for this issue to re-surface.

PR #834, opened 2018-05-30
Add web3 clientVersion endpoint

davesque:
### What was wrong?

Lacked an endpoint for `web3.clientVersion` call.

### How was it fixed?

Added it.

#### Cute Animal Picture


Nathapon8899:
#835

Issue #775, opened 2018-05-24
Trinity `ListDeserializationError` during fast sync


pipermerriam:
### What is wrong?

```bash
concurrent.futures.process._RemoteTraceback:
"""
Traceback (most recent call last):
  File "/Users/piper/.pyenv/versions/3.6.3/lib/python3.6/concurrent/futures/process.py", line 175, in _process_worker
    r = call_item.fn(*call_item.args, **call_item.kwargs)
  File "/Users/piper/sites/py-evm/p2p/p2p_proto.py", line 56, in decode
    raw_decoded = cast(Dict[str, int], super(Disconnect, self).decode(data))
  File "/Users/piper/sites/py-evm/p2p/protocol.py", line 85, in decode
    return self.decode_payload(data[1:])
  File "/Users/piper/sites/py-evm/p2p/protocol.py", line 72, in decode_payload
    data = rlp.decode(rlp_data, sedes=decoder)
  File "/Users/piper/python-environments/trinity/lib/python3.6/site-packages/rlp/codec.py", line 220, in decode
    obj = sedes.deserialize(item, **kwargs)
  File "/Users/piper/python-environments/trinity/lib/python3.6/site-packages/eth_utils/functional.py", line 22, in inner
    return callback(fn(*args, **kwargs))
  File "/Users/piper/python-environments/trinity/lib/python3.6/site-packages/rlp/sedes/lists.py", line 84, in deserialize
    raise ListDeserializationError('List has wrong length', serial)
rlp.exceptions.ListDeserializationError: List has wrong length
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 739, in connect
    self.cancel_token)
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 143, in handshake
    await peer.do_p2p_handshake()
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 222, in do_p2p_handshake
    cmd, msg = await self.read_msg()
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 324, in read_msg
    token=self.cancel_token)
  File "/Users/piper/sites/py-evm/p2p/cancel_token.py", line 110, in wait_with_token
    return done.pop().result()
rlp.exceptions.ListDeserializationError: List has wrong length
```

This error occurs semi-regularly during fast sync.

### How can it be fixed

Explicit error handling.

gsalgado:
[...]

PR #839, opened 2018-05-30
Handle ListDeserializationError during connection

pipermerriam:
### What was wrong?

#775 

### How was it fixed?

Catch the `rlp` error within the `Disconnect` command class and log a warning, then raise a new exception `MalformedMessage` which is caught within `Peer.connect` as one of the *expected* exceptions.

#### Cute Animal Picture

cute-baby-donkey

### DIFF
--- a/p2p/protocol.py
+++ b/p2p/protocol.py
@@ -41,6 +41,10 @@ class Command:
         self.cmd_id = cmd_id_offset + self._cmd_id
 
     @property
+    def logger(self):
+        return logging.getLogger("p2p.protocol.{0}".format(self.__class__.__name__))
+
+    @property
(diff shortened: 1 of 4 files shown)


p2p/protocol.py
```diff
@@ -40,6 +40,10 @@ def __init__(self, cmd_id_offset: int) -> None:
         self.cmd_id_offset = cmd_id_offset
         self.cmd_id = cmd_id_offset + self._cmd_id
 
+    @property
+    def logger(self):
+        return logging.getLogger("p2p.protocol.{0}".format(self.__class__.__name__))
```
carver: Maybe worth saving this logger to the instance on the first usage.

Issue #711, opened 2018-05-16
LightChain integration test is broken


gsalgado:
Some LightChain methods (e.g. `get_canonical_block_by_number()`) have been removed recently and now `tests/p2p/test_lightchain_integration.py` is broken.

Not sure if there's a plan to add them back or if the functionality is being moved somewhere else

carver:
Did someone ignore a CI failure, or is that not being tested in circle?

carver:
Found it: https://github.com/ethereum/py-evm/pull/685/files#diff-354708b81c89737c927333afd3f068fdL257

Unless someone else wants to jump in, I'll take a look at reviving these methods tomorrow morning. I'm thinking that some of them might go into another class that orchestrates the retrieval and local storage of block/transaction data.

carver:
These all got added back in. Closing...

PR #841, opened 2018-05-31
p2p: Move on-demand retrieval methods from LESPeer to LightPeerChain

gsalgado:
This is in preparation for another PR that will redesign the handling of peers in PeerSubscriber instances

Closes: #711

### DIFF
--- a/p2p/lightchain.py
+++ b/p2p/lightchain.py
@@ -1,6 +1,8 @@
 import asyncio
 import logging
 from typing import (
+    Any,
+    Callable,
     cast,
     Dict,
     List,
@@ -11,22 +13,29 @@ from typing import (
 
 from async_lru import alru_cache
 
+import rlp
+
 from eth_typing import (
     Address,
     Hash32,
 )
 
+from eth_hash.auto import keccak
+
 from eth_utils import (
     encode_hex,
 )
 
+from trie import HexaryTrie
+
 from evm.constants import GENESIS_BLOCK_NUMBER
+from evm.exceptions import BlockNotFound, HeaderNotFound
 from evm.rlp.accounts import Account
-from evm.rlp.blocks import BaseBlock
 from evm.rlp.headers import BlockHeader
 from evm.rlp.receipts import Receipt
 
 from p2p.exceptions import (
+    BadLESResponse,
     EmptyGetBlockHeadersReply,
     LESAnnouncementProcessingError,
     OperationCancelled,
@@ -34,20 +43,23 @@ from p2p.exceptions import (
     UnexpectedMessage,
 )
 from p2p import les
+from p2p import protocol
 from p2p.cancel_token import (
     CancelToken,
     wait_with_token,
 )
+from p2p.constants import REPLY_TIMEOUT
 from p2p.peer import (
     BasePeer,
     LESPeer,
     PeerPool,
     PeerPoolSubscriber,
 )
+from p2p.rlp import BlockBody
 from p2p.service import (
     BaseService,
 )
-from p2p.utils import unclean_close_exceptions
+from p2p.utils import gen_request_id, unclean_close_exceptions
 
 if TYPE_CHECKING:
     from trinity.db.header import BaseAsyncHeaderDB  # noqa: F401
@@ -56,6 +68,7 @@ if TYPE_CHECKING:
 class LightPeerChain(PeerPoolSubscriber, BaseService):
     logger = logging.getLogger("p2p.lightchain.LightPeerChain")
     max_consecutive_timeouts = 5
+    reply_timeout = REPLY_TIMEOUT
     headerdb: 'BaseAsyncHeaderDB' = None
 
     def __init__(self, headerdb: 'BaseAsyncHeaderDB', peer_pool: PeerPool) -> None:
@@ -65,6 +78,7 @@ class LightPeerChain(PeerPoolSubscriber, BaseService):
         self._announcement_queue: asyncio.Queue[Tuple[LESPeer, les.HeadInfo]] = asyncio.Queue()
         self._last_processed_announcements: Dict[LESPeer, les.HeadInfo] = {}
         self._running_peers: Set[LESPeer] = set()
+        self._pending_replies: Dict[int, Callable[[protocol._DecodedMsgType], None]] = {}
 
     def register_peer(self, peer: BasePeer) -> None:
         asyncio.ensure_future(self.handle_peer(cast(LESPeer, peer)))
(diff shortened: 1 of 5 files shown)


p2p/lightchain.py
```diff
@@ -95,12 +108,17 @@ def register_peer(self, peer: BasePeer) -> None:
...
-            # We currently implement only the LES commands for retrieving data (apart from
-            # Announce), and those should always come as a response to requests we make so will be
-            # handled by LESPeer.handle_sub_proto_msg().
+
             if isinstance(cmd, les.Announce):
                 peer.head_info = cmd.as_head_info(msg)
                 self._announcement_queue.put_nowait((peer, peer.head_info))
+            elif isinstance(msg, dict):
+                request_id = msg.get('request_id')
+                if request_id is not None and request_id in self._pending_replies:
```
pipermerriam: Can we document the case where `request_id == None` so it's clear why we silently do nothing in that case

p2p/lightchain.py
```diff
@@ -245,35 +263,105 @@ def peer_block_height(peer: LESPeer) -> int:
...
-        return await peer.get_block_header_by_hash(block_hash, self.cancel_token)
+        return await self._get_block_header_by_hash(peer, block_hash)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
     async def get_block_body_by_hash(self, block_hash: Hash32) -> BaseBlock:
         peer = await self.get_best_peer()
         self.logger.debug("Fetching block %s from %s", encode_hex(block_hash), peer)
-        return await peer.get_block_by_hash(block_hash, self.cancel_token)
+        request_id = gen_request_id()
+        peer.sub_proto.send_get_block_bodies([block_hash], request_id)
```
pipermerriam: This pattern of generating the `request_id` independent of issuing the request is odd to me.  Is there a reason why `sub_proto.send_get_block_bodies` doesn't generate the `request_id` and return it?

p2p/lightchain.py
```diff
@@ -245,35 +263,105 @@ def peer_block_height(peer: LESPeer) -> int:
...
-        self.logger.debug("Fetching header %s from %s", encode_hex(block_hash), peer)
-        return await peer.get_block_header_by_hash(block_hash, self.cancel_token)
+        return await self._get_block_header_by_hash(peer, block_hash)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
     async def get_block_body_by_hash(self, block_hash: Hash32) -> BaseBlock:
         peer = await self.get_best_peer()
         self.logger.debug("Fetching block %s from %s", encode_hex(block_hash), peer)
-        return await peer.get_block_by_hash(block_hash, self.cancel_token)
+        request_id = gen_request_id()
```
pipermerriam: Also, I was surprised to see how `gen_request_id` is implemented (I sort of expected it to return a `uuid.uuid4()` or something.  Is there a reason for it's current implementation?

p2p/lightchain.py
```diff
@@ -245,35 +263,105 @@ def peer_block_height(peer: LESPeer) -> int:
...
+        return rlp.decode(rlp_account, sedes=Account)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
     async def get_contract_code(self, block_hash: Hash32, key: bytes) -> bytes:
         peer = await self.get_best_peer()
-        return await peer.get_contract_code(block_hash, key, self.cancel_token)
+        request_id = gen_request_id()
+        peer.sub_proto.send_get_contract_code(block_hash, key, request_id)
+        reply = await self._wait_for_reply(request_id)
+        if not reply['codes']:
```
pipermerriam: It seems like we should be looking up the `account.code_hash` and verifying that the returned code is indeed the code which hashes to the expected `code_hash`?

p2p/lightchain.py
```diff
@@ -245,35 +263,105 @@ def peer_block_height(peer: LESPeer) -> int:
...
+
+    async def _get_block_header_by_hash(self, peer: LESPeer, block_hash: Hash32) -> BlockHeader:
+        self.logger.debug("Fetching header %s from %s", encode_hex(block_hash), peer)
+        request_id = gen_request_id()
+        max_headers = 1
+        peer.sub_proto.send_get_block_headers(block_hash, max_headers, request_id)
+        reply = await self._wait_for_reply(request_id)
+        if not reply['headers']:
+            raise HeaderNotFound("Peer {} has no block with hash {}".format(peer, block_hash))
+        return reply['headers'][0]
```
pipermerriam: Same here, I expected to see a check that the returned `header.hash` actually matches the requested `block_hash`

p2p/lightchain.py
gsalgado: Sure, it's because not all LES messages include a request_id. I've added a comment explaining

p2p/lightchain.py
gsalgado: The only drawback I can think of is that these sub_proto methods would then no longer conform to the standard `send_*` method signature, which doesn't return anything. Filed #843 for it

p2p/lightchain.py
gsalgado: Nope. I've changed it to return `big_endian_to_int(uuid4().bytes)`

gsalgado: Actually, `uuid4().int`

p2p/lightchain.py
gsalgado: Well, this is the light client, so we can't assume we'll have the account in our db, right?

p2p/lightchain.py
gsalgado: I've added a check raising an exception if the received header's hash does not match `block_hash`

p2p/lightchain.py
gsalgado: And it turns out a uuid4 is too long (16 bytes) to use in the request_id field. I'd need to use only 8 bytes of it, so I guess I'll stick with os.urandom() for now, unless you have another suggestion?

pipermerriam: No other suggestions 👍

p2p/lightchain.py
pipermerriam: Right right right, context... 👍

pipermerriam: Um, but wait, we could still fetch the account which IIRC comes back as a merkle proof which is verifiable and then we can verify the code against the code hash.

gsalgado: Yeah, I guess so. Filed #849

gsalgado:
I'm sneaking a fix to the lightchain integration test here as well. It's a trivial change, but if you want to have a look and have any suggestions I'd be happy to deal with them post merge

Issue #780, opened 2018-05-25
Code duplication in peer handling services


jannikluhn:
### What is wrong?

As identified [here](https://github.com/ethereum/py-evm/pull/758#discussion_r190611644), various services (`FastChainSyncer`, `StateDownloader`, `LightPeerChain`, and `ShardSyncer`) implement the same peer handling pattern:

- They subscribe to a peer pool
- They start a `handle_peer` task whenever a new peer registers
- `handle_peer` takes messages from the peer's message queue and processes them depending on the message type
- When the services are stopped they cancel all peer handling tasks

### How can it be fixed

Add a common parent class (`PeerHandlingService`?) that implements that functionality.

Related: #746, #768

Issue #822, opened 2018-05-30
LightPeerChain must not cancel() its peers on exit


gsalgado:
LightPeerChain calls `.cancel()` on all connected peers when it is shutting down. That is the job of `PeerPool`'s shutdown routine

PR #842, opened 2018-05-31
p2p: Redesign our services' peer-handling code

gsalgado:
Depends on #841, please review that one first and then you can review just the last commit here.

They used to spawn one background coroutine for every peer, reading from
the peer's msg queue, but that is wasteful, complex and made it impossible
for peers to support multiple subscribers.

Now each service (as a PeerPoolSubscriber) provides their own msg queue and
all peers write to them, so the services just need to consume messages from
that queue

Closes: #780, #822


p2p/chain.py
```diff
@@ -193,28 +174,25 @@ def register_peer(self, peer: BasePeer) -> None:
         # find the common ancestor between our chain and the peer's.
         start_at = max(0, head.block_number - eth.MAX_HEADERS_FETCH)
         while not self._sync_complete.is_set():
-            if peer.is_finished:
+            if not peer.is_running:
```
pipermerriam: Why the inversion?  The `not` seems to make this slightly harder to read.

p2p/chain.py
```diff
@@ -443,9 +410,7 @@ def _handle_block_headers(self, headers: List[BlockHeader]) -> None:
         self.logger.debug("Got Receipts for %d blocks from %s", len(receipts_by_block), peer)
         loop = asyncio.get_event_loop()
         iterator = map(make_trie_root_and_nodes, receipts_by_block)
-        receipts_tries = await wait_with_token(
-            loop.run_in_executor(self._executor, list, iterator),
-            token=self.cancel_token)
+        receipts_tries = await self.wait_first(loop.run_in_executor(self._executor, list, iterator))
```
pipermerriam: Can we add a comment explaining what is happening here.  It makes sense to me because I know `map` is lazy and thus, the `list` application pushes the evaluation down into the executor.  Not sure how clear that will be to others reading the codebase.

p2p/state.py
```diff
@@ -58,7 +58,7 @@ def __init__(self,
         self.peer_pool = peer_pool
         self.root_hash = root_hash
         self.scheduler = StateSync(root_hash, account_db)
-        self._running_peers: Set[ETHPeer] = set()
+        self._msg_queue: asyncio.Queue[Tuple[BasePeer, protocol.Command, protocol._DecodedMsgType]] = asyncio.Queue()  # noqa: E501
```
pipermerriam: This should probably be moved to the `__init__` of `PeerPoolSubscriber` and done in a `super()` call.  I think it is repeated in 4+ locations.

p2p/chain.py
gsalgado: Because I got rid of the `self.finished` event and I don't think it makes sense to have both `is_running` and `is_finished` properties.

p2p/chain.py
gsalgado: Sure, that is indeed not clear

p2p/state.py
gsalgado: Problem is, `PeerPoolSubscriber` is mainly used in classes that have multiple parents and we already `super()` to the other parent's `__init__()`, so we'd need a separate, explicit call to this one's. I'd prefer to keep this small repetition

pipermerriam: Thoughts on making it lazily initialize itself within a `property` method?

gsalgado: Sounds good, done!

p2p/service.py
```diff
@@ -27,18 +26,35 @@ def __init__(self, token: CancelToken=None) -> None:
...
         """Await for the service's _run() coroutine.
 
-        Once _run() returns, set the finished event, call cleanup() and
+        Once _run() returns, triggers the cancel token, call cleanup() and
         finished_callback (if one was passed).
         """
         if self.is_running:
             raise RuntimeError("Cannot start the service while it's already running")
-        elif self.is_finished:
-            raise RuntimeError("Cannot restart a service after it has completed")
```
carver: This check has come in handy at least once. Instead of removing it, how about:

```py
        elif self.cancel_token.triggered:
            raise RuntimeError("Cannot restart a service after it has been canceled")
```

p2p/service.py
```diff
@@ -50,45 +66,32 @@ def __init__(self, token: CancelToken=None) -> None:
...
-        """Run the service's _cleanup() coroutine."""
+        """Run the service's _cleanup() coroutine and sets the cleaned_up event."""
         await self._cleanup()
-
         self.cleaned_up.set()
 
     async def cancel(self):
-        """Trigger the CancelToken and wait for the run() coroutine to finish."""
-        if self.is_finished:
-            self.logger.warning("Tried to cancel %s, but it was already finished", self)
```
carver: Same thing here. How about:

```py
        if self.cancel_token.triggered:
            self.logger.warning("Tried to cancel %s, but it was already cancelled", self)
```

p2p/state.py
```diff
@@ -138,15 +120,9 @@ def idle_peers(self) -> List[ETHPeer]:
...
-        self.logger.info("Waiting for %d running peers to finish", len(self._running_peers))
-        while time.time() < start_at + 1:
-            if not self._running_peers:
-                break
-            await asyncio.sleep(0.1)
-        else:
-            self.logger.info("Waited too long for peers to finish, exiting anyway")
+        # We don't need to cancel() anything, but we yield control just so that the coroutines we
+        # run in the background notice the cancel token has been triggered and return.
+        await asyncio.sleep(0)
```
carver: Isn't the promise of cleanup that all subservices are shut down by the time it returns?

If I understand that right, then I would expect a pattern like `await asyncio.gather(*[subservice.cleaned_up.wait() for subservice in subservices])` here.

(plus the same thing in p2p/chain)

p2p/chain.py
```diff
@@ -705,8 +666,8 @@ def _test() -> None:
 
     async def exit_on_sigint():
         await sigint_received.wait()
-        await syncer.cancel()
         await peer_pool.cancel()
+        await syncer.cancel()
```
carver: Why is this order important? Seems worth a comment.

p2p/service.py
gsalgado: Sure, I'll include that in one branch I'm currently working on

p2p/state.py
gsalgado: StateDownloader has no sub-services. It only runs a few coroutines in the background and I plan to have a new method on `BaseService` so that we can easily keep track of those and wait for them to return on `_cleanup()`, but right now there's no way to do that

p2p/chain.py
gsalgado: This is just the script to manually test the syncer, and I don't remember why I did that. In practice the syncer will usually be shut down before the `PeerPool` -- that's what happens in `FullNodeSyncer`. IOW, it doesn't really matter

PR #844, opened 2018-05-31
Easy profiling of trinity

pipermerriam:
### What was wrong?

We need to address performance.  Profiling lets us do this in a data driven manner.

### How was it fixed?

You can now run `trinity --profile` which will dump `cProfile` data into files that can then be parsed.

#### Cute Animal Picture

baby donkey-2


trinity/cli_parser.py
```diff
@@ -51,6 +51,7 @@ def __call__(self, parser, namespace, values, option_string=None):
 network_parser = parser.add_argument_group('network')
 syncing_parser = parser.add_argument_group('sync mode')
 chain_parser = parser.add_argument_group('chain')
+debug_parser = parser.add_argument_group('debug')
```
carver: Is this used anywhere?

PR #850, opened 2018-06-01
remove min_peers

pipermerriam:
### What was wrong?

`PeerPool.min_peers` was no longer used

### How was it fixed?

Removed it

#### Cute Animal Picture

cute-baby-donkeys-11

### DIFF
--- a/tests/p2p/integration_test_helpers.py
+++ b/tests/p2p/integration_test_helpers.py
@@ -31,9 +31,10 @@ class LocalGethPeerPool(HardCodedNodesPeerPool):
                  network_id: int,
                  privkey: datatypes.PrivateKey,
                  ) -> None:
-        min_peers = 1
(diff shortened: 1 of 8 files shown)


tests/p2p/integration_test_helpers.py
```diff
@@ -31,9 +31,8 @@ def __init__(self,
                  network_id: int,
                  privkey: datatypes.PrivateKey,
                  ) -> None:
-        min_peers = 1
         discovery = None
-        super().__init__(peer_class, chaindb, network_id, privkey, discovery, min_peers)
+        super().__init__(peer_class, chaindb, network_id, privkey, discovery)
```
gsalgado: We should use max_peers=1 here so that it doesn't unnecessarily try to connect to more peers

PR #851, opened 2018-06-01
Start implementing SMC service

jannikluhn:
WIP

#### Cute Animal Picture


p2p/smc_service.py
```diff
@@ -0,0 +1,153 @@
...
+class SMCService(BaseService):
+    w3: Web3 = None
+    config = get_sharding_config()
+
+    def __init__(self,
+                 smc_address: Address,
+                 shard_ids: List[int]=None,
+                 private_key: Optional[datatypes.PrivateKey]=None,
+                 token: CancelToken=None) -> None:
+        if self.w3 is None:
```
pipermerriam: Hrm, class level variable here seems potentially problematic.  Maybe this should be part of the `__init__` params.?

p2p/smc_service.py
```diff
@@ -0,0 +1,153 @@
...
+    ShardTracker,
+)
+from sharding.contracts.utils.config import (
+    get_sharding_config,
+)
+
+
+class SMCService(BaseService):
+    w3: Web3 = None
+    config = get_sharding_config()
```
pipermerriam: Same here with having this be a class level attribute.  It'd be pretty easy to keep this at the instance level.  Probably desirable since testing situations probably require different configuration.

p2p/smc_service.py
```diff
@@ -0,0 +1,153 @@
...
+        return 11 * 1000 * 1000 * 1000
+
+    def start_watching_shard(self, shard_id: int) -> asyncio.Queue[CollationHeader]:
+        if shard_id in self._shard_trackers:
+            raise ValueError("Already watching shard {}".format(shard_id))
+
+        self.logger.debug("Start watching shard {}".format(shard_id))
+        shard_tracker = ShardTracker(self.config, shard_id, self._log_handler, self.smc_address)
+        queue: asyncio.Queue[CollationHeader] = asyncio.Queue()
+        self._shard_trackers[shard_id] = shard_tracker
```
pipermerriam: Instead of two data structures, what about using `collections.namedtuple` for these two fields.

p2p/smc_service.py
```diff
@@ -0,0 +1,153 @@
...
+                        propser_address=proposer,
+                    )
+
+                    if header.hash in self._processed_header_hashes:
+                        continue
+
+                    self.logger.debug("Got AddHeader log for header %s", header)
+                    self._processed_header_hashes.add(header.hash)
+                    for added_header_queue in self._added_header_queues[shard_id]:
+                        await added_header_queue.put(header)
```
pipermerriam: curious if there is a reason to `put` instead of [`put_nowait`](https://docs.python.org/3/library/asyncio-queue.html#asyncio.Queue.put_nowait)

p2p/smc_service.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def add_header(self, collation_header: CollationHeader) -> None:
+        if self.private_key is None:
+            raise ValueError("No private key has been configured")
+        address = self.private_key.public_key.to_canonical_address
```
pipermerriam: missing function call here (`to_canonical_address()`)

carver: :+1: Wasn't this always raising a `ValidationError`? I'm surprised that didn't show up in testing.

p2p/smc_service.py
carver: The `await` seems cleaner than trying to catch a QueueFull and then wait for the queue to clear, then continue adding.

jannikluhn:
@pipermerriam Don't bother looking at the actual code and tests yet, but does the Tox/CircleCI setup makes sense like this? Alternatively, I thought about adding a command line flag to the tests (something like "--flaky-dependencies") and `skipif` the tests unless the flag is set which is done only in a special run. However, the `--ignore` option seemed to simpler for now (I don't think that there will be too many tests that require `eth-tester`).

pipermerriam:
General setup seems fine.  Probably the most useful thing we can have is clear documentation, maybe as comments in the appropriate places that helps someone understand why this is failing when it fails.

tox.ini
```diff
@@ -19,7 +19,8 @@ passenv =
     TRAVIS_EVENT_TYPE
 commands=
     core: py.test {posargs:tests/core/}
-    trinity: py.test {posargs:tests/trinity/core/}
+    trinity: py.test {posargs:tests/trinity/core/ --ignore tests/trinity/core/sharding/test_smc_service.py}
+    smc: py.test {posargs:tests/trinity/code/sharding/test_smc_service.py}
```
carver: This feels a little brittle, for example if we want to add another SMC test file. If it's worth running a separate SMC job (how long does it take to run?), maybe we can put it in its own test folder, like:

```py
    trinity: py.test {posargs:tests/trinity/core/}
    smc: py.test {posargs:tests/trinity/smc/}
```

jannikluhn:
- Put the tests in their own directory as suggested by @carver
- Added a comment in the tox.ini file, even though I don't think it will get a lot of visibility there (don't see a better location for it though)
- Added `eth-tester` and the `sharding` as dependencies (in their own section in `setup.py`, for further separation)
- Redesigned the log subscription mechanism slightly

tox.ini
```diff
@@ -14,13 +14,17 @@ ignore=
...
+install_command = pip install {opts} --process-dependency-links {packages}
 passenv =
     PYTEST_ADDOPTS
     TRAVIS_EVENT_TYPE
 commands=
     benchmark: {toxinidir}/scripts/benchmark/run.py
     core: py.test {posargs:tests/core/}
     trinity: py.test {posargs:tests/trinity/core/}
+    # The SMC tests may fail as they depend on eth-tester which does not necessarily work with
+    # py-EVMs master branch.
```
carver: If tests are known to be failing, in a way that should change in a future release, I think the best way to communicate that is with an xfail.

In this case, something like: `@pytest.mark.xfail(reason="depends on recent eth-tester, which doesn't work with current py-evm")` on the test that's failing.

trinity/smc_service.py
```diff
@@ -0,0 +1,181 @@
...
+        self._latest_complete_period = -1  # not even period 0 is complete yet
+
+        self._shard_trackers: Dict[int, ShardTracker] = {}
+        self._subscriptions: SubscriptionsDictType = collections.defaultdict(list)
+
+        self._processed_header_hashes: Set[Hash32] = set()
+
+    @property
+    def add_header_gas_price(self) -> int:
+        return 11 * 1000 * 1000 * 1000
```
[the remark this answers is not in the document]
jannikluhn: Just some more or less random placeholder value. In practice this should be dynamic, to react to changing gas prices, that's why I put it in a `@property`. But :+1:, for now we can make it an configurable constant.

PR #853, opened 2018-06-01
Enforce max_peers limit on inbound connections

pipermerriam:
### What was wrong?

The `max_peers` limit is not being enforced for inbount peer connections.

### How was it fixed?

Added enforcement to the `Server` class which handles these.

#### Cute Animal Picture

idkrrlunnt2gi722xp4qtmqswuvquncobhhx_-mydew


gsalgado:
While you're at it, would you mind removing https://github.com/ethereum/py-evm/blob/master/p2p/peer.py#L656-L662 as well? That's completely unnecessary as the main loop won't attempt to connect when the pool is already full

PR #854, opened 2018-06-01
Early exit when peer pool is full during _connect_to_nodes

pipermerriam:
### What was wrong?

Yet another code path that didn't respect the max peers.  In the case where max peers is set to 10 and we are currently connected to 9 peers, the code would connect to *up-to* 10 more peers, allowing us to end up with approximately `max_peers * 2 - 1` peer connections.

### How was it fixed?

Early break from the loop in `_connect_to_nodes` when the peer pool is full.

#### Cute Animal Picture

flat 800x800 070 f

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -726,6 +726,11 @@ class PeerPool(BaseService):
             # as discussed in
             # https://github.com/ethereum/py-evm/pull/139#discussion_r152067425
             peer = await self.connect(node)
+            if self.is_full:
+                self.logger.debug("Peer pool is full: disconnecting from %s", peer)
+                peer.disconnect(DisconnectReason.other)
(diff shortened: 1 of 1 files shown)


p2p/peer.py
```diff
@@ -726,6 +726,11 @@ def _get_random_bootnode(self) -> Generator[Node, None, None]:
             # as discussed in
             # https://github.com/ethereum/py-evm/pull/139#discussion_r152067425
             peer = await self.connect(node)
+            if self.is_full:
+                self.logger.debug("Peer pool is full: disconnecting from %s", peer)
+                peer.disconnect(DisconnectReason.other)
```
carver: Isn't there a `too many peers` disconnect reason?

PR #855, opened 2018-06-01
Yet another fix for enforcing max peers

pipermerriam:
### What was wrong?

And yet another stab at fixing max peer enforcement.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -726,16 +726,16 @@ class PeerPool(BaseService):
             # as discussed in
             # https://github.com/ethereum/py-evm/pull/139#discussion_r152067425
             peer = await self.connect(node)
-            if self.is_full:
+
+            if peer is None:
+                continue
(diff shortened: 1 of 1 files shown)


p2p/peer.py
```diff
@@ -726,16 +726,16 @@ def _get_random_bootnode(self) -> Generator[Node, None, None]:
             # as discussed in
             # https://github.com/ethereum/py-evm/pull/139#discussion_r152067425
             peer = await self.connect(node)
-            if self.is_full:
+
+            if peer is None:
+                continue
+            elif self.is_full:
```
gsalgado: We should simply refrain from attempting to self.connect() when we're full -- it's only for incoming connections that we should disconnect after the handshake.

pipermerriam: will work on getting this cleaned up today.

PR #861, opened 2018-06-04
p2p: No longer catch "unclean" exceptions anywhere

gsalgado:
Those used to happen when a trinity-proxied Chain/DB service was terminated
before the networking process, but that shouldn't happen anymore.

Also logs more info in trinity when terminating sub-processes

### DIFF
--- a/trinity/utils/ipc.py
+++ b/trinity/utils/ipc.py
@@ -14,33 +14,39 @@ def wait_for_ipc(ipc_path: str, timeout: int=1) -> None:
 
 
 def kill_process_gracefully(process: Process,
-                            logger: Logger=None,
+                            logger: Logger,
                             SIGINT_timeout: int=5,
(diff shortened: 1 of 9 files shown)


trinity/utils/ipc.py
```diff
@@ -14,33 +14,39 @@ def wait_for_ipc(ipc_path: str, timeout: int=1) -> None:
 
 
 def kill_process_gracefully(process: Process,
-                            logger: Logger=None,
+                            logger: Logger,
```
pipermerriam: Thoughts on just using a local logger for this?  Reason to send in an outside logger?

gsalgado: Not sure; I've just made the logger required to ensure we always log what goes on here, since that is important. Before this change `trinity.main()` was not passing a logger so it was all lost

pipermerriam:
good to :shipit: at your discretion.

PR #867, opened 2018-06-05
Implement peerCount JSON-RPC API

cburgdorf:
# THIS IS STILL WIP, NEEDS TESTS AND EATS KITTEN

### What was wrong?

The JSON-RPC API for `peerCount` wasn't implemented.

### How was it fixed?

Implemented it.

#### Cute Animal Picture


trinity/rpc/modules/net.py
```diff
@@ -10,3 +10,21 @@ def version(self):
         Returns the current network ID.
         """
         return str(self._chain.network_id)
+
+    def peerCount(self):
+        """
+        Return the number of peers that are currently connected to the node
+        """
+        if self._p2p_server.peer_pool is None:
+            return '0x0'
```
pipermerriam: I'm curious about this logic and why it's here.  Seems like the `net_` API should not be runnable if there is no peer pool.

trinity/rpc/modules/net.py
```diff
@@ -10,3 +10,21 @@ def version(self):
...
+        return hex(self._p2p_server.peer_pool.connected_peer_count)
+
+    def listening(self):
+        """
+        Return `True` if the client is actively listening for network connections
+        """
+        if self._p2p_server.peer_pool is None:
+            return False
+
+        return not self._p2p_server.peer_pool.is_full
```
pipermerriam: Are we sure this is the right logic.  We are technically still listening (have an outbound socket open that accepts incoming connections) even when the peer pool is full.  I think the `discovery` service is where we'd need to look to determine if we're listening or not.

trinity/rpc/modules/net.py
cburgdorf: Well, parts of the `net` module are independent from the peer pool (e.g. version) and by the time that the peer pool module is created, we don't have a peer pool yet. The peer pool is created asynchronously and it takes several seconds until it is set here: https://github.com/ethereum/py-evm/blob/f66d296074c0603eefa4e4c718421956d70d183e/p2p/server.py#L262

In other words, when you start trinity, the `net` module is already set up and answers to `net_peerCount` request and does therefore return `0x0` for when the `peer_pool` is still `None`. Which at least technically is correct :)

trinity/rpc/modules/main.py
```diff
@@ -1,8 +1,9 @@
 class RPCModule:
     _chain = None
 
-    def __init__(self, chain=None):
+    def __init__(self, chain=None, p2p_server=None):
```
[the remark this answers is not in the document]
cburgdorf: [See my comment](https://github.com/ethereum/py-evm/pull/867#discussion_r193181748) here, by the time that the module is initialized the `peer_pool` isn't available yet

p2p/peer.py
```diff
@@ -582,7 +582,11 @@ def __init__(self,
 
     @property
     def is_full(self) -> bool:
-        return len(self.connected_nodes) >= self.max_peers
+        return self.connected_peer_count >= self.max_peers
+
+    @property
+    def connected_peer_count(self) -> int:
```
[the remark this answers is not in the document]
cburgdorf: Sounds pythonic to me :+1:

trinity/rpc/modules/net.py
cburgdorf: I'm gonna double check that. Here is what pyethapp does which isn't the same as what we do but not that far away either :sweat_smile: 

https://github.com/ethereum/pyethapp/blob/9af6e64e6140104ab13fdbc5a6b22a915c412adc/pyethapp/jsonrpc.py#L696

trinity/rpc/modules/net.py
pipermerriam: Hrm, can you open an issue to track this as I think that we should fix that.

pipermerriam: Maybe if `peer_pool is None` should return a JSON-RPC error indicating that it's not ready?

trinity/rpc/modules/main.py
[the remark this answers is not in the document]
carver: We might need it anyway, to add `eth_syncing`.

carver:
If you haven't yet, I would recommend checking if this works on the light client. I suspect there will be problems because it doesn't track the peer pool the same way.

trinity/rpc/modules/net.py
pipermerriam: > in fact our p2p subsystem is still bootstrapping and we have in fact 0 connected peers at that time.

👍

cburgdorf: Wouldn't `hex(0)` cut it if you dislike the hardcoded string?

cburgdorf: @pipermerriam I created #877 regarding the `PeerPool` issue.

cburgdorf:
@pipermerriam @carver 

Here's a summary of the state of this:

- I implemented `__len__` for `PeerPool` as suggested
- I refactored and added more tests. E.g. the tests now cover the case when the `PeerPool` is still `None` during bootstrapping etc.
- I tested the PR against `--light` and found the added APIs to be working fine

I'm still not sure what to do about the implementation logic for `net_listening`. Parity always returns `true`, `pyethapp` returns `true` if the number of peers is lower than the `min_peers`. I don't know what kind of implications our answer on this API endpoint has so for now it strikes me as balanced to say we are listening as long as we haven't reached the max cap of peers.

pipermerriam:
I think `net_listening` should be `true` when we are listening (even if the peer pool is full).

trinity/rpc/modules/net.py
carver: > `// (this may not be true now -- 26 september 2016)`

lol

carver:
> I think net_listening should be true when we are listening (even if the peer pool is full).

I agree, [the spec says](https://github.com/ethereum/wiki/wiki/JSON-RPC#net_listening):
> `Returns true if client is actively listening for network connections.`

We are listening for connections even if full, just sending back a drop message.

Returning `False` before the p2p server has bound to the address/port makes sense, but I'm not sure how easy it is to get at that info, and it doesn't strike me as critical to implement right away.

PR #871, opened 2018-06-06
Reintroduce validate_seal method of Chain class

jannikluhn:
### What was wrong?

`MainnetTesterChain` was overriding `validate_seal` to do nothing, but this had no effect as it was not being used by `validate_block`.

### How was it fixed?

Add `validate_seal` in the `Chain` again.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -251,6 +251,10 @@ class BaseChain(Configurable, metaclass=ABCMeta):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
+    def validate_seal(self, header: BlockHeader) -> None:
+        raise NotImplementedError("Chain classes must implement this method")
+
+    @abstractmethod
     def validate_gaslimit(self, header: BlockHeader) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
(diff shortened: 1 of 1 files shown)


cburgdorf:
I *think* rather than reintroducing `validate_seal` on the `Chain` class, the `MainnetTesterChain` should rather ensure that `validate_seal` is monkey patched on the VMs.

jannikluhn:
@cburgdorf I can see your point: `validate_seal` feels like an internal method that a subclass should not rely on. But monkey-patching the VMs also sounds kind of dirty. So I'd rather leave it like this, also considering that it originally was supposed to work like this.

evm/chains/base.py
```diff
@@ -606,12 +610,17 @@ def validate_block(self, block: BaseBlock) -> None:
...
-
+        self.validate_seal(block.header)
         self.validate_uncles(block)
         self.validate_gaslimit(block.header)
 
+    def validate_seal(self, header: BlockHeader) -> None:
+        """
+        Validate the seal on the given header.
+        """
+        vm = self.get_vm()
```
carver: I believe this should be `self.get_vm(header)` so that we validate according to the correct fork's rules.

carver: Actually, I'm already working in the area, so I'll take care of it.

Issue #768, opened 2018-05-24
Peer must (eventually) handle multiple subscribers


gsalgado:
### What is wrong?

`PeerPool` provides an API to add subscribers (which get notified about new peers), and those subscribers then use `Peer.read_sub_proto_msg()` to read the messages sent by the remote. That works fine when we have a single subscriber but since all subscribers get a reference to the same `Peer` instance and that instance uses a single `asyncio.Queue` in `read_sub_proto_msg()`, it cannot handle multiple subscribers.

Right now this is not really a problem as our `PeerPool` instances never seem to have more than one subscriber at the same time.

### How can it be fixed

`Peer` must use a separate `asyncio.Queue` to store the sub-proto msgs for every subscriber. That may also require us to drop the `read_sub_proto_msg()` method and add a new API to subscribe, which would then return the subscriber's queue

PR #873, opened 2018-06-06
p2p: Peer now supports multiple subscribers

gsalgado:
Closes: #768

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -13,6 +13,7 @@ from abc import (
 
 from typing import (
     Any,
+    cast,
     Dict,
     Generator,
     Iterator,
@@ -21,7 +22,6 @@ from typing import (
     TYPE_CHECKING,
     Tuple,
     Type,
-    cast,
 )
 
 import sha3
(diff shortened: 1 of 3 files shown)


p2p/peer.py
```diff
@@ -169,6 +168,7 @@ def __init__(self,
         self.headerdb = headerdb
         self.network_id = network_id
         self.inbound = inbound
+        self._subscribers: List['asyncio.Queue[PEER_MSG_TYPE]'] = []
```
cburgdorf: If you define `PEER_MSG_TYPE = Tuple['BasePeer', protocol.Command, protocol._DecodedMsgType]`

At the beginning of the file, you can get rid of all the string type references (aka forward refs)

You can see this (or even pick from) here: https://github.com/cburgdorf/py-evm/commit/f5d534dfe062b5ebc140c54ac4462d6cf727de19

gsalgado: I cannot, because `asyncio.Queue` is an actual type and python will give a type error (TypeError: 'type' object is not subscriptable)

cburgdorf: Oh, you're right. Never mind.

gsalgado:
> Any thoughts on setting some reasonable upper bounds on queue size so ensure we don't somehow end up in a situation where a queue has tons and tons of items in it which aren't being consumed?

The tricky thing is deciding what would be *reasonable*, but I agree that's probably a good thing. I'm going with 10k for now, and I'm extending the pool's stats reporter to also log extra info (# of subscribers and length of longest subscriber queue) about every peer.

PR #878, opened 2018-06-06
apt-get install

MysticRyuujin:
### What was wrong?
Missing Keyword Install

### How was it fixed?
Added Keyword Install


cburgdorf:
Thanks!

PR #883, opened 2018-06-07
Validate headers for lightchain

carver:
### What was wrong?

Not much validation was happening on the lightchain headers. This caused us to track onto the ETC chain by accident (sometimes).

This is an alternative approach to #869 

### How was it fixed?

Roughly:
- Add new `validate_header` classmethod to the VM
- Pass in `Chain` class to `LightPeerChain` to get the VM at the appropriate block number and validate
- If header fails validation:
  - do not save header, stop importing from the peer
  - disconnect from peer
- added dao fork header rules to VM's `validate_header` for that fork choice
- some basic dao fork header tests
- did a basic trinity run check
- get ETC fork header for 1920000, for real chain validation in tests
- switch to validating headers with VM class instead of VM instance (to fix the broken `*-*-blockchain` tests that were getting an `evm.exceptions.HeaderNotFound: No uncles found for hash b'\x81\xa9\x9a\x84\x8cg\xca\x84C\x99\x8b\x17\x07\t\xe1\xb8$\x8b\xaf\xd5\xb6\xbc\xd5\xe9\x94/W1\xd2T\x1c\x12'`)

Planning to punt these to another PR:
- validate the genesis header against the genesis settings for the current network: #892
- blacklist peer from reconnecting: #893

#### Cute Animal Picture


p2p/lightchain.py
```diff
@@ -210,19 +224,62 @@ def peer_block_height(peer: LESPeer) -> int:
...
+        VM.validate_header(header, parent_header)
+
+    async def _get_parent_header(self, header):
+        parent_hash = header.parent_hash
+        # TODO extract 500 to constant
+        for attempt in range(500):
+            try:
+                return await self.headerdb.coro_get_block_header_by_hash(parent_hash)
+            except KeyError:
+                self.wait(asyncio.sleep(0.05))
```
gsalgado: Why would we get a KeyError and not a HeaderNotFound here? Also, in which situation is the retrying necessary?

p2p/lightchain.py
```diff
@@ -210,19 +224,62 @@ def peer_block_height(peer: LESPeer) -> int:
...
+        for header in headers:
+            try:
+                await self._validate_header(header)
+            except ValidationError as exc:
+                self.logger.info(
+                    "Dropping Peer %s for sending invalid header %s: %s",
+                    peer,
+                    header,
+                    exc,
+                )
```
gsalgado: I think it'd be better if you raised a `LESAnnouncementProcessingError` here, including the DisconnectReason there. That will cause a warning to be logged and the peer to be disconnected. Doing this you also won't need the `peer.connected` check and the `new_tip is not None` checks in `process_announcement()`

p2p/peer.py
```diff
@@ -283,6 +283,10 @@ def close(self):
         self.reader.feed_eof()
         self.writer.close()
 
+    @property
+    def connected(self):
+        return not self.reader.at_eof()
```
gsalgado: This worries me as a peer may be connected but no longer running. This may not be necessary with the changes I suggest above, but if something like it really is needed, we should consider using peer.is_running instead.

evm/vm/base.py
```diff
@@ -686,7 +673,52 @@ def validate_block(self, block):
...
+        # Some validations are different for the genesis block
+        if header.is_genesis:
+            if parent_header is None:
+                # TODO validate a variety of things about the genesis block, like a matching:
+                # - mix_hash
+                # - difficulty
+                # - timestamp
+                # - coinbase
+                # - extra_data
+                # - gas_limit
```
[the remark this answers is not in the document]
carver: I guess I have to look up more about how the genesis setup works. Does a genesis file always fully specify a header, or might it only partially specify the header?

If partial, it seems possible to have competing genesis headers based on the same file, where I suppose you would pick based on total difficulty of the tip of the chain. If that's right, then it seems like we to store genesis fields somewhere (in `Chain`?), for this check.

If always completely specified, then the only point of downloading is to make sure that we're on the same network, so the header should already be set, and we're just validating against the existing one. Then checking the hash works. Also, at `trinity` launch time: I suppose we need to create and store the genesis header if it's not already in the db.

p2p/lightchain.py
carver: Hm, for some reason I thought we might receive headers out of order, like:

```
Peer A: request 1-5
Peer B: request 6-10
Peer B: respond header 6-10
```

Anyway, that doesn't appear to be true (at least for now) so I can remove this whole method.

carver:
Seems like forward progress at this point. Merging, and will fix forward with any follow-up comments.

p2p/lightchain.py
```diff
@@ -65,13 +71,18 @@ class LightPeerChain(PeerPoolSubscriber, BaseService):
...
+            headerdb: 'BaseAsyncHeaderDB',
+            peer_pool: PeerPool,
+            Chain: Type[BaseChain]) -> None:
         super().__init__()
         self.headerdb = headerdb
         self.peer_pool = peer_pool
         self._announcement_queue: asyncio.Queue[Tuple[LESPeer, les.HeadInfo]] = asyncio.Queue()
         self._last_processed_announcements: Dict[LESPeer, les.HeadInfo] = {}
         self._pending_replies: Dict[int, Callable[[protocol._DecodedMsgType], None]] = {}
+        self.Chain = Chain
```
gsalgado: Why a capitalized variable name? Shouldn't it be `chain_class` to be consistent with other places?

p2p/lightchain.py
```diff
@@ -218,11 +229,37 @@ def peer_block_height(peer: LESPeer) -> int:
...
+            try:
+                await self._validate_header(header)
+            except ValidationError as exc:
+                raise LESAnnouncementProcessingError("Peer {} sent invalid header {}: {}".format(
+                    peer,
+                    header,
+                    exc,
+                ))
+            else:
+                await self.headerdb.coro_persist_header(header)
```
gsalgado: Need to use `self.wait()` here

p2p/lightchain.py
```diff
@@ -218,11 +229,37 @@ def peer_block_height(peer: LESPeer) -> int:
...
+                new_tip = header.block_number
+        return new_tip
+
+    async def _validate_header(self, header):
+        if header.is_genesis:
+            parent_header = None
+        else:
+            parent_header = await self.headerdb.coro_get_block_header_by_hash(header.parent_hash)
```
gsalgado: Need to use `self.wait()` here

evm/vm/base.py
```diff
@@ -249,6 +249,11 @@ def get_transaction_class(cls):
     def validate_block(self, block):
         raise NotImplementedError("VM classes must implement this method")
 
+    @classmethod
+    @abstractmethod
+    def validate_header(cls, header: BlockHeader, parent_header: BlockHeader) -> None:
```
pipermerriam: Looking at how this is called in `validate_block`, shouldn't this be `parent_header: Optional[BlockHeader]`?

evm/vm/base.py
```diff
@@ -686,7 +673,52 @@ def validate_block(self, block):
...
+                # - extra_data
+                # - gas_limit
+                pass
+            else:
+                raise ValidationError("Must not set parent header when validating genesis block")
+        else:
+            if parent_header is None:
+                raise ValidationError("Must set parent header if block is not genesis block")
+            else:
+                cls._validate_post_genesis_header(header, parent_header)
```
pipermerriam: Some thoughts on how to cleanup the readability here:

```python
if header.is_genesis and parent_header is not None:
    raise ...
elif parent_header is None and not header.is_genesis:
    raise ...
elif header.is_genesis:
    self._validate_genesis_header(...)
else:
    self._validate_post_genesis_header(...)
```

carver: #895 took this a fairly different direction and said `validate_header` is only to be used on non-genesis headers, which cleaned things up a lot.

Issue #877, opened 2018-06-06
Check if we can initialize the PeerPool earlier


cburgdorf:
### What is wrong?

This is a follow up to the discussion @pipermerriam and I had [here](https://github.com/ethereum/py-evm/pull/867#discussion_r193187256).

Currently the `PeerPool` isn't available until some other async tasks are resolved. This issue serves as a reminder to check whether the `PeerPool` could be created synchronously earlier.

### How can it be fixed

A possible idea would be to create the `PeerPool` early on without the dependencies that first need to be resolved asynchronously and then as a second step, once these dependencies are resolved, call some initialization API on the `PeerPool`. Before that, the `PeerPool` would already exist but would reside in some sort of `bootstrapping` mode until that API is called.

Not sure, if I like that tbh. ¯\_(ツ)_/¯

gsalgado:
ISTM that the problem here is the fact that `Server.peer_pool` only gets set in `Server._run()` and not in `Server.__init__()`, as it should. IIUC we do that because of some (far from ideal) server or sharding tests. If we fix those tests we could then initiate a `PeerPool` and assign it to `Server.peer_pool` in `Server.__init__()`, as I think we should do

cburgdorf:
@gsalgado That is correct but the `PeerPool` depends on the `DiscoveryProtocol`

https://github.com/ethereum/py-evm/blob/af3ac1ddd3c5cf18253b79ce4c9ea356556c83f1/p2p/server.py#L266

Which depends on the external ip address

https://github.com/ethereum/py-evm/blob/af3ac1ddd3c5cf18253b79ce4c9ea356556c83f1/p2p/server.py#L263-L264

Which we get from the `upnp_dev`

https://github.com/ethereum/py-evm/blob/af3ac1ddd3c5cf18253b79ce4c9ea356556c83f1/p2p/server.py#L252

Which we `await` on in the beginning.

https://github.com/ethereum/py-evm/blob/af3ac1ddd3c5cf18253b79ce4c9ea356556c83f1/p2p/server.py#L249

So what I meant to say is that it seems to me that we can create the `PeerPool` early on in `__init__` but we either have to pass in something like an `Awaitable[DiscoveryProtocol]` and `await` that inside or we introduce some method to set the `DiscoveryProtocol` after creation. Either way it seems that this would leave the `PeerPool` in a state where it can't fully operate yet until the `DiscoveryProtocol` (which is waiting on the external IP) is resolved.

Am I reading that wrong?

[...]

PR #886, opened 2018-06-08
Make PeerPool available earlier

cburgdorf:
### What was wrong?

Currently the `PeerPool` on the server appears asynchronously some time later after initialization. See #877. This turns out as being problematic as we often want to pass the `PeerPool` somewhere else without passing the entire server.

### How was it fixed?

The reason the `PeerPool` doesn't immediately appear on the server today is because the server asynchronously awaits on the upnpdevice before it can build the `DiscoveryProtocol` that the ` `PeerPool` depends on.

Now what this PR does is that it allows the `PeerPool` to accept a `Future[DiscoveryProtocol]` instead of a `DiscoveryProtocol`.

I'm not super sold on that because it essentially just pushes the problem one level down (now the `PeerPool` will *sometimes* have the `DiscoveryProtocol` appearing asynchronously sometime after initialization.

The other alternative would be to make all methods that depend on the `DiscoveryProtocol` async but I'm not sure how far that would creep through all depending APIs.

As I'm working on the minimal Tx pool, I figured that I may need the `PeerPool` to be available early so I'm laying this out here for further discussion.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -22,6 +22,7 @@ from typing import (
     TYPE_CHECKING,
     Tuple,
     Type,
+    Union,
 )
 
 import sha3
@@ -578,7 +579,7 @@ class PeerPool(BaseService):
                  headerdb: 'BaseAsyncHeaderDB',
                  network_id: int,
                  privkey: datatypes.PrivateKey,
-                 discovery: DiscoveryProtocol,
+                 discovery: Union[DiscoveryProtocol, 'asyncio.Future[DiscoveryProtocol]'],
                  max_peers: int = DEFAULT_MAX_PEERS
                  ) -> None:
         super().__init__()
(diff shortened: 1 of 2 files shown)


gsalgado:
What if instead we simplified `PeerPool` by moving the responsibility of doing peer discovery to some other component (e.g. `p2p.Server`)?  That way `PeerPool` would only need to provide APIs to add new peers to the pool and notify its subscribers about them. I think it doesn't make sense for `PeerPool` to handle discovery anyway, and that would allow us to fix this issue in a cleaner way

cburgdorf:
@gsalgado I agree on that and in fact I was considering it for a moment but I then wasn't sure how feasible that would be. Now that *you* bring it up I'm inclined to give that more thought.

However, since we have #898 waiting on this and also #904 making changes to the `PeerPool` would be ok to take it as is and open up a ticket to follow up with that refactoring later?

gsalgado:
How urgent is #898? I'll have a look at moving discovery to `p2p.Server` today

cburgdorf:
@gsalgado would be nice if it makes it into the next release (I think we are aiming for one this week) but if it doesn't that's not the end of the world either. Anyway, if you're looking into this today, we'll wait. :+1:

cburgdorf:
Replaced by #934. Closing.

PR #889, opened 2018-06-11
Add guide on high level Trinity architecture

cburgdorf:
### What was wrong?

The documentation is lacking a high level architecture guide for Trinity (e.g. explain layer, processes and other core things)

### How was it fixed?

Wrote such a guide.

**Teaser**

guide_tease_2

Known issues:

- the graphics aren't that professional. I'm not a designer and these should act as stubs until someone who knows this stuff steps in
- still missing plenty of things that would fit into the topic
- links to some APIs don't work yet because we don't have them in our API docs yet
- links to some Python APIs don't work yet because I assume we need to configure something for that)

#### Cute Animal Picture


carver:
> links to some Python APIs don't work yet because I assume we need to configure something for that

Which python API reference is not linking as expected? Intersphinx is set up correctly for linking out to python, afaict.

cburgdorf:
@carver turns out I just had the linking syntax wrong. Links to `BaseManager` and `BaseProxy` are now working as intended.

PR #895, opened 2018-06-11
validate genesis header in light sync

### LINKED ISSUE #892: Validate genesis header
### What is wrong?

Right now, we aren't doing anything to validate that the genesis header we download is compatible with the genesis setup used in `trinity`. (not that we support arbitrary genesis settings now, but we will at some point)

### How can it be fixed

See some discussion at: https://github.com/ethereum/py-evm/pull/883#discussion_r194305867

Evaluate the genesis file setup we'll use (I think an EIP is in draft or merged on this). Does it fully specify the header? If so, force header creation before launching p2p. If not, save the genesis fields for later validation.

carver:
### What was wrong?

Fixes #892 

### How was it fixed?

When receiving a genesis header, check if it matches the canonical header already at block number 0. If they are not equal, raise a validation error that will drop the peer. 

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -609,6 +609,8 @@ class Chain(BaseChain):
         Since block validation (specifically the uncle validation) must have
         access to the ancestor blocks, this validation must occur at the Chain
         level.
+
+        Cannot be used to validate genesis block.
         """
         VM = self.get_vm_class_for_block_number(BlockNumber(block.number))
         parent_block = self.get_block_by_hash(block.header.parent_hash)
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -612,6 +612,8 @@ class Chain(BaseChain):
 
         Cannot be used to validate genesis block.
         """
+        if block.is_genesis:
+            raise ValidationError("Cannot validate genesis block this way")
         VM = self.get_vm_class_for_block_number(BlockNumber(block.number))
         parent_block = self.get_block_by_hash(block.header.parent_hash)
         VM.validate_header(block.header, parent_block.header)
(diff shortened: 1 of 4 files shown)


p2p/lightchain.py
```diff
@@ -253,12 +253,18 @@ def peer_block_height(peer: LESPeer) -> int:
 
     async def _validate_header(self, header):
         if header.is_genesis:
-            parent_header = None
+            genesis_header = await self.headerdb.coro_get_canonical_block_header_by_number(0)
```
gsalgado: As I said in https://github.com/ethereum/py-evm/issues/892#issuecomment-396387941, I think we should never download the genesis header as we must already have it on our first run, and we'll be comparing its hash with that of every peer we connect to. I think what we need to do is fix `LightPeerChain.get_sync_start_block()` to never request the genesis header

evm/chains/base.py
```diff
@@ -609,6 +609,8 @@ def validate_block(self, block: BaseBlock) -> None:
         Since block validation (specifically the uncle validation) must have
         access to the ancestor blocks, this validation must occur at the Chain
         level.
+
+        Cannot be used to validate genesis block.
```
pipermerriam: Should this be explicitely handled to raise an exception if `block.header.is_genesis`?

carver:
Ready for re-review

evm/vm/base.py
```diff
@@ -636,10 +636,10 @@ def validate_block(self, block):
             )
 
         if block.is_genesis:
-            parent_header = None
+            validate_length_lte(block.header.extra_data, 32, title="BlockHeader.extra_data")
```
pipermerriam: Not to be fixed now, but this inline header validation bugs me.  It'd be nice to just have dedicated `validate_genesis_header` and `validate_chain_header` functions or something similar.  Otherwise we cannot support alternate genesis header formats without fully overriding this function.

PR #896, opened 2018-06-11
Send disconnect reasons

carver:
### What was wrong?

We weren't accurately reporting the disconnect reason to peers.

### How was it fixed?

- Switched `subprotocol_error` from 0xc to 0x10
- Removed old usages of `error` reason
- Used more specific reasons in places that used to have `other`
- Let `TimeoutError` bubble up, and disconnect from peer with timeout reason

#### Cute Animal Picture

### DIFF
--- a/p2p/lightchain.py
+++ b/p2p/lightchain.py
@@ -49,6 +49,9 @@ from p2p.exceptions import (
 from p2p import les
 from p2p import protocol
 from p2p.constants import REPLY_TIMEOUT
+from p2p.p2p_proto import (
+    DisconnectReason,
+)
 from p2p.peer import (
     BasePeer,
     LESPeer,
@@ -87,6 +90,10 @@ class LightPeerChain(PeerPoolSubscriber, BaseService):
     def register_peer(self, peer: BasePeer) -> None:
         peer = cast(LESPeer, peer)
         self._announcement_queue.put_nowait((peer, peer.head_info))
+
+    async def disconnect_peer(self, peer: LESPeer, reason: DisconnectReason) -> None:
+        peer.disconnect(reason)
+        await self.drop_peer(peer)
 
     async def drop_peer(self, peer: LESPeer) -> None:
         self._last_processed_announcements.pop(peer, None)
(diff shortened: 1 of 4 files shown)


p2p/lightchain.py
```diff
@@ -162,10 +169,14 @@ def peer_block_height(peer: LESPeer) -> int:
                 self._last_processed_announcements[peer] = head_info
             except OperationCancelled:
                 self.logger.debug("Asked to stop, breaking out of run() loop")
+                await self.disconnect_peer(peer, DisconnectReason.client_quitting)
```
gsalgado: This is already done in `PeerPool` (although it should send the disconnect reason), as that is where the disconnection will actually take place. Here it is possible we quit the LightChain without stopping the PeerPool, in which case we wouldn't want to disconnect from the peer

p2p/lightchain.py
```diff
@@ -202,8 +213,8 @@ def peer_block_height(peer: LESPeer) -> int:
                 raise LESAnnouncementProcessingError(
                     "No common ancestors found between us and {}".format(peer))
             except TooManyTimeouts:
-                raise LESAnnouncementProcessingError(
-                    "Too many timeouts when fetching headers from {}".format(peer))
+                self.logger.debug("Too many timeouts when getting common ancestors from %s", peer)
```
gsalgado: I don't see the point of the debug log here given that you let the exception bubble up and then log a warning with it further up.

p2p/lightchain.py
```diff
@@ -227,8 +238,8 @@ def peer_block_height(peer: LESPeer) -> int:
                 # block to work around https://github.com/ethereum/go-ethereum/issues/15447
                 batch = await self.fetch_headers(start_block, peer)
             except TooManyTimeouts:
-                raise LESAnnouncementProcessingError(
-                    "Too many timeouts when fetching headers from {}".format(peer))
+                self.logger.debug("Too many timeouts when fetching headers from %s", peer)
```
gsalgado: Same here

p2p/lightchain.py
carver: Ah, ok, I'll take a look at sending the `client_quitting` when `PeerPool` shuts down.

When would we want to quit the `LightPeerChain` but leave the `PeerPool` running?

p2p/lightchain.py
carver: Mostly so we could see the context of what it was working on when it timed out. The warning log will only have whatever message got raised during `TooManyTimeouts`. I'm pretty neutral on this, though, so if you still think it's not helpful info, I'm okay to drop. :+1:

Issue #866, opened 2018-06-04
Trinity should relay transactions


pipermerriam:
### What is wrong?

Currently we do nothing when people send us transactions.

### How can it be fixed

Basic transaction relaying should be a good minimal first step towards a transaction pool.

1. A new long running service, `TransactionPool`
2. The pool should maintain something like a `Queue` where batches of transactions may be placed.
3. The pool should maintain individual bloom filters for each peer to track what transactions we've sent to each peer.

When we receive transactions, push them onto the `TransactionPool` queue.  Ensure that we update the bloom filter for the peer who sent us the transactions with the newly received transactions.

In the main loop of the `TransactionPool` service:

- pull a batch of transactions from the queue.
- for each peer, filter the transactions to only ones that have not been transmitted to that peer and send them the transactions.

pipermerriam:
Partially wanting to prioritize this because it'll clean up the DEBUG logs a lot, but it also helps make us a better network citizen.

PR #898, opened 2018-06-12
Minimale Viable Tx Pool

cburgdorf:
### What was wrong?

As specified in #866 we need a minimal viable tx pool that may not work as a proper pool at first but at least relay transactions to peers. This PR serves as a stub to signal that I'm on it.

### How was it fixed?

Implemented minimal viable transaction pool that simply relays transactions among connected peers.

#### Cute Animal Picture

### DIFF
--- a/p2p/eth.py
+++ b/p2p/eth.py
@@ -5,12 +5,13 @@ from rlp import sedes
 
 from evm.rlp.headers import BlockHeader
 from evm.rlp.receipts import Receipt
+from evm.rlp.transactions import BaseTransactionFields
 
 from p2p.protocol import (
     Command,
     Protocol,
 )
-from p2p.rlp import BlockBody, P2PTransaction
+from p2p.rlp import BlockBody
 from p2p.sedes import HashOrNumber
 
 
@@ -40,7 +41,7 @@ class NewBlockHashes(Command):
 
 class Transactions(Command):
     _cmd_id = 2
-    structure = sedes.CountableList(P2PTransaction)
+    structure = sedes.CountableList(BaseTransactionFields)
 
 
 class GetBlockHeaders(Command):
@@ -72,7 +73,7 @@ class NewBlock(Command):
     _cmd_id = 7
     structure = [
         ('block', sedes.List([BlockHeader,
-                              sedes.CountableList(P2PTransaction),
+                              sedes.CountableList(BaseTransactionFields),
                               sedes.CountableList(BlockHeader)])),
         ('total_difficulty', sedes.big_endian_int)]
 
@@ -178,7 +179,7 @@ class ETHProtocol(Protocol):
         header, body = cmd.encode(receipts)
         self.send(header, body)
 
-    def send_transactions(self, transactions: List[P2PTransaction]) -> None:
+    def send_transactions(self, transactions: List[BaseTransactionFields]) -> None:
         cmd = Transactions(self.cmd_id_offset)
         header, body = cmd.encode(transactions)
         self.send(header, body)
--- a/p2p/eth.py
+++ b/p2p/eth.py
@@ -177,3 +177,8 @@ class ETHProtocol(Protocol):
         cmd = Receipts(self.cmd_id_offset)
         header, body = cmd.encode(receipts)
         self.send(header, body)
+
+    def send_transactions(self, transactions: List[P2PTransaction]) -> None:
+        cmd = Transactions(self.cmd_id_offset)
+        header, body = cmd.encode(transactions)
+        self.send(header, body)
--- a/p2p/rlp.py
+++ b/p2p/rlp.py
@@ -1,3 +1,7 @@
+from eth_hash.auto import (
+    keccak
+)
+
 import rlp
 from rlp import sedes
 
@@ -9,6 +13,10 @@ from evm.rlp.transactions import BaseTransaction
 # instantiated.
 class P2PTransaction(rlp.Serializable):
     fields = BaseTransaction._meta.fields
+
+    @property
+    def hash(self) -> bytes:
+        return keccak(rlp.encode(self))
 
 
 class BlockBody(rlp.Serializable):
--- a/p2p/rlp.py
+++ b/p2p/rlp.py
@@ -1,26 +1,13 @@
-from eth_hash.auto import (
-    keccak
-)
-
 import rlp
 from rlp import sedes
 
+
 from evm.rlp.headers import BlockHeader
-from evm.rlp.transactions import BaseTransaction
-
-
-# This is needed because BaseTransaction has several @abstractmethods, which means it can't be
-# instantiated.
-class P2PTransaction(rlp.Serializable):
-    fields = BaseTransaction._meta.fields
-
-    @property
-    def hash(self) -> bytes:
-        return keccak(rlp.encode(self))
+from evm.rlp.transactions import BaseTransactionFields
 
 
 class BlockBody(rlp.Serializable):
     fields = [
-        ('transactions', sedes.CountableList(P2PTransaction)),
+        ('transactions', sedes.CountableList(BaseTransactionFields)),
         ('uncles', sedes.CountableList(BlockHeader))
     ]
--- a/tests/p2p/test_tx_pool.py
+++ b/tests/p2p/test_tx_pool.py
@@ -8,8 +8,8 @@ from p2p.peer import (
 from p2p.eth import (
     Transactions
 )
-from p2p.rlp import (
-    P2PTransaction
+from evm.rlp.transactions import (
+    BaseTransactionFields
 )
 
 from trinity.tx_pool.pool import (
@@ -140,7 +140,7 @@ def create_tx_recorder(monkeypatch, peer):
 
 
 def create_random_tx():
-    return P2PTransaction(
+    return BaseTransactionFields(
         nonce=0,
         gas_price=1,
         gas=21000,
--- a/tests/p2p/test_tx_pool.py
+++ b/tests/p2p/test_tx_pool.py
@@ -1,0 +1,153 @@
+import asyncio
+import pytest
+import random
+
+from p2p.peer import (
+    ETHPeer,
+)
+from p2p.eth import (
+    Transactions
+)
+from p2p.rlp import (
+    P2PTransaction
+)
+
+from trinity.tx_pool.pool import (
+    TxPool,
+)
+
+from tests.p2p.peer_helpers import (
+    get_directly_linked_peers,
+    MockPeerPoolWithConnectedPeers,
+)
+
+# TODO: Move this file into the trinity tests (Requires refactor of peer_helpers)
+
+
+class TxsRecorder():
+    def __init__(self):
+        self.recorded_tx = []
+        self.send_count = 0
+
+    def send_txs(self, txs):
+        self.recorded_tx.extend(txs)
+        self.send_count = self.send_count + 1
+
+
+@pytest.mark.asyncio
+async def test_tx_propagation(monkeypatch, request, event_loop):
+    peer1, peer2 = await get_directly_linked_peers(
+        request,
+        event_loop,
+        peer1_class=ETHPeer,
+        peer2_class=ETHPeer,
+    )
+
+    # We intercept sub_proto.send_transactions to record detailed information
+    # about which peer received what and was invoked how often.
+    peer1_txs_recorder = create_tx_recorder(monkeypatch, peer1)
+    peer2_txs_recorder = create_tx_recorder(monkeypatch, peer2)
+
+    pool = TxPool(MockPeerPoolWithConnectedPeers([peer1, peer2]))
+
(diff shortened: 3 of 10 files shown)


trinity/nodes/full.py
```diff
@@ -23,6 +26,9 @@ def __init__(self, chain_config: ChainConfig) -> None:
         self._node_key = chain_config.nodekey
         self._node_port = chain_config.port
         self._max_peers = chain_config.max_peers
+        self._tx_pool = TxPool()
+        self.add_service(self._tx_pool)
```
carver: Shouldn't the light node also need to propagate the transaction pool? If so, this can go in `Node` instead of `FullNode`.

cburgdorf: That was indeed, one of the questions I had in mind. I wasn't sure if a light node relays transactions similar to a full node. Thanks for jumping in to clarify!

trinity/tx_pool/base.py
```diff
@@ -0,0 +1,68 @@
...
+
+    async def _handle_tx(self, peer: ETHPeer, msg: List[P2PTransaction]) -> None:
+        self.logger.info("Wohoo, let's do work!")
+        self.logger.info(str(msg))
+
+        for receiving_peer in self.peer_pool.peers:
+            receiving_peer = cast(ETHPeer, receiving_peer)
+            if receiving_peer is not peer:
+                receiving_peer.sub_proto.send_transactions(
+                    self._filter_tx_for_peer(receiving_peer, msg)
```
pipermerriam: The filtering should probably happen first, and then only send if the result of filtering is not an empty list.

cburgdorf: Yes, good catch. I plan to cover all of that with test as well.

trinity/tx_pool/pool.py
```diff
@@ -0,0 +1,82 @@
...
+class TxPool(BaseService, PeerPoolSubscriber):
+    """
+    Minimal viable Transaction Pool
+    """
+    logger = logging.getLogger("trinity.tx_pool.TxPool")
+
+    def __init__(self, peer_pool: PeerPool) -> None:
+        super().__init__()
+        self._peer_pool = peer_pool
+        self._bloom = BloomFilter()
```
pipermerriam: Can we be explicit about the bloom filter parameters as well as having a comment here which documents what our collision rates are likely to be with the given parameters.

p2p/rlp.py
```diff
@@ -1,18 +1,20 @@
...
+# )
+
 import rlp
 from rlp import sedes
 
-from evm.rlp.headers import BlockHeader
-from evm.rlp.transactions import BaseTransaction
-
+# from eth_typing import (
+#     Address
```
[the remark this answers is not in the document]
cburgdorf: Refactoring leftover. Will fix.

p2p/eth.py
```diff
@@ -40,7 +41,7 @@ class NewBlockHashes(Command):
 
 class Transactions(Command):
     _cmd_id = 2
-    structure = sedes.CountableList(P2PTransaction)
+    structure = sedes.CountableList(BaseTransactionCommonFields)
```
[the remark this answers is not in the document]
cburgdorf: They are related (because I need `hash()` on the `P2PTransaction` and getting rid of that `P2PTransaction` alltogether seems like a good thing to me. But I'm happy to move that into a seperate commit to make the core of the tx pool change stand out better.

trinity/tx_pool/pool.py
cburgdorf: >what our collision rates are likely to be with the given parameters

 ¯\_(ツ)_/¯

tests/p2p/test_tx_pool.py
```diff
@@ -0,0 +1,154 @@
...
+from trinity.tx_pool.pool import (
+    TxPool,
+)
+
+from tests.p2p.peer_helpers import (
+    get_directly_linked_peers,
+    MockPeerPoolWithConnectedPeers,
+)
+
+# TODO: This whole file should live in the trinity tests but can't
```
cburgdorf: @pipermerriam any advice how to deal with that? My limited Python fu tells me the problem is that most of the directories within `tests` aren't actually Python modules (no `__init__.py`) and that is the reason why I can't cross import it (`..relative.lookup` also doesn't work).

But I read somewhere that NOT making the test directories proper python modules is actually the right way to handle test directories because of some pytest madness. Any advice how to handle this?

pipermerriam: Maybe relocating `peer_helpers` to something like `p2p.utils.testing.peer_helpers`.  That will cause it to be shipped with the actual `p2p` module but I don't see that as a big deal as long as the overhead remains small (like not including large fixture files or something like that).

cburgdorf: Mmh...not terribly excited about that :sweat_smile: I digged a bit further and noticed that `tests` already had a `__init__.py` but `tests/p2p` did not. So, I tried what would happen if we make that a module and this seems to work: https://github.com/ethereum/py-evm/pull/898/commits/882780972fae53777b5e129d1f23b51851a6563d

Still has some things failing but I think that could be worked out. But maybe we leave this out of this PR and I address that in a follow up?

p2p/server.py
```diff
@@ -233,24 +237,37 @@ def _make_syncer(self, peer_pool: PeerPool) -> BaseService:
...
-            discovery,
+            fut_discovery,
             max_peers=self.max_peers,
         )
 
-    async def _run(self) -> None:
-        self.logger.info("Running server...")
-        upnp_dev = await self._discover_upnp_device()
+    async def _make_discovery(self,
+                              upnp_device: Awaitable[upnpclient.Device]) -> DiscoveryProtocol:
```
carver: Passing in a single upnp device will not be sufficient for discovery. Take a look at the proposed refactor here, which loops over multiple devices: https://github.com/ethereum/py-evm/pull/905

cburgdorf: Yep, that commit basically just tries to solve the problem of not having the `PeerPool` available with the instantiating of the `Server`. It's a small refactor which doesn't change the rest of the process. I can rebase that in case #905 lands first.

tests/p2p/test_tx_pool.py
cburgdorf: I slightly changed that comment and plan to address that test refactoring in a follow up PR to not grow this PR with more unrelated changes.

p2p/server.py
```diff
@@ -233,24 +237,37 @@ def _make_syncer(self, peer_pool: PeerPool) -> BaseService:
...
+
+        upnp_dev = await upnp_device
         external_ip = '0.0.0.0'
         if upnp_dev is not None:
-            external_ip = upnp_dev.WANIPConn1.GetExternalIPAddress()['NewExternalIPAddress']
             await self._add_nat_portmap(upnp_dev)
+            external_ip = upnp_dev.WANIPConn1.GetExternalIPAddress()['NewExternalIPAddress']
+
+        addr = Address(external_ip, self.port, self.port)
```
pipermerriam: Just realized this will collide with our `eth_typing.Address`.  Just pointing it out.

trinity/tx_pool/pool.py
```diff
@@ -0,0 +1,98 @@
...
+
+        This is a minimal vialble implementation that only relays transactions but doesn't actually
+        hold on to them yet. It's still missing many features of a grown up transaction pool.
+    """
+    logger = logging.getLogger("trinity.tx_pool.TxPool")
+
+    def __init__(self, peer_pool: PeerPool) -> None:
+        super().__init__()
+        self._peer_pool = peer_pool
+        self._bloom = BloomFilter(max_elements=10000)
```
pipermerriam: some quick checks suggest `10,000` transactions is equivalent to about 90 blocks at current mainnet volume.

Lets bump this to something like 1 million, which looks like it will consume about 1mb of memory.  This also suggests a rolling bloom filter (maybe per peer) solution is going to be a priority very quickly.

p2p/server.py
cburgdorf: `Address` is a pretty generic name so I expect it to exist in many different modules. That said, in this module only `kademlia.Address` is used. Do you want me to change that to `kademlia.Address` to make that more explicit?

trinity/tx_pool/pool.py
cburgdorf: Will bump it up to 1m. I'd also like to check what geth and Parity do. If they do a rolling bloom filter or something else and how exactly. But yeah, can make that a high priority.

p2p/server.py
pipermerriam: No, just wanted to point it out as we'll probably make the mistake eventually.

cburgdorf:
@pipermerriam @gsalgado This is now rebased on top of the refactored `PeerPool`. I also bumped up the bloom filter size as suggested.

trinity/tx_pool/pool.py
```diff
@@ -0,0 +1,100 @@
...
+
+    async def _handle_tx(self, peer: ETHPeer, txs: List[BaseTransactionFields]) -> None:
+
+        self.logger.debug('Received transactions from %r: %r', peer, txs)
+
+        self._add_txs_to_bloom(peer, txs)
+
+        for receiving_peer in self._peer_pool.peers:
+            receiving_peer = cast(ETHPeer, receiving_peer)
+            if receiving_peer is not peer:
```
pipermerriam: I've found that in loops like this the following is easier to read.  Thoughts?

```python
for thing in many_things:
    if not some_required_condition(thing):
        continue

    do_for_loop_body_logic(thing)

# instead of
for thing in many_things:
    if some_required_condition(thing):
        do_for_loop_body_logic(thing)
```

Basically, rather than making the context deeper, trying to keep the context as shallow as possible by putting the early exit conditions first and then maintaining the shallow context level for the main function body.

For bodies with a single conditional this isn't a huge improvement, but when we get to having two conditional checks:

```python
for thing in many_things:
    if not some_required_condition(thing):
        continue
    if some_forbidden_condition(thing):
        continue

    do_for_loop_body_logic(thing)

# instead of
for thing in many_things:
    if some_required_condition(thing):
        if not some_forbidden_condition(thing):
            do_for_loop_body_logic(thing)
```

trinity/tx_pool/pool.py
```diff
@@ -0,0 +1,100 @@
...
+                    )
+                    receiving_peer.sub_proto.send_transactions(filtered_tx)
+                    self._add_txs_to_bloom(receiving_peer, filtered_tx)
+
+    def _filter_tx_for_peer(
+            self,
+            peer: BasePeer,
+            txs: List[BaseTransactionFields]) -> List[BaseTransactionFields]:
+
+        return [val for val in txs if self._construct_bloom_entry(peer, val) not in self._bloom]
```
pipermerriam: *nitpick*

I find these easier to read as

```python
return [
    val 
    for val 
    in txs 
    if self._construct_bloom_entry(peer, val) not in self._bloom
]
# or if you're more into brevity

return [
    val for val in txs 
    if self._construct_bloom_entry(peer, val) not in self._bloom
]
```

trinity/tx_pool/pool.py
cburgdorf: I can live with that. I'm personally more used to something like this (python-pseudocode):

```python

things.filter(lambda thing: return thing.is_cool())
          .filter(lambda thing: return thing.rocks())
          .filter(lambda thing: return thing.doesn_it_kitten())
          .for_each(lamda thing: do_stuff(thing))
```

I find this far more readable but I'm not sure if something similarish exists. I guess with lambdas being restricted to single lines in Python this probably doesn't work so well in this ecosystem.

trinity/tx_pool/pool.py
cburgdorf: yep, same here, I'm more used to 

```python
txs.filter(lambda tx: return self._construct_bloom_entry(peer, tx) not in self._bloom)
```

I like the second of your suggestions more than the first.

trinity/tx_pool/pool.py
cburgdorf: Oh and I do agree that rightwards drift is ugly and should be avoided. I'm just not used that much to that loop skipping style.

pipermerriam: You can accomplish this pretty easily with `toolz/cytoolz` but the result is far less readable.  

```
tuple(pipe(
    things,
    functools.partial(filter, operator.methodcaller('is_cool')),
    functools.partial(filter, operator.methodcaller('rocks')),
    functools.partial(filter, operator.methodcaller('does_it_kitten')),
    functools.partial(map, do_stuff)
))
```

I suspect there is a data pipeline library out there somewhere which allows for a smoother api.

cburgdorf: >but the result is far less readable

Well...yes...that is.....GOD DAMN UGLY hahaha!

It's fine, it's the ecosystem I signed up for. If that's the price I pay for working on exciting stuff, I'm happy to pay it ;)

PR #905, opened 2018-06-13
Support multiple UPnP devices

### LINKED ISSUE #787: Port mapping fails if the first UPnP device found is on a different subnet
### What is wrong?

The server grabs the first device it can find and runs with it. So if the UPnP device on the same `/24` address space is the second one listed, NAT port mapping fails.

### How can it be fixed

If the first device mapping fails, keep iterating through the devices until one succeeds (or you run out of devices).

I have a very dirty PR on my machine. I haven't pushed it yet because it's not critical path to get the alpha out the door.

carver:
### What was wrong?

If the first UPnP device you found was not working (say, because it was on a different subnet), then we just gave up.

Fixes #787 

### How was it fixed?

- If setting up NAT on the first device fails, attempt to set up on the others
- Extracted UPnP logic out of p2p.server

Bonus fix:
- bugfix when canceling an already canceled service (now it will only warn instead of raise an exception)

#### Cute Animal Picture

### DIFF
--- a/p2p/nat.py
+++ b/p2p/nat.py
@@ -13,6 +13,7 @@ from p2p.cancel_token import (
     CancelToken,
 )
 from p2p.exceptions import (
+    NoInternalAddressMatchesDevice,
     OperationCancelled,
 )
 import netifaces
@@ -27,11 +28,6 @@ UPNP_DISCOVER_TIMEOUT_SECONDS = 30
 class PortMapping(NamedTuple):
     internal: str  # of the form "192.2.3.4:56"
     external: str  # of the form "192.2.3.4:56"
-
-
-class NoInternalAddressMatchesDevice(Exception):
-    def __init__(self, *args, device_hostname=None, **kwargs):
-        self.device_hostname = device_hostname
 
 
 def find_internal_ip_on_device_network(upnp_dev: upnpclient.upnp.Device) -> str:
--- a/p2p/nat.py
+++ b/p2p/nat.py
@@ -5,7 +5,7 @@ from concurrent.futures import (
 import ipaddress
 from typing import (
     AsyncGenerator,
-    Tuple,
+    NamedTuple,
 )
 from urllib.parse import urlparse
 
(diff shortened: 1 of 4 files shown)


p2p/nat.py
```diff
@@ -0,0 +1,178 @@
...
+# UPnP discovery can take a long time, so use a loooong timeout here.
+UPNP_DISCOVER_TIMEOUT_SECONDS = 30
+
+
+class PortMapping(NamedTuple):
+    internal: str  # of the form "192.2.3.4:56"
+    external: str  # of the form "192.2.3.4:56"
+
+
+class NoInternalAddressMatchesDevice(Exception):
```
pipermerriam: Is there a reason this isn't in `p2p.exceptions`?

p2p/server.py
```diff
@@ -246,11 +142,11 @@ def _make_peer_pool(self, discovery: DiscoveryProtocol) -> PeerPool:
...
     async def _run(self) -> None:
         self.logger.info("Running server...")
-        upnp_dev = await self._discover_upnp_device()
-        external_ip = '0.0.0.0'
-        if upnp_dev is not None:
-            external_ip = upnp_dev.WANIPConn1.GetExternalIPAddress()['NewExternalIPAddress']
-            await self._add_nat_portmap(upnp_dev)
+        mapped_external_ip = await self.upnp_service.add_nat_portmap()
+        if mapped_external_ip is None:
+            external_ip = '0.0.0.0'
```
pipermerriam: Wondering if we can move away from this and instead do what geth does and just not show an IP address.

p2p/nat.py
carver: Everything in `p2p.exceptions` seems to be related to peers, and this doesn't really have anything to do with it. Maybe `nat` should be moved to `trinity`.

pipermerriam: I think it is safe to assume anyone who wants to setup an app which uses the DevP2P protocol will want this functionality so `p2p` seems more appropriate.

Issue #901, opened 2018-06-12
BadAckMessage: Auth ack msg too short: 0


carver:
### What is wrong?

```
    INFO  06-12 13:10:22        peer  Adding peer: LESPeer <Node(0x053d@79.98.29.154)>
   DEBUG  06-12 13:10:22        peer  Connecting to <Node(0x94c1@192.81.208.223)>...
   DEBUG  06-12 13:10:24  lightchain  fetched headers from 1 to 192
    INFO  06-12 13:10:24        peer  Got bad Ack during peer connection
   DEBUG  06-12 13:10:24        peer  Got bad Ack during peer connection
Traceback (most recent call last):
  File "/home/jcarver/code/py-evm/p2p/peer.py", line 654, in connect
    self.cancel_token))
  File "/home/jcarver/code/py-evm/p2p/service.py", line 32, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/jcarver/code/py-evm/p2p/service.py", line 51, in wait_first
    return await wait_with_token(*awaitables, token=token_chain, timeout=timeout)
  File "/home/jcarver/code/py-evm/p2p/cancel_token.py", line 119, in wait_with_token
    return done.pop().result()
  File "/home/jcarver/code/py-evm/p2p/peer.py", line 126, in handshake
    ) = await auth.handshake(remote, privkey, token)
  File "/home/jcarver/code/py-evm/p2p/auth.py", line 59, in handshake
    initiator, reader, writer, token)
  File "/home/jcarver/code/py-evm/p2p/auth.py", line 81, in _handshake
    ephemeral_pubkey, responder_nonce = initiator.decode_auth_ack_message(auth_ack)
  File "/home/jcarver/code/py-evm/p2p/auth.py", line 195, in decode_auth_ack_message
    raise BadAckMessage("Auth ack msg too short: {}".format(len(ciphertext)))
p2p.exceptions.BadAckMessage: Auth ack msg too short: 0

```

### How can it be fixed

:man_shrugging: 

#885 fixed a similar-sounding bug. I was running from 0fe47acb737dea65a which includes that merge.

@gsalgado probably has a better idea where to start

veox:
It is a Parity node with no matching capabilities:

```
   DEBUG  06-12 23:32:48        peer  Connecting to <Node(0x94c1@192.81.208.223)>...
   DEBUG  06-12 23:32:48  lightchain  fetched headers from 0 to 191
[...]

PR #907, opened 2018-06-13
p2p: Correctly handle peers that disconnect before sending the auth ack

gsalgado:
Closes: #901

### DIFF
--- a/p2p/auth.py
+++ b/p2p/auth.py
@@ -22,8 +22,9 @@ from p2p import kademlia
 from p2p.cancel_token import CancelToken, wait_with_token
 from p2p.constants import REPLY_TIMEOUT
 from p2p.exceptions import (
-    DecryptionError,
     BadAckMessage,
+    DecryptionError,
+    HandshakeFailure,
 )
 from p2p.utils import (
     sxor,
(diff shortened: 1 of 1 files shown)


p2p/auth.py
```diff
@@ -78,6 +79,11 @@
         token=token,
         timeout=REPLY_TIMEOUT)
 
+    if reader.at_eof():
+        # This is what happens when Parity nodes have blacklisted us
+        # (https://github.com/ethereum/py-evm/issues/901).
```
carver: I'm not a fan of linking out to github from a comment usually, but I may be alone on that. My logic is that putting the link can be used as a crutch to avoid fully describing what's going on. But most people won't follow the link, so they end up without as clear a picture of the situation.

gsalgado:
The first time we try to connect Parity will do the auth handshake and disconnect with a useless peer msg, but if we keep trying then it drops the connection as soon as it gets our node ID, which makes sense. Also, the first time we tried to connect we also detected that the peer was useless (since Parity sent us the P2P Hello message), so we should definitely use that info and not keep trying to connect -- instead of doing that only for peers that disconnect from us with a useless peer msg.

p2p/auth.py
gsalgado: Right, but I think the code here is straightforward enough to be understood even without the comment. The only reason I added the comment is so that, if necessary, one can understand why it had to be written in the first place

PR #909, opened 2018-06-13
Add ERC20 deployment/interaction benchmark

cburgdorf:
### What was wrong?

As specified in #592, our benchmark suite should test the deployment of an ERC20 Token.

### How was it fixed?

This adds a new benchmark that deploys an ERC20 token and simply calls the `totalSupply()` API to prove that everything works as expected.

Known issues:

- does not yet test further interaction with the contract
- ~~dumped solidity code in one large string~~

I hope that we can live with these known issues for now as I'm planning to clean them up at a later point in the future.

#### Cute Animal Picture


setup.py
```diff
@@ -46,12 +46,14 @@
...
         "hypothesis==3.44.26",
+        "py-solc==3.1.*",
         "pytest~=3.2",
         "pytest-asyncio==0.8.0",
         "pytest-cov==2.5.1",
         "pytest-xdist==1.18.1",
         "pytest-watch>=4.1.0,<5",
-        "termcolor==1.1.*"
+        "termcolor==1.1.*",
+        "web3>=4.1.0,<5.0.0",
```
carver: py-solc and web3 dependencies can go into the tox dependency list for py36-benchmark, to keep the general test requirements shorter.

.circleci/config.yml
```diff
@@ -26,7 +26,7 @@ common: &common
           - cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
     - run:
         name: install dependencies
-        command: pip install --user tox
+        command: ./.circleci/fetch_deps.sh
```
carver: For testing speed, maybe a separate dependency install that only triggers on the benchmark job?

```yml
     - run:
         name: install dependencies
         command: pip install --user tox
     - run:
         name: install benchmark dependencies
         command: if [ "$CIRCLE_JOB" == "py36-benchmark" ]; then ./.circleci/fetch_deps.sh; fi
```

setup.py
cburgdorf: Good point. However, moving them somewhere into the tox setup makes them very subtle though. It should be most obvious for someone new to the code base how to install these dependencies. So how about I introduce a new `deps['benchmark']` section that is not included in `dev` and then make sure to install that one only for the benchmark CI run?

.circleci/config.yml
cburgdorf: Yep, that looks good!

.circleci/fetch_deps.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# used by all CI jobs
+pip install --user tox
+
+# only used by the benchmark job
+if [ "$CIRCLE_JOB" == "py36-benchmark" ]; then
```
cburgdorf: @carver I've put the condition into that `fetch_deps` file because all jobs need tox and so things seem a bit more DRY and organized to me.

scripts/benchmark/contract_data/erc20.sol
```diff
@@ -0,0 +1,274 @@
+pragma solidity ^0.4.23;
```
cburgdorf: @carver I actually tried to import the openzeppelin repo as submodule and then directly use their example file but `solc ---allow-paths` was driving me bat shit crazy. Need to revisit. For now, this is basically a manual copy and paste act to resolve all imports and concat things into one file

tox.ini
```diff
@@ -72,3 +71,10 @@ commands=
     mypy --follow-imports=silent --ignore-missing-imports --check-untyped-defs --disallow-incomplete-defs -p p2p -p trinity
     # benchmark uses more restrictive type checking
     mypy --follow-imports=silent --ignore-missing-imports --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics scripts/benchmark
+
+
+[testenv:py36-benchmark]
+deps = .[evm-extras, benchmark]
```
cburgdorf: That's the only job using the `benchmark` dependencies

.circleci/fetch_deps.sh
carver: I was assuming you would remove the tox install from `fetch_deps` altogether so there would be duplication.

I slightly prefer the other approach. It keeps some circle-specific things out of the script, which could theoretically end up being moved/reused for other setup things. Even so, this is a tiny thing, and only an opinion, so please code it however you like. :)

.circleci/fetch_deps.sh
```diff
@@ -0,0 +1,18 @@
...
+  # to get the add-apt-repository command
+  sudo apt-get install software-properties-common
+
+  sudo add-apt-repository -y ppa:ethereum/ethereum
+  sudo add-apt-repository -y ppa:ethereum/ethereum-dev
+  sudo sed -i 's/cosmic/bionic/g' /etc/apt/sources.list.d/ethereum-ubuntu-ethereum-cosmic.list
+  sudo sed -i 's/cosmic/bionic/g' /etc/apt/sources.list.d/ethereum-ubuntu-ethereum-dev-cosmic.list
+  sudo apt-get -y update
+  sudo apt-get install -y --allow-unauthenticated solc
+fi
```
carver: *nitpick* I (and github) prefer trailing newlines :)

setup.py
```diff
@@ -51,12 +51,16 @@
         "pytest-cov==2.5.1",
         "pytest-xdist==1.18.1",
         "pytest-watch>=4.1.0,<5",
-        "termcolor==1.1.*"
```
carver: Was `termcolor` first added by the other benchmark PR?

tox.ini
```diff
@@ -72,3 +71,10 @@ commands=
...
+basepython=python3.6
+commands=
+    benchmark: {toxinidir}/scripts/benchmark/run.py
```
carver: *nitpick* trailing newline, please

scripts/benchmark/checks/deploy_erc20.py
```diff
@@ -0,0 +1,167 @@
...
+
+        self.num_blocks = num_blocks
+        self.num_tx = num_tx
+
+        compiled_sol = solc.compile_files([SOLIDITY_SRC_FILE])
+        self.contract_interface = compiled_sol['{}:SimpleToken'.format(SOLIDITY_SRC_FILE)]
+
+        # Stop Web3 from trying to call out to external APIs
+        TRANSACTION_DEFAULTS['gas'] = lambda web3, tx: 0
+        TRANSACTION_DEFAULTS['gasPrice'] = lambda web3, tx: 0
```
carver: Global monkeypatches creep me out. Also, we really shouldn't have to reach into `utils` from another package.

If we pass in `{'gas': 0, 'gasPrice': 0}` to the `buildTransaction()` calls, that should prevent web3 from trying to generate defaults. Does anywhere else fail when you remove these lines?

scripts/benchmark/checks/deploy_erc20.py
```diff
@@ -0,0 +1,167 @@
...
+        logging.debug('Applying Transaction {}'.format(tx))
+
+        block, receipt, computation = chain.apply_transaction(tx)
+        deployed_contract_address = computation.msg.storage_address
+
+        assert computation.is_success
+
+        # Interact with the deployed contract by calling the totalSupply() API
+        simple_token = self.w3.eth.contract(
+            address=Web3.toChecksumAddress(encode_hex(deployed_contract_address)),
```
carver: In the next release, web3py will accept `bytes` addresses, but it doesn't seem worth a wait (or rushing out the release), for a minor nicety.

setup.py
cburgdorf: Yep, it was

scripts/benchmark/checks/deploy_erc20.py
cburgdorf: Ah, will try that!

cburgdorf: That worked. I wasn't happy with monkeypatching either. Happy to see there's a better way to use the API to avoid that :+1:

.circleci/fetch_deps.sh
cburgdorf: >It keeps some circle-specific things out of the script, which could theoretically end up being moved/reused for other setup things

True that! Changed it.

cburgdorf:
Addressed all of @carver feedback. Since he's off for the rest of the week. Anyone else up for a review? @davesque ?

.circleci/install_benchmark_sys_deps.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# to get the add-apt-repository command
+sudo apt-get install software-properties-common
+
+sudo add-apt-repository -y ppa:ethereum/ethereum
+sudo add-apt-repository -y ppa:ethereum/ethereum-dev
+sudo sed -i 's/cosmic/bionic/g' /etc/apt/sources.list.d/ethereum-ubuntu-ethereum-cosmic.list
+sudo sed -i 's/cosmic/bionic/g' /etc/apt/sources.list.d/ethereum-ubuntu-ethereum-dev-cosmic.list
+sudo apt-get -y update
+sudo apt-get install -y --allow-unauthenticated solc
```
pipermerriam: This is almost certainly going to cause a future CI test failure.  Can we just pre-compile the binary and keep note of the compiler version?  Also, using different compiler versions will end up with different benchmarking results due to different resulting bytecode.

scripts/benchmark/checks/deploy_erc20.py
```diff
@@ -0,0 +1,162 @@
...
+
+class DeployErc20(BaseBenchmark):
+
+    def __init__(self, num_blocks: int = 100, num_tx: int = 2) -> None:
+        super().__init__()
+
+        self.num_blocks = num_blocks
+        self.num_tx = num_tx
+
+        compiled_sol = solc.compile_files([SOLIDITY_SRC_FILE])
```
pipermerriam: Just leaving the note here too.  I think we should package the pre-compiled binary rather than having compilation be part of this.

cburgdorf: Good point. I did this mostly for maintainability but I see your point. Makes CI faster and more reliable.

Would you be ok if I put in a script to generate the binary and save it next to the source? We could even make this a flag on the benchmark script (e.g. `--compile-contracts`).

.circleci/install_benchmark_sys_deps.sh
cburgdorf: True that. Cost me far too long to figure out the right CI setup but I'm not married to that approach. I see your point. See my second comment further down.

cburgdorf:
>Can you get an issue open to complete this benchmark with more than totalSupply calls?

I leave #592 intentionally open which covers our requirements in more detail. I check/strike through in this ticket as I land more PRs on this. Is this alright?

scripts/benchmark/checks/deploy_erc20.py
pipermerriam: Yes, but I'd like to fully drop the `py-solc` dependency.

cburgdorf:
@pipermerriam mind taking another look at the last WIP commit.

- benchmark uses pre-compiled contracts
- pre-compiles can be generated with `--compile-contracts`
- `py-solc` dependency is dropped
- there's an implicit `solc` system dependency for when one wants to run `--compile-contracts` (but that comes only into play when benchmarks are added/changed, so no runtime dependency)

PR #913, opened 2018-06-13
Light client: validate contract code with account

### LINKED ISSUE #849: LightPeerChain should confirm that contract code retrieved matches what was requested
Currently `LightPeerChain.get_contract_code()` simply return whatever is served to us by a peer, but we should confirm that it matches the code_hash for the account, as suggested in https://github.com/ethereum/py-evm/pull/841#discussion_r192200495

carver:
### What was wrong?

Fixes #849 

### How was it fixed?

- Modify API to take the address of the contract to check
- Request the account info to get the code_hash to validate against
- Drop a peer with subprotocol error if they give us a bad code hash

#### Cute Animal Picture

### DIFF
--- a/p2p/lightchain.py
+++ b/p2p/lightchain.py
@@ -144,14 +144,33 @@ class LightPeerChain(PeerPoolSubscriber, BaseService):
         return rlp.decode(rlp_account, sedes=Account)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
-    async def get_contract_code(self, block_hash: Hash32, key: bytes) -> bytes:
+    async def get_contract_code(self, block_hash: Hash32, address: Address) -> bytes:
+        """
+        :param block_hash: find code as of the block with block_hash
+        :param address: which contract to look up
+
+        :return: bytecode of the contract, ``b''`` if no code is set
+        """
         peer = cast(LESPeer, self.peer_pool.highest_td_peer)
         request_id = gen_request_id()
-        peer.sub_proto.send_get_contract_code(block_hash, key, request_id)
+        peer.sub_proto.send_get_contract_code(block_hash, keccak(address), request_id)
         reply = await self._wait_for_reply(request_id)
+
         if not reply['codes']:
-            return b''
-        return reply['codes'][0]
+            bytecode = b''
+        else:
+            bytecode = reply['codes'][0]
+
+        # validate bytecode against a proven account
+        account = await self.get_account(block_hash, address)
+
+        if account.code_hash == keccak(bytecode):
+            return bytecode
+        else:
+            # disconnect from this bad peer
+            await self.disconnect_peer(peer, DisconnectReason.subprotocol_error)
+            # try again with another peer
+            return await self.get_contract_code(block_hash, address)
 
     async def _get_block_header_by_hash(self, peer: LESPeer, block_hash: Hash32) -> BlockHeader:
         self.logger.debug("Fetching header %s from %s", encode_hex(block_hash), peer)
--- a/p2p/lightchain.py
+++ b/p2p/lightchain.py
@@ -38,7 +38,13 @@ from p2p.exceptions import (
 )
 from p2p.cancel_token import CancelToken
 from p2p import protocol
-from p2p.constants import REPLY_TIMEOUT
+from p2p.constants import (
+    COMPLETION_TIMEOUT,
+    REPLY_TIMEOUT,
+)
+from p2p.p2p_proto import (
+    DisconnectReason,
+)
 from p2p.peer import (
     LESPeer,
     PeerPool,
@@ -47,6 +53,7 @@ from p2p.peer import (
 from p2p.rlp import BlockBody
 from p2p.service import (
     BaseService,
+    service_timeout,
 )
 from p2p.utils import gen_request_id
 
@@ -144,6 +151,7 @@ class LightPeerChain(PeerPoolSubscriber, BaseService):
         return rlp.decode(rlp_account, sedes=Account)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
+    @service_timeout(COMPLETION_TIMEOUT)
     async def get_contract_code(self, block_hash: Hash32, address: Address) -> bytes:
         """
         :param block_hash: find code as of the block with block_hash
--- a/p2p/lightchain.py
+++ b/p2p/lightchain.py
@@ -149,7 +149,7 @@ class LightPeerChain(PeerPoolSubscriber, BaseService):
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
     async def get_account(self, block_hash: Hash32, address: Address) -> Account:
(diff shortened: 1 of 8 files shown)


p2p/lightchain.py
```diff
@@ -333,14 +333,33 @@ def callback(r):
...
+        # validate bytecode against a proven account
+        account = await self.get_account(block_hash, address)
+
+        if account.code_hash == keccak(bytecode):
+            return bytecode
+        else:
+            # disconnect from this bad peer
+            await self.disconnect_peer(peer, DisconnectReason.subprotocol_error)
+            # try again with another peer
+            return await self.get_contract_code(block_hash, address)
```
gsalgado: A recursive call like this could potentially send us in an infinite loop; I think it'd be preferable to raise something here and have callers deal with it whichever way they want.

p2p/lightchain.py
```diff
@@ -333,14 +333,33 @@ def callback(r):
         return rlp.decode(rlp_account, sedes=Account)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
-    async def get_contract_code(self, block_hash: Hash32, key: bytes) -> bytes:
+    async def get_contract_code(self, block_hash: Hash32, address: Address) -> bytes:
+        """
+        :param block_hash: find code as of the block with block_hash
+        :param address: which contract to look up
+
+        :return: bytecode of the contract, ``b''`` if no code is set
```
gsalgado: Currently (partly because of #916), this will also return b'' if the peer we pick below doesn't have the given block. Maybe mention that here, with a fixme that we should instead raise an exception if none of our peers have that block?

p2p/lightchain.py
carver: How about we just retry, but with a timeout? (a single timeout across all retries, of course)

Dealing with timeouts makes sense as a caller. Dealing with a bad faith response from some specific peer that the caller didn't pick feels awkward.

pipermerriam: I still think this should be handled by the caller.  This feels like an anti pattern.  This function should only be responsible for retrieval/validation,  orchestration is something that seems like it should be higher up.

p2p/lightchain.py
carver: Cool, thoughts on the new approach?

p2p/lightchain.py
```diff
@@ -315,14 +320,46 @@ def callback(r):
...
+        elif bytecode == b'':
+            # TODO disambiguate failure types here, and raise the appropriate exception
+            # An (incorrectly) empty bytecode might indicate a bad-acting peer, or it might not
+            raise NoEligiblePeers("Our best peer incorrectly responded with an empty code value")
+        else:
+            # a bad-acting peer sent an invalid non-empty bytecode
+            # disconnect from the peer
+            await self.disconnect_peer(peer, DisconnectReason.subprotocol_error)
+            # try again with another peer
+            return await self.get_contract_code(block_hash, address)
```
pipermerriam: I still feel reasonably strongly that this pattern is going to bite us.  Recursive function calls are generally in a similar bucket as `global` and `eval` and this doesn't strike me as one of the situations where it's necessary.  It doesn't seem that big a deal to add proper exception handling and orchestration to the call sites for this function.

cburgdorf: >Recursive function calls are generally in a similar bucket as global and eval

Yep, strong disagree here, too ;) I find recursion to be quite elegant tbh and it is in fact, the [go to approach in the functional programming camp](https://softwareengineering.stackexchange.com/questions/149167/are-functional-languages-better-at-recursion/149189#149189)

That said, recursion can be dangerous especially in languages that don't do [tail call optimizations](https://stackoverflow.com/questions/310974/what-is-tail-call-optimization)

Anyway, I'm not saying I'm in favor for one solution over the other here. Just wanted to throw in my opinion on recursion.

pipermerriam: I'll back down on putting `globals/eval` and recursiveness in the same bucket.  The point I was trying to make here was that the **current** implementation allows for an infinite amount of recursion and I very much dislike that.

To @carver point about forgetting to catch the exception.  Fair, and very valid.

At minimum, I'd like to go with the exception based approach for now, and document it's deficiencies and search for a 3rd option that we all like and that doesn't suffer from either of these problems.  I prefer something we might forget over possibly infinitely recursive async calls.

And one last argument against the current approach (which I think I already touched on).  It mixes two types of logic that I don't think belong together.  The *business* part of fetching and validating, and the orchestration part of managing re-issuing the request.

And one last suggestion for a temporary 3rd option.  don't make it recursive.  Just put the whole thing in a loop and only allow a max of X retries and then fail hard if they all fail.

carver:
I believe the only reason that the code got a :-1: from @pipermerriam was the concern about recursion. The latest code has no recursion (and no way to loop indefinitely). So I think it should be good to merge.

In order to test this, I had to fix an existing bug in the lightchain integration test since #869 -- The peer was trying to check the dao fork on a Ropsten sync.

carver:
I'm signing off for a week, so I'm happy if you two choose to do any of these things:

- merge as is
- fixup and merge
- wait for me to come back

carver:
@pipermerriam can you review again? (especially the retry pattern that was of concern)

p2p/lightchain.py
```diff
@@ -136,22 +149,90 @@ def callback(r: protocol._DecodedMsgType) -> None:
...
+        else:
+            bytecode = reply['codes'][0]
+
+        # validate bytecode against a proven account
+        if code_hash == keccak(bytecode):
+            return bytecode
+        elif bytecode == b'':
+            # TODO disambiguate failure types here, and raise the appropriate exception
+            # An (incorrectly) empty bytecode might indicate a bad-acting peer, or it might not
+            raise NoEligiblePeers("Our best peer incorrectly responded with an empty code value")
```
pipermerriam: I'm not sure if I'm splitting hairs but this exception seems wrong.  If the code came into existence at block 5, and our best peer is at block 10, but some of our other peers are at blocks 6 and 8, then we still have eligible peers, but there is something wrong with our *best* peers.

Mostly, my issue is with the name of the exception and whether we need some other exception to capture the concept of what has gone wrong here.

p2p/lightchain.py
```diff
@@ -136,22 +149,90 @@ def callback(r: protocol._DecodedMsgType) -> None:
...
+                peer,
+                encode_hex(address),
+                encode_hex(block_hash),
+            )) from exc
         return rlp.decode(rlp_account, sedes=Account)
 
     @alru_cache(maxsize=1024, cache_exceptions=False)
-    async def get_contract_code(self, block_hash: Hash32, key: bytes) -> bytes:
-        peer = cast(LESPeer, self.peer_pool.highest_td_peer)
+    @service_timeout(COMPLETION_TIMEOUT)
```
pipermerriam: Any reason this was not applied to other methods like `get_account`?

carver: Not any especially good ones. Added everywhere.

PR #915, opened 2018-06-14
Move header state into special MiningChain

### LINKED ISSUE #902: Trinity fails to switch to "regular" sync upon finishing state sync, then hangs indefinitely
### What is wrong?

Upon completing state sync (on Ropsten), Trinity attempts to switch to a "regular" rolling sync, but the expected API does not match the one present.

In the log, notice `^C` and the hour-long window in between the timestamps.

Additionally, on subsequent run, starting regular sync fails for the same reason.

``` python
    INFO  06-12 19:22:40       state  Scheduler queue is empty, sleeping a bit
    INFO  06-12 19:22:41       state  Finished state sync with root hash 0x469a557209df85602c0488f90865fa15fa486ec3338fd349f80c685955835421
    INFO  06-12 19:22:41        sync  Starting regular sync; current head: #3423335
   ERROR  06-12 19:22:41     service  Unexpected error in <p2p.sync.FullNodeSyncer object at 0x7fd4cdb38a20>, exiting
  File "/home/veox/src/py-evm/p2p/service.py", line 68, in run
    await self._run()
  File "/home/veox/src/py-evm/p2p/sync.py", line 66, in _run
    new_chain = type(self.chain)(self.base_db)
TypeError: __init__() missing 1 required positional argument: 'serializer'
    INFO  06-12 19:22:41      server  Closing server...
    INFO  06-12 19:22:41   discovery  stopping discovery
```

### Wholly optional animal pic

Source: [kumaran@peepeth](https://peepeth.com/Kumaran/peeps/QmXiw8HwbufWSG8iSABpeFEY6Dw5PbTVPhE3hrnWiooXxe)

carver:
Keeping a pending block and header only really matters for mining. Remove that quirky part of state. With header gone, all critical state is held in the chaindb. Chain objects in different processes
will no longer diverge.

Note that although the chain now doesn't maintain a concept of a pending header, the VM must still be initialized on a pending header before importing a block. That's because VMs must not change block number during any operation. (So they have to "start" on the pending block they are about to import, somewhat counter-intuitively.)

If incremental activity like applying transactions are required, then use the new MiningChain class or the new `Chain.build_block_with_transactions()`.

### What was wrong?

#902 pointed out some challenges from keeping header state in the `Chain` object.

### How was it fixed?

- Pushed all the header state into `MiningChain`
- Do not explicitly manage a pending header in `Chain`
- Drop `pending` support from RPC (supporting it has weird corner cases anyway, and is a source of pain for `geth`, so maybe best to just never support it from the start)
- New convenience method: `Chain.build_block_with_transactions()` to build prospective blocks for import (ignoring PoW)
- Fixes #902 by reusing the same chain instance instead of trying to make a copy.

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -109,7 +109,6 @@ class BaseChain(Configurable, ABC):
     chaindb = None  # type: BaseChainDB
     chaindb_class = None  # type: Type[BaseChainDB]
     vm_configuration = None  # type: Tuple[Tuple[int, Type[BaseVM]], ...]
-    header = None  # type: BlockHeader
     network_id = None  # type: int
 
     @abstractmethod
@@ -140,10 +139,6 @@ class BaseChain(Configurable, ABC):
     def from_genesis_header(cls,
                             base_db: BaseDB,
                             genesis_header: BlockHeader) -> 'BaseChain':
-        raise NotImplementedError("Chain classes must implement this method")
-
-    @abstractmethod
-    def get_chain_at_block_parent(self, block: BaseBlock) -> 'BaseChain':
         raise NotImplementedError("Chain classes must implement this method")
 
     #
@@ -234,19 +229,11 @@ class BaseChain(Configurable, ABC):
     # Execution API
     #
     @abstractmethod
-    def apply_transaction(self, transaction):
-        raise NotImplementedError("Chain classes must implement this method")
-
-    @abstractmethod
     def estimate_gas(self, transaction: BaseTransaction, at_header: BlockHeader=None) -> int:
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
     def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseBlock:
-        raise NotImplementedError("Chain classes must implement this method")
-
-    @abstractmethod
-    def mine_block(self, *args: Any, **kwargs: Any) -> BaseBlock:
         raise NotImplementedError("Chain classes must implement this method")
 
     #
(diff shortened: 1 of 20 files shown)


carver:
@cburgdorf I changed the chains to mining chains in the benchmark for simplicity. It could make sense to use the new `build_block_with_transactions` instead, but it didn't seem urgent. Any thoughts on the changes.

cburgdorf:
@carver that looks good to me. I think that this change will also have consequences for the "Understanding the mining process" guide as that will probably also need to use the new `MiningChain` now. https://github.com/ethereum/py-evm/blob/master/docs/guides/evm/understanding_the_mining_process.rst

I'm not super confident to give this the proper review it deserves. I think I understand the reasoning for this change but this may have implications that I don't foresee well enough so I'd be happy if @gsalgado could have another look at this.

evm/chains/base.py
```diff
@@ -310,6 +301,12 @@ def get_chaindb_class(cls) -> Type[BaseChainDB]:
     def get_vm_configuration(cls) -> Tuple[Tuple[int, Type['BaseVM']], ...]:
         return cls.vm_configuration
 
+    def default_header(self, header: BlockHeader=None) -> BlockHeader:
```
gsalgado: Having a method with this name that takes an argument feels odd to me. Even more so since it simply returns the argument passed to it when it's not None. Is this really needed?

tests/p2p/integration_test_helpers.py
```diff
@@ -62,7 +64,7 @@ class FakeAsyncMainnetChain(MainnetChain):
     coro_import_block = coro_import_block
 
 
-class FakeAsyncChain(Chain):
+class FakeAsyncChain(MiningChain):
```
gsalgado: Why do we need a MiningChain here?

gsalgado:
Oh, and if you include http://ix.io/1eYn here you can close #902 as well!

carver:
> Oh, and if you include http://ix.io/1eYn here you can close #902 as well!

:+1: @gsalgado done -- any other comments?

Issue #916, opened 2018-06-14
Consolidate APIs to select peers from which to request data


gsalgado:
We have several APIs for selecting a peer when we want to fetch data: `LightPeerChain.get_best_peer()`, `PeerPool.get_random_peer()`, `PeerPool.peers`, `StateDownloader.get_idle_peer()` and a couple places in `FastChainSyncer` that sort/filter `PeerPool.peers` based on some criteria. 

Apart from `StateDownloader.get_idle_peer()`, I think we should be able to have a single API on `PeerPool` that returns a peer known to have a certain block (defined by its number or TD). That would avoid duplication and ensure we always request data to peers known to have them.

PR #917, opened 2018-06-14
p2p: Consolidate PeerPool APIs to get peers from which to fetch data

gsalgado:
Closes: #916

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -26,6 +26,8 @@ from typing import (
 
 import sha3
 
+from cytoolz import groupby
+
 import rlp
 from rlp import sedes
 
@@ -61,6 +63,7 @@ from p2p.exceptions import (
     DecryptionError,
     HandshakeFailure,
     MalformedMessage,
+    NoConnectedPeers,
     NoEligibleNodes,
     NoMatchingPeerCapabilities,
     OperationCancelled,
@@ -145,6 +148,8 @@ class BasePeer(BaseService):
     listen_port = 30303
     # Will be set upon the successful completion of a P2P handshake.
     sub_proto: protocol.Protocol = None
+    head_td: int = None
+    head_hash: Hash32 = None
 
     def __init__(self,
                  remote: Node,
@@ -296,7 +301,7 @@ class BasePeer(BaseService):
             try:
                 cmd, msg = await self.read_msg()
             except (PeerConnectionLost, TimeoutError) as e:
-                self.logger.info(
+                self.logger.debug(
                     "%s stopped responding (%s), disconnecting", self.remote, repr(e))
                 return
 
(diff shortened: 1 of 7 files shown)


p2p/peer.py
```diff
@@ -734,18 +745,21 @@ def _peer_finished(self, peer: BaseService) -> None:
...
+        if not self.connected_nodes:
+            raise NoConnectedPeers()
+        peers_by_td = groupby(operator.attrgetter('head_td'), self._peers)
+        max_td = max(peers_by_td.keys())
+        return random.choice(peers_by_td[max_td])
+
+    def get_peers(self, min_td: int) -> List[BasePeer]:
+        if not self.connected_nodes:
+            raise NoConnectedPeers()
+        return [peer for peer in self._peers if peer.head_td >= min_td]
```
pipermerriam: Can we use a `tuple` instead.

p2p/shard_syncer.py
```diff
@@ -99,7 +99,7 @@ def propose(self) -> Collation:
         self.shard.add_collation(collation)
 
         # broadcast collation
-        for peer in self.peer_pool.peers:
+        for peer in self.peer_pool._peers:
```
pipermerriam: We shouldn't be accessing private properties.  Maybe we should implement `PeerPool.__iter__` so that we can do `for peer in peer_pool`

p2p/state.py
```diff
@@ -64,7 +64,8 @@ def register_peer(self, peer: BasePeer) -> None:
 
     @property
     def idle_peers(self) -> List[ETHPeer]:
-        peers = set([cast(ETHPeer, peer) for peer in self.peer_pool.peers])
+        # FIXME: Should probably use get_peers() and pass the TD of our head?
+        peers = set([cast(ETHPeer, peer) for peer in self.peer_pool._peers])
```
pipermerriam: Same here, lets not access private properties.

p2p/shard_syncer.py
gsalgado: So, the reason why I made that private is because I want to have, eventually, all PeerPool subscribers using one of the APIs that select peers based on a given criteria instead of going through all connected peers, but maybe there are valid use cases for accessing all connected peers, like here, where we want to broadcast to all... 

I'd prefer to keep the `.peers` property, though, instead of implementing `__iter__`, but can't really think of a good reason for going one way or the other, so happy to do `__iter__` if you feel strong about it

pipermerriam: Don't feel strongly.  `PeerPool.peers` is fine with me.

gsalgado: Ok, I've reverted it back to `.peers`; ready for another look

p2p/peer.py
```diff
@@ -735,17 +746,20 @@ def _peer_finished(self, peer: BaseService) -> None:
...
+
+    @property
+    def highest_td_peer(self) -> BasePeer:
+        if not self.connected_nodes:
+            raise NoConnectedPeers()
+        peers_by_td = groupby(operator.attrgetter('head_td'), self.peers)
+        max_td = max(peers_by_td.keys())
+        return random.choice(peers_by_td[max_td])
+
+    def get_peers(self, min_td: int) -> List[BasePeer]:
```
pipermerriam: Maybe rename to `get_peers_above_height` or `get_peers_by_td`

gsalgado: I'd considered that, but since the `min_td` argument is required it seemed to be clear enough already, and including that in the name would not make it any clearer. I'd prefer to leave it as is, if you're ok with that

PR #920, opened 2018-06-15
WIP attempt to resolve #904

### LINKED ISSUE #904: Limit number of peer connections from the same IP address
should fix: https://github.com/ethereum/py-evm/issues/880#issuecomment-395369943

### What is wrong?

https://www.cs.bu.edu/%7Egoldbe/projects/eclipseEth

We should add protection against eclipse attacks.

### How can it be fixed

Two things:

- The peer pool should be configurable for how many connections may be made to the same IP address.  I think a default of `2` should be acceptable here.
- The peer pool should require that a certain percentage of peer connections be outgoing (dial out).  Lets require `1/4` of our peer connections to be outbound (with a strict lower bound of `1` if the peer pool max peers is `>2` and no lower bound if `max_peers == 1`).

williambannas:
### What was wrong?
Peer nodes could perform and eclipse attack 

### How was it fixed?
limiting the number of incoming connections to 75% of total connections
no more then 2 nodes with the same IP are allowed to be connected

I think there are ways that this could have been done better. I thought that this point would be a nice time to get some feedback. I am still a bit confused on how to create tests for this, but I used ipdb.set_trace() to try to verify that things were working correctly (maybe there is a better way to do this as well). Any and all criticism is greatly appreciated. Thank you for your time.

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -1,11 +1,9 @@
 import asyncio
-import collections
 import contextlib
 import logging
 import operator
 import random
 import struct
-import time
 from abc import (
     ABC,
     abstractmethod
@@ -15,10 +13,8 @@ from typing import (
     Any,
     cast,
     Dict,
-    Generator,
     Iterator,
     List,
-    Sequence,
     TYPE_CHECKING,
     Tuple,
     Type,
@@ -38,7 +34,6 @@ from cryptography.hazmat.primitives.constant_time import bytes_eq
 from eth_utils import (
     decode_hex,
     encode_hex,
-    to_tuple,
 )
 
 from eth_typing import BlockNumber, Hash32
@@ -55,7 +50,6 @@ from evm.rlp.headers import BlockHeader
 
 from p2p import auth
 from p2p import ecies
-from p2p.discovery import DiscoveryProtocol
 from p2p.kademlia import Address, Node
 from p2p import protocol
 from p2p.exceptions import (
@@ -64,7 +58,6 @@ from p2p.exceptions import (
     HandshakeFailure,
     MalformedMessage,
     NoConnectedPeers,
-    NoEligibleNodes,
     NoMatchingPeerCapabilities,
     OperationCancelled,
     PeerConnectionLost,
@@ -567,18 +560,13 @@ class PeerPool(BaseService):
     PeerPool maintains connections to up-to max_peers on a given network.
     """
     logger = logging.getLogger("p2p.peer.PeerPool")
-    _connect_loop_sleep = 2
     _report_interval = 60
-    _discovery_lookup_running = asyncio.Lock()
-    _discovery_last_lookup: float = 0
-    _discovery_lookup_interval: int = 30
 
     def __init__(self,
                  peer_class: Type[BasePeer],
                  headerdb: 'BaseAsyncHeaderDB',
                  network_id: int,
                  privkey: datatypes.PrivateKey,
-                 discovery: DiscoveryProtocol,
                  max_peers: int = DEFAULT_MAX_PEERS
                  ) -> None:
         super().__init__()
@@ -586,7 +574,6 @@ class PeerPool(BaseService):
         self.headerdb = headerdb
         self.network_id = network_id
         self.privkey = privkey
-        self.discovery = discovery
         self.max_peers = max_peers
         self.connected_nodes: Dict[Node, BasePeer] = {}
         self._subscribers: List[PeerPoolSubscriber] = []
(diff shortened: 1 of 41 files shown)


p2p/peer.py
```diff
@@ -598,8 +598,20 @@ def __len__(self):
     def is_full(self) -> bool:
         return len(self) >= self.max_peers
 
+    def is_valid_connection_candidate(self, candidate: Node) -> bool:
+        max_matching_ip = 2
+        matching_ip = 0
+        for peer in self.connected_nodes.values():
```
pipermerriam: We tend to aim for near zero mutability in our code which generally means avoiding things like `+=` or really writing to a variable more than once.  Also, avoiding things like appending to lists and mutating dictionaries.  Here's some untested code that should accomplish what you're going for here.

```python
import operator
from cytoolz import groupby

def is_valid_connection_candidate(self, candidate: Node) -> bool:
    if candidate == self.this_node:
        return False

    nodes_by_ip = groupby(operator.attrgetter('address.ip'), self.connected_nodes.values())
    matching_ip_nodes = nodes_by_ip.get(candidate.address.ip, [])
    
    return len(matching_ip_nodes) < 2
```

p2p/server.py
```diff
@@ -355,8 +355,21 @@ def _make_peer_pool(self, discovery: DiscoveryProtocol) -> PeerPool:
             inbound=True,
         )
 
+        total_peers_inbound = 0.0
+        total_peers = 0.0
+
+        for current_peers in self.peer_pool.connected_nodes.values():
```
pipermerriam: Same here for avoiding mutability.  Here's some untested code that will do this a bit cleaner with less mutation.  You'll notice the approach is very similar.

```python
import operator
from cytoolz import groupby

total_peer_count = len(self.peer_pool)
inbound_peer_count = len([peer for peer in self.peer_pool.peers if peer.inbound])

if total_peer_count > 1 and inbound_peer_count // total_peers > 0.75:
    ...
```

p2p/server.py
```diff
@@ -355,8 +355,21 @@ def _make_peer_pool(self, discovery: DiscoveryProtocol) -> PeerPool:
...
+            total_peers += 1
+            if current_peers.inbound:
+                total_peers_inbound += 1
+        # make sure to have atleast 1/4 outbound connections
+
         if self.peer_pool.is_full:
```
pipermerriam: This check should be moved above the inbound/outbound peer connection logic.

p2p/server.py
```diff
@@ -257,9 +257,18 @@ def _make_peer_pool(self, discovery: DiscoveryProtocol) -> PeerPool:
...
         else:
-            # We use self.wait() here as a workaround for
-            # https://github.com/ethereum/py-evm/issues/670.
-            await self.wait(self.do_handshake(peer))
+            total_peers = len(self.peer_pool.connected_nodes)
+            inbound_peers = len(
+                [peer for peer in self.peer_pool.connected_nodes.values() if peer.inbound])
+            if total_peers > 1 and inbound_peers / total_peers > 0.75:
+                # make sure to have at least 1/4 outbound connections
+                peer.disconnect(DisconnectReason.useless_peer)
```
cburgdorf: Maybe double check what `DisconnectReason` geth and parity use for that case. `useless_peer` doesn't strike me as right but the protocol doesn't specify something that matches really well. Maybe `too_many_peers` would make more sense?

tests/p2p/test_server.py
```diff
@@ -44,8 +44,15 @@ def get_open_port():
 INITIATOR_REMOTE = Node(INITIATOR_PUBKEY, INITIATOR_ADDRESS)
 
 
-class MockPeerPool:
+class MockPeerPool():
     is_full = False
+    connected_nodes = {}
+
+    def is_valid_connection_candidate(self, node):
```
cburgdorf: @williambannas can you look into adding tests to cover the functionality of the PR

p2p/peer.py
```diff
@@ -598,8 +598,17 @@ def __len__(self):
...
         return len(self) >= self.max_peers
 
+    def is_valid_connection_candidate(self, candidate: Node) -> bool:
+        # connect to no more then 2 nodes with the same IP
+        nodes_by_ip = groupby(
+            operator.attrgetter('remote.address.ip'), self.connected_nodes.values())
+        matching_ip_nodes = nodes_by_ip.get(candidate.address.ip, [])
+        return len(matching_ip_nodes) <= 2
+
     def get_nodes_to_connect(self) -> Generator[Node, None, None]:
```
cburgdorf: @williambannas not really related to this PR but as you already touch the bulk of this function you could change the return type to `Iterable[Node]` which seems more idiomatic for this use case

pipermerriam:
Replaced by https://github.com/ethereum/py-evm/pull/935

thanks for getting the majority of this done @williambannas

PR #923, opened 2018-06-18
Add ewasm support

lrettig:
Tie together the software packages (py-evm, juno, pywasm) and test
detecting WASM code and passing data and receiving data. One
stEWASMtests test (`useGas`) is passing so data is flowing in both
directions.

For now the nastiness lives mostly in another package called [Juno](https://github.com/lrettig/juno) (which is the Latin name of Hera, which is the [EVMC "glue code" in C++](https://github.com/ewasm/hera)). One of the things @poemm and I have been discussing is whether this "glue" code belongs in py-evm or not.

@poemm's pywebassembly lives at https://github.com/poemm/pywebassembly.

#### Cute Animal Picture


setup.py
```diff
@@ -12,6 +12,7 @@
         "eth-bloom>=1.0.0,<2.0.0",
         "eth-utils>=1.0.1,<2.0.0",
         "eth-typing>=1.0.0,<2.0.0",
+        "juno>=0.1.0",
```
cburgdorf: I'm confused, https://pypi.org/project/juno/ seems to be something different, no? Where is that dependency coming from?

cburgdorf: Ah, ok I see it is here: https://github.com/lrettig/juno But then, how does that fit in with that `juno` package on pypi that seems to be occupied by something else, no?

lrettig: It doesn't. Two different things, no connection. The name can be changed :) I'm just running this locally, no pypi.

pipermerriam:
This seems like a good place to dump some brief thoughts.

1. The code quality of the python ewasm interpreter is pretty poor.  Acceptable for a POC, will require lots of work for much beyond that.
2. In this model, the `computation` object no longer contains accurate information about what happened during computation.  This is going to be problematic in many places and is something we will need to explore fixing.  Losing this information is not something I'm inclined to accept lightly as it is very valuable information.
3. Very cool to see how easy this was to glue together, though I expect (as stated in my previous two points) that real integration may be significantly more complex.

lrettig:
> In this model, the computation object no longer contains accurate information about what happened during computation. This is going to be problematic in many places and is something we will need to explore fixing. Losing this information is not something I'm inclined to accept lightly as it is very valuable information.

Completely agree. I wouldn't even call this hacky thing a "model." Agree this is important to maintain if possible.

poemm:
Dear @pipermerriam ,


Thank you for your message. My responses follow.

1. I sincerely apologize for not explaining the motivations behind the way in which PyWebAssembly was written. PyWebAssembly follows the Wasm Spec document closely, referencing where each function was defined. This is useful for the following reasons.
   i) The Spec document can be used as a user's manual.
   ii) PyWebAssembly can be audited alongside the Spec document.
   iii) When Wasm 1.1 is released, PyWebAssembly can be updated alongside the Spec document.
   iv) PyWebAssembly does not introduce invariants or design decisions that are not in the Spec document. There are many subtleties in the 150 page Spec document, and invariants may be difficult to maintain, as I have learned.
   v) Implementing the Spec document has allowed me to find errors and submit fixes to the Spec document, and I have more fixes coming.

PyWebAssembly is also structured for my personal economy-of-thought as it is being developed and studied for errors. I do not know any better than to follow the Spec document, but I have a lot to learn, and I am eager to improve myself.

In any case, PyWebAssembly was brought to your attention because it is 100% finished for your purposes, modulo tracing which is to-be-done, and tests which are not applicable to your application. This is not to speak of aesthetics.

2. Tracing Wasm execution is possible at the granularity of opcodes. When testing, I step through execution of each opcode, printing the operand stack and other info. I am open to your suggestions on how to implement tracing in a way which meets your parameters and specifications.

3. Lane is skilled at quickly gluing code together. Full integration may be complex because Wasm has different structure and properties than EVM, and it may be difficult to honor the invariants and designs which you have specified for EVM. Because of this complexity and resulting difficulty, I am willing to offer assistance in your project, although I do not wish to hinder your progress, nor do I wish to create any feelings of obligation.

I wish you only the best success.


Sincerely,
@poemm

Issue #882, opened 2018-06-07
Trinity throws MalformedMessage


pipermerriam:
### What is wrong?

This is escaping unhandled during runs of trinity.

```python
   ERROR  06-07 03:32:06      server  Unexpected error handling handshake
Traceback (most recent call last):
  File "/home/ubuntu/py-evm/p2p/p2p_proto.py", line 59, in decode
    raw_decoded = cast(Dict[str, int], super(Disconnect, self).decode(data))
  File "/home/ubuntu/py-evm/p2p/protocol.py", line 95, in decode
    return self.decode_payload(data[1:])
  File "/home/ubuntu/py-evm/p2p/protocol.py", line 82, in decode_payload
    data = rlp.decode(rlp_data, sedes=decoder)
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/rlp/codec.py", line 220, in decode
    obj = sedes.deserialize(item, **kwargs)
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/eth_utils/functional.py", line 22, in inner
    return callback(fn(*args, **kwargs))
  File "/home/ubuntu/py-evm/venv/lib/python3.6/site-packages/rlp/sedes/lists.py", line 84, in deserialize
    raise ListDeserializationError('List has wrong length', serial)
rlp.exceptions.ListDeserializationError: List has wrong length

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/py-evm/p2p/server.py", line 287, in receive_handshake
    await self._receive_handshake(reader, writer)
  File "/home/ubuntu/py-evm/p2p/server.py", line 363, in _receive_handshake
    await self.wait(self.do_handshake(peer))
  File "/home/ubuntu/py-evm/p2p/service.py", line 32, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/ubuntu/py-evm/p2p/service.py", line 51, in wait_first
    return await wait_with_token(*awaitables, token=token_chain, timeout=timeout)
  File "/home/ubuntu/py-evm/p2p/cancel_token.py", line 119, in wait_with_token
    return done.pop().result()
  File "/home/ubuntu/py-evm/p2p/server.py", line 366, in do_handshake
    await peer.do_p2p_handshake(),
  File "/home/ubuntu/py-evm/p2p/peer.py", line 242, in do_p2p_handshake
    cmd, msg = await self.read_msg()
[...]

PR #924, opened 2018-06-19
Catch and log exception about malformed msg

cburgdorf:
### What was wrong?

As described in #882 there's a `MalformedMessage` error escaping.

### How was it fixed?

Caught and logged exception. I'm not entirely sure if that is the right level to handle this but it's inspired by how we treat the exact same error in [`peer`](https://github.com/ethereum/py-evm/blob/7b660dc30fa6fd62df97e38c5d43dcf29d102437/p2p/peer.py#L679)

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -122,6 +122,8 @@ async def handshake(remote: Node,
          ) = await auth.handshake(remote, privkey, token)
     except (ConnectionRefusedError, OSError) as e:
         raise UnreachablePeer(e)
+    except MalformedMessage as e:
+        raise HandshakeFailure(e)
     peer = peer_class(
         remote=remote, privkey=privkey, reader=reader, writer=writer,
         aes_secret=aes_secret, mac_secret=mac_secret, egress_mac=egress_mac,
@@ -638,7 +640,6 @@ class PeerPool(BaseService):
             PeerConnectionLost,
             TimeoutError,
             UnreachablePeer,
-            MalformedMessage,
         )
         try:
             self.logger.debug("Connecting to %s...", remote)
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -121,9 +121,9 @@ async def handshake(remote: Node,
          writer
          ) = await auth.handshake(remote, privkey, token)
     except (ConnectionRefusedError, OSError) as e:
-        raise UnreachablePeer(e)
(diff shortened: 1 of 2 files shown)


gsalgado:
I think it'd be better to catch that in `handshake()` (in peer.py) and `Server.do_handshake()` and have them raise a `HandshakeFailure` instead. Then we can also get rid of `MalformedMessage` from `PeerPool.connect()`'s `expected_exceptions`

cburgdorf:
@gsalgado Like this? https://github.com/ethereum/py-evm/pull/924/commits/c760d31c28b9537b0f15206224f2265daaa63ba4

cburgdorf:
@carver 

>Maybe raise HandshakeFailure() from e

Fancy!

I squash commits before merging as usual.

p2p/server.py
```diff
@@ -267,10 +268,16 @@ def _make_peer_pool(self) -> PeerPool:
         else:
             # We use self.wait() here as a workaround for
             # https://github.com/ethereum/py-evm/issues/670.
-            await self.wait(self.do_handshake(peer))
+            try:
+                await self.wait(self.do_handshake(peer))
+            except MalformedMessage as e:
+                self.logger.debug("Could not complete handshake with %r: %s", peer.remote, repr(e))
```
gsalgado: Now we should get rid of this, right?

cburgdorf: Right 🙈

Should be good to merge now?

p2p/peer.py
```diff
@@ -121,7 +121,9 @@
          writer
          ) = await auth.handshake(remote, privkey, token)
     except (ConnectionRefusedError, OSError) as e:
-        raise UnreachablePeer(e)
+        raise UnreachablePeer() from e
+    except MalformedMessage as e:
```
gsalgado: It turns out this is the wrong place to catch this exception; it is raised by `Peer.do_p2p_handshake()` and not `auth.handshake()`

carver:
:/ merged with three commits called `WIP` :(

cburgdorf:
Nope, they were squashed before the merge.

carver:
Ah, squashed via github's squash UI I guess, which seems to mean the commits still show up in the commit list of the PR. :+1:

PR #925, opened 2018-06-19
Docs: Move release notes into their own section

cburgdorf:
### What was wrong?

Release notes right beneath the quickstart doesn't feel right to me.

### How was it fixed?

image

Moved into their own section. Notice that this is prepared to be split into Py-EVM/Trinity. It's just that we don't have release notes for Py-EVM yet.

On a related note: How would everyone feel if we adopt a commit message convention like the one from the Angular project and then have release notes auto generated.

Here's how the generated changelog for Angular looks like: https://github.com/angular/angular/blob/master/CHANGELOG.md

And this is the convention they are following:
https://github.com/angular/angular/blob/master/CONTRIBUTING.md#commit

I adopted that convention for other projects before and it works well. We can use something like [clog](https://github.com/clog-tool/clog-cli) (shameless plug: written by me and friends) to generate the changelog. It's written in Rust and we could simply put mac/linux binaries in a tools directory to have zero external dependencies. The binary is rather small.

Thoughts?

#### Cute Animal Picture


pipermerriam:
I'm very hesitant to require a specific commit format.  I believe this will add a barrier to entry to both the research teams as well as new contributors.  Manually generating the changelog is not currently that cumbersome, though I admit that since it is done manually, it is likely to be missed on occasion.

cburgdorf:
>I believe this will add a barrier to entry to both the research teams as well as new contributors

Yes, that's pretty much the only downside to it. To fully benefit from a generated changelog, people need to craft nice semantic commits that follow a convention, hence they need to become more proficient in git ("can you make that two commits, one with x and one with y").

On the other hand, I think that a high quality changelog is very valuable and I can imagine it will become more and more important for us in the future. Anyway, let's stick to the manual changelog for now and revisit in the future :+1:

PR #927, opened 2018-06-19
Add type hints for entire Trinity code base and enable stricter mypy checks

cburgdorf:
### What was wrong?

- The Trinity code base mixed typed defs with untyped defs and the
- `mypy` didn't protect us from further untyped defs landing in the code base
- `mypy` didn't protect us from implicit `Any` in generics (e.g. `Queue` which implicitly becomes `Queue[Any]`)
- there were lots of places that used `# type: ignore` effectively silencing `mypy`

### How was it fixed?

- added all missing type annotations for Trinity
- enabled  `--disallow-untyped-defs` and `--disallow-untyped-defs` for Trinity to protect the code base from further untyped defs to land
- removed almost all `# type: ignore` comments from Trinity

### Known issues

- There are some places typed as `Any` that could be further improved
- There's one `# type: ignore` because `EmptyService` isn't a proper substitute for `IPCServer` (no common base clase, no matching `run(...)` signature. I didn't want to address this in this PR though.

#### Cute Animal Picture

### DIFF
--- a/evm/vm/base.py
+++ b/evm/vm/base.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 from abc import (
     ABCMeta,
-    abstractmethod
+    abstractmethod,
 )
 import contextlib
 import functools
@@ -69,6 +69,11 @@ class BaseVM(Configurable, metaclass=ABCMeta):
     fork = None  # type: str
     chaindb = None  # type: BaseChainDB
     _state_class = None  # type: Type[BaseState]
+
+    @property
+    @abstractmethod
+    def state(self):
+        raise NotImplementedError("VM classes must implement this property")
 
     @abstractmethod
     def __init__(self, header, chaindb):
--- a/trinity/utils/xdg.py
+++ b/trinity/utils/xdg.py
@@ -11,28 +11,28 @@ from .filesystem import (
 )
 
 
-def get_home():
+def get_home() -> str:
     try:
         return os.environ['HOME']
     except KeyError:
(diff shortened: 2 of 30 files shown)


trinity/utils/xdg.py
```diff
@@ -11,28 +11,28 @@
...
-def get_xdg_config_home():
+def get_xdg_config_home() -> str:
     try:
         return os.environ['XDG_CONFIG_HOME']
     except KeyError:
         return os.path.join(get_home(), '.config')
 
 
-def get_xdg_data_home():
+def get_xdg_data_home() -> str:
```
pipermerriam: not for this PR but we should convert these to use `pathlib.Path` objects.

cburgdorf:
@pipermerriam mind giving that another go because it changed *a lot* since your last review.

evm/vm/base.py
```diff
@@ -70,6 +71,10 @@ class BaseVM(Configurable, metaclass=ABCMeta):
     chaindb = None  # type: BaseChainDB
     _state_class = None  # type: Type[BaseState]
 
+    @abstractproperty
```
pipermerriam: This has been deprecated since `python3.3`

https://docs.python.org/3.5/library/abc.html#abc.abstractproperty

appropriate way to do this is just

```python
@property
@abstractmethod
def the_method(self):
    ...
```

cburgdorf: Ah! Fixed it. Will merge once CI goes green :fire:

PR #928, opened 2018-06-19
Use Path.exists() instead of os.path.exists(str)

cburgdorf:
### What was wrong?

Path handling used old fashioned `os.path.exists(str)` style even though we are dealing with a `pathlib.Path`

### How was it fixed?

This is just a tiny refactoring making use of `Path.exists()` rather than `os.path.exists(str)`. Stumbled over this while working on #927 but I want keep #927 free from any changes that aren't type hints so that its less pain to review.

#### Cute Animal Picture

### DIFF
--- a/trinity/utils/ipc.py
+++ b/trinity/utils/ipc.py
@@ -1,14 +1,15 @@
 import os
+import pathlib
 import signal
 import time
 from multiprocessing import Process
 from logging import Logger
 
 
-def wait_for_ipc(ipc_path: str, timeout: int=1) -> None:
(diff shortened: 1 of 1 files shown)


pipermerriam:
Feel free to assign pull requests to people for review.  With there being multiple of us I think it's good practice to be more explicit about who is handling what as well as explicitly requesting review.  (just in case this has not previously been communicated to you).

cburgdorf:
@pipermerriam right..I usually wait until someone picks it up and then usually that someone is you ;) I'll start proactively requesting reviews from now and then balance that a bit between us.

PR #934, opened 2018-06-20
p2p: Move peer-discovery out of PeerPool

gsalgado:
We now have a separate service (DiscoveryService) that performs
discovery, when needed, and adds the peers to its PeerPool

Motivated by the discussion in #886, but also because discovery
shouldn't really be the responsibility of PeerPool

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -6,13 +6,17 @@ listening nodes.
 More information at https://github.com/ethereum/devp2p/blob/master/rlpx.md#node-discovery
 """
 import asyncio
+import collections
 import logging
+import random
 import time
 from typing import (
     Any,
     Callable,
-    Generator,
+    Dict,
+    Iterator,
     List,
+    Sequence,
     Tuple
 )
 
@@ -22,6 +26,7 @@ from eth_utils import (
     text_if_str,
     to_bytes,
     to_list,
+    to_tuple,
 )
 
 from eth_utils import (
@@ -35,8 +40,10 @@ from eth_keys import datatypes
 from eth_hash.auto import keccak
 
 from p2p.cancel_token import CancelToken
-from p2p.exceptions import OperationCancelled
+from p2p.exceptions import NoEligibleNodes, OperationCancelled
 from p2p import kademlia
+from p2p.peer import PeerPool
+from p2p.service import BaseService
 
 # UDP packet constants.
 MAC_SIZE = 256 // 8  # 32
@@ -93,7 +100,13 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
     async def lookup_random(self, cancel_token: CancelToken) -> List[kademlia.Node]:
         return await self.kademlia.lookup_random(self.cancel_token.chain(cancel_token))
 
-    def get_random_nodes(self, count: int) -> Generator[kademlia.Node, None, None]:
+    def get_random_bootnode(self) -> Iterator[kademlia.Node]:
+        if self.bootstrap_nodes:
+            yield random.choice(self.bootstrap_nodes)
+        else:
+            self.logger.warning('No bootnodes available')
+
+    def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
         return self.kademlia.routing.get_random_nodes(count)
 
     @property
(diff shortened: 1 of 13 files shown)


p2p/discovery.py
```diff
@@ -93,7 +100,13 @@ def __init__(self,
     async def lookup_random(self, cancel_token: CancelToken) -> List[kademlia.Node]:
         return await self.kademlia.lookup_random(self.cancel_token.chain(cancel_token))
 
-    def get_random_nodes(self, count: int) -> Generator[kademlia.Node, None, None]:
+    def get_random_bootnode(self) -> Generator[kademlia.Node, None, None]:
```
cburgdorf: `def get_random_bootnode(self) -> Iterable[kademlia.Node]` seems like a better fit for the signature

p2p/discovery.py
```diff
@@ -93,7 +100,13 @@ def __init__(self,
...
+        if self.bootstrap_nodes:
+            yield random.choice(self.bootstrap_nodes)
+        else:
+            self.logger.warning('No bootnodes available')
+
+    def get_nodes_to_connect(self, count: int) -> Generator[kademlia.Node, None, None]:
```
cburgdorf: Same here, can we make that `Iterable[kademlia.Node]`

p2p/discovery.py
```diff
@@ -234,6 +247,131 @@ def send_neighbours(self, node: kademlia.Node, neighbours: List[kademlia.Node])
...
+        if preferred_nodes is not None:
+            self.preferred_nodes = preferred_nodes
+        else:
+            self.logger.debug('Operating with no preferred nodes')
+            self.preferred_nodes = tuple()
+
+        self._preferred_node_tracker = collections.defaultdict(lambda: 0)
+
+    @to_tuple
+    def _get_eligible_preferred_nodes(self) -> Generator[kademlia.Node, None, None]:
```
cburgdorf: Same here, `Iterable[kademlia.Node]`

p2p/discovery.py
```diff
@@ -234,6 +247,131 @@ def send_neighbours(self, node: kademlia.Node, neighbours: List[kademlia.Node])
...
+        """
+        Return a random node from the preferred list.
+        """
+        eligible_nodes = self._get_eligible_preferred_nodes()
+        if not eligible_nodes:
+            raise NoEligibleNodes("No eligible preferred nodes available")
+        node = random.choice(eligible_nodes)
+        return node
+
+    def get_random_bootnode(self) -> Generator[kademlia.Node, None, None]:
```
cburgdorf: and here `Iterable[kademlia.Node]`

p2p/discovery.py
```diff
@@ -234,6 +247,131 @@ def send_neighbours(self, node: kademlia.Node, neighbours: List[kademlia.Node])
...
+        Return a single node to bootstrap, preferring nodes from the preferred list.
+        """
+        try:
+            node = self._get_random_preferred_node()
+            self._preferred_node_tracker[node] = time.time()
+            yield node
+        except NoEligibleNodes:
+            yield from super().get_random_bootnode()
+
+    def get_nodes_to_connect(self, count: int) -> Generator[kademlia.Node, None, None]:
```
cburgdorf: and here `Iterable[kademlia.Node]`

p2p/discovery.py
```diff
@@ -234,9 +247,134 @@ def send_neighbours(self, node: kademlia.Node, neighbours: List[kademlia.Node])
...
+                 privkey: datatypes.PrivateKey,
+                 address: kademlia.Address,
+                 bootstrap_nodes: Tuple[kademlia.Node, ...],
+                 preferred_nodes: Sequence[kademlia.Node] = None) -> None:
+        super().__init__(privkey, address, bootstrap_nodes)
+
+        if preferred_nodes is not None:
+            self.preferred_nodes = preferred_nodes
+        else:
+            self.logger.debug('Operating with no preferred nodes')
```
pipermerriam: Might be worth going ahead and doing a debug log in the other conditional branch to log what preferred nodes we are using.

p2p/discovery.py
```diff
@@ -234,9 +247,134 @@ def send_neighbours(self, node: kademlia.Node, neighbours: List[kademlia.Node])
...
+    async def _run(self) -> None:
+        connect_loop_sleep = 2
+        asyncio.ensure_future(self.proto.bootstrap())
+        while True:
+            await self.maybe_connect_to_more_peers()
+            await self.wait(asyncio.sleep(connect_loop_sleep))
+
+    async def maybe_connect_to_more_peers(self) -> None:
+        """Connect to more peers if we're not yet maxed out to max_peers"""
+        num_connected_nodes = len(self.peer_pool)
```
pipermerriam: Can you move this inside the `if` statement (or maybe just inline with the logging statement)

p2p/discovery.py
gsalgado: I'll move it outside of the if block, always logging self.preferred_nodes, and change it into a INFO as this is only going to show up once at startup

PR #939, opened 2018-06-21
Make register_peer of PeerPoolSubscriber non-abstract

cburgdorf:
### What was wrong?

Three out of five `PeerPoolSubscriber` do not actually need to act upon the registration of other peers (counting the TxPool as well that still sits in a PR).

Instead, the majority of `PeerPoolSubscriber` provides the following implementation for `register_peer`

```python
    def register_peer(self, peer: BasePeer) -> None:
        pass
```
This strikes me as a hint that `register_peer` isn't a great candidate for being an abstract method on `PeerPoolSubscriber`.

### How was it fixed?

I considered moving it to a whole different abstract class but it strikes me as most natural to keep it on `PeerPoolSubscriber` but just don't dictate to overwrite it.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -536,9 +536,16 @@ class ETHPeer(BasePeer):
 class PeerPoolSubscriber(ABC):
     _msg_queue: 'asyncio.Queue[PEER_MSG_TYPE]' = None
 
-    @abstractmethod
     def register_peer(self, peer: BasePeer) -> None:
-        raise NotImplementedError("Must be implemented by subclasses")
+        """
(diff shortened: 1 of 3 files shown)


pipermerriam:
I see now that we don't have an `unregister_peer` callback.  Do you think it makes sense to add this.  The use case I have in mind is `TxnPool` using it to cleanup peer specific bloom filters.

Not for this PR, but wanted to toss the idea out to see if it sticks.

cburgdorf:
@pipermerriam was wondering the same but thought we should add it when we need it. The bloom filter use case seems legit!

PR #947, opened 2018-06-22
Add type hints for p2p package and enable stricter checks

cburgdorf:
### What was wrong?

P2P was missing many type hints and didn't have the `--disallow-untyped-defs` check yet which means that new code could potentially land without type hints (making the fight for 100 % coverage a bit of a fight against windmills)

### How was it fixed?

Added type hints and enabled `--disallow-untyped-defs` check

#### Cute Animal Picture

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -89,6 +89,7 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
             # Our handle_msg() method runs cpu-intensive tasks in sub-processes so that the main
             # loop can keep processing msgs, and that's why we use ensure_future() instead of
             # awaiting for it to finish here.
+            peer = cast(ETHPeer, peer)
             asyncio.ensure_future(self.handle_msg(peer, cmd, msg))
 
     async def handle_msg(self, peer: ETHPeer, cmd: protocol.Command,
@@ -108,7 +109,7 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
         with self.subscribe(self.peer_pool):
             while True:
                 peer_or_finished = await self.wait_first(
-                    self._sync_requests.get(), self._sync_complete.wait())
+                    self._sync_requests.get(), self._sync_complete.wait())  # type: Any
 
                 # In the case of a fast sync, we return once the sync is completed, and our caller
                 # must then run the StateDownloader.
@@ -403,7 +404,9 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
         # the executor when the list() is applied to it.
         receipts_tries = await self.wait(loop.run_in_executor(self._executor, list, iterator))
         downloaded: List[DownloadedBlockPart] = []
-        for (receipts, (receipt_root, trie_dict_data)) in zip(receipts_by_block, receipts_tries):
+        # TODO: figure out why mypy is losing the type of the receipts_tries
+        # so we can get rid of the ignore
+        for (receipts, (receipt_root, trie_dict_data)) in zip(receipts_by_block, receipts_tries):  # type: ignore # noqa: E501
             await self.wait(self.chaindb.coro_persist_trie_data_dict(trie_dict_data))
             downloaded.append(DownloadedBlockPart(receipts, receipt_root))
         self._downloaded_receipts.put_nowait((peer, downloaded))
@@ -419,7 +422,10 @@ class FastChainSyncer(BaseService, PeerPoolSubscriber):
         transactions_tries = await self.wait(
             loop.run_in_executor(self._executor, list, iterator))
         downloaded: List[DownloadedBlockPart] = []
-        for (body, (tx_root, trie_dict_data)) in zip(bodies, transactions_tries):
+
+        # TODO: figure out why mypy is losing the type of the transactions_tries
+        # so we can get rid of the ignore
+        for (body, (tx_root, trie_dict_data)) in zip(bodies, transactions_tries):  # type: ignore # noqa: E501
             await self.wait(self.chaindb.coro_persist_trie_data_dict(trie_dict_data))
             uncles_hash = await self.wait(self.chaindb.coro_persist_uncles(body.uncles))
             downloaded.append(DownloadedBlockPart(body, (tx_root, uncles_hash)))
@@ -698,7 +704,7 @@ def _test() -> None:
     for sig in [signal.SIGINT, signal.SIGTERM]:
         loop.add_signal_handler(sig, sigint_received.set)
 
-    async def exit_on_sigint():
+    async def exit_on_sigint() -> None:
         await sigint_received.wait()
         await peer_pool.cancel()
         await syncer.cancel()
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -13,11 +13,14 @@ import time
 from typing import (
     Any,
     Callable,
+    cast,
     Dict,
     Iterator,
     List,
     Sequence,
-    Tuple
+    Tuple,
+    Text,
+    Union,
 )
 
 import rlp
@@ -69,7 +72,7 @@ class Command():
         # Elements beyond this length must be trimmed.
         self.elem_count = elem_count
 
-    def __repr__(self):
+    def __repr__(self) -> str:
         return 'Command(%s:%d)' % (self.name, self.id)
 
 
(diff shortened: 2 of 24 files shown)


p2p/chain.py
```diff
@@ -403,7 +404,9 @@ def _handle_block_headers(self, headers: List[BlockHeader]) -> None:
         # the executor when the list() is applied to it.
         receipts_tries = await self.wait(loop.run_in_executor(self._executor, list, iterator))
         downloaded: List[DownloadedBlockPart] = []
-        for (receipts, (receipt_root, trie_dict_data)) in zip(receipts_by_block, receipts_tries):
+        # TODO: figure out why mypy is losing the type of the receipts_tries
+        # so we can get rid of the ignore
```
gsalgado: I'm not sure what this means. Shouldn't mypy treat receipts_tries's type as Any since that's what self.wait() returns?

p2p/discovery.py
```diff
@@ -125,25 +128,23 @@ def _get_handler(self, cmd: Command) -> Callable[[kademlia.Node, List[Any], byte
...
 
-    def _get_max_neighbours_per_packet(self):
+    def _get_max_neighbours_per_packet(self) -> int:
         if self._max_neighbours_per_packet_cache is not None:
             return self._max_neighbours_per_packet_cache
         self._max_neighbours_per_packet_cache = _get_max_neighbours_per_packet()
         return self._max_neighbours_per_packet_cache
 
-    def connection_made(self, transport):
+    def connection_made(self, transport: Any) -> None:
```
gsalgado: Can't we use an actual transport type here?

p2p/eth.py
```diff
@@ -108,7 +108,7 @@ class ETHProtocol(Protocol):
     cmd_length = 17
     logger = logging.getLogger("p2p.eth.ETHProtocol")
 
-    def send_handshake(self, head_info):
+    def send_handshake(self, head_info: Any) -> None:
```
gsalgado: Why use Any here?

p2p/les.py
```diff
@@ -86,7 +95,7 @@ def decode_payload(self, rlp_data: bytes) -> Generator[Tuple[str, Any], None, No
             else:
                 yield key, value
 
-    def encode_payload(self, data):
+    def encode_payload(self, data: Dict[str, bytes]) -> bytes:  # type: ignore
```
gsalgado: Why do we need the `type: ignore` comment here?

p2p/les.py
```diff
@@ -254,7 +263,7 @@ class LESProtocol(Protocol):
     _commands = [Status, Announce, BlockHeaders, BlockBodies, Receipts, Proofs, ContractCodes]
     cmd_length = 15
 
-    def send_handshake(self, head_info):
+    def send_handshake(self, head_info: Any) -> None:
```
gsalgado: Should be able to use `HeadInfo` here, no?

p2p/les.py
```diff
@@ -354,7 +363,7 @@ class LESProtocolV2(LESProtocol):
     _commands = [StatusV2, Announce, BlockHeaders, BlockBodies, Receipts, ProofsV2, ContractCodes]
     cmd_length = 21
 
-    def send_handshake(self, head_info):
+    def send_handshake(self, head_info: Any) -> None:
```
gsalgado: And here

p2p/lightchain.py
```diff
@@ -257,14 +260,14 @@ def register_peer(self, peer: BasePeer) -> None:
             # TODO push validation into process pool executor
             VM.validate_header(header, parent_header)
 
-    async def _cleanup(self):
+    async def _cleanup(self) -> None:
         self.logger.info("Stopping LightPeerChain...")
 
-    async def _wait_for_reply(self, request_id: int) -> Dict[str, Any]:
+    async def _wait_for_reply(self, request_id: int) -> Any:
```
gsalgado: Why this change?

p2p/peer.py
```diff
@@ -238,11 +239,12 @@ def remove_subscriber(self, subscriber: 'asyncio.Queue[PEER_MSG_TYPE]') -> None:
         if subscriber in self._subscribers:
             self._subscribers.remove(subscriber)
 
-    async def do_sub_proto_handshake(self):
+    async def do_sub_proto_handshake(self) -> None:
         """Perform the handshake for the sub-protocol agreed with the remote peer.
 
         Raises HandshakeFailure if the handshake is not successful.
         """
+        self.base_protocol = cast(P2PProtocol, self.base_protocol)
```
gsalgado: `BasePeer.__init__()` does `self.base_protocol = P2PProtocol(self)`, so I'm wondering why this is needed

p2p/peer.py
```diff
@@ -258,11 +260,12 @@ def remove_subscriber(self, subscriber: 'asyncio.Queue[PEER_MSG_TYPE]') -> None:
         await self.process_sub_proto_handshake(cmd, msg)
         self.logger.debug("Finished %s handshake with %s", self.sub_proto, self.remote)
 
-    async def do_p2p_handshake(self):
+    async def do_p2p_handshake(self) -> None:
         """Perform the handshake for the P2P base protocol.
 
         Raises HandshakeFailure if the handshake is not successful.
         """
+        self.base_protocol = cast(P2PProtocol, self.base_protocol)
```
gsalgado: Ditto

p2p/peer.py
```diff
@@ -347,7 +350,7 @@ def close(self):
                 self.logger.debug("%s disconnected: %s", self, e)
                 return
 
-    async def read_msg(self) -> Tuple[protocol.Command, protocol._DecodedMsgType]:
+    async def read_msg(self) -> Tuple[protocol.Command, Dict[str, Any]]:
```
gsalgado: Why this change?  This method can return things other than a dict, that's why it used `protocol._DecodedMsgType`

p2p/peer.py
```diff
@@ -361,12 +364,13 @@ def close(self):
...
         # ProcessPoolExecutor(). Need to make sure we don't use all CPUs in the machine for that,
         # though, otherwise asyncio's event loop can't run and we can't keep up with other peers.
-        decoded_msg = cmd.decode(msg)
+        decoded_msg = cast(Dict[str, Any], cmd.decode(msg))
         self.logger.debug("Successfully decoded %s msg: %s", cmd, decoded_msg)
         return cmd, decoded_msg
 
     def handle_p2p_msg(self, cmd: protocol.Command, msg: protocol._DecodedMsgType) -> None:
         """Handle the base protocol (P2P) messages."""
+        self.base_protocol = cast(P2PProtocol, self.base_protocol)
```
gsalgado: Another cast I wouldn't think to be needed.

p2p/peer.py
```diff
@@ -478,6 +482,7 @@ def disconnect(self, reason: DisconnectReason) -> None:
 
         :param reason: An item from the DisconnectReason enum.
         """
+        self.base_protocol = cast(P2PProtocol, self.base_protocol)
```
gsalgado: And here too

p2p/discovery.py
cburgdorf: Ah, good hint. It's one of these things that I skipped and never came back to. So this represents something like a `asyncio.DatagramTransport` right?

p2p/les.py
cburgdorf: Ah, good point. I skipped that because I didn't notice `HeadInfo` actually exists.

cburgdorf: Ah, turns out it's actually `ChainInfo` there (because `genesis_hash` doesn't exist on `HeadInfo`) 

https://github.com/ethereum/py-evm/pull/947/commits/69892c3ff720ad437d76ca29d1c94df81c5380e1

Notice, I'll rebase these changes into the right place once this passed a complete review.

p2p/les.py
cburgdorf: Both places fixed.

p2p/chain.py
cburgdorf: [`wait(...)`](https://github.com/cburgdorf/py-evm/blob/69892c3ff720ad437d76ca29d1c94df81c5380e1/p2p/service.py#L45-L52) is generic now and should return the right type which works in principle but fails at this place.

p2p/discovery.py
cburgdorf: Fixed by https://github.com/ethereum/py-evm/pull/947/commits/87d585801d1da8b99dbe60595299dab5734c671f

p2p/peer.py
cburgdorf: True, some temporary change that made it through my own review.

p2p/les.py
cburgdorf: It doesn't conform to the signature of the super type.

We could also do this:
```
    def encode_payload(self, data: Union[_DecodedMsgType, sedes.CountableList]) -> bytes:
        response = [
            (key, self.items_sedes[key].serialize(value))
            for key, value
            in sorted(data.items())  # type: ignore
        ]
        return super().encode_payload(response)
```
So, we conform to the interface but we then ignore that `data.items()` is actually invalid because `items()` doesn't exist on a Tuple.

p2p/lightchain.py
cburgdorf: Good point, fixed and explained here https://github.com/ethereum/py-evm/pull/947/commits/6cf408c432c74f78d2029ced398774a6c85f142d

p2p/peer.py
cburgdorf: Yep, we can do better here. Checkout https://github.com/ethereum/py-evm/pull/947/commits/a8632048824259f6816b57aca92c7c72d8c36d24

Basically we:

- get rid of the type of `self.base_protocol` on the class level so that it is really only defined by `__init__`
- return early in `get_protocol_command_for` so that the intermediate variable doesn't have to be types as `Union[Protocol, P2PProtocol]`

p2p/peer.py
cburgdorf: solved, see previous comment

p2p/peer.py
cburgdorf: solved, see previous comment

p2p/peer.py
cburgdorf: solved, see previous comment

p2p/eth.py
cburgdorf: Fixed by https://github.com/ethereum/py-evm/pull/947/commits/7ba3b72ffe102aec18bf3095159841326ac79afe

cburgdorf:
@gsalgado Thanks for the sharp review. I was able to tighten types for a lot of places. I think I addressed all your comments. Once you approve, I'll make sure to move all the wip commits into the first one before merging.

That said, CircleCI seems to have hiccups right now so this won't go green until Circle runs smoothly again.

p2p/peer.py
gsalgado: Are you going to revert it?

cburgdorf: Sorry, fixed by https://github.com/ethereum/py-evm/pull/947/commits/1bbab42b47e15f6453894dcc993e1d9bab925a05

cburgdorf:
Oh, just noticed its not CircleCI that has a hiccup, it's actually blocked by https://github.com/ethereum/py-evm/issues/953

PR #948, opened 2018-06-22
Piper/abstract networking api

pipermerriam:
### What was wrong?

Currently, in some of our tests we actually open network connections and do actual network communication.  This is problematic as it is fragile as well as doesn't operate in isolation.

### How was it fixed?

Creating a new `network` abstraction which defaults to `asyncio` but allows for a mock interface to be slotted in which does all of the communication in-memory.

#### Cute Animal Picture

TODO


pipermerriam:
I'll be out of town all next week.  If anyone likes this and wants to take it somewhere, feel free.

p2p/auth.py
```diff
@@ -118,8 +122,10 @@ def pubkey(self) -> datatypes.PublicKey:
         return self.privkey.public_key
 
     async def connect(self) -> Tuple[asyncio.StreamReader, asyncio.StreamWriter]:
+        from p2p.tools import network
         return await wait_with_token(
-            asyncio.open_connection(host=self.remote.address.ip, port=self.remote.address.tcp_port),
+            #asyncio.open_connection(host=self.remote.address.ip, port=self.remote.address.tcp_port),
+            network.mock_network.open_connection(host=self.remote.address.ip, port=self.remote.address.tcp_port),
```
pipermerriam: This needs to come from `self.network` which defaults to `asyncio`

p2p/server.py
```diff
@@ -109,11 +111,17 @@ def __init__(self,
         if not bootstrap_nodes:
             self.logger.warn("Running with no bootstrap nodes")
 
+    @property
+    def network(self):
+        from p2p.tools import network
+        return network.mock_network
```
pipermerriam: This needs to be settable in the constructor and to default to `asyncio`.

p2p/tools/network.py
```diff
@@ -0,0 +1,144 @@
...
+    client_reader.__id = 'rr'
+    client_writer.__id = 'rw'
+
+    return (
+        server_reader, server_writer,
+        client_reader, client_writer,
+    )
+
+
+# TODO: Need a `Router` and a `Network`.  Network is aware of IP address and Router
```
pipermerriam: This is needed so that `writer.get_extra_info('peername')` can return the proper data.  That means that we instantiate a `Network` object for each `IP` address (and this `Network` object is what gets passed to the `Server` and `PeerPool`.  The `Router` object is more of a global and handles managing connections between `Network` objects.

p2p/tools/network.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+class MockNetwork:
+    servers: Dict[Address, MockServer] = None
```
pipermerriam: this property probably stays on the `Network` object.

p2p/tools/network.py
```diff
@@ -0,0 +1,144 @@
...
+    connections: Set[Tuple[Address, Address]] = None
```
pipermerriam: This potentially moves to the `Router` object.

pipermerriam:
This is now a working POC.  There is one *oddity* that I know of which is commented in the code, related to the behavior of `writer.drain()`, but otherwise, this looks like a viable way for us to substitute fully in-memory networking in our tests.

p2p/tools/network.py
```diff
@@ -0,0 +1,189 @@
...
+
+    reader = asyncio.StreamReader()
+
+    # TODO: Calls to `writer.drain` will block until the corresponding data has
+    # actually been read by the reader.  This is not how things actually behave
+    # in a real networked environment as the call to `writer.drain` will return
+    # once the data has been sent over the protocol, so...
+
+    # Preliminary investigation suggests that this must be done in the
+    # `Protocol` class
```
pipermerriam: @gsalgado I'd like to get your thoughts on this. My current general thoughts on how to approach this are:

Setup two `reader,writer` pairs, `r1,w1` and `r2,w2` such that `w2` is the main writer and `r1` is the main reader.  Data flows `w2 -> r2 -> w1 -> r1`.  Then we create a long running asynchronous loop which continually reads from `r2` and writes what it reads to `w1`.

You can think of this as buffering the data from the main writer to the main reader via an intermediate step (which sort of unintentionally simulates the actual networking).

gsalgado: That sounds ok to me, but I worry a bit that all the extra complexity added by the in-memory mock network will make it even harder to debug failing tests. Some of our tests are already pretty hard to debug because they are implemented as integration tests that rely on multiple components connecting (sometimes over the network) and talking to each other, but I'm not sure simply removing network access from the equation is the right solution -- I think it'd be better to try and rewrite those as unit tests so that they setup/exercise just the pieces they want to test. I understand this is usually tricky and not the most exciting work, so don't let that deter you from working on this mock-networking approach, but I thought it was worth bringing it up

pipermerriam:
@gsalgado and @carver this is ready for review.  I'm strongly leaning towards extracting this out into a stand-alone package which can just become a testing dependency.

carver:
The idea is we don't have to use it for testing the `eth` module, so it's okay to not support py35?

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+            self.write(b'\n')
+
+    def write_eof(self) -> None:
+        self._reader.feed_eof()
+
+    def can_write_eof(self) -> bool:
+        return True
+
+    def is_closing(self) -> bool:
+        return False
```
carver: Should return True after `close()` is called?

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+    reader = asyncio.StreamReader()
+
+    transport = MemoryTransport(reader)
+    protocol = asyncio.StreamReaderProtocol(reader)
+
+    writer = asyncio.StreamWriter(
+        transport=transport,
+        protocol=protocol,
+        reader=reader,
+        loop=asyncio.get_event_loop(),
```
carver: Probably want the ability to override the default loop with an optional kwarg to `direct_pipe`.

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+    NamedTuple,
+    List,
+    Tuple,
+)
+
+import pytest
+
+from p2p.cancel_token import CancelToken, wait_with_token
+from p2p.exceptions import OperationCancelled
+from p2p.service import BaseService
```
carver: You mentioned wanting to pull this module to its own pypi distribution. I assume that means removing all these p2p dependencies. Any ideas/plans for that?

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+        writer.write_eof()
+
+    if reader.at_eof():
+        reader.feed_eof()
+
+
+class Router(BaseService):
+    hosts: Dict[str, 'Host']
+    networks = Dict[str, 'Network']
+    connections = Dict[CancelToken, asyncio.Task]
```
carver: I *think* `CancelToken` could be replaced with `asyncio.Event` in a pretty straightforward way.

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+    def get_network(self, name: str) -> 'Network':
+        if name not in self.networks:
+            self.networks[name] = Network(name, self)
+        return self.networks[name]
+
+    def get_connected_readers(self, address: Address) -> ReaderWriterPair:
+        external_reader, internal_writer = direct_pipe()
+        internal_reader, external_writer = addressed_pipe(address)
+
+        token = CancelToken(str(address)).chain(self.cancel_token)
```
carver: I guess this is the biggest pain point for replacing with `asyncio.Event`. If you want to pull the network mocking into its own distro, maybe the canceltoken stuff has to get pulled out too so we can depend on it in both p2p and mock_network?

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+        self.name = name
+        self.router = router
+        self.default_host = default_host
+
+    @property
+    def default_host(self) -> str:
+        if self._default_host is not None:
+            return self._default_host
+        else:
+            return os.environ.get('P2P_DEFAULT_HOST', '127.0.0.1')
```
carver: This env variable name probably makes less sense when mock-network is in its own distro.

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+                continue
+            elif port in self.servers:
+                continue
+            else:
+                break
+        return port
+
+    async def start_server(self, client_connected_cb: ConnectionCallback, port: int) -> Server:
+        if port in self.servers:
+            raise OSError('Address already in use')
```
carver: Probably want to include the port number in the error

p2p/tools/local_network.py
```diff
@@ -0,0 +1,378 @@
...
+        server = Server(client_connected_cb, address)
+        self.servers[port] = server
+        return server
+
+    def receive_connection(self, port: int) -> ReaderWriterPair:
+        address = Address(self.host, port)
+        if port not in self.servers:
+            raise ConnectionRefusedError("No server running at {0}:{1}".format(self.host, port))
+        elif address.port in self.connections:
+            raise OSError('Address already in use')
```
carver: Ditto, although I guess you might want to match *exatly* what python does, even though it kinda sucks. I think showing the port is better, but I get either choice.

p2p/tools/local_network.py
pipermerriam: Yes, cancel token is going to get it's own pypy distribution `asyncio-cancel-token, and then the new library which I'm calling `pytest-asyncio-networking-simulator` will depend on that distribution.

p2p/tools/local_network.py
pipermerriam: Hrm, the thing I don't think it replaces is the `wait_with_token` functionality which would have to be reproduced to keep the various `awaits` in the connection loop from hanging when things get cancelled.

PR #955, opened 2018-06-25
trinity: New BlockImporter service to workaround issue 902

gsalgado:
Shouldn't be needed once #915 lands, but if that takes longer than we expect we might want to go with this as a temporary solution


carver:
Since #915 is merged, are we good to close this issue?

PR #957, opened 2018-06-25
Lock web3 dependency to 0.4.3

cburgdorf:
### What was wrong?

Lock down web3.py dependency to fix dependency issue with eth-pm. Eth-PM is a dependency of web3.py from 0.4.4 onwards and it's currently demanding for a lower version py-evm version, leaving us with a broken install.

### How was it fixed?

Stick to version 0.4.3 for now

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,7 +42,7 @@ deps = {
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.4",
         "coincurve>=7.0.0,<8.0.0",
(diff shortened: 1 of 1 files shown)


cburgdorf:
Making @davesque and @njgheorghita aware that this is a thing. However, my current thinking is that the way to ensure this isn't affecting us again and again seems to really lock down the web3 dependency to some version that we know is safe to use.

njgheorghita:
@cburgdorf 👍

PR #960, opened 2018-06-26
trinity: Use default preferred peers when none are provided

gsalgado:
The recently introduced PreferredNodeDiscoveryProtocol doesn't look up
the default preferred peers like the old PreferredNodePeerPool did, so
now we do that in ChainConfig when no preferred peers are specified in
the command line

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -264,15 +264,11 @@ class PreferredNodeDiscoveryProtocol(DiscoveryProtocol):
                  privkey: datatypes.PrivateKey,
                  address: kademlia.Address,
                  bootstrap_nodes: Tuple[kademlia.Node, ...],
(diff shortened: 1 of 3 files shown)


p2p/server.py
```diff
@@ -100,7 +100,7 @@ def __init__(self,
         self.max_peers = max_peers
         self.bootstrap_nodes = bootstrap_nodes
         self.preferred_nodes = preferred_nodes
-        if self.preferred_nodes is None and network_id in DEFAULT_PREFERRED_NODES:
+        if not self.preferred_nodes and network_id in DEFAULT_PREFERRED_NODES:
```
carver: If we want to specify that there should be no preferred nodes, an empty sequence seems like a reasonable way to do it. So I think the original `is None` check is reasonable here.

gsalgado: Indeed, reverting that

PR #961, opened 2018-06-26
Create MsgQueueExposer/MsgQueueSubscriber as generic pattern

cburgdorf:
### What was wrong?

We currently use `asyncio.Queue` as a mean to do async communication in our code base. The way we use that in the `PeerPool` is to implement a `PeerPoolSubscriber` class and have the `PeerPool` expose `subscribe` and `unsubscribe` methods.

We may need the same pattern in the tx pool and probably other places as well. (Depends if other areas of our code are interested in certain events of the tx pool)

### How was it fixed?

This is still a wip but I extracted the reusable bits into generic `MsgQueueExposer` and `MsgQueueSubscriber` classes that can be reused.

Notice that I'm not quite happy with the typing just yet because I would really like to get rid of the currently needed casts.

Also notice that I *think* if we want this pattern to be extracted, then I think it should live somewhere outside of this repo to be reusable in other places. Maybe put it in `eth_utils` idk.

#### Cute Animal Picture

### DIFF
--- a/p2p/msg_queue.py
+++ b/p2p/msg_queue.py
@@ -1,0 +1,44 @@
+from abc import (
+    ABC,
+    abstractmethod
+)
+import asyncio
+import contextlib
+from typing import (
+    Iterator,
+    Generic,
+    TypeVar
+)
+
+_TMsg = TypeVar('_TMsg')
+
+
+class MsgQueueExposer(ABC, Generic[_TMsg]):
+
+    @abstractmethod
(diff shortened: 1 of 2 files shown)


cburgdorf:
I'm gonna close this one. As a reference point. I left a question on SO back then to figure out some related mypy stuff. Leaving it here: https://stackoverflow.com/questions/51039507/get-mypy-to-accept-subtype-of-generic-type-as-a-method-argument

Issue #954, opened 2018-06-25
Block importing crashes with KeyError when validating uncles


gsalgado:
### What is wrong?

When validating a newly imported block, we do a `self.get_vm(uncle)` for every uncle of the block, but that will crash when the uncle has not been previously persisted to the DB. That happens because the `get_vm()` call will try to load the uncle's transactions from the DB and will crash with a KeyError when those are not found. This is a regression that has been introduced by #740 -- before that we didn't have the `self.get_vm(uncle)` call as we were using the imported block's VM to perform the uncle validation

### How can it be fixed

Not sure

gsalgado:
This is how it shows up in the logs: 

```
   ERROR  06-23 08:26:27  base_events  Task exception was never retrieved
future: <Task finished coro=<FastChainSyncer.sync() done, defined at /home/salgado/src/py-evm/p2p/chain.py:122> exception=KeyError(b'(e\xed\x84\xeark\xf1\x8c\x9d\xb8/\xce\x9f\x
a0l\xe8\x11\xef\x88\x83\x933<4\xab\x15{\xc9+\xfa1',)>
Traceback (most recent call last):
  File "/home/salgado/src/py-evm/p2p/chain.py", line 135, in sync
    await self._sync(peer)
  File "/home/salgado/src/py-evm/p2p/chain.py", line 184, in _sync
    head_number = await self._process_headers(peer, headers)
  File "/home/salgado/src/py-evm/p2p/chain.py", line 610, in _process_headers
    await self.wait(self.chain.coro_import_block(block, perform_validation=True))
  File "/home/salgado/src/py-evm/p2p/service.py", line 32, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/salgado/src/py-evm/p2p/service.py", line 51, in wait_first
    return await wait_with_token(*awaitables, token=token_chain, timeout=timeout)
  File "/home/salgado/src/py-evm/p2p/cancel_token.py", line 119, in wait_with_token
    return done.pop().result()
  File "/home/salgado/src/py-evm/trinity/utils/mp.py", line 27, in method
    args,
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.6/multiprocessing/managers.py", line 772, in _callmethod
    raise convert_to_error(kind, result)
[...]

PR #962, opened 2018-06-26
Turn VM.validate_uncle() into a @classmethod

gsalgado:
There's no reason why it should be a VM method in the first place,
and by having it as a standalone function we don't need to attempt to
get a VM for an uncle block (in Chain.validate_uncles()), which is not
supposed to work as that requires having the block's body in our DB and
we never have block bodies for uncle blocks.

Closes: #954

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -707,8 +707,8 @@ class Chain(BaseChain):
                     "Uncle ancestor not found: {0}".format(uncle.parent_hash)
                 )
 
-            uncle_vm = self.get_vm(uncle)
-            uncle_vm.validate_uncle(block, uncle, uncle_parent)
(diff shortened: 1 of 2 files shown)


cburgdorf:
Does that mean that uncle validation is set in stone and isn't subject to change between forks? At least this change seems to give up flexibility regarding different uncle validation strategies between forks.

pipermerriam:
Seconding what @cburgdorf .  Lets just make `VM.validate_uncle` a `classmethod` and use `Chain.get_vm_class_for_block_number` instead of `Chain.get_vm`.  This maintains the ability to modify these rules per-fork and should fix #954

gsalgado:
Makes sense! Please review again when you guys have a chance

PR #963, opened 2018-06-26
Remove superfluous tx gas validation

cburgdorf:
### What was wrong?

The `transaction.validate()` call in the next
line has exactly the same logic making this a
superfluous extra check.

https://github.com/ethereum/py-evm/blob/042cd4bc215d8567d560f43564dbee2aba5c970d/evm/rlp/transactions.py#L106-L113

### How was it fixed?

Removed superfluous check

#### Cute Animal Picture

### DIFF
--- a/evm/rlp/transactions.py
+++ b/evm/rlp/transactions.py
@@ -108,7 +108,7 @@ class BaseTransaction(BaseTransactionFields, BaseTransactionMethods):
         Hook called during instantiation to ensure that all transaction
         parameters pass validation rules.
         """
-        if self.intrinsic_gas > self.gas:
+        if self.gas < self.intrinsic_gas:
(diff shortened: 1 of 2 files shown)


cburgdorf:
> Also, the check in evm/rlp/transactions reads like a yoda condition

:joy: Are we talking yoda from star wars?

Just to make sure we mean the same thing. You (and I) would prefer the check to be:

```python
if self.gas < self.intrinsic_gas: 
    raise ValidationError("Insufficient gas") 
```

cburgdorf:
@voith ha! It's a real thing!

carver:
Haha yup, that's the one! :)

carver:
"Intrinsic gas more than the transaction gas, this is" ~Yoda validating a transaction

Issue #945, opened 2018-06-21
Prevent relaying of invalid transactions (possible DDOS attacks)


pipermerriam:
### What is wrong?

The receipt and processing of transactions from peers is a potential DDOS vector.  A peer could spam us with bad transactions which could overload our node as *checking* the validity of a transaction isn't cheap.

This can also cause us to be disconnected from other peers if we relay these bad transactions.

### How can it be fixed

- We should check the validity of a transaction before relating it.
- We should disconnect from peers who send us too many invalid transactions.

cburgdorf:
Sounds good, do you want that addressed as part of #898 before it lands or can it land in a follow up?

pipermerriam:
Follow up, just tracking the issue.

Issue #893, opened 2018-06-11
Ability to blacklist peers


carver:
### What is wrong?

When disconnecting from a peer, some kinds of disconnects suggest that we should not try to reconnect to it later (is peer myself? is it replying with protocol-breaking headers? etc). It would be best to not try to connect to it again later, but nothing is preventing us from retrying.

### How can it be fixed

- Add a peer blacklist
- Check peer against blacklist before trying to connect to it
- Add peer to blacklist for certain kinds of peer failures (like a protocol error)

The blacklist can be in-memory, resetting on every `trinity` reboot.

pipermerriam:
It'd be nice to use something with constant size storage, otherwise this will be a DOS vector.  Lets use a bloom filter and a salt to ensure that remotes cannot generate their own hash collisions.

veox:
Two wei:

-----

> some kinds of disconnects suggest that we should not try to reconnect to it later (is peer myself? is it replying with protocol-breaking headers? etc).

A common case on the ETH public mainnet is other-genesis peers with the same network ID that otherwise follow all [sub-]protocol rules correctly.

This can get annoying/overwhelming, especially since it consumes precious resources. `geth`/`parity` APIs aren't very comprehensive on this, so in a few cases I've resorted to firewall-level connection filtering (think `iptables`/`shorewall`/`fail2ban`).

-----

> Add a peer blacklist

May I suggest more generic/neutral language?.. Like: blocklist; but really: tier, filter, sieve.

The reason is that these things tend to get layered. A not entirely fictional scenario:

* first I want to just block the irrelevant;
* but then I want a limited set of "priority" peers (e.g. a VPN ring specified in a text file);
* then there happens to be "customers" that maybe I'd drop on a rainy day, but they're paying, so I try not to;
* eventually, there's the "too slow" category that gets dropped during low load periods, and when I myself get dropped for the same reason during high load, this is the pool I'd like to dip into, instead of thrashing around randomly.

I guess the gist is: all of these involve peer management, are application-specific and can't be guessed by developers in advance.

A blacklist/blocklist is fine as a start, but long-term this is best given away to the node operator's middleware.

carver:
> A common case on the ETH public mainnet is other-genesis peers with the same network ID that otherwise follow all [sub-]protocol rules correctly.

[...]

PR #964, opened 2018-06-26
Ensure Transaction Pool does not relay invalid transactions

cburgdorf:
### What was wrong?

Currently the transaction pool relays transactions even if they are invalid making that an attack vector (see #945) 

### How was it fixed?

- added a `validate_signed_transaction` API to the Chain
- make the `TxPool` take a `Callable[[BaseTransactionFields], None]` in its constructor as a validation method
- make the `TxPool` validate txs against the passed validation function

#### Cute Animal Picture

### DIFF
--- a/tests/p2p/test_tx_pool.py
+++ b/tests/p2p/test_tx_pool.py
@@ -33,9 +33,7 @@ class TxsRecorder():
         self.recorded_tx.extend(txs)
         self.send_count = self.send_count + 1
 
-
-@pytest.mark.asyncio
-async def test_tx_propagation(monkeypatch, request, event_loop):
+async def bootstrap_test_setup(monkeypatch, request, event_loop):
     peer1, peer2 = await get_directly_linked_peers(
         request,
         event_loop,
@@ -55,6 +53,16 @@ async def test_tx_propagation(monkeypatch, request, event_loop):
     def finalizer():
         event_loop.run_until_complete(pool.cancel())
     request.addfinalizer(finalizer)
+
+    return peer1, peer1_txs_recorder, peer2, peer2_txs_recorder, pool
+
+@pytest.mark.asyncio
+async def test_tx_propagation(monkeypatch, request, event_loop):
+    peer1, peer1_txs_recorder, peer2, peer2_txs_recorder, pool = await bootstrap_test_setup(
+        monkeypatch,
+        request,
+        event_loop
+    )
 
     txs_broadcasted_by_peer1 = [create_random_tx()]
 
--- a/tests/p2p/test_tx_pool.py
+++ b/tests/p2p/test_tx_pool.py
@@ -1,6 +1,6 @@
 import asyncio
 import pytest
-import random
+import uuid
 
 from p2p.peer import (
     ETHPeer,
@@ -8,14 +8,21 @@ from p2p.peer import (
 from p2p.eth import (
     Transactions
 )
-from evm.rlp.transactions import (
-    BaseTransactionFields
+
+from evm.utils.address import (
+    force_bytes_to_address
 )
 
 from trinity.tx_pool.pool import (
     TxPool,
 )
+from trinity.tx_pool.validators import (
+    DefaultTransactionValidator
+)
 
+from tests.conftest import (
+    funded_address_private_key
+)
 from tests.p2p.peer_helpers import (
     get_directly_linked_peers,
     MockPeerPoolWithConnectedPeers,
@@ -33,7 +40,8 @@ class TxsRecorder():
         self.recorded_tx.extend(txs)
         self.send_count = self.send_count + 1
 
-async def bootstrap_test_setup(monkeypatch, request, event_loop):
+
+async def bootstrap_test_setup(monkeypatch, request, event_loop, chain, tx_validator):
     peer1, peer2 = await get_directly_linked_peers(
         request,
         event_loop,
@@ -46,7 +54,10 @@ async def bootstrap_test_setup(monkeypatch, request, event_loop):
     peer1_txs_recorder = create_tx_recorder(monkeypatch, peer1)
     peer2_txs_recorder = create_tx_recorder(monkeypatch, peer2)
 
-    pool = TxPool(MockPeerPoolWithConnectedPeers([peer1, peer2]))
+    pool = TxPool(
+        MockPeerPoolWithConnectedPeers([peer1, peer2]),
+        tx_validator
+    )
 
     asyncio.ensure_future(pool.run())
 
@@ -56,15 +67,28 @@ async def bootstrap_test_setup(monkeypatch, request, event_loop):
 
     return peer1, peer1_txs_recorder, peer2, peer2_txs_recorder, pool
 
+
+@pytest.fixture
+def tx_validator(chain_with_block_validation):
+    return DefaultTransactionValidator(chain_with_block_validation, 0)
+
+
 @pytest.mark.asyncio
-async def test_tx_propagation(monkeypatch, request, event_loop):
+async def test_tx_propagation(monkeypatch,
+                              request,
+                              event_loop,
+                              chain_with_block_validation,
+                              tx_validator):
+
     peer1, peer1_txs_recorder, peer2, peer2_txs_recorder, pool = await bootstrap_test_setup(
         monkeypatch,
         request,
-        event_loop
+        event_loop,
+        chain_with_block_validation,
+        tx_validator
     )
 
-    txs_broadcasted_by_peer1 = [create_random_tx()]
+    txs_broadcasted_by_peer1 = [create_random_tx(chain_with_block_validation)]
 
     # Peer1 sends some txs
     await pool._handle_tx(peer1, txs_broadcasted_by_peer1)
@@ -91,7 +115,10 @@ async def test_tx_propagation(monkeypatch, request, event_loop):
     assert peer1_txs_recorder.send_count == 0
 
     # Peer2 sends old + new tx
-    txs_broadcasted_by_peer2 = [create_random_tx(), txs_broadcasted_by_peer1[0]]
+    txs_broadcasted_by_peer2 = [
+        create_random_tx(chain_with_block_validation),
+        txs_broadcasted_by_peer1[0]
+    ]
     await pool._handle_tx(peer2, txs_broadcasted_by_peer2)
 
     # Check that Peer1 receives only the one tx that it didn't know about
@@ -101,7 +128,35 @@ async def test_tx_propagation(monkeypatch, request, event_loop):
 
 
 @pytest.mark.asyncio
-async def test_tx_sending(request, event_loop):
+async def test_does_not_propagate_invalid_tx(monkeypatch,
+                                             request,
+                                             event_loop,
+                                             chain_with_block_validation,
+                                             tx_validator):
+
+    peer1, peer1_txs_recorder, peer2, peer2_txs_recorder, pool = await bootstrap_test_setup(
+        monkeypatch,
+        request,
+        event_loop,
+        chain_with_block_validation,
+        tx_validator
+    )
+
+    txs_broadcasted_by_peer1 = [
+        create_random_tx(chain_with_block_validation, is_valid=False),
+        create_random_tx(chain_with_block_validation)
+    ]
+
+    # Peer1 sends some txs
+    await pool._handle_tx(peer1, txs_broadcasted_by_peer1)
+
+    # Check that Peer2 received only the second tx which is valid
+    assert len(peer2_txs_recorder.recorded_tx) == 1
+    assert peer2_txs_recorder.recorded_tx[0].hash == txs_broadcasted_by_peer1[1].hash
+
+
+@pytest.mark.asyncio
+async def test_tx_sending(request, event_loop, chain_with_block_validation, tx_validator):
     # This test covers the communication end to end whereas the previous
     # focusses on the rules of the transaction pool on when to send tx to whom
     peer1, peer2 = await get_directly_linked_peers(
@@ -114,7 +169,7 @@ async def test_tx_sending(request, event_loop):
     peer2_subscriber = asyncio.Queue()
     peer2.add_subscriber(peer2_subscriber)
 
-    pool = TxPool(MockPeerPoolWithConnectedPeers([peer1, peer2]))
+    pool = TxPool(MockPeerPoolWithConnectedPeers([peer1, peer2]), tx_validator)
 
     asyncio.ensure_future(pool.run())
 
@@ -122,7 +177,7 @@ async def test_tx_sending(request, event_loop):
         event_loop.run_until_complete(pool.cancel())
     request.addfinalizer(finalizer)
 
-    txs = [create_random_tx()]
+    txs = [create_random_tx(chain_with_block_validation)]
 
     peer1.sub_proto.send_transactions(txs)
 
@@ -147,15 +202,14 @@ def create_tx_recorder(monkeypatch, peer):
     return recorder
 
 
-def create_random_tx():
-    return BaseTransactionFields(
+def create_random_tx(chain, is_valid=True):
+    return chain.create_unsigned_transaction(
         nonce=0,
         gas_price=1,
-        gas=21000,
-        data=b'',
-        to=b'',
-        value=random.randint(0, 1000),
-        r=random.randint(0, 1000),
-        s=random.randint(0, 1000),
-        v=random.randint(0, 1000)
-    )
+        gas=2100000000000 if is_valid else 0,
+        # For simplicity, both peers create tx with the same private key.
+        # We rely on unique data to create truly unique txs
+        data=uuid.uuid4().bytes,
+        to=force_bytes_to_address(b'\x10\x10'),
+        value=1,
+    ).as_signed_transaction(funded_address_private_key())
--- a/trinity/nodes/ropsten.py
+++ b/trinity/nodes/ropsten.py
@@ -1,5 +1,6 @@
 from evm.chains.ropsten import (
     RopstenChain,
+    BYZANTIUM_ROPSTEN_BLOCK
 )
 
 from trinity.chains.ropsten import (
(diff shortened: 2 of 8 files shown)


carver:
So an implicit question here is: which VM should we use to validate a transaction in the pool? I suspect we can't even just validate on one VM.

1. What if the transaction is valid on the current head but invalid at the next fork and we are synced? (easy: treat it as valid)
2. What if the transaction is valid on the current head but invalid at the next fork and we are syncing? (probably treat invalid, assuming at least one of the peers has reached the next fork)
2. What if the transaction is invalid on the current head but valid at the next imminent fork? (probably treat as valid)
2. What if the transaction is invalid on the current head but valid at the next fork that is weeks/months away? (probably treat as invalid)
3. When is the cutoff between the previous two questions?
3. What if there are two forks coming up in quick succession? (we can probably roughly ignore this very unusual case)
5. Probably a lot more...

I'm not saying that we need to implement this all right now, but it helps inform the architecture. The most important takeaway is that there is a lot of logic around deciding which VM to validate with, and how to handle results from multiple VMs. So I think the validation should probably not be in `Chain`, and be a classmethod in `BaseVM` instead. Then `TxPool` can decide which VMs to check, and check them individually.

cburgdorf:
@carver These are great points. :+1:  I have to give them some more thought tomorrow.

cburgdorf:
@carver These points sound all very reasonable. However, I haven't found geth or parity doing these checks as a pre-condition to tx relaying (found tons of other interesting things though :sweat_smile:)

But even if they don't (which I'm not implying) I do see that we want to allow this kind of flexibility.

>So I think the validation should probably not be in Chain, and be a classmethod in `BaseVM` instead.

Adding `validate_transaction` there sounds reasonable but now that I think of it these validations are stateless and hence `vm.get_transaction_class().validate(...)` may equally cut it for now. Especially since a `validate_transaction` method on the VM directly could be misinterpreted as a stateful transaction validation (e.g. checking the balance of the sender).

These stateful validations are run as part of the `apply_transaction(...)` API but it is my understanding that for the tx relaying only stateless validations (the one that can be done with the class method `validate`  on the transaction class) should be considered. Is that correct?

https://github.com/ethereum/py-evm/blob/88fd7887cfd8929b9d7fd6b3c1ee17f6fba736d6/evm/vm/forks/frontier/validation.py#L6-L21

cburgdorf:
>What if the transaction is valid on the current head but invalid at the next fork and we are syncing? (probably treat invalid, assuming at least one of the peers has reached the next fork)

I just thought about this again and it got me thinking if this is actually what we want:

In the `DefaultTransactionValidator` (so that's not hardcoded but a mainnet thing) basically always do `ByzantiumTransaction.validate(transaction)` **except** if our head is already further along than Byzantium. In that case we would use whatever `get_Vm().get_transaction_class()` is.

Because, even if we just started fresh from genesis, we wouldn't want to relay any transactions that do not pass `ByzantiumTransaction.validate(transaction)`, no?. After all, just because our head is still on `Frontier` that doesn't mean we want to relay transactions that do not conform to the **current** criteria. Am I looking at this wrong?

carver:
> (found tons of other interesting things though sweat_smile)

You've piqued my interest, please share! :)

> These stateful validations are run as part of the apply_transaction(...) API but it is my understanding that for the tx relaying only stateless validations (the one that can be done with the class method validate on the transaction class) should be considered. Is that correct?

Right, I don't think we should test against our current state database. So `vm.get_transaction_class().validate(...)` sounds reasonable.

>> What if the transaction is valid on the current head but invalid at the next fork and we are syncing? (probably treat invalid, assuming at least one of the peers has reached the next fork)
> 
> Because, even if we just started fresh from genesis, we wouldn't want to relay any transactions that do not pass ByzantiumTransaction.validate(transaction), no?.

Right, we're saying the same thing, here. While we are syncing (so, far behind peers), we should treat a transaction as valid only if it is valid on the fork that our top peer is on.

Also, after we add a Constantinople VM to the Mainnet config, the node should continue validating against Byzantium, until one of the node's peers have crossed the block number into Constantinople (and maybe only forward transactions to the peers that have crossed over). So I don't think hard-coding will work. We still have to dynamically choose which VM to validate with.

cburgdorf:
>the fork that our top peer is on

Just so that I'm 100 % clear I understand it correctly. What fork our top peer is on, is a question of interpreting their highest block number, right? So, basically we would see that peer x is on block 7m and then lookup in our mapping and say "oh block 7m, that is fork LazorRazorSuperman".

If that is the case, doesn't that mean we are putting a lot of trust into whatever that peer tells us their block number is. My understanding is, that during that time (**we** aren't synced yet!), we haven't validated the entire blockchain yet and hence can't make a statement about whether that peer is trying to fool us or not. E.g. if a peer would trick us to believe that a fork has already happened even if it hasn't, the same peer could trick us to relay lots of transactions that are invalid on Byzantium but are valid on the next fork which is still far out (just because that one peer made us believe that fork has already happened but which is a lie)

>You've piqued my interest, please share! :)

I guess a lot of the things aren't much of a surprise to you but where interesting to me. Here's a super quick brain dump (because I'm about to leave for ECDC Day 2)

1. Having a scoring mechanism (needs more investigation)
2. Logic around replacing tx (nonce, gas price)
3. I think if I didn't misread, Parity seems to relay on a fixed interval rather than immediately as they arrive which seems to make sense and is probably something we also want to do (seems more resource friendly)

Probably tons of other things that I forgot again. I will continue to research parity and geth as we progress with the tx pool.

cburgdorf:
@carver @pipermerriam I pushed a new version.

This doesn't yet take the fork of our top peer into account (see my concerns in the previous comment) but other than that I think it may be a good first start. Here's how it works:

1. There's an abstract `BaseTransactionValidator` that one can implement to tell the tx pool how to handle tx validation.

2. The `TxPool` expects to get an instance of `BaseTransactionValidator` passed upon instantiation

3. There's a `DefaultTransactionValidator` that works as follows:

    A. It expects a `Chain(Proxy)` and a `BaseTransaction` that serves as initial transaction class.
    B. On validation, it will check if the chain is operating on a newer fork than what the initial transaction 
        class belongs to. If yes, then it will use that class, if not then it will use the class that was passed as 
       `initial_tx_class`

4. The class that is used as `initial_tx_class` isn't hardcoded but instead is defined in `MainnetFullNode`, `MainnetLightNode`, `RopstenFullNode` and `RopstenLightNode`. It is set to `ByzantiumTransaction` as that is the minimal criteria that we know every tx must satisfy in order to be relayed on the ethereum network today.

trinity/tx_pool/validators.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    The :class:`~trinity.tx_pool.validators.BaseTransactionValidator` class is the base class for
+    transaction validation done by the :class:`~trinity.tx_pool.pool.TxPool`.
+
+        .. note::
+        This is an abstract class that must be implemented in subclasses
+    """
+
+    @abstractmethod
+    def is_valid(self, transaction: BaseTransactionFields) -> bool:
```
pipermerriam: classes with a single method are generally an anti-pattern for me.  Instead of using a class for this, can we instead just change to just passing in a callable?  

We can still implement more complex validation using classes if desired and just implement `__call__` for the check.

cburgdorf: I generally agree and I had it a plain callable in the beginning but then changed it to a class because I wanted the tx pool to provide a default implementation that takes constructor arguments. We can also solve that using currying if you prefer.

cburgdorf: Ah, or the default one continues to be a class and implements `__call__` as you suggests and we just drop the abstract class in favor for the more lightweight `Callable[[TX], bool]` interface. I guess that would be my preference.

pipermerriam: I'm good with either a class with `__call__` **or** a curried callable.

cburgdorf: Yep, I pushed the call based solution

tests/p2p/test_tx_pool.py
```diff
@@ -93,7 +131,35 @@ def finalizer():
...
+    peer1, peer1_txs_recorder, peer2, peer2_txs_recorder, pool = await bootstrap_test_setup(
+        monkeypatch,
+        request,
+        event_loop,
+        chain_with_block_validation,
+        tx_validator
+    )
+
+    txs_broadcasted_by_peer1 = [
+        create_random_tx(chain_with_block_validation, False),
```
pipermerriam: I'm allergic to boolean literals as positional arguments.  Can you update to calling this using a keyword so it is clear what the `False` does here?

trinity/tx_pool/pool.py
```diff
@@ -41,9 +42,17 @@ class TxPool(BaseService, PeerPoolSubscriber):
...
-    def __init__(self, peer_pool: PeerPool, token: CancelToken = None) -> None:
+    def __init__(self,
+                 peer_pool: PeerPool,
+                 tx_validation_fn: Callable[[BaseTransactionFields], bool],
+                 token: CancelToken = None) -> None:
         super().__init__(token)
         self._peer_pool = peer_pool
+
+        if tx_validation_fn is None:
+            raise ValueError('Must pass a validator instance')
```
pipermerriam: This exception message looks like it needs to be updated.

trinity/tx_pool/pool.py
```diff
@@ -96,6 +105,8 @@ def _filter_tx_for_peer(
         return [
             val for val in txs
             if self._construct_bloom_entry(peer, val) not in self._bloom
+            # TODO: we need to keep track of invalid txs and eventually blacklist nodes
```
pipermerriam: Is there an issue open for this?  Did a quick search and didn't find one.  Seems like something that could be done as a "Good first issue" if you provide a sufficiently detailed spec for it.

trinity/tx_pool/validators.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+
+    def __init__(self, chain: BaseChain, initial_tx_class: Type[BaseTransaction]) -> None:
+        self.chain = chain
+        self._initial_tx_class = initial_tx_class
+        self._ordered_tx_classes = [
+            vm_class.get_transaction_class()
+            for _, vm_class in chain.get_vm_configuration()
+        ]
+        self._initial_tx_index = self._ordered_tx_classes.index(initial_tx_class)
```
pipermerriam: This will throw an error if the provided `initial_tx_class` is not in the `vm_configuration`.  Probably worth checking existence as I think it should be a validation error if the provided *initial* class is not found in the `vm_configuration`.

trinity/tx_pool/validators.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    The :class:`~trinity.tx_pool.validators.DefaultTransactionValidator` class is responsible to
+    decide wether transactions should be relayed to peers or not. This implementation validates
+    transactions against an initially provided transaction class but will switch to a different
+    one as soon as the tip of the chain uses a more up to date transaction class than the one
+    that was initially provided.
+    """
+
+    def __init__(self, chain: BaseChain, initial_tx_class: Type[BaseTransaction]) -> None:
+        self.chain = chain
```
pipermerriam: `chain` only appears to be used to `get_vm_configuration`.  What do you think about changing this class to be initialized directly with the` vm_configuration`.

trinity/tx_pool/validators.py
```diff
@@ -0,0 +1,59 @@
...
+        self._initial_tx_class = initial_tx_class
```
pipermerriam: If we instead used a `block_number` to specify which vm should be used, then we could eliminate the need to validate whether the `initial_tx_class` is present in the `vm_configuration` and just extract the appropriate *initial* vm from the `vm_configuration` based on the provided block number.

tests/p2p/test_tx_pool.py
cburgdorf: Agree. Need to get used to keyword arguments. :+1:

trinity/tx_pool/pool.py
cburgdorf: Yep.

trinity/tx_pool/pool.py
cburgdorf: We have touched on it here. https://github.com/ethereum/py-evm/issues/893 Still need to settle on a design (but I was about to put out a draft in a PR soon)

trinity/tx_pool/validators.py
cburgdorf: We need the chain to get the current VM (or blocknumber theoretically) to validate the tx against the fork at the time of tx relaying. https://github.com/ethereum/py-evm/pull/964/files/9f8056e83067f6b38a251b8e1cf777043cdd6b87#diff-580a9356192014f1ac5f18afe2008b5aR50

trinity/tx_pool/validators.py
cburgdorf: I like that :+1:

trinity/tx_pool/validators.py
cburgdorf: Let's go with the block number as you suggested.

trinity/tx_pool/validators.py
pipermerriam: Ok, do what you think is best but.. maybe it would be good if the `tx_validation_fn` had a different signature that provided this contextual information?

- `Callable[[block_number, transaction], bool]`
- `Callable[[chain, transaction], bool]`

cburgdorf: >maybe it would be good if the tx_validation_fn had a different signature that provided this contextual information

I'm not too crazy about that tbh because that seems too much of an implementation detail of a specific validator to me. In general, I'd like to keep the validator fn `Callable[[tx], bool]` because that really seems like the most common interface that all possible validators would share. E.g. imagine you want a validator that takes a proof of authority approach. It wouldn't care about the block number and maybe not even about the chain but instead check if the tx sender is part of some predefined list.

Of course, we could say such a validator may just ignore `chain` or `block_number` as superfluous contextual information but it feels like a smell to me that we don't want it in the interface at all.

cburgdorf:
@pipermerriam I think I have implemented all your suggestions. Are we good to merge?

trinity/tx_pool/validators.py
```diff
@@ -0,0 +1,62 @@
...
+        # active fork (syncing), ensure that we use the specified initial tx class
+        if self._is_outdated_tx_class(current_tx_class):
+            return self._initial_tx_class
+
+        return current_tx_class
+
+    def _is_outdated_tx_class(self, tx_class: Type[BaseTransaction]) -> bool:
+        return self._ordered_tx_classes.index(tx_class) < self._initial_tx_class_index
+
+    def _get_tx_class_for_block_number(self, block_number: int) -> Type[BaseTransaction]:
```
pipermerriam: I question the necessity of this method over just inlining `self.chain.get_vm_class_for_block_number(block_number).get_transaction_class()` though they both admittedly are not great.  Probably fine either way.

trinity/tx_pool/validators.py
```diff
@@ -0,0 +1,62 @@
...
+        transaction_class = self.get_appropriate_tx_class()
+        tx = transaction_class(**transaction.as_dict())
+        try:
+            tx.validate()
+        except ValidationError:
+            return False
+        else:
+            return True
+
+    def get_appropriate_tx_class(self) -> Type[BaseTransaction]:
```
pipermerriam: This could probably be cached.  Maybe use a time-based caching approach and expire the cache after a few seconds (like 5?).  Probably save us a **lot** of inter-process communication talking back and forth with the chain for every single transaction we want to validate.

cburgdorf: Sounds like a good idea. Is there a standard approach to this? I've found this one but I'm not sure if it's worth the dependency for this single place of use?

http://cachetools.readthedocs.io/en/latest/#cachetools.func.ttl_cache

Anyway, I will create an issue for this so that I can merge before we get to decide this.

trinity/tx_pool/validators.py
cburgdorf: I agree that in this particular case it's not too much of an improvement over the inlined version. I often do these things for improved readability. Anyway, I'm going to leave this as is now to be able to merge without another CI run.

trinity/nodes/ropsten.py
```diff
@@ -11,7 +12,9 @@
 
 class RopstenFullNode(FullNode):
     chain_class = RopstenChain
+    initial_tx_validation_block_number = BYZANTIUM_ROPSTEN_BLOCK
 
 
 class RopstenLightNode(LightNode):
     chain_class = RopstenLightDispatchChain
+    initial_tx_validation_block_number = BYZANTIUM_ROPSTEN_BLOCK
```
carver: This solution is high maintenance: someone has to remember to update this variable in a bunch of different places at every fork. Or, realistically, no one does the maintenance. Then we get a totally silent bug: during syncing, we blacklists peers who send valid Constantinople transactions, if they are invalid in Byzantium.

One alternative is: we always drop transactions that fail validation based on our local node's canonical head. But we treat the peers differently during fast syncing and real-time sync...

- fast sync: do not drop or blacklist peers who send invalid transactions
- real-time sync: drop and blacklist peers who send invalid transactions

cburgdorf: >someone has to remember to update this variable in a bunch of different places at every fork

Yes, that's true. It's a bit better already in the PR that makes the transaction pool a plugin. It's more centralized there (hooray for centralization haha)

```python
    def activate(self, context: PluginContext) -> None:
        if isinstance(self.chain, MainnetChain):
            validator = DefaultTransactionValidator(self.chain, BYZANTIUM_MAINNET_BLOCK)
        elif isinstance(self.chain, RopstenChain):
            validator = DefaultTransactionValidator(self.chain, BYZANTIUM_ROPSTEN_BLOCK)
        else:
            # TODO: We could hint the user about e.g. a --tx-pool-no-validation flag to run the
            # tx pool without tx validation in this case
            raise ValueError("The TxPool plugin does only support MainnetChain or RopstenChain")

        tx_pool = TxPool(self.peer_pool, validator, self.cancel_token)
        asyncio.ensure_future(tx_pool.run())
```

We could probably even maintain some constant that points to whatever the *current fork* is at some central place such as `evm.constants`. It's not ideal but the alternatives aren't that compelling either imho. Given that forks happen rarely I feel it might be not a huge deal if can reduce it to one centrally maintained constant.

Issue #891, opened 2018-06-11
Get rid of code duplication between FastChainSyncer and LightPeerChain


gsalgado:
### What is wrong?

They both implement logic to sync block headers, triggered by peer announcements

### How can it be fixed

Although one uses the LES sub-protocol and the other uses ETH, we should be able to abstract that by having `BasePeer` provide a standard API, and then we can reuse `FastChainSyncer` in `LightPeerChain`

gsalgado:
Closed by #975

PR #966, opened 2018-06-27
Refactor FastChainSyncer and implement a light syncer based on it

gsalgado:
The new BaseHeaderChainSyncer is an abstract class that can be used to
implement all forms of chain syncing (light, full-fast and full-regular)

This is the first step towards fixing #891; next I'm going to use this in LightNode and delete LightPeerChain's header-syncing code


p2p/chain.py
```diff
@@ -20,64 +22,64 @@
...
     # the round-trip times from our download requests.
     _reply_timeout = 60
 
     def __init__(self,
+                 chain: AsyncChain,
                  chaindb: AsyncChainDB,
                  peer_pool: PeerPool,
                  token: CancelToken = None) -> None:
         super().__init__(token)
+        self.logger = logging.getLogger("p2p.chain." + self.__class__.__name__)
```
carver: This is the default logger added by BaseService now, so could be removed.

p2p/chain.py
```diff
@@ -569,10 +601,12 @@ def _handle_get_node_data(self, peer: ETHPeer, msg: List[Hash32]) -> None:
             nodes.append(node)
         peer.sub_proto.send_node_data(nodes)
 
-    async def _process_headers(self, peer: ETHPeer, headers: List[BlockHeader]) -> int:
+    async def _process_headers(
+            self, peer: HeaderRequestingPeer, headers_tuple: Tuple[BlockHeader, ...]) -> int:
         # This is needed to ensure after a state sync we only start importing blocks on top of our
         # current head, as that's the only one whose state root is present in our DB.
-        for header in headers.copy():
+        headers = list(headers_tuple)
```
carver: Maybe `headers` could be named `new_headers` to indicate that we're building the list of headers we don't already have.

p2p/peer.py
```diff
@@ -615,6 +647,10 @@ class ETHPeer(BasePeer):
         self.head_td = msg['td']
         self.head_hash = msg['best_hash']
 
+    def request_block_headers(self, block_number_or_hash: Union[int, bytes],
+                              max_headers: int, reverse: bool = True) -> None:
+        self.sub_proto.send_get_block_headers(block_number_or_hash, max_headers, reverse)
```
carver: It took me a little digging to convince my self that it wasn't a bug that these methods named `send_get_block_headers` have different positional arguments. Maybe rename one of them?

p2p/peer.py
```diff
@@ -515,12 +516,22 @@ def __repr__(self) -> str:
...
 
+    @property
+    def max_headers_fetch(self) -> int:
+        return les.MAX_HEADERS_FETCH
+
+    def handle_sub_proto_msg(self, cmd: protocol.Command, msg: protocol._DecodedMsgType) -> None:
+        if isinstance(cmd, les.Announce):
+            self.head_info = cmd.as_head_info(msg)
+            self.head_td = self.head_info.total_difficulty
+            self.head_hash = self.head_info.block_hash
```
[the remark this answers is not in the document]
gsalgado: For the moment we still need `LESPeer.head_info` as it's used in `LightPeerChain`, but once I replace that with the new `LightChainSyncer` we'll be able to get rid of `head_info` and keep only the `head_td` and `head_hash` attributes

p2p/peer.py
gsalgado: Well, both sub-protocols (LES and ETH) use the same msg name (GetBlockHeaders), but the msg format itself is different and that's why the method signatures are different. I'd rather keep the method names as they are -- that is, just prefixing the msg name with `send_`

Issue #959, opened 2018-06-26
trinity attach command should check for existing IPC handles and exit cleanly if it can't find them


cburgdorf:
### What is wrong?

This is a follow-up to the discussion in https://github.com/ethereum/py-evm/issues/950.
The `trinity attach` command currently doesn't recognize when it fails to attach to a running `trinity` instance leaving users with unhandled exceptions upon `w3` usage.

### How can it be fixed

Check if we find the IPC handles that `trinity attach` assumes to exist. If we don't find them, print a user friendly message and exit with error code.

cburgdorf:
This landed a while back and we forgot to close it.

PR #968, opened 2018-06-27
Log an informative error message when IPC file is not found

glaksmono:
### What was wrong?
The `trinity attach` command currently doesn't recognize when it fails to attach to a running `trinity` instance leaving users with unhandled exceptions upon `w3` usage.

Issue Reference: https://github.com/ethereum/py-evm/issues/959

### How was it fixed?
Throw `FileNotFoundError` if the `ipc_file` not found when executing `trinity attach` command

#### Cute Animal Picture

### DIFF
--- a/trinity/console.py
+++ b/trinity/console.py
@@ -5,6 +5,9 @@ from typing import (
     Any,
     Dict,
 )
+from trinity.utils.log_messages import (
+    create_missing_ipc_error_message,
+)
 
 from cytoolz import merge
 
@@ -58,6 +61,10 @@ def console(ipc_path: Path,
     """
     if env is None:
         env = {}
+
+    # if ipc_path is not found, raise an exception with a useful message
+    if not ipc_path.exists():
+        raise FileNotFoundError(create_missing_ipc_error_message(ipc_path))
 
     # cast needed until https://github.com/ethereum/web3.py/issues/867 is fixed
     w3 = web3.Web3(web3.IPCProvider(str(ipc_path)))
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -129,7 +129,7 @@ def main() -> None:
 
     # if console command, run the trinity CLI
     if args.subcommand == 'attach':
-        console(chain_config.jsonrpc_ipc_path, use_ipython=not args.vanilla_shell)
+        run_console(chain_config, not args.vanilla_shell)
         sys.exit(0)
 
     # start the listener thread to handle logs produced by other processes in
@@ -168,7 +168,7 @@ def main() -> None:
 
     try:
         if args.subcommand == 'console':
-            console(chain_config.jsonrpc_ipc_path, use_ipython=not args.vanilla_shell)
+            run_console(chain_config, not args.vanilla_shell)
         else:
             networking_process.join()
     except KeyboardInterrupt:
@@ -190,6 +190,15 @@ def main() -> None:
         import time; time.sleep(0.2)  # noqa: E702
         kill_process_gracefully(networking_process, logger)
         logger.info('Networking process (pid=%d) terminated', networking_process.pid)
+
+
+def run_console(chain_config: ChainConfig, vanilla_shell_args: bool) -> None:
+    logger = logging.getLogger("trinity")
+    try:
+        console(chain_config.jsonrpc_ipc_path, use_ipython=vanilla_shell_args)
+    except FileNotFoundError as err:
+        logger.error(str(err))
+        sys.exit(1)
 
 
 @setup_cprofiler('run_database_process')
--- a/tests/trinity/core/cli/test_trinity_repl.py
+++ b/tests/trinity/core/cli/test_trinity_repl.py
@@ -1,15 +1,21 @@
 import pytest
 
 from trinity.console import console
+from pathlib import Path
+from trinity.utils.log_messages import (
+    create_missing_ipc_error_message,
+)
 
 
-def test_console(jsonrpc_ipc_pipe_path):
-    # Test running the console, actually start it.
-    with pytest.raises(OSError, match='^reading .* stdin .* captured$'):
-        console(jsonrpc_ipc_pipe_path)
+def test_console(caplog, jsonrpc_ipc_pipe_path):
+    # if ipc_path is not found, raise an exception with a useful message
+    with pytest.raises(FileNotFoundError):
+        console(Path(jsonrpc_ipc_pipe_path))
+        assert create_missing_ipc_error_message(jsonrpc_ipc_pipe_path) in caplog.text
 
 
-def test_python_console(jsonrpc_ipc_pipe_path):
-    # Test running the default python REPL, actually start it.
-    with pytest.raises(OSError, match='^reading .* stdin .* captured$'):
-        console(jsonrpc_ipc_pipe_path, use_ipython=False)
+def test_python_console(caplog, jsonrpc_ipc_pipe_path):
+    # if ipc_path is not found, raise an exception with a useful message
+    with pytest.raises(FileNotFoundError):
+        console(Path(jsonrpc_ipc_pipe_path), use_ipython=False)
+        assert create_missing_ipc_error_message(jsonrpc_ipc_pipe_path) in caplog.text
--- a/trinity/utils/log_messages.py
+++ b/trinity/utils/log_messages.py
@@ -1,0 +1,11 @@
+from pathlib import Path
+
+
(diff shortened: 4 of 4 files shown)


glaksmono:
If we exit after figuring out the `ipc_path` file is not found, then it will fail at `py36-trinity`: https://circleci.com/gh/ethereum/py-evm/34147?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

glaksmono:
@cburgdorf I'm aware that there's an error on `py36-trinity` test [here](https://circleci.com/gh/ethereum/py-evm/34256?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), but I feel that we need to fix the test instead - thoughts?

cburgdorf:
@glaksmono thanks for jumping on this. :+1: :tada: 

I think we want to leave the user with a proper message that explains that we are unable to find a running trinity to attach to. We should also hint that the the user may try to run `trinity attach` under a different user than the main process.

Also, I think we should rather move the check here and don't raise any exception at all:

https://github.com/ethereum/py-evm/blob/53f2aab566e55df1ee092ea906e2553835d363f4/trinity/main.py#L131-L133

There you can do something like:

```python
logger.error(TRINITY_ATTACH_IPC_NOT_FOUND_ERROR)
sys.exit(1)
```

What do you think about that?

carver:
> Also, I think we should rather move the check here and don't raise any exception at all:

I agree with half of this.

Any method that calls `console()` will probably want/expect the check to happen, so having the check inside the method seems reasonable.

Definitely agree that we want more information and instruction about the missing path:

```py
if not path.exists():
    logging.error("There is no IPC server at {0}, make sure your node is still running, and start it with something like `TRINITY_JSONRPC_IPC={0} trinity".format(path))
    return
```

carver:
Also :+1: for updating the test. You can use pytest's `caplog` to confirm that the log is emitted.

glaksmono:
@carver @cburgdorf I feel like this [tests](https://github.com/ethereum/py-evm/blob/master/tests/trinity/core/cli/test_trinity_repl.py) should do more coverages:
* Make sure that `trinity attach` works
* Make sure that in the case when IPC file is not found, then it prints error log (currently)

But we should probably create another task for it?

trinity/console.py
```diff
@@ -56,9 +57,17 @@ def console(ipc_path: Path,
     Method that starts the chain, setups the trinity CLI and register the
     cleanup function.
     """
+    logger = logging.getLogger('trinity')
     if env is None:
         env = {}
 
+    # if ipc_path is not found, print a useful error message
+    if not Path(ipc_path).exists():
```
cburgdorf: `ipc_path` is already a `Path` object so you can do `ipc_path.exists()` directly.

trinity/console.py
```diff
@@ -56,9 +57,17 @@ def console(ipc_path: Path,
...
+        logger.error("The IPC path at %s is not found. "
```
cburgdorf: @glaksmono if you build this error message from a function e.g. `create_missing_ipc_error_message(ipc_path: Path) -> str:` then you could import that function in your tests and ensure we don't have to maintain that string as a duplicate in the test.

cburgdorf:
@carver 

>Any method that calls console() will probably want/expect the check to happen, so having the check inside the method seems reasonable

Yes, fair point. I just noticed we have multiple places calling `console(...)` so it makes even more sense to do the check inside.

Also simply returning instead of raising and then catching in `main` seems reasonable but let me bring up that this will print the error and close the process with exit code `0` aka "no error".

I *think* the correct thing to do would actually be:

1. Do raise an error in `def console`
2. In `main` wrap the `console(...)` invocations in a common method that catches the error and does a `sys.exit(1)`

But feels like bike shedding to me and I'm fine with just returning and leaving the program close without error code.  ¯\_(ツ)_/¯

@glaksmono 

Yes, having a test that ensures that `trinity attach` actually works and maybe even going as far to check that some basic `w3` operations work would be **very** valuable.

This stuff is still so easy to break with only user feedback left for us to become aware of it. But as you said, this can totally be tackled in another PR.

glaksmono:
Done, and created this follow up issue https://github.com/ethereum/py-evm/issues/973

trinity/console.py
carver: On the other hand, if that method started returning `""` then the test would still pass. I kind of like the style of just checking for one or two key words in the error during the test. (not opposed to extracting the message to a method, btw, just think there are risks to using the same method to generate as to test)

cburgdorf: I see your point but an empty string seems to be the only return value that would imply that risk and it seems unlikely such a change would ever slip through a review. If in doubt, I'm usually more on the DRY side. No strong opinion though.

cburgdorf:
I tested this locally and it does its job

```
$ trinity --data-dir /tmp/foobar attach
    INFO  06-30 08:19:59        main  
      ______     _       _ __       
                           /____/   
    INFO  06-30 08:19:59        main  Trinity/0.2.0a26/linux/cpython3.6.5
    INFO  06-30 08:19:59        main  Trinity DEBUG log file is created at /tmp/foobar/logs/trinity.log
   ERROR  06-30 08:19:59     console  The IPC path at /tmp/foobar/jsonrpc.ipc is not found. Please run trinity attach using --data-dir param to specify the IPC path.
```

However, I think we can make the error message more user friendly. Something like:


>Failed to attach to a running `trinity` node. The expected IPC file {filename} does not exist. In case the trinity node that you want to attach to runs under a different user account, you can run `trinity --data-dir <path-to-running-nodes-data-dir> attach` to make the `attach` command aware of it. 
 

Doesn't need to be exactly that text but something in that direction would be helpful imho. I was considering for a second to even include that the second user obviously needs read/write access to the other users directory but couldn't word it in nicely and dropped it eventually ;)

cburgdorf:
@pipermerriam any objections to merge this as is?

cburgdorf:
@pipermerriam I think that basically echos my [previous comment](https://github.com/ethereum/py-evm/pull/968#issuecomment-400944731)

>I think the correct thing to do would actually be:
>1. Do raise an error in def console
>2. In main wrap the `console(...)` invocations in a common method that catches the error and does a `sys.exit(1)`

In that case, I'd prefer to keep this PR pending and introduce the changes that way directly. Reasoning is that:

1. It's not urgent
2. It's simple to do

@glaksmono Are you ok to make that final change?

glaksmono:
Done @cburgdorf @pipermerriam

trinity/main.py
```diff
@@ -171,6 +174,8 @@ def main() -> None:
             console(chain_config.jsonrpc_ipc_path, use_ipython=not args.vanilla_shell)
         else:
             networking_process.join()
+    except FileNotFoundError:
```
cburgdorf: Can you wrap the `console(...)` invocation in one common method that catches and does the `sys.exit(1)`

trinity/main.py
```diff
@@ -192,6 +192,13 @@ def main() -> None:
         logger.info('Networking process (pid=%d) terminated', networking_process.pid)
 
 
+def run_console(chain_config: ChainConfig, vanilla_shell_args: bool) -> None:
+    try:
+        console(chain_config.jsonrpc_ipc_path, use_ipython=vanilla_shell_args)
+    except FileNotFoundError:
+        sys.exit(1)
```
pipermerriam: I think this is missing a logging statement?  Maybe I'm dense but I don't see anywhere where the error will make it's way to the console.

cburgdorf: @pipermerriam yes, that's right. Seems it got lost in that refactoring. 

@glaksmono 
The message is only inside the exception now but it still needs to be printed to the user.

glaksmono: My bad. Just fixed it @pipermerriam @cburgdorf

trinity/console.py
```diff
@@ -56,9 +60,16 @@ def console(ipc_path: Path,
...
     cleanup function.
     """
+    logger = logging.getLogger("trinity")
     if env is None:
         env = {}
 
+    # if ipc_path is not found, raise an exception with a useful message
+    if not ipc_path.exists():
+        ipc_error_message = create_missing_ipc_error_message(ipc_path)
+        logger.error(ipc_error_message)
```
pipermerriam: My apologies for not being clear.  I do not think the logging message should be emitted within the `console` function, but instead, at the location where the exception is caught just before the `sys.exit(1)`.  You can use the exception message for the logging message.

```python
try:
    console(...)
except FileNotFoundError as err:
    logger.error(str(err))
    sys.exit(1)
```

PR #971, opened 2018-06-28
Added benchmarks for erc20 transfer, approve, and transferfrom

williambannas:
### What was wrong?

The existing ERC20 benchmark did only deploy but not interact with the ERC20 contract.

### How was it fixed?

- Created four benchmarks:

  1.One that only deploys the contract
  2. One that, based on a deployed contract, calls `transfer`
  3. One that, based on a deployed contract, calls `approve`
  4. One that, based on a deployed contract with pre approval, calls `transferFrom`

- All four benchmarks share some common code.
- A `BaseERC20Benchmark` provides the base layer and provides a hook to perform some preparation work on the chain that does not count into the measured time
- Inherited classes perform each task in isolation

#### Cute Animal Picture


scripts/benchmark/checks/erc20_interact.py
```diff
@@ -0,0 +1,233 @@
...
+        return 'ERC20 ' + self.interact
+
+    def execute(self) -> DefaultStat:
+        total_stat = DefaultStat()
+
+        for chain in get_all_chains():
+
+            # deploy and mine the simple token contract
+            self.simple_token = self.deploy_simple_token(chain)
+            if (self.interact == "transferFrom"):
```
cburgdorf: Magic strings scare me because the type system can't help you if you just mistype. Let's use an `enum` for the different modes here.

```python
from enum import Enum, auto

class ERC20BenchmarkMode(Enum):

    DeployOnly = auto()
    Transfer = auto()
    SomeOtherMode = auto()
```

Then you can check `self.mode == ERC20BenchmarkMode.DeployOnly` etc.

carver:
> save the state of the chain after contract deployment in some immutable way to allow cloning it for each run of these other benchmarks to use it as a starting point.

The first approach that comes to mind is to rewind the chain to an old header. In pytest, you could try something like:

```py
@pytest.fixture(scope="module")
def chain_with_contract_setup():
    # setup stuff
    return mining_chain

@pytest.fixture
def chain_reset_to_contract_setup(chain_with_contract_setup):
    # on every test, capture the original header and reset it on exit

    orig_header = chain_with_contract_setup.header
    yield chain_with_contract_setup

    # the original header is reset on the chain instance that is reused on all tests in the module
    chain_with_contract_setup.header = orig_header
```

cburgdorf:
@carver Oh sweet! This looks easier than expected :)

@williambannas instead of resetting the chain using pytest, we need to reset it from within the benchmark. We may also want to tune the way how the benchmark runner currently works to make it more convenient to run multiple benchmarks from one `BaseBenchmark` inherited class. Does this help you to move things further?

williambannas:
@cburgdorf  I think so, the edits I was planning to make this change were going to go into chain_plumbing.py, so that we are only creating one chain per vm, and then get_all_chains() would yield the current chains instead of creating new ones.

cburgdorf:
>and then `get_all_chains()` would yield the current chains instead of creating new ones.

That's a risky thing because *in general* you do want as much isolation as possible so I'd prefer `get_all_chains()` to not recycle any existing chain.

I think the recycling of this deployed contract is something very specific to that benchmark so I'd prefer to have some code directly in that benchmark to handle that.

williambannas:
okay that makes a lot more sense, will do :)

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -42,41 +44,70 @@
...
+SECOND_TX_GAS_LIMIT = 60000
 
 CONTRACT_FILE = 'scripts/benchmark/contract_data/erc20.sol'
 CONTRACT_NAME = 'SimpleToken'
 
 W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
 
 
-class DeployErc20(BaseBenchmark):
+class ERC20BenchmarkMode(Enum):
```
pipermerriam: I think this would be better structured if we dropped the `mode` concept and instead used subclasses of base class to implement each benchmark.

```
class BaseERC20Benchmark(BaseBenchmark):
    def execute(self):
        ... # setup code here
        self._do_benchmark(...)

class ERC20TransferBenchmark(BaseERC20Benchmark):
    name = 'ERC20 Transfer Benchmark'

    def _do_benchmark(self):
        ... # logic for the `tranfer` test
```

scripts/benchmark/contract_data/ContractFactory.sol
```diff
@@ -0,0 +1,30 @@
+pragma solidity ^0.4.19;
+
+contract ContractFactory{
```
pipermerriam: This looks like it may have snuck in un-intentionally?

williambannas: No, this is intentional, this is the contract that will be used for testing the DOS benchmarks, however it is currently poorly named

pipermerriam: Ah, could we keep that separate and focus on just the ERC20 benchmarks in this PR and do the DOS benchmarks in a new PR?

scripts/benchmark/checks/erc20_interact.py
cburgdorf: @pipermerriam the one thing to keep in mind here though is that we are currently thinking of running these benchmarks in a defined order, with the benchmark that deploys the contract as the first one, and then snapshot the chain to perform the transfer/approve benchmarks on top of an already deployed contract rather than redeploying the contract again (to have a better isolation of *what* is being tested).

Not saying, this can not be done in the inheritance model but it may be slightly more pragmatic in the one single instance, three different calls model.

No strong opinion, just weighing in here to give some perspective. But I'd say we can shoot for *any* approach that achieves the "only test `transfer` / `approve` without redeploying the contract" criteria and then look at how to maybe refactor that into the architecture we want.

williambannas: @cburgdorf so after making the persistent chain, I'm not sure if it makes the most sense, as it requires the overhead of holding onto the chains through multiple benchmarks in memory, and the difference in benchmark execution time is very very small, (I can get an exact measurement if needed). 

I also don't think that the persistent chain really helps with isolation, because wthe initial setup transactions do not get included in the benchmarked data at all.  The setup transactions are sent and mined before any of the timing, and benchmarking even begins, and is not included in any of the benchmarked data or any of the totals. 

I also think that it makes things a bit harder to understand, but I think that is more my fault then anything, but still something that needs to be addressed.

side note: when I was testing initially, before realized to revert the canonical head of the chain, I found that if I ran the full transfer benchmark twice, the second time it would run more then a second slower in every VM version, because the head wouldn't reset, and the chain would start mining on block 201. (this is fixed by resetting the canonical head of the chain) but just wanted to point that out. Im not sure if that time difference is noteworthy but it seemed so to me

cburgdorf: >The setup transactions are sent and mined before any of the timing, and benchmarking even begins, and is not included in any of the benchmarked data or any of the totals.

Yes, I think moving the contract deployment into a setup step that is run before the measurement actually starts will also work (as long as the setup time is small enough which is what you suggest). So, I'm ok to use that as an approach especially as it makes the benchmarks less inter dependent.

>side note: when I was testing initially, before realized to revert the canonical head of the chain, I found that if I ran the full transfer benchmark twice, the second time it would run more then a second slower in every VM version, because the head wouldn't reset, and the chain would start mining on block 201

Interesting. There's probably a reason for that. One second slower seems like a big difference here but my guess would be that there is a strong difference between starting on an empty chain vs starting on a chain that already has a couple of hundred blocks but that the difference won't scale linearly onwards from here. Just guessing though.

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -0,0 +1,354 @@
...
+class ERC20DeployBenchmark(BaseERC20Benchmark):
+    def __init__(self, persist: Persistant) -> None:
+        super().__init__(persist)
+
+    @property
+    def name(self) -> str:
+        return 'ERC20 deployment'
+
+    def _setup_benchmark(self, chain: MiningChain) -> None:
+        return
```
cburgdorf: Not looking into the actual code yet (because I guess it will change again to not reset the header). Just commenting on the code style: For an empty implementation you write `pass` instead of `return`

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -0,0 +1,354 @@
...
+        super().__init__(persist)
+
+    @property
+    def name(self) -> str:
+        return 'ERC20 Transfer'
+
+    def _setup_benchmark(self, chain: MiningChain) -> None:
+        self.deploy_simple_token(chain)
+        chain.mine_block()
+        return
```
cburgdorf: No need to explicitly write `return`. Every method returns implicitly with `None` if no `return <something>` is given

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -0,0 +1,354 @@
...
+
+    @property
+    def name(self) -> str:
+        return 'ERC20 Approve'
+
+    def _setup_benchmark(self, chain: MiningChain) -> None:
+        self.deploy_simple_token(chain)
+        self.erc_transfer(self.addr1, chain)
+        chain.mine_block()
+        return
```
cburgdorf: No need to explicitly write `return`. Every method returns implicitly with `None` if no `return <something>` is given

scripts/benchmark/checks/erc20_interact.py
williambannas: on my machine running only the ERC20 benchmarks blocks = 100, tx/block = 2:

Persistent chain benchtime : 87.543 seconds
Persistent chain total run time : 87.548 seconds

new chain benchtime : 86.435 seconds
new chain total run time : 87.269 seconds

cburgdorf:
Hey @williambannas I went through this and performed a full review. It looks pretty good to me :+1: 

I just applied some minor fine tuning on top of this. At the end we can squash it all into one commit of yours so that there will be only one clean commit from you.

However, I'd like to get one more review (looking at you @carver) just to be sure ;)

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -71,18 +73,12 @@ def __init__(self, num_blocks: int = 100, num_tx: int = 2) -> None:
         self.addr1 = Web3.toChecksumAddress(FUNDED_ADDRESS)
         self.addr2 = Web3.toChecksumAddress(SECOND_ADDRESS)
 
-    @property
-    @abstractmethod
-    def name(self) -> DefaultStat:
```
cburgdorf: @williambannas `name` is abstract in `BaseBenchmark` so it continues to be abstract in inherited classes until it is implemented. No need to implement it as abstract again.

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -71,18 +73,12 @@ def __init__(self, num_blocks: int = 100, num_tx: int = 2) -> None:
...
-        raise NotImplementedError(
-            "Must be implemented by subclasses"
-        )
-
-    @abstractmethod
```
cburgdorf: @williambannas I made `_setup_benchmark` non-abstract because not every child needs that hook.

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -258,7 +259,6 @@ def name(self) -> str:
 
     def _setup_benchmark(self, chain: MiningChain) -> None:
         self._deploy_simple_token(chain)
-        self._erc_transfer(self.addr1, chain)
```
cburgdorf: @williambannas seems we don't need the `transfer` here so I removed that

cburgdorf:
@carver 

> I favor a compositional approach over subclassing

I actually agree 100 %. In fact, my traditional programming style (like, outside this code base because is the first real Python code base that I work with) is far less inheritance friendly. I usually try to avoid most of inheritance. I adapted to the style of this code base to use inheritance more but I will say, that I think, in general, we should consider to use less inheritance and more composition.

However, for the sake of moving this forward and having @williambannas land this, even if we want to refactor that, we could do that in a follow up PR. In fact, I guess @williambannas would have fun trying to take a stab at it :hocho: :smile:

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -0,0 +1,283 @@
...
+
+FIRST_TX_GAS_LIMIT = 1400000
+SECOND_TX_GAS_LIMIT = 60000
+TRANSFER_AMOUNT = 1000
+TRANSER_FROM_AMOUNT = 1
+
+CONTRACT_FILE = 'scripts/benchmark/contract_data/erc20.sol'
+CONTRACT_NAME = 'SimpleToken'
+
+W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
```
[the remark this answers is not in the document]
cburgdorf: We need to set *something* in order to prevent web3 trying to call out via RPC (which it isn't connect to anything).

```
  File "./scripts/benchmark/run.py", line 89, in <module>
    run()
  File "./scripts/benchmark/run.py", line 83, in run
    total_stat = total_stat.cumulate(benchmark.run(), increment_by_counter=True)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/base_benchmark.py", line 53, in run
    stat = self.execute()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/erc20_interact.py", line 98, in execute
    lambda: self.mine_blocks(chain, self.num_blocks, self.num_tx)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/base_benchmark.py", line 58, in as_timed_result
    return time_call(fn)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/utils/meters.py", line 18, in time_call
    return_value = fn()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/erc20_interact.py", line 98, in <lambda>
    lambda: self.mine_blocks(chain, self.num_blocks, self.num_tx)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/erc20_interact.py", line 116, in mine_blocks
    block = self.mine_block(chain, i, num_tx)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/erc20_interact.py", line 126, in mine_block
    self._apply_transaction(chain)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/erc20_interact.py", line 233, in _apply_transaction
    self._deploy_simple_token(chain)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/scripts/benchmark/checks/erc20_interact.py", line 136, in _deploy_simple_token
    w3_tx = SimpleToken.constructor().buildTransaction(W3_TX_DEFAULTS)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/utils/decorators.py", line 14, in _wrapper
    return self.method(obj, *args, **kwargs)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/contract.py", line 862, in buildTransaction
    return fill_transaction_defaults(self.web3, built_transaction)
  File "cytoolz/functoolz.pyx", line 232, in cytoolz.functoolz.curry.__call__
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/utils/transactions.py", line 55, in fill_transaction_defaults
    default_val = default_getter(web3, transaction)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/utils/transactions.py", line 26, in <lambda>
    'gas': lambda web3, tx: web3.eth.estimateGas(tx),
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/eth.py", line 303, in estimateGas
    [transaction],
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/manager.py", line 107, in request_blocking
    response = self._make_request(method, params)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/web3/manager.py", line 99, in _make_request
    params,

```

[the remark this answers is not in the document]
carver: For the stack trace you gave: one option is to move the `gas` value from https://github.com/ethereum/py-evm/pull/971/files#diff-373c9624e8e0573238de9774d078a2c6R143
up a line to https://github.com/ethereum/py-evm/pull/971/files#diff-373c9624e8e0573238de9774d078a2c6R136

cburgdorf:
Hey @williambannas I squashed everything in one commit (you as the author obviously) and landed it. :tada: Thanks for driving this home :pray:

Issue #967, opened 2018-06-27
Code to figure out common ancestor with peer is wasteful


gsalgado:
### What is wrong?

When we start a sync with a given peer, our `ChainSyncer` requests a chain of (192) headers, in reverse order, starting from our current head, then validates that chain and fetches/stores bodies for them. It does that in order to avoid having to deal with chain reorgs, and although that was a known corner we cut (there's a FIXME comment in the code for it), it is very wasteful specially on a fully synced node as it will end up fetching/processing 192 blocks when we are missing just one block.

### How can it be fixed

We could fetch a header chain starting MAX_REORG_DEPTH blocks before our current local head, discard all the ones already present in our DB and process all the others normally

PR #974, opened 2018-06-28
Improve logic to find common ancestor when starting sync

gsalgado:
When starting a sync with a peer, ChainSyncer used to always fetch the 192
headers prior to the current local head and reprocess them, which is very
wasteful specially when we're already fully synced. Now it only fetches
MAX_REORG_DEPTH previous headers and skips processing the ones we already
have.

Depends on #966
Closes: #967

### DIFF
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -62,3 +62,6 @@ ROPSTEN_BOOTNODES = (
 
 # Maximum peers number, we'll try to keep open connections up to this number of peers
 DEFAULT_MAX_PEERS = 25
+
+# Maximum allowed depth for chain reorgs.
+MAX_REORG_DEPTH = 24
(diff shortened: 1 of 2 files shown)


p2p/constants.py
```diff
@@ -62,3 +62,6 @@
 
 # Maximum peers number, we'll try to keep open connections up to this number of peers
 DEFAULT_MAX_PEERS = 25
+
+# Maximum allowed depth for chain reorgs.
+MAX_REORG_DEPTH = 24
```
carver: Is 24 used by other clients? If I was making up a number, it would be bigger. But also it would be nice to have a framework to decide a reasonable reorg maximum.

gsalgado: No, geth uses a somewhat complicated algorithm that I didn't understand well. I filed #978 so that we don't forget about this

Issue #747, opened 2018-05-22
LightPeerChain should serve chain data as well


gsalgado:
### What is wrong?

Currently, our `LightPeerChain` implementation only has APIs for retrieving chain data from other peers, but it simply drops any requests for chain data from peers

### How can it be fixed

It should handle those requests for chain data. That will require our `LESPeer` implementation to grow a few APIs, some of which already exist in `tests/p2p/test_lightchain.py`

gsalgado:
Closing since the new `LightChainSyncer` serves headers, which I believe is all we can do when running a light node, right?

PR #975, opened 2018-06-29
Use the new LightChainSyncer in trinity

gsalgado:
Also remove the sync-related code from LightPeerChain, turning it into just
a service that only fetches data on demand. (Closes: 891)

As a bonus, extend LightChainSyncer to serve headers as well (Part of #747)

Depends on #974


carver:
I may have rushed through my first review on my way out the door. Some of the things that got pulled from `LightPeerChain` don't appear to be reproduced anywhere else now, like `LESAnnouncementProcessingError` and `TooManyTimeouts` which previously caused us to disconnect from a peer. Is that a bug or do we not need them anymore for some reason?

gsalgado:
The code that raised/caught those exceptions was the original light-sync loop I wrote a long time ago, which was buggy and used a sub-optimal approach to timeouts from peers. The whole point of this PR was replacing that with the loop from `BaseHeaderChainSyncer`, which is functionally equivalent but less buggy. So yeah, those exceptions are no longer needed, I just forgot to remove them

PR #976, opened 2018-06-29
A few small fixes

gsalgado:
## Fix ChainSyncer handling of body/receipts/node requests from peers

- When handling body/receipts requests for blocks we don't know, we'd
 fail with an undhandled HeaderNotFound
- Handling of trie Node requests didn't work on trinity because it
 used a ChainDB method that was not exposed in trinity's ChainDBProxy

## Re-enable PeerPool's periodically reporting of peer stats

It had been accidentally removed.

## Make logging less noisy by using a new (LOG_TRACE) level

Peer was logging (with DEBUG level) all received/sent msgs, making it so
noisy that we had to forcibly set its level to INFO on trinity. This
changes it to log those on a different level that is currently never shown
when running trinity

Similaryly, change all debug() logs in evm/vm to use the new level as those
make trinity's log pretty much useless when syncing in regular
(block processing) mode

### DIFF
--- a/evm/db/account.py
+++ b/evm/db/account.py
@@ -5,7 +5,7 @@ from abc import (
 from uuid import UUID
 import logging
 from lru import LRU
-from typing import Set, Tuple  # noqa: F401
+from typing import cast, Set, Tuple  # noqa: F401
 
 from eth_typing import (
     Address,
@@ -43,6 +43,9 @@ from evm.validation import (
     validate_canonical_address,
 )
 
+from evm.utils.logging import (
+    TraceLogger
+)
 from evm.utils.numeric import (
     int_to_big_endian,
 )
@@ -174,7 +177,7 @@ class BaseAccountDB(ABC):
 
 class AccountDB(BaseAccountDB):
 
-    logger = logging.getLogger('evm.db.account.AccountDB')
+    logger = cast(TraceLogger, logging.getLogger('evm.db.account.AccountDB'))
 
     def __init__(self, db, state_root=BLANK_ROOT_HASH):
         r"""
@@ -403,7 +406,7 @@ class AccountDB(BaseAccountDB):
         self._journaltrie.commit(trie_changeset)
 
     def make_state_root(self) -> Hash32:
-        self.logger.debug("Generating AccountDB trie")
+        self.logger.trace("Generating AccountDB trie")
         self._journaldb.persist()
         self._journaltrie.persist()
         return self.state_root
@@ -424,7 +427,7 @@ class AccountDB(BaseAccountDB):
                 else:
                     accounts_displayed.add(address)
                     account = self._get_account(address)
-                    self.logger.debug(
+                    self.logger.trace(
                         "Account %s: balance %d, nonce %d, storage root %s, code hash %s",
                         encode_hex(address),
                         account.balance,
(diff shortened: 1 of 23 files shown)


evm/constants.py
```diff
@@ -1,10 +1,16 @@
...
 from eth_typing import (
     Address,
     Hash32
 )
 from eth_utils import denoms
 
 
+# FIXME: Need to move this somewhere else and possibly provide a custom Logger instance with a
+# method (e.g. .trace()) that uses this level.
+LOG_TRACE = logging.DEBUG - 1
```
cburgdorf: We do have a `TraceLogger` that defines log level `5` for tracing

https://github.com/ethereum/py-evm/blob/6708b9eedde39b3b8d4c0db57a7bd2b1770c72d9/evm/utils/logging.py#L8-L14

So, I think we should be able to just use `logger.trace(...)` where needed. In fact, the `computation` is already doing that.

https://github.com/ethereum/py-evm/blob/6708b9eedde39b3b8d4c0db57a7bd2b1770c72d9/evm/vm/computation.py#L543-L548

evm/db/account.py
```diff
@@ -403,7 +404,7 @@ def commit(self, changeset: Tuple[UUID, UUID]) -> None:
         self._journaltrie.commit(trie_changeset)
 
     def make_state_root(self) -> Hash32:
-        self.logger.debug("Generating AccountDB trie")
+        self.logger.log(LOG_TRACE, "Generating AccountDB trie")
```
cburgdorf: See, my comment above. I think that can simple be `self.logger.trace(...)` as we do it in the `computation` class.

evm/db/account.py
```diff
@@ -424,7 +425,8 @@ def _log_pending_accounts(self) -> None:
                 else:
                     accounts_displayed.add(address)
                     account = self._get_account(address)
-                    self.logger.debug(
+                    self.logger.log(
+                        LOG_TRACE,
```
cburgdorf: same here

evm/vm/computation.py
```diff
@@ -228,7 +229,8 @@ def extend_memory(self, start_position: int, size: int) -> None:
         before_cost = memory_gas_cost(before_size)
         after_cost = memory_gas_cost(after_size)
 
-        self.logger.debug(
+        self.logger.log(
+            LOG_TRACE,
```
cburgdorf: Same here, we are using `logger.trace(...)` in that same file already. Can't we use that?

gsalgado:
I didn't know about `TraceLogger`! Thanks for pointing it out, @cburgdorf. Updated to use it

evm/db/chain.py
```diff
@@ -606,6 +610,12 @@ def exists(self, key: bytes) -> bool:
         """
         return self.db.exists(key)
 
+    def get(self, key: bytes) -> bytes:
+        """
+        Returns the value for the given key or a KeyError if it doesn't exist in the database.
```
cburgdorf: nitpick: we are trying to make all these in imperative present tense:

>Return the value for the given key or a KeyError if it doesn't exist in the database.

cburgdorf:
@pipermerriam even better, since that typo made that `ignore` useless, I guess it can be removed entirely as it seems we don't need to `ignore` there.

PR #980, opened 2018-07-02
Use AsyncHeaderDB in BaseHeaderChainSyncer

gsalgado:
This is necessary because BaseHeaderChainSyncer must depend on a DB containing only headers, so that we can use it for LightChainSyncer.

The other syncers (FastChainSyncer/RegularChainSyncer) still use an AsyncChainDB

Depends on #981


p2p/__init__.py
```diff
@@ -0,0 +1,2 @@
+# This is to ensure we call setup_trace_logging() before anything else.
+import evm  # noqa: F401
```
pipermerriam: Just pointing out that this won't work as a long term solution as when `p2p` is extracted from his codebase it won't have access to the `evm` module.  Maybe we should package up our trace logging facilities in a small re-usable package.

p2p/chain.py
```diff
@@ -67,12 +72,12 @@ class BaseHeaderChainSyncer(BaseService, PeerPoolSubscriber):
...
     def __init__(self,
                  chain: AsyncChain,
-                 chaindb: AsyncChainDB,
+                 db: 'AsyncHeaderDB',
                  peer_pool: PeerPool,
                  token: CancelToken = None) -> None:
         super().__init__(token)
         self.chain = chain
-        self.chaindb = chaindb
+        self.db = db
```
pipermerriam: This rename looks like it may cause confusion.  Our current naming conventions have been

```
- `db` for `BaseDB` type objects
- `chaindb` for `ChainDB` type objects
- `headerdb` for `HeaderDB` type objects.
```

If I saw this I'd assume that `db` was a `BaseDB` instance.

pipermerriam:
Can you rebase now that https://github.com/ethereum/py-evm/pull/976 is merged.

gsalgado:
Rebased, but it still includes the commits from #981 so it's probably best to review just https://github.com/ethereum/py-evm/pull/980/commits/6ce8261fc644a229cd5aa83962f9f9822458c1d1

p2p/__init__.py
gsalgado: Maybe `eth_utils`?

PR #981, opened 2018-07-02
ChainDB now inherits from HeaderDB to avoid duplication

gsalgado:
ChainDB unnecessarily duplicated all methods from HeaderDB; by inheriting we avoid that duplication. More importantly, it allows us to use a HeaderDB in LightChainSyncer and a ChainDB in FastChainSyncer/RegularChainSyncer (#980)

Depends on #976


pipermerriam:
Can you rebase now that https://github.com/ethereum/py-evm/pull/976 is merged.

gsalgado:
Rebased

PR #984, opened 2018-07-03
Christoph/docs/api docs more more

cburgdorf:
### What was wrong?

Docs need :heart: 

### How was it fixed?

- fixed some small inconsistencies
- moved the *forks* in the docs to follow the source hierarchy
- Added lots of new API docs
- Changed settings to also include members that (currently) do not have a docstring (not to be confused with private members which continue to stay out of the docs)

Let me elaborate on the last item:

While we aim for great docstrings, hiding methods entirely because they have no doc string can be very confusing, especially in extrem cases where classes may otherwise be displayed without any members at all.

Displaying members even if they do not contain a nice description (so far) is also a great way to find public members that are missing a docstring or such that should rather be private.

#### Cute Animal Picture


pipermerriam:
Impatiently went ahead and merged cause I want to see these live!

PR #987, opened 2018-07-04
StateDownloader now handles data requests from peers

gsalgado:
I noticed a low peer count during a state sync (which makes it quite slow) and it turns out that's because peers attempt to fetch data from us and disconnect as we fail to reply. This should ensure we keep a high peer count, speeding up the state sync significantly

Depends on #980

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -564,54 +564,16 @@ class RegularChainSyncer(FastChainSyncer):
         elif isinstance(cmd, eth.GetBlockHeaders):
             await self._handle_get_block_headers(peer, cast(Dict[str, Any], msg))
         elif isinstance(cmd, eth.GetBlockBodies):
-            await self._handle_get_block_bodies(peer, cast(List[Hash32], msg))
+            await handle_get_block_bodies(
+                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
         elif isinstance(cmd, eth.GetReceipts):
-            await self._handle_get_receipts(peer, cast(List[Hash32], msg))
+            await handle_get_receipts(
+                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
         elif isinstance(cmd, eth.GetNodeData):
-            await self._handle_get_node_data(peer, cast(List[Hash32], msg))
+            await handle_get_node_data(
+                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
         else:
             self.logger.debug("%s msg not handled yet, need to be implemented", cmd)
-
-    async def _handle_get_block_bodies(self, peer: ETHPeer, msg: List[Hash32]) -> None:
-        bodies = []
-        # Only serve up to eth.MAX_BODIES_FETCH items in every request.
-        hashes = msg[:eth.MAX_BODIES_FETCH]
-        for block_hash in hashes:
-            try:
-                header = await self.wait(self.db.coro_get_block_header_by_hash(block_hash))
-            except HeaderNotFound:
-                self.logger.debug("%s asked for block we don't have: %s", peer, block_hash)
-                continue
-            transactions = await self.wait(
-                self.db.coro_get_block_transactions(header, BaseTransactionFields))
-            uncles = await self.wait(self.db.coro_get_block_uncles(header.uncles_hash))
-            bodies.append(BlockBody(transactions, uncles))
-        peer.sub_proto.send_block_bodies(bodies)
-
-    async def _handle_get_receipts(self, peer: ETHPeer, msg: List[Hash32]) -> None:
-        receipts = []
-        # Only serve up to eth.MAX_RECEIPTS_FETCH items in every request.
-        hashes = msg[:eth.MAX_RECEIPTS_FETCH]
-        for block_hash in hashes:
-            try:
-                header = await self.wait(self.db.coro_get_block_header_by_hash(block_hash))
-            except HeaderNotFound:
-                self.logger.debug(
-                    "%s asked receipts for block we don't have: %s", peer, block_hash)
-                continue
-            receipts.append(await self.wait(self.db.coro_get_receipts(header, Receipt)))
-        peer.sub_proto.send_receipts(receipts)
-
-    async def _handle_get_node_data(self, peer: ETHPeer, msg: List[Hash32]) -> None:
-        nodes = []
-        for node_hash in msg:
-            try:
-                node = await self.db.coro_get(node_hash)
-            except KeyError:
-                self.logger.debug("%s asked for a trie node we don't have: %s", peer, node_hash)
-                continue
-            nodes.append(node)
-        peer.sub_proto.send_node_data(nodes)
 
     async def _process_headers(
             self, peer: HeaderRequestingPeer, headers: Tuple[BlockHeader, ...]) -> int:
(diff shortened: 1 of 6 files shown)


pipermerriam:
The other two PRs this is based on have been 👍

I'll review this once those are merged and this has been rebased (hard to parse as is)

gsalgado:
Rebased

p2p/chain.py
```diff
@@ -741,6 +633,144 @@ def _is_receipts_empty(header: BlockHeader) -> bool:
...
+        block_number = header.block_number
+    elif isinstance(block_number_or_hash, int):
+        block_number = block_number_or_hash
+    else:
+        raise TypeError(
+            "Unexpected type for 'block_number_or_hash': %s",
+            type(block_number_or_hash),
+        )
+
+    limit = max(max_headers, eth.MAX_HEADERS_FETCH)
```
pipermerriam: What do you think about splitting this part of the body (the pure part) into a stand alone utility function for which we can write some simple tests.

p2p/chain.py
```diff
@@ -589,19 +529,9 @@ def request_receipts(self, target_td: int, headers: List[BlockHeader]) -> int:
...
-            self.logger.debug(
-                "Peer %r requested starting header %r that is unavailable, returning nothing",
-                peer,
-                header_request['block_number_or_hash'],
-            )
-            block_numbers = []
-
-        headers = [header async for header in self._generate_available_headers(block_numbers)]
+        headers = await lookup_headers(
+            self.db, header_request['block_number_or_hash'], header_request['max_headers'],
```
pipermerriam: These call signatures are quite large.  What do you think about something like the following.

```
class PeerRequestHandler:
    def __init__(self, db, logger, cancel_token):
        self.db = db
        ...

    def lookup_headers(self, ...):
        return lookup_headers(db=self.db, ..., logger=self.logger, token=self.cancel_token)

    ...
```

Then the `BaseChainSyncer` classes can just set an instance of this class as a local variable and the call turns into `headers = await self.handler.lookup_headers(bn_or_hash, max_headers, skip, reverse)` and eliminates the extra 3 arguments that are always the same.

p2p/chain.py
```diff
@@ -625,55 +555,17 @@ class RegularChainSyncer(FastChainSyncer):
         elif isinstance(cmd, eth.GetBlockHeaders):
             await self._handle_get_block_headers(peer, cast(Dict[str, Any], msg))
         elif isinstance(cmd, eth.GetBlockBodies):
-            await self._handle_get_block_bodies(peer, cast(List[Hash32], msg))
+            await handle_get_block_bodies(
+                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
         elif isinstance(cmd, eth.GetReceipts):
-            await self._handle_get_receipts(peer, cast(List[Hash32], msg))
+            await handle_get_receipts(
```
pipermerriam: In conjunction with my previous comment, these would now be `await self.handler.get_receipts(peer, ...)`

p2p/chain.py
```diff
@@ -741,6 +633,144 @@ def _is_receipts_empty(header: BlockHeader) -> bool:
...
+        receipts.append(block_receipts)
+    peer.sub_proto.send_receipts(receipts)
+
+
+async def handle_get_node_data(
+        chaindb: 'AsyncChainDB', peer: ETHPeer, node_hashes: List[Hash32],
+        logger: logging.Logger, token: CancelToken) -> None:
+    nodes = []
+    # Only serve up to eth.MAX_STATE_FETCH items in every request.
+    for node_hash in node_hashes[:eth.MAX_STATE_FETCH]:
```
pipermerriam: My gut says that the truncation of which hashes we retrieve should happen at a higher level where this function is actually called, and that this function should blindly return all of the data that was requested.

I think this is a pretty minor thing, but architecturally, I think it is more correct.

- simplifies this function to be less complex.
- separates different classes of business logic.
- allows this function to be used for larger retrieval sizes (if that ever becomes a requirement)

gsalgado: That makes sense, but this is where we call the sub-proto method that actually sends the data, so would need to extract this part into a separate unit that just returns the trie nodes. It'd also have to become a generator otherwise malicious peers could easily DoS us as we'd be always retrieving all the data from the DB, regardless of the limit.

How would you feel about just adding a new `limit` argument to the `handle_*()` methods for now, and having them default to the current values we use? That way we don't need to break these into even smaller methods (which is not necessary yet), but we keep them flexible and easy to refactor should it become necessary

p2p/chain.py
gsalgado: I'd prefer to save that for another PR, as I'm just shuffling this code around in this one

p2p/chain.py
gsalgado: Yeah, that's a nice improvement

p2p/chain.py
pipermerriam: My intention was not to truncate the return value, but to truncate `node_hashes` before it was passed into this function.

```python
# at the call-site
trie_data = self._handler.handle_get_node_data(..., node_hashes=requested_hashes[:MAX_STATE_FETCH])
```

I think I prefer this over the `limit` argument, but both would be fine.

p2p/chain.py
```diff
@@ -720,6 +610,138 @@ class RegularChainSyncer(FastChainSyncer):
...
+    def __init__(self, db: 'AsyncHeaderDB', logger: TraceLogger, token: CancelToken) -> None:
+        self.db = db
+        self.logger = logger
+        self.cancel_token = token
+
+    async def handle_get_block_bodies(self, peer: ETHPeer, block_hashes: List[Hash32]) -> None:
+        chaindb = cast('AsyncChainDB', self.db)
+        bodies = []
+        # Only serve up to eth.MAX_BODIES_FETCH items in every request.
+        for block_hash in block_hashes[:eth.MAX_BODIES_FETCH]:
```
pipermerriam: Same thought here about moving the limits up a level to the call site of this function.

```python
bodies = self._handler.handle_get_block_bodies(peer, requested_block_hashes[:MAX_BODIES_FETCH])
```

p2p/chain.py
```diff
@@ -720,6 +610,138 @@ class RegularChainSyncer(FastChainSyncer):
...
 
 
+class PeerRequestHandler:
+
+    def __init__(self, db: 'AsyncHeaderDB', logger: TraceLogger, token: CancelToken) -> None:
+        self.db = db
+        self.logger = logger
+        self.cancel_token = token
+
+    async def handle_get_block_bodies(self, peer: ETHPeer, block_hashes: List[Hash32]) -> None:
```
pipermerriam: Since the class has the name `Handler` what do you think about dropping the `handle_` prefix for these methods?

p2p/chain.py
```diff
@@ -720,6 +610,138 @@ class RegularChainSyncer(FastChainSyncer):
...
+        chaindb = cast('AsyncChainDB', self.db)
+        bodies = []
+        # Only serve up to eth.MAX_BODIES_FETCH items in every request.
+        for block_hash in block_hashes[:eth.MAX_BODIES_FETCH]:
+            try:
+                header = await wait_with_token(
```
pipermerriam: Maybe put the `self.wait(...)` API on this class.  Maybe we can introduce that as a mixin class that 

```python
class Cancelable:
    cancel_token: CancelToken = None

    def wait(...):
        ...

    def wait_first(...):
        ...
```

Or Just inline a duplication of the `wait` function from our `Service` class with a comment/issue to remove the duplication at some point.

p2p/chain.py
```diff
@@ -720,6 +610,138 @@ class RegularChainSyncer(FastChainSyncer):
...
+                "Peer requested starting header %r that is unavailable, returning nothing",
+                block_number_or_hash)
+            block_numbers = []
+
+        headers = [header async for header in self._generate_available_headers(block_numbers)]
+        return headers
+
+    async def _get_block_numbers_for_request(
+            self, block_number_or_hash: Union[int, bytes], max_headers: int,
+            skip: int, reverse: bool) -> List[BlockNumber]:
```
pipermerriam: Mostly unrelated.  I think go-ethereum recently patched a bug which exploited overflowing their integer with very large `skip` values.  While we don't suffer from the same overflow problems, it makes me think that we should enforce a reasonable upper bound on the skip size.

p2p/chain.py
gsalgado: Well, if I drop the `handle_` prefix the names will kinda suggest the methods are getters (e.g. `get_block_bodies()`), so I'd rather keep the prefix

p2p/chain.py
gsalgado: Oh, I see! Went with your proposed solution

p2p/chain.py
gsalgado: I like the mixin idea! Just thought I'd add a `Mixin` suffix to its name to make that clear, though

PR #990, opened 2018-07-04
Don't try to call method on non-existing discovery

cburgdorf:
### What was wrong?

If we try to exit `trinity` before the `discovery` exists we are running into the following exception

```
   ERROR  07-04 14:20:20     service  Unexpected error in <trinity.nodes.mainnet.MainnetFullNode object at 0x7fe22f4e8eb8>, exiting
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/p2p/service.py", line 82, in run
    await self._run()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/nodes/base.py", line 128, in _run
    await self.get_p2p_server().run()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/p2p/service.py", line 92, in run
    await self.cleanup()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/p2p/service.py", line 99, in cleanup
    await self._cleanup()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/p2p/server.py", line 183, in _cleanup
    await asyncio.gather(self.peer_pool.cleaned_up.wait(), self.discovery.cleaned_up.wait())
AttributeError: 'Server' object has no attribute 'discovery'
```

Important part:

```
await asyncio.gather(self.peer_pool.cleaned_up.wait(), self.discovery.cleaned_up.wait())
AttributeError: 'Server' object has no attribute 'discovery'
```

### How was it fixed?

Ensure that we only try to call `self.discovery.cleaned_up.wait()` is `self.discovery` is not `None`

#### Cute Animal Picture

### DIFF
--- a/p2p/server.py
+++ b/p2p/server.py
@@ -72,6 +72,7 @@ class Server(BaseService):
     _udp_listener = None
 
     peer_pool: PeerPool = None
+    discovery: DiscoveryService = None
 
     def __init__(self,
                  privkey: datatypes.PrivateKey,
@@ -180,7 +181,12 @@ class Server(BaseService):
 
     async def _cleanup(self) -> None:
         self.logger.info("Closing server...")
-        await asyncio.gather(self.peer_pool.cleaned_up.wait(), self.discovery.cleaned_up.wait())
+        cleanup_tasks = [self.peer_pool.cleaned_up.wait()]
+
+        if self.discovery is not None:
+            cleanup_tasks.append(self.discovery.cleaned_up.wait())
+
+        await asyncio.gather(*cleanup_tasks)
         await self._close()
 
     async def receive_handshake(


p2p/server.py
```diff
@@ -180,7 +181,12 @@ def _make_peer_pool(self) -> PeerPool:
 
     async def _cleanup(self) -> None:
         self.logger.info("Closing server...")
-        await asyncio.gather(self.peer_pool.cleaned_up.wait(), self.discovery.cleaned_up.wait())
+        cleanup_tasks = [self.peer_pool.cleaned_up.wait()]
```
cburgdorf: I tried to avoid the list and wanted to use a tuple and then just do `cleanup_tasks += (self.discovery.cleaned_up.wait())` but hell broke loose once I tried that.

p2p/server.py
```diff
@@ -72,6 +72,7 @@ class Server(BaseService):
     _udp_listener = None
 
     peer_pool: PeerPool = None
+    discovery: DiscoveryProtocol = None
```
gsalgado: This is a `DiscoveryService`, I think

gsalgado:
I've been meaning to add a new method to `p2p.BaseService` that takes another `BaseService` instance, runs it in the background (with `asyncio.ensure_future()`) and ensures we wait for it to finish upon exit. If you'd like to take a stab at it, it'd solve this as well, and in a more elegant/generic way. Otherwise we can go with what you've got here until I get around to add the new API

p2p/server.py
cburgdorf: yep

cburgdorf:
@gsalgado ah so you'r thinking that instead of 

```python
        await self._start_udp_listener(discovery_proto)
        self.discovery = DiscoveryService(discovery_proto, self.peer_pool, self.cancel_token)
        asyncio.ensure_future(self.peer_pool.run())
        asyncio.ensure_future(self.discovery.run())
        asyncio.ensure_future(self.upnp_service.run())
```

we would have something like

```python
        await self._start_udp_listener(discovery_proto)
        self.discovery = DiscoveryService(discovery_proto, self.peer_pool, self.cancel_token)
        self.schedule_child(self.peer_pool)
        self.schedule_child(self.discovery)
        self.schedule_child(self.upnp_service)
```

And then on cleanup we would wait for all child services to be finished implicitly? Makes sense to me. I can take a look.

gsalgado:
Yep, that's what I had in mind, but I think `run_service_in_background()` would be a better name for such a method

cburgdorf:
Closing in favor for #991

Issue #401, opened 2018-02-23
Add tx.hash global variable


k06a:
### What is wrong?

Sometimes it is needed to store hash of the transaction in blockchain. I am sure it is possible to make `tx.hash` available in EVM.

### How can it be fixed

Add `tx.hash` support to EVM

pipermerriam:
@k06a this isn't the correct place for this type of request.

Changes to the core protocol are done via the EIP process here: https://github.com/ethereum/EIPs

I believe this functionality is already part of the sharding roadmap.

k06a:
@pipermerriam thanks: https://github.com/ethereum/EIPs/issues/901

PR #991, opened 2018-07-04
Don't try to call method on non-existing discovery

cburgdorf:
### What was wrong?

Alternative to #990 

### How was it fixed?

This introduces a new method `schedule_child_service` on the `BaseService` which will schedule other `BaseService`' to get executed together with the main service.

The advantage of that is, that we can wait for all child services to properly cleanup without explicitly specifying that.

#### Cute Animal Picture


p2p/service.py
```diff
@@ -94,8 +97,21 @@ def __init__(self, token: CancelToken=None) -> None:
...
+        self._child_services.append(child_service)
+
+    def _run_child_services(self) -> None:
+        for service in self._child_services:
+            asyncio.ensure_future(service.run())
+
     async def cleanup(self) -> None:
         """Run the service's _cleanup() coroutine and sets the cleaned_up event."""
+
+        await asyncio.gather(*[svc.cleaned_up.wait() for svc in self._child_services])
```
cburgdorf: @gsalgado do you think it is important that we wait on the child service cleanup first or could we do that in parallel as in:

```python
await asyncio.gather(*[svc.cleaned_up.wait() for svc in self._child_services], self._cleanup())
```

p2p/service.py
```diff
@@ -94,8 +97,21 @@ def __init__(self, token: CancelToken=None) -> None:
             if finished_callback is not None:
                 finished_callback(self)
 
+    def schedule_child_service(self, child_service: 'BaseService') -> None:
+        """
+        Schedule a child service to run with the main service. On shutdown the main service
```
cburgdorf: @gsalgado @pipermerriam let's do some bike shedding over the name.

The reason why I'd like to have `child` in their is because it implies that hierachy that we are essentially creating (child services run when main service runs, main service waits for child services to cleanup).

The reason why I use `schedule` rather than `run` is because this method really doesn't run the services yet but just puts them in a list. They are all run after the `_run()` method of the main service. 

Ok, but now that I'm thinking about that...*maybe* we don't want to run them immediately because if we run it directly we could also start such child services after `_run()` was called. That's not possible with the current design

How about I change that and then we call the method `run_child_service`?

p2p/service.py
gsalgado: I think in parallel is fine as long as it is clearly documented that we don't wait for them in any specific order

p2p/service.py
```diff
@@ -15,8 +17,15 @@
 from p2p.exceptions import OperationCancelled
 
 
+if TYPE_CHECKING:
```
pipermerriam: Can we drop the `if TYPE_CHECKING` since this won't cause a circular import and we can just mark it with the `noqa: #401` to prevent linting errors?

p2p/service.py
```diff
@@ -94,9 +103,27 @@ def __init__(self, token: CancelToken=None) -> None:
...
+        """
+        Run the ``_cleanup()`` coroutine and set the ``cleaned_up`` event after the service as
+        well as all child services finished their cleanup.
+
+        The ``_cleanup()`` coroutine is invoked before the child services may have finished
+        their cleanup.
+        """
+
+        await asyncio.gather(*[
+            svc.cleaned_up.wait()
```
pipermerriam: Thoughts on just `svc -> child_service`?  (I have grown to dislike most abbreviations)

p2p/service.py
cburgdorf: true that!

p2p/service.py
cburgdorf: Fine by me!

PR #993, opened 2018-07-05
FastChainSyncer now handles data requests from peers

gsalgado:
It used to handle only GetBlockHeader requests, and now handles the rest as well

Includes commits from other PRs, so please review just https://github.com/ethereum/py-evm/pull/993/commits/f80d8ba0708137fee153b860f2b94fb75b17b267

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -289,6 +289,7 @@ class LightChainSyncer(BaseHeaderChainSyncer):
         query = msg['query']
         headers = await self._handler.lookup_headers(
             query.block_number_or_hash, query.max_headers, query.skip, query.reverse)
+        self.logger.trace("Replying to %s with %d headers", peer, len(headers))
         peer.sub_proto.send_block_headers(headers, buffer_value=0, request_id=msg['request_id'])
 
     async def _process_headers(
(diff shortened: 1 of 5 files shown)


p2p/chain.py
```diff
@@ -480,8 +482,24 @@ def request_receipts(self, target_td: int, headers: List[BlockHeader]) -> int:
...
+        elif isinstance(cmd, eth.GetNodeData):
+            await handle_get_node_data(
+                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
+        elif isinstance(cmd, eth.Transactions):
+            # Transactions msgs are handled by our TxPool service.
+            pass
+        elif isinstance(cmd, eth.NodeData):
+            # When doing a chain sync we never send GetNodeData requests, so peers should not send
+            # us NodeData msgs.
+            self.logger.warn("Unexpected NodeData msg from %s", peer)
```
pipermerriam: Should we also disconnect from the peer?  Otherwise a malicious peer could flood the terminal with these messages.

p2p/chain.py
```diff
@@ -543,28 +561,10 @@ class RegularChainSyncer(FastChainSyncer):
...
-                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
-        elif isinstance(cmd, eth.GetNodeData):
-            await handle_get_node_data(
-                self.db, peer, cast(List[Hash32], msg), self.logger, self.cancel_token)
-        else:
-            self.logger.debug("%s msg not handled yet, need to be implemented", cmd)
+    async def _handle_block_receipts(
+            self, peer: ETHPeer, receipts_by_block: List[List[eth.Receipt]]) -> None:
+        # When doing a regular sync we never request receipts.
+        self.logger.warn("Unexpected BlockReceipts msg from %s", peer)
```
pipermerriam: Any reason this was not inlined in the `if/elif` block on lines 485-500 the same way that the `NodeData` warning is inlined.  Similarly, should we disconnect in this case?

gsalgado: lines 485-500 are in `FastChainSyncer`, and during a fast sync we send `GetReceipts` requests and expect `Receipts` responses. This is `RegularChainSyncer`, and during a regular (i.e. block-processing) sync we don't send `GetReceipts`. I guess it'd make sense to disconnect as well, yeah

gsalgado: `NodeData`, on the other hand, is unexpected in both syncers -- it is only the `StateDownloader` that requests those

pipermerriam: I mis-read the diff and didn't see that those were two different classes.

p2p/peer.py
```diff
@@ -485,6 +488,8 @@ def disconnect(self, reason: DisconnectReason) -> None:
         self.logger.debug("Disconnecting from remote peer; reason: %s", reason.name)
         self.base_protocol.send_disconnect(reason.value)
         self.close()
+        if self.is_running:
+            await self.cancel()
```
pipermerriam: cc @carver as you added a `wait_disconnect` method in your other PR and I think this is a better pattern.

PR #994, opened 2018-07-05
ChainDB.persist_block() now only persists uncles when not empty

gsalgado:


evm/db/chain.py
```diff
@@ -268,9 +272,14 @@ def persist_block(self, block: 'BaseBlock') -> None:
             for index, transaction_hash in enumerate(self.get_block_transaction_hashes(header)):
                 self._add_transaction_to_canonical_chain(transaction_hash, header, index)
 
-        if hasattr(block, "uncles"):
```
gsalgado: All our `BaseBlock` implementations have a `uncles` attribute, so this would always return True

Issue #529, opened 2018-04-04
ChainSyncer: add tx lookup entries to DB


gsalgado:
### What is wrong?

Currently (when doing a fast sync) it generates the transactions trie manually (using `make_trie_root_and_nodes()`) and persists that to the DB, so we don't get the DB entries that allow us to lookup the block of a given tx by hash.

### How can it be fixed

We could redesign `ChainSyncer` so that it uses `ChainDB.persist_block()`, but that would certainly affect performance significantly because that trie generation would be done in the DB process and that is quite expensive.

Another option would be to duplicate (or refactor) parts of `persist_block()` so that `ChainSyncer` calls `ChainDB._add_transaction_to_canonical_chain()` after it calls `ChainDB.persist_header()`

pipermerriam:
linking to #507 as I think this topic might be tangentially related.

The DB process could be modified to only expose the base level DB connection which would allow the `ChainSyncer` to use it's own adapter or re-use an existing adapter while still giving us control over where the actual processing happens.

gsalgado:
Actually, `persist_block()` only generates the trie to validate the transaction_root. If we [move that to a validation method](https://github.com/ethereum/py-evm/pull/533) we should be able to use `persist_block()` in ChainSyncer without too much of a performance penalty

PR #995, opened 2018-07-05
Use persist_block() instead of persist_header() in FastChainSyncer

gsalgado:
That ensures we add lookup entries from TX hash to block/idx.

Closes: #529


p2p/chain.py
```diff
@@ -423,10 +423,17 @@ def __init__(self,
             'receipt')
         self.logger.debug("Got block receipts for chain segment")
 
-        # FIXME: Get the bodies returned by self._download_block_parts above and use persit_block
-        # here.
         for header in headers:
-            await self.wait(self.db.coro_persist_header(header))
+            uncles: Tuple[BlockHeader, ...] = tuple()
```
pipermerriam: *nitpick* I'd rather this be in an `else` block of the following `if` statement than to overwrite the local variable.

Issue #979, opened 2018-06-30
Plugin architecture


pipermerriam:
### What is the big idea!?

I have been thinking about a new API for *plugins* in the trinity client.

Plugins could range from things like simple internal optional components such as a gas estimation engine up to full blown applications like daemon processes for something like augur or ethereum alarm clock.

I also think that we might benefit from dog-fooding this API for many of our internal things such as the JSON-RPC server, the transaction pool, peer management for rating and blocking peers.

Here is a quick list of things that I think fit well into a *plugin* type architecture.

1. Application which manages paid slots for light peers
2. Transaction Pool
3. Rating peers
4. Blocking peers
5. Whisper
6. Deamon processes (augur, ethereum alarm clock, truebit)
7. The python console.

APIs that would need to be exposed for these things would be:

- hooks to add new sections of CLI arguments 
- hooks to add new CLI commands.
- high level Web3.py connection
- low level chain and database access
- hooks for interacting with core components like peer pool
- hooks for launching JSON-RPC server namespaces.

We might even be able to go as far as proving the ability to run new DevP2P services which would allow running something like whisper.

And finally, while this is the absolute extreme version of this idea, we may even be able to move syncing to be relegated to run within this system.

### How can it be fixed

This probably starts with a POC for either the transaction pool or the planned system for rating and persisting information about peers.

cburgdorf:
There also needs to be a generic way for plugins to raise events that other parts of the system (either core stuff or other plugins) can react upon. 

One idea could be to have a generic *event bus* in Trinity that is exposed to the plugins and where plugins can listen for events but can also raise events. An event bus has been my standard pattern for this kind of problem in the past (which is why I'm bringing it up here) but there may be other well suited patterns as well.

cburgdorf:
Just leaving this thought here: I think that the nodes itself (full (ropsten/mainnet), light (ropsten / mainnet)) are pretty good candidates to fall into the plugin space themselves. We can probably make these things more fine grained and then provide plugins to expose the same functionality as today but with the added benefit to make it really easy for people to e.g. add support for *Kovan* or whatever.

cburgdorf:
Guess we can close. Feel free to re-open if anyone disagrees.

PR #996, opened 2018-07-05
Provide a rudimentary plugin architecture and make the TxPool a plugin

cburgdorf:
### What was wrong?

As we aim for Trinity to be highly flexible we've identified that having a plugin architecture will likely be our best way forward to achieve this. Read #979 for more info.

### How was it fixed?

This brings in a first rudimentary plugin system. The core parts are:

**`PluginManager`**

Responsible for the overall management of the plugins.

**`BasePlugin`**

Abstract base class that plugins need to inherit from.

**`BaseActivationEvent`**
Abstract base class for all activation events. Explained further down.

**FAQ**:

**How are plugins registered / discovered?**

Currently, there's a hardcoded list of enabled plugins in `plugins.registry` that is feed to `PluginManager.register_plugins(...)`. However, the idea would be to have some sort of auto-discovery that works based on some convention / configuration scheme.

**How get plugins activated?**

Plugins can respond to certain types of *activation events*. The concept is inspired by Visual Studio code which does the same thing (certainly much much different though). The idea is that plugins can activate based on a rich set of events. For instance:

- some CLI flag was passed
- some resource became available (e.g. `PeerPool`)
- some other plugin was activated

Activation events can be anything derived from `BaseActivationEvent`. Existing activation events are:

- `TrinityStartupActivationEvent`
- `PluginActivatedActivationEvent`
- `ResourceAvailableActivationEvent`

That said, in the current model, plugins often may only want to get activated after a series of *activation events* have happened (e.g. application startup with a certain flag + `PeerPool` became activate). Currently, plugins need to keep track of these events themselves and activate as soon as all criteria match. There's probably room to improve this.

**What can a plugin do?**

Apart from being able to run an activation function with potentially any kind of side effects, a plugin can also hook into the `ArgumentParser` to add to or mess with the existing CLI commands and arguments.

This is achieved by overwriting `def configure_parser(self, arg_parser: ArgumentParser) -> None:...` in the plugin and performing whatever action is needed on `arg_parser`

### Caveats

- The `PluginManager` does currently live **inside** the networking process. It should probably live in its own process and allow plugins to specify if they want to run in a shared plugin process or in their own dedicated process. It also needs to provide the right facilities for inter plugin communication then.

- plugins are still hardcoded in a `ENABLED_PLUGINS` list. There's no auto-discovery based on some convention / configuration scheme.

- it may be easy to overlook if a plugin does not activate because its requirements were not fulfilled. The plugin manager prints an info whenever a plugin gets activated but it may be easily overlooked if a plugin **does not get activated**. One reason to deal with this could be to let plugins specify a timeout and if they did not activate within the specified timeout it could throw an exception.

## Making the TxPool a plugin

To make all of this a little more useful this PR dogfoods the new plugin architecture by wrapping the `TxPool` as a plugin and moving it into the new top level `plugins` module.

This also includes that the transaction pool can now be disabled by providing a `--disable-tx-pool` flag on the command line. This is partly to show case the new plugin architecture but also because I think it may prove useful in case we notice that the transaction pool isn't that stable / useful yet and people may want to disable it entirely.

#### Cute Animal Picture


cburgdorf:
Just wanted to add that the proposed system is likely enough to also move the `console` and `attach` commands into the plugin space making our `main` slightly cleaner. I'm giving that a try in a follow up PR.

pipermerriam:
Looking at the class names:

- `TrinityStartupActivationEvent`
- `PluginActivatedActivationEvent`
- `ResourceAvailableActivationEvent`

Seems like we could drop the `Activation` and just refer to them as events?

- `TrinityStartupEvent`
- `PluginActivatedEvent`
- `ResourceAvailableEvent`

cburgdorf:
>Seems like we could drop the `Activation` and just refer to them as events?

Yep, I also began questioning the super long event names. SGTM :+1: 

I also agree with your comment regarding moving the plugins.

What's your thought regarding the overall approach? Did you also see the follow up PR that makes the `console` / `attach` command a plugin (still has failing tests but works in general).

I think the main thing to sort out next are the multi process aspects of this. My current thinking is that the plugin manager should live in its own process and that plugins should be able to decide whether they want to run in that *shared plugin process* or run in their own dedicated process. But that will have further implications which still need to be sorted out.

I'm mostly interested in you signaling if you are ok with the current start and idea how to move further or if it is not going into the direction that you envisioned.

plugins/tx_pool/plugin.py
```diff
@@ -0,0 +1,90 @@
...
+            help="Disable the Transaction Pool",
+        )
+
+    def should_activate(self, activation_event: BaseActivationEvent) -> bool:
+        if isinstance(activation_event, TrinityStartupActivationEvent):
+            self.is_enabled = not activation_event.args.disable_tx_pool
+        if isinstance(activation_event, ResourceAvailableActivationEvent):
+
+            if activation_event.resource_type is PeerPool:
+                self.peer_pool, self.cancel_token = activation_event.resource
```
pipermerriam: I find it very odd that `should_activate` isn't a pure function, largely based on the name.  I don't love that it has side effects.

Maybe this should be split into two things.

1. `BasePlugin.handle_event`
2. `BasePlugin.should_activate`

That allows for segregation of the mutative code from the logic of whether the plugin should be activated.

pipermerriam:
> Did you also see the follow up PR that makes the `console` / `attach` command a plugin

❤️ 

> I'm mostly interested in you signaling if you are ok with the current start and idea how to move further or if it is not going into the direction that you envisioned.

👍 (I like the direction this is going)

> I think the main thing to sort out next are the multi process aspects of this.

I think this is related to the async comment I made about expecting the plugins to have more `async def` methods.  My thought on this is roughly as follows.

- Plugins should have at least one method which is expected to be long running like `BasePlugin.start`
- Maybe plugins should be a subclass of `p2p.service.Service`
- Plugins which want to run their own processes can spawn them within the `_run` method and handle shutting the process down as part of the cancel token handling.
- Plugins which are *light* and can run on their own can just implement their logic locally.
- We can probably wrap the subprocess based plugin logic up nicely in a `DedicatedProcessPlugin` base class.

trinity/extensibility/plugin_manager.py
```diff
@@ -0,0 +1,83 @@
...
+        self._active_plugins: List[BasePlugin] = []
+        self._logger = logging.getLogger("trinity.extensibility.plugin_manager.PluginManager")
+
+    def register_plugin(self, plugin: BasePlugin) -> None:
+        """
+        Register a :class:`~trinity.extensibility.plugin.BasePlugin` with the plugin manager.
+        """
+        self._plugin_store.append(plugin)
+
+    def register_plugins(self, plugins: Iterable[BasePlugin]) -> None:
```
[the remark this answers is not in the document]
cburgdorf: Yep, let's make this take a `Union[BasePlugin, Iterable[BasePlugin]` and just call it `register` then.

plugins/tx_pool/plugin.py
cburgdorf: I like your way of thinking. It didn't feel good to me to mix concerns here either and I very much welcome your suggestion to split this into `handle_event` and `should_activate` APIs.

Btw, I have a somewhat radical™ idea around how to handle these events in an entirely different way but it's probably too radical to get adopted :sweat_smile: Soon™

trinity/extensibility/activation_events.py
```diff
@@ -0,0 +1,54 @@
...
+    Namespace,
+)
+
+from trinity.config import (
+    ChainConfig,
+)
+
+
+if TYPE_CHECKING:
+    from trinity.extensibility import (  # noqa: F401
```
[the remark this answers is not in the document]
cburgdorf: It doesn't sit very well with me to have the plugins (even not the builtin ones) sit in the same directory where the infrastructure code is. Those are consumers of the extensibility infrastructure and I want to promote a clear separation as much as possible (e.g. not importing something from `extensibility.utils` in a plugin which may feel less of an issue if they all sit in a shared space).

My preference would be to keep `trinity/plugins` seperated from `trinity/extensibility` if you don't feel strongly about the merge.

cburgdorf:
>The activate verb isn't sitting well with me. I am leaning towards something like `run` or `start`. Thoughts?

I like the symmetry of `start` and `stop`. :+1: 

Regarding the async stuff, here are my current thoughts (which are already inspired by what you wrote):

Maybe we have different kind of base plugins: `BaseMainProcessPlugin` (runs in the main process such as `console` / `attach`), `BaseSharedProcessPlugin` (runs in the shared plugins process), `BaseOwnProcessPlugin` runs in their own process.

The `BaseSharedProcessPlugin` would need to provide an async API to play well in a shared process that is largely (or entirely) run in an event-looped fashion using `asyncio` APIs.

Both the `BaseSharedProcessPlugin` and the `BaseSharedProcessPlugin` would need to provide an API to create a proxy object to allow inter-process communication between plugins. 

It's still vague and needs more exploration though.

cburgdorf:
@pipermerriam I think I addressed all your points. But maybe it would be ok to land these in another PR?

trinity/extensibility/plugin_manager.py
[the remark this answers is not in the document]
pipermerriam: So callable as:

- `register(plugin)`
- `register([plugin_a, plugin_b])

and not callable as

- `register(plugin_a, plugin_b)`

Now that I'm looking at this, my feeling is that we should just make this simple and only support `register(plugin)` and if people want to register multiple, then they do it with their own loop.

trinity/extensibility/plugin_manager.py
```diff
@@ -0,0 +1,79 @@
...
+    def register(self, plugins: Union[BasePlugin, Iterable[BasePlugin]]) -> None:
+        """
+        Register one or multiple instances of :class:`~trinity.extensibility.plugin.BasePlugin`
+        with the plugin manager.
+        """
+
+        new_plugins = [plugins] if isinstance(plugins, BasePlugin) else plugins
+
+        for val in new_plugins:
+            self._plugin_store.append(val)
```
pipermerriam: loop can be replaced with `self._plugin_store.extend(new_plugins)`

This will go away completely if you choose to go with only supporting a single plugin at a time for registration.

trinity/extensibility/plugin_manager.py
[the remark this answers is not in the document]
cburgdorf: What's your reasoning? That would force more boiler plate on the consumer. Sure, it's easy to call this in a loop but it's even easier for us to eliminate that boilerplate and provide a nicer API.

trinity/main.py
```diff
@@ -250,3 +274,11 @@ def run_service_until_quit(service: BaseService) -> None:
     asyncio.ensure_future(service.run())
     loop.run_forever()
     loop.close()
+
+
+def setup_plugins() -> PluginManager:
+    plugin_manager = PluginManager()
+    # TODO: Implement auto-discovery of plugins based on some convention/configuration scheme
```
pipermerriam: I think this is what we want to do.

https://packaging.python.org/guides/creating-and-discovering-plugins/

Specifically this section

https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata

trinity/plugins/builtin/tx_pool/plugin.py
```diff
@@ -0,0 +1,87 @@
...
+        if isinstance(activation_event, ResourceAvailableEvent):
+            if activation_event.resource_type is PeerPool:
+                self.peer_pool, self.cancel_token = activation_event.resource
+            elif activation_event.resource_type is BaseChain:
+                self.chain = activation_event.resource
+
+    def should_start(self) -> bool:
+        return all(
+            val is not None and val is not False
+            for val in [self.peer_pool, self.chain, self.is_enabled]
```
pipermerriam: I think this is cleaner as:

```python
return all((self.peer_pool, self.chain, self.is_enabled))

# or
if not self.peer_pool:
    return False
elif not self.chain:
    return False
elif not self.is_enabled:
    return False
else:
    return True
```

Either would be 👍 from me.

trinity/main.py
cburgdorf: Thanks for the pointer! Will look into it.

trinity/plugins/builtin/tx_pool/plugin.py
cburgdorf: The first one looks good. I just learned that Python has the concept of Truthy/Falsey as well. Thought that's a JavaScript thing only :+1:

trinity/extensibility/plugin_manager.py
cburgdorf: Ah, nice. Let's use `extend` then and continue to support batch registration if you don't mind :unicorn:

trinity/plugins/builtin/tx_pool/plugin.py
cburgdorf: Oh turns out, `all(...)` does expect a proper `boolean` so that doesn't work. However, it can be shortened to this:

```python
        return all((self.peer_pool != None, self.chain != None, self.is_enabled))
```

carver: > all(...) does expect a proper boolean

Truth-y things work in `all()`. See:

```py
assert all([object(), 'tru', 1, True])
```

cburgdorf:
Maybe arbitrary object instances aren't truthy then? Whatever it is, it
didn't work 😅

PR #998, opened 2018-07-06
Implement missing abstract methods in LightDispatchChain

cburgdorf:
### What was wrong?

Exception due to missing methods on the `LightDispatchChain`

```
  File "/usr/lib/python3.6/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python3.6/multiprocessing/process.py", line 93, in run
    self._target(*self._args, **self._kwargs)
  File "/home/salgado/src/py-evm/trinity/utils/profiling.py", line 31, in inner
    return fn(*args, **kwargs)
  File "/home/salgado/src/py-evm/trinity/utils/logging.py", line 118, in inner
    return fn(*args, **inner_kwargs)
  File "/home/salgado/src/py-evm/trinity/main.py", line 235, in launch_node
    node = NodeClass(chain_config)
  File "/home/salgado/src/py-evm/trinity/nodes/light.py", line 54, in __init__
    self.create_and_add_tx_pool()
  File "/home/salgado/src/py-evm/trinity/nodes/base.py", line 106, in create_and_add_tx_pool
    DefaultTransactionValidator(self.get_chain(), self.initial_tx_validation_block_number),
  File "/home/salgado/src/py-evm/trinity/nodes/light.py", line 76, in get_chain
    self._chain = self.chain_class(self._headerdb, peer_chain=self._peer_chain)
TypeError: Can't instantiate abstract class RopstenLightDispatchChain with abstract methods get_vm_class, get_vm_configuration
```

### How was it fixed?

Added methods.

#### Cute Animal Picture

### DIFF
--- a/trinity/chains/light.py
+++ b/trinity/chains/light.py
@@ -1,5 +1,4 @@
 import asyncio
-import inspect
 from typing import (  # noqa: F401
     Any,
     Optional,
@@ -22,12 +21,7 @@ from eth_typing import (
 )
 
 from evm.chains.base import (
-    AccountState,
-    BaseChain,
-)
-from evm.db.backends.base import BaseDB
-from evm.db.chain import (
-    BaseChainDB,
+    Chain,
 )
 from evm.db.header import (
     BaseHeaderDB,
@@ -37,17 +31,6 @@ from evm.rlp.blocks import (
 )
 from evm.rlp.headers import (
     BlockHeader,
-    HeaderParams,
-)
-from evm.rlp.receipts import (
-    Receipt
-)
-from evm.rlp.transactions import (
-    BaseTransaction,
-    BaseUnsignedTransaction,
-)
-from evm.vm.computation import (
-    BaseComputation
 )
 
 from p2p.lightchain import (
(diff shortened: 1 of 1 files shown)


gsalgado:
Hmm, did you try running `trinity --light` after those changes? I see the TxPool validator calling `chain.get_vm_class()`, which would cause a NotImplementedError, no?

cburgdorf:
@gsalgado I admit being guilty of not doing that ;) I misunderstood were that ends up. Now that I looked into it more shouldn't that actually inherit from `Chain` rather than `BaseChain` and then just overwrite the methods that it wants to change. Otherwise we end up duplicating quite a bit of code.

cburgdorf:
@gsalgado I updated the PR with what I think is the better fix. The `LightDispatchChain` now inherits from `Chain` and overwrites only the methods that it cares to change. This seems less fragile and fixes the problem by deleting code rather than adding new one.

cburgdorf:
>provided you ran `trinity --light` this time ;-)

I did :sweat_smile: :see_no_evil:

PR #1000, opened 2018-07-09
Make trinity console / attach a plugin

cburgdorf:
### What was wrong?

This builds on #996 and makes the `trinity console` as well as the `trinity attach` command a plugin. Both commands don't need to be in the core and in fact help to make the main file less verbose.

### How was it fixed?

- Created a new `NetworkProcessReadyActivationEvent` to give plugins more fine grained control
- changed the `BasePlugin` to be able to amend sub command configs
- moved exist console logic into plugin space and wrapped it as `ConsolePlugin`

#### Cute Animal Picture


trinity/extensibility/plugin.py
```diff
@@ -30,12 +40,19 @@ def name(self) -> str:
             "Must be implemented by subclasses"
         )
 
+    @property
+    def process(self) -> PluginProcess:
```
cburgdorf: This could actually be a `classmethod`

cburgdorf:
I'm signing off until Monday. Feel free to review, merge or drive it further from here

trinity/plugins/builtin/console/plugin.py
```diff
@@ -0,0 +1,91 @@
...
+    def should_start(self):
+        return (self.is_console and self.network_process_ready) or self.is_attach
+
+    def start(self, context: PluginContext) -> None:
+        self.run_console(self.chain_config, not self.is_vanilla)
+
+        # If we run `attach` the console will block the node from booting and we want
+        # to shut it down to not continue booting when the user leaves the console
+        if self.is_attach:
+            sys.exit(0)
```
carver: It makes sense how we got here. But I'm worried that a pattern where a plugin takes over execution and then exits is a pretty special case. Maybe such a special case that it's not even really a plugin anymore, but a completely different app that's running.

I don't really have a suggestion here other than to maybe add a comment that it's unusual and should be discouraged for plugins to block indefinitely on `start()`, and also discouraged to `exit()`. If two plugins tried to do that at the same time, they would be racing to go first.

trinity/plugins/builtin/console/plugin.py
```diff
@@ -0,0 +1,91 @@
...
+    console
+)
+
+
+class ConsolePlugin(BasePlugin):
+
+    def __init__(self) -> None:
+        self.is_attach = False
+        self.is_console = False
+        self.is_vanilla = False
```
pipermerriam: I think we should drop the vanilla/ipython distinction and just make them separate plugins.  We can autodetect which plugin to load by default based on whether ipython is available or not.

trinity/plugins/builtin/console/plugin.py
pipermerriam: In the model I proposed (in another comment) we would get rid of the `start` and `should_start` mechanisms entirely in favor of the plugin having a formal route to take over the whole process. (this however requires we drop the `console` command which I'm generally in favor of because I think running both the node and the console in the same process is weird.

carver: It *is* weird. Although I do find myself doing it fairly regularly with geth. I guess it's fairly easy to simulate with something like:

```sh
trinity >>info.log &
trinity attach
pkill trinity
```

Probably the thing I would miss the most is making a node with a custom IPC location and being able to start it up with a console without having to retype the IPC location on the second command.

pipermerriam: I have found that `console` has so much logging noise that it is borderline un-usable.  IMHO it is a weird carryover from geth that shouldn't exist (combined REPL with a bunch of stdout messages clobbering the UI)

carver: > I have found that console has so much logging noise that it is borderline un-usabl

Yeah, I basically always do this:

```
geth ... console 2>geth.log
```

cburgdorf:
>I suggest we adopt the following convention for plugins which need to take full control of the process.

I will look into this!

>In the model I proposed (in another comment) we would get rid of the start and should_start mechanisms entirely in favor of the plugin having a formal route to take over the whole process

I think we still need that pattern because we want pugins to start based on various kind of events that happend. E.g. you may want to write a plugin that starts only based on the fact that another plugin started (e.g. the transaction pool). I think this will allow for more complex scenarios to be covered. We even use that today for the transaction pool itself that only starts as soon as the `PeerPool` got broadcasted as an available resource.

pipermerriam:
It looks like one-of-two-things

1. https://github.com/ethereum/py-evm/blob/master/eth/__init__.py#L24 is no longer sufficiently large and needs to be increased
2. something is actually causing infinite recursion

I suspect it is the first though I don't know why your PR would have caused this.

cburgdorf:
@pipermerriam thanks! Seems to be the second. Increasing the recursion limit doesn't help. I'll dig deeper tomorrow.

trinity/plugins/builtin/attach/plugin.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+
+        attach_parser.set_defaults(func=self.run_console)
+
+    def run_console(self, args: Namespace, chain_config: ChainConfig) -> None:
+        try:
+            console(chain_config.jsonrpc_ipc_path, use_ipython=self.use_ipython)
+        # FIXME: That doesn't seem to work very well as we leave behind the
+        # JSON IPC file by default it seems. Maybe we can make this more reliable
+        # with the upcoming PidFile support.
```
carver: I think this is just `IPCServer` leaving behind the file, no matter how it's started. You can remove this comment, I'll post a fixup PR.

trinity/main.py
```diff
@@ -137,80 +134,66 @@ def main() -> None:
...
-        args=(
-            chain_config,
-            LevelDB,
-        ),
-        kwargs=extra_kwargs,
-    )
+    if hasattr(args, 'func'):
+        args.func(args, chain_config)
+    else:
+        display_launch_logs(chain_config)
```
carver: The `main()` method is pretty long. Maybe this `else` block belongs in a new `run_node()` method. (This has the side benefit of not requiring the huge indentation change, especially if you insert it right after `main()`)

trinity/cli_parser.py
```diff
@@ -189,31 +184,3 @@ def __call__(self,
         "Enables profiling via cProfile."
     ),
 )
-
-
-#
-# Add `console` sub-command to trinity CLI.
-#
-console_parser = subparser.add_parser(
-    'console', help='run the chain and start the trinity REPL')
```
carver: Maybe add a comment in this config file that plugins may add subcommands that set a `func` attribute to be run instead of the main node?

carver:
Hm, no idea yet why that would affect state tests.

trinity/main.py
cburgdorf: Yeah, I had that initially. But then, having to pass 6 or 7 arguments to that function turned me off and made me roll it back. Another idea to get rid of the `else` and the indentation would be to do.

```python
if hasattr(args, 'func'):
    return args.func(args, chain_config)
```

cburgdorf:
@carver thanks. Here's the output of `pytest -s tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py -k Call1024BalanceTooLow_d0g0v0_Byzantium`

```
main.py                    130 WARNING  RPC method caused exception
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/rpc/main.py", line 118, in _get_result
    result = method(*params)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/rpc/format.py", line 102, in formatted_func
    return func(self, *formatted)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/rpc/modules/evm.py", line 43, in applyBlockFixture
    _, _, rlp_encoded = apply_fixture_block_to_chain(block_info, self._chain)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/tools/fixture_tests.py", line 644, in apply_fixture_block_to_chain
    mined_block = chain.import_block(block)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/chains/base.py", line 573, in import_block
    imported_block = self.get_vm(base_header_for_import).import_block(block)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/base.py", line 471, in import_block
    new_header, receipts, _ = self.apply_all_transactions(block.transactions, self.block.header)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/base.py", line 427, in apply_all_transactions
    transaction,
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/base.py", line 348, in apply_transaction
    state_root, computation = self.state.apply_transaction(transaction)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/state.py", line 219, in apply_transaction
    computation = self.execute_transaction(transaction)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/forks/frontier/state.py", line 190, in execute_transaction
    return executor(transaction)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/state.py", line 244, in __call__
    computation = self.build_computation(message, valid_transaction)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/forks/frontier/state.py", line 129, in build_computation
    transaction_context).apply_message()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/forks/frontier/computation.py", line 71, in apply_message
    self.transaction_context,
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/computation.py", line 551, in apply_computation
    opcode_fn(computation=computation)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/logic/call.py", line 115, in __call__
    child_computation = computation.apply_child_computation(child_msg)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/computation.py", line 345, in apply_child_computation
    child_computation = self.generate_child_computation(child_msg)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/computation.py", line 360, in generate_child_computation
    self.transaction_context,
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/forks/frontier/computation.py", line 71, in apply_message
    self.transaction_context,
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/computation.py", line 551, in apply_computation
    opcode_fn(computation=computation)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/logic/call.py", line 115, in __call__
    child_computation = computation.apply_child_computation(child_msg)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/computation.py", line 345, in apply_child_computation
    child_computation = self.generate_child_computation(child_msg)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/eth/vm/computation.py", line 360, in generate_child_computation
    self.transaction_context,
(...goes on until recursion limit is hit...)
```
So, something is causing it to infinitely apply child computations but I don't see the connection to this PR yet. However, it *is* this last commit that is causing it. Anyway, I'll find out. In fact, I just noticed that just **not loading** the attach plugin makes the error go away. :thinking:

cburgdorf:
Ok, I found the cause. It's the `get_suited_attach_plugin` function that is either loading the IPython or Vanilla shell based attach plugin that is causing this. If I hardcoded one of these plugins it works, but the function that checks if IPython is available to dynamically decide, breaks it.

cburgdorf:
Ok, so it turns out that just `import IPython` breaks pytest. The simplest fix I've found so far is to just do:

```python
ENABLED_PLUGINS = [
    VanillaShellAttachPlugin() if "pytest" in sys.modules else get_suited_attach_plugin(),
    TxPlugin(),
]
```

Means, if we are in a `pytest` session, just hardcode the attach plugin, otherwise run the dynamic decision logic.

trinity/main.py
cburgdorf: Ok, moved some code around and now we are down to five parameters which I think is still ok.

cburgdorf:
@carver @pipermerriam The issues are fixed. Waiting for another :+1: (especially because of the pytest weirdness and the workaround) before merging.

pipermerriam:
You might try using this for the Ipython thing:

```
pkg_resources.get_distribution("ipython")
```

trinity/main.py
```diff
@@ -139,21 +140,29 @@ def main() -> None:
...
 
+    # Plugins can provide a subcommand with a `func` which does then control
+    # the entire process from here.
+    if hasattr(args, 'func'):
+        args.func(args, chain_config)
+    else:
+        continue_default_boot(args, chain_config, extra_kwargs, listener, logger)
+
+
+def continue_default_boot(args: Namespace,
```
pipermerriam: The `continue_` prefix is odd to me.  Maybe we could just call this:

- `trinity_main`
- `trinity_boot`
- `trinity_run`
- `trinity_default`

trinity/plugins/builtin/attach/__init__.py
```diff
@@ -0,0 +1,12 @@
+from trinity.plugins.builtin.attach.plugin import (
+    AttachPlugin
+)
+
+
+def get_suited_attach_plugin() -> AttachPlugin:
+    try:
+        import IPython  # noqa: F401
```
pipermerriam: Maybe since this has been determined to have a side effect we can instead use the following pattern.

```python
def is_ipython_available():
    try:
        pkg_resources.get_distribution('IPython')
        return False
    else:
        return True
```

I think there **might** be some weirdness in whether `DistributionNotFound` is available but that might be isolated to older python versions.

trinity/plugins/registry.py
```diff
@@ -8,5 +16,8 @@
 # config file which plugin is enabled or not
 
 ENABLED_PLUGINS = [
-    TxPlugin()
+    # Importing IPython (even just to check if it is available) breaks pytest.
+    # We detect if we are in a pytest session and just hardcoded the plugin then.
+    AttachPlugin() if "pytest" in sys.modules else get_suited_attach_plugin(),
```
pipermerriam: I'd love to drop this `if "pytest" in sys.modules` in favor of the `pkg_resources` approach suggested above.

cburgdorf:
@pipermerriam I made the requested changes but rebasing on latest master unveiled another issue. The `fix-unclean-shutdown` was implemented as a subcommand but multiple subcommand parser are seem to be illegal. Just for the sake of doing a quick rebase I turned that into a `--fix-unclean-shutdown` flag but I guess the proper fix would be to go back to the approach were trinity sets up one subcommand parser and then passes that on to the plugins.

I guess the rest of how we currently handling things (e.g. the `args.func` thing) would be unaffected of that.

cburgdorf:
@pipermerriam I'll fix that right away...

cburgdorf:
@pipermerriam ok, this should now be ready for another review. I added another commit *before* the commit that introduces the `attach` plugin, that refactores the plugins to expect the `subparser` passed. Then, `fix-unclean-shutdown` can continue to be a subcommand. However, I think that command should also be a plugin rather than being backed in. Main reasoning is basically to keep the core code base lighter.

pipermerriam:
> However, I think that command should also be a plugin rather than being backed in. Main reasoning is basically to keep the core code base lighter.

👍

PR #1001, opened 2018-07-09
[WIP] issue #592: creation of a DoScontract: deploy empty contract/ sstore / revert sst…

GagziW:
…ore / revert empty contract deployment and adjusted as well as added the respective benchmark. 

The contract is a first draft. I am not 100% sure if this implementation of the contract fullfils the requirements of issue #592 DOS style transaction spam. Please let me know!

Thank you! :)


scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,314 @@
...
+                   num_tx: int) -> BaseBlock:
+
+        for i in range(1, num_tx + 1):
+            self.apply_transaction(chain)
+
+        return chain.mine_block()
+
+    def apply_transaction(self, chain: MiningChain) -> None:
+
+        self._do_benchMark(chain)
```
cburgdorf: Casing is wrong here.

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,314 @@
...
+    
+    def deploy_DoScontract(self, chain: MiningChain) -> None:
```
cburgdorf: I think that should better be cased as `deploy_dos_contract`

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,314 @@
...
+
+class DosContractDepolyBenchmark(BaseDoScontractBenchmark):
+    def __init__(self) -> None:
+        super().__init__()
+
+    @property
+    def name(self) -> str:
+        return 'DoScontract deployment'
+
+    def _setup_benchmark(self, chain: MiningChain) -> None:
```
cburgdorf: let's make `_setup_benchmark` non-abstract in the base contract and then we don't need this empty implementation here. In other words, provide the empty implementation in the base contract.

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,314 @@
...
+            from_=FUNDED_ADDRESS,
+            to=self.deployed_contract_address,
+            amount=0,
+            gas=FIFTH_TX_GAS_LIMIT,
+            data=decode_hex(w3_tx5['data']),
+        )
+
+        block, receipt, computation = chain.apply_transaction(tx5)
+
+class DosContractDepolyBenchmark(BaseDoScontractBenchmark):
```
cburgdorf: Not entirely sure if we have a convention in place but I'd rather have `DoS` written as `DOS` or even `Dos` making this ether `BaseDOSContractBenchmark` or `BaseDosContractBenchmark`

scripts/benchmark/run.py
```diff
@@ -13,8 +13,13 @@
     SimpleValueTransferBenchmark,
 )
 
-from checks.deploy_erc20 import (
-    DeployErc20
+
+from checks.deploy_dos import (
+    DOSContractDepolyBenchmark,
```
cburgdorf: typo `DOSContractDepolyBenchmark` -> `DOSContractDeployBenchmark`

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,312 @@
...
+
+CONTRACT_FILE = 'scripts/benchmark/contract_data/DOSContract.sol'
+CONTRACT_NAME = 'DOSContract'
+
+W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
+
+
+class BaseDOSContractBenchmark(BaseBenchmark):
+
+    def __init__(self, num_blocks: int = 100, num_tx: int = 2) -> None:
```
cburgdorf: Depending on the time these benchmarks take maybe we could bump up the transactions to really *fill* the blocks. I know that's a bit of a balance act because the benchmarks still need to finish timely to not slow down CI too much. That said, maybe we want to shift towards *less* blocks but with *more* transactions so that the benchmark is less impacted by the general block forging costs and more by the actual costs of transaction processing?

Maybe have @pipermerriam weigh in on that.

GagziW: I increased the TX per block to 10 and decrease the number of blocks to 20. It now takes around 0.6-0.8 seconds longer to build all the 200 TXs.

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,296 @@
...
+    get_compiled_contract
+)
+from utils.reporting import (
+    DefaultStat,
+)
+from utils.tx import (
+    new_transaction,
+)
+
+from abc import (
```
cburgdorf: I hate to be nitpicking about these things and in the future we'll probably use something like `black` to auto format but for the time being, can you move this import to the very top. Imports should be ordered according to our style guide: https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#overall-ordering

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,296 @@
...
+FORTH_TX_GAS_LIMIT = 21272
+FIFTH_TX_GAS_LIMIT = 21272
+
+CONTRACT_FILE = 'scripts/benchmark/contract_data/DOSContract.sol'
+CONTRACT_NAME = 'DOSContract'
+
+W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
+
+
+class BaseDOSContractBenchmark(BaseBenchmark):
```
cburgdorf: I guess that we could even turn some of this into a `BaseContractInteractionBenchmark` class that is reused for the ERC20 token benchmarks as well.

Doesn't need to be addressed in this PR though.

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,296 @@
...
+        return total_stat
+
+    def mine_blocks(self, chain: MiningChain, num_blocks: int, num_tx: int) -> Tuple[int, int]:
+        total_gas_used = 0
+        total_num_tx = 0
+
+        for i in range(1, num_blocks + 1):
+            block = self.mine_block(chain, i, num_tx)
+            total_gas_used = total_gas_used + block.header.gas_used
+            total_num_tx = total_num_tx + len(block.transactions)
```
pipermerriam: *mutability nitpick*

Here is how you can do this without all of the incremental mutation.

```python
blocks = tuple(self.mine_block(chain, i, num_tx) for i in range(1, num_blocks + 1))
total_gas_used = sum(block.header.gas_used for block in blocks)
total_num_tx = sum(len(block.transactions) for block in blocks)
```

I find this style easier to read and reason about.

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -0,0 +1,296 @@
...
+            from_=FUNDED_ADDRESS,
+            to=CREATE_CONTRACT_ADDRESS,
+            amount=0,
+            gas=FIRST_TX_GAS_LIMIT,
+            data=decode_hex(w3_tx1['data']),
+        )
+
+        logging.debug('Applying Transaction {}'.format(tx))
+
+        block, receipt, computation = chain.apply_transaction(tx)
```
pipermerriam: doesn't look like we use either `block` or `receipt`.  Normally we use a placeholder variable of a single underscore for un-used variables like this.

```python
_, _, computation = chain.apply_transaction(tx)
```

cburgdorf: These placeholders don't work with mypy if both variables are of different types

williambannas:
I don't mind dealing with merge conflicts, and refactor in pull request #1046

scripts/benchmark/checks/deploy_dos.py
GagziW: fixed :)

scripts/benchmark/checks/deploy_dos.py
GagziW: so I just left it as it was?

GagziW:
>a single large transaction (or a few large transactions) that are full of heavy operations.

would it be alright to leave that for a seperate PR?

Is there anything else I need to change to get this PR merged? :)

cburgdorf:
>Is there anything else I need to change to get this PR merged? :)

@GagziW Nope, we can merge as is and tune it with follow up PRs. Merged it! :tada: Thank you very much for your work on this!

Issue #956, opened 2018-06-25
Exception information lost when it passes through multiprocessing-based services


gsalgado:
When we call a method on one of our multiprocessing-based services and that method raises an exception, the multiprocessing module catches and re-raises it, but then we end up with a traceback that doesn't show the code which originally raised the exception, like this:

```
   ERROR  06-23 08:26:27  base_events  Task exception was never retrieved
future: <Task finished coro=<FastChainSyncer.sync() done, defined at /home/salgado/src/py-evm/p2p/chain.py:122> exception=KeyError(b'(e\xed\x84\xeark\xf1\x8c\x9d\xb8/\xce\x9f\x
a0l\xe8\x11\xef\x88\x83\x933<4\xab\x15{\xc9+\xfa1',)>
Traceback (most recent call last):
  File "/home/salgado/src/py-evm/p2p/chain.py", line 135, in sync
    await self._sync(peer)
  File "/home/salgado/src/py-evm/p2p/chain.py", line 184, in _sync
    head_number = await self._process_headers(peer, headers)
  File "/home/salgado/src/py-evm/p2p/chain.py", line 610, in _process_headers
    await self.wait(self.chain.coro_import_block(block, perform_validation=True))
  File "/home/salgado/src/py-evm/p2p/service.py", line 32, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/salgado/src/py-evm/p2p/service.py", line 51, in wait_first
    return await wait_with_token(*awaitables, token=token_chain, timeout=timeout)
  File "/home/salgado/src/py-evm/p2p/cancel_token.py", line 119, in wait_with_token
    return done.pop().result()
  File "/home/salgado/src/py-evm/trinity/utils/mp.py", line 27, in method
    args,
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.6/multiprocessing/managers.py", line 772, in _callmethod
    raise convert_to_error(kind, result)
KeyError: b'(e\xed\x84\xeark\xf1\x8c\x9d\xb8/\xce\x9f\xa0l\xe8\x11\xef\x88\x83\x933<4\xab\x15{\xc9+\xfa1'
```

In the log above, the `KeyError` was generated inside `coro_import_block()`, but one has no idea exactly where. We need to somehow retain/log that info so that we can debug those errors, otherwise it is really hard

[...]

PR #1004, opened 2018-07-10
trinity: Proxied services now record remote exceptions' tracebacks 

gsalgado:
Closes: #956

### DIFF
--- a/trinity/chains/__init__.py
+++ b/trinity/chains/__init__.py
@@ -4,8 +4,14 @@ from multiprocessing.managers import (  # type: ignore
     BaseManager,
     BaseProxy,
 )
+import inspect
 import os
+import traceback
+from types import TracebackType
 from typing import (
+    Any,
+    Callable,
+    List,
     Type
 )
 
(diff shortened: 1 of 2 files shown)


trinity/chains/__init__.py
```diff
@@ -144,6 +148,33 @@ def initialize_database(chain_config: ChainConfig, chaindb: AsyncChainDB) -> Non
             )
 
 
+class TracebackKeeper:
```
pipermerriam: *naming nitpick*

maybe `TracebackPreserver` or `TracebackRecorder`?

cburgdorf: :+1:  on `TracebackRecorder`

trinity/chains/__init__.py
```diff
@@ -144,6 +148,33 @@ def initialize_database(chain_config: ChainConfig, chaindb: AsyncChainDB) -> Non
...
+
+    def __dir__(self) -> List[str]:
+        return dir(self.obj)
+
+    def __getattr__(self, name: str) -> Any:
+        attr = getattr(self.obj, name)
+        if not inspect.ismethod(attr):
+            return attr
+
+        def wrapper(*args: Any, **kwargs: Any) -> None:
```
pipermerriam: **maybe** lift this out of the function body to the top level of the module and then this function body would be 

```
if not inspect.ismethod(attr):
    return attr
else:
    return record_traceback_on_error(attr)
```

gsalgado:
I've improved this a bit based on [this trick](https://bugs.python.org/issue13831)

It uses exception chaining, which means we don't need any extra hacks to special case the logging of those exceptions. This is how it looks like:

```
   ERROR  07-11 11:45:15     service  Unexpected error in <p2p.chain.FastChainSyncer object at 0x7f7fa2b3cf60>, exiting
trinity.chains.RemoteTraceback:
"""
  File "/home/salgado/src/py-evm/trinity/chains/__init__.py", line 176, in wrapper
    return attr(*args, **kwargs)
  File "/home/salgado/src/py-evm/evm/db/header.py", line 86, in get_canonical_block_hash
    validate_block_number(block_number, title="Block Number")
  File "/home/salgado/src/py-evm/evm/validation.py", line 201, in validate_block_number
    validate_is_integer(block_number, title="Block Number")
  File "/home/salgado/src/py-evm/evm/validation.py", line 37, in validate_is_integer
    "{title} must be a an integer.  Got: {0}".format(type(value), title=title)
evm.exceptions.ValidationError: Block Number must be a an integer.  Got: <class 'bytes'>
"""

The above exception was the direct cause of the following exception:

  File "/home/salgado/src/py-evm/p2p/service.py", line 55, in run
    await self._run()
  File "/home/salgado/src/py-evm/p2p/chain.py", line 120, in _run
    await self.wait(self.db.coro_get_canonical_block_hash(b'foo'))
  File "/home/salgado/src/py-evm/p2p/cancel_token.py", line 102, in wait
    return await self.wait_first(awaitable, token=token, timeout=timeout)
  File "/home/salgado/src/py-evm/p2p/cancel_token.py", line 123, in wait_first
    return await wait_with_token(*awaitables, token=token_chain, timeout=timeout)
  File "/home/salgado/src/py-evm/p2p/cancel_token.py", line 160, in wait_with_token
    return done.pop().result()
  File "/home/salgado/src/py-evm/trinity/utils/mp.py", line 27, in method
    args,
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 56, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/lib/python3.6/multiprocessing/managers.py", line 772, in _callmethod
    raise convert_to_error(kind, result)
evm.exceptions.ValidationError: Block Number must be a an integer.  Got: <class 'bytes'>
 ```

PR #1005, opened 2018-07-10
Inherit BaseChain from LightDispatchChain

carver:
### What was wrong?

#998 introduced inheritance of `LightDispatchChain` -> `Chain`. For many methods, the `Chain` implementations will are broken if run in a `LightDispatchChain`. They expect a `chaindb`, but the dispatch chain only has a `headerdb`. It seems safer, easier to reason about, and quicker to isolate problems, if we only inherit `BaseChain`.

### How was it fixed?

- Move some key methods that don't depend on `chaindb` from `Chain` to `BaseChain`.
- Require header as part of `get_vm_class()` so that it can be implemented in `BaseChain`

#### Cute Animal Picture

### DIFF
--- a/evm/chains/base.py
+++ b/evm/chains/base.py
@@ -124,9 +124,8 @@ class BaseChain(Configurable, ABC):
         raise NotImplementedError("Chain classes must implement this method")
 
     @classmethod
-    @abstractmethod
     def get_vm_configuration(cls) -> Tuple[Tuple[int, Type['BaseVM']], ...]:
-        raise NotImplementedError("Chain classes must implement this method")
+        return cls.vm_configuration
 
     #
     # Chain API
@@ -149,9 +148,11 @@ class BaseChain(Configurable, ABC):
     #
     # VM API
     #
-    @abstractmethod
-    def get_vm_class(self, header: BlockHeader=None) -> Type['BaseVM']:
-        raise NotImplementedError("Chain classes must implement this method")
+    def get_vm_class(self, header: BlockHeader) -> Type['BaseVM']:
+        """
+        Returns the VM instance for the given block number.
+        """
+        return self.get_vm_class_for_block_number(header.block_number)
 
     @abstractmethod
     def get_vm(self, header: BlockHeader=None) -> 'BaseVM':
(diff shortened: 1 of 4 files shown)


carver:
Hi @cburgdorf does this alternative approach to #998 look okay to you?

cburgdorf:
I rage merged this because it coincidentally contains a fix for #1007 (Unrelated to the main inheritance issue discussed here).

On the topic:

Your main point seems to be that inheriting from `Chain` makes the `LightDispatchChain` inherit methods that will be broken for it. That's a valid concern, so I will say that these methods aren't used anyway and if they throw an exception because we re-implement them to throw or if they throw because `chaindb` isn't there doesn't make a huge difference in the result. I do admit it feels cleaner to have these methods throw the proper `NotImplementedError` though.

However, it comes at a cost of *missing out* methods that the `LightDispatchChain` *should* inherit, some of them which this PR has moved into the `BaseChain` to tackle that. I wasn't sure if we want any implementations in the `BaseChain` at all or if it should stay entirely abstract. In that case, it may make sense to introduce yet another layer such as `BaseChainMethods` where these implementations could go?

Another thing to note is, that with the `LightDispatchChain` inheriting directly from `BaseChain` we will likely see this break again until we have integration tests for Trinity simply because it is easy to forget to update the `LightDispatchChain` whenever the `BaseChain` changes.

That said, I volunteer to do the integration tests for Trinity next week if no one beats me to it.

carver:
> In that case, it may make sense to introduce yet another layer such as BaseChainMethods where these implementations could go?

Eventually, maybe. The time when we have enough information to do that cleanly is probably when we want to inherit from `BaseChain` for the Casper/sharding beacon chain. That will force us to focus on which assumptions need to be re-examined. At that point, we'll have a clearer idea of what should be defined at different inheritance layers (or we might come up with an entirely different solution by breaking it apart and composing classes).

> we will likely see this break again until we have integration tests for Trinity

Yes, I agree this is a huge problem. I also agree that the best solution is adding integration tests. :)

> That said, I volunteer to do the integration tests for Trinity next week if no one beats me to it.

Great! I'll race through Grace Hopper issues to see if I can beat you there. I just created #1008 to track. If I'm able to start taking a look, I'll "assign" it to myself. If I'm not assigned, it's all yours.

carver:
@cburgdorf if you merge my branches, would you mind using a merge commit instead of a squash commit? One reason I care is that I use (roughly) this shell script to clean up my local branches that have been merged upstream:

```sh
git branch -d $(git branch --merged master | grep -v master)
```

Using a squash commit defeats this approach :/

Obviously a minor thing, but it would be nice.

cburgdorf:
No problem!

PR #1006, opened 2018-07-11
p2p: Announce serveHeaders/txRelay during LES handshake

gsalgado:
Our light client was recently extended to serve headers and relay
transactions, so we should advertise that during the handshake

### DIFF
--- a/p2p/les.py
+++ b/p2p/les.py
@@ -1,4 +1,4 @@
-from typing import Any, cast, Dict, Generator, List, Tuple, TYPE_CHECKING, Union
+from typing import Any, cast, Dict, Iterator, List, Tuple, TYPE_CHECKING, Union
 
 import rlp
 from rlp import sedes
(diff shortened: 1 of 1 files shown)


p2p/les.py
```diff
@@ -276,6 +289,9 @@ def send_handshake(self, chain_info: 'ChainInfo') -> None:
             'headHash': chain_info.block_hash,
             'headNum': chain_info.block_number,
             'genesisHash': chain_info.genesis_hash,
+            'serveHeaders': None,
+            'serveChainSince': 0,
+            'txRelay': None,
```
pipermerriam: Is it correct that this should be set to a truthy value (`1`?) when we have a working transaction pool?

gsalgado: Both `txRelay` and `serveHeaders` have no value -- they just need to be present to signal that we do those things. I thought we already had a working tx pool?  If that's not the case I should not include txRelay yet

cburgdorf:
It should be ready and it can be disabled via flag if things turn out bad. Maybe just add a comment to these `None` values to explain that they just need to be present. Or if it doesn't matter, maybe set them to `True`?

PR #1009, opened 2018-07-12
Track peer performance, favor fast peers

### LINKED ISSUE #868: Don't let a single slow peer kill performance of chain sync
### What is wrong?

When doing a chain sync, for each chain segment we split the missing block parts (bodies, receipts) in N batches (where N is the number of connected peers that have those block parts) and request each batch to a different peer. Quite often the peers will not reply with all requested items (because they implement a byte-size cap in responses), so we need to wait for replies from all peers in order to figure out what block parts we're still missing so that we can make another round of requests.

This obviously doesn't perform well if we have a slow peer.

### How can it be fixed

Need to keep track of the avg response time (or throughput?) for peers and request body parts only to the ones we deem fast enough (e.g. top 2/3rds ranked by response time, or below a certain response time threshold)

carver:
### What was wrong?

Related to #868 -- slow peers can dramatically slow down syncing

### How was it fixed?

- track throughput performance of peers
- favor high-throughput peers, sending them bigger requests
- if performance of a peer is bad enough, we will eventually not request anything from them (this happens naturally in the current approach, because the proportional number of block parts requested is `floor`ed and so you can end up with 0 parts requested if a peer is slow enough)

As mentioned in #868 a better solution is eventually to stream all block part requests, rather than batching them into header chunks of 192. This is a performance patch on the way there, although significant parts of PR would be obsoleted by such a solution. It would take a while to get that solution right, and probably belongs in the next release instead of the current one.

#### Cute Animal Picture

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -7,7 +7,6 @@ from enum import (
 )
 from functools import partial
 import logging
-import math
 import time
 from typing import (
     Any,
@@ -24,7 +23,6 @@ from typing import (
 )
 
 from cytoolz import (
-    take,
     unique,
 )
 
@@ -58,6 +56,7 @@ from p2p.service import BaseService
 from p2p.utils import (
     ThroughputTracker,
     get_process_pool_executor,
+    get_scaled_batches,
 )
 
 
@@ -347,7 +346,7 @@ class FastChainSyncer(BaseHeaderChainSyncer):
             smoothing_factor=THROUGHPUT_SMOOTHING_FACTOR,
         )
         # mypy is confused -- https://github.com/python/typeshed/issues/2329
-        self._throughput_stats = {part: defaultdict(create_stats) for part in BlockPartEnum}  # type: ignore # noqa: E501
+        self._peer_stats = {part: defaultdict(create_stats) for part in BlockPartEnum}  # type: ignore # noqa: E501
 
     async def _calculate_td(self, headers: Tuple[BlockHeader, ...]) -> int:
         """Return the score (total difficulty) of the last header in the given list.
@@ -449,7 +448,7 @@ class FastChainSyncer(BaseHeaderChainSyncer):
             unexpected = received_keys.difference(key_func(header) for header in headers)
 
             work_size = len(received_keys.difference(unexpected))
-            self._get_throughput_stats(peer, part_name).complete_work(work_size)
+            self._get_peer_stats(peer, part_name).complete_work(work_size)
 
             parts.extend(received)
             pending_replies -= 1
@@ -467,16 +466,16 @@ class FastChainSyncer(BaseHeaderChainSyncer):
 
         return dict((part.unique_key, part.part) for part in parts)
 
-    def _get_throughput_stats(self, peer: BasePeer, body_part: BlockPartEnum) -> ThroughputTracker:
-        return self._throughput_stats[body_part][peer]
+    def _get_peer_stats(self, peer: BasePeer, body_part: BlockPartEnum) -> ThroughputTracker:
+        return self._peer_stats[body_part][peer]
 
     def deregister_peer(self, peer: BasePeer) -> None:
         # Delete all in-progress stats for the given peer. Otherwise, if a peer disconnects before
         # sending a block part, stats would fail when starting to track throughput the 2nd time.
         # Also, we don't want a perpetually growing dict of stats.
         for part in BlockPartEnum:
-            if peer in self._throughput_stats[part]:
-                del self._throughput_stats[part][peer]
+            if peer in self._peer_stats[part]:
+                del self._peer_stats[part][peer]
 
     def _request_block_parts(
             self,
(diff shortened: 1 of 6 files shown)


p2p/chain.py
```diff
@@ -437,20 +486,50 @@ def __init__(self,
...
+            block_part: BlockPartEnum,
+    ) -> int:
         peers = self.peer_pool.get_peers(target_td)
         if not peers:
             raise NoEligiblePeers()
-        length = math.ceil(len(headers) / len(peers))
-        batches = list(partition_all(length, headers))
+        batches = self._select_headers_by_peer(headers, peers, block_part)
         for peer, batch in zip(peers, batches):
+            self._get_throughput_stats(peer, block_part).begin_work()
```
pipermerriam: Should this handle the case where work was already begun but never completed?  Not entirely sure but it seems like a peer could disconnect and leave us in a *middle* state at which point this will fail everytime we try to request things from a given peer.

p2p/chain.py
```diff
@@ -302,6 +313,36 @@ class LightChainSyncer(BaseHeaderChainSyncer):
...
+        if self._last_start is not None:
+            raise ValidationError("Cannot start the ThroughputTracker again without completing it")
+        self._last_start = time.perf_counter()
+
+    def complete_work(self, work_completed: Union[int, float]) -> None:
+        if self._last_start is None:
+            raise ValidationError("Cannot end the ThroughputTracker without starting it")
+        time_elapsed = time.perf_counter() - self._last_start
+        last_throughput = work_completed / time_elapsed
+        self._throughput = (self._throughput * (1 - self._alpha)) + (last_throughput * self._alpha)
```
pipermerriam: This function body looks like it's missing a line to reset `self._last_start` to `None`.

p2p/chain.py
carver: Added a `deregister_peer` to remove the stats when a peer logs off. This also has the nice effect of limiting the total size of the stats we track.

carver:
Actually, one more fixup: have to handle the 0 headers requested case correctly.

p2p/chain.py
```diff
@@ -437,19 +465,60 @@ def __init__(self,
...
+            self,
+            headers: List[BlockHeader],
+            peers: List[BasePeer],
+            block_part: BlockPartEnum,
+    ) -> List[List[BlockHeader]]:
+        speeds = [self._get_throughput_stats(peer, block_part).get_throughput() for peer in peers]
+        speed_sum = sum(speeds)
+        fractional_speeds = [speed / speed_sum for speed in speeds]
+
+        # request headers proportionally, so faster peers are asked for more headers than slow peers
```
gsalgado: The peers will actually be asked for body parts for certain headers, and not for headers themselves, right? The comment kind of suggests the latter

p2p/chain.py
```diff
@@ -437,19 +465,60 @@ def __init__(self,
...
 
+    def _get_throughput_stats(self, peer: BasePeer, body_part: BlockPartEnum) -> ThroughputTracker:
+        return self._throughput_stats[body_part][peer]
+
+    def deregister_peer(self, peer: BasePeer) -> None:
+        # Delete all in-progress stats for the given peer. Otherwise, if a peer disconnects before
+        # sending a block part, stats would fail when starting to track throughput the 2nd time.
+        # Also, we don't want a perpetually growing dict of stats.
+        for part in BlockPartEnum:
+            del self._throughput_stats[part][peer]
```
gsalgado: Could this raise a KeyError if the peer never sent us anything?

carver: Ah yes, a bug in my mental representation of `defaultdict`

carver:
> am having a hard time as I'm not sure if the PR description is still reflecting the actual changes, and it claims to fix #868 but I'm under the impression it doesn't implement everything described in #868 (comment)

Totally fair. I'll drop the claim that it fixes that issue.

It _does_ help speed things up if you have a few peers that are great (say on your LAN) or a few peers that are terrible (someone out there running a node on dialup? :P)

But it doesn't approach broader change: to stream the block part requests instead of batching them. I think that should be pushed out of the Grace Hopper milestone.

p2p/utils.py
```diff
@@ -47,3 +54,33 @@ def get_process_pool_executor() -> ProcessPoolExecutor:
...
         cpu_count = os_cpu_count - 1
     return ProcessPoolExecutor(cpu_count)
+
+
+class ThroughputTracker:
+    """
+    Tracks throughput using an exponential moving average.
+    https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average
+    """
+    def __init__(self, default_throughput: Union[int, float], smoothing_factor: float) -> None:
```
cburgdorf: nitpick: Is there a reason for using `Union[int, float]` over simply sticking to `float` entirely for this API?

p2p/chain.py
```diff
@@ -438,19 +467,61 @@ def __init__(self,
...
+        for peer, fractional_speed in zip(peers, fractional_speeds):
+            num_to_take = math.floor(fractional_speed * num_headers)
+            header_batch = list(take(num_to_take, header_iter))
+            batches.append(header_batch)
+
+        # any headers missed due to rounding error will go to the fastest peer
+        fastest_peer_idx = fractional_speeds.index(max(fractional_speeds))
+        batches[fastest_peer_idx] += list(header_iter)
+
+        return batches
```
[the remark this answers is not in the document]
pipermerriam: I'd suggest pulling this out into a utility method to accomplish this.  Way easier to test.

p2p/utils.py
carver: I guess it's just a claim that both will be fine. No objection to dropping to float, though.

carver:
Should be ready for another look.

If it made sense to do another iteration, I was thinking the util should just take the amount of work to do as an int, and return the amount of work each worker should do as an int. (rather than passing in the headers and returning the headers)

But this should work fine, and I think there are other more pressing matters. If we really want to reuse this method for a different thing, we can refactor it then.

Issue #932, opened 2018-06-20
K-th header PoW verification


gsalgado:
### What is wrong?

PoW verification is computationally expensive (even for high-performant languages like Go)

### How can it be fixed

instead of verifying the PoW on every header, geth does so on a randomly chosen subset of them, in a way that the
probability of a forgery is smaller than the probability of an attacker producing a SHA3 collision.

pipermerriam:
Regarding label usage.  My intention for the `p2p` label is everything that is not pure (or almost pure) DevP2P protocol.  Anything that's part of the ethereum node syncing code should be `trinity`

carver:
> My intention for the p2p label is everything that is not pure (or almost pure) DevP2P protocol.

Extra *not* in there?

So p2p label would be for "everything that **is** ~not~ pure (or almost pure) DevP2P protocol."

gsalgado:
Fair enough, I thought I was the only one using the `p2p` label, and was doing so to keep track of anything related to the `p2p` package

pipermerriam:
I've been wanting to move most of the *trinity* code from the `p2p` package into `trinity` for a while.  Probably something that should be done sooner than later.

PR #1010, opened 2018-07-12
During fast/light sync, check seal on a small subset of headers

gsalgado:
For every header batch, pick a random subset and check the seal only on those.

Before this change: ```Imported 192 headers in 1.33 seconds, new head: #3583663```

After this change: ```Imported 192 headers in 0.53 seconds, new head: #3585558```

Closes: #932

### DIFF
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -70,9 +70,8 @@ DEFAULT_MAX_PEERS = 25
 # Maximum allowed depth for chain reorgs.
 MAX_REORG_DEPTH = 24
 
-# Seal check frequency of headers downloaded during light/fast sync. Apparently 100 was the
-# optimal value determined by geth devs
+# Random sampling rate (i.e. every K-th) for header seal checks during light/fast sync. Apparently
+# 100 was the optimal value determined by geth devs
 # (https://github.com/ethereum/go-ethereum/pull/1889#issue-47241762), but in order to err on the
-# side of caution, we use 192/4 to ensure we check 4 headers out of every batch (of length
-# MAX_HEADERS_FETCH).
-FAST_SYNC_SEAL_CHECK_FREQUENCY = 192 // 4
+# side of caution, we use a higher value.
+SEAL_CHECK_RANDOM_SAMPLE_RATE = 48
--- a/p2p/constants.py
+++ b/p2p/constants.py
@@ -69,3 +69,10 @@ DEFAULT_MAX_PEERS = 25
 
 # Maximum allowed depth for chain reorgs.
 MAX_REORG_DEPTH = 24
+
+# Seal check frequency of headers downloaded during light/fast sync. Apparently 100 was the
+# optimal value determined by geth devs
+# (https://github.com/ethereum/go-ethereum/pull/1889#issue-47241762), but in order to err on the
+# side of caution, we use 192/4 to ensure we check 4 headers out of every batch (of length
+# MAX_HEADERS_FETCH).
+FAST_SYNC_SEAL_CHECK_FREQUENCY = 192 // 4
--- a/setup.py
+++ b/setup.py
@@ -46,7 +46,7 @@ deps = {
     ],
     'test': [
         "hypothesis==3.44.26",
-        "pytest~=3.2",
+        "pytest~=3.3",
         "pytest-asyncio==0.8.0",
         "pytest-cov==2.5.1",
         "pytest-xdist==1.18.1",
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -274,7 +274,8 @@ class BaseChain(Configurable, ABC):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def validate_chain(self, chain: Tuple[BlockHeader, ...], seal_check_frequency: int = 1) -> None:
+    def validate_chain(
+            self, chain: Tuple[BlockHeader, ...], seal_check_random_sample_rate: int = 1) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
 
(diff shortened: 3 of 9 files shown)


setup.py
```diff
@@ -46,7 +46,7 @@
     ],
     'test': [
         "hypothesis==3.44.26",
-        "pytest~=3.2",
+        "pytest~=3.3",
```
gsalgado: This is needed because we now have some tests that use the `caplog` fixture, which is not available in 3.2

p2p/constants.py
```diff
@@ -69,3 +69,10 @@
 
 # Maximum allowed depth for chain reorgs.
 MAX_REORG_DEPTH = 24
+
+# Seal check frequency of headers downloaded during light/fast sync. Apparently 100 was the
+# optimal value determined by geth devs
+# (https://github.com/ethereum/go-ethereum/pull/1889#issue-47241762), but in order to err on the
+# side of caution, we use 192/4 to ensure we check 4 headers out of every batch (of length
+# MAX_HEADERS_FETCH).
+FAST_SYNC_SEAL_CHECK_FREQUENCY = 192 // 4
```
carver: Maybe it's clearer to specify the period here, like:

```py
# check one in 4 headers
FAST_SYNC_SEAL_CHECK_PERIOD = 4
```

Then, we aren't assuming this value is used with 192 headers, either.

evm/chains/base.py
```diff
@@ -686,15 +687,26 @@ def validate_uncles(self, block: BaseBlock) -> None:
...
-    def validate_chain(self, chain: Tuple[BlockHeader, ...]) -> None:
+    def validate_chain(self, chain: Tuple[BlockHeader, ...], seal_check_frequency: int = 1) -> None:
         parent = self.chaindb.get_block_header_by_hash(chain[0].parent_hash)
-        for header in chain:
+        all_indices = list(range(len(chain)))
+        if seal_check_frequency == 1:
+            headers_to_check_seal = all_indices
+        else:
+            headers_to_check_seal = random.sample(
+                all_indices, len(all_indices) // seal_check_frequency)
```
carver: This looks like it's written as if the seal_check constant is a period. Otherwise, `192 // (192/4)` would only check 4 headers total in a chain of 192.

evm/chains/base.py
```diff
@@ -686,15 +687,26 @@ def validate_uncles(self, block: BaseBlock) -> None:
...
 
-    def validate_chain(self, chain: Tuple[BlockHeader, ...]) -> None:
+    def validate_chain(self, chain: Tuple[BlockHeader, ...], seal_check_frequency: int = 1) -> None:
         parent = self.chaindb.get_block_header_by_hash(chain[0].parent_hash)
-        for header in chain:
+        all_indices = list(range(len(chain)))
+        if seal_check_frequency == 1:
+            headers_to_check_seal = all_indices
+        else:
+            headers_to_check_seal = random.sample(
```
pipermerriam: I went as far as writing up a big comment with a *secure* version of `random.sample` implemented using `secrets.choice`.  But then I deleted it because it felt a bit overly paranoid/tinfoil-hat.  But then I still wanted to raise the issue.

I respect that there is very little likelihood that we're exposing attack surface by using `random.sample` here, but given the potential severity of a security exploit in this type of software, how would you feel about replacing this with such a small utility that uses *secure* sources of randomness.  My *basic* implementation is below:

```python
def samp(pop: Set, size: int) -> Set:  # could probably use a generic for the items in the set `Set[G]`
    if size >= len(pop):
        return pop
    rejects = set()
    for _ in range(len(pop) - size):
        # tuple cast is needed to use `secrets.choice`
        rejects.add(secrets.choice(tuple(pop.difference(rejects))))
    return pop.difference(rejects)
````

File: evm/chains/base.py [hunk 5]
```diff
@@ -686,15 +687,26 @@ def validate_uncles(self, block: BaseBlock) -> None:
...
+                all_indices, len(all_indices) // seal_check_frequency)
+
+        for i, header in enumerate(chain):
             if header.parent_hash != parent.hash:
                 raise ValidationError(
                     "Invalid header chain; {} has parent {}, but expected {}".format(
                         header, header.parent_hash, parent.hash))
             vm_class = self.get_vm_class_for_block_number(header.block_number)
-            vm_class.validate_header(header, parent)
+            if i in headers_to_check_seal:
```
pipermerriam: *semantic nitpick*

When I see inclusion checks I default to wanting the population object to be a `set` type.  Thoughts on casting `headers_to_check_seal` to a `set`?

evm/chains/base.py
```diff
@@ -686,15 +687,26 @@ def validate_uncles(self, block: BaseBlock) -> None:
...
+                check_seal = True
+            else:
+                check_seal = False
```
pipermerriam: Maybe just embed the call to `vm_class.validate_header(header, parent, check_seal=True)` here and get rid of the `check_seal` variable?

evm/vm/base.py
```diff
@@ -712,7 +713,8 @@ def validate_block(self, block):
             )
 
     @classmethod
-    def validate_header(cls, header: BlockHeader, parent_header: BlockHeader) -> None:
+    def validate_header(
+            cls, header: BlockHeader, parent_header: BlockHeader, check_seal: bool) -> None:
```
carver: I think it's okay to default `check_seal` to `True`, which should remove some of the boilerplate changes.

evm/chains/base.py
gsalgado: I initially wrote this using `secrets.choice()`, but changed it to `random` because we need to be python3.5 compatible as this is in the `evm` package, no?

pipermerriam: You are correct it seems.  👍

evm/chains/base.py
gsalgado: Yes, that's what I intended. In geth they check every 100-th header, and we're doing every 48th

p2p/constants.py
gsalgado: AIUI, geth checks 1 in 100, and as I mentioned below we're doing 1 in 48. Why would we do 1 in 4 instead?

carver: Sorry, I see now that the 'frequency` name was throwing me off. So I think my only issue with this is that it should be renamed to period. Frequency would be 1/period, ~0.02.

evm/chains/base.py
gsalgado: Well, `validate_header()` does more than checking the seal, and we still want to perform those for every header -- it's only the seal check that we want to skip sometimes

gsalgado: Oh, duh, now I understand what you mean. Will do that

evm/chains/base.py
gsalgado: Makes sense

p2p/constants.py
gsalgado: What about `seal_check_random_sample_rate`? 

```
            sample_size = len(all_indices) // seal_check_random_sample_rate
            headers_to_check_seal = set(random.sample(all_indices, sample_size))
```

PR #1011, opened 2018-07-13
p2p/discovery: Workaround a parity bug that causes pongs with wrong token

gsalgado:
Parity 1.10 (and earlier) nodes send the wrong token on pong messages
(https://github.com/paritytech/parity/issues/8038). This change keeps
a mapping of wrong tokens to the ones we expect, and uses that whenever
we get a pong with one of those wrong tokens.

### DIFF
--- a/p2p/kademlia.py
+++ b/p2p/kademlia.py
@@ -13,6 +13,7 @@ from typing import (
     Any,
     Callable,
     cast,
+    Dict,
     Hashable,
     Iterable,
     Iterator,
@@ -23,6 +24,8 @@ from typing import (
     TYPE_CHECKING,
 )
 from urllib import parse as urlparse
+
+import cytoolz
 
 from eth_utils import (
     big_endian_to_int,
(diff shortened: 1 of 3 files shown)


p2p/kademlia.py
```diff
@@ -421,6 +423,23 @@ def recv_pong(self, remote: Node, token: bytes) -> None:
...
+            token = self.parity_pong_tokens.pop(token)
+        else:
+            # This is a pong from a non-buggy node, so just cleanup self.parity_pong_tokens.
+            parity_token = None
+            for k, v in self.parity_pong_tokens.items():
+                if v == token:
+                    parity_token = k
+                    break
+            if parity_token is not None:
+                del self.parity_pong_tokens[parity_token]
```
carver: Another option if you like it better:
```py
token_pairs = cytoolz.valfilter(lambda val: val == token, self.parity_pong_tokens)
for parity_token, _token in token_pairs:
    del self.parity_pong_tokens[parity_token]
```

pipermerriam: How about just:

```
for k, v in tuple(self.parity_pong_tokens.items()):
    if v == token:
        del self.parity_pong_tokens[parity_token]
        break
```

pipermerriam: or a one-liner

```python
self.parity_pong_tokens = cytoolz.valfilter(lambda val: val != token, self.parity_pong_tokens)
```

PR #1013, opened 2018-07-13
Logs to stderr

carver:
### What was wrong?

logs to stdout are annoying, especially when invoking `trinity console`

### How was it fixed?

- move logs to stderr
- add warning and error options to cli logging parser
- bonus syntax fix in upnp exception

#### Cute Animal Picture

### DIFF
--- a/trinity/cli_parser.py
+++ b/trinity/cli_parser.py
@@ -38,10 +38,12 @@ class ValidateAndStoreEnodes(argparse.Action):
 
 DEFAULT_LOG_LEVEL = 'info'
 LOG_LEVEL_CHOICES = (
+    'trace',
     'debug',
     'info',
     'warning',
     'error',
+    'critical',
 )
 
 
--- a/trinity/cli_parser.py
+++ b/trinity/cli_parser.py
@@ -40,6 +40,8 @@ DEFAULT_LOG_LEVEL = 'info'
 LOG_LEVEL_CHOICES = (
     'debug',
     'info',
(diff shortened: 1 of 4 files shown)


trinity/cli_parser.py
```diff
@@ -40,6 +40,8 @@ def __call__(self,
 LOG_LEVEL_CHOICES = (
     'debug',
     'info',
+    'warning',
+    'error',
```
pipermerriam: maybe add `'trace'`?

cburgdorf:
@carver 

>logs to stdout are annoying

Do you mind elaborating on this? What's the benefit of writing standard logs (non error logs) to `stderr`? I think I've seen this pattern before (I remember I once wondered why `curl` was logging something to `stderr` that I expected on `stdout`) but I'm not sure I understand the reasoning behind it. This may be a great opportunity to learn :)

carver:
> Do you mind elaborating on that?

Gladly. :)

If an executable is outputting anything that *isn't* logs, then it's convenient to separate the two streams (executable output & logs). Here, that comes up in `trinity console` which outputs the result of your console commands, and the logs. The console is essentially unusable unless you turn logging down to `error`, because it's too noisy. Even then, it's pretty annoying to have text output into your console session. Splitting the streams makes it easy to do:

```sh
trinity console 2>trinity.log
```

Then, you get only the console output in your terminal, and your logs are viewable in the `trinity.log` file.

Issue #1012, opened 2018-07-13
Fix leaky `_child_services` property on Service class


pipermerriam:
### What is wrong?

https://github.com/ethereum/py-evm/blob/0ff7bc0f820250d7f15e804bc09870776ea34eb5/p2p/service.py#L21

The `Service` class has list set as a class property.  This will cause all instances of the `Service` class to share this list which is not the desired behavior.

cc @cburgdorf (this is a common mistake in python, classes should rarely have mutable data structures set at the class level)

### How can it be fixed

Remove the assignment of the list from the class level into the `__init__` constructor.

```python
class Service:
    _child_services: List['BaseService']

    def __init__(self, ...):
        ...
        self._child_services = []
```

cburgdorf:
Haha, I'm tempted to post a picture of us sitting on the porch of our
airbnb, enjoying some drinks over the sunset while listening to my 5yo
explaining the world to us ;)

PR #1015, opened 2018-07-14
fixes issue #1012

eddybenkhin:
### What was wrong?

 classes should rarely have mutable data structures set at the class level

### How was it fixed?

moving the instantiation to the init constructor


pipermerriam:
@eddybenkhin thanks for cleaning this up for us 😄

Issue #1016, opened 2018-07-14
Trinity logs lots of "socket.send() raised exception" warnings


pipermerriam:
on current master @ 32cc41d5315003289b1c86b31680f2501894da06

### What is wrong?

When running `trinity` a **lot** of these warnings happen.

```bash
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
 WARNING  07-14 03:48:12  selector_events  socket.send() raised exception.
```

### How can it be fixed

It looks like it may related to this issue found here on stack overflow:

https://stackoverflow.com/questions/32175217/how-to-detect-write-failure-in-asyncio

However, brief experimentation with adding a `await asyncio.sleep(0)` to [the only place in our code that calls `writer.drain()`](https://github.com/ethereum/py-evm/blob/32cc41d5315003289b1c86b31680f2501894da06/p2p/server.py#L239) didn't fix anything....

pipermerriam:
Here is a call-site which throws this error:

```python
Traceback (most recent call last):
  File "/Users/piper/sites/py-evm/p2p/service.py", line 54, in run
    await self._run()
  File "/Users/piper/sites/py-evm/trinity/plugins/builtin/tx_pool/pool.py", line 72, in _run
    await self._handle_tx(peer, msg)
  File "/Users/piper/sites/py-evm/trinity/plugins/builtin/tx_pool/pool.py", line 95, in _handle_tx
    receiving_peer.sub_proto.send_transactions(filtered_tx)
  File "/Users/piper/sites/py-evm/p2p/eth.py", line 193, in send_transactions
    self.send(header, body)
  File "/Users/piper/sites/py-evm/p2p/protocol.py", line 132, in send
    self.peer.send(header, body)
  File "/Users/piper/sites/py-evm/p2p/peer.py", line 479, in send
    self.writer.write(self.encrypt(header, body))
[...]

PR #1018, opened 2018-07-16
Fix transaction pool send loop logic

pipermerriam:
### What was wrong?

- The tx pool loop was *probably* causing us to occasionally time out during sync
- The tx pool loop was trying to send to peers which had already disconnected causing lots of console warnings

### How was it fixed?

- made the loop yield for each iteration
- changed how peers are iterated over to account for in-flight disconnections.
- fixed some places where `peer.disconnect` was not being *awaited*

#### Cute Animal Picture

TODO

### DIFF
--- a/trinity/plugins/builtin/tx_pool/pool.py
+++ b/trinity/plugins/builtin/tx_pool/pool.py
@@ -87,9 +87,6 @@ class TxPool(BaseService, PeerPoolSubscriber):
             if len(filtered_tx) == 0:
                 continue
 
-            if receiving_peer.is_closing:
-                continue
-
             self.logger.trace(
                 'Sending %d transactions to %s',
                 len(filtered_tx),
--- a/trinity/plugins/builtin/tx_pool/pool.py
+++ b/trinity/plugins/builtin/tx_pool/pool.py
@@ -87,7 +87,14 @@ class TxPool(BaseService, PeerPoolSubscriber):
             if len(filtered_tx) == 0:
                 continue
 
-            self.logger.trace('Sending %d transactions to %s', len(filtered_tx), receiving_peer)
+            if receiving_peer.is_closing:
+                continue
+
+            self.logger.trace(
+                'Sending %d transactions to %s',
+                len(filtered_tx),
+                receiving_peer,
+            )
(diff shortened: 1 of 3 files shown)


pipermerriam:
link: https://github.com/ethereum/py-evm/issues/1016

pipermerriam:
Even after rebaseing on #1022 this is still throwing lots of the `selector_events` errors.

gsalgado:
I wonder if it could be something other than the txpool that is causing them... Do you still get them when running with the txpool disabled as well?

gsalgado:
I went through my old logs and I have lots of those warnings as well. They all happen immediately after we try to send a tx to a peer (I get a `DEBUG pool  Sending transactions to ETHPeer(...)` log entry before each of those), so I guess there's no question it is the txpool that's causing them. However, I've been running a full node on ropsten for more than 1h (with ~15 peers) and didn't get any more socket.send() errors

pipermerriam:
So far it is looking like disabling the tx pool removes the errors so loose confirmation that tx pool is indeed the culprit.  I'm now trying to figure out how the errors are still leaking through as I'm still seeing them in large quantities.

pipermerriam:
Alright, bef4cd0 seems to have cleaned up the rest of those errors.   Lemme clean this up and it'll be ready for review.

trinity/plugins/builtin/tx_pool/pool.py
```diff
@@ -87,6 +87,9 @@ def register_peer(self, peer: BasePeer) -> None:
             if len(filtered_tx) == 0:
                 continue
 
+            if receiving_peer.is_closing:
```
gsalgado: is_closing is a method

gsalgado: Also, as I mentioned in https://github.com/ethereum/py-evm/issues/1016#issuecomment-405488411 I believe the real issue is in `PeerPool` and once it's fixed there this will not be needed

gsalgado: With #1022 this should not be needed because `ConnectedPeersIterator.__anext__()` will perform this exact check for every peer it returns, and here there's nowhere where we yield control between the check in `__anext__()` and this check

PR #1022, opened 2018-07-16
PeerPool is now an asynchronous iterable

gsalgado:
This way callsites are less likely to get a list of peers and hold them while making blocking calls, as these peers may disconnect in the meantime.

Closes: #1016

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -11,6 +11,8 @@ from abc import (
 
 from typing import (
     Any,
+    AsyncIterable,
+    AsyncIterator,
     cast,
     Dict,
     Iterator,
@@ -333,6 +335,10 @@ class BasePeer(BaseService):
         self.reader.feed_eof()
         self.writer.close()
 
+    @property
+    def is_closing(self) -> bool:
+        return self.writer.transport.is_closing()
+
     async def _cleanup(self) -> None:
         self.close()
 
@@ -476,6 +482,9 @@ class BasePeer(BaseService):
     def send(self, header: bytes, body: bytes) -> None:
         cmd_id = rlp.decode(body[:1], sedes=sedes.big_endian_int)
         self.logger.trace("Sending msg with cmd_id: %s", cmd_id)
+        if self.is_closing:
+            self.logger.error("Attempted to send %s to disconnected peer", cmd_id)
+            return
         self.writer.write(self.encrypt(header, body))
 
     async def disconnect(self, reason: DisconnectReason) -> None:
@@ -708,7 +717,7 @@ class PeerPoolSubscriber(ABC):
             peer_pool.unsubscribe(self)
 
 
-class PeerPool(BaseService):
+class PeerPool(BaseService, AsyncIterable[BasePeer]):
     """
     PeerPool maintains connections to up-to max_peers on a given network.
     """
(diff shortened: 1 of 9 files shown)


p2p/peer.py
```diff
@@ -879,6 +891,24 @@ def get_peers(self, min_td: int) -> List[BasePeer]:
                 break
 
 
+class ConnectedPeersIterator(AsyncIterator[BasePeer]):
+
+    def __init__(self, peers: Tuple[BasePeer, ...]) -> None:
+        self.iter = iter(peers)
+
+    async def __anext__(self) -> BasePeer:
```
cburgdorf: Could this be written as an [Asynchronous Generator](https://www.python.org/dev/peps/pep-0525/#improvements-over-asynchronous-iterators) instead of an `AsyncIterator`?

They seem to be lighter in syntax while being 2.3x as fast.

gsalgado: Maybe, but I don't think it's worth spending time on that given this is not a performance-critical code, given the number of simultaneous peers we can handle

cburgdorf: Yes, it's mostly the lighter syntax that is appealing.

pipermerriam: Second to the lighter syntax.  I think there is less cognitive overhead in the generator syntax since it can be implemented inline in the `PeerPool.peers` property.

tests/p2p/test_peer.py
```diff
@@ -52,6 +53,30 @@ def test_sub_protocol_selection():
...
+    assert peer3 in peers
+
+    peers = []
+    asyncio.ensure_future(peer2.disconnect(DisconnectReason.disconnect_requested))
+    async for peer in pool:
+        peers.append(peer)
+
+    assert len(peers) == 2
+    assert peer1 in peers
+    assert peer2 not in peers
```
pipermerriam: Should there be an assertion that `peer2` is still found in the connected nodes?  Otherwise we aren't testing what we think we're testing if the disconnect has already been processed.

p2p/peer.py
gsalgado: Not sure I follow. How would that look like?

tests/p2p/test_peer.py
gsalgado: It won't be in `connected_nodes`, but that's not really what we want to ensure with this test. The point of this change is to force callsites to iterate asynchronously over the peers in the pool so that on every iteration we can process disconnections, which in turn allow us to skip stale items from the _snapshot_ we get when the iteration starts -- and this test is ensuring we're skipping those stale items from the snapshot.

p2p/peer.py
```diff
@@ -333,6 +335,9 @@ def close(self) -> None:
         self.reader.feed_eof()
         self.writer.close()
 
+    def is_closing(self) -> bool:
```
pipermerriam: Just realized I already made the mistake that this pattern causes (at least for me).

I have grown to dislike functions that return booleans.  As you pointed out in my PR, I incorrectly used this as a property rather than calling it.  Can we instead convert this to a `property`?

This way if we *incorrectly* try calling it we'll get an error, where-as, with this pattern, if we incorrectly treat it like a property, it silently always returns truthy.

pipermerriam: I've updated this in #1018 so if we want to just merge that as the union of these two I'm good with that.

p2p/peer.py
pipermerriam: I tinkered with it in my branch and it disn't seem like it was doable in a way that is forward compatible with the planned changes in python 3.7

disregard.

p2p/peer.py
gsalgado: I implemented it as a method to match the underlying one, and because I intended to get rid of this and use `StreamWriter.is_closing()` directly once we migrate to 3.7 (the version in which it was introduced), but I guess that will be a while and yeah, I agree this makes more sense as a property, so I'll change it

p2p/peer.py
cburgdorf: @pipermerriam I'm just curious what kind of changes in Python 3.7 would prevent that?

pipermerriam: I think I got my reasoning backwards but I'm honestly not sure right now.  I think I was wrong and that we *could* implement this as follows directly on the `PeerPool` class

```python
async def __aiter__(self):
    for peer in ...:
        ...
```

Issue #1021, opened 2018-07-16
install py-evm fatal error: secp256k1_ecdh.h: no such file


johnys222:
* py-evm Version: 0.2.0a26
* OS: linux ubuntu 16.04 64bit
* Python Version (python --version):3.6.5
* Environment (output of `pip freeze`):pip 8.1.1 from /usr/lib/python3/dist-packages (python 3.6)

### What is wrong?
when i use souce code to install
use  command "python3 setup.py install to install"  and Get the following error during the installation process
like this:
build/temp.linux-x86_64-3.6/_libsecp256k1.c:493:28: fatal error: secp256k1_ecdh.h: no such file
compilation terminated.
error: Setup script exited with error: command 'x86_64-linux-gnu-gcc' failed with exit status 1
and  I found a solution to use command pip install -r requirements.txt but in source files I did not find this file  how to fix it.？ thanks

cburgdorf:
Are you looking to work on the source of just want to run a Trinity node?

If you like to install from source, I recommend to follow the [contributing guide](http://py-evm.readthedocs.io/en/latest/contributing.html). If you just want to run a Trinity node, the [quickstart](http://py-evm.readthedocs.io/en/latest/quickstart.html) would best answer this.

Please let me know if this helps or if there's anything else we could do to get you unstuck.

pipermerriam:
Can you try this: https://github.com/JoinMarket-Org/joinmarket/wiki/Installing-the-libsecp256k1-binding#installing-the-libsecp256k1-binding-on-linuxosx

johnys222:
No, I don't need just want to run a Trinity node, I just want to test a python script, this script needs to rely on the py-evm library, so I want to install this library

pipermerriam:
@johnys222 the link I sent you is instructions on how to install the necessary system dependencies to fix the error you encountered.

johnys222:
@pipermerriam  thanks  let me try it

johnys222:
@cburgdorf  I still don't quite understand how to install py-evm? I have downloaded the source code to my local, and have already created a pure Python environment, but I don't understand how to install the py-evm library, because I am a newbie, thank you

cburgdorf:
@johnys222 gotcha. So you basically want to write your own program that uses `py-evm` as a library. You should be able to just specify `py-evm` as a `install_requires` in your `setup.py` file and then install dependencies through pip.

[...]

PR #1029, opened 2018-07-17
Add simple guide to show how to use Py-EVM as a library

cburgdorf:
### What was wrong?

So, far, we don't have a guide that shows how to use `Py-EVM` as an actual library to build things. Frankly, most of this isn't really `Py-EVM` specific and falls more into a "How to get started with Python" category but I feel it doesn't hurt to be as helpful as possible especially when it comes to onboarding new peoplel to the ecosystem.

### How was it fixed?

Added simple guide that shows how to clone a boilerplate repo, add `Py-EVM` as a dependency, install it, and code a very very simple script that uses Py-EVM

**The code is written based on the latest release. Might be worth cutting a new release soon and then rebasing this PR (as opposed to merging and breaking silently soon** :sweat_smile: 

#### Cute Animal Picture

### DIFF
--- a/docs/api/eth/db/api.db.chain.rst
+++ b/docs/api/eth/db/api.db.chain.rst
@@ -12,9 +12,3 @@ ChainDB
 
 .. autoclass:: eth.db.chain.ChainDB
   :members:
-
-AsyncChainDB
-------------
-
-.. autoclass:: eth.db.chain.AsyncChainDB
-  :members:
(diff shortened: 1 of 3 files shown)


cburgdorf:
Happy to hear this guide already helped @johnys222 to move forward (#1021).

Initially, I was tempted to keep this PR open until we cut a new Py-EVM release (because `eth` namespace changed) but I realize that this guide won't break as the py-evm version is locked down. So while it may become outdated, it won't break at least. Merging.

PR #1043, opened 2018-07-18
Convert IPCServer to a BaseService

### LINKED ISSUE #689: Consider using p2p.BaseService in trinity.rpc.ipc.IPCServer

carver:
### What was wrong?

Fixes #689

Also, the ipc socket file is left dangling when the service is cleaned up.

### How was it fixed?

- convert to BaseService
- can set custom asyncio loop on BaseService
- cancel a service running on a different thread
- remove ipc socket file on server shutdown
- some helpful contextual info: show the target data folder that fix-unclean-shutdown is using to fix (I was trying to clean a bad `trinity --ropsten` with a `trinity fix-unclean-shutdown`. I needed `trinity --ropsten fix-unclean-shutdown` but that wasn't obvious without a little extra logging)

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -2,7 +2,6 @@ from abc import ABC, abstractmethod
 import asyncio
 import functools
 import logging
-import time
 from typing import (
     Any,
     Callable,
@@ -77,6 +76,7 @@ class BaseService(ABC, CancellableMixin):
                 # XXX: Only added to help debug https://github.com/ethereum/py-evm/issues/1023;
                 # should be removed eventually.
                 self.logger.warn("%s finished but had no finished_callback", self)
+            self.logger.debug("%s halted cleanly", self)
 
     def run_child_service(self, child_service: 'BaseService') -> 'asyncio.Future[Any]':
         """
@@ -122,6 +122,15 @@ class BaseService(ABC, CancellableMixin):
     @property
     def is_running(self) -> bool:
         return self._run_lock.locked()
+
+    async def threadsafe_cancel(self) -> None:
+        """
+        Cancel service in another thread. Block until service is cleaned up.
+
+        :param poll_period: how many seconds to wait in between each check for service cleanup
+        """
+        asyncio.run_coroutine_threadsafe(self.cancel(), loop=self.loop)
+        await asyncio.wait_for(self.cleaned_up.wait(), timeout=self._wait_until_finished_timeout)
 
     @abstractmethod
     async def _run(self) -> None:
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 import asyncio
 import functools
 import logging
+import time
 from typing import (
     Any,
     Callable,
@@ -22,7 +23,10 @@ class BaseService(ABC, CancellableMixin):
     # Number of seconds cancel() will wait for run() to finish.
     _wait_until_finished_timeout = 5
 
-    def __init__(self, token: CancelToken=None) -> None:
+    # the custom event loop to run in, or None if the default loop should be used
+    loop: asyncio.AbstractEventLoop = None
+
+    def __init__(self, token: CancelToken=None, loop: asyncio.AbstractEventLoop = None) -> None:
         if self.logger is None:
             self.logger = cast(
                 TraceLogger, logging.getLogger(self.__module__ + '.' + self.__class__.__name__))
@@ -31,7 +35,9 @@ class BaseService(ABC, CancellableMixin):
         self.cleaned_up = asyncio.Event()
         self._child_services = []
 
-        base_token = CancelToken(type(self).__name__)
+        self.loop = loop
+        base_token = CancelToken(type(self).__name__, loop=loop)
+
         if token is None:
             self.cancel_token = base_token
         else:
--- a/trinity/nodes/base.py
+++ b/trinity/nodes/base.py
@@ -8,7 +8,6 @@ from threading import Thread
 from typing import (
     List,
     Type,
-    Union
 )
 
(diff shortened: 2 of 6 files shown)


p2p/service.py
```diff
@@ -22,7 +23,10 @@ class BaseService(ABC, CancellableMixin):
     # Number of seconds cancel() will wait for run() to finish.
     _wait_until_finished_timeout = 5
 
-    def __init__(self, token: CancelToken=None) -> None:
+    # the custom event loop to run in, or None if the default loop should be used
+    target_loop: asyncio.AbstractEventLoop = None
```
pipermerriam: Thoughts on just naming this `loop`?  It's pretty common across all of the `asyncio` libraries for this to just be named `loop`

p2p/service.py
```diff
@@ -31,7 +35,12 @@ def __init__(self, token: CancelToken=None) -> None:
         self.cleaned_up = asyncio.Event()
         self._child_services = []
 
-        base_token = CancelToken(type(self).__name__)
+        if loop is None:
+            base_token = CancelToken(type(self).__name__)
+        else:
+            self.target_loop = loop
+            base_token = CancelToken(type(self).__name__, loop=loop)
```
pipermerriam: I think you can drop this `if/else`.  Passing in `loop=None` to `CancelToken` should be equivalent to not including the argument.

trinity/rpc/ipc.py
```diff
@@ -128,32 +130,32 @@ def strip_non_json_prefix(raw_request: str) -> Tuple[str, str]:
...
 
-    async def run(self, loop: asyncio.AbstractEventLoop=None) -> None:
-        ipc_path = str(self.ipc_path)
-        self.cancel_token = CancelToken('IPCServer', loop=loop)
+    async def _run(self) -> None:
         self.server = await asyncio.start_unix_server(
             connection_handler(self.rpc.execute, self.cancel_token),
-            ipc_path,
-            loop=loop,
+            str(self.ipc_path),
```
pipermerriam: reason for the `str` casting?

trinity/nodes/base.py
```diff
@@ -116,36 +114,43 @@ def notify_resource_available(self) -> None:
...
 
         await self.get_p2p_server().run()
 
     async def _cleanup(self) -> None:
-        if isinstance(self._ipc_server, IPCServer):
-            await self._ipc_server.stop()
+        # IPC Server requires special handling because it's running in its own loop & thread
+        if self.has_ipc_server:
+            self._ipc_server.threadsafe_cancel()
+            # trigger ipc server thread to close out
```
pipermerriam: Maybe clarify a bit as it wasn't immediately clear what this meant.

> By stopping the event loop for the ipc server the thread running the server will exit.

Something like that?

p2p/service.py
```diff
@@ -112,6 +121,24 @@ def run_child_service(self, child_service: 'BaseService') -> 'asyncio.Future[Any
     def is_running(self) -> bool:
         return self._run_lock.locked()
 
+    def threadsafe_cancel(self, poll_period: float=0.1) -> None:
```
pipermerriam: I was surprised that this wasn't a coroutine.  Is there a reason for that?

carver: Hm... I could try it. It requires being cautious to make sure everything is scheduling where you intend it to: that you are `await`-ing in one loop & thread, but canceling and awaiting on a different loop and thread.

p2p/service.py
```diff
@@ -117,6 +124,24 @@ def run_child_service(self, child_service: 'BaseService') -> 'asyncio.Future[Any
...
+        """
+        Cancel service in another thread. Block until service is cleaned up.
+
+        :param poll_period: how many seconds to wait in between each check for service cleanup
+        """
+        asyncio.run_coroutine_threadsafe(self.cancel(), loop=self.loop)
+
+        start_time = time.monotonic()
+        while time.monotonic() - start_time < self._wait_until_finished_timeout:
+            if not self.cleaned_up.is_set():
```
pipermerriam: I think you can drop the whole loop here and replace it with:

```
asyncio.wait_for(self.cleaned_up.wait(), timeout=self._wait_until_finished_timeout)
```

This should allow `asyncio` to manage the timeout.

carver: Hm, not that the current implementation necessarily solves this, but should we be worried that [`asyncio.Event` is marked as "not thread safe"](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event)?

I tried it out, and seems to work okay. As long as it's safe, it's clearly better (less code).

p2p/service.py
```diff
@@ -22,7 +22,10 @@ class BaseService(ABC, CancellableMixin):
     # Number of seconds cancel() will wait for run() to finish.
     _wait_until_finished_timeout = 5
 
-    def __init__(self, token: CancelToken=None) -> None:
+    # the custom event loop to run in, or None if the default loop should be used
+    loop: asyncio.AbstractEventLoop = None
```
cburgdorf: @carver @pipermerriam doesn't assigning `None` on the class level introduce the same issue as discussed here? https://github.com/ethereum/py-evm/issues/1012 So, essentially making all instances run in the same `loop` which may not always be what is assumed?

pipermerriam: Yes, but that is the desired behavior.  If you want a service running in a different event loop you have to be explicit about that.  Generally, we should only have a single event loop per process.

PR #1045, opened 2018-07-18
Fix header number sequence generation

### LINKED ISSUE #1036: Broken algorithm for generating block header numbers
### What is wrong?

https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/chain.py#L775-L782

The algorithm linked above and included here is not generating block numbers according to protocol.

```python
if reverse:
    low = max(0, block_number - limit)
    high = block_number + 1
    block_numbers = reversed(range(low, high, step))
else:
    low = block_number
    high = block_number + limit
block_numbers = range(low, high, step)
```

For example, if requesting `5` headers starting at block number `0`, with a skip of `1`:

- expected correct: `0, 2, 4, 6, 8`
- actual: `0, 2, 4` (doesn't return 5 headers)

### How can it be fixed

Need to fix how the limits are computed to take into account the skip size so that we still return the appropriate number of headers.

carver:
### What was wrong?

Fixes #1036 

### How was it fixed?

- Extract to a utility for easier testing
- Added the example from the issue, plus a few corner cases

Could probably do a few more cases, but ran out of time today. If any obviously missing cases jump out, please comment!

#### Cute Animal Picture

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -46,6 +46,7 @@ from p2p.rlp import BlockBody
 from p2p.service import BaseService
 from p2p.utils import (
     get_process_pool_executor,
+    sequence_builder,
 )
 
 
@@ -688,14 +689,14 @@ class PeerRequestHandler(CancellableMixin):
             self.logger.debug(
                 "Peer requested starting header %r that is unavailable, returning nothing",
                 block_number_or_hash)
-            block_numbers = []
+            block_numbers = tuple()  # type: ignore
 
         headers = [header async for header in self._generate_available_headers(block_numbers)]
         return headers
 
     async def _get_block_numbers_for_request(
             self, block_number_or_hash: Union[int, bytes], max_headers: int,
-            skip: int, reverse: bool) -> List[BlockNumber]:
+            skip: int, reverse: bool) -> Tuple[BlockNumber]:
         """
         Generates the block numbers requested, subject to local availability.
         """
@@ -713,19 +714,10 @@ class PeerRequestHandler(CancellableMixin):
             )
 
         limit = min(max_headers, eth.MAX_HEADERS_FETCH)
-        step = skip + 1
-        if reverse:
-            low = max(0, block_number - limit)
-            high = block_number + 1
-            block_numbers = reversed(range(low, high, step))
-        else:
-            low = block_number
-            high = block_number + limit
-            block_numbers = iter(range(low, high, step))  # mypy thinks range isn't iterable
-        return list(block_numbers)
+        return sequence_builder(block_number, limit, skip, reverse)
 
     async def _generate_available_headers(
-            self, block_numbers: List[BlockNumber]) -> AsyncGenerator[BlockHeader, None]:
+            self, block_numbers: Tuple[BlockNumber]) -> AsyncGenerator[BlockHeader, None]:
         """
         Generates the headers requested, halting on the first header that is not locally available.
         """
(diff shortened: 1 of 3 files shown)


p2p/utils.py
```diff
@@ -33,6 +36,25 @@ def get_devp2p_cmd_id(msg: bytes) -> int:
     return rlp.decode(msg[:1], sedes=rlp.sedes.big_endian_int)
 
 
+@to_tuple
+def sequence_builder(start_number: int, max_length: int, skip: int, reverse: bool) -> Iterator[int]:
+    remaining_elements = max_length
+    next_element = start_number
+    step_size = skip + 1
```
pipermerriam: We could drop the `if reverse` in the loop if you changed this to be:

```
step_size = -1 * (skip + 1) if reverse else (skip + 1)
```

p2p/utils.py
```diff
@@ -33,6 +36,25 @@ def get_devp2p_cmd_id(msg: bytes) -> int:
...
+
+    while remaining_elements > 0:
```
pipermerriam: We could drop the loop mutation and just do `for _ in range(max_length)`.  That should drop the inner `iff/else` as well as the need to decrement `remaining_elements` in each loop.

p2p/chain.py
```diff
@@ -713,19 +714,10 @@ def __init__(self, db: 'AsyncHeaderDB', logger: TraceLogger, token: CancelToken)
             )
 
         limit = max(max_headers, eth.MAX_HEADERS_FETCH)
```
gsalgado: This should actually be `min(max_headers, eth.MAX_HEADERS_FETCH)`

Maybe add a test to ensure we return the requested `max_headers`. That is important because currently when geth asks us for the DAO fork block header we respond with 192 headers and it doesn't like that

p2p/utils.py
```diff
@@ -33,6 +36,25 @@ def get_devp2p_cmd_id(msg: bytes) -> int:
...
+        if next_element < 0:
```
pipermerriam: This should also check that `next_element` is not greater than the maximum UINT256 size.

p2p/utils.py
carver: :+1: for dropping the mutation, that was embarrassing : P

We still need the if check for dropping below 0, for example:

```py
@pytest.mark.parametrize(
    'start_num, max_length, skip, reverse, expected',
    (
        ...,
        (1, 9, 0, True, (1, 0)),
    ),
)       
def test_sequence(start_num, max_length, skip, reverse, expected):
    assert sequence_builder(start_num, max_length, skip, reverse) == expected
```

Are you talking about a different `iff/else`?

p2p/chain.py
carver: > This should actually be min(max_headers, eth.MAX_HEADERS_FETCH

fix inbound...

> Maybe add a test to ensure we return the requested max_headers.

I'm not really sure where to start on writing that test. It seems like it belongs in `tests/p2p/test_sync.py` somewhere, but that reads a full sync integration test. I didn't see anything low-level like:
- trigger some request in the client
- assert that request received in the server

Any suggestions?

p2p/utils.py
pipermerriam: > We still need the if check for dropping below 0, for example:

I don't know what I was talking about... carry on and :shipit:

p2p/chain.py
gsalgado: Needed some refactoring, but pretty straightforward: #1051

PR #1046, opened 2018-07-19
ERC20 benchmark refactored for composition

williambannas:
### What was wrong?
enhancement, to make code more maintainable in the future

@cburgdorf @carver is this what you two were looking for?

### How was it fixed?
refactored to no longer use inheritance, and instead try to use composition


carver:
Yeah, heading a great direction. :+1: 

The dependency on the `ERC20BenchmarkConfig` in the setup and benchmark run methods stands out to me. Let's see if we can remove that dependency completely. Then each method gets access only to the pieces it needs, and returns the pieces needed by the next step. Also, let's see if we can find a way to remove state storage on the `config` instance between setup and benchmark run.

One option (which would require changes to the core benchmark runner) would be something like:

```py
def transfer_benchmark(w3, chain):
  # setup
  undeployed_contract = make_contract_for_deployment(w3)
  token_contract = erc20_deploy(w3, chain, undeployed_contract)
  txn_data = w3...buildTransaction().data
  yield

  # benchmark
  erc20_transfer(chain, txn_data, token_contract.address)
  yield

  # cleanup
  # (but no cleanup to do on this one, I think)
  yield

  # never hit this point
```

The test runner would roughly do:

```
def benchmark_run(benchmark: Callable):
  test_lifecycle = executor(w3, chain)
  next(test_lifecycle)  # setup

  start_clock ...
  next(test_lifecycle)  # run
  end_clock ...

  next(test_lifecycle)  # cleanup
```

cburgdorf:
It seems that this PR has unaddressed feedback and development has not progressed for a while. I'm closing this for now and created an issue for the idea to change the runner.

https://github.com/ethereum/py-evm/issues/1306

Feel free to continue and re-open at any time.

Issue #1025, opened 2018-07-17
Do not ignore data requests while performing DAO fork check


gsalgado:
### What is wrong?

Currently, while we're waiting for a remote to send us the DAO fork block headers in order to determine if we're on the same side of the fork, we ignore any data requests they make. That seems to cause them treat us as a useless peer and disconnect.

It'd explain why when running on ropsten trinity can keep a high peer count but on mainnet most nodes disconnect shortly after the initial handshake.

### How can it be fixed

Need to handle data requests while performing the DAO fork check. Or maybe keep ignoring them during the DAO fork check, but re-queue them once the check is completed

gsalgado:
@pipermerriam https://github.com/gsalgado/py-evm/tree/misc has a WIP fix for this. I've been running it on mainnet and it's syncing reliably keeping a peer count of 25, but it still hasn't synced up to the DAO fork block and I know geth will treat peers differently depending on whether their announced chain HEAD is past that, so it'd be great if you could give it a spin in case you have a DB synced past the DAO fork block

pipermerriam:
I unfortunately do not have such a data dir.  I'm working on syncing a trinity node now but it's slow going.

pipermerriam:
@gsalgado Going to try and ingest this issue tomorrow.

Side note, we should cache information about which side of the fork peers are on.

gsalgado:
Another way of fixing this would be to let peers be added to `PeerPool` without the DAO fork check, and then perform the check in `ChainSyncer` by keeping a list of peers for which the check is pending and having its `_handle_msg_loop()` do the following for every msg read from the queue:
   1. If peer has passed DAO fork check, process msg normally  
   2. Else if cmd is not BlockHeaders, add it to the list of msgs received by that peer and return
   3. Else if cmd is BlockHeaders
    3.1 If DAO fork check is successful, handle all previously received messages from the peer
    3.2 Else disconnect from peer and discard all previous messages

PR #1048, opened 2018-07-19
p2p: Perform DAO fork check only when peer is added to PeerPool

gsalgado:
Also, instead of ignoring non-BlockHeaders messages received while the DAO
fork check is pending, we store them and re-add to subscribers' msg queues
when the fork check is successful and the peer is actually added to the
pool

Closes: #1025

### DIFF
--- a/p2p/protocol.py
+++ b/p2p/protocol.py
@@ -1,5 +1,6 @@
 import logging
 import struct
+from abc import ABC, abstractmethod
 from typing import (
     Any,
     Dict,
@@ -14,6 +15,7 @@ import rlp
 from rlp import sedes
 
 from eth.constants import NULL_BYTE
+from eth.rlp.headers import BlockHeader
 
 from p2p.utils import get_devp2p_cmd_id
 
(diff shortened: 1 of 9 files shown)


gsalgado:
Running this on mainnet, with a chain synced past the DAO fork block, and it's been syncing at a good pace, maxed out on connected peers (25) and very infrequent peer disconnections. The approach I used here may feel a bit hackish but I think it's better than [the alternative I could think of](https://github.com/ethereum/py-evm/issues/1025#issuecomment-406216692)

p2p/protocol.py
```diff
@@ -135,6 +136,12 @@ def __repr__(self) -> str:
         return "(%s, %d)" % (self.name, self.version)
 
 
+class BaseBlockHeaders(Command):
+
+    def extract_headers(self, msg: _DecodedMsgType) -> Tuple[BlockHeader, ...]:
```
pipermerriam: Should this be decorated with an `@abstractmethod`?

Issue #1031, opened 2018-07-17
TxPool does nothing when in light mode


gsalgado:
### What is wrong?

The TxPool reads every msg delivered into its queue but only handles the `eth.Transactions` ones, but those can only come from an `ETHPeer`, which is only used when running a full node.

### How can it be fixed

If we want to handle transactions when in light mode, we need to implement the `SendTx` command in `LESProtocol` and handle that in TxPool. Otherwise we should just disable the TxPool when in light mode

PR #1049, opened 2018-07-19
Add integration tests for Trinity

cburgdorf:
### What was wrong?

We currently do not have any integration tests for Trinity, that really test it on the command line level. That is, run Trinity the same way as a user would run it and check whether it behaves the way we assume it should behave.

This has biten us over and over as it is **very hard** to guarantee to ship software that works™ when one has to rely on manual human testing for these things.

### How was it fixed?

This adds some new testing facilities to run Trinity, the same way a user would do, and then do certain assertions about the things it responds.

In particular we test that,

- Trinity logs out sequence of expected messages
- Trinity does not log out `Traceback (most recent call last)` in the first 20 seconds of booting

These assertions are provided for the following invocations

```
trinity,
trinity --light
trinity --ropsten
trinity --light --ropsten
```

Additionally, there are tests that run Trinity as a full node with the tx pool enabled and observe whether it is starting.

**I've ran these tests on CI over and over again and haven't seen a single sign of flakiness**

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/conftest.py
+++ b/tests/trinity/conftest.py
@@ -29,7 +29,7 @@ def xdg_trinity_root(monkeypatch, tmpdir):
     """
     Ensure proper test isolation as well as protecting the real directories.
     """
-    dir_path = tmpdir.mkdir('xdg_trinity_root')
+    dir_path = tmpdir.mkdir('trinity')
     monkeypatch.setenv('XDG_TRINITY_ROOT', str(dir_path))
 
     assert not is_under_path(os.path.expandvars('$HOME'), get_xdg_trinity_root())
--- a/tests/trinity/integration/test_boot.py
+++ b/tests/trinity/integration/test_boot.py
@@ -1,0 +1,123 @@
+import pytest
+
+from trinity.tools.async_process_runner import AsyncProcessRunner
+from trinity.utils.async_iter import (
+    contains_all
+)
+
+
+# IMPORTANT: Test names are intentionally short here because they end up
+# in the path name of the isolated Trinity paths that pytest produces for
+# us.
+# e.g. /tmp/pytest-of-circleci/pytest-0/popen-gw3/test_light_boot_comman0/xdg/mainnet/jsonrpc.ipc)
+#
+# However, UNIX IPC paths can only be 100 chars which means long paths
+# *WILL* break these tests. See: https://unix.stackexchange.com/q/367008
+
+# This fixture provides a tear down to run after each test that uses it.
+# This ensures the AsyncProcessRunner will never leave a process behind
+@pytest.fixture(scope="function")
+def async_process_runner():
+    runner = AsyncProcessRunner(
+        # This allows running pytest with -s and observing the output
+        debug_fn=lambda line: print(line)
+    )
+    yield runner
+    runner.kill()
+
+# Great for debugging the AsyncProcessRunner
+# @pytest.mark.asyncio
+# async def test_ping(async_process_runner):
+#     await async_process_runner.run(['ping', 'www.google.de'])
+#     assert await contains_all(async_process_runner.iterate_stdout(), ['byytes from'])
+
+
+@pytest.mark.parametrize(
+    'command',
+    (
+        ['trinity'],
+        ['trinity', '--ropsten'],
+    )
+)
+@pytest.mark.asyncio
+async def test_full_boot(async_process_runner, command):
+    # UPNP discovery can delay things, we use a timeout longer than the discovery timeout
+    await async_process_runner.run(command, timeout_sec=40)
+    assert await contains_all(async_process_runner.stderr, {
+        "Started DB server process",
+        "Started networking process",
+        "Running server",
+        "IPC started at",
+    })
+
+
+@pytest.mark.parametrize(
+    'command',
+    (
+        ['trinity', '--tx-pool'],
+        ['trinity', '--tx-pool', '--ropsten'],
+    )
+)
+@pytest.mark.asyncio
+async def test_txpool_full_boot(async_process_runner, command):
+    # UPNP discovery can delay things, we use a timeout longer than the discovery timeout
+    await async_process_runner.run(command, timeout_sec=40)
+    assert await contains_all(async_process_runner.stderr, {
+        "Started DB server process",
+        "Started networking process",
+        "Running Tx Pool",
+        "Running server",
+        "IPC started at",
+    })
+
+
+@pytest.mark.parametrize(
+    'command',
+    (
+        ['trinity', '--light'],
+        ['trinity', '--light', '--ropsten'],
+    )
+)
+@pytest.mark.asyncio
+async def test_light_boot(async_process_runner, command):
+    # UPNP discovery can delay things, we use a timeout longer than the discovery timeout
+    await async_process_runner.run(command, timeout_sec=40)
+    assert await contains_all(async_process_runner.stderr, {
+        "Started DB server process",
+        "Started networking process",
+        "IPC started at",
+    })
+
+
+@pytest.mark.parametrize(
+    'command',
+    (
+        # mainnet
+        ['trinity'],
+        ['trinity', '--light'],
+        # ropsten
+        ['trinity', '--ropsten'],
+        ['trinity', '--light', '--ropsten'],
+    )
+)
+@pytest.mark.asyncio
+async def test_does_not_throw(async_process_runner, command):
+    # This is our last line of defence. This test basically observes the first
+    # 20 seconds of the Trinity boot process and fails if Trinity logs any exceptions
+    lines_since_error = 0
+    await async_process_runner.run(command, timeout_sec=20)
+    async for line in async_process_runner.stderr:
+
+        # We detect errors by some string at the beginning of the Traceback and keep
+        # counting lines from there to be able to read and report more valuable info
+        if "Traceback (most recent call last)" in line and lines_since_error == 0:
+            lines_since_error = 1
+        elif lines_since_error > 0:
+            lines_since_error += 1
+
+        # Keep on listening for output for a maxmimum of 100 lines after the error
+        if lines_since_error >= 100:
+            break
+
+    if lines_since_error > 0:
+        raise Exception("Exception during Trinity boot detected")
--- a/trinity/tools/async_process_runner.py
+++ b/trinity/tools/async_process_runner.py
@@ -1,0 +1,65 @@
+import asyncio
+import os
+import signal
+from typing import (
+    AsyncIterable,
+    Awaitable,
+    Callable,
+    Iterable,
+)
+
+
+class AsyncProcessRunner():
+
+    def __init__(self, debug_fn: Callable[[bytes], None] = None) -> None:
+        self.debug_fn = debug_fn
+
+    @classmethod
+    async def create_and_run(cls, cmds: Iterable[str], timeout_sec: int=10) -> 'AsyncProcessRunner':
+        runner = cls()
+        await runner.run(cmds, timeout_sec)
+        return runner
+
+    async def run(self, cmds: Iterable[str], timeout_sec: int=10) -> None:
+        proc = await asyncio.create_subprocess_exec(
+            *cmds,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            # We need this because Trinity spawns multiple processes and we need to take down
+            # the entire group of processes.
+            preexec_fn=os.setsid
+        )
+        self.proc = proc
+        asyncio.ensure_future(self.kill_after_timeout(timeout_sec))
+
+    @property
+    async def stdout(self) -> AsyncIterable[str]:
+        async for line in self._iterate_until_empty(self.proc.stdout.readline):
+            yield line
+
+    @property
+    async def stderr(self) -> AsyncIterable[str]:
+        async for line in self._iterate_until_empty(self.proc.stderr.readline):
+            yield line
+
+    async def _iterate_until_empty(
+            self,
+            awaitable_bytes_fn: Callable[[], Awaitable[bytes]]) -> AsyncIterable[str]:
+
+        while True:
+            line = await awaitable_bytes_fn()
+            if self.debug_fn:
+                self.debug_fn(line)
+            if line == b'':
+                return
+            else:
+                yield line.decode('utf-8')
+
+    async def kill_after_timeout(self, timeout_sec: int) -> None:
+        await asyncio.sleep(timeout_sec)
+        self.kill()
+        raise TimeoutError('Killed process after {} seconds'.format(timeout_sec))
+
+    def kill(self) -> None:
+        # TODO: investigate if we could do this more gracefully
+        os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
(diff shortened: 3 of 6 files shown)


trinity/tools/async_process_runner.py
```diff
@@ -0,0 +1,64 @@
...
+                yield line.decode('utf-8')
+
+    async def kill_after_timeout(self, timeout_sec: int) -> None:
+        await asyncio.sleep(timeout_sec)
+        self.kill()
+        raise TimeoutError('Killed process after {} seconds'.format(timeout_sec))
+
+    def kill(self) -> None:
+        # TODO: investigate if we could do this more gracefully
+        os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
```
pipermerriam: look into using one of the utils from here: https://github.com/ethereum/py-evm/blob/master/trinity/utils/ipc.py#L18

trinity/tools/utils/async_generator_utils.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Check wether an ``AsyncIterable[str]`` contains all of the given keywords. The keywords
+    can be embedded in some larger string. Return ``True`` as soon as all keywords were matched.
+    If not all keywords were matched by the time that the async iterable is done, return ``False``.
+    """
+    left_keywords = keywords
+    async for line in async_gen:
+        for check in left_keywords:
+            if check in line:
+                left_keywords.remove(check)
```
pipermerriam: I'm a fan of an alternate pattern that you may like.

```
original_values = [1, 2, 3]
seen_values = set()
for thing in things:
    if thing in set(original_values).difference(seen_things):
        seen_things.add(thing)
```

Things I like about this.

- maintains list of what we've seen (explicitely)
- doesn't require duplicating and mutating the original list.

cburgdorf:
@pipermerriam afaik pytest is already doing that.

carver:
> afaik pytest is already doing that.

Do you know where/how that is happening?

cburgdorf:
@carver yes, I believe it is this `autouse` fixture.

https://github.com/ethereum/py-evm/blob/30d28d013f62866e8275fd4b44dc13a91e4e2614/tests/trinity/conftest.py#L27-L37

We already rely on this for other tests. Also, if you run these integration test with `-s` you can see that this is working by observing the Trinity output.

```
Trinity DEBUG log file is created at /tmp/pytest-of-cburgdorf/pytest-224/test_trinity_full_boot_command0/xdg_trinity_root/mainnet/logs/trinity.log
```

trinity/tools/async_process_runner.py
cburgdorf: I removed the comment because I think we can actually keep it as is, because:

1. This is using `killpg` to kill the entire process group which isn't what our utils do
2. I *think* there is nothing we can gain from sending a `SIGTERM` first because these Trinity instances are throwaway onetime instances anyway. That means, the only effect a `SIGTERM` followed by a `SIGKILL` after some seconds would have are slower running tests.
3. Since I'm using `asyncio.subprocess.Process` I can't use `join` to wait on the shutdown and I believe waiting in an async fashion would complicate the tear down of the tests.

Overall, I think it's good as it is :)

trinity/tools/async_process_runner.py
```diff
@@ -0,0 +1,63 @@
...
+
+    async def _iterate_until_empty(
+            self,
+            awaitable_bytes_fn: Callable[[], Awaitable[bytes]]) -> AsyncIterable[str]:
+
+        while True:
+            line = await awaitable_bytes_fn()
+            # Leaving this print here allows running pytest with -s and observing the output
+            # There's probably a better way to handle that
+            print(line)
```
cburgdorf: @pipermerriam @carver This bothers me but I've tried to use a logger and even with `logger.warning` and various kind of pytest arguments I wasn't able to see the output of the integration tests. I wonder if that's related to us logging on `STDERR`?

So, for now I kept the `print` here because it's really useful to be able to run these tests with `pytest -s` and observe the Trinity output. Any ideas are welcome :)

cburgdorf:
@pipermerriam @carver @gsalgado I believe that this is good enough as a start and can be given a review.

Notice that I've run these tests on CI (and locally obviously) over and over again and haven't seen a single sign of flakiness. 

The next thing that I'd like to do is to add a test that spins up two instance of Trinity, one with `attach` that does some basic `web3` interaction. There's probably also lots of value in providing some tests that connect to a geth node with a preset data directory and check for some more advanced things to happen but I haven't gotten that far yet. That said, these tests seem like a good start to improve on robustness already and should have caught many of the things that were biting us in the past.

For instance, I went back to https://github.com/ethereum/py-evm/commit/7d5f0413 and cherry-picked these integration tests on top to figure out if these tests would have caught that bug and they are indeed caught (by that test that runs for 20 seconds and fails if it detects any unhandled exceptions)

trinity/tools/utils/async_generator_utils.py
```diff
@@ -0,0 +1,21 @@
+from typing import (
+    AsyncIterable,
+    Set,
+)
+
+
+async def contains_all(async_gen: AsyncIterable[str], keywords: Set[str]) -> bool:
```
cburgdorf: Not sure if this file should rather live in `trinity/utils` as it really isn't directly coupled to `trinity/tools` in any way other then `trinity/tools` using it. That said, not sure how often I caught myself thinking "Damn, all I wish for was a function that takes an `AsyncIterable[str]` and returns a boolean if it contains all of some given keywords"....probably never :thinking:

trinity/tools/async_process_runner.py
```diff
@@ -0,0 +1,63 @@
...
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE,
+            # We need this because Trinity spawns multiple processes and we need to take down
+            # the entire group of processes.
+            preexec_fn=os.setsid
+        )
+        self.proc = proc
+        asyncio.ensure_future(self.kill_after_timeout(timeout_sec))
+
+    async def iterate_stdout(self) -> AsyncIterable[str]:
```
cburgdorf: I wonder if the following three methods were better named `yield_*` instead of `iterate_*` because now that I'm thinking about it, I find the `iterate` prefix rather wrong. A *consumer* would iterate over the produced `AsyncIterable` but that has nothing to do with the method itself which just asynchronously yield stuff. Thoughts?

pipermerriam: How about just making them properties which return async iterators.  Since the class is `AsyncProcessRunner` I think it's not too big of a stretch for the `runner.stdout` to return something that supports async iteration.

trinity/tools/async_process_runner.py
pipermerriam: Without seeing how you had logging setup, I'm not sure I can be of much help.

trinity/tools/async_process_runner.py
```diff
@@ -0,0 +1,60 @@
...
+            else:
+                yield line.decode('utf-8')
+
+    async def kill_after_timeout(self, timeout_sec: int) -> None:
+        await asyncio.sleep(timeout_sec)
+        self.kill()
+        raise TimeoutError('Killed process after {} seconds'.format(timeout_sec))
+
+    def kill(self) -> None:
+        os.killpg(os.getpgid(self.proc.pid), signal.SIGKILL)
```
pipermerriam: This seems like it should use `trinity.utils.ipc.kill_process_id_gracefully`

trinity/tools/async_process_runner.py
carver: I agree. Other options:
- `stdout_lines` <- my fave
- `stdout_buffer`
- `stdout_iterable`

trinity/tools/utils/async_generator_utils.py
pipermerriam: I don't think `trinity.tools` needs its own `utils` module.  I'm good with this living under `trinity.utils.async` or something similar.

trinity/tools/async_process_runner.py
carver: (with a little tweaking, in order to kill a whole process group, instead of a single process id)

pipermerriam: I now see that little `g` hanging around in there that I missed on first pass.

trinity/tools/async_process_runner.py
cburgdorf: @pipermerriam I created a new logger like this:

```python
def __init__(self):
    self.logger = logging.getLogger('trinity.tools.async_process_runner.AsyncProcessRunner')

and then used `self.logger.info(...) or even self.logger.warning(...)` instead of the `print`
```

trinity/tools/async_process_runner.py
cburgdorf: Did you see my past comment here? https://github.com/ethereum/py-evm/pull/1049#discussion_r203996646

tests/trinity/conftest.py
```diff
@@ -29,7 +29,7 @@ def xdg_trinity_root(monkeypatch, tmpdir):
     """
     Ensure proper test isolation as well as protecting the real directories.
     """
-    dir_path = tmpdir.mkdir('xdg_trinity_root')
+    dir_path = tmpdir.mkdir('xdg')
```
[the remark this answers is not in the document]
cburgdorf: I can give that a try, notice though that the actual directory in `/tmp` ends up much longer though:

```
/tmp/pytest-of-circleci/pytest-0/popen-gw3/test_light_boot_comman0/xdg
```

tests/trinity/integration/test_boot.py
```diff
@@ -0,0 +1,107 @@
...
+
+@pytest.mark.parametrize(
+    'command',
+    (
+        # mainnet
+        ['trinity'],
+        ['trinity', '--light'],
+        # ropsten
+        ['trinity', '--ropsten'],
+        ['trinity', '--light', '--ropsten'],
```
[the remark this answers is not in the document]
cburgdorf: There's a separate test for `--tx-pool` because we also want to check for tx pool related output in that case https://github.com/ethereum/py-evm/pull/1049/files#diff-4efbd71fd82c31c1cbb0ade62cadfa3aR65

trinity/tools/async_process_runner.py
cburgdorf: Ok, I got rid of this hardcoded `print` inside the `AsyncProcessRunner` and instead allowed passing a`debug_fn` in the `__init__`. That way, at least the `print` got moved into the tests directly.

cburgdorf:
Alright, I *think* this may be good enough to merge:

- I moved the helper function to `trinity.utils.async_iter`
- I got rid of that hardcoded `print` by allowing to pass a `debug_fn` in the `__init__` of `AsyncProcessRunner`
- I turned `iterate_stderr` and `iterate_stdout` into properties `stderr` and `stdout`
- I reduced the concurrency of **only these tests** to one process to not run into errors due to multiple Trinity instances competing for the same ports

Regarding, gracefully killing the Trinity process, let me echo what I wrote in a comment before.

I *think* there is nothing we can gain from sending a `SIGTERM` followed by a `SIGKILL` after some timeout. While this is of course the right way of shutting Trinity down for a normal user, I believe that for these *throwaway* instances, there is nothing to gain from that. At worst, it will just make these test run slower as we have to wait longer on each shutdown.

Also since we are using `asyncio.subprocess.Process` we can't use `join` to wait on the shutdown and I believe waiting in an async fashion would complicate the tear down of the tests.

So, in general, I think this would be **good enough** to merge. Obviously, there's a lot left to do and in fact I've already started on more complex tests that spin up multiple Trinity instances in parallel to test `web3` interaction via `trinity attach`.

trinity/utils/async_iter.py
```diff
@@ -0,0 +1,21 @@
+from typing import (
+    AsyncIterable,
+    Set,
+)
+
+
+async def contains_all(async_gen: AsyncIterable[str], keywords: Set[str]) -> bool:
```
pipermerriam: This appears to be duplicated both here in `trinity.utils.async_iter` and under `trinity.tools.utils.async_iter`

tests/trinity/integration/test_boot.py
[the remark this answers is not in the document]
carver: Yes, but that test doesn't:
- look for exceptions
- run light sync
- run light ropsten sync

carver:
What's the story with merging with a :x: ? Some reason a rush is warranted?

cburgdorf:
@carver it was discussed in chat. The rejection was only due to that extra file.

pipermerriam:
@cburgdorf @carver worth a quick discussion.  I've typically used the following:

- Comment: Not a full review
- Approve: Nothing that **must** be changed (but likely many suggested changes)
- Request Changes: Something that **must** be changed (or discussed and resolved).

If the change is clearly defined in a "Request Changes", then I'm fine with merge prior to re-approval once the change has been executed, but that does leave us with merged pull requests that aren't "approved".  Not sure what to do about this.  It's more of a process question.

Should we use "Approve" to also encompass "some basic changes are required but in general this is good to go"?

cburgdorf:
@pipermerriam I've been wondering the same. I usually try to be extra careful to reassure everyone is okay with me merging something which is also why I pinged you in chat again.

At the same time I don't want to be too spamy with (re)review requests when I feel that the substantial feedback has been addressed.

That said, I think in this particular case I guess a positive review with a clear comment that the file deletion must be addressed would probably have worked better (even though it does violate these rules you pointed out) 

I'm on mobile, not going into much more detail now.

carver:
Yeah, I've (informally) been using this approach:

- Comment: undecided / not a full review
- Approve: Good to go, or I may have pointed out a specific concern.
  - If the concern is unclear to OP, then please clarify before merging
  - If the OP comes up with a different solution to the concern, go ahead and merge
  - If the concern is clearly a misunderstanding, go ahead and merge. OP should be on the lookout for followup clarifications (like the one below the break)
  - If the OP acknowledges the concern, but thinks it's not problematic, please discuss before merging.
- Request Changes: 
  - There's a concern, but I (reviewer) haven't identified a satisfactory solution. OP should post a solution and wait for re-review
  - There's a concern, and a potential solution is suggested. There are subtleties: alternative solutions may be unacceptable. OP should post a solution and wait for re-review.

This approach provides a lot of clarity to OP about next steps: :heavy_check_mark: is completely in the hands of OP to merge at their discretion (although follow-up may be expected). A :x: always implies further discussion or review is expected.

---

BTW, I wanted to make sure this comment didn't get lost in the thread: https://github.com/ethereum/py-evm/pull/1049#discussion_r204514460

pipermerriam:
@carver :+1: Maybe we can update the tactical manual to reflect this approach for better clarity.

tests/trinity/integration/test_boot.py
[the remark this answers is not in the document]
cburgdorf: I'll comment on that later. Mobile s*cks!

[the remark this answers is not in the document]
cburgdorf: >run light mainnet sync
>run light ropsten sync

I wasn't exactly sure what we want for the light mode in regard of the transaction pool. Currently, the transaction pool doesn't work for the light mode anyway (see https://github.com/ethereum/py-evm/issues/1031) so I'm inclined to catch that in the tx pool plugin and then add a test that proves that the tx pool **does not start** in light mode (as in, if `--tx-pool` is specified you'll get a warning that tx pool doesn't work yet in light mode).

>look for exceptions

True that. I'll add the `--txpool` setup to the test that scans for exceptions.

PR #1050, opened 2018-07-20
Docs: Fix wrong hyperlinks

qpakzk:
### What was wrong?

There're lots of wrong links in docs. They're connected to unknown urls.

### How was it fixed?

Fix them to be entered to right location.

#### Cute Animal Picture


carver:
Did you happen to build the docs locally to verify the new links?

qpakzk:
No. On github, hyperlinks in `docs` directory find a wrong path. So I sent Pull Request.

docs/api/index.rst
```diff
@@ -1,7 +1,7 @@
 API
 =====
 
-This section aims to provide a detailed description of all APIs. If you are looking for something more hands-on or higher-level check out the existing `guides </guides>`_.
+This section aims to provide a detailed description of all APIs. If you are looking for something more hands-on or higher-level check out the existing `guides </docs/guides>`_.
```
cburgdorf: This doesn't work either. This works: `:doc:`guides </guides/index>`

cburgdorf:
I can confirm that these links are broken but unfortunately the proposed changes, doesn't work either. I left a comment to demonstrate how these links can be fixed. If you pull down the repository and change into `docs` you can run `sphinx-build -T . _build/html` and then open the docs from the generated `_build/html` directory to check if things work as they should.

Would you like to update the PR with correct links?

qpakzk:
@cburgdorf Thank you for reviewing. Due to your advice, I can test with right way and modify to correct links.

PR #1063, opened 2018-07-20
Implement stronger linkage between requests for headers and their responses

### LINKED ISSUE #1035: Matching between requested headers and received headers.
### What is wrong?

The syncing code in `p2p.chain` under `BaseHeaderChainSyncer._sync` makes an assumption about the received headers which is not necessarily correct.

### What I think is going on (please confirm)

In trying to understand where these are coming from, I'm working to understand the syncing logic and want to be sure I have it correct.

- [`p2p.chain.BaseHeaderChainSyncer._sync#L205`](https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/chain.py#L205) requests headers from the chosen `peer`
- [`p2p.chain.BaseHeaderChainSyncer._fetch_missing_headers#L253`](https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/chain.py#L253) issues a request to the peer for some headers.
- [`p2p.chain.BaseHeaderChainSyncer._fetch_missing_headers#L256`](https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/chain.py#L256) waits for new headers to show up in the `_new_headers` queue.
- When the new headers show up, [`p2p.chain.BaseHeaderChainSyncer._handle_block_headers#L276`](https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/chain.py#L276) puts the new headers in the `_new_headers` queue.

So, IIUC, if **any** peer sends us a `BlockHeaders` message, the new headers will end up in the `_new_headers` queue and the `headers` returned in [`_fetch_missing_headers`](https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/chain.py#L256) are not guaranteed to originate from the peer they were requested from.

Alternate explanation of what I'm trying to understand.

1. We request headers from peer `A`
2. Peer `B` sends us a `BlockHeaders` message before peer `A` responds
3. The headers from peer `B` end up in the queue.
4. We treat the headers received from peer `B` as if they were sent from peer `A`.

If this is the case then I believe we have a systemic problem that we need to solve.

### How we can solve it (maybe)

My understanding is that there is no `id` for these types of requests so we cannot do `id` based linking between requests and responses.

Without an `id` based way to link requests and responses, I think we have to do it in a probabilistic manner.

1. instead of a single queue for headers, lets use a queue per peer.
2. use the [`PeerPoolSubscriber.deregister_peer`](https://github.com/ethereum/py-evm/blob/3dbad6b8ea82b2f325b3fce6d96a7f0dc2b5642f/p2p/peer.py#L706) API to clear out stale queues.
3. maintain list of *header requests* that are awaiting responses.
4. on receipt of `BlockHeaders` message, we only put it in the queue if it has a corresponding *header request* which is satisfied by the headers contained in the message.

Defining *header request*:

- A *header request* would be defined as the requesting parameters in the `GetBlockHeaders` message.  
- A match would be defined as a non-empty set of headers which is a subset of the full set of headers that could be generated by the request.

Any received headers that cannot be correlated to a specific peer and header request should be logged and discarded.

pipermerriam:
fixes #1035 

### What was wrong?

When issuing request for headers to `ETHPeer` peers, there is no `id` as part of the request.  This means we have no direct linkage between the request for headers and the response.

### How was it fixed?

- Implemented a `HeaderRequest` object which wraps the request parameters and exposes two apis.
  - generation of block numbers that satisfy the request
  - validation of a list of block numbers (or headers) is a correct subset of the expected numbers

What I did **not** do in this PR that will happen in a subsequent PR.

- Some better organization.  The `HeaderRequest` living the `p2p.peer` module isn't ideal.  I want to get the `HeaderRequest` object and the `BlockHeaders` command class co-located.
- An API on the `Peer` class so that you can `await` the response to a request.
- A full `await` API that issues the request as well as waiting for the response

The last piece here will simplify our syncing code quite a bit because we can get rid of a lot of code which issues a request and then waits for an asynchronous response.

#### Cute Animal Picture

cuteowl


pipermerriam:
This is me tinkering with fixing #1035

p2p/chain.py
```diff
@@ -203,6 +202,10 @@ def register_peer(self, peer: BasePeer) -> None:
                 self.logger.warn("Timeout waiting for header batch from %s, aborting sync", peer)
                 await peer.disconnect(DisconnectReason.timeout)
                 break
+            except ValidationError:
+                self.logger.warn("Invalid header response sent by peer.")
```
gsalgado: It'd be good to include the original exception in the log, no? I'd also mention we'll disconnect from the peer because of that

p2p/chain.py
```diff
@@ -695,34 +723,46 @@ def __init__(self, db: 'AsyncHeaderDB', logger: TraceLogger, token: CancelToken)
...
+                                             request: HeaderRequest) -> Tuple[BlockNumber, ...]:
+        """
+        Generate the block numbers for a given `HeaderRequest`.
+        """
+        if isinstance(request.block_number_or_hash, bytes):
+            header = await self.wait(
+                self.db.coro_get_block_header_by_hash(cast(Hash32, request.block_number_or_hash)))
+            return request.generate_block_numbers(header.block_number)
+        elif isinstance(request.block_number_or_hash, int):
+            return request.generate_block_numbers()
```
gsalgado: Without having read `HeaderRequest.generate_block_numbers()`, it feels odd to see it called both with/without a block number argument. Maybe it'll make sense later on, but thought I'd mention

p2p/les.py
```diff
@@ -314,8 +315,12 @@ def send_get_block_bodies(self, block_hashes: List[bytes], request_id: int) -> N
         header, body = GetBlockBodies(self.cmd_id_offset).encode(data)
         self.send(header, body)
 
-    def send_get_block_headers(self, block_number_or_hash: Union[int, bytes],
-                               max_headers: int, request_id: int, reverse: bool = True
+    def send_get_block_headers(self,
+                               request_id: int,
```
gsalgado: I don't care too much about having `request_id` as the first or last argument, but all `LESProtocol` methods have it as the last, so if you feel strong about having it as the first, we should change the others as well, to be consistent

p2p/peer.py
```diff
@@ -143,6 +143,106 @@
...
 
+class HeaderRequest(NamedTuple):
+    block_number_or_hash: BlockIdentifier
+    max_headers: int
+    skip: int
+    reverse: bool
+
+    def generate_block_numbers(self,
+                               block_number: BlockNumber=None) -> Tuple[BlockNumber, ...]:
+        if block_number is None and not self.is_numbered:
```
gsalgado: Ok, now it's clear why we sometimes have to pass the block number here. Could we change `HeaderRequest` to have a `block_number: int` attribute and force callsites to to the header lookup before instantiating a `HeaderRequest`, when the peer specifies a block hash?

I'd really like to address this somehow because it's very easy to misuse the current API and get a `TypeError` when calling `generate_block_numbers()`

p2p/peer.py
```diff
@@ -143,6 +143,106 @@
...
+
+    def validate_headers(self,
+                         headers: Tuple[BlockHeader, ...]) -> None:
+        if not headers:
+            # empty response is always valid (TODO: deal with this)
+            return
+        elif not self.is_numbered:
+            first_header = headers[0]
+            if first_header.hash != self.block_number_or_hash:
+                raise ValueError("TODO: mismatched first block hash")
```
gsalgado: TODO?

p2p/peer.py
```diff
@@ -143,6 +143,106 @@
...
+            raise ValueError(
+                'Unexpected numbers: {0}'.format(unexpected_numbers))
+
+        # check that the numbers are correctly ordered.
+        expected_order = tuple(sorted(
+            block_numbers,
+            reverse=self.reverse,
+        ))
+        if block_numbers != expected_order:
+            raise ValueError('Invalid ordering')
```
gsalgado: Maybe include the expected and seen orders in the exception?

p2p/peer.py
```diff
@@ -143,6 +143,106 @@
...
+        )
+
+    @property
+    def is_numbered(self) -> bool:
+        return isinstance(self.block_number_or_hash, int)
+
+    def validate_headers(self,
+                         headers: Tuple[BlockHeader, ...]) -> None:
+        if not headers:
+            # empty response is always valid (TODO: deal with this)
```
gsalgado: How would we deal with it?  Asking because if you already have ideas it'd be nice to describe them here so that others can address the TODO

p2p/peer.py
```diff
@@ -143,6 +143,106 @@
...
+        # sequence.
+        iter_expected = iter(expected_numbers)
+        for number in block_numbers:
+            for value in iter_expected:
+                if value == number:
+                    break
+            else:
+                raise ValueError('Unmatched number')
+
+    def is_valid_headers(self, headers: Tuple[BlockHeader, ...]) -> bool:
```
gsalgado: This doesn't seem to be used anywhere

p2p/peer.py
```diff
@@ -143,6 +143,106 @@
...
+        try:
+            self.validate_headers(headers)
+        except ValueError:
+            return False
+        else:
+            return True
+
+    def is_valid_sequence(self, block_numbers: Tuple[BlockNumber, ...]) -> bool:
```
gsalgado: Nor this

p2p/state.py
```diff
@@ -135,7 +136,14 @@ def msg_queue_maxsize(self) -> int:
                 # A node may be received more than once, so pop() with a default value.
                 self._pending_nodes.pop(node_key, None)
         elif isinstance(cmd, eth.GetBlockHeaders):
-            await self._handle_get_block_headers(peer, cast(Dict[str, Any], msg))
+            query = cast(Dict[Any, Union[bool, int]], msg)
```
gsalgado: Why do we need to use `Any` for the dict's key here?

p2p/les.py
pipermerriam: :+1: I'd like it to be consistently ordered.  Last works for me.

p2p/peer.py
pipermerriam: Placeholder for a more descriptive error message.  Will update.

p2p/peer.py
pipermerriam: Leftover todo that I apparently forgot to cleanup.  With your suggestion in the other PR about only dealing with a single in-flight request at a time, this is less of an issue.

p2p/peer.py
pipermerriam: I think we can drop this API with the planned change in the next iteration to only support a single in-flight request for a given command at any time.

p2p/state.py
pipermerriam: No, will update to be `str`

pipermerriam:
@gsalgado can you give this another pass.

p2p/peer.py
```diff
@@ -955,7 +955,11 @@ def _add_peer(self,
                 break
 
             start_block = vm_class.dao_fork_block_number - 1
-            request = peer.request_block_headers(start_block, max_headers=2)  # type: ignore
+            request = cast(ETHPeer, peer).request_block_headers(
```
gsalgado: This could be a `LESPeer` as well. Not sure if the `# type ignore` comment is any better, but the cast is misleading

pipermerriam: I updated to use an `ignore` comment and added a comment indicating that this *can* be an `LESPeer` and that this issue will be fixed when the full awaitable round-trip API is in place.

p2p/chain.py
```diff
@@ -244,22 +253,39 @@ def register_peer(self, peer: BasePeer) -> None:
...
+            in headers
+            if not (await self.wait(self.db.coro_header_exists(header.hash)))
+        ])
+        if len(missing_headers) != len(headers):
+            self.logger.debug(
+                "Discarding %d / %d headers that we already have",
+                len(headers) - len(missing_headers),
+                len(headers),
+            )
+        return headers
```
carver: Should this return `missing_headers`?

pipermerriam: Yeah, looking at these now the logic isn't nearly equivalent, I'll clean it up in #1089

pipermerriam: Updated here with a comment to make it clearer what the logic of the loop is doing.

https://github.com/ethereum/py-evm/pull/1089/commits/3996fcd695153a533fb43801440556f10adefa1c

PR #1070, opened 2018-07-23
Make it easy to profile the ChainSyncer

gsalgado:
Also, when profiling, do not run anything in an asyncio executor as
those would not be profiled if we did.

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -1,7 +1,6 @@
 import asyncio
 import logging
 import math
-import time
 from abc import abstractmethod
 from typing import (
     Any,
@@ -46,7 +45,8 @@ from p2p.rlp import BlockBody
 from p2p.service import BaseService
 from p2p.utils import (
     get_block_numbers_for_request,
-    get_process_pool_executor,
+    get_asyncio_executor,
+    Timer,
 )
 
 
@@ -88,7 +88,7 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
         self._sync_complete = asyncio.Event()
         self._sync_requests: asyncio.Queue[HeaderRequestingPeer] = asyncio.Queue()
         self._new_headers: asyncio.Queue[Tuple[BlockHeader, ...]] = asyncio.Queue()
-        self._executor = get_process_pool_executor()
+        self._executor = get_asyncio_executor()
 
     @property
     def msg_queue_maxsize(self) -> int:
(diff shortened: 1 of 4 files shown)


p2p/utils.py
```diff
@@ -66,7 +66,7 @@ def get_block_numbers_for_request(
     return sequence_builder(block_number, limit, skip, reverse)
 
 
-def get_process_pool_executor() -> ProcessPoolExecutor:
+def get_asyncio_executor() -> Executor:
```
cburgdorf: @gsalgado what's the reason for this name change? Isn't `get_process_pool_executor` quite accurate?

gsalgado: This function returns the executor we use in conjunction with `loop.run_in_executor()`, and an `Executor` is what we want there -- the fact that we use a `ProcessPoolExecutor` by default is just an implementation detail. We might change this to return a `ThreadPoolExecutor`, for instance, if we're running on a single-core machine with limited resources

cburgdorf: >the fact that we use a `ProcessPoolExecutor` by default is just an implementation detail.

Totally, but imagine we also had a `ThreadPoolExecutor`, then I guess having the functions named `get_process_pool_executor` and `get_thread_pool_executor` would make even more sense to differentiate. Now that we only have one I agree that `get_asyncio_executor` isn't wrong, I just think that the previous name was a better fit **especially** if we consider that there **could** be different functions, returning different kind of implementations.

p2p/chain.py
```diff
@@ -304,10 +304,14 @@ class LightChainSyncer(BaseHeaderChainSyncer):
 
     async def _process_headers(
             self, peer: HeaderRequestingPeer, headers: Tuple[BlockHeader, ...]) -> int:
+        start = time.time()
```
pipermerriam: What do you think about writing a simple utility for this.

```
with timer() as stopwatch:
    ...  # do things
self.logger.info("thing took %d", stopwatch.elapsed)

# Or explicit start/stop
stopwatch = timer()
self.logger.info("thing took %d", stopwatch.elapsed)
```

I think there are multiple places in our code where we coul

p2p/utils.py
gsalgado: We won't have two different executors/functions for the `ChainSyncer` to choose from -- we'd still have a single function that returns the executor, depending on user config, machine resources, whatever. I can't help but wonder if such a trivial change really warrants that much bikeshedding, though...

p2p/chain.py
cburgdorf: We have something like that already.

https://github.com/ethereum/py-evm/blob/462ae35cbfa8bede00baeec1db92bdb720fede2b/scripts/benchmark/utils/meters.py#L11-L20

gsalgado: Added one under p2p/ as at this time we only seem to need that functionality there.

gsalgado: `TimedResult` doesn't seem like a good fit here as in the p2p code we're usually timing just a few statements. And in one instance (`StateSyncer`) we actually want to keep the start time it as an instance attribute so that we can use it on the stats we report periodically

p2p/utils.py
cburgdorf: > can't help but wonder if such a trivial change really warrants that much bikeshedding

Oh, I didn't mean to do that. I was just interested to hear the reason for the rename and I continue to believe the previous name was well suited. However, naming things is highly subjective anyway so no need to continue bikeshedding here :wink:

PR #1071, opened 2018-07-23
Prefix trinity version with eth if trinity not installed

pipermerriam:
### What was wrong?

When `trinity` is not installed we fall back to the `py-evm` version for the version number.  We don't want to have any confusion when viewing this number

### How was it fixed?

When this happens, prefix the version string with `eth-`

#### Cute Animal Picture

smiling-gorilla

### DIFF
--- a/trinity/__init__.py
+++ b/trinity/__init__.py
@@ -5,7 +5,9 @@ __version__: str
 try:
     __version__ = pkg_resources.get_distribution("trinity").version
 except pkg_resources.DistributionNotFound:
-    __version__ = pkg_resources.get_distribution("py-evm").version
+    __version__ = "eth-{0}".format(
+        pkg_resources.get_distribution("py-evm").version,
+    )
 
 from .main import (  # noqa: F401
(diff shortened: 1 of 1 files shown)


trinity/__init__.py
```diff
@@ -5,7 +5,9 @@
 try:
     __version__ = pkg_resources.get_distribution("trinity").version
 except pkg_resources.DistributionNotFound:
-    __version__ = pkg_resources.get_distribution("py-evm").version
+    __version__ = "eth-{0}".format(
```
carver: Since this is inside the trinity module, should we just let it crash?

If the concern is confusion from user-reported versions, I still think we might run into problems. If the display on their screen is `eth-v1.2.3` and we ask them the version, it seems totally plausible that they just say "v1.2.3".

pipermerriam: Merging as-is and punting on a better change which pulls the local git commit and/or reads the version string from `setup_trinity.py`

PR #1078, opened 2018-07-24
Also tests that scan for exceptions when running tx pool

cburgdorf:
### What was wrong?

As discussed [here](https://github.com/ethereum/py-evm/pull/1049#discussion_r204514460) our integration test that scans for exception during the Trinity boot do not cover Trinity runs with `--tx-pool` yet.

### How was it fixed?

This adds tests for `trinity --tx-pool` and `trinity --tx-pool --ropsten` to check for any thrown exceptions during boot. Notice that this does not yet add such tests for `--light` because the `TxPool` currently [doesn't work in light mode](https://github.com/ethereum/py-evm/issues/1031) and I'm tempted to add a test that proves it doesn't even start in light mode instead (with a raised warning).

#### Cute Animal Picture


cburgdorf:
Pinging @carver for a review. Notice that this is mildly related to https://github.com/ethereum/py-evm/pull/1079 which still needs a bit of work.

PR #1079, opened 2018-07-24
Deactivate TxPool in light mode

cburgdorf:
### What was wrong?

Currently the transaction pool starts but doesn't work when in light mode. (#1031)

### How was it fixed?

This ensures the user gets a warning that the transaction pool doesn't work yet in light mode.

#### Cute Animal Picture


trinity/plugins/builtin/tx_pool/plugin.py
```diff
@@ -61,7 +61,10 @@ def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAct
 
     def handle_event(self, activation_event: BaseEvent) -> None:
         if isinstance(activation_event, TrinityStartupEvent):
-            self.is_enabled = activation_event.args.tx_pool
+            light_mode = activation_event.args.sync_mode == 'light'
+            self.is_enabled = activation_event.args.tx_pool and not light_mode
```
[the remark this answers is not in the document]
cburgdorf: Mmh, I feel different about that.

>Did you consider changing trinity's `launch_node()` to modify `args.tx_pool` when in light mode

Since that code is currently in the core Trinity code base it should not know about the transaction pool plugin. Basically, none of the Trinity core code should know about any plugins imho.

On the other hand, I feel less bad about *this particular* transaction pool plugin making assumptions of the `sync_mode`. After all, the whole idea is that someone else could write another transaction pool plugin which works just fine with light mode.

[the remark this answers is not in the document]
gsalgado: Hmm, indeed, I guess it makes more sense to have the plugins themselves decide when they should be enabled. In this case my only suggestion is to use a constant when comparing to `.args.sync_mode`. I believe we might even have that already: `trinity.consntants.SYNC_LIGHT`

tests/trinity/integration/test_boot.py
```diff
@@ -71,6 +71,25 @@ def async_process_runner():
...
+    'command',
+    (
+        ['trinity', '--light', '--tx-pool'],
+        ['trinity', '--light', '--ropsten', '--tx-pool'],
+    )
+)
+@pytest.mark.asyncio
+async def test_txpool_deactivated(async_process_runner, command):
+    # UPNP discovery can delay things, we use a timeout longer than the discovery timeout
+    await async_process_runner.run(command, timeout_sec=40)
```
gsalgado: Can't we disable UPNP altogether for this test? I also wonder if an integration test is the right approach to ensure the tx pool is deactivated...

trinity/plugins/builtin/tx_pool/plugin.py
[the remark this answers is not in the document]
cburgdorf: Ah, good point. Will update to use that constant.

tests/trinity/integration/test_boot.py
cburgdorf: >Can't we disable UPNP altogether for this test

Oh, indeed we don't have to wait for UPNP discovery here. Will update.

>I also wonder if an integration test is the right approach to ensure the tx pool is deactivated

I have mixed feelings about this test myself. After all, I guess the main thing that we want to achieve is to ensure someone running `trinity --light --tx-pool` won't be left with something totally broken which is what the *exception scanning* test tries to achieve.

But then I think it doesn't hurt to go one step beyond that and test that the tx pool doesn't start at all and yields a warning. I think it's hard to test this somewhere else without making the test somewhat artificial and redundant. Testing this on the integration level gives us a 100 % guarantee and only adds up a couple of seconds.

Not sure. How does the rest feel about it? @pipermerriam @carver

trinity/plugins/builtin/tx_pool/plugin.py
```diff
@@ -61,7 +64,10 @@ def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAct
 
     def handle_event(self, activation_event: BaseEvent) -> None:
         if isinstance(activation_event, TrinityStartupEvent):
-            self.is_enabled = activation_event.args.tx_pool
+            light_mode = activation_event.args.sync_mode == SYNC_LIGHT
+            self.is_enabled = activation_event.args.tx_pool and not light_mode
+            if light_mode:
+                self.logger.warning('The transaction pool is not yet available in light mode')
```
pipermerriam: I think this should be an error and cause the process to terminate.  I'd rather the user re-run the command without the flag so there is no ambiguity.

cburgdorf:
Maybe give plug-ins a generic way to exit trinity and hence giving other
processes the chance to gracefully shut down?

pipermerriam:
> Maybe give plug-ins a generic way to exit trinity

Maybe using a built-in event that any plugin could emit?

cburgdorf:
>Maybe using a built-in event that any plugin could emit

I've thought about that but there's something about it that doesn't sit well with me. I think I'm not a fan of *raising an event* being used as the cause of something. Well, obviously events will always *cause* other things but if my **intend** is to do something - shutting down trinity - in this case, I feel like a method call would be more suitable.

In the initial plugin API, there's a stub for a `PluginContext` which is meant as an API that every plugin has access to. With that, one can think of a plugin calling something like `self.context.shutdown_trinity()`. That in turn could cause every plugin to receive an event (e.g. `TrinityShutdown`) that they may react upon. Alternatively, we could also imagine there wouldn't be an event because we can also just call  `cancel()` on every plugin to give them a formal way to cleanup.

With the new plugin API as described in https://github.com/ethereum/py-evm/issues/1090 we could still think about injecting such a `PluginContext` or `PluginExternalAPI` or whatever one wants to call it, as part of `Plugin.__init__(args, api, token)`.

**Alternatively**, another more esoteric way could be to not *inject* a formal API into plugins but rather write another plugin to expose a common API. E.g. think of a `CommonAPIPlugin` that provides methods such as `shutdown_trinity()` (and what not) and that other plugins may use. The upside of this approach would be that a new API could be introduced without breaking the old one. As in: a newer API would just be yet another plugin that may even run side by side with the old one.

The downside of this is that it feels kinda weird and I think injecting *some sort* of API into plugins is rather fairly standard. ¯\_(ツ)_/¯

pipermerriam:
I'm :+1: for the `PluginContext` API.

cburgdorf:
I tried to move this forward and started refactoring the plugin API so that:

1. Plugins are initialized by the plugin manager
2. Plugins get a `PluginContext` injected which serves as an API for plugins to perform certain things including shutting down trinity

However, I hit a road block again. Naively, I first tried to just do `sys.exit(exit_code)` and thought this would propagate and eventually cause a clean shutdown similar to hitting `CTRL + C` on a running Trinity node.

However, that doesn't work and I *think* what we really need is to built the plugin architecture to properly spread across both processes. We currently already have two instances of the `PluginManager`, one in the main process and one in the networking process but that's more a hack than a solution. I guess we really need to distinct parts that are connected so they can talk because then, the `PluginManager` from the networking process (or any non-main process really) could just talk to the other part that is sitting in the main process and tell it to shut the process down. And then, the part sitting in main would be responsible to shut everything down as it's the highest process in the hierarchy and knows about the subprocesses.

pipermerriam:
Maybe we can do something with a *root* cancel token which gets created in the primary trinity process.  Then we can pass a handle of that into the `PluginManager` and then it can call `token.trigger()` which if wired up to the `Node` class cancel tokens will cause a cascading cancellation and exit the process.

cburgdorf:
>Maybe we can do something with a root cancel token which gets created in the primary trinity process. Then we can pass a handle of that into the `PluginManager` and then it can call `token.trigger()` which if wired up to the Node class cancel tokens will cause a cascading cancellation and exit the process.

I'll probably have to tinker with a reduced multi process example but I'd assume that you can not simply call `token.trigger()` in process B and expect it to get triggered in process A. Probably needs some sort of proxy to work.

cburgdorf:
I updated this to be partly based on #1212  (cherry-picked one commit). There are still issues (that may turn out to be upstream lahja issues)

cburgdorf:
@pipermerriam this was already approved a while back but it didn't fully work back then and I made substantial changes now to get it working.

This has one commit from #1212 cherry-picked (so that will fall out of that other PR if we merge this first), which is the one that brings in isolated plugins and sets up the eventbus architecture. While isolated plugins aren't used in this PR, the event bus structure is used to propagate a `ShutdownRequest` back to the main process where the shutdown is then handled.

This works reasonable well. There *is* room for improvement because it can happen that the db process shuts down while the networking process is trying to connect to it and then we are left with an ugly unhandled exception but other than that, it shuts down smoothly. Getting rid of that error would take a bit of rethinking the node boot process I guess. However, imho, we don't need to address this now. This works reasonable well from my point of view.

trinity/main.py
```diff
@@ -206,27 +242,62 @@ def trinity_boot(args: Namespace,
...
+        kill_trinity_gracefully(
+            logger,
+            database_server_process,
+            networking_process,
+            plugin_manager,
+            event_bus
+        )
+
+
+def kill_trinity_gracefully(logger: logging.Logger,
```
cburgdorf: I wonder if it's worth making a named tuple (e.g `BootArtifacts`) to tighten the function signature

trinity/main.py
```diff
@@ -206,27 +242,62 @@ def trinity_boot(args: Namespace,
     networking_process.start()
     logger.info("Started networking process (pid=%d)", networking_process.pid)
 
+    main_endpoint.subscribe(
```
cburgdorf: Here is where we listen for the `ShutdownRequest` and then call `kill_trinity_gracefully`. Sidenote about lahja: You could also say `main_endpoint.stream(ShutdownRequest)` to receive the event but this is an async API that needs to be awaited and hence will introduce async/await creep. The `subscribe` API is perfect for these cases as it gives you a callback based way to still receive events asynchronously without introducing async/await creep. I'm not saying I encourage that usage over the other (as I quite like `async` / `await`) but it is handy to use in situations such as here.

cburgdorf:
Right now, it is only the `main` process that listens for this event and then calls `kill_trinity_gracefully` which is the same takedown routine we already have for when a user presses `CTRL+C`.

So, I believe we do actually have the same behavior already if you press `CTRL+C` right in that moment similar to when the `TxPool` calls `shutdown_host()` on the plugin context.

What happens right now (I believe) is that `kill_trinity_gracefully` is called *before* the networking process is setup but after the db process already started. Then, inside `kill_trinity_gracefully` we start of taking down the db process and shortly after the networking process is starting trying to connect to the db. So that should theoretically be exploitable right now if you `CTRL+C` at the right moment in time.

>this gets a little more complicated with multiple processes, but doable: the db process could track which processes are listening to it, and only shut down after it hears a termination event from every listening process

Yep, totally doable and probably worth following up on especially if we want isolated plugins to become popular and hence have a shitload of processes to manage (which btw, a modern browser also does, so I'm not worried about that).

However, I'd appreciate if we could push that into another issue. I can write down a quick issue and link to this thread.

trinity/extensibility/plugin.py
```diff
@@ -70,6 +156,35 @@ def stop(self) -> None:
         pass
 
 
+class BaseIsolatedPlugin(BasePlugin):
+
+    def __init__(self) -> None:
+        super().__init__()
+        self._process: Process = None
```
pipermerriam: This seems like it could be a class level attribute allowing you to remove this `__init__` method entirely.

trinity/extensibility/plugin_manager.py
```diff
@@ -74,7 +113,55 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+                boot_kwargs: Dict[str, Any] = None) -> None:
+        for plugin in self._plugin_store:
+
+            if not self._is_responsible_for_plugin(plugin):
+                continue
+
+            context = self._create_context_for_plugin(plugin)
+            context.args = args
+            context.boot_kwargs = boot_kwargs
+            context.chain_config = chain_config
```
pipermerriam: Maybe roll this into `_create_context_for_plugin` with `**kwargs`?

trinity/extensibility/plugin_manager.py
```diff
@@ -74,7 +113,55 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+            plugin.set_context(context)
+
+    def shutdown(self) -> None:
+        for plugin in self._started_plugins:
+            plugin.stop()
```
pipermerriam: Maybe wrap this in a `try/except` so that if a single plugin throws it doesn't prevent a clean shutdown on the remaining plugins?

cburgdorf: Seems right. And then just issue a warning, right?

trinity/extensibility/plugin_manager.py
cburgdorf: Not a big **kwargs fan because of the lower type safety. Would explicit parameter passing be ok with you?

trinity/extensibility/plugin.py
cburgdorf: wait, but then it becomes bound to that class and not the instance, no? We will have many instances of such plugins and each needs to address its very own process instance. Am I understanding you wrong?

cburgdorf: Oh wow, I just realized that my mental model for how class attributes work was still wrong. I just read through this [super detailed guide](https://www.toptal.com/python/python-class-attributes-an-overly-thorough-guide) and now I understand.

So, while it is fine to have `_process: Process = None` as a class attribute **and still maintain different `_process` instances on the instance level** it would fall apart to have something like `_process_list: List[Process] = []` on the class level if the reference isn't *overwritten* but *mutated* in instances e.g. by  `self._process_list.append(some_process)`. In that case, that process would appear in the `_process_list` of all other plugin instances as well.

With that in mind, class level attributes have more legitimate usage that I thought they have.

cburgdorf: I'm just posting this here for anyone who (as me) did not know this. I'm aware this won't be news for most of you reading this but maybe it will help someone else.

```python
from typing import (
    List
)
from multiprocessing import (
    Process
)

class Foo:

    # FINE
    _proc: Process = None

    # BAD! DON'T DO THIS
    _proc_list: List[Process] = []
    
    def do_something(self):
        self._proc = Process()
        self._proc_list.append(self._proc)


foo1 = Foo()
foo1.do_something()

foo2 = Foo()
foo2.do_something()


print("shared _proc instance? {}".format(foo1._proc is foo2._proc))
print("shared _proc_list instance? {}".format(foo1._proc_list is foo2._proc_list))
```

prints:

```
$ python class_attributes.py 
shared _proc instance? False
shared _proc_list instance? True
```

pipermerriam:
@cburgdorf I never got this across the line but here's my initial attempt at cross process cancellation.  However, I am now skeptical that this is the right approach because it's taking an asyncio primitive and trying to turn it into a multiprocessing primitive and I suspect there's a more optimal/simple approach.

PR #1089, opened 2018-07-25
Further flesh out round trip request/response API

pipermerriam:
### What was wrong?

Still too much manual work to be done for round trip requests for block headers.

### How was it fixed?

- established base class for p2p requests for data in `p2p.peer.BaseRequest`
- setup data structure on `ETHPeer` which houses a mapping between `Command -> Future`.
- In the `handle_sub_process_msg`, if there is a waiting future for a given command, it will set the future result.
- New API `ETHPeer.get_block_headers` which does the full request/response round trip.

This has a really nice side effect of not needing the chain syncers to handle the `BlockHeaders` response, and instead can just `await peer.get_block_headers(...)`, removing a decent amount of complexity from the class.  If we continue this pattern for the rest of the block body parts I think the chain syncer is going to look a lot cleaner and easier to grok.

#### Cute Animal Picture

ab6jyf-1

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -43,9 +43,9 @@ from p2p import eth
 from p2p import les
 from p2p.cancellable import CancellableMixin
 from p2p.constants import MAX_REORG_DEPTH, SEAL_CHECK_RANDOM_SAMPLE_RATE
-from p2p.exceptions import NoEligiblePeers
+from p2p.exceptions import NoEligiblePeers, ValidationError
 from p2p.p2p_proto import DisconnectReason
-from p2p.peer import BasePeer, ETHPeer, LESPeer, HeaderRequest, PeerPool, PeerSubscriber
+from p2p.peer import BasePeer, ETHPeer, LESPeer, PeerPool, PeerSubscriber
 from p2p.rlp import BlockBody
 from p2p.service import BaseService
 from p2p.utils import (
@@ -91,7 +91,6 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
         self._syncing = False
         self._sync_complete = asyncio.Event()
         self._sync_requests: asyncio.Queue[HeaderRequestingPeer] = asyncio.Queue()
-        self._new_headers: asyncio.Queue[Tuple[BlockHeader, ...]] = asyncio.Queue()
         self._executor = get_asyncio_executor()
 
     @property
@@ -207,7 +206,7 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
                 self.logger.warn("Timeout waiting for header batch from %s, aborting sync", peer)
                 await peer.disconnect(DisconnectReason.timeout)
                 break
-            except ValueError as err:
+            except ValidationError as err:
                 self.logger.warn(
                     "Invalid header response sent by peer %s disconnecting: %s",
                     peer, err,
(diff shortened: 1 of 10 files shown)


p2p/peer.py
```diff
@@ -694,6 +698,13 @@ def handle_sub_proto_msg(self, cmd: protocol.Command, msg: protocol._DecodedMsgT
             if actual_td > self.head_td:
                 self.head_hash = actual_head
                 self.head_td = actual_td
+        elif isinstance(cmd, eth.BlockHeaders):
+            # try to match with a request
+            for request, waiter in self.pending_requests.items():
```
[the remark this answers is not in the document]
pipermerriam: I like that.

pipermerriam:
@gsalgado this is ready for a second pass.

p2p/chain.py
```diff
@@ -313,9 +302,6 @@ class LightChainSyncer(BaseHeaderChainSyncer):
                           msg: protocol._DecodedMsgType) -> None:
         if isinstance(cmd, les.Announce):
             self._sync_requests.put_nowait(peer)
-        elif isinstance(cmd, les.BlockHeaders):
-            msg = cast(Dict[str, Any], msg)
-            self._handle_block_headers(tuple(cast(Tuple[BlockHeader, ...], msg['headers'])))
```
gsalgado: Maybe it's worth leaving a comment here that `BlockHeaders` messages are not handled here because they're handled in a Peer method? Otherwise someone reading this code might have a hard time figuring out why those messages are not handled like the others

p2p/chain.py
```diff
@@ -538,9 +525,7 @@ def request_receipts(self, target_td: int, headers: List[BlockHeader]) -> int:
     async def _handle_msg(self, peer: HeaderRequestingPeer, cmd: protocol.Command,
                           msg: protocol._DecodedMsgType) -> None:
         peer = cast(ETHPeer, peer)
-        if isinstance(cmd, eth.BlockHeaders):
-            self._handle_block_headers(tuple(cast(Tuple[BlockHeader, ...], msg)))
-        elif isinstance(cmd, eth.BlockBodies):
```
gsalgado: Same here

p2p/eth.py
```diff
@@ -70,6 +73,31 @@ class GetBlockHeaders(Command):
...
+                 block_number_or_hash: BlockIdentifier,
+                 max_headers: int,
+                 skip: int,
+                 reverse: bool) -> None:
+        self.block_number_or_hash = block_number_or_hash
+        self.max_headers = max_headers
+        self.skip = skip
+        self.reverse = reverse
+
+    def validate_response(self, response: Any) -> None:
```
gsalgado: Do we need to use `Any` here, even though you raise a ValidationError if it's not a tuple?

p2p/exceptions.py
```diff
@@ -159,3 +159,10 @@ class NoInternalAddressMatchesDevice(BaseP2PError):
     def __init__(self, *args: Any, device_hostname: str=None, **kwargs: Any) -> None:
         super().__init__(*args, **kwargs)
         self.device_hostname = device_hostname
+
+
+class ValidationError(BaseP2PError):
```
gsalgado: I'm a bit worried about this as it's the same name of the exception from the `eth` package, and I believe we'll often end up writing code that catches one when in fact is the other that will be raised. Like we just saw with the `OperationCancelled` exception after the move to the cancel_token lib

p2p/peer.py
```diff
@@ -693,6 +641,17 @@ def handle_sub_proto_msg(self, cmd: protocol.Command, msg: protocol._DecodedMsgT
...
+        if cmd_type in self.pending_requests:
+            request, future = self.pending_requests[cmd_type]
+            try:
+                request.validate_response(msg)
+            except ValidationError:
+                pass
+            else:
+                future.set_result(msg)
+                self.pending_requests.pop(cmd_type)
+
```
gsalgado: We could move the whole block above into `BasePeer.handle_sub_proto_msg()` and then avoid having to duplicate it in both `ETHPeer` and `LESPeer`, no?

p2p/protocol.py
```diff
@@ -116,6 +122,120 @@ def encode(self, data: _DecodedMsgType) -> Tuple[bytes, bytes]:
...
+
+class BaseHeaderRequest(BaseRequest):
+    block_number_or_hash: BlockIdentifier
+    max_headers: int
+    skip: int
+    reverse: bool
+
+    @property
+    @abstractmethod
+    def MAX_HEADERS_FETCH(self) -> int:
```
gsalgado: I find it weird to have an all-uppercase property. Any reason for that?

p2p/protocol.py
```diff
@@ -116,6 +122,120 @@ def encode(self, data: _DecodedMsgType) -> Tuple[bytes, bytes]:
...
+        pass
```
gsalgado: shouldn't this raise NotImplementedError()? We definitely don't want it to return None, right?

tests/p2p/test_header_request_object.py
```diff
@@ -13,6 +14,23 @@
 BLOCK_HASH = b'\x01' * 32
 
 
+class HeaderRequest(BaseHeaderRequest):
+    MAX_HEADERS_FETCH = 192
```
gsalgado: I thought you couldn't overwrite an `@abstractmethod` by defining a class attribute in a subclass. Am I missing something?

p2p/peer.py
```diff
@@ -620,10 +531,21 @@ def max_headers_fetch(self) -> int:
...
         if isinstance(cmd, les.Announce):
             self.head_info = cmd.as_head_info(msg)
             self.head_td = self.head_info.total_difficulty
             self.head_hash = self.head_info.block_hash
+        elif cmd_type in self.pending_requests:
+            request, future = self.pending_requests[cmd_type]
+            try:
+                request.validate_response(msg)
+            except ValidationError:
+                pass
```
gsalgado: We should log this, probably as a warning even, no?

p2p/chain.py
gsalgado: Actually, we should probably leave the `elif` block here otherwise it will be handled in the `else` block, which logs a debug msg and may be confusing to someone reading the logs

p2p/eth.py
pipermerriam: I believe `Any` is what we want.  It is allowed to pass anything into this function, but it will raise a validation error if it's not the correct type or it isn't well formed.

That said, I can see an argument for dropping both `isinstance` checks here if we think that `mypy` is sufficient to catch those cases (Which is the way I'm leaning)

p2p/exceptions.py
pipermerriam: cc @carver 

how would the two of you feel about a common/shared validation library that exposed a common `ValidationError` exception that we could use across all of our libraries.  It is one of those things that we have implemented almost everywhere.

p2p/protocol.py
pipermerriam: it's a constant so I mirrored the casing.  I'm fine lower casing it.

p2p/protocol.py
pipermerriam: The `abstractmethod` decorator should take care of this for us, preventing us from instantiating the class if this property has not been overridden.

tests/p2p/test_header_request_object.py
pipermerriam: Brief testing shows this works as expected.  The parent class with the abstract property disallows instantiation.  The child class with the property set, allows instantiation and proper attribute access.

p2p/peer.py
pipermerriam: Yes, though I've been questioning the only one request of a single type in-flight per message type restriction.  I'm going to keep it in place, but unless the spec says that it's disallowed, I'd like to support it eventually.  Sending concurrent requests to high performing peers seems like a way to speed up sync.

p2p/exceptions.py
pipermerriam: https://github.com/ethereum/py-evm/issues/1125

pipermerriam:
@gsalgado PR review pushed.  It includes some minor logic changes to ensure we don't end up with dangling requests in `Peer.pending_requests`.  Mind giving another :eyes:

p2p/peer.py
gsalgado: Fair enough, but IIRC geth doesn't do that, so it might be worth finding out why before we consider doing it ourselves.

Issue #1074, opened 2018-07-24
Trinity times out peer connections during state sync


pipermerriam:
### What is wrong?

While resuming state sync my node stalled while processing *something*.

Here are the logs:  https://gist.githubusercontent.com/pipermerriam/f33802d8d8890af3c6c7a128e2745e75/raw/d83f3abefabeb3dcef5e6d12518db1755f66b666/trinity-state-sync-stalled.log

Event occured between timestamps `07-23 21:20:34` and `07-23 21:21:28` (approximately a full minute of *dead* time.

### How can it be fixed

See what blocked the process for so long and figure out how to move it into a process pool executor or something.

pipermerriam:
It looks like the code location of this may be here:

https://github.com/ethereum/py-evm/blob/master/p2p/state.py#L130

The call out to `StateSyncer.process`.  I believe that we need to do something like push this out to a process pool of workers.

gsalgado:
Good catch! I'd seen this kind of behaviour in the past and wrongly thought it was because some of our requests were timing out (like, in your logs, peer `0x0282` never seem to reply to us[1]), but as you point out the main thread is blocked in the calls to `StateSyncer.process()` so even when the peers actually reply the event loop will not be able to consume them in time since it is blocked. That can even lead us to drop peers as the connection will be idle for too long.

Notice that `StateSyncer` cannot be pickled as it holds a reference to a `LevelDB` instance so we cannot run `.process()` in our `ProcessPoolExecutor`. Also, batching won't help because `StateSyncer.process` processes each item individually.

Given that, I think our best bet for now is to change the loop calling `.process()` to yield control after every N items

[1]

```
   DEBUG  07-23 21:20:42       state  Requesting 384 trie nodes to ETHPeer <Node(0x0282@219.150.218.187)>
   DEBUG  07-23 21:20:42       state  Requesting 384 trie nodes to ETHPeer <Node(0x9a1c@40.118.145.245)>
   DEBUG  07-23 21:20:42       state  Requesting 384 trie nodes to ETHPeer <Node(0x5669@127.0.0.1)>
   DEBUG  07-23 21:20:42       state  Got 0 NodeData entries from ETHPeer <Node(0x9a1c@40.118.145.245)>
   DEBUG  07-23 21:20:42       state  Requesting 384 trie nodes to ETHPeer <Node(0x9a1c@40.118.145.245)>
[...]

PR #1094, opened 2018-07-26
Add an async sleep to the trie node processing loop

pipermerriam:
### What was wrong?

The loop which processes trie nodes can take a long time and none of the loop internals yield control to the event loop.

### How was it fixed?

Added an `asyncio.sleep(0)` to the loop.  Local testing of this did not appear to have any significant effect on sync performance.

#### Cute Animal Picture

4 3 15-dogs-and-baby-animals0

### DIFF
--- a/p2p/state.py
+++ b/p2p/state.py
@@ -132,6 +132,8 @@ class StateDownloader(BaseService, PeerSubscriber):
                     # This means we received a node more than once, which can happen when we
                     # retry after a timeout.
                     pass
+                else:
+                    await asyncio.sleep(0)
                 # A node may be received more than once, so pop() with a default value.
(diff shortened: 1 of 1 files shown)


pipermerriam:
Fixes https://github.com/ethereum/py-evm/issues/1074

gsalgado:
I did test a similar change yesterday as well, but I was seeing a lot more nodes timing out and being retried. I think I've found the cause but I want to test it a bit more. I also have a WIP for a proper fix to #1074 (i.e. making `StateSync`'s APIs asynchronous), but that's going to be more involved

gsalgado:
A similar change has been merged in another PR, closing

Issue #1077, opened 2018-07-24
StateDownloader should retry immediately after receiving empty reply from peer


gsalgado:
### What is wrong?

During a state sync, sometimes a peer won't have the trie nodes we requested, and so will send us an empty `NodeData` msg.  When that happens we just move on and those nodes are re-requested only after the timeout (20s) has passed.

### How can it be fixed

Whenever we get an empty `NodeData` from a peer, we should look up the nodes that were originally requested to that peer and re-request them to a different peer.

pipermerriam:
Seems like we need to keep some sort of tracking around which peers do not have certain trie nodes so that we can get to a point where we know "no peers have this trie node".  I believe this is needed for us to get out of a stalled state sync as noted in #1073

Issue #1073, opened 2018-07-24
Trinity very slow to resume state sync


pipermerriam:
### What is wrong?

My understanding is that the state sync works as follows.

A breadth first traversal of the state trie is performed, requesting missing nodes as they are encountered.  Each time we receive new nodes, we dig deeper into the state trie, eventually walking the whole thing.

When I stopped my state sync after about 2-3 million trie nodes had been synced and then resumed it took ____still_waiting____ for it to walk down through the tree far enough to start finding new nodes to request.  This delay seems like it could be reduced.

### How can it be fixed

Figure out a way to *resume* with potentially a secondary process that walks from the root of the tree to find any intermediate nodes that we are still missing.

pipermerriam:
1. need to detect this state.  Probably some additional tracking on the missing trie nodes to ensure we don't just re-request the same node to the same peers endlessly.
2. when we detect this. Probably need to sync more of the header chain and then try incrementally newer state roots to do the state sync from.

Assuming this is the desired strategy, this is another case for concurrent state and chain data syncing.

pipermerriam:
@carver I removed this from the Margaret Hamilton milestone as I think it's a more significant refactor to the `FullSyncer` service than we can reasonably do for this release.  Can you add it to the next milestone when you create it?

pipermerriam:
Latest thoughts on how we can deal with this.

1. Change `FastChainSyncer` to first download the header chain (no block bodies or receipts)
2. Run both the `FastChainSyncer` and the `StateDownloader` concurrently once the header chain has been synced.
3. Anytime the state root that the `StateDownloader` is using is more than X blocks behind the current head of the chain, update to a newer state root.

We may need to build a mechanism for *smooth* transition from one state root to another in the state downloader.  Simply re-starting the state download with the new state root will potentially incur significant overhead.  Ideally we can preserve whatever data from the existing state downloader and populate the new one, filtering out any `SyncRequest` that is for a node-key which is not present under the new state root.

PR #1097, opened 2018-07-26
state-sync: don't request trie nodes to peers known not to have them

gsalgado:
Keep track of which nodes we requested to every peer so that we know which
ones they're missing when they reply with just a subset of what was
requested. That way we can avoid re-requesting those missing nodes to peers
that don't have them. #1073

Also, when a reply from a peer contains only a subset of what was
requested, we immediately re-request the missing ones. Closes: #1077

And a quick workaround to avoid losing peers during state sync (#1074)

### DIFF
--- a/p2p/state.py
+++ b/p2p/state.py
@@ -73,7 +73,7 @@ class StateDownloader(BaseService, PeerSubscriber):
         self.root_hash = root_hash
         self.scheduler = StateSync(root_hash, account_db)
         self._handler = PeerRequestHandler(self.chaindb, self.logger, self.cancel_token)
-        self._requested_nodes: Dict[ETHPeer, Tuple[float, List[Hash32]]] = {}
+        self._active_requests: Dict[ETHPeer, Tuple[float, List[Hash32]]] = {}
         self._peer_missing_nodes: Dict[ETHPeer, List[Hash32]] = collections.defaultdict(list)
         self._executor = get_asyncio_executor()
 
@@ -94,7 +94,7 @@ class StateDownloader(BaseService, PeerSubscriber):
         """Return an idle peer that may have any of the trie nodes in node_keys."""
         async for peer in self.peer_pool:
             peer = cast(ETHPeer, peer)
-            if peer in self._requested_nodes:
+            if peer in self._active_requests:
                 self.logger.trace("%s is not idle, skipping it", peer)
                 continue
             if node_keys.difference(self._peer_missing_nodes[peer]):
@@ -141,7 +141,7 @@ class StateDownloader(BaseService, PeerSubscriber):
             pass
         elif isinstance(cmd, eth.NodeData):
             msg = cast(List[bytes], msg)
-            if peer not in self._requested_nodes:
+            if peer not in self._active_requests:
                 # This is probably a batch that we retried after a timeout and ended up receiving
                 # more than once, so ignore but log as an INFO just in case.
                 self.logger.info(
@@ -150,7 +150,7 @@ class StateDownloader(BaseService, PeerSubscriber):
                 return
 
             self.logger.debug("Got %d NodeData entries from %s", len(msg), peer)
-            _, requested_node_keys = self._requested_nodes.pop(peer)
+            _, requested_node_keys = self._active_requests.pop(peer)
 
             loop = asyncio.get_event_loop()
             node_keys = await loop.run_in_executor(self._executor, list, map(keccak, msg))
@@ -208,7 +208,7 @@ class StateDownloader(BaseService, PeerSubscriber):
             candidates = list(not_yet_requested.difference(self._peer_missing_nodes[peer]))
             batch = candidates[:eth.MAX_STATE_FETCH]
             not_yet_requested = not_yet_requested.difference(batch)
-            self._requested_nodes[peer] = (time.time(), batch)
+            self._active_requests[peer] = (time.time(), batch)
             self.logger.debug("Requesting %d trie nodes to %s", len(batch), peer)
             peer.sub_proto.send_get_node_data(batch)
 
@@ -218,12 +218,12 @@ class StateDownloader(BaseService, PeerSubscriber):
             oldest_request_time = now
             timed_out = []
             # Iterate over a copy of our dict's items as we're going to mutate it.
-            for peer, (req_time, node_keys) in list(self._requested_nodes.items()):
+            for peer, (req_time, node_keys) in list(self._active_requests.items()):
                 if now - req_time > self._reply_timeout:
                     self.logger.debug(
                         "Timed out waiting for %d nodes from %s", len(node_keys), peer)
                     timed_out.extend(node_keys)
-                    self._requested_nodes.pop(peer)
+                    self._active_requests.pop(peer)
                 elif req_time < oldest_request_time:
                     oldest_request_time = req_time
             if timed_out:
@@ -276,7 +276,7 @@ class StateDownloader(BaseService, PeerSubscriber):
     async def _periodically_report_progress(self) -> None:
         while self.is_running:
             requested_nodes = sum(
-                len(node_keys) for _, node_keys in self._requested_nodes.values())
+                len(node_keys) for _, node_keys in self._active_requests.values())
             self.logger.info("====== State sync progress ========")
             self.logger.info("Nodes processed: %d", self._total_processed_nodes)
             self.logger.info("Nodes processed per second (average): %d",
--- a/p2p/state.py
+++ b/p2p/state.py
@@ -1,17 +1,18 @@
 import asyncio
+import collections
 import logging
-import secrets
 import time
 from typing import (
     Any,
     cast,
     Dict,
+    Iterable,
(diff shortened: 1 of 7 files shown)


p2p/state.py
```diff
@@ -161,32 +186,36 @@ def msg_queue_maxsize(self) -> int:
...
+    async def request_nodes(self, node_keys: Iterable[Hash32]) -> None:
+        not_yet_requested = set(node_keys)
+        while not_yet_requested:
+            try:
+                peer = await self.get_peer_for_request(not_yet_requested)
+            except NoEligiblePeers:
+                self.logger.debug(
+                    "No idle peers have any of the trie nodes we want, sleeping a bit")
+                await self.wait(asyncio.sleep(0.2))
+                continue
```
pipermerriam: If none of the peers we are connected to have a node we're looking for this loop will never terminate.

p2p/state.py
```diff
@@ -161,32 +186,36 @@ def msg_queue_maxsize(self) -> int:
...
             timed_out = []
-            for node_key, req_time in self._pending_nodes.items():
+            # Iterate over a copy of our dict's items as we're going to mutate it.
+            for peer, (req_time, node_keys) in list(self._requested_nodes.items()):
                 if now - req_time > self._reply_timeout:
-                    timed_out.append(node_key)
+                    self.logger.debug(
+                        "Timed out waiting for %d nodes from %s", len(node_keys), peer)
+                    timed_out.extend(node_keys)
+                    self._requested_nodes.pop(peer)
```
pipermerriam: Can we re-structure this in a non-mutative manner.

```python
timed_out = cytoolz.valfilter(lambda v: time.time() - v[0] > self._reply_timeout, self.requested_nodes)
self._requested_nodes = cytoolz.dissoc(self._request_nodes, *timed_out.keys())
timed_out_nodes = cytoolz.concat(node_keys for _, node_keys in timed_out.values())
```

p2p/state.py
```diff
@@ -201,7 +230,7 @@ def msg_queue_maxsize(self) -> int:
             now = time.time()
             sleep_duration = (oldest_request_time + self._reply_timeout) - now
             try:
-                await self.wait_first(asyncio.sleep(sleep_duration))
+                await self.wait(asyncio.sleep(sleep_duration))
```
pipermerriam: More places where `self.sleep` would be more concise

p2p/state.py
```diff
@@ -238,18 +267,19 @@ def msg_queue_maxsize(self) -> int:
 
     async def _periodically_report_progress(self) -> None:
         while self.is_running:
+            requested_nodes = sum(
+                len(node_keys) for _, node_keys in self._requested_nodes.values())
             self.logger.info("====== State sync progress ========")
```
pipermerriam: Whatcha think about going ahead and re-structuring this output to be a single line similar to geth's?

```
processed: %11d  tnps: %5s committed: %11d scheduled: %7d timeout: %10d duplicates: %11d
```

p2p/state.py
```diff
@@ -71,7 +71,8 @@ def __init__(self,
         self.root_hash = root_hash
         self.scheduler = StateSync(root_hash, account_db)
         self._handler = PeerRequestHandler(self.chaindb, self.logger, self.cancel_token)
-        self._peers_with_pending_requests: Dict[ETHPeer, float] = {}
+        self._requested_nodes: Dict[ETHPeer, Tuple[float, List[Hash32]]] = {}
```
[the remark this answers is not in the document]
gsalgado: Sounds good to me

p2p/state.py
```diff
@@ -81,23 +82,29 @@ def msg_queue_maxsize(self) -> int:
...
-    async def get_idle_peer(self) -> ETHPeer:
-        idle_peers = await self._get_idle_peers()
-        while not idle_peers:
-            await self.wait(asyncio.sleep(0.02))
-            idle_peers = await self._get_idle_peers()
-        return secrets.choice(idle_peers)
+    async def get_peer_for_request(self, node_keys: Set[Hash32]) -> ETHPeer:
+        """Return an idle peer that may have any of the trie nodes in node_keys."""
+        async for peer in self.peer_pool:
+            peer = cast(ETHPeer, peer)
```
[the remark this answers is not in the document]
gsalgado: Do we actually need that or can we instead have a separate `PeerPool` instance for every peer type, with separate services that subscribe only to the pool they need?

p2p/state.py
gsalgado: Right... Maybe for now we could have another exception to signal that none of our peers have any of the nodes we're about to request. Then we could do one/more of the following:
1. If PeerPool is full, disconnect from one or more peers to make room for others that might have the nodes
2. Increase sleep delay
3. Request another batch of nodes

What do you think? Any other ideas?

p2p/state.py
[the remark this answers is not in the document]
pipermerriam: :+1: 

This seems like a good idea, since we're likely to also have different policies on how we treat different peers and we won't want to conflate those things.

p2p/state.py
pipermerriam: I'm fine punting on the long term solution for this (but lets document it with another issue).  I think option 3 is our best and simplest solution for now.

p2p/state.py
```diff
@@ -116,24 +138,27 @@ def msg_queue_maxsize(self) -> int:
         if isinstance(cmd, ignored_commands):
             pass
         elif isinstance(cmd, eth.NodeData):
+            msg = cast(List[bytes], msg)
+            if peer not in self._requested_nodes:
+                # This is probably a batch that we retried after a timeout and ended up receiving
+                # more than once, so ignore but log as an INFO just in case.
+                self.logger.info(
+                    "Got %d NodeData entries from %s that were not expected, ignoring them",
```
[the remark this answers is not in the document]
gsalgado: Agreed. Filed #1098 for it, although in my (limited) testing, we're getting a lot less timeouts with those changes

p2p/state.py
gsalgado: I'd rather avoid doing that now, as it's tricky to exercise this code path and the non-mutative approach is not as clear as the current one. I'm sure I'd end up introducing bugs and we wouldn't uncover them easily. If only we had proper unit tests for those things... I really should take the time to write those

p2p/state.py
gsalgado: Sure, but what's `tnps` and why the different formatting lengths for every item?

p2p/state.py
pipermerriam: Maybe this would be easier to test if we isolated this functionality into something stand-alone

```
class StateSyncTracker:
   ...
```

It could house all of the statefullness of state syncing which would allow us to test it's APIs without needing to spin up the full state syncer?

pipermerriam: (let me know if it's unclear what I mean by this and I can try to flesh out an example a bit further)

gsalgado: Yeah, it'll definitely need some refactoring in order to be unit tested

p2p/state.py
pipermerriam: `tnps -> tree-nodes-per-second`, the numbers are the fixed width spacing so that each line is consistently easy to read (and I tried to pick numbers appropriate for the expected scale of the values we expect to see in those places)

p2p/state.py
gsalgado: I can take a stab at it, but I'm done for the day and will be off tomorrow, so this will have to wait until next week. I'll at least try to fix the issue above with the infinite loop when no peers have the nodes we want

p2p/state.py
gsalgado: Turns out it would take a big redesign of the whole `StateDownloader` in order to request a different batch as we don't have something like a back-burner area to put those node keys that no peers seem to have. I guess we'll have to go with option 1 and 2 when none our peers have the nodes. I'll give it some more thought, though, maybe I'll come up with something

gsalgado: I went ahead and implemented the above so now when no peers have the nodes in a batch, they're added to a list of missing node keys and retried after a while.

p2p/state.py
```diff
@@ -109,6 +130,21 @@ def msg_queue_maxsize(self) -> int:
...
+                self.scheduler.process([(node_key, node)])
+            except SyncRequestAlreadyProcessed:
+                # This means we received a node more than once, which can happen when we
+                # retry after a timeout.
+                pass
+            if idx % 10 == 0:
+                # XXX: This is a quick workaround for
+                # https://github.com/ethereum/py-evm/issues/1074, which will be replaced soon
+                # with a proper fix.
+                await self.sleep(0)
```
pipermerriam: should this be `self.sleep(0)`?

pipermerriam:
:shipit:

p2p/state.py
gsalgado: hmm, what do you mean? that's what it is

pipermerriam: No clue, I must have looked at this and read it as `asyncio.sleep`.  Disregard.

PR #1099, opened 2018-07-26
Use new network simulator library in tests

pipermerriam:
### What was wrong?

Some of our `p2p` tests use real networking.

### How was it fixed?

Change to use the new library `pytest-asyncio-network-simulator` which mocks out the networking.

#### Cute Animal Picture

cat-monkey-shutterstock-76988095-webonly

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,8 @@ deps = {
         "pytest-cov==2.5.1",
         "pytest-watch>=4.1.0,<5",
         "pytest-xdist==1.18.1",
+        # only needed for p2p
+        "pytest-asyncio-network-simulator==0.1.0a1;python_version>='3.6'",
     ],
     'lint': [
         "flake8==3.5.0",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -50,7 +50,7 @@
         "pytest-xdist==1.18.1",
         "pytest-watch>=4.1.0,<5",
         # only needed for p2p
-        "pytest-asyncio-network-simulator==0.1.0a1",
+        "pytest-asyncio-network-simulator==0.1.0a1;python_version>='3.6'",
```
cburgdorf: Does this syntax mean: Only install this if we are in a python 3.6 environment?

tests/p2p/conftest.py
```diff
@@ -30,3 +33,10 @@ def p2p_logger():
 
     return logger
 """
+
+
+@pytest.fixture
```
carver: Probably an `autouse=True` here?

pipermerriam: I find it hilarious that I forgot that.  Think I'll add a small test that verifies that asyncio is indeed patched.

PR #1100, opened 2018-07-26
convert to use CancelToken from libary

pipermerriam:
### What was wrong?

The `CancelToken` has been extracted into [a separate library `asyncio-cancel-token`](https://asyncio-cancel-token.readthedocs.io/en/latest/index.html).

### How was it fixed?

Convert all code to use the new library version.

#### Cute Animal Picture

cute-animals-animals-9403180-585-437


pipermerriam:
@gsalgado if you have any feedback I'll clean it up post merge.

p2p/discovery.py
```diff
@@ -42,7 +42,8 @@
 
 from eth_hash.auto import keccak
 
-from p2p.cancel_token import CancelToken
+from cancel_token import CancelToken
+
 from p2p.exceptions import NoEligibleNodes, OperationCancelled
```
gsalgado: We need to get rid of `p2p.exceptions.OperationCancelled` and use the one from cancel_token instead

pipermerriam: https://github.com/ethereum/py-evm/pull/1113

Issue #951, opened 2018-06-23
The --trinity-root-dir argument doesn't seem to have any effect at all


cburgdorf:
### What is wrong?

Trinity promotes `--trinity-root-dir` to overwrite the trinity root dir but we actually don't pick that up anywhere. In other words, passing `--trinity-root-dir` is absolutely useless right now unless I'm missing something.

### How can it be fixed

Revisit in combination with #922

pipermerriam:
closed by #1268

pipermerriam:
not quite merged, got confused.

Issue #922, opened 2018-06-17
Trinity does not attempt to create directories


MysticRyuujin:
So does Trinity also not create log files? 😄 

```
root@trinity01:~# ls /root/trinity/
logs
root@trinity01:~# trinity --port 30304 --data-dir /root/trinity
   ERROR  06-17 08:21:50        main  
It appears that /root/trinity/logs/trinity.log does not exist.
Trinity does not attempt to create directories outside of its root path
Either manually create the path or ensure you are using a data directory
inside the XDG_TRINITY_ROOT path
```

MysticRyuujin:
So obviously I need to set `--trinity-root-dir` but I think this error is still relevant

cburgdorf:
As long as the directory `/root/trinity/logs/` exists, that *should* work fine **on latest master** (was fixed by https://github.com/ethereum/py-evm/commit/4958c6ab9ba8819048c99877675d3da98a110326).

Can you double check if that works on latest master if the directory `/root/trinity/logs/` exists?

MysticRyuujin:
@cburgdorf I just recently (like 30 minutes ago) did a `pip3 install -U trinity` so I have whatever it most recently released?

See the top line of my output, the folder exists

```
root@trinity01:~# ls /root/trinity/
data  logs
root@trinity01:~# trinity --port 30304 --trinity-root-dir /root/trinity --data-dir /root/trinity/data
   ERROR  06-17 08:38:38        main  
It appears that /root/trinity/data/logs/trinity.log does not exist.
Trinity does not attempt to create directories outside of its root path
Either manually create the path or ensure you are using a data directory
inside the XDG_TRINITY_ROOT path
```

MysticRyuujin:
@cburgdorf if you've got quick instructions how to update the pip3 installed version to the github master version I'd be happy to run / /test it (Ubuntu)

cburgdorf:
I just checked the latest release and it didn't have the fix for this issue in yet. We plan to cut a new release this week though.

If you like to get your node on latest master, just follow along the [contributing guide](http://py-evm.readthedocs.io/en/latest/contributing.html).

carver:
Although the fix isn't released yet, also note that the logs are going inside the `--data-dir`, so you would have to manually create the folder `/root/trinity/data/logs` before the above `trinity` runner. (instead of `/root/trinity/logs`)

pipermerriam:
[...]

PR #1108, opened 2018-07-29
Implementing --trinity-root-dir

glaksmono:
### What was wrong?
Trinity promotes `--trinity-root-dir` to overwrite the trinity root dir but we actually don't pick that up anywhere. In other words, passing `--trinity-root-dir` is absolutely useless right now.

Issue Reference: https://github.com/ethereum/py-evm/issues/951, https://github.com/ethereum/py-evm/issues/922

### How was it fixed?
Implementing `--trinity-root-dir`

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/integration/test_trinity_cli.py
+++ b/tests/trinity/integration/test_trinity_cli.py
@@ -1,9 +1,31 @@
+import os
 import pytest
+import shutil
 
 from trinity.tools.async_process_runner import AsyncProcessRunner
 from trinity.utils.async_iter import (
-    contains_all
+    contains_all,
 )
+from trinity.utils.xdg import (
+    get_xdg_runtime_home,
+)
+
+
+XDG_RUNTIME_DIR = os.path.join(get_xdg_runtime_home(), '.test')
+
+
+@pytest.fixture(scope="function")
+def xdg_runtime_dir(monkeypatch):
+    assert not os.path.exists(XDG_RUNTIME_DIR)
+    os.makedirs(XDG_RUNTIME_DIR)
+    assert os.path.exists(XDG_RUNTIME_DIR)
+    monkeypatch.setenv('XDG_RUNTIME_DIR', XDG_RUNTIME_DIR)
+    try:
+        yield XDG_RUNTIME_DIR
+    finally:
+        assert os.path.exists(XDG_RUNTIME_DIR)
+        shutil.rmtree(XDG_RUNTIME_DIR)
+        assert not os.path.exists(XDG_RUNTIME_DIR)
 
 
 # IMPORTANT: Test names are intentionally short here because they end up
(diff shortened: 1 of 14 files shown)


glaksmono:
Before getting any further, would like to get an early feedback on the implementation. If ok, then go further to complete the design, implementing the root folder to the log folder, etc.

pipermerriam:
> I'm not so sure if TRINITY_DATA_DIR == TRINITY_ROOT_DIR

They are not the same

- `TRINITY_DATA_DIR` -> The root directory that trinity will use for the various chain data like the IPC connection for JSON-RPC, the location of the chain database, etc.
- `TRINITY_ROOT_DIR` -> The root directory that trinity will use to put data directories for the different chains it runs.

By default:

- `TRINITY_ROOT_DIR` -> `~/.local/share/trinity`
- `TRINITY_DATA_DIR`  -> `$TRINITY_ROOT_DIR/<chain-name>`

glaksmono:
@pipermerriam so we don't really need [this](https://github.com/ethereum/py-evm/blob/master/trinity/utils/xdg.py#L42) then?

glaksmono:
@pipermerriam so that means we won't default the `TRINITY_ROOT_DIR` to the `$HOME/.local/share/trinity` then ?

pipermerriam:
> so that means we won't default the `TRINITY_ROOT_DIR` to the `$HOME/.local/share/trinity` then ?

Yes, but the default value is based on the XDG spec

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

So I say we still need `get_xdg_trinity_root` since the root dir is defined as `$XDG_DATA_HOME/trinity`.

glaksmono:
> Yes, but the default value is based on the XDG spec

@pipermerriam sorry, can you please clarify what you meant by this. That is because `get_xdg_trinity_root` will return `$HOME/.local/share/trinity` as default.

pipermerriam:
Lemme try to clarify.

We need `get_xdg_trinity_root` for the case where it hasn't been overridden in the command line or via environment variable.

Currently, `trinity.utils.chains.get_local_data_dir` uses `get_xdg_trinity_root` to determine the appropriate data directory.  Since we're allowing the trinity root dir to be set via CLI, we need to do one of two things.

1. When `--trinity-root-dir` has been specified in the command line, we populate the environment variable `XDG_TRINITY_ROOT` with the provided value so that `get_xdg_trinity_root` returns the correct value.
2. Change the call signature to `get_local_data_dir` to instead require that the trinity root dir be explicitely provided.

I'm in favor of option 2 as I think it is less magic.

So, we need to modify all call sites to `get_local_data_dir` to provide the `trinity_root_dir`.  This is probably primarily in the `ChainConfig` object and tests.

tests/trinity/integration/test_trinity_cli.py
```diff
@@ -35,8 +35,8 @@ def async_process_runner():
 @pytest.mark.parametrize(
     'command',
     (
-        ('trinity',),
-        ('trinity', '--ropsten',),
+        ('trinity', '--trinity-root-dir .',),
+        ('trinity', '--trinity-root-dir .', '--ropsten',),
```
carver: Pretty sure this should be: `('trinity', '--trinity-root-dir', '.', '--ropsten',),`

Also: I think we should test trinity with & without the root dir specified:
```py
 ('trinity',),
 ('trinity', '--trinity-root-dir .',),
```

glaksmono:
So I would like to ask for a suggestion to name the functions in [this](https://github.com/ethereum/py-evm/pull/1108/files#diff-65654463a4049baf3b3285cb6316d27d) file

Before I made this change, I was getting this error: https://gist.github.com/glaksmono/6c01f72fbc0e83e8b7e208b821c0d052

To be specific, `AF_UNIX` path too long. That is the path to the IPC pipe is too long. The problem though, is that even if I'm just using the name of `test_t1`, it's giving that error already, so I'm not sure what kind of short name we want here.

cburgdorf:
@glaksmono 

>The problem though, is that even if I'm just using the name of `test_t1`, it's giving that error already,

Mmh, in that case, I *think* we should look how we can hook into the pytest directory generation. The current path should be something like `/private/var/folders/j1/r4dzzr892pd5v6wk_v78_kvc0000gn/T/pytest-of-glaksmono/pytest-23/popen-gw0/test_does_not_throw_command5_0/trinity-root-dir/<name-of-ipc-pipe>` and I *think* if we can hook into the process of *how* these are generated we may be able to trim it down (while still providing unique temp paths) to something like `/private/var/folders/j1/r4dzzr892pd5v6wk_v78_kvc0000gn/T/pytest-glaksmono/<uuid>/<name-of-ipc-pipe>`

I haven't looked into that myself but maybe the following link can be a good start.
https://docs.pytest.org/en/latest/reference.html#tmpdir-factory-api

pipermerriam:
Alternative is to use the stdlib [`tempfile`](https://docs.python.org/3.5/library/tempfile.html) library which I believe creates shorter paths.

pipermerriam:
replaced by #1268

PR #1109, opened 2018-07-30
Test core with py3.7

### LINKED ISSUE #970: Add testing for Python 3.7
### What is wrong?

Python 3.7 stable has been released: https://pythoninsider.blogspot.com/2018/06/python-3.html
We are not currently running any tests against it.

### How can it be fixed

Start running the core tests against py3.7 in addition to the current python envs.

veox:
### What was wrong?

Nothing; Python 3.7 is out, and it would be nice to start testing for compatibility.

### How was it fixed?

~~Implements~~ Closes #970.

The most basic set of tests, `core`, is now also run with a `circleci/python:3.7` image.

The image lacks some dependencies to build wheels locally (`gmp`, `leveldb`, likely others); so some packages are bumped to versions where wheels have Python 3.7 support. These are:

* `plyvel` - https://plyvel.readthedocs.io/en/latest/news.html#plyvel-1-0-5
* `coincurve` - https://pypi.org/project/coincurve/#id4

### Cute Animal Picture

Source: [friendlyferret](https://www.friendlyferret.com/how-many-types-of-ferrets-are-there/)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,6 @@
 [tox]
 envlist=
+    py37-core
     py{35,36}-{core,database,transactions,vm,native-blockchain}
     py{36}-{benchmark,p2p,trinity}
     py{36}-rpc-blockchain
@@ -45,6 +46,7 @@ deps = .[eth-extras, test]
 basepython =
     py35: python3.5
     py36: python3.6
+    py37: python3.7
 
 
 [testenv:py36-docs]
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -256,37 +256,37 @@ workflows:
   test:
     jobs:
       - py37-core
-      # - py36-docs
-      # - py36-native-state-byzantium
-      # - py36-native-state-frontier
-      # - py36-native-state-homestead
-      # - py36-native-state-eip150
-      # - py36-native-state-eip158
-      # - py36-rpc-state-byzantium
-      # - py36-rpc-state-frontier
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -1,5 +1,6 @@
 [tox]
 envlist=
+    py37-core
```
cburgdorf: To provide equal confidence for all three different Python versions I guess we'd need to change everything that runs on `35, 36` to `35, 36, 37` and everything that's currently been on `36` only to be `36, 37`. That will totally ruin our CI times though.

veox: Yes, seems so.

For ref, this PR (running in my CircleCI, not the `ethereum` org's) [took about an hour](https://circleci.com/workflow-run/897b3219-db10-4102-a59d-cb1ba6900397).

pipermerriam: We can increase our circle CI capacity so I'd recommend we expand the test suite out to fully cover the python versions we want to test against for all of our runs.

veox: I haven't checked the rest of the test suite with Python 3.7.

Do clarify: are you two suggesting it be done in this PR, or a follow-up one?..

pipermerriam:
Follow up would be fine with me, are you good having this merged as-is?

veox:
@pipermerriam Yes, that was my original intent: to have an indication that (at least) there is support for 3.7 by all the dependencies.

Issue #1060, opened 2018-07-20
Organize p2p and trinity modules


pipermerriam:
### What is wrong?

Conceptually the `p2p` module is intended to be *pure* DevP2P library code, and the `trinity` module should be all of the components built on top of the `p2p` module which are used to construct our ethereum node.

Currently, much of the Trinity code is living in the `p2p` module.

### How can it be fixed

Get all of the following moved to live under the `trinity` module.

- everything in `p2p.eth`
- everything in `p2p.les`
- everything in `p2p.lightchain`
- `p2p.peer`
  - minor refactor to `p2p.peer.BasePeer` to relocate `genesis` and `_local_chain_info`.
  - `p2p.peer..LESPeer`
  - `p2p.peer..EthPeer`
  - `p2p.peer..PeerPoolSubscriber`
  - `p2p.peer..PeerPool`
- everything in `p2p.rlp` ( 
- everything in `p2p.sedes`
- everything in `p2p.server`
- everything in `p2p.sharding_peer` and `p2p.sharding_protocol`, `p2p.shard_syncer`
- everything in `p2p.state`
- everything in `p2p.sync`
- `p2p.utils`
  - `p2p.utils.sequence_builder`

pipermerriam:
This is largely done.  There is still some functionality to be extracted from the `p2p` module but last I checked, there were some small things standing it the way.  The remaining cleanup can happen organically, or whenever we get the impetus to fully extract the `trinity` and `p2p` modules.

PR #1120, opened 2018-07-31
WIP: Start moving the ETH p2p protocol stuff into the trinity module

pipermerriam:
### What was wrong?

https://github.com/ethereum/py-evm/issues/1060

### How was it fixed?

- duplicated `p2p.rlp` and `p2p.sedes` into `trinity.rlp.block_body` and `p2p.rlp.sedes` respectively.
- setup new trinity module `trinity.protocols.eth`
- moved `p2p.peer.ETHPeer` -> `trinity.protocols.eth.peer`
- moved `p2p.eth`:
  - `trinity.protocols.eth.commands`
  - `trinity.protocols.eth.proto`
  - `trinity.protocols.eth.constants`
  - `trinity.protocols.eth.requests` (contains the `HeaderRequest`) object.
- same moves happened for `p2p.les -> trinity.protocol.les`

For `p2p` modules that still need access to these the following was done:

- for type hints, used `if TYPE_CHECKING` and used string version of value.
- for module level use, inlined imports within function bodies.

#### Cute Animal Picture

cute-animals-my-favorite-snuggle-buddy

### DIFF
--- a/p2p/chain.py
+++ b/p2p/chain.py
@@ -39,13 +39,11 @@ from eth.rlp.transactions import BaseTransaction, BaseTransactionFields
 from eth.utils.logging import TraceLogger
 
 from p2p import protocol
-from p2p import eth
-from p2p import les
 from p2p.cancellable import CancellableMixin
 from p2p.constants import MAX_REORG_DEPTH, SEAL_CHECK_RANDOM_SAMPLE_RATE
 from p2p.exceptions import NoEligiblePeers, ValidationError
 from p2p.p2p_proto import DisconnectReason
-from p2p.peer import BasePeer, ETHPeer, LESPeer, PeerPool, PeerSubscriber
+from p2p.peer import BasePeer, PeerPool, PeerSubscriber
 from p2p.rlp import BlockBody
 from p2p.service import BaseService
 from p2p.utils import (
@@ -57,9 +55,12 @@ from p2p.utils import (
 if TYPE_CHECKING:
     from trinity.db.chain import AsyncChainDB  # noqa: F401
     from trinity.db.header import AsyncHeaderDB  # noqa: F401
+    from trinity.protocol.eth.peer import ETHPeer  # noqa: F401
+    from trinity.protocol.les.peer import LESPeer  # noqa: F401
+    from trinity.protocol.base_request import BaseHeaderRequest  # noqa: F401
 
 
-HeaderRequestingPeer = Union[LESPeer, ETHPeer]
+HeaderRequestingPeer = Union['LESPeer', 'ETHPeer']
 
 
 class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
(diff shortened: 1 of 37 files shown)


pipermerriam:
There should be zero functionality changes in this PR.  Only moving code around.

I'm running this locally for a while to do some hands on testing for whether anything goes sideways.  Would like to merge soon so I can start on another section.  This is the type of work that is really hard to rebase.

p2p/chain.py
```diff
@@ -57,9 +55,12 @@
 if TYPE_CHECKING:
     from trinity.db.chain import AsyncChainDB  # noqa: F401
     from trinity.db.header import AsyncHeaderDB  # noqa: F401
+    from trinity.protocol.eth.peer import ETHPeer  # noqa: F401
+    from trinity.protocol.les.peer import LESPeer  # noqa: F401
+    from trinity.protocol.base_request import BaseHeaderRequest  # noqa: F401
```
carver: The goal is to cut all dependencies from p2p to trinity, right? Maybe all of p2p/chain needs to move to trinity. Or... and I hesitate to say it... should there be an eth-p2p?

p2p/shard_syncer.py
```diff
@@ -192,6 +188,8 @@ def msg_queue_maxsize(self) -> int:
         # to header existence at the moment. In the future we won't transfer collations anyway but
         # only collation bodies (headers are retrieved from the main chain), so there's no need to
         # add this at the moment.
+        from trinity.protocol.les.utils import gen_request_id
```
carver: Cross-package utils import?

p2p/sharding_peer.py
```diff
@@ -92,6 +89,7 @@ def handle_sub_proto_msg(self, cmd: Command, msg: protocol._DecodedMsgType) -> N
     async def get_collations(self,
                              collation_hashes: List[Hash32],
                              cancel_token: CancelToken) -> Set[Collation]:
+        from trinity.protocol.les.utils import gen_request_id
```
carver: cross-package utils import?

trinity/rlp/block_body.py
```diff
@@ -0,0 +1,13 @@
+import rlp
+from rlp import sedes
+
+
+from eth.rlp.headers import BlockHeader
+from eth.rlp.transactions import BaseTransactionFields
+
+
+class BlockBody(rlp.Serializable):
```
carver: Instead of copying this in both packages, maybe it could go into: https://github.com/ethereum/eth-rlp

pipermerriam:
@carver 

> I'm having a hard time seeing where this is going, unless it's to move a lot of p2p to trinity

This is correct.  The `p2p` module is going to get a lot smaller, ideally only generic DevP2P things.  Everything else moves to the `trinity` module.

This PR is me doing it incrementally because it's too much to do all at once, I'm planning on trying to railroad this through by end of week so we have another week to shake bugs out.

p2p/sharding_peer.py
pipermerriam: temporary, will be removed once everything has been migrated.

p2p/shard_syncer.py
pipermerriam: temporary for the duration of the migration.

p2p/chain.py
pipermerriam: This whole module will get moved, probably in the next wave.

trinity/rlp/block_body.py
pipermerriam: The duplication is simply a half-step towards moving it entirely into `trinity` as the `p2p` module should no longer need this.  I don't know of another codebase other than `trinity` that needs this.

PR #1124, opened 2018-07-31
Move HexaryTrieSync from py-trie here, and make it async

gsalgado:
Closes: #1074

### DIFF
--- a/tests/trinity/core/p2p-proto/test_state_sync.py
+++ b/tests/trinity/core/p2p-proto/test_state_sync.py
@@ -1,4 +1,3 @@
-import asyncio
 import os
 import random
 import time
@@ -19,8 +18,8 @@ from eth.db.backends.memory import MemoryDB
 from eth.db.account import AccountDB
 from eth.utils.logging import TraceLogger
 
+from trinity.sync.full.hexary_trie import HexaryTrieSync
 from trinity.sync.full.state import StateSync, TrieNodeRequestTracker
-from trinity.sync.full.trie import HexaryTrieSync
 
 from tests.trinity.core.integration_test_helpers import FakeAsyncMemoryDB
 
@@ -51,7 +50,7 @@ def make_random_trie(random):
 @settings(max_examples=10)
 @example(random=RandomWithSeed(EXAMPLE_37968))
 @example(random=RandomWithSeed(EXAMPLE_809368))
-def test_trie_sync(random):
+def test_trie_sync(random, event_loop):
 
     # Apparently hypothesis tests cannot be used in conjunction with pytest-asyncio yet, so do it
     # like this for now. https://github.com/HypothesisWorks/hypothesis/pull/1343
@@ -70,7 +69,7 @@ def test_trie_sync(random):
         for key, value in contents.items():
             assert dest_trie[key] == value
 
-    asyncio.get_event_loop().run_until_complete(_test_trie_sync())
+    event_loop.run_until_complete(_test_trie_sync())
 
 
 def make_random_state(n):
--- a/tests/trinity/core/p2p-proto/test_state_sync.py
+++ b/tests/trinity/core/p2p-proto/test_state_sync.py
@@ -1,12 +1,76 @@
+import asyncio
 import os
 import random
 import time
+
+import pytest
+
+from hypothesis import (
+    given,
+    settings,
+    strategies,
+    example,
+)
+from hypothesis.types import RandomWithSeed
+
+from trie import HexaryTrie
 
 from eth.db.backends.memory import MemoryDB
 from eth.db.account import AccountDB
 from eth.utils.logging import TraceLogger
 
 from trinity.sync.full.state import StateSync, TrieNodeRequestTracker
+from trinity.sync.full.trie import HexaryTrieSync
+
+from tests.trinity.core.integration_test_helpers import FakeAsyncMemoryDB
+
+
+# produces a branch node with an extention node who's encoding is less than 32
+# bytes in length so it is inlined.
+EXAMPLE_37968 = 37968
+
+# produces an top level extension node who's encoding is less than 32 bytes in
+# length so it gets inlined.
+EXAMPLE_809368 = 809368
+
+
+def make_random_trie(random):
+    trie = HexaryTrie({})
+    contents = {}
+    for _ in range(1000):
+        key_length = random.randint(2, 32)
+        key = bytes([random.randint(0, 255) for _ in range(key_length)])
+        value_length = random.randint(2, 64)
+        value = bytes([random.randint(0, 255) for _ in range(value_length)])
+        trie[key] = value
+        contents[key] = value
+    return trie, contents
+
+
+@given(random=strategies.randoms())
+@settings(max_examples=10)
+@example(random=RandomWithSeed(EXAMPLE_37968))
+@example(random=RandomWithSeed(EXAMPLE_809368))
+def test_trie_sync(random):
+
+    # Apparently hypothesis tests cannot be used in conjunction with pytest-asyncio yet, so do it
+    # like this for now. https://github.com/HypothesisWorks/hypothesis/pull/1343
+    async def _test_trie_sync():
+        src_trie, contents = make_random_trie(random)
+        dest_db = FakeAsyncMemoryDB()
+        scheduler = HexaryTrieSync(src_trie.root_hash, dest_db, TraceLogger("test"))
+        requests = scheduler.next_batch()
+        while len(requests) > 0:
+            results = []
+            for request in requests:
+                results.append([request.node_key, src_trie.db[request.node_key]])
+            await scheduler.process(results)
+            requests = scheduler.next_batch(10)
+        dest_trie = HexaryTrie(dest_db, src_trie.root_hash)
+        for key, value in contents.items():
+            assert dest_trie[key] == value
+
+    asyncio.get_event_loop().run_until_complete(_test_trie_sync())
 
 
 def make_random_state(n):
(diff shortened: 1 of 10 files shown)


gsalgado:
@pipermerriam I think this may have slipped under your radar?

tests/trinity/core/p2p-proto/test_state_sync.py
```diff
@@ -1,12 +1,76 @@
...
+            results = []
+            for request in requests:
+                results.append([request.node_key, src_trie.db[request.node_key]])
+            await scheduler.process(results)
+            requests = scheduler.next_batch(10)
+        dest_trie = HexaryTrie(dest_db, src_trie.root_hash)
+        for key, value in contents.items():
+            assert dest_trie[key] == value
+
+    asyncio.get_event_loop().run_until_complete(_test_trie_sync())
```
pipermerriam: Should we instead use the `event_loop` fixture provided by pytest?  I know these are supposed to be equivalent during runtime, but it seems slightly more correct to use the fixture version of this.

trinity/sync/full/trie.py
```diff
@@ -0,0 +1,231 @@
...
+        batch = list(reversed((self.queue[-n:])))
+        self.queue = self.queue[:-n]
+        return batch
+
+    async def schedule(self, node_key: Hash32, parent: SyncRequest, depth: int,
+                       leaf_callback: Callable[[bytes, 'SyncRequest'], Awaitable[None]],
+                       is_raw: bool = False) -> None:
+        """Schedule a request for the node with the given key."""
+        if node_key in self._existing_nodes:
+            self.logger.debug("Node %s already exists in db" % encode_hex(node_key))
```
pipermerriam: Logging statements should not do string formatting.  Can you change to `self.logger.debug("the %s message", encode_hex(nodekey))`

trinity/sync/full/trie.py
```diff
@@ -0,0 +1,231 @@
...
+            return
+        if await self.db.coro_exists(node_key):
+            self._existing_nodes.add(node_key)
+            self.logger.debug("Node %s already exists in db" % encode_hex(node_key))
```
pipermerriam: Same here, drop the string formatting and just provide the hex encoded `node_key` as a positional argument.

trinity/sync/full/trie.py
```diff
@@ -0,0 +1,231 @@
...
+        bisect.insort(self.queue, request)
+
+    def get_children(self, request: SyncRequest
+                     ) -> Tuple[List[Tuple[int, Hash32]], List[bytes]]:
+        """Return all children of the node retrieved by the given request.
+
+        :rtype: A two-tuple with one list containing the children that reference other nodes and
+        another containing the leaf children.
+        """
+        node = decode_node(request.data)
```
pipermerriam: This stands out as a potential source of trouble.  `request.data` can be `None` in cases where the `SyncRequest` has not been processed.  I trust that this code path is only called *after* the `data` attribute has been populated, but the implicitness of this approach makes me think we should look for an alternate pattern, maybe something like a `SyncResponse` object which takes the request and the data as an argument and exposes `response.data` which is always guaranteed to be populated.

trinity/sync/full/trie.py
```diff
@@ -0,0 +1,231 @@
...
+
+            if request.data is not None:
+                raise SyncRequestAlreadyProcessed("%s has been processed already" % request)
+
+            request.data = data
+            if request.is_raw:
+                await self.commit(request)
+                continue
+
+            references, leaves = self.get_children(request)
```
pipermerriam: Related to my other comment about `request.data` being potentially problematic.  At this call-site we'd instead pass in a `response` object so that we have a guarantee that the `data` attribute is present.

trinity/sync/full/trie.py
```diff
@@ -0,0 +1,231 @@
...
+                await self.schedule(ref, request, depth, request.leaf_callback)
+
+            if request.leaf_callback is not None:
+                for leaf in leaves:
+                    await request.leaf_callback(leaf, request)
+
+            if request.dependencies == 0:
+                await self.commit(request)
+
+    async def commit(self, request: SyncRequest) -> None:
```
pipermerriam: This function would also change to take a `SyncResponse` object.

trinity/sync/full/trie.py
gsalgado: get_children() is called precisely to find the children of the node we're processing, and `process()` does set `request.data` immediately before calling it. I guess if I inlined this method into `process()` it wouldn't raise a red flag, and we wouldn't make `process()` any more complex or less readable as this is a two-line method that just decodes the request data and passes it to `_get_children()`

pipermerriam: :+1: if it gets rid of the mutation of the `request` object and thus the implicit availability of the `request.data` attribute.

gsalgado: In order to completely get rid of the mutation of `request.data` I'll have to get rid of the check in `process()` that raises `SyncRequestAlreadyProcessed` when request.data is not None, but that is useless anyway as it is caught and ignored in `StateDownloader`

gsalgado: Hmm, spoke too soon. I thought I'd be able to have `process()` pass the `data` to `commit()` but that doesn't work because we don't always commit at the end of `process()` -- we only commit once all the child nodes have been retrieved and committed. IOW, we need to keep (somewhere) the data received for every request, until that request is eventually committed. I could get rid of `request.data`, but then I'd have to store the data for every request in an instance variable of `HexaryTrieSync`. The `SyncResponse` you suggest would also have to be stored somewhere, and since it's disjoint from the actual request I'm not sure it gives us any guarantees really -- we could still try to commit a request for which there's no `SyncResponse` in the same way we can try to commit one which has `.data` set to None

pipermerriam: If you can't find a way to do it that seems good, my fallback request would be to add some comments to document this deficiency.  Given that if `request.data` is `None` the two methods that access it will very likely fail loudly I'm ok with it staying in.

gsalgado: Ok, I've added a docstring to `commit()`, and the other method is gone as there was no point in having it anyway

gsalgado:
@pipermerriam I think I've addressed all your comments. wanna have another look?

trinity/sync/full/hexary_trie.py
```diff
@@ -0,0 +1,232 @@
...
+            self.logger.debug(
+                "Already requesting %s, will just update parents list", node_key)
+            existing.parents.append(parent)
+            return
+
+        request = SyncRequest(node_key, parent, depth, leaf_callback, is_raw)
+        # Requests get added to both self.queue and self.requests; the former is used to keep
+        # track which requests should be sent next, and the latter is used to avoid scheduling a
+        # request for a given node multiple times.
+        self.logger.debug("Scheduling retrieval of %s", encode_hex(request.node_key))
```
pipermerriam: This should probably be `TRACE` level logging?

trinity/sync/full/hexary_trie.py
```diff
@@ -0,0 +1,232 @@
...
+        batch = list(reversed((self.queue[-n:])))
+        self.queue = self.queue[:-n]
+        return batch
+
+    async def schedule(self, node_key: Hash32, parent: SyncRequest, depth: int,
+                       leaf_callback: Callable[[bytes, 'SyncRequest'], Awaitable[None]],
+                       is_raw: bool = False) -> None:
+        """Schedule a request for the node with the given key."""
+        if node_key in self._existing_nodes:
+            self.logger.debug("Node %s already exists in db", encode_hex(node_key))
```
pipermerriam: If this code path is likely to be encountered a lot we should move this to `TRACE` level logging.  Same with the one in the next `if` block.

trinity/sync/full/hexary_trie.py
```diff
@@ -0,0 +1,232 @@
...
+        """Process request results.
+
+        :param results: A list of two-tuples containing the node's key and data.
+        """
+        for node_key, data in results:
+            request = self.requests.get(node_key)
+            if request is None:
+                # This may happen if we resend a request for a node after waiting too long,
+                # and then eventually get two responses with it.
+                self.logger.debug(
```
pipermerriam: trace

trinity/sync/full/hexary_trie.py
```diff
@@ -0,0 +1,232 @@
...
+                    "No SyncRequest found for %s, maybe we got more than one response for it",
+                    encode_hex(node_key))
+                return
+
+            if request.data is not None:
```
pipermerriam: I think I've got a solution that'll work.  Lets remove the `data` property from the `SyncRequest` object and just use a boolean property like `is_done`.  We can set `is_done` here in this method, and just pass the raw `data` value into `self.commit`.  It looks like there are no other places where `request.data` is accessed.  This has the added benefit of reducing the memory overhead of this class.

trinity/sync/full/hexary_trie.py
```diff
@@ -0,0 +1,232 @@
...
+    def __init__(self, root_hash: Hash32, db: AsyncBaseDB, logger: TraceLogger) -> None:
+        self.queue: List[SyncRequest] = []
+        self.requests: Dict[Hash32, SyncRequest] = {}
+        self.db = db
+        self.root_hash = root_hash
+        self.logger = logger
+        # A cache of node hashes we know to exist in our DB, used to avoid querying the DB
+        # unnecessarily as that's the main bottleneck when dealing with a large DB like for
+        # ethereum's mainnet/ropsten.
+        self._existing_nodes: Set[Hash32] = set()
```
pipermerriam: This worries me a bit as it has `O(n)` storage cost.  For a full state trie sync I believe that means storing roughly 180 million hashes which is 5gb of memory.  Do you think we can get by with an LRU cache or something on the existence check and not actually store all of them?

trinity/sync/full/hexary_trie.py
gsalgado: We don't always commit at the end of `process()`, so we need to keep the data for a given request until we have committed all of its children; see my previous comment: https://github.com/ethereum/py-evm/pull/1124#discussion_r207978869

pipermerriam: Ah, missed that.  Ok, :shipit: unless you see a clean way around this.

trinity/sync/full/hexary_trie.py
gsalgado: I had a deja-vu feeling here so did a bit of digging and found https://github.com/ethereum/py-trie/pull/43#pullrequestreview-109756575

I've opened #1154 for it

PR #1129, opened 2018-07-31
[WIP] Test everything with py3.7 (not just core)

veox:
### What was wrong?

PR #1109 enabled testing with Python 3.7 for the `core` tests only.

### How was it fixed?

* New CircleCI jobs defined (mostly as copy-paste from Python 3.6 jobs).
* New `tox` environments defined, along with an "abstraction" of the common parts.

### Notes

* PR #1109 had a CircleCI run time of 1 hour for me. This is what external contributors could expect prior to this PR.
* With all jobs enabled, it takes 1:33:00.

### WIP

- [x] Does `docs` need to run with all Pythons?.. Currently, only `py36` is done (same as before).
- [x] Does `benchmmark` need to run with all Pythons?.. Currently, `py36` and `py37` are done (only `py36` was done previously).
- [x] Is linting with all Pythons required?.. Currently, all three are enabled (i.e. `py37` was added).
- [ ] After first fully-successful build, will rebase to clean up dirty commits, and get another measurement on total execution time.

### Cute Animal Picture

Source: [EidolonWeasel@flickr](https://www.flickr.com/photos/45043374@N07/4556825624)


carver:
IMO, we could maybe get away with only adding py37 to the trinity-integration tests and core. Since state & rpc tests take so long. That would mean that:

>    Does docs need to run with all Pythons?.. Currently, only py36 is done (same as before).

No
>    Does benchmmark need to run with all Pythons?.. Currently, py36 and py37 are done (only py36 was done previously).

No, only 3.6 is fine.
>    Is linting with all Pythons required?.. Currently, all three are enabled (i.e. py37 was added).

No, part of the reason for the split is that py-evm supports 3.5+ and p2p/trinity supports 3.6+, so they are each linted at the minimum supported version.

pipermerriam:
Generally agree with what carver said.  I'm good dropping the state tests (both core and rpc based).

veox:
> IMO, we could maybe get away with only adding py37 to the trinity-integration tests and core. Since state & rpc tests take so long.

Sounds reasonable to me. It's late here, will continue another day. (That also gets us a "longest run" measurement.)

veox:
Closing in favour of PR #1135.

PR #1112, opened 2018-07-30
Implement bitwise shift opcodes for Constantinople

cburgdorf:
### What was wrong?

As stated in https://github.com/ethereum/py-evm/issues/1104, the upcoming Constantinople fork adds three new opcodes for bitwise shifting operations. Namely `SHL`, `SHR` and `SAR`. This PR adds support for the `SHL` and `SHR` opcodes.

This does **not** include support for the `SAR` opcode yet which is expected to land in another PR. https://github.com/ethereum/py-evm/pull/1134

### How was it fixed?

1. Created a new `test_opcodes` file and wrote a couple of tests for existing opcodes. This isn't finished but from the perspective of this PR it shouldn't matter and we have an open issue to not forget about testing other existing opcodes #1111 

2. Implemented `SHR` and `SHL` logic in `eth.vm.logic.arithmetic`
3. Added all test cases from the [EIP](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-145.md#test-cases)
3. Added opcodes that expose the functionality to the constantinople opcodes

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1134, opened 2018-08-01
Implement SAR opcode

cburgdorf:
### What was wrong?

This is a follow up to #1112 that also includes the `SAR` opcode. I branched this off because I feel less confident about this one compared to the other ones. That said, it passes all test cases that are specified in the EIP.

### How was it fixed?

Implemented new opcode.

#### Cute Animal Picture


eth/vm/logic/arithmetic.py
```diff
@@ -205,3 +205,18 @@ def shr(computation):
...
+
+def sar(computation):
+    """
+    Arithmetic bitwise right shift
+    """
+    shift_length, value = computation.stack_pop(num_items=2, type_hint=constants.UINT256)
+    value = unsigned_to_signed(value)
+
+    if shift_length >= 256:
+        result = 0 if value >= 0 else signed_to_unsigned(-1)
```
pipermerriam: Can we *constantify* `signed_to_unsigned(-1)`

eth/vm/logic/arithmetic.py
```diff
@@ -205,3 +205,18 @@ def shr(computation):
         result = (value >> shift_length) & constants.UINT_256_MAX
 
     computation.stack_push(result)
+
+
+def sar(computation):
+    """
+    Arithmetic bitwise right shift
+    """
+    shift_length, value = computation.stack_pop(num_items=2, type_hint=constants.UINT256)
```
pipermerriam: Side note, I just realized that we use keyword style calling for all stack operations, and a while back @davesque uncovered that calling things with positional arguments is significantly faster than keyword style calling.  We should look into updating all of these call sites to use positional arguments.

eth/vm/logic/arithmetic.py
cburgdorf: Makes sense! However, I'm not sure if I got the naming right because my bits and bytes fu is pretty bad.

I [added](https://github.com/ethereum/py-evm/pull/1134/files#diff-1a697f3a2088726b12d5e69c8fee879eR16) the following constant to `eth.constants`

```py
UINT_255_NEGATIVE_ONE = -1 + UINT_256_CEILING
```
My understanding is that this is representing the number `-1` according to the rules of the [two's complement](https://en.wikipedia.org/wiki/Two's_complement) which uses the most significant bit to represent the sign and leaving us with 255 other bits to represent the number.

Btw, the constant basically inlines what `signed_to_unsigned` does because I can't import it here (circular import)

cburgdorf: @pipermerriam interpreted :+1: as indicator that the constant name seems reasonable and merged it

PR #1135, opened 2018-08-01
Run trinity integration tests with py3.7

veox:
### What was wrong?

PR #1109 enabled testing with Python 3.7 for the `core` tests only. This is useful to show support by dependencies, but not much else.

### How was it fixed?

This PR enables `trinity{,-integration}` tests. As discussed in #1129, the rest aren't very useful (yet), and would prolong CI runs (by about 50%).

Closes #1129 as superceded. (This is a squashed/rebased version.)

### Note: two new `tox` configuration features used!

(Some [discussion below](https://github.com/ethereum/py-evm/pull/1135#discussion_r207134409) that urged me to add this, so the fact doesn't slip through.)

#### Factor-conditional settings

See [relevant `tox` doc section](https://tox.readthedocs.io/en/latest/config.html#factors-and-factor-conditional-settings).

**TL;DR:** a "factor" is a segment of a `testenv`'s name. E.g., `testenv:py37-trinity-integration` has 3 factors: `py37`, `trinity`, and `integration`. A factor can be used (e.g. by the "generic" `testenv`) to apply settings conditionally to other environments that have it in their name.

These were already being specified in the "generic" `testenv`, as:

https://github.com/veox/py-evm/blob/27ebdf83481540098413de90af54d98077a7e8ed/tox.ini#L46-L49

``` python
basepython =
    py35: python3.5  # factor `py35`: use `basepython=python3.5`
    py36: python3.6  # factor `py36`: use `basepython=python3.6`
    py37: python3.7  # factor `py37`: use `basepython=python3.7`
```

However, other `testenv`s still had their own `basepython` lines, effectively overriding what's set here.

This PR removes `basepython` from all `testenv`s but this one, meaning that the setting is picked up based on a factor.

#### Other-section substitutions

See [relevant `tox` doc section](https://tox.readthedocs.io/en/latest/config.html#substitution-for-values-from-other-sections).

**TL;DR:** curly-brackety syntax like `X = {[S]Y}` will substitute the value of var `Y` from section `S` to var `X`.

Otherwise, there would be some copy-paste repetition, e.g. in `py3{6,7}-trinity-integration`; so I moved the common part to a section called `common-trinity-integration` (_not_ `testenv:common-trinity-integration`, since it's not an actual `testenv` - just a helper).

This doesn't reduce the line count (actually, increases it!..), but I thought it's less error-prone than copy-paste sections.

The same could have been achieved using factor-conditional settings - for example, for the `lint`/`integration` `testenv`s touched in this PR, we could've had:

``` python
[testenv]
# ...
deps=
    integration: .[eth-extras, test]
    lint: .[lint]
setenv=
    lint: MYPYPATH={toxinidir}:{toxinidir}/stubs
passenv=
    integration: ACTUALLY NOT SURE ABOUT VALID SYNTAX WITHOUT NEWLINES
# ...
```

But that looks much less readable to me, and piles settings from various environments into the same spot...

### Cute Animal Picture

baby black-footed ferret

([Image source](https://www.abqjournal.com/814838/black-footed-ferrets-return-to-where-they-held-out-in-wild.html))

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -72,20 +72,17 @@ commands=
     py.test -n 1 {posargs:tests/trinity/integration/}
 
 [testenv:py36-trinity-integration]
-basepython=python3.6
 deps = {[common-trinity-integration]deps}
 passenv = {[common-trinity-integration]passenv}
 commands = {[common-trinity-integration]commands}
 
 [testenv:py37-trinity-integration]
-basepython=python3.7
 deps = {[common-trinity-integration]deps}
 passenv = {[common-trinity-integration]passenv}
 commands = {[common-trinity-integration]commands}
 
 
 [testenv:py36-benchmark]
-basepython=python3.6
 deps = .[eth-extras, benchmark]
 commands=
     benchmark: {toxinidir}/scripts/benchmark/run.py
@@ -98,7 +95,6 @@ commands=
     flake8 {toxinidir}/eth
 
 [testenv:lint-py35]
-basepython=python3.5
 deps = {[common-lint]deps}
 setenv = {[common-lint]setenv}
 commands=
@@ -108,7 +104,6 @@ commands=
     mypy --follow-imports=silent --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
 
 [testenv:lint-py36]
-basepython=python3.6
 deps = {[common-lint]deps}
 setenv = {[common-lint]setenv}
 commands=
--- a/tox.ini
+++ b/tox.ini
@@ -1,11 +1,11 @@
 [tox]
 envlist=
-    py37-core
     py{35,36}-{core,database,transactions,vm,native-blockchain}
     py{36}-{benchmark,p2p,trinity}
     py{36}-rpc-blockchain
     py{36}-rpc-state-{frontier,homestead,eip150,eip158,byzantium,quadratic}
     py{35,36}-native-state-{frontier,homestead,eip150,eip158,byzantium,constantinople,metropolis}
+    py37-{core,trinity,trinity-integration}
     lint-py{35,36}
 
 [flake8]
@@ -61,9 +61,8 @@ commands=
     make validate-docs
 
 
-[testenv:py36-trinity-integration]
+[common-trinity-integration]
 deps = .[eth-extras, test]
-basepython=python3.6
 passenv =
     PYTEST_ADDOPTS
     TRAVIS_EVENT_TYPE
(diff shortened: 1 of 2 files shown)


pipermerriam:
Somethign looks wrong, this seems to only be running 7 CI runs when there should be something like 20+

veox:
Yeah, should be 37. I think the CircleCI integration doesn't _display_ them as it usually does, but adds them in one-by-one. It's 11 by now (4 pending, 7 successful).

pipermerriam:
> but is adding them in one-by-one.

Well that's super confusing.

veox:
What's even more confusing is CircleCI's website showing 37 successful jobs in this run, whereas Github lists only 31. :/

I'll bump this with a dummy commit to force re-run.

tox.ini
```diff
@@ -72,37 +70,47 @@ commands=
     # due to multiple Trinity instances competing for the same ports
     py.test -n 1 {posargs:tests/trinity/integration/}
 
+[testenv:py36-trinity-integration]
+deps = {[common-trinity-integration]deps}
+passenv = {[common-trinity-integration]passenv}
+commands = {[common-trinity-integration]commands}
```
cburgdorf: @veox Nice! Didn't know that syntax was possible. Am I right to assume that we are now entirely relying on the Python version that is available on the system (e.g. just taking the one that is defined in the circle image)? So, if one runs these `tox` jobs locally `testenv:py36-trinity-integration` and `testenv:py37-trinity-integration` will not use different Python versions. If so, that would be a small regression. Not sure how important that is for anyone. I'm also not sure if there's maybe another command line argument one could pass to `tox` to specify the python version for each job anyway. I'm a bit in a hurry so not looking it up.

veox: As I understand, specifying `basepython` in `testenv:py36-trinity-integration` was overriding what was specified in `testenv`. Effectively, `testenv` had (and [still has](https://github.com/veox/py-evm/blob/27ebdf83481540098413de90af54d98077a7e8ed/tox.ini#L46)) instructions to use `basepython=python3.6` for all environments/tests that have `py36` in their name (such as `testenv:py36-trinity-integration`).

That's my understanding - but I could be wrong. Need to check...

veox: In [a different branch](https://github.com/veox/py-evm/commits/run-trinity-integration-tests-with-py3.7-sloppy-check), I've set `basepython` to `false` for `py36` jobs (see commit https://github.com/veox/py-evm/commit/5de2944f16b361a12ec11b1882ce17591dc30536).

All CircleCI jobs that were running `tox` `testenv`s with `py36` in their name failed (e.g. [this one](https://circleci.com/gh/veox/py-evm/846)). They are certainly not using the `python` available on the system, but do indeed pick up the one specified in the "generic" `testenv`.

cburgdorf:
Ah! That makes sense!

tox.ini
veox: To clarify, the particular line you're commenting on, with curly syntax like `{[common-trinity-integration]deps}`, is a different configuration feature, [substitutions](https://tox.readthedocs.io/en/latest/config.html#substitution-for-values-from-other-sections). It instructs to use the value from a different section, `common-trinity-integration`.

PR #1136, opened 2018-08-01
Finish moving trinity code from p2p module

pipermerriam:
### What was wrong?

continues work on #1060 

### How was it fixed?

Migrated the remaining code (no functionality changes).

#### Cute Animal Picture

1967a305c291ffe0507f5bb10dcc356c


pipermerriam:
Next after this is to do some deeper re-factoring to `BasePeer` to move all Trinity functionality to a new common base class for `LESPeer` and `ETHPeer`

PR #1137, opened 2018-08-01
PeerSubscribers now specify what messages they are interested in

### LINKED ISSUE #1055: Consider allowing PeerPoolSubscribers to specify types of messages they're not interested in
### What is wrong?

Some PeerPoolSubscribers (e.g. `LightPeerChain`, `TxPool`) are only meant to handle a subset of msg types, so it is rather wasteful to have `PeerPool` deliver all messages to their queues.

### How can it be fixed

We could make it possible for `PeerPoolSubscribers` to specify what types of msgs they aren't interested in, so that they never get those. Or they could specify the types of msgs they are interested in, but the problem with that is we need to remember to update the list if/when they're extended to handle more msg types

pipermerriam:
### What was wrong?

fixes #1055

### How was it fixed?

All `PeerSubscriber` classes must now implement a `subscription_msg_types` which defines which types of messages it will add to it's queue.  Any message not the specified types is discarded.  Including the base `p2p.protocol.Command` in the list results in all messages being included in the queue.

#### Cute Animal Picture

wenn21048138

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -2,6 +2,7 @@ import asyncio
 import collections
 import contextlib
 import datetime
+import functools
 import logging
 import operator
 import random
@@ -20,6 +21,7 @@ from typing import (
     Dict,
     Iterator,
     List,
+    Set,
     TYPE_CHECKING,
     Tuple,
     Type,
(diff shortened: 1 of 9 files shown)


pipermerriam:
@gsalgado I still want to make some changes to the API so hold off on reviewing for now.  Sorry for the pre-mature request.

pipermerriam:
ok, ready for review

pipermerriam:
Two things

1. Something on `master` recently is causing a significant performance hit.  I'm seeing block import times on master up to around 45s or more for batches of 192 blocks.
2. This PR seems to provide a measurable performance increase, as block import times on this branch are closer to the 30s mark.

Will investigate further tomorrow but wanted to make note of it here.

p2p/peer.py
```diff
@@ -547,10 +558,28 @@ def __hash__(self) -> int:
...
 
+    @property
+    @abstractmethod
+    def subscription_msg_types(self) -> Set[Type[protocol.Command]]:
+        """
+        The `p2p.protocol.Command` types that this class subscribes to.  Any
+        command which is not in this set will not be passed to this subscriber.
+
+        The base command class `p2p.protocol.Command` can be used to enable
+        **all** command types.
```
gsalgado: It might be worth mentioning that only sub-proto messages are delivered this way -- the base protocol ones are handled exclusively at the peer level

p2p/peer.py
```diff
@@ -577,16 +606,30 @@ def msg_queue(self) -> 'asyncio.Queue[PEER_MSG_TYPE]':
     def queue_size(self) -> int:
         return self.msg_queue.qsize()
 
-    def add_msg(self, msg: 'PEER_MSG_TYPE') -> None:
+    def add_msg(self, msg: 'PEER_MSG_TYPE') -> bool:
         peer, cmd, _ = msg
+
+        if not self.is_subscription_command(type(cmd)):
+            if hasattr(self, 'logger'):
```
gsalgado: Which subscribers don't provide a logger? Should we change them so they do?

pipermerriam: This is actually for the tests.  The `PeerSubscriber` itself doesn't provide a logger.  I'll remove this and change the ones in the test to have loggers.

PR #1138, opened 2018-08-01
Establish pattern for generic round trip request response handling

pipermerriam:
### What was wrong?

Currently, the way that we implement the `Peer.get_block_headers` API is very boilerplate heavy.  Doing it this way for all of the various requests and response pairs is not going to be maintainable.

### How was it fixed?

- Created new container class that I'm calling a `RequestResponseHandler` which exists on the `Peer` class under the `Peer.handler` property
- Create a new base class `BaseRequestManager` which implements the generic logic for doing request/response handling.
- Converted the `get_block_headers` API on `LESPeer` and `ETHPeer` to instead use implementations of these APIs.

Still to be done:  Figure out the types.

#### Cute Animal Picture


pipermerriam:
Any of @carver or @cburgdorf , can I get a high level review on this?  It's basically ready for a real review, but I'm struggling a bit with how to do the type hinting in a sane way.

carver:
Integration test, tail end of stack trace:

```
'  File "/home/circleci/repo/p2p/peer.py", line 684, in start_peer\n'
'    msgs = await self.ensure_same_side_on_dao_fork(peer)\n'
'  File "/home/circleci/repo/p2p/peer.py", line 793, in ensure_same_side_on_dao_fork\n'
'    request = peer.request_block_headers(  # type: ignore\n'
"AttributeError: 'ETHPeer' object has no attribute 'request_block_headers'\n"
```

trinity/protocol/common/handlers.py
```diff
@@ -0,0 +1,43 @@
...
+                    "Unable to set manager on attribute `{0}` which is already "
+                    "present on the class: {1}".format(attr, getattr(self, attr))
+                )
+            manager = request_manager_cls(self._peer, self.cancel_token)
+            setattr(self, attr, manager)
+
+    async def _run(self) -> None:
+        for attr in self._managers.keys():
+            manager = getattr(self, attr)
+            self.run_child_service(manager)
```
carver: So we're talking about `num_peers * num_commands` services running, right? Not sure how many commands there will be, but let's say eventually 50. If someone wanted to experiment with 200 peers, we're talking 10,000 different services running. Not sure when overhead of multiple tasks starts to overwhelm asyncio, but I wouldn't be surprised if it was an issue at 10k.

trinity/protocol/common/handlers.py
```diff
@@ -0,0 +1,43 @@
...
+
+        while not self.cancel_token.triggered:
```
carver: or the other common construction: `while self.is_running:`

trinity/protocol/common/managers.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def __init__(self, peer: PeerClass, token: CancelToken) -> None:
+        self._peer = peer
+        super().__init__(token)
+
+    #
+    # Service API
+    #
+    async def _run(self) -> None:
+        self.logger.debug("Running %s for peer %s", self.__class__.__name__, self._peer)
```
carver: Maybe: ~Running~ Launching

trinity/protocol/common/managers.py
```diff
@@ -0,0 +1,147 @@
...
+                    self._handle_msg(cast(ResponseType, msg))
+                else:
+                    self.logger.warning("Unexpected message type: %s", cmd.__class__.__name__)
+
+    async def _cleanup(self) -> None:
+        pass
+
+    def _handle_msg(self, msg: ResponseType) -> None:
+        if self.pending_request is None:
+            return
```
carver: Maybe worth a debug message that a peer sent an unexpected msg.

pipermerriam:
@carver thnx, got a fix in for that, still something wrong that I haven't sorted out, but the overall structure of this is still fair game to review.

p2p/peer.py
```diff
@@ -790,47 +787,40 @@ def _add_peer(self,
...
+
+                with msg_buffer.subscribe_peer(peer):
+                    headers = await peer.handler.get_block_headers(  # type: ignore
+                        start_block,
+                        max_headers=2,
+                        reverse=False,
+                        timeout=CHAIN_SPLIT_CHECK_TIMEOUT,
+                    )
+
+                msgs = [msg_buffer.msg_queue.get_nowait()[1:] for _ in range(msg_buffer.queue_size)]
```
carver: Took me a bit to parse/remember that this was for capturing inbound requests for later processing. `MsgBuffer` and this is a lot of stuff that doesn't really have to do with the DAO fork. Maybe push it into `BasePeer`, which might get reused elsewhere:

```py
with peer.buffer_inbound_commands() as buffer:
  headers = await peer.handler.get_block_headers(...

msgs = buffer.get_messages()
```

p2p/peer.py
```diff
@@ -790,47 +787,40 @@ def _add_peer(self,
...
             except ValidationError as err:
                 raise DAOForkCheckFailure(
                     "Invalid header response during DAO fork check: {}".format(err)
-                )
+                ) from err
 
-            if len(headers) != 2:
-                raise DAOForkCheckFailure(
-                    "Peer failed to return all requested headers for DAO fork check"
-                )
```
carver: Is this removed so we can connect to peers who haven't sync'd that far yet?

trinity/protocol/eth/peer.py
```diff
@@ -20,21 +14,28 @@
...
 
 
 class ETHPeer(BasePeer):
     _supported_sub_protocols = [ETHProtocol]
     sub_proto: ETHProtocol = None
 
+    _handler: ETHRequestResponseHandler = None
+
+    @property
+    def handler(self) -> ETHRequestResponseHandler:
```
carver: Handler alone is such a generic word: it could mean connection handler, service handler, etc.

Could it just be named for the sub protocol it uses, like:

```py
peer = EthPeer(...)
peer.eth.get_block_headers(...)
```

and

```py
peer = LESPeer(...)
peer.les.get_proof(...)
```
etc. Reading it would always be explicit about which subprotocol is being used.

trinity/protocol/common/handlers.py
pipermerriam: Formula is a little off but it doesn't discount your concern entirely.

`num_peers * `num_command_pairs`

Currently I think the `eth` protocol has 4-5 command pairs and the LES protocol has maybe 8-10.  So only request/response command pairs. That at least pushes the overhead of this down by a decent bit, and for now, I really like how clean it is for this to be self contained.  We can reduce the overhead by moving the subscription from the `Manager` classes down to the `Handler` classes and having the handler dispatch to each of the mangers which moves it from `O(n)` to `O(1)` per peer.

trinity/protocol/common/handlers.py
pipermerriam: Realizing I can drop the inner loop (and thus some task switching overhead) by removing this loop and replacing with `await self.cancel_token.wait()`

p2p/peer.py
pipermerriam: Ooh, I like this.

p2p/peer.py
pipermerriam: No, it's removed because the response will have already been validated to contain exactly 2 headers as part of the core API for request/response validation.

pipermerriam:
@carver  This is ready for another pass.  Changes should be in the newest commit

- `Peer` now starts running before we enter the dao fork check, but only gets added to the `PeerPool` on successful check.  This has be benefit of being able to leverage the peer's subscription API for requesting headers rather than requiring manual message processing.
- `Peer.handler` has been renamed to `Peer.requests`
- New API on peer, `Peer.collect_sub_proto_msgs` which can be used like a context manager to gather up all sub protocol messages within a specific context block.

pipermerriam:
One more change pushed which allows adding a `finished_callback` after a service has been started.  This allows us to only add the finished callback to the `Peer` if it passes the dao fork check.

pipermerriam:
Looking at the diff I think I'm going to tease this apart into 2-3 pull requests.  I think I can lift some of the functionality that's been added to the peer class and the service class to it's own pull requests which will clean this up quite a bit.

pipermerriam:
I was optimistic I could take this apart into smaller pieces but upon trying they are all kind of woven together.

p2p/peer.py
```diff
@@ -695,28 +700,33 @@ def unsubscribe(self, subscriber: PeerSubscriber) -> None:
...
     async def start_peer(self, peer: BasePeer) -> None:
+        asyncio.ensure_future(peer.run())
+        await self.wait(peer.events.started.wait(), timeout=1)
         try:
             # Although connect() may seem like a more appropriate place to perform the DAO fork
             # check, we do it here because we want to perform it for incoming peer connections as
             # well.
-            msgs = await self.ensure_same_side_on_dao_fork(peer)
+            with peer.collect_sub_proto_messages() as buffer:
+                await self.ensure_same_side_on_dao_fork(peer)
```
cburgdorf: I know it's unrelated to this PR and I remember silently reading a discussion around this a while back so please forgive if my comment doesn't make sense but it bothers me (probably same as everyone else) that we have this dao check built right into the `PeerPool`. Did we consider to allow passing something like a `precondition_predicate` (name tbd) to the `PeerPool` that the `PeerPool` uses to check whether peers qualify for further communication. That way, we don't hardcode the dao check here and keep things flexible for other chains.

trinity/protocol/common/managers.py
```diff
@@ -0,0 +1,154 @@
...
+        else:
+            future.set_result(self._normalize_response(msg))
+            self.pending_request = None
+
+    @abstractmethod
+    def _normalize_response(self, msg: ResponseType) -> ReturnType:
+        pass
+
+    @abstractmethod
+    def __call__(self) -> ReturnType:  # type: ignore
```
cburgdorf: You can drop that `ignore` here. Ignoring the violating places is enough. However, maybe we can get rid of the violations altogether. I'll elaborate in another comment.

trinity/protocol/eth/managers.py
```diff
@@ -0,0 +1,57 @@
...
+    Tuple[BlockHeader, ...],
+]
+
+
+class GetBlockHeadersRequestManager(BaseRequestManager):
+    msg_queue_maxsize = 100
+
+    _response_msg_type: Type[Command] = BlockHeaders
+
+    async def __call__(self,  # type: ignore
```
cburgdorf: The signature doesn't align with the parent. The signature here (the child) is

```py
 __call__(self, # type: ignore
                       block_number_or_hash: BlockIdentifier,
                       max_headers: int = None,
                       skip: int = 0,
                       reverse: bool = True,
                       timeout: int = None) -> Tuple[BlockHeader, ...]
```

The signature in the parent is:

```py
def __call__(self) -> ReturnType:
```

While the `ReturnType` matches just fine, notice that the parent method is defined as one without any arguments whereas the child implementation does have arguments. That is violating the type safety rules.

What we could do is to define `__call__` in the parent as `__call__(self, req_params: TRequest) -> TReturn:`
Then, in the child it's perfectly valid to implement it as `__call__(self, req_params: HeaderRequest) -> Tuple[BlockHeader, ...]:`. So, instead of passing the arguments as individual arguments that mypy doesn't know about, we use the type of the request that the subclass is specialized for (as specified via `BaseRequestManager[T1, T2, T3, T4]`

Hope that helps.

cburgdorf: Btw, there is also the alternative to define `__call__` in parent and child as `*args: Any` but that gives up type safety whereas the previous solution doesn't.

p2p/peer.py
pipermerriam: @cburgdorf one of the tasks that I don't think there is a detailed issue for but is likely to be a core part of one of the subsequent milestones is to tease out the rest of this EVM logic from the `p2p` module.  Exactly how isn't clear but it might look something like you mentioned.

trinity/protocol/eth/managers.py
pipermerriam: This is intentional.  I spent a decent amount of time trying to figure out a sane way to do this and this is the result.  Here are the properties I was able to achieve.

- the `__call__` method is required to be implemented on subclasses.
- call sites which call this class still do correct type checks on the function signature (i.e. we don't loose type safety when this function is being called)
- all `__call__` implementations on subclasses require a `# type: ignore` comment.
- we don't have to manually instantiate a `Request` object in order to use the API (**this one is important**)

I've added a comment to both of the `Manager` subclasses noting this.

I'm open to alternate solutions but all that I've seen have trade-offs that I'm not ok with (like requiring a `Request` object to use the API.

pipermerriam: It's not a big deal, but it exposes the underlying implementation which bothers me.  It means every single call site for `foo` must import `Request` when really the `Request` class is an internal API that callers should not be exposed to.

cburgdorf: I see. It doesn't *have* to be `Request` though. We could add one more generic type which could be set to something as lightweight as `Tuple[int, int, str]` and make call sites look like `foo((1, 2, 'bar'))`.

I wanted to find out if there's any other way how we could constrain an abstract method on only the return type and created a question on SO. But there doesn't seem to be anything and someone also [pointed out](https://stackoverflow.com/questions/51720280/is-it-possible-to-constrain-an-abstract-base-method-only-on-the-return-type-bu/51724793#comment90401496_51720280) that it's violating the [Liskov Substitution Principle](https://en.wikipedia.org/wiki/Liskov_substitution_principle).

But anyway, I didn't mean to bikeshed about it. If it works for you, it works for me :)

pipermerriam: I don't think the `Tuple[...]` example works because the signatures of these will vary widely in both argument type and count, so we'd have to use `Tuple[Any, ...]`

With this we'd end up with satisfying the type checker while losing a degree of type safety on the actual function calls (at least I think we would).

Right now this works for me.  I'm open to an alternative if someone wants to take the time to figure it out.

PR #1140, opened 2018-08-03
Simplify sequence_builder

carver:
### What was wrong?

Nothing really, just thought it could be a little simpler and easier to read

### How was it fixed?

Used a generator comprehension instead of a generator function with a `@to_tuple` decorator. 

I suppose it technically is slower on sequences which go significantly out of bounds, but that should be unusual.

#### Cute Animal Picture


carver:
Ready for re-review.

I thought how to justify any given limit. It seemed like in-memory size of the object was probably the key limiter. I chose a somewhat arbitrary 10MB maximum for a given sequence tuple. That limits the sequence to ~1.25M entries.

carver:
Ok, not sure how much it protects us. Also, it carries some risk that it catches us by surprise in a few decades when everything starts exploding... So I'll leave it out for now.

Issue #719, opened 2018-05-18
EIP-1052 EXTCODEHASH opcode


pipermerriam:
### What is wrong?

https://eips.ethereum.org/EIPS/eip-1052

### How can it be fixed

Implement it.

PR #1147, opened 2018-08-06
Implement EXTCODEHASH opcode for Constantinople

cburgdorf:
### What was wrong?

Missing the EXTCODEHASH opcode as specified for Constantinople (#719)

### How was it fixed?

Implemented it. 

#### Known issues

- `test_opcodes` got more messy as we are reusing one computation setup strategy across all tests. (though, I think it's not **that** bad yet)
- some [test cases](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1052.md#test-cases) are rather vague to me and I couldn't find geth or parity covering these either

#### Cute Animal Picture


eth/vm/opcode_values.py
```diff
@@ -56,6 +56,7 @@
 GASPRICE = 0x3a
 EXTCODESIZE = 0x3b
 EXTCODECOPY = 0x3c
+EXTCODEHASH = 0x3f
```
carver: *nitpick*

The other opcodes are in byte order, maybe put this after `RETURNDATACOPY`?

cburgdorf: I was thinking back and forth for a minute when I wrote it but then settled with alphabetical order as that is the chosen order at most other places (e.g. in mnemonics). But now that I think about it again, would probably have made sense to use byte order here to stay consistent within this location.

Issue #1076, opened 2018-07-24
Formalize API and use common executor pool for `p2p.Service.run_in_executor`


pipermerriam:
### What is wrong?

The `_run_in_executor` API from the chain and state syncers needs to be formalized onto `BaseService`.  In addition, by defaults, all services should use a common executor pool

### How can it be fixed

Fill this in if you know how to fix it.

PR #1149, opened 2018-08-06
Expand round trip to GetNodeData and NodeData commands

pipermerriam:
### What was wrong?

The `GetnodeData` and `NodeData` commands can be done as a singe awaitable round trip call using the new round trip peer request API

### How was it fixed?

Setup a new request manager for `NodeData` requests/response and integrated it into the `StateSyncer`

#### Cute Animal Picture

wenn21648221_hejbmq

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -27,7 +27,7 @@ LOGGING_NAMESPACES = ('eth', 'p2p', 'trinity')
 
 
 @pytest.fixture(autouse=True, scope="session")
-def vm_logger(namespaces=LOGGING_NAMESPACES):
+def _stdout_logging(namespaces=LOGGING_NAMESPACES):
     for namespace in namespaces:
         logger = logging.getLogger(namespace)
 
@@ -43,14 +43,14 @@ def vm_logger(namespaces=LOGGING_NAMESPACES):
 
         logger.addHandler(handler)
 
-        return logger
+        logger.info('Set level for logger: %s', namespace)
 
 
 # Uncomment this to have logs from tests written to a file.  This is useful for
 # debugging when you need to dump the VM output from test runs.
 """
 @pytest.yield_fixture(autouse=True)
-def vm_file_logger(request):
+def _file_logging(request):
     import datetime
     import os
 
--- a/tests/trinity/core/p2p-proto/test_node_data_request_object.py
+++ b/tests/trinity/core/p2p-proto/test_node_data_request_object.py
@@ -1,0 +1,74 @@
+import os
+import random
+
+import pytest
+
+from eth_utils import keccak
+
+from p2p.exceptions import ValidationError
+
+from trinity.protocol.eth.requests import NodeDataRequest
+
+
+def mk_node():
+    node_length = random.randint(0, 2048)
+    node = os.urandom(node_length)
+    return node
+
+
+def mk_node_data(n):
+    if n == 0:
+        return tuple(), tuple()
+    nodes = tuple(set(mk_node() for _ in range(n)))
+    node_keys = tuple(keccak(node) for node in nodes)
+    return node_keys, nodes
+
+
+def test_node_data_request_empty_response_is_valid():
+    node_keys, _ = mk_node_data(10)
+    request = NodeDataRequest(node_keys)
+
+    request.validate_response(tuple())
+
+
+def test_node_data_request_with_full_response():
+    node_keys, nodes = mk_node_data(10)
+    request = NodeDataRequest(node_keys)
+    node_data = tuple(zip(node_keys, nodes))
+
+    request.validate_response(node_data)
+
+
+def test_node_data_request_with_partial_response():
+    node_keys, nodes = mk_node_data(10)
+    request = NodeDataRequest(node_keys)
+    node_data = tuple(zip(node_keys, nodes))
+
+    request.validate_response(node_data[3:])
+    request.validate_response(node_data[:3])
+    request.validate_response((node_data[1], node_data[8], node_data[4]))
+
+
+def test_node_data_request_with_fully_invalid_response():
+    node_keys, nodes = mk_node_data(10)
+    request = NodeDataRequest(node_keys)
+
+    # construct a unique set of other nodes
+    other_nodes = tuple(set(mk_node() for _ in range(10)).difference(nodes))
+    other_node_data = tuple((keccak(node), node) for node in other_nodes)
+
+    with pytest.raises(ValidationError):
+        request.validate_response(other_node_data)
+
+
+def test_node_data_request_with_extra_unrequested_nodes():
+    node_keys, nodes = mk_node_data(10)
+    request = NodeDataRequest(node_keys)
+    node_data = tuple(zip(node_keys, nodes))
+
+    # construct a unique set of other nodes
+    other_nodes = tuple(set(mk_node() for _ in range(10)).difference(nodes))
+    other_node_data = tuple((keccak(node), node) for node in other_nodes)
+
+    with pytest.raises(ValidationError):
+        request.validate_response(node_data + other_node_data)
--- a/trinity/protocol/common/managers.py
+++ b/trinity/protocol/common/managers.py
@@ -3,10 +3,10 @@ import asyncio
 from typing import (
     cast,
     Generic,
-    TypeVar,
     Set,
     Tuple,
     Type,
+    TypeVar,
 )
 
 from cancel_token import CancelToken
@@ -65,14 +65,14 @@ class BaseRequestManager(PeerSubscriber, BaseService, Generic[TPeer, TRequest, T
                     self.logger.error("Unexpected peer: %s  expected: %s", peer, self._peer)
                     continue
                 elif isinstance(cmd, self._response_msg_type):
-                    self._handle_msg(cast(TResponse, msg))
+                    await self._handle_msg(cast(TResponse, msg))
                 else:
                     self.logger.warning("Unexpected message type: %s", cmd.__class__.__name__)
 
     async def _cleanup(self) -> None:
         pass
 
-    def _handle_msg(self, msg: TResponse) -> None:
+    async def _handle_msg(self, msg: TResponse) -> None:
         if self.pending_request is None:
             self.logger.debug(
                 "Got unexpected %s message from %", self.response_msg_name, self._peer
@@ -82,20 +82,31 @@ class BaseRequestManager(PeerSubscriber, BaseService, Generic[TPeer, TRequest, T
         request, future = self.pending_request
 
         try:
-            request.validate_response(msg)
+            response = await self._normalize_response(msg)
         except ValidationError as err:
             self.logger.debug(
-                "Response validation failure for pending %s request from peer %s: %s",
+                "Malformed response for pending %s request from peer %s: %s",
+                self.response_msg_name,
+                self._peer,
+                err,
+            )
+            return
+
+        try:
+            request.validate_response(response)
+        except ValidationError as err:
+            self.logger.debug(
+                "Response validation failed for pending %s request from peer %s: %s",
                 self.response_msg_name,
                 self._peer,
                 err,
             )
         else:
-            future.set_result(self._normalize_response(msg))
+            future.set_result(response)
             self.pending_request = None
 
     @abstractmethod
-    def _normalize_response(self, msg: TResponse) -> TReturn:
+    async def _normalize_response(self, msg: TResponse) -> TReturn:
         pass
 
     @abstractmethod
--- a/trinity/protocol/common/managers.py
+++ b/trinity/protocol/common/managers.py
@@ -12,8 +12,10 @@ from typing import (
 from cancel_token import CancelToken
 
 from p2p.exceptions import (
+    MalformedMessage,
     ValidationError,
 )
+from p2p.p2p_proto import DisconnectReason
 from p2p.peer import BasePeer, PeerSubscriber
 from p2p.protocol import (
     Command,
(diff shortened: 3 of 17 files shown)


trinity/protocol/eth/requests.py
```diff
@@ -40,3 +48,42 @@ def validate_response(self, response: Any) -> None:
...
+
+        if len(actual_node_keys) != len(nodes):
+            # TODO: is this something that needs to be enforced.
+            raise ValidationError("Response may not contain duplicate nodes")
+
+        unexpected_keys = actual_node_keys.difference(expected_node_keys)
+
+        if unexpected_keys:
+            raise ValidationError(
+                "Response contains {0} unexpected nodes".format(len(unexpected_keys))
```
pipermerriam: @gsalgado Is this correct? Should we reject a response that contains nodes that we didn't request?

trinity/protocol/eth/requests.py
```diff
@@ -40,3 +48,42 @@ def validate_response(self, response: Any) -> None:
...
+        elif not all(isinstance(item, bytes) for item in response):
+            raise ValidationError("Response must be a tuple of `BlockHeader` objects")
+        return self.validate_nodes(cast(Tuple[bytes, ...], response))
+
+    def validate_nodes(self, nodes: Tuple[bytes, ...]) -> None:
+        if not nodes:
+            # an empty response is always valid
+            return
+
+        # TODO: move to executor since keccak is expensive
```
gsalgado: I think this is very likely to cause the main thread to block for too long, thus preventing the event loop from consuming received messages fast enough which in turn will lead to us dropping peers, as we've seen recently. We should either address this as part of this PR or immediately afterwards

gsalgado: Also, does this mean we're now hashing nodes twice? Once here and once in `_request_and_process_nodes()`? That'd be worrisome as this is indeed an expensive operation and we don't want to do it twice for every NodeData we receive

trinity/sync/full/state.py
```diff
@@ -253,11 +230,42 @@ def deregister_peer(self, peer: BasePeer) -> None:
...
+        # check for missing nodes and re-schedule them
+        missing = set(batch).difference(node_keys)
+
+        self._peer_missing_nodes[peer].update(missing)
+        if missing:
+            await self.request_nodes(missing)
+
+        # check for unexpected nodes and log if there are any
+        unexpected = set(node_keys).difference(batch)
+        if unexpected:
```
gsalgado: Can this actually happen or would a ValidationError be raised by the `get_node_data()` call above if a node sent us unexpected nodes?

trinity/sync/full/state.py
```diff
@@ -253,11 +230,42 @@ def deregister_peer(self, peer: BasePeer) -> None:
...
             self.request_tracker.active_requests[peer] = (time.time(), batch)
-            self.logger.debug("Requesting %d trie nodes to %s", len(batch), peer)
-            peer.sub_proto.send_get_node_data(batch)
+            asyncio.ensure_future(self._request_and_process_nodes(peer, batch))
+
+    async def _request_and_process_nodes(self, peer: ETHPeer, batch: Tuple[Hash32, ...]) -> None:
+        self.logger.debug("Requesting %d trie nodes from %s", len(batch), peer)
+        nodes = await peer.requests.get_node_data(batch)
+
+        self.logger.debug("Got %d NodeData entries from %s", len(nodes), peer)
```
gsalgado: Should we also include here how many nodes were originally requested? I know it's in the previous log but there will usually be lots of noise between them so it'd be good to have that info here as well.

trinity/protocol/eth/requests.py
gsalgado: I think so, but maybe log a warning just in case a bug in our code could be causing them

trinity/protocol/eth/requests.py
pipermerriam: > We should either address this as part of this PR or immediately afterwards

:+1: I'm going to prioritize https://github.com/ethereum/py-evm/issues/1076 today and try to get it in before merging this.

> does this mean we're now hashing nodes twice?

Yes, I'm still puzzling over what to do about this.  It's even worse in the `GetReceipts/Receipts` case because we have to generate the receipt tries to validate the response against the requested `header.receipt_root` properties.  Still not sure what I want to do about this but I know it isn't tenable to incur that overhead twice.

pipermerriam:
Depends on https://github.com/ethereum/py-evm/pull/1155 which should be merged before trying to review.

trinity/protocol/common/managers.py
```diff
@@ -82,7 +85,12 @@ def _handle_msg(self, msg: TResponse) -> None:
         request, future = self.pending_request
 
         try:
-            request.validate_response(msg)
+            response = await self._normalize_response(msg)
```
pipermerriam: @gsalgado this is the mechanism that now allows us to do avoid doing heavy computations on response data both during validation and processing.  Whatever *pre-processing* that *must* be done on response data in order to validate it is now required to be done before validation.  Each `Manager` class will then implement the pre-processing step via a `_normalize_response` method, and the data returned is in whatever format is deemed appropriate for the given response type.

Here, for `NodeData` responses, the returned format is a `Tuple[Tuple[Hash32, bytes], ...]`, aka a tuple of `(keccak(node), node)` pairs.  For `Receipts` I think this is going to be a 3-tuple of `[receipts, receipts_root, trie_kv_data]`, ie. the receipts paired up with the return of `mk_trie_nodes_and_data`.  Happy to hear feedback on this as it's the best I could come up with.

trinity/protocol/eth/managers.py
```diff
@@ -56,5 +68,41 @@ class GetBlockHeadersRequestManager(BaseRequestManager):
...
+    async def _normalize_response(self,
+                                  msg: Tuple[bytes, ...]
+                                  ) -> Tuple[Tuple[Hash32, bytes], ...]:
+        if not isinstance(msg, tuple):
+            raise ValidationError("Invalid msg, must be tuple of byte strings")
+        elif not all(isinstance(item, bytes) for item in msg):
+            raise ValidationError("Invalid msg, must be tuple of byte strings")
+
+        node_keys = await self._run_in_executor(tuple, map(keccak, msg))
+        return tuple(zip(node_keys, msg))
```
pipermerriam: @gsalgado here you can see the *normalization* of the `NodeData` response, pairing the nodes with their computed hashes.

pipermerriam:
@gsalgado this is ready for another round of review.  If https://github.com/ethereum/py-evm/pull/1155 has not yet been reviewed/merged, then you can just look at the latest commit as this is built on that PR.

trinity/protocol/eth/requests.py
```diff
@@ -30,13 +32,33 @@ def __init__(self,
...
         """
         Core `Request` API used for validation.
         """
-        if not isinstance(response, tuple):
-            raise ValidationError("Response to `HeaderRequest` must be a tuple")
-        elif not all(isinstance(item, BlockHeader) for item in response):
-            raise ValidationError("Response must be a tuple of `BlockHeader` objects")
+        return self.validate_node_data(response)
+
+    def validate_node_data(self, node_data: Tuple[Tuple[Hash32, bytes], ...]) -> None:
```
carver: Not really a problem, but why the extra layer of function call here? Seems reasonable to inline this in `validate_response`

trinity/protocol/les/requests.py
```diff
@@ -1,20 +1,21 @@
...
-
-from trinity.protocol.common.requests import BaseHeaderRequest
+from trinity.protocol.common.requests import (
+    BaseHeaderRequest,
+)
 
 from .constants import MAX_HEADERS_FETCH
 
 
+BlockHeaders_R = Tuple[BlockHeader, ...]
```
carver: How is this used?

trinity/sync/full/state.py
```diff
@@ -245,11 +221,42 @@ def deregister_peer(self, peer: BasePeer) -> None:
...
             not_yet_requested = not_yet_requested.difference(batch)
             self.request_tracker.active_requests[peer] = (time.time(), batch)
-            self.logger.debug("Requesting %d trie nodes to %s", len(batch), peer)
-            peer.sub_proto.send_get_node_data(batch)
+            asyncio.ensure_future(self._request_and_process_nodes(peer, batch))
+
+    async def _request_and_process_nodes(self, peer: ETHPeer, batch: Tuple[Hash32, ...]) -> None:
+        self.logger.debug("Requesting %d trie nodes from %s", len(batch), peer)
+        node_data = await peer.requests.get_node_data(batch)
+        self.request_tracker.active_requests.pop(peer)
```
carver: How do we know there's no `KeyError` here?
Why store the `time.time()` just to drop it?

trinity/sync/full/state.py
```diff
@@ -245,11 +221,42 @@ def deregister_peer(self, peer: BasePeer) -> None:
...
+        else:
+            node_keys = tuple()
+
+        # check for missing nodes and re-schedule them
+        missing = set(batch).difference(node_keys)
+
+        # TODO: this doesn't actually mean the peer doesn't have them, just
+        # that they didn't respond with them this time.  We should explore
+        # alternate ways to do this since a false negative here will result in
+        # not requesting this node from this peer again.
```
carver: Maybe if we have marked the node as missing from *all* connected peers, then we drop the missing node from all peers and try again?

trinity/protocol/common/managers.py
```diff
@@ -82,20 +82,31 @@ def _handle_msg(self, msg: TResponse) -> None:
...
+        except ValidationError as err:
+            self.logger.debug(
+                "Malformed response for pending %s request from peer %s: %s",
+                self.response_msg_name,
+                self._peer,
+                err,
+            )
```
carver: Can/should we tread this peer as malicious/broken and drop them?

tests/trinity/core/p2p-proto/test_node_data_request_object.py
```diff
@@ -0,0 +1,74 @@
...
+def test_node_data_request_with_extra_unrequested_nodes():
+    node_keys, nodes = mk_node_data(10)
+    request = NodeDataRequest(node_keys)
+    node_data = tuple(zip(node_keys, nodes))
+
+    # construct a unique set of other nodes
+    other_nodes = tuple(set(mk_node() for _ in range(10)).difference(nodes))
+    other_node_data = tuple((keccak(node), node) for node in other_nodes)
+
+    with pytest.raises(ValidationError):
```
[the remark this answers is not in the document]
pipermerriam: I think `ValidationError` is the semantically correct one here.

- `MalformedMessage`: The data payload sent is not well formed and thus cannot be decoded/parsed/etc.
- `ValidationError`: The data is well formed, but the actual content of it does not pass some validation check (like that they sent us the stuff we asked for).

Also, this exception doesn't escape the Manager's internal message loop and is used a signal from `Request.validate_response` to indicate that the response does not validate and thus should not fulfil the request.

That does however point out that the `_normalize_response`  API should probably be raising `MalformedMessage` exceptions since the checks it performs are about whether the data is well formed.  Will fix.

trinity/protocol/eth/requests.py
pipermerriam: I think is a left-over from some type checking hoop jumping which has now been removed/resolved.  Will remove the extra function call layer.

trinity/protocol/les/requests.py
pipermerriam: Looks like another leftover from type hint hoop jumping.  Will remove.

trinity/sync/full/state.py
pipermerriam: > How do we know there's no KeyError here?

We don't.  I'll add some exception handling here for the interim until this API gets removed entirely.  My plan is to do another round of refactoring on the `StateDownloader` to remove this `request_tracker` API because I don't believe it is needed anymore.

trinity/sync/full/state.py
pipermerriam: That could at least be a decent short term solution to the larger problem which is that we need to have state syncing continually update to newer state roots, otherwise it's basically impossible to finish because the nodes you are connected to are pruning old state history and in my experiments, you always end up with a few missing trie nodes, even if you have 99% of the state db already downloaded because of the time it takes to process the state trie and find the few nodes you are missing.

tests/conftest.py
```diff
@@ -27,7 +27,7 @@
 
 
 @pytest.fixture(autouse=True, scope="session")
-def vm_logger(namespaces=LOGGING_NAMESPACES):
+def _stdout_logging(namespaces=LOGGING_NAMESPACES):
```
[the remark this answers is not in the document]
pipermerriam: For trinity proper, yes, it's `stderr` but this one specifically is `stdout`.

trinity/protocol/eth/managers.py
```diff
@@ -56,5 +68,41 @@ class GetBlockHeadersRequestManager(BaseRequestManager):
     def _send_sub_proto_request(self, request: HeaderRequest) -> None:
         self._peer.sub_proto.send_get_block_headers(request)
 
-    def _normalize_response(self, response: Tuple[BlockHeader, ...]) -> Tuple[BlockHeader, ...]:
-        return response
+    async def _normalize_response(self,
+                                  msg: Tuple[BlockHeader, ...]
+                                  ) -> Tuple[BlockHeader, ...]:
```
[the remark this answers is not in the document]
pipermerriam: I find that style harder to read than this one.  Thoughts?

pipermerriam:
> I feel like I have to hop back and forth between several different files to figure out what's going on. That's not too bad in a PR, but not great when just reading the final code. I don't think it needs to block this PR, but hopefully we can address it in a follow-up refactor.

Agreed. There's some deficiencies in how this is structured that I'm still hoping will shake out into something that's easier to understand and work with.

trinity/protocol/common/managers.py
gsalgado: That sounds good to me!

trinity/protocol/eth/managers.py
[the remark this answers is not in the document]
carver: Yeah, it comes up a *lot* more with all the type boilerplate.

My biggest issue is that it means the intend is tied to the length of the method name, so changing the method name:
- triggers a much larger diff than necessary
- takes more work to re-indent the args, depending on your local IDE setup

Also, for long method names, it is awkward to read variables dangling way off to the right. It might not even avoid the line width limit that well.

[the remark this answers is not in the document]
cburgdorf: I personally like the `self` staying on the top line but then again I think we may soon migrate to something like black anyway ;)

PR #1150, opened 2018-08-06
Handle decryption error

pipermerriam:
### What was wrong?

`Peer.read_msg` can throw a decryption error.

```
  File "/home/piper/projects/py-evm/p2p/service.py", line 76, in run
    await self._run()
  File "/home/piper/projects/py-evm/p2p/peer.py", line 334, in _run
  File "/home/piper/projects/py-evm/p2p/peer.py", line 348, in read_msg
    header = self.decrypt_header(header_data)
  File "/home/piper/projects/py-evm/p2p/peer.py", line 460, in decrypt_header
    expected_header_mac, header_mac))
```

### How was it fixed?

Handle it in the main `_run` loop.

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -332,9 +332,16 @@ class BasePeer(BaseService):
         while True:
             try:
                 cmd, msg = await self.read_msg()
-            except (PeerConnectionLost, TimeoutError) as e:
+            except (PeerConnectionLost, TimeoutError) as err:
                 self.logger.debug(
-                    "%s stopped responding (%s), disconnecting", self.remote, repr(e))
+                    "%s stopped responding (%r), disconnecting", self.remote, err)
+                return
+            except DecryptionError as err:
+                self.logger.warn(
+                    "Unable to decrypt message from %s, disconnecting: %r",
+                    self, err,
+                    exc_info=True,
+                )
                 return
 
             try:


p2p/peer.py
```diff
@@ -332,9 +332,15 @@ def is_closing(self) -> bool:
...
                 cmd, msg = await self.read_msg()
-            except (PeerConnectionLost, TimeoutError) as e:
+            except (PeerConnectionLost, TimeoutError) as err:
                 self.logger.debug(
-                    "%s stopped responding (%s), disconnecting", self.remote, repr(e))
+                    "%s stopped responding (%r), disconnecting", self.remote, err)
+                return
+            except DecryptionError as err:
+                self.logger.debug(
+                    "Unable to decrypt message from %s, disconnecting: %r",
```
carver: Hm, any way this could be a spoofed message from a different peer? Just thinking that if a malicious peer could spoof messages from other peers (even with broken encryption), and cause us to drop all other peers.

pipermerriam: Possibly, I was planning on maybe bumping this to `WARN` level for a while to see if it shows up again so we don't loose it in our debug logs.

gsalgado:
> Is raising an exception all that needs to be done to cleanly close the connection?

Once the `_run()` method returns, `BaseService` will call `BasePeer._cleanup()` and that will call `.close()`

p2p/peer.py
pipermerriam: Upped it to `WARN` level and included a stack-trace in the logging output for good measure.

Issue #772, opened 2018-05-24
Changes to `Chain.import_block`


pipermerriam:
Blocks: https://github.com/ethereum/py-evm/issues/733

### What is wrong?

Currently `chain.import_block(..)` returns the block that was imported.  Sometimes, this will cause a chain re-org, but this API doesn't expose sufficient information to determine this.

### How can it be fixed

Change `Chain.import_block` to return two iterables:

- blocks which are *new* to the canonical chain.
- blocks which were *previously* part of the canonical chain (but no longer are).

With this, callers can inspect the return value and determine that a re-org happened when more than a single block is returned.

This will also require changing `ChainDB.persist_block` to return the same data format, an iterable of blocks which are part of the new canonical chain.

carver:
With HeaderDB, we ended up also returning a list of dropped headers on re-org, right? That might make sense here, too.

voith:
registering this for myself.

voith:
did not get time to work on this. If this is blocking anyone, I have no problem with anyone else picking this up!

glaksmono:
Already confirmed with @voith that I will take on this task 🍴

glaksmono:
Haven't had a chance working on this. Would be ok if anyone could take on it

carver:
I expect this to be a common question, so wanted to add the conversation from chat:
>  How can there be multiple blocks that are new? Because ChainDB.import_block is importing 1 block at a time

It's returning the blocks that are new to the canonical chain. Let's say you have chain:

```
A -- B - C
  `- D
```

where C is the current canonical head.

Then you import E with parent D, where the total difficulty of E is higher than C. Now, you have added D and E to the canonical chain, and removed B and C.

voith:
Is this issue still valid? It seems like logging when a re-org happens has been added in #974. So I assume that whatever logic was needed to resolve the blocker has been addressed.

carver:
> Is this issue still valid?

Yes, `import_block()` still only returns one block.

carver:
Closed by #1151

PR #1151, opened 2018-08-06
WIP changes chain.import_block to return new blocks and blocks evicte…

eddybenkhin:
…d issue #772

### What was wrong?
evicted blocks were not being returned 

### How was it fixed?
created a list that holds old blocks inside _set_as_canonical_chain_head() and returns both new and evicted blocks. 
one thing that was unclear was why the tests required the imported block even if it wasn't added to the chain, so I added a bool import_success to track this, but not sure if this is best practice.   

#### Cute Animal Picture

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -255,7 +255,7 @@ class BaseChain(Configurable, ABC):
     def import_block(self,
                      block: BaseBlock,
                      perform_validation: bool=True,
-                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...], bool]:
+                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...]]:
         raise NotImplementedError("Chain classes must implement this method")
 
     #
@@ -560,9 +560,9 @@ class Chain(BaseChain):
     def import_block(self,
                      block: BaseBlock,
                      perform_validation: bool=True
-                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...], bool]:
+                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...]]:
         """
-        Imports a complete block.
+        Imports a complete block and returns new_blocks and orphaned_canonical_blocks.
         """
 
         try:
@@ -585,23 +585,26 @@ class Chain(BaseChain):
             ensure_imported_block_unchanged(imported_block, block)
             self.validate_block(imported_block)
 
-        new_canonical_headers, old_headers = self.chaindb.persist_block(imported_block)
+        (
+            new_canonical_header_hashes,
+            orphaned_header_hashes,
+        ) = self.chaindb.persist_block(imported_block)
+
         self.logger.debug(
             'IMPORTED_BLOCK: number %s | hash %s',
             imported_block.number,
             encode_hex(imported_block.hash),
         )
 
-        new_canonical_blocks = [
-            self.get_block_by_header(header) for header in new_canonical_headers]
-        old_canonical_blocks = [
-            self.get_block_by_header(header) for header in old_headers]
+        new_blocks = tuple(
+            self.get_block_by_hash(header_hash) for header_hash in new_canonical_header_hashes)
+        orphaned_canonical_blocks = tuple(
+            self.get_block_by_hash(header_hash) for header_hash in orphaned_header_hashes)
+        # add imported block to new block when no chain re-org
+        if not new_blocks:
+            new_blocks = (imported_block, )
 
-        import_success = True
-        if not new_canonical_blocks:
-            new_canonical_blocks.append(imported_block)
-            import_success = False
-        return tuple(new_canonical_blocks), tuple(old_canonical_blocks), import_success
+        return new_blocks, orphaned_canonical_blocks
 
     #
     # Validation API
@@ -770,12 +773,12 @@ class MiningChain(Chain):
     def import_block(self,
                      block: BaseBlock,
                      perform_validation: bool=True
-                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...], bool]:
-        new_canonical_blocks, old_canonical_blocks, import_success = super().import_block(
+                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...]]:
+        new_canonical_blocks, orphaned_canonical_blocks = super().import_block(
             block, perform_validation)
 
         self.header = self.ensure_header()
-        return new_canonical_blocks, old_canonical_blocks, import_success
+        return new_canonical_blocks, orphaned_canonical_blocks
 
     def mine_block(self, *args: Any, **kwargs: Any) -> BaseBlock:
         """
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -8,12 +8,13 @@ import operator
 import random
 from typing import (  # noqa: F401
     Any,
-    Optional,
     Callable,
     cast,
     Dict,
     Generator,
     Iterator,
+    List,
+    Optional,
     Tuple,
     Type,
     TYPE_CHECKING,
@@ -251,7 +252,10 @@ class BaseChain(Configurable, ABC):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseBlock:
+    def import_block(self,
+                     block: BaseBlock,
+                     perform_validation: bool=True,
+                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...], bool]:
         raise NotImplementedError("Chain classes must implement this method")
 
     #
@@ -553,10 +557,14 @@ class Chain(BaseChain):
         with self.get_vm(at_header).state_in_temp_block() as state:
             return self.gas_estimator(state, transaction)
 
-    def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseBlock:
+    def import_block(self,
+                     block: BaseBlock,
+                     perform_validation: bool=True
+                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...], bool]:
         """
         Imports a complete block.
         """
+
         try:
             parent_header = self.get_block_header_by_hash(block.header.parent_hash)
         except HeaderNotFound:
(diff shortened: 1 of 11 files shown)


eddybenkhin:
Hello. I tried to resolve this issue to the best of my abilities but I'm not sure if it was done exactly as expected. Looking for feedback when possible. Thanks!

pipermerriam:
taking a look now, sorry I missed this when first submitted.

eth/chains/base.py
```diff
@@ -577,13 +585,23 @@ def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseB
...
+        new_canonical_blocks = [
+            self.get_block_by_header(header) for header in new_canonical_headers]
+        old_canonical_blocks = [
+            self.get_block_by_header(header) for header in old_headers]
+
+        import_success = True
+        if not new_canonical_blocks:
+            new_canonical_blocks.append(imported_block)
+            import_success = False
+        return tuple(new_canonical_blocks), tuple(old_canonical_blocks), import_success
```
pipermerriam: I think I'd like to drop the `import_success` since it can be derived from the returned data.

eth/chains/base.py
```diff
@@ -577,13 +585,23 @@ def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseB
...
             'IMPORTED_BLOCK: number %s | hash %s',
             imported_block.number,
             encode_hex(imported_block.hash),
         )
-        return imported_block
+
+        new_canonical_blocks = [
+            self.get_block_by_header(header) for header in new_canonical_headers]
+        old_canonical_blocks = [
+            self.get_block_by_header(header) for header in old_headers]
```
pipermerriam: You can drop the `tuple` casting for the return statement by just embedding the comprehension here in a `tuple(...)`

```python
squares = tuple(i ** 2 for i in range(10))
```

eth/chains/base.py
```diff
@@ -577,13 +585,23 @@ def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseB
             ensure_imported_block_unchanged(imported_block, block)
             self.validate_block(imported_block)
 
-        self.chaindb.persist_block(imported_block)
+        new_canonical_headers, old_headers = self.chaindb.persist_block(imported_block)
```
pipermerriam: Can we use the suffix `_hashes` on these as to not accidentally confuse them to be `BlockHeader` objects.

- `new_canonical_header_hashes`
- `old_header_hashes`

Also, maybe instead of the term `old` we could use one of

- `orphaned`
- `abandoned`
- `discarded`

cc @carver as he normally has thoughts about the naming of things.

eth/db/chain.py
```diff
@@ -245,6 +252,7 @@ def _set_as_canonical_chain_head(self, header: BlockHeader) -> Tuple[BlockHeader
                 break
             else:
                 old_header = self.get_block_header_by_hash(old_hash)
+                old_headers.append(old_header)
                 for transaction_hash in self.get_block_transaction_hashes(old_header):
                     self._remove_transaction_from_canonical_chain(transaction_hash)
                     # TODO re-add txn to internal pending pool (only if local sender)
```
pipermerriam: When you update this branch would you mind removing this `TODO` comment as well as the following `pass` statement which is no longer needed.

pipermerriam:
Good looking pull request.  It would be nice if we could eliminate the various places where lists are being mutated without sacrificing readability.  Otherwise, with some test cases to cover the functionality this will be good to merge.

eddybenkhin:
Thanks for the feedback. I'll get to work on these changes, but could use some guidance or examples on creating the test for this.

pipermerriam:
Test is likely going to require a bit of discovery work to figure 1-2 things out.  Here is the basic structure.

- Mine a chain of a few blocks (2-3) using an in-memory database.  This is chain `A`
- Make a full copy of the database used in chain `A`, and create a separate chain instance with this db.  This is chain `B`
- Mine 3 more blocks on chain `A`
- Mine 5 more blocks on chain `B`.  Make sure the first block you mine is different from the first block that was just mined on chain `A`.  This can be done by including a transaction in the block.

Now we have what we need to test our API.

- extract the blocks from chain `B` that you mined which are not in chain `A`.
- In chain `A` use the `import_block` API to import the blocks one-by-one.

You should be able to then make the necessary assertions on the return values of the `import_block` calls.

eddybenkhin:
and where should all of this logic live file structure wise?

pipermerriam:
New test file here:

`tests/core/chain-objects/test_import_block.py`

eth/chains/base.py
carver: :+1: to hashes. All of the `old` alternatives are pretty good. I think `orphaned` is my favorite: it's used pretty regularly for exactly this type of header/block: no longer on the longest/main chain.

eth/db/chain.py
carver: I guess the TxPool will have to track this information and handle it internally. Maybe an event should get broadcast to plugins with the new and orphaned header hashes.

pipermerriam: :+1: to broadcasting across the plugin event bus.

eddybenkhin:
started putting together the tests following the structure given in the earlier comment. ran into an issue where chain b block 4 isn't getting imported to chain a because the score is too low and i'm not sure how the score exactly works, any suggestions or thoughts on how to proceed?

carver:
Yeah, it makes sense that the 4th block on B will have lower total difficulty than the 6th block on A. Just keep importing the next blocks, and the total difficulty on fork B should eventually be higher than an A, and they'll all get imported to the canonical chain.

pipermerriam:
@eddybenkhin to expand on what @carver said.

When we import a block which **does not** cause a re-org, i.e. it is part of a different fork and the block being imported doesn't score higher than the current `HEAD` block, we want all things in the import to succeed **except** for it should be skipping/bypassing any logic that sets things as *canonical*.  Then, at the point where you import the block which causes the minority fork to score higher than the current head, all of the setting of things as canonical should happen then.  It would be expected behavior for the import to result in the block **not** being part of the chain, however, if you are getting an error during import then we may need to fix something.

eddybenkhin:
Okay thank you! I actually just figured this out, am performing some cleanup and will be committing shortly :)

eth/db/chain.py
eddybenkhin: not quite sure how to do this but would be willing to try given a bit more information

eddybenkhin:
I think I finished up all that this issue asked for but any feedback on how to improve is much appreciated.

eth/chains/base.py
```diff
@@ -553,10 +557,14 @@ def estimate_gas(self, transaction: BaseTransaction, at_header: BlockHeader=None
         with self.get_vm(at_header).state_in_temp_block() as state:
             return self.gas_estimator(state, transaction)
 
-    def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseBlock:
+    def import_block(self,
+                     block: BaseBlock,
+                     perform_validation: bool=True
+                     ) -> Tuple[Tuple[BaseBlock, ...], Tuple[BaseBlock, ...]]:
         """
         Imports a complete block.
```
pipermerriam: When you come back in here will you update this docstring to indicate what the two return values represent.

eth/chains/base.py
```diff
@@ -577,13 +585,24 @@ def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseB
             ensure_imported_block_unchanged(imported_block, block)
             self.validate_block(imported_block)
 
-        self.chaindb.persist_block(imported_block)
+        new_canonical_header_hashes,\
```
pipermerriam: *formatting nitpick*  we don't have any of these style line continuations and I'd like to avoid adding any.  Can you update this to:

```python
(
    new_canonical_header_hashes,
    orphaned_header_hashes,
) = self.chaindb.persist_block(imported_block)
```

eth/chains/base.py
```diff
@@ -577,13 +585,24 @@ def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseB
...
+            orphaned_header_hashes = self.chaindb.persist_block(imported_block)
+
         self.logger.debug(
             'IMPORTED_BLOCK: number %s | hash %s',
             imported_block.number,
             encode_hex(imported_block.hash),
         )
-        return imported_block
+
+        new_blocks = tuple([
```
pipermerriam: The inner list comprehension isn't needed (you can remove the brackets `[`, `]`).

eth/chains/base.py
```diff
@@ -577,13 +585,24 @@ def import_block(self, block: BaseBlock, perform_validation: bool=True) -> BaseB
...
+            self.get_block_by_hash(header_hash) for header_hash in new_canonical_header_hashes])
+        orphaned_canonical_blocks = tuple([
```
pipermerriam: same here, remove inner list brackets.

eth/db/chain.py
```diff
@@ -186,7 +188,9 @@ def get_block_uncles(self, uncles_hash: Hash32) -> List[BlockHeader]:
 
     # TODO: This method should take a chain of headers as that's the most common use case
     # and it'd be much faster than inserting each header individually.
```
pipermerriam: cc @gsalgado just remembering we have this pending optimization that we can still do.

eth/db/chain.py
```diff
@@ -214,17 +218,22 @@ def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
         try:
             head_score = self.get_score(self.get_canonical_head().hash)
         except CanonicalHeadNotFound:
-            new_headers = self._set_as_canonical_chain_head(header)
+            new_canonical_headers,\
```
pipermerriam: same style here to remove the `\` line continuation

eth/db/chain.py
```diff
@@ -214,17 +218,22 @@ def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
...
+                orphaned_canonical_headers = self._set_as_canonical_chain_head(header)
         else:
             if score > head_score:
-                new_headers = self._set_as_canonical_chain_head(header)
+                new_canonical_headers,\
```
pipermerriam: same style here to remove the `\` line continuation

eth/db/chain.py
```diff
@@ -288,6 +298,11 @@ def persist_block(self, block: 'BaseBlock') -> None:
             raise ValidationError(
                 "Block's uncles_hash (%s) does not match actual uncles' hash (%s)",
                 block.header.uncles_hash, uncles_hash)
+        new_canonical_header_hashes = tuple([header.hash for header in new_canonical_headers])
+        orphaned_canonical_header_hashes = tuple(
+            [header.hash for header in orphaned_canonical_headers])
```
pipermerriam: inner list brackets should be removed

eth/db/chain.py
```diff
@@ -288,6 +298,11 @@ def persist_block(self, block: 'BaseBlock') -> None:
             raise ValidationError(
                 "Block's uncles_hash (%s) does not match actual uncles' hash (%s)",
                 block.header.uncles_hash, uncles_hash)
+        new_canonical_header_hashes = tuple([header.hash for header in new_canonical_headers])
```
pipermerriam: inner list brackets should be removed

eth/db/header.py
```diff
@@ -198,13 +204,24 @@ def _set_as_canonical_chain_head(self, block_hash: Hash32) -> Tuple[BlockHeader,
...
+        old_headers = []
+
+        for h in new_canonical_headers:
+            try:
+                old_hash = self.get_canonical_block_hash(h.block_number)
+            except HeaderNotFound:
+                # no old block, and no more possible
+                break
+            else:
+                old_header = self.get_block_header_by_hash(old_hash)
```
pipermerriam: Can we update and use the same naming convention `orphaned_` in this function body?

eth/db/chain.py
cburgdorf: Jumped on it https://github.com/ethereum/py-evm/pull/1192

eddybenkhin:
Made the cleanups requested. Any suggestions for a good next issue to work on or any help needed with the TODO's?

pipermerriam:
Very closely related to this one: https://github.com/ethereum/py-evm/issues/1185  

If you want to give that a stab.  I *think* it will be reasonably straight forward but we'll have to see once you get into implementation.

PR #1155, opened 2018-08-07
Implement ServiceContext API for shared resources across services

pipermerriam:
unlikely-sleeping-buddies-animal-friendship-301__880
### What was wrong?

fixes https://github.com/ethereum/py-evm/issues/1076

In the *chain* and *state* syncers there is an API for running computationally heavy workloads in a process pool of workers.  The worker pool is automatically sized according to the number of cpus available on the machine.  This pattern is quite useful, however, if used concurrently by two services, it would result in multiple pools being created, and would likely overwhelm the CPU load on the machine.

### How was it fixed?

Implemented a `ServiceContext` class which is a required argument for instantiating any `BaseService` subclass.  This class currenly **only** has the executor on it, however, things like the chain config look like good candidates for this API.

#### Cute Animal Picture


carver:
My first reaction is that it seems like a lot of overhead to add the init argument. Especially when every invocation in the repo is `ServiceContext()`. At first glance, the potential future change we're talking about (being sure that we don't launch too many processes in the process pool) seems like it will have to be a global thing, like a single shared `ProcessPoolExecutor`. So I don't see the benefit of each service having its own instance here.

How about: dropping `_run_in_executor()` and the `ServiceContext` argument from `BaseService`, and adding a utils method:

```py
async def parallel_execute(service, func, *args, executor=None):
  if executor is None:
    executor = get_asyncio_executor()  # should that result be a global/singleton?

  loop = service.get_event_loop()
  return await service.wait(loop.run_in_executor(executor, func, *args)
```

Then any `_run_in_executor` call would be replaced with:

```diff
- self._run_in_executor(callable, arg)
+ parallel_execute(self, callable, arg)
```

p2p/service.py
```diff
@@ -102,6 +138,10 @@ def run_child_service(self, child_service: 'BaseService') -> 'asyncio.Future[Any
         self._child_services.append(child_service)
         return asyncio.ensure_future(child_service.run())
 
+    async def _run_in_executor(self, callback: Callable[..., Any], *args: Any) -> Any:
+        loop = self.get_event_loop()
+        return await self.wait(loop.run_in_executor(self.context._executor, callback, *args))
```
carver: `context.executor`?

pipermerriam:
After thinking about this a bit I think you're right (that it's a bit much just to add this API).  I'm going to take another stab at this tomorrow and see if I can come up with something minimal that gets the common executor pool in place without being as heavy handed.

PR #1156, opened 2018-08-07
Reduce logging about no idle peers

### LINKED ISSUE #1133: DEBUG logs are dominated by "No idle peers have any of the trie nodes we want"
### What is wrong?

Currently, the debug logs are dominated by this logging statement during state sync.

https://github.com/ethereum/py-evm/blob/2f8f89a6abfec1757ab2090c98e4c1df03f2b3f1/p2p/state.py#L219-L221

### How can it be fixed

We either need to move that to `TRACE` level, increase the sleep duration (probably bad for performance), or ideally, find a clean way to throttle that message so that it only shows up at most once every 1-2 seconts, or something else I haven't thought of.

pipermerriam:
### What was wrong?

Fixes #1133 

### How was it fixed?

Reduced the logging level to `TRACE`.  After running stat sync continuously for days now, these messages provide no useful information in the debug logs and merely serve as noise.

#### Cute Animal Picture

unlikely-sleeping-buddies-animal-friendship-fb

### DIFF
--- a/trinity/sync/full/state.py
+++ b/trinity/sync/full/state.py
@@ -232,7 +232,7 @@ class StateDownloader(BaseService, PeerSubscriber):
             try:
                 peer = await self.get_peer_for_request(not_yet_requested)
             except NoIdlePeers:
-                self.logger.debug(
+                self.logger.trace(
                     "No idle peers have any of the %d trie nodes we want, sleeping a bit",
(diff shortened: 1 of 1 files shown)


pipermerriam:
@gsalgado given the simplicity of this, I'm merging as-is and will address any feedback you have post-merge.

PR #1157, opened 2018-08-08
Use shared ProcessPoolExecutor across all services

pipermerriam:
revamp of #1155 

### What was wrong?

fixes https://github.com/ethereum/py-evm/issues/1076

In the *chain* and *state* syncers there is an API for running computationally heavy workloads in a process pool of workers.  The worker pool is automatically sized according to the number of cpus available on the machine.  This pattern is quite useful, however, if used concurrently by two services, it would result in multiple pools being created, and would likely overwhelm the CPU load on the machine.

### How was it fixed?

The `_run_in_executor` API is now available for all `BaseService` subclasses.  This implements a quick-and-easy solution using a global executor pool which will be re-used across all `BaseService` subclasses within a single process.

This PR also contains additional cleanup:

- The `BasePeer` class is now instantiated with the cancel token from the `PeerPool`.
- More consistent usage of the `BaseService.loop` in the various asyncio APIs.
- Minor linting cleanup

#### Cute Animal Picture

unlikely-sleeping-buddies-animal-friendship-301__880


tests/trinity/core/p2p-proto/test_state_sync.py
```diff
@@ -47,7 +47,10 @@ def make_random_trie(random):
 
 
 @given(random=strategies.randoms())
-@settings(max_examples=10)
+@settings(
+    max_examples=10,
+    deadline=4000,
```
carver: Interesting, was this taking a really long time with just a few examples?

pipermerriam: I think it's because it runs 10 examples, each which populate a trie with 1000 nodes, which as we know is a computationally expensive operation.

PR #1158, opened 2018-08-08
Allow granular configuration of logging

### LINKED ISSUE #806: Trinity should support logger-specific levels
The log level specified via the command line is used for all loggers. It should be possible to specify logger names/levels pairs as well. Something like `-l p2p.chain:debug,p2p.peer.Peer:warn`

pipermerriam:
### What was wrong?

Fixes #806 

### How was it fixed?

You can now do all of the following.

- `trinity --stderr-log-level` -> configures the level at which things will get printed to stderr
- `trinity --file-log-level` -> configures the level at which things will get printed to the debug file based logs
- `trinity --log-level INFO` -> sets the logging level on the root logger.
- `trinity --log-level trinity.sync=INFO --log-level p2p.kademlia=TRACE` -> configures the individual loggers.

#### Cute Animal Picture

xmtvj0d_syex


trinity/cli_parser.py
```diff
@@ -23,28 +25,107 @@ class ValidateAndStoreEnodes(argparse.Action):
...
+    '50': logging.CRITICAL,
+    # lowercase versio,
+    'trace': TRACE_LEVEL_NUM,
+    'debug': logging.CRITICAL,
+    'info': logging.INFO,
+    'warn': logging.WARNING,
+    'warning': logging.WARNING,
+    'error': logging.ERROR,
+    'critical': logging.CRITICAL,
+    # capcase versio,
```
cburgdorf: Instead of maintaining upper and lower case versions I think it would be better to simply decide on one casing and then do `value.lower()` or `value.upper()` before the lookup.

trinity/cli_parser.py
```diff
@@ -23,28 +25,107 @@ class ValidateAndStoreEnodes(argparse.Action):
...
+    def __call__(self,
+                 parser: argparse.ArgumentParser,
+                 namespace: argparse.Namespace,
+                 value: Any,
+                 option_string: str=None) -> None:
+        if value is None:
+            return
+
+        # this is a global log level.
+        if value in LOG_LEVEL_CHOICES:
```
cburgdorf: So, this should become `value.upper()`or `value.lower()` imho

PR #1159, opened 2018-08-08
Expand round trip api to ETH-protocol receipts

pipermerriam:
Builds on https://github.com/ethereum/py-evm/pull/1149

### What was wrong?

The round trip API for peer request/responses had not been implemented for block receipts.

### How was it fixed?

Expanded the API to cover receipts and integrated the new api into the `FastChainSyncer`.

#### Cute Animal Picture

cow-attacks-and-dogs-770x405

### DIFF
--- a/trinity/protocol/eth/proto.py
+++ b/trinity/protocol/eth/proto.py
@@ -1,8 +1,10 @@
 import logging
 from typing import (
+    cast,
     List,
     Tuple,
     TYPE_CHECKING,
+    Union,
 )
 
 from eth_typing import Hash32
@@ -35,6 +37,7 @@ from . import constants
 from .requests import (
     HeaderRequest,
     NodeDataRequest,
+    ReceiptsRequest,
 )
 
 if TYPE_CHECKING:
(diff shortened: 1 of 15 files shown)


trinity/protocol/eth/proto.py
```diff
@@ -115,7 +122,12 @@ def send_block_bodies(self, blocks: List[BlockBody]) -> None:
         header, body = cmd.encode(blocks)
         self.send(header, body)
 
-    def send_get_receipts(self, block_hashes: List[Hash32]) -> None:
+    def send_get_receipts(self,
+                          request: Union[ReceiptsRequest, Tuple[Hash32, ...]]) -> None:
```
carver: Is this type Union (and the `send_get_node_data` one) just to make testing easier? If so, it might be nice to simplify this to only `ReceiptsRequest` and do a bit more setup work in the tests. (also get to drop the `isinstance` which is occasionally slow, in tight loops)

trinity/sync/full/chain.py
```diff
@@ -155,6 +144,87 @@ def __init__(self,
...
+
+            # issue requests to all of the peers and wait for all of them to respond.
+            requests = tuple(
+                self._get_receipts(peer, batch)
+                for peer, batch
+                in zip(peers, batches)
+            )
+            responses = await self.wait(asyncio.gather(
+                *requests,
+                loop=self.get_event_loop(),
```
carver: I guess if we start seeing this a lot, we can add a `BaseService.gather()` to turn this into:

```py
responses = await self.gather(*requests)
```

trinity/sync/full/chain.py
```diff
@@ -155,6 +144,87 @@ def __init__(self,
...
+            headers = tuple(concat(all_missing_headers))
+
+            # process all of the returned receipts, storing their trie data
+            # dicts in the database
+            receipts, trie_roots_and_data_dicts = zip(*receipt_bundles)
+            trie_roots, trie_data_dicts = zip(*trie_roots_and_data_dicts)
+            for trie_data in trie_data_dicts:
+                await self.wait(self.db.coro_persist_trie_data_dict(trie_data))
+        else:
+            self.logger.debug("Got receipts batch for %d headers", len(all_headers))
```
carver: This doesn't need to be in an `else` block, right? I did a double-take looking for a `break` statement.

pipermerriam:
Closing in favor of doing all the PR cleanup in https://github.com/ethereum/py-evm/pull/1169 which includes this PR.

Issue #1154, opened 2018-08-07
HexaryTrieSync._existing_nodes may use a ton of memory


gsalgado:
That is a cache we use to speed things up, but it has no limit on how much it can grow, and for a full state sync that may end up consuming a lot of memory. See https://github.com/ethereum/py-evm/pull/1124#discussion_r208231117 for more info

gsalgado:
The reason why we need this optimization is because existence checks on a large (e.g. 30GB) LevelDB is very expensive, so maybe we could use a separate LevelDB where we'd keep just the hash of nodes known to be present in the chain DB. This separate LevelDB would be created when the StateDownloader starts, passed on to HexaryTrieSync, and destroyed when StateDownloader exits

PR #1161, opened 2018-08-09
Use a LevelDB instance as the node cache in StateDownloader

gsalgado:
Closes: #1154

### DIFF
--- a/trinity/sync/full/state.py
+++ b/trinity/sync/full/state.py
@@ -264,7 +264,7 @@ class StateDownloader(BaseService, PeerSubscriber):
 
             retriable_missing = self.request_tracker.get_retriable_missing()
             if retriable_missing:
-                self.logger.debug("Re-requesting %d missing trie nodes", len(timed_out))
+                self.logger.debug("Re-requesting %d missing trie nodes", len(retriable_missing))
                 try:
                     await self.request_nodes(retriable_missing)
                 except OperationCancelled:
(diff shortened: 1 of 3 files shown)


trinity/sync/full/state.py
```diff
@@ -82,7 +85,11 @@ def __init__(self,
         self.chaindb = chaindb
         self.peer_pool = peer_pool
         self.root_hash = root_hash
-        self.scheduler = StateSync(root_hash, account_db, self.logger)
+        # We use a LevelDB instance for the nodes cache because a full state download, if run
+        # uninterrupted will visit more than 180M nodes, making an in-memory cache unfeasible.
+        self._nodes_cache_dir = tempfile.TemporaryDirectory(prefix="pyevm-state-sync-cache")
```
pipermerriam: Just wanted to note that *if* we wanted to re-use this cache we should put it in `$XDG_CACHE_HOME`, though from what I can tell I don't think this is going to be easy to re-use without some fancy orchestration.

gsalgado:
This is faster because the DB is much smaller as it contains only the hashes of the state trie nodes we already have in the actual chain DB: https://github.com/ethereum/py-evm/issues/1154#issuecomment-411746777

trinity/sync/full/state.py
gsalgado: Yeah, I think reusing this cache is a bad idea as then we'd need to always ensure it is in lockstep with the actual chain DB

PR #1164, opened 2018-08-09
Lightchain integration test

### LINKED ISSUE #969: Add lightchain integration test to CI
### What is wrong?

The lightchain integration test is not currently running in CI, so it breaks regularly, and for extended periods.

(eg~ broken from #869 to #913)

### How can it be fixed

Instructions for a manual local run are here: https://github.com/ethereum/py-evm/pull/913/files#diff-19d5ae6516fbe4a30f0bcc2f45e8dc00R36

Run it locally first to be sure that tests pass on your master.

One time prep as described in the comment in the code linked above:
- Set up geth with a >=11 blocks of `geth --testnet --syncmode full` (should happen essentially instantly after you find a peer). Try to run it as short as possible while still having the first 11 blocks
- Note the `enode` in the geth logs
- tar up the geth data dir (it should be very small with so few blocks)
- add the geth data dir to a new file called `tests/p2p/integration/geth-data-ropsten.tar.gz`
- move the lightchain integration test from `tests/p2p/test_lightchain_integration.py` to `tests/p2p/integration/test_lightchain.py`

Add a new circle job that runs just that integration test. In that job only:
- untar the geth data dir
- install geth, and cache it (see web3.py integration tests with geth for some examples)
- run geth with discovery off: `geth --testnet --lightserv 90 --nodiscover`
- run only this integration test `pytest --integration --capture=no --enode="enode://$ENODE_PUBKEY_FROM_EARLIER@127.0.0.1:30303" tests/p2p/integration/test_lightchain.py`

carver:
### What was wrong?

Fixes #969 

### How was it fixed?

Other than futzing with CircleCI, this is ready for review.

- Added ~150kb geth datadir with the first few blocks from Ropsten
- Launch geth during that integration test
- Add tox test
- Add test to Circle, building geth if necessary

#### Cute Animal Picture


.circleci/config.yml
```diff
@@ -223,6 +266,14 @@ jobs:
...
         environment:
           TOXENV: py36-trinity-integration
+  py36-trinity-lightchain_integration:
+    <<: *common
+    <<: *geth_steps
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-trinity-lightchain_integration
+          GETH_VERSION: v1.8.1
```
pipermerriam: Why such an old version?

tests/trinity/integration/test_lightchain_integration.py
```diff
@@ -33,31 +42,121 @@
 )
 
 
+@pytest.fixture(scope='module')
+def geth_port():
+    sock = socket.socket()
+    sock.bind(('127.0.0.1', 0))
+    port = sock.getsockname()[1]
+    sock.close()
+    return port
```
pipermerriam: I found this the other day which might be of interest:

https://github.com/pytest-dev/pytest-asyncio#unused_tcp_port

The `pytest-asyncio` library has a pytest fixture for doing this.

tests/trinity/integration/test_lightchain_integration.py
```diff
@@ -33,31 +42,121 @@
...
+def geth_command_arguments(geth_binary, geth_ipc_path, geth_datadir, geth_port):
+    return (
+        geth_binary,
+        '--testnet',
+        '--datadir', str(geth_datadir),
+        '--ipcpath', geth_ipc_path,
+        '--lightserv', '90',
+        '--nodiscover',
+        '--fakepow',
+        '--port', str(geth_port),
```
pipermerriam: Maybe we should bake these options into a config file so that we can also include our trinity node's enode in the `StaticNodes` or `TrustedNodes` configuration so that geth maintains a connection slot for our trinity node (which I don't think can be done via CLI arguments".

.circleci/config.yml
carver: I think it's the latest supported by py-geth: https://github.com/ethereum/py-geth#installing-specific-versions-of-geth

tests/trinity/integration/test_lightchain_integration.py
carver: I think we can skip that work for this test. This enode has never been connected to the network and has `--nodiscover` on. So AIUI, the node should never connect to any peers but ours.

Issue #1061, opened 2018-07-20
Implement 8 more of the core eth_* read-only RPC endpoints


carver:
This is a tracking issue for the next alpha release milestone. The endpoints are a subset of #612, and can be chosen from that list as desired. Feel free to make individual PRs for each endpoint chosen (or group them if there are a few that are trivial to group). Link the PRs back to this issue for tracking.

  - [x] `eth_accounts` (always return `[]`)
  - [ ] `eth_syncing`
  - [ ] `eth_gasPrice` (should get its own issue to plan the algorithm)
  - [x] `eth_getBlockTransactionCountByHash`
  - [x] `eth_getBlockTransactionCountByNumber`
  - [x] `eth_getUncleCountByBlockNumber` block string
  - [x] `eth_call`
  - [x] `eth_estimateGas`
  - [ ] `eth_getTransactionByHash`
  - [x] `eth_getTransactionByBlockHashAndIndex`
  - [x] `eth_getTransactionByBlockNumberAndIndex`
  - [ ] `eth_getTransactionReceipt`
  - [x] `eth_getUncleByBlockHashAndIndex`
  - [x] `eth_getUncleByBlockNumberAndIndex` block int
  - [x] `eth_getUncleByBlockNumberAndIndex` block string
  - [ ] `eth_getLogs`

voith:
@carver I couldn't find `eth_sign`, nor did I find it #612

pipermerriam:
@voith we probably won't be implementing `eth_sign` since Trinity will not have any account management.  We can however implement something that always raises an exception since there is no way to have an unlocked account.

voith:
@pipermerriam Are there any plans to implement the personal module?
Also, would it be possible to mock this behaviour in `eth-tester`?

pipermerriam:
`personal_` will be lower priority for a while here.

1. because we don't plan to do key management.
2. because it's a non-standard API that only exists in `geth`

We can add or mock any behavior in `eth-tester` but it's a bit ambiguous to me how to best implement something like the `personal_` module functionality in a *modular* fasion.  Since it's non-standard JSON-RPC API, I wouldn't want it to be a *core* piece, but rather something that could be enabled.

carver:
Sweet, I guess someone just forgot to check off some methods after implementing them. With those, and the three more implemented, we now have 8 more than we thought we had. Just under the wire for today's release. :)

PR #1165, opened 2018-08-10
[WIP] add chain.get_transaction_result and rpc eth_call

carver:
### What was wrong?

Need an eth_call endpoint, according to #1061 

### How was it fixed?

Add a new `Chain.get_transaction_result()` method. Test that method and the identity precompile.

TODO:
- [x] test calling a contract over RPC
- test calling a contract with a few error conditions:
  - [x] out of gas
  - [x] solidity revert
- [x] figure out how to handle non-zero gas price calls from accounts without funds. Current ideas:
  - [x] **Current Winner** Monkey-patch the execution context to use the original gas price, cut price to 0 for everything else
  - We could try to monkey-patch validation
  - If we force the gas price to 0, it could affect execution
  - If we add balance to the sending account, it could affect execution
  - We could force the gas to 0, then inspect the trace for any `GASPRICE` opcodes, and bail if found
  - We could add to the user balance, then inspect the trace for any `BALANCE` opcodes, and bail if found
  - we could try each of the previous two in succession, then bail if BOTH fail.

~~I'll probably bail on this last todo and put it in an issue to handle later.~~

#### Cute Animal Picture

### DIFF
--- a/tests/core/chain-object/test_contract_call.py
+++ b/tests/core/chain-object/test_contract_call.py
@@ -1,0 +1,149 @@
+from cytoolz import (
+    assoc,
+)
+from tests.core.helpers import (
+    new_transaction,
+)
+
+from eth_utils import (
+    decode_hex,
+    function_signature_to_4byte_selector,
+    to_bytes,
+)
+import pytest
+
+from eth.exceptions import (
+    InvalidInstruction,
+    OutOfGas,
+)
+
+
+@pytest.fixture
+def chain(chain_with_block_validation):
+    return chain_with_block_validation
+
+
+@pytest.fixture
+def simple_contract_address():
+    return b'\x88' * 20
+
+
+@pytest.fixture
+def genesis_state(base_genesis_state, simple_contract_address):
+    """
+    Includes runtime bytecode of compiled Solidity:
+
+        pragma solidity ^0.4.24;
+
+        contract GetValues {
+            function getMeaningOfLife() public pure returns (uint256) {
+                return 42;
+            }
+            function getGasPrice() public view returns (uint256) {
+                return tx.gasprice;
+            }
+            function getBalance() public view returns (uint256) {
+                return msg.sender.balance;
+            }
+            function doRevert() public pure {
+                revert("always reverts");
+            }
+            function useLotsOfGas() public view {
+                uint size;
+                for (uint i = 0; i < 2**255; i++){
+                    assembly {
+                        size := extcodesize(0)
+                    }
+                }
+            }
+        }
+    """
+    return assoc(
+        base_genesis_state,
+        simple_contract_address,
+        {
+            'balance': 0,
+            'nonce': 0,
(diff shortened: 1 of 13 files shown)


pipermerriam:
> figure out how to handle non-zero gas price calls from accounts without funds. Current ideas: 

How about something that monkeypatches the execution context so that we can set a gas price of 0 externally, but internal to the EVM the `GASPRICE` opcode still returns the original value?

tests/core/chain-object/test_contract_call.py
```diff
@@ -0,0 +1,119 @@
...
+import pytest
+
+
+@pytest.fixture
+def chain(chain_without_block_validation):
+    return chain_without_block_validation
+
+
+@pytest.fixture
+def loaded_simple_contract(chain, funded_address, funded_address_private_key):
```
pipermerriam: Instead of loading this into the chain by sending a transaction, what if you instead just seeded it in the initial state.  We could alter the `chain_with_block_validation` fixtures to allow injection of extra state data.

```python
# in conftest.py
@pytest.fixture
def base_genesis_state(funded_address):
    return {...}

@pytest.fixture
def genesis_state(base_genesis_state):
    return base_genesis_state

@pytest.fixture
def chain_without_block_validation(genesis_state, ...):
    # we use the `genesis_state` here to seed the initial chain database.

# in another `test_something_about_the_evm.py` file
@pytest.fixture
def genesis_state(base_genesis_state):
    return merge(base_genesis_state, {...<custom-state-things})
```

This allows any test module to define a `genesis_state` fixture and pytest will prioritize the local one to the one in the `conftest.py` file, allowing easy injection of genesis state into those fixtures.

tests/core/chain-object/test_contract_call.py
```diff
@@ -0,0 +1,119 @@
...
+
+        transactions = base_block.transactions + (transaction, )
+        receipts = base_block.get_receipts(chain.chaindb) + (receipt, )
+
+        block = vm.set_block_transactions(base_block, new_header, transactions, receipts)
+    computation.raise_if_error()
+    return (block.header, computation.msg.storage_address)
+
+
+def get_function_selector(signature):
```
pipermerriam: [eth_utils.function_signature_to_4byte_selector]https://github.com/ethereum/eth-utils#function_signature_to_4byte_selectorfunction_signature---bytes)

tests/trinity/core/json-rpc/test_ipc.py
```diff
@@ -73,6 +83,60 @@ def can_decode_json(potential):
...
+    if isinstance(chain, MiningChain):
+        block, _, computation = chain.apply_transaction(transaction)
+    else:
+        pytest.skip('can only test on mining chain')
+
+    computation.raise_if_error()
+    return (block.header, computation.msg.storage_address)
+
+
+def get_function_selector(signature):
```
pipermerriam: same here

trinity/utils/validation.py
```diff
@@ -30,3 +34,11 @@ def validate_transaction_call_dict(transaction_dict: Dict[str, Any], vm: BaseVM)
                 list(sorted(spec_keys)),
             )
         )
+
+
+def validate_transaction_call_dict(transaction_dict: Dict[str, Any], vm: BaseVM) -> None:
+    validate_transaction_gas_estimation_dict(transaction_dict, vm)
+
+    # 'to' is required in a call, but not a gas estimation
+    if not is_address(transaction_dict.get('to', '')):
```
pipermerriam: *nitpick* Do we need the default empty string here?  Wouldn't the default return of `None` also work?

pipermerriam:
Test failure is odd.  Can you reproduce locally?  Does it persist if you change back to using the block validating chain fixture?

carver:
> Test failure is odd. Can you reproduce locally?

No :(

> Does it persist if you change back to using the block validating chain fixture?

Hm, will try that in a sec.

pipermerriam:
Also, the test isn't failing, but rather it is crashing the worker that pytest is using to run the test.  You could try removing the parallelization in ci to see if you can get at why it is failing?

carver:
Grrrr, when I run it with `-n 1` in CI, it passes. Going to SSH in and run with several different options to try to hunt it down.

pipermerriam:
I believe that the specific test in question may do a **lot** of looped EVM calls.  It is entirely possible that it's running out of memory or something like that.

pipermerriam:
Also, same weird failure showing up here: https://github.com/ethereum/py-evm/pull/1177

carver:
I was at least able to get:

> 
```
Slave 'gw3' crashed while running 'tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py::test_rpc_against_fixtures[/home/circleci/repo/fixtures/BlockchainTests/GeneralStateTests/stStaticCall/static_Call50000_identity2_d1g0v0.json:static_Call50000_identity2_d1g0v0_Byzantium-Chain]'
```

carver:
Well, it's not a great solution, but it appears if I skip that specific test then the rest passes. The test passes solo.

We haven't generally solved the problem of making sure we pass the slow tests consistently. So this test will just go into that "open problem" bucket for now.

Issue #1128, opened 2018-07-31
Start tracking peer latency


pipermerriam:
### What is it?

With the introduction of round trip request response in #1063 and #1089 we now can do very clean tracking of peer latency for various requests.

### How can it be fixed

Lets update [`Peer.get_block_headers()`](https://github.com/ethereum/py-evm/blob/f01c6563967f7f6c4c0903027d1c36c78016c776/p2p/peer.py#L636-L642) to track the following metrics about requests and responses.

- round trip time
- total number of objects in the response
- total number of requests
- average round trip time normalized for number of returned objects.

These metrics should be logged on a per-command (or maybe command pair?) basis, so that we have distinct metrics for each command we track.

Then lets update the DEBUG log reporting of peers to include this information.  Something like:

```bash
Peer: <peer-information-here>
  - BlockHeaders: count=<total-num-requests> avg=<avg-round-trip-time> avg-time-per-obj=<normalized-round-trip-time-per-object-returned>
```

PR #1166, opened 2018-08-10
Track and report peer response size/time

gsalgado:
Closes: #1128

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -201,6 +201,9 @@ class BasePeer(BaseService):
         mac_cipher = Cipher(algorithms.AES(mac_secret), modes.ECB(), default_backend())
         self.mac_enc = mac_cipher.encryptor().update
 
+    def get_extra_stats(self) -> List[str]:
+        return []
+
     @abstractmethod
     async def send_sub_proto_handshake(self) -> None:
         raise NotImplementedError("Must be implemented by subclasses")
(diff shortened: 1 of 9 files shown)


trinity/protocol/common/managers.py
```diff
@@ -33,6 +35,26 @@
...
+        total_msgs = len(self.responses)
+        total_response_time = sum([item[0] for item in self.responses])
+        total_items = sum([item[1] for item in self.responses])
+        avg_rtt = total_response_time / total_msgs
+        per_item_rtt = total_response_time / total_items
+        return 'count=%d, avg_rtt=%.2f, avg_time_per_item=%.5f' % (
+            total_msgs, avg_rtt, per_item_rtt)
+
+    def add(self, time: float, size: int) -> None:
+        self.responses.append((time, size))
```
pipermerriam: This looks like it has `O(n)` storage/memory footprint.  Unless I'm missing something we need to aggregate these stats using a running average and total count of number of responses.

p2p/peer.py
```diff
@@ -1056,21 +1063,17 @@ def _test() -> None:
             peer_pool.logger.info("Waiting for peer connection...")
             await asyncio.sleep(0.2)
         peer = peer_pool.highest_td_peer
-        block_hash = decode_hex(
-            '0x59af08ab31822c992bb3dad92ddb68d820aa4c69e9560f07081fa53f1009b152')
+        headers = await peer.requests.get_block_headers(2440319, max_headers=100)  # type: ignore
+        hashes = [header.hash for header in headers]
```
pipermerriam: clever!

trinity/protocol/common/managers.py
gsalgado: I thought this wouldn't be a problem as it's just a two-tuple for every *response* (not every msg received) and I've rarely seen peer connections lasting more than a few hours. Do you think we need to deal with this now or would a separate, low-priority issue be enough?

pipermerriam: > Do you think we need to deal with this now or would a separate, low-priority issue be enough?

That works for me.  Can you add it to the next milestone to be sure we address it quickly.

gsalgado: I'll actually do it here and save an unnecessary PR roundtrip

gsalgado: Done

pipermerriam:
Just thought of adding a timeout counter to this as well to track number of failed requests.

trinity/protocol/common/managers.py
```diff
@@ -33,6 +34,27 @@
...
+    def __init__(self) -> None:
+        self.total_msgs = 0
+        self.total_items = 0
+        self.total_response_time = 0.0
+
+    def get_stats(self) -> str:
+        if not self.total_msgs:
+            return 'None'
+        avg_rtt = self.total_response_time / self.total_msgs
+        per_item_rtt = self.total_response_time / self.total_items
```
pipermerriam: This is succeptable to division by zero if the peer is returning empty responses.

gsalgado: I've changed the stats to include the total number of items as well, and use 0 for per_item_rtt when self.total_items is 0

Issue #1075, opened 2018-07-24
Trinity: Need a transparent way to run services in their own process.


pipermerriam:
### What is wrong?

The problem statement:  Our current model for exposing APIs across processes is very verbose, requiring a lot of boilerplate.  The overhead for supporting an API across a subprocess boundary is high.  High overhead like that isn't sustainable if we're going to expose a lot of cross process APIs which I think we need.

It needs to be lower, ideally much lower.  We don't need 100% deep access to everything across service boundaries, but we do need it to be easier to expose access to APIs in a sustainable way.

### How can it be fixed

I'm not sure, but I have ideas. (see comments)

pipermerriam:
Thoughs in no particular order.

- It should be declarative
- We need both sync and async apis
- Minimal orchestration like setting up managers, and additional IPC files.
- It needs to be really hard for two systems to deadlock waiting for each other in some way.

Nice to have
- type checks work across API boundaries.
- ways to *simulate* in a testing environment without the subprocesses.
- easy and concise way to do calls with timeouts.

cburgdorf:
Whatever that will be, it will be needed for the plugins as well which could either mean, there are different kind of base plugins that use that, or that the plugin manager will use it or idk. I think it makes sense to come up with something that doesn't *live* inside the plugin facilities but is designed to work well with them (whatever that will mean in practice)

>type checks work across API boundaries

I'm tempted to raise priority for that. The loose typing of the proxy types right now is really nasty and already caused us quite some bugs in the past. I remember I once looked into it and there was something strange going that prevented me from coming up with something to raise type safety.

pipermerriam:
Noting here that #1101 is probably a great issue to use to figure this pattern out.

cburgdorf:
I was about to create a separate issue but now I think a comment here may be better suited. I started thinking about how we could move the transaction pool into it's own process. The main problem here is the interaction with the `PeerPool` which currently lives in the same process as the transaction pool but would then live outside.

Currently, the transaction pool contains code such as:

https://github.com/ethereum/py-evm/blob/e9f1216d8d3efe0c465fe0c5cb2a79b0fa3f2154/trinity/plugins/builtin/tx_pool/pool.py#L90-L105

This is going to be problematic in a multiprocess world as the peers are stateful objects that can't just be serialized/deserialized without messing up that state (e.g. state about in-flight requests etc)

[...]

PR #1172, opened 2018-08-13
PoC: Using the lahja eventbus for event based inter process communication

cburgdorf:
## DISCLAIMER

**This is not intended to get merged. This is a PROOF OF CONCEPT to demonstrate how to use the [lahja eventbus](https://github.com/cburgdorf/lahja/blob/master/README.md) for decoupled event-based inter process communication**

### What is this?

This is a PoC that is based on https://github.com/ethereum/py-evm/pull/1079 which means it also 
 contains some unrelated code. The reason this is based on https://github.com/ethereum/py-evm/pull/1079 is because that PR already lays out a potential use case for this event bus. That is, we want plugins to be able to tell the main process to gracefully shut down Trinity. This is really just one among many use cases that we will probably run into once we try to embrace the plugin infrastructure more.

**Notice that this code is full of bugs and unresolved issues so it is really just to demonstrate the potential approach**

### Keypoints

1. We setup an event bus in the main process

```Python
    event_bus = EventBus()
    main_endpoint = event_bus.create_endpoint('main')
    networking_endpoint = event_bus.create_endpoint('networking')
```

Notice, that we are creating two `Endpoints` to this eventbus. Each process is intended to hold one endpoint to the eventbus which can be used to `subscribe`, `stream` and `broadcast` events.

2. We pass the `networking_endpoint` into the networking process

```Python
networking_process = multiprocessing.Process(
        target=launch_node,
        args=(args, chain_config, networking_endpoint,),
        kwargs=extra_kwargs,
    )
```

3. In the main process, we subscribe to the `ShutdownRequested` event, which can be raised by any process that holds an endpoint to the eventbus.

```Python
    main_endpoint.subscribe(
        ShutdownRequested,
        lambda event: kill_trinity_gracefully(
            database_server_process,
            networking_process,
            logger
        )
    )

The main process listens to the event and causes a clean shutdown.
```

4. The `PluginContext` that each plugin has access to also holds one endpoint. Currently the `networking_endpoint` since all plugins run into that shared process.

```Python
class PluginContext:

    def __init__(self, endpoint: Endpoint):
        self.eventbus = endpoint

    def shutdown_trinity(self, exit_code: int = 0) -> None:
        self.eventbus.broadcast(ShutdownRequested('meh'))
```

5. The `TxPool` plugin calls `shutdown_trinity` on the plugin context

```Python
if light_mode:
    self.logger.warning('The transaction pool is not yet available in light mode')
    self.context.shutdown_trinity(exit_code=1)
```
(Theoretically plugins will most likely also hold a direct reference to the eventbus endpoint and only go through API calls on the context for "well defined" stuff such as a Trinity shutdown)

We can see this already working on this PR when running `trinity --tx-pool --light`. The networking process will broadcast the `ShutdownRequested` event, the main process will get it and shutdown Trinity.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ deps = {
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.5",
         "web3==4.4.1",
+        "lahja==0.1.0-alpha.0",
     ],
     'test': [
         "hypothesis==3.44.26",
--- a/tests/trinity/integration/test_trinity_cli.py
+++ b/tests/trinity/integration/test_trinity_cli.py
@@ -74,6 +74,24 @@ async def test_txpool_full_boot(async_process_runner, command):
 @pytest.mark.parametrize(
     'command',
     (
+        ('trinity', '--light', '--tx-pool',),
+        ('trinity', '--light', '--ropsten', '--tx-pool',),
+    )
+)
+@pytest.mark.asyncio
+async def test_txpool_deactivated(async_process_runner, command):
+    await async_process_runner.run(command)
+    assert await contains_all(async_process_runner.stderr, {
+        "Started DB server process",
+        "Started networking process",
+        "The transaction pool is not yet available in light mode",
+        "IPC started at",
+    })
+
+
+@pytest.mark.parametrize(
+    'command',
+    (
         ('trinity', '--light',),
         ('trinity', '--light', '--ropsten',),
     )
--- a/trinity/events.py
+++ b/trinity/events.py
@@ -1,0 +1,6 @@
+from lahja import (
+    BaseEvent
+)
+
+class ShutdownRequested(BaseEvent):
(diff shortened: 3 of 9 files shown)


pipermerriam:
One thing that I haven't seen mentioned is how we'll deal with this across process boundaries.  This might however just be something we have to figure out when we do https://github.com/ethereum/py-evm/issues/1075

cburgdorf:
@pipermerriam this *is* across process boundaries. The event is raised in the network process and caught in the main process.

Here's a more focused example with three processes:

https://github.com/cburgdorf/lahja/blob/master/examples/inter_process_ping_pong.py

pipermerriam:
Ah, I see!

cburgdorf:
I'm going to close this as #1211 is a more modern version showcasing the same thing.

Issue #1118, opened 2018-07-30
Manage background tasks in BaseService


carver:
### What is wrong?

BaseService handles background tasks awkwardly right now. For example:
- have to `await asyncio.sleep(0)` in cleanup to make sure that all tasks cancel
- often need to catch and drop `OperationCancelled` exceptions

This is what kicked off the thought: https://github.com/ethereum/py-evm/pull/1097/files#diff-3942908a5c806895979f03283db24df3R206

An toy example with these quirks:

```py
class MyService(BaseService):
  async def check_connection(self):
    while self.is_running:
      if not is_connected():
        await reconnect()
      try:
        await self.sleep(60.0)
      except OperationCancelled:
        break

  async def _run(self):
    asyncio.ensure_future(self.check_connection())

  async def _cleanup(self):
    # We don't need to cancel() anything, but we yield control just so that the coroutines we
    # run in the background notice the cancel token has been triggered and return.
    # Also, don't use self.sleep() here as the cancel token will be triggered and that will
    # raise OperationCancelled.
    await asyncio.sleep(0)
```

### How can it be fixed

Add a `BaseService.loop_in_background(my_task)` method that handles these things for you. Then services can define something like:

```py
class MyService(BaseService):
  async def check_connection(self):
    if not is_connected():
      await reconnect()
    await self.sleep(60.0)

  async def _run(self):
    self.loop_in_background(self.check_connection())
```

pipermerriam:
Conceptually, :+1: 

naming thoughts as the `loop_` prefix is a bit confusing to me

- `run_in_background`
- `ensure_future` (to borrow semantics from the `asyncio` API)
- `handle_background_task`
- `handle_task`
- `run_background_task`

I like the `background` terminology.  The word `daemon` is shorted and also a well known concept.

- `run_daemon`

carver:
The important semantic part of `loop` (as a verb) is that `BaseService` would keep running the method over and over until the service shut down. I haven't found another word that conveys that as well. `daemon` itself kind of implies that, but usually a `daemon` would handle its own looping, which would be a lot less convenient. Maybe:

- `cycle_task`
- `repeat_task`
- `repeat_forever`

gsalgado:
[...]

PR #1173, opened 2018-08-13
Implement BaseService.run_task(awaitable)

gsalgado:
It runs the given awaitable in the background, catches/ignores
OperationCancelled and ensures the task is finished before the
service is finished.

Also removes trinity.Node.add_service() as the same
functionality is provided by BaseService.run_child_service()

Closes: #1118


p2p/service.py
```diff
@@ -114,12 +119,26 @@ def get_event_loop(self) -> asyncio.AbstractEventLoop:
...
+        The awaitable should return whenever this service's cancel token is triggered.
+
+        If it raises OperationCancelled, that is caught and ignored.
+        """
+        async def f() -> None:
+            try:
+                await awaitable
+            except OperationCancelled:
+                pass
+        self._tasks.add(asyncio.ensure_future(f()))
```
pipermerriam: I'm not seeing anywhere that we cleanup finished tasks.  Things like `Discovery.mayber_connect_to_more_peers` which run periodically will result in an ever growing memory footprint.  It looks like we need something to do periodic garbage collection for tasks that are complete.

gsalgado: That's why I used a `WeakSet` for `self._tasks`; once a task is done there should be no strong reference to it left, and then it is discarded from the set.

gsalgado: https://github.com/ethereum/py-evm/pull/1173/files#diff-bea0ad88fecb099146d53835e5f8fc78R35

pipermerriam: That's fancy!  :+1:

PR #1175, opened 2018-08-13
Reduce logging level of BaseService cancellation message

pipermerriam:
### What was wrong?

The logging message the the `BaseService` emits on cancellation is now too noisy with the increased number of services being run.

### How was it fixed?

Reduced the logging level to `DEBUG`

#### Cute Animal Picture

baby-deer-puppy-cute-animal-picture-588632

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -94,7 +94,7 @@ class BaseService(ABC, CancellableMixin):
                 self.events.started.set()
                 await self._run()
         except OperationCancelled as e:
-            self.logger.info("%s finished: %s", self, e)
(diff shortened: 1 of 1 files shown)


gsalgado:
I think we should keep it as an INFO log as for most services we want to know when/why they return. Maybe refactor this code somehow so that we use a DEBUG only for `*RequestManager`s?

pipermerriam:
@gsalgado you don't think it's sufficient to have these in the `DEBUG` logs?  They are still accessible and in the event that we're trying to understand why things start/stop aren't we already likely to be looking in the debug logs?

PR #1176, opened 2018-08-13
Move handling of MalformedMessage to the main Peer._run

pipermerriam:
### What was wrong?

Turns out we only handled `MalformedMessage` during DAO fork checks.  If a peer sends a malformed message outside of this context, it isn't caught.

### How was it fixed?

Moved the handling of `MalformedMessage` directly into the `Peer._run` loop.

#### Cute Animal Picture

animal-odd-couples

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -356,6 +356,9 @@ class BasePeer(BaseService):
                     exc_info=True,
                 )
                 return
+            except MalformedMessage as err:
+                await self.disconnect(DisconnectReason.bad_protocol)
+                return
 
             try:
                 self.process_msg(cmd, msg)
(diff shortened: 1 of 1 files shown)


p2p/peer.py
```diff
@@ -356,6 +356,9 @@ def is_closing(self) -> bool:
                     exc_info=True,
                 )
                 return
+            except MalformedMessage as err:
+                await self.disconnect(DisconnectReason.bad_protocol)
```
carver: If we receive an unexpected response, do we raise a `MalformedMessage`?

Like if we do:
- connect to peer
- request header
- disconnect from peer
- connect to same peer
- request block body
- receive header... `raise MalformedMessage` ?

Most `MalformedMessage`s seem like a good reason to disconnect, but that ^ doesn't.

pipermerriam: No, malformed message is literally a message that we cannot parse, invalid RLP encoding, unable to decrypt, etc.  A message that when we pass it through the msg decoder, something fundamental fails.  Nothing to do with validation, or the request/response round trips.

PR #1177, opened 2018-08-13
Piper/fix cancellation error in full sync service

pipermerriam:
### What was wrong?

The `full` syncer service throws errors about not being able to restart a cancelled service.

Many of our services used `while True` in their `_run` loop.  

### How was it fixed?

The full syncer `_run` loop now has exit clauses between the different types of syncing to exit if cancellation has been triggered.

Also changed all of the `while True` statements to a more correct way to  `while self.is_running` so that the loop automatically exits upon cancellation.

#### Cute Animal Picture

bb0949ad8ea8fcbe9c61a2267aa2993c

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -277,11 +277,11 @@ def run_database_process(chain_config: ChainConfig, db_class: Type[BaseDB]) -> N
         base_db = db_class(db_path=chain_config.database_dir)
 
         manager = get_chaindb_manager(chain_config, base_db)
+        server = manager.get_server()  # type: ignore
 
         def _sigint_handler(*args: Any) -> None:
             server.stop_event.set()
 
-        server = manager.get_server()  # type: ignore
         signal.signal(signal.SIGINT, _sigint_handler)
 
         try:
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -28,7 +28,7 @@ from trinity.exceptions import (
 from trinity.chains import (
     initialize_data_dir,
     is_data_dir_initialized,
-    serve_chaindb,
+    get_chaindb_manager,
 )
 from trinity.cli_parser import (
     parser,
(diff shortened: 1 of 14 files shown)


pipermerriam:
@carver same strange test failure here: https://circleci.com/gh/ethereum/py-evm/59546?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

trinity/main.py
```diff
@@ -294,6 +306,7 @@ def exit_because_ambigious_filesystem(logger: logging.Logger) -> None:
     await sigint_received.wait()
     try:
         await service_to_exit.cancel()
+        service_to_exit._executor.shutdown(wait=True)
```
carver: Maybe this should be in `BaseService`'s `cancel()`.

pipermerriam:
> The _start_queue_management_thread is a big enough mystery to me, that I'm not ready to +1 for merge yet.

I'll get an expanded comment in there to explain in more detail why that is there.  In short, it is two independent hacks to ensure that the worker processes shut themselves down in a quiet manner during state sync.  First hack is to ensure that `SIGINT` is ignored at the time that the worker processes start up since processes inherit their signal handlers from their parent process.  This is required because the `concurrent.futures.ProcessPoolExecutor` doesn't expose an `initializer` callback like the `multiprocessing.ProcessPool` does.  The second is to call that private method, which is what initializes all of the worker processes.  The `ProcessPoolExecutor` class is lazy and doesn't actually spin up the processes until the first batch of work is submitted.  Reading the source for that module, this was a very clean way to trigger that initialization.

trinity/main.py
```diff
@@ -276,7 +276,19 @@ def run_database_process(chain_config: ChainConfig, db_class: Type[BaseDB]) -> N
     with chain_config.process_id_file('database'):
         base_db = db_class(db_path=chain_config.database_dir)
 
-        serve_chaindb(chain_config, base_db)
+        manager = get_chaindb_manager(chain_config, base_db)
+
+        def _sigint_handler(*args: Any) -> None:
+            server.stop_event.set()
```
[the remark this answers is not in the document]
pipermerriam: wow, that's totally a copy paste accident and I'm amazed it works if it does.  And if it doesn't, there isn't any exception escaping..., will investigate tomorrow morning.

trinity/main.py
carver: Hah, right, makes sense...

carver:
You can rebase to get the byzantium state test passing

trinity/main.py
pipermerriam: Which btw, and not really related to this PR:

I've been thinking about mapping out our service hierarchy sometime soon.  I think it'd be really useful to have a graph that shows the hierarchy of services as well as the peer subscriptions.  Otherwise, visualizing that stuff takes a **lot** of domain knowledge.

carver: I suppose we could *generate* this service hierarchy for display at runtime, which could be neat and not require maintenance as we add more services.

pipermerriam: Maybe we could do this as a plugin which gathered this information from the events that get broadcast over the event bus.  Subscriptions and running child services both strike me as reasonable events to be fired off and with that it'd be pretty easy to scrape the application hierarchy.

Alternatively, we would need to add some extra API surface area to the Service and subscription API to make this data available.  Gonna put this on the back-burner for now, but I like the idea.

Issue #1003, opened 2018-07-10
Trinity does not accept incoming peer connections when in light mode


gsalgado:
It also does not set up NAT port mapping. If we did both, we should be able to keep a higher peer count and faster syncs. But do we want that?

pipermerriam:
What are the trade-offs here?  My intuition is:

- lower peer count -> more performant light node assuming it finds a sufficient number of peers serving LES
- higher peer count -> client is a more *responsible* participant in the network, serving headers when it can.

gsalgado:
> lower peer count -> more performant light node assuming it finds a sufficient number of peers serving LES

Do you mean that support the LES protocol or that have `serveHeaders` set to True? In light mode we currently connect to any peers that support the LES protocol, regardless of whether or not they have `serveHeaders` set to True.

We could accept incoming connections but have a low default value for `max_peers`

pipermerriam:
I wasn't aware of the `serveHeaders` flag so not that.

carver:
:+1: to allowing light peers to connect to us, with a low-ish limit on leeching light peers (and probably a rate-limiter on light serving requests)

PR #1179, opened 2018-08-14
Use trinity.Server in LightNode

gsalgado:
This avoids duplicating some logic to run a PeerPool/DiscoveryService,
but also causes LightNode to setup NAT port forwarding and accept
incoming peer connections.

Also changed the default peer limit when running in light mode to half
of the limit used when in full mode.

Closes: #1003

### DIFF
--- a/tests/trinity/core/chain-management/test_initialize_data_dir.py
+++ b/tests/trinity/core/chain-management/test_initialize_data_dir.py
@@ -13,7 +13,7 @@ from trinity.config import (
 
 @pytest.fixture
 def chain_config():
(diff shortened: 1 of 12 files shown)


carver:
> Also changed the default peer limit when running in light mode to half
of the limit used when in full mode.

What's the intuition behind this change?

gsalgado:
@carver a lower default limit seemed to make sense for light mode as that generally means low-end hardware. if we don't accept incoming connections that probably doesn't matter as with discv4 we'd probably never find more than a handful peers, but now that we accept incoming connections we're more likely to reach the limit

PR #1180, opened 2018-08-14
Add docker support

cburgdorf:
### What was wrong?

No docker support

### How was it fixed?

Added docker support.

#### Cute Animal Picture


cburgdorf:
I didn't. But should work just fine I guess as it will technically be
running on an Ubuntu that's just hosted on a windows machine. From the
perspective of the image, the host machine should be entirely transparent.

pipermerriam:
> Just curious: have you had a chance to try this on a Windows machine?

Time for us to try out https://www.appveyor.com/ again?  See if we can get a minimal subset of our tests running there and have proper windows support?!

carver:
> Time for us to try out https://www.appveyor.com/ again?

I'd say it's a must have before reaching `beta`. To start, maybe we can run the trinity launch tests.

cburgdorf:
It looks like we control `ethereum/trinity` on docker hub now. I'm currently pushing a test image to see if things work as expected and in case it does, I'll apply the final finish here to wrap things up.

cburgdorf:
@pipermerriam @carver I just pushed to tags `latest` and `0.1.0-alpha.15` to the docker hub (Same image)

https://hub.docker.com/r/ethereum/trinity/tags/

I actually wanted to post a screenshot demoing running Trinity on Windows but boy, installing Docker on Windows is such a pain in the ass. I gave up. But for anyone who has Docker already installed ([with support to run Linux containers](https://docs.docker.com/docker-for-windows/#switch-between-windows-and-linux-containers)) running trinity should be as easy as `docker run -it ethereum/trinity`.

carver:
> installing Docker on Windows is such a pain in the ass. I gave up.

Hm, this doesn't inspire confidence in our short-term "it works on Windows, because it works in Docker" plan.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+RUN apt-get update && apt-get install -y libssl-dev
+RUN apt-get update && apt-get install -y python3-pip
+RUN apt-get update && apt-get install -y pandoc
+
+#Copy the application py-evm to the /usr/src/app folder
+COPY . /usr/src/app
+
+# Remove existing virtualenv directory
+RUN rm -rf venv_docker_build
```
carver: Isn't the point of a dockerfile that it starts from scratch at the beginning of the file, so this folder would never exist?

cburgdorf: I'm not an expert on Dockerfiles but when this docker file runs to create the image it is based on the actual source code. So whatever you have on disk is taken into account. Hence, you can create an image even from some temporary state that isn't even committed. So it can happen that a previous `venv_docker_build` exists. That btw, is the reason why I'm not using `venv` here to not clash or delete an existing `venv` that may be around.

carver: Hm, it still doesn't sound right to me. Maybe you created that virtualenv folder in your host machine outside of docker? Changes in the Dockerfile at line 18 shouldn't propagate back to the host disk, or be still present on a rerun of the Dockerfile.

PR #1181, opened 2018-08-14
Enable state tests for Constantinople

cburgdorf:
### What was wrong?

We need to run the global state tests from `ethereum/test` for the upcoming constantinople fork

### How was it fixed?

- updated `fixtures` against https://github.com/ethereum/tests/commit/10ab37c095bb87d2e781bcf112b6104912fccb44 which is a WIP branch containing some of the new constantinople tests. We'll most likely have to update this to a different version prior to merge
- configured `test_state.py` to handle Constantinople tests

We need to keep this PR open and continue to rebase it until we get all tests passing. Also notice that there are some tests failing for other forks as well which either means, we have some evm bugs or just need to adjust the test setup somehow.

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,12 @@ jobs:
       - image: circleci/python:3.5
         environment:
           TOXENV: py35-native-state-byzantium
+  py35-native-state-constantinople:
+    <<: *common
+    docker:
+      - image: circleci/python:3.5
+        environment:
+          TOXENV: py35-native-state-constantinople
   py35-native-state-frontier:
     <<: *common
     docker:
@@ -175,6 +181,12 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: py36-native-state-byzantium
+  py36-native-state-constantinople:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-native-state-constantinople
   py36-native-state-frontier:
     <<: *common
     docker:
@@ -205,6 +217,12 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: py36-rpc-state-byzantium
+  py36-rpc-state-constantinople:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-rpc-state-constantinople
   py36-rpc-state-frontier:
     <<: *common
     docker:
@@ -340,11 +358,13 @@ workflows:
       - py37-beacon
 
       - py36-native-state-byzantium
+      - py36-native-state-constantinople
       - py36-native-state-frontier
       - py36-native-state-homestead
       - py36-native-state-eip150
       - py36-native-state-eip158
       - py36-rpc-state-byzantium
+      - py36-rpc-state-constantinople
       - py36-rpc-state-frontier
       - py36-rpc-state-homestead
       - py36-rpc-state-eip150
@@ -364,6 +384,7 @@ workflows:
       - py36-beacon
 
       - py35-native-state-byzantium
+      - py35-native-state-constantinople
       - py35-native-state-frontier
       - py35-native-state-homestead
       - py35-native-state-eip150
--- a/eth/__init__.py
+++ b/eth/__init__.py
@@ -21,7 +21,7 @@ from eth.chains import (  # noqa: F401
 #
 #  Ensure we can reach 1024 frames of recursion
 #
-EVM_RECURSION_LIMIT = 1024 * 10
+EVM_RECURSION_LIMIT = 1024 * 12
 sys.setrecursionlimit(max(EVM_RECURSION_LIMIT, sys.getrecursionlimit()))
 
 
--- a/eth/tools/fixtures/helpers.py
+++ b/eth/tools/fixtures/helpers.py
@@ -40,6 +40,7 @@ from eth.vm.base import (
     BaseVM,
 )
 from eth.vm.forks import (
+    ConstantinopleVM,
     ByzantiumVM,
     TangerineWhistleVM,
     FrontierVM,
@@ -122,6 +123,10 @@ def chain_vm_configuration(fixture: Dict[str, Any]) -> Iterable[Tuple[int, Type[
     elif network == 'Byzantium':
         return (
             (0, ByzantiumVM),
+        )
+    elif network == 'Constantinople':
+        return (
+            (0, ConstantinopleVM),
         )
     elif network == 'FrontierToHomesteadAt5':
         HomesteadVM = BaseHomesteadVM.configure(support_dao_fork=False)
--- a/eth/vm/logic/system.py
+++ b/eth/vm/logic/system.py
@@ -20,12 +20,9 @@ from eth.utils.numeric import (
     ceil32,
 )
 from eth.vm import mnemonics
-from eth.vm.computation import (
-    BaseComputation
-)
-from eth.vm.opcode import (
-    Opcode,
-)
+from eth.vm.computation import BaseComputation
+from eth.vm.message import Message
+from eth.vm.opcode import Opcode
 
 from .call import max_child_gas_eip150
 
@@ -193,13 +190,16 @@ class Create(Opcode):
             code=call_data,
             create_address=contract_address,
         )
+        self.apply_create_message(computation, child_msg)
 
+    def apply_create_message(self, computation: BaseComputation, child_msg: Message) -> None:
         child_computation = computation.apply_child_computation(child_msg)
 
         if child_computation.is_error:
             computation.stack_push(0)
         else:
-            computation.stack_push(contract_address)
+            computation.stack_push(child_msg.storage_address)
+
         computation.return_gas(child_computation.get_gas_remaining())
 
 
@@ -240,3 +240,22 @@ class Create2(CreateByzantium):
             stack_data.salt,
             call_data
         )
+
+    def apply_create_message(self, computation: BaseComputation, child_msg: Message) -> None:
+        # We need to ensure that creation operates on empty storage **and**
+        # that if the initialization code fails that we revert the account back
+        # to its original state root.
+        snapshot = computation.state.snapshot()
+
+        computation.state.account_db.delete_storage(child_msg.storage_address)
+
+        child_computation = computation.apply_child_computation(child_msg)
+
+        if child_computation.is_error:
+            computation.state.revert(snapshot)
+            computation.stack_push(0)
+        else:
+            computation.state.commit(snapshot)
+            computation.stack_push(child_msg.storage_address)
+
+        computation.return_gas(child_computation.get_gas_remaining())
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -23,10 +23,13 @@ from eth.vm.forks.spurious_dragon import SpuriousDragonVM
 # Uncomment this to have logs from tests written to a file.  This is useful for
 # debugging when you need to dump the VM output from test runs.
 """
+import datetime
+import logging
+import os
+from eth.tools.logging import TRACE_LEVEL_NUM
+
 @pytest.yield_fixture(autouse=True)
 def _file_logging(request):
-    import datetime
-    import os
 
     logger = logging.getLogger('eth')
 
--- a/tests/json-fixtures/test_blockchain.py
+++ b/tests/json-fixtures/test_blockchain.py
@@ -45,6 +45,11 @@ INCORRECT_UPSTREAM_TESTS = {
     # The result is in conflict with the yellow-paper:
     # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418800369
     ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Byzantium'),  # noqa: E501
+    ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Constantinople'),  # noqa: E501
+    # The CREATE2 variant seems to have been derived from the one above - it, too,
+    # has a "synthetic" state, on which py-evm flips.
+    # * https://github.com/ethereum/py-evm/pull/1181#issuecomment-446330609
+    ('GeneralStateTests/stCreate2/RevertInCreateInInitCreate2_d0g0v0.json', 'RevertInCreateInInitCreate2_d0g0v0_Constantinople'),  # noqa: E501
 }
 
 
@@ -84,8 +89,6 @@ def fixture(fixture_data):
         fixture_key,
         normalize_blockchain_fixtures,
     )
-    if fixture['network'] == 'Constantinople':
-        pytest.skip('Constantinople VM rules not yet supported')
     return fixture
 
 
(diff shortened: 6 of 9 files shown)


veox:
The `py3{5,6}-vm` failures are... "interesting".

(A page of reasoning shredded as irrelevant...)

The failure tested for by `exp8.json` is present in `py-evm`. Good work, @ehildenb! :D

Will submit micro-fix PR...

veox:
Regarding the `state` CI run failures:

I've tried running `test_state_fixtures` using latest `ethereum/tests` branch `develop` (at https://github.com/ethereum/tests/commit/691680aa7fca44a77447ee701a78ed0062837218) and latest `py-evm` branch `master` (at fcf73d071278fab5f85c221f89cb95fcbeb942a3), since

* it's starting to see Constantinople-related PRs merged (e.g. https://github.com/ethereum/tests/pull/491),
* that's the branch [recommended in the `README`](https://github.com/ethereum/tests#i-want-to-test-my-client-which-version-shall-i-use),
* I wanted something less WIP'y.

There were:

```
77 failed, 9960 passed, 1124 skipped, 30633 deselected
```

It _might_ be prudent to tease apart the commits updating fixtures and enabling Constantinople.

pipermerriam:
#1217 has been merged so this can be rebased.

veox:
PR #1224 bumped fixtures, so this PR should be rebased against `master`. Should show 74 failures in CI job `py36-rpc-state-byzantium` (I think).

veox:
@cburgdorf I meant they'd show if you rebased against master (to incorporate the few fixes) _and_ bumped `fixtures` back up - either to latest commit in upstream's `develop` branch, or that WIP branch you were using originally.

It would then show 74 failures, instead of the 77 it was showing previously. That is, if there weren't _new_ tests added upstream (I think there were) that would fail on `py-evm` (e.g. because "not implemented", as you note).

cburgdorf:
@veox Ah gotcha! Thanks for clarifying!

pipermerriam:
I suspect we can remove the blocked tag from this?

cburgdorf:
@pipermerriam 

>I suspect we can remove the blocked tag from this?

Yes, I think so. I'll try to move this forward today. Btw, is it correct to assume that we can not activate the constantinople **rpc** test before the Constantinople VM actually landed in the mainnet config with a block number. Because, these tests are based on the `MainnetFullChain` so I suspect they can not work against Constantinople anyway, no?

https://github.com/ethereum/py-evm/blob/1aa06dac1784bd55efc9e284242aad273e4f9031/tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py#L389

cburgdorf:
@pipermerriam Here's a summary of the current state of this PR.

1. There are some tests failing for `native-state-constantinople` and I have a hard time telling if it is our implementation or the tests that are at fault. Would greatly appreciate any help.

2. https://github.com/ethereum/py-evm/pull/1181/commits/bc485452074117c001fd4f047fe24184ab286e78 doesn't seem to be needed anymore (also I could swear it was at some point it was). However, I think it is actually more correct. E.g. without this fix, the `py36-native-state-homestead` runs has **2432 tests passing and 34 skipped**. With this fix it is **2031 tests passing and 30 skipped**. The higher number of tests without this fix is due to the `-k Homestead` glob not going precisely only against tests that are meant for this fork but also for those that are e.g. meant to run against `Byzantium` but where the test name or directory has contains the string `Homestead`.

Since the `-k` runs against something like `tests/json-fixtures/test_state.py::</path/to/test.json>:<TestName>:<ForkName>:0` including the `:` is a simple way to ensure we ony match against the actual `<ForkName> and not the fork name appearing in the name of the directory or test.

3. I'm not sure if activation of the `rpc-state-constantinople` test is currently useful or not. Since the Constantinople VM isn't yet active in the mainnet and these tests are based on the `MainnetFullChain` I suspect there's no point in running them, no?

https://github.com/ethereum/py-evm/blob/1aa06dac1784bd55efc9e284242aad273e4f9031/tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py#L389

pipermerriam:
1) I can help you with these now/today if you'd like
2) bc48545 makes sense.  It'd be cool if we could make this even more precise somehow but not worth investing much time in this I don't think.
3) we should still run these, we just need to wire it up with a version of that chain with a different `vm_configuration` that has Constantinople enabled.

cburgdorf:
>I can help you with these now/today if you'd like

Thank you. I'm afraid I won't get to it today. If you have a urge to investigate on your own don't wait for me. If not, we may find some time in Prague to work on this together.

>It'd be cool if we could make this even more precise somehow but not worth investing much time in this I don't think.

In fact, I had something where we explicitly passed `--fork X` to `pytest` and then directly filtered the tests to not generate the wrong ones in the first place. In the end I didn't like it that much so I threw it away (could dig in my reflog to look for it though)

>we should still run these, we just need to wire it up with a version of that chain with a different vm_configuration that has Constantinople enabled.

Ok, makes sense.

pipermerriam:
> I had something where we explicitly passed `--fork X`

I at least like this on the surface level.  Better than the `-k` mechanism.

cburgdorf:
@pipermerriam I digged it out of the dirt https://github.com/ethereum/py-evm/pull/1181/commits/9047fac1821eed2ce87acac41cd6a865c7f31161

cburgdorf:
>Can you split this out into something stand-alone?

@pipermerriam There you go https://github.com/ethereum/py-evm/pull/1470

cburgdorf:
Right, I remember hunting it as far down to noticing that it is somehow related to gas costs. Btw, I used ethereum-js to get a trace output for that just because I figured that would be easier for me and they have these tests passing already. 

I eventually gave up on it though. It kinda makes me feel bad that it ended up on your table now but at the same time I guess this PR can give you a lot of insights about the state of testing/debugging and what could be improved about it :)

veox:
Perhaps of interest is [EIP-1014](https://eips.ethereum.org/EIPS/eip-1014) (introducing `CREATE2`), and the test case's [filler](https://github.com/ethereum/tests/blob/v6.0.0-beta.2/src/GeneralStateTestsFiller/stBadOpcode/badOpcodesFiller.json).

The contract at test is rather simple:

`0x60016001600160016001600160016001`**`XX`**`60006000f3`

... where `XX` are 120 different opcodes. (There is an oddball variant - see under break.)

The contract pushes 8 single-byte `0x01` elements to the stack (in a wasteful manner), executes `XX` (opcode under test), then `RETURN`s "nothing" if the operation passes.

-----

The [very first case](https://github.com/ethereum/tests/blob/v6.0.0-beta.2/src/GeneralStateTestsFiller/stBadOpcode/badOpcodesFiller.json#L63) is slightly different, and might be a typo:

`0x60016001600160016001600160016001`**`5800`**`006000f3`

That is, instead of pushing a `0x00` to stack (to be used as argument in `RETURN`), it has a `STOP`.

Not sure if this should be brought upstream to `ethereum/tests`.

veox:
@lithp 

Looking at the traces, the interpretation should be as follows IMO:

`geth`:

```
PUSH1           pc=00000012 gas=345614 cost=3
PUSH1           pc=00000014 gas=345611 cost=3
CREATE2         pc=00000016 gas=345608 cost=32009  ;; 32000 (CREATE) + 6 (CREATE2-specific hash-cost) + 3 (memory expansion: 0->32)
STOP            pc=00000000 gas=308700 cost=0      ;; bogus: not part of execution; contract created
                                                   ;; gas at this point: 313599
PUSH1           pc=00000017 gas=313599 cost=3
PUSH1           pc=00000019 gas=313596 cost=3
RETURN          pc=00000021 gas=313593 cost=0
```

`py-evm`:

```
GAS CONSUMPTION: 345614 - 3 -> 345611 (PUSH1)
GAS CONSUMPTION: 345611 - 3 -> 345608 (PUSH1)
GAS CONSUMPTION: 345608 - 6 -> 345602 (CREATE2)     ;; CREATE2-specific hash-cost
GAS CONSUMPTION: 345602 - 3 -> 345599 (Expanding memory 0 -> 32)
GAS CONSUMPTION: 345599 - 340200 -> 5399 (CREATE2)  ;; 340200?.. what?..
GAS RETURNED: 5399 + 340200 -> 345599               ;; ... :/
                                                    ;; gas at this point: 345599
GAS CONSUMPTION: 345599 - 3 -> 345596 (PUSH1)
GAS CONSUMPTION: 345596 - 3 -> 345593 (PUSH1)
```

The trace formatting is not identical, so you should consider per-opcode segments, not single lines.

The difference 345599 - 313599 = 32000, or the cost of a `CREATE` that `py-evm` seems to be missing.

Possibly `get_gas_cost()` not calling its `super()`?..

cburgdorf:
yes, if I recall correctly the implementation only charges the dynamic cost and misses to add the fixed cost of a `CREATE`.

https://github.com/ethereum/py-evm/blob/5ef07d9319017282dbccd53d2e55ba0ec65f1cc4/eth/vm/logic/system.py#L229-L230

But I *think* I remember that I tried this locally and even then the test still failed at which point I gave up on it :see_no_evil: :hear_no_evil:

veox:
The `340200` in `CREATE2` is mighty confusing. Where does it come from? Perhaps `py-evm` sets up a nested call frame, which it enters?.. I'll take a look within a week if this is still stuck.

veox:
Can confirm what @cburgdorf wrote:

``` diff
diff --git a/eth/vm/logic/system.py b/eth/vm/logic/system.py
index a6ad3aaa..9a27a452 100644
--- a/eth/vm/logic/system.py
+++ b/eth/vm/logic/system.py
@@ -227,7 +227,7 @@ class Create2(CreateByzantium):
         return CreateOpcodeStackData(endowment, memory_start, memory_length, salt)
 
     def get_gas_cost(self, data: CreateOpcodeStackData) -> int:
-        return constants.GAS_SHA3WORD * ceil32(data.memory_length) // 32
+        return constants.GAS_CREATE + constants.GAS_SHA3WORD * ceil32(data.memory_length) // 32
 
     def generate_contract_address(self,
                                   stack_data: CreateOpcodeStackData,
```

... makes the gas consumption identical to that of `geth`'s, but the test still fails (state root mismatch).

Blind guess: perhaps because class `Create2` does not implement `__call__()`, with `CREATE2`-specific operation?..

veox:
After some hacking around `test_blockchain_fixtures()` - need to document this!.. - here's our diff with the [state as declared in the test JSON](https://github.com/ethereum/tests/blob/v6.0.0-beta.2/BlockchainTests/GeneralStateTests/stBadOpcode/badOpcodes_d115g0v0.json#L163):

```
E           AssertionError: State DB did not match expected state on 1 values:
E           0x6295ee1b4f6dd65047762f924ecd367c17eabf8f(nonce) | Actual: 1 | Expected: 2
```

`0x62..8f` is the empty-`to` destination to which the transaction submitter `0xa9..0b` pushes its transient payload.

pipermerriam:
@cburgdorf or @lithp (whomever ends up running with this)

Looking at the current failures after the `CREATE2` fix was merged by @veox it looks like we may need to do something along the lines of clearing storage for the newly created account as well.

veox:
The recursive `CREATE2` test that overflows the test run's log is likely related to the issue that's been submitted as a CVE with much publicity recently?..

veox:
@pipermerriam I'm not convinced the CI run has been rebased on current `master`. I've run the tests locally, and some of those shown as failing by CircleCI pass just fine on my end. :/

pipermerriam:
@veox I just rebased on current master and pushed.

veox:
`RevertInCreateInInitCreate2` fails because of

```
E           AssertionError: State DB did not match expected state on 2 values:
E           0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba(balance) | Actual: 2000000000000113058 | Expected: 2000000000000128058 | Delta: 15000
E            - 0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b(balance) | Actual: 429496616542 | Expected: 429496601542 | Delta: -15000
```

(`0x2adc..` is miner, `0xa94f..` is tx origin.)

Trace shows that the post-`REVERT` there are these `SSTORE`s:

```
   TRACE  12-10 23:26:06            logging.py  OPCODE: 0x55 (SSTORE) | pc: 19
   TRACE  12-10 23:26:06            logging.py  GAS CONSUMPTION: 42949585131 - 200 -> 42949584931 (SSTORE: 0x6295ee1b4f6dd65047762f924ecd367c17eabf8f[2] -> 0 (current: 0 / original: 0))
...
   TRACE  12-10 23:26:06            logging.py  OPCODE: 0x55 (SSTORE) | pc: 23
   TRACE  12-10 23:26:06            logging.py  GAS CONSUMPTION: 42949584926 - 5000 -> 42949579926 (SSTORE: 0x6295ee1b4f6dd65047762f924ecd367c17eabf8f[0] -> 32 (current: 1 / original: 1))
...
   TRACE  12-10 23:26:06            logging.py  OPCODE: 0x55 (SSTORE) | pc: 36
   TRACE  12-10 23:26:06            logging.py  GAS CONSUMPTION: 42949579902 - 20000 -> 42949559902 (SSTORE: 0x6295ee1b4f6dd65047762f924ecd367c17eabf8f[1] -> 1122867 (current: 0 / original: 0))
```

That is, the last `SSTORE` consumes too much gas.

pipermerriam:
I'm good with skipping the tests mentioned by @sorpaas (but lets be sure to add a beefy comment explaining why and linking at minimum to this thread).

veox:
> Just pushed [010b840](https://github.com/ethereum/py-evm/commit/010b840d533fc342154322730acbbcae5c0c4a40) which _in theory_ fixes a few of the remaining failures but causes some other tests to fail that were not failing previously...

I can not corroborate this, at least not when running `tox py36-native-state-constantinople`.

The diff between before commit 010b840 and after is:

```
<  test_state_fixtures[/home/veox/src/py-evm/fixtures/GeneralStateTests/stCreate2/create2collisionStorage.json:create2collisionStorage:Constantinople:0] 
<  test_state_fixtures[/home/veox/src/py-evm/fixtures/GeneralStateTests/stCreate2/create2collisionStorage.json:create2collisionStorage:Constantinople:1] 
<  test_state_fixtures[/home/veox/src/py-evm/fixtures/GeneralStateTests/stCreate2/create2collisionStorage.json:create2collisionStorage:Constantinople:2] 
```

(That is, three nodeIDs of a single test `create2collisionStorage` have been fixed.)

That said, my summary line (for the "post" run) is

```
============ 176 failed, 5142 passed, 30 skipped in 1561.03 seconds ============
```

..., whereas [the CircleCI run](https://circleci.com/gh/ethereum/py-evm/108904?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) shows

```
============= 10 failed, 2146 passed, 30 skipped in 58.59 seconds ==============
```

-----

I'll repeat the experiment - but running `pytest tests/json-fixtures/test_state.py --fork Constantinople`, like the CircleCI run, - and report here with the findings.

sorpaas:
@veox It may be still be that issue. Parity skipped two, one is `RevertInCreateInInit` and the other is `RevertInCreateInInitCreate2`.

If you want to confirm, check the gas cost for `SSTORE` in those two tests. For those artificial items accessed, geth/aleth acted as if the original value is zero, but parity/py-evm acted as if the original value is the artificial values set.

veox:
@sorpaas You are right - I didn't look at the link you posted, and so talked past you. Sorry.

In [same comment, second section](https://github.com/ethereum/py-evm/pull/1181#issuecomment-446330609) I confirm what you're saying.

veox:
I've fixed the conflicts and continued in PR #1579.

PR #1183, opened 2018-08-14
Fix full sync service cancellation detection

pipermerriam:
### What was wrong?

The previous fix for this in #1177 did not take the async nature into account when detecting whether cancellation has been requested.

### How was it fixed?

Added a `self.sleep(0)` before the `self.is_running` check.

#### Cute Animal Picture

android-animals-hed2-2015

### DIFF
--- a/trinity/sync/full/service.py
+++ b/trinity/sync/full/service.py
@@ -47,6 +47,7 @@ class FullNodeSyncer(BaseService):
                 self.chain, self.chaindb, self.peer_pool, self.cancel_token)
             await chain_syncer.run()
 
+        await self.sleep(0)
         if not self.is_running:
             return
 
@@ -59,6 +60,7 @@ class FullNodeSyncer(BaseService):
                 self.chaindb, self.base_db, head.state_root, self.peer_pool, self.cancel_token)
             await downloader.run()
 
+        await self.sleep(0)
         if not self.is_running:
             return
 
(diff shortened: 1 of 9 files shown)


trinity/sync/full/service.py
```diff
@@ -47,6 +47,7 @@ def __init__(self,
                 self.chain, self.chaindb, self.peer_pool, self.cancel_token)
             await chain_syncer.run()
 
+        await self.sleep(0)
```
gsalgado: I don't think this will have the effect you want. This code is inside `_run()` and `BaseService.is_running` will return True as long as `_run()` is executing. I think the correct thing to do here is to check if the token has been triggered instead of checking for `self.is_running`

trinity/sync/full/service.py
```diff
@@ -47,7 +47,8 @@ def __init__(self,
                 self.chain, self.chaindb, self.peer_pool, self.cancel_token)
             await chain_syncer.run()
 
-        if not self.is_running:
+        await self.sleep(0)
```
gsalgado: The sleep() is not needed here

trinity/sync/full/service.py
```diff
@@ -59,7 +60,8 @@ def __init__(self,
                 self.chaindb, self.base_db, head.state_root, self.peer_pool, self.cancel_token)
             await downloader.run()
 
-        if not self.is_running:
+        await self.sleep(0)
+        if self.cancel_token.triggered:
```
gsalgado: Nor here

PR #1184, opened 2018-08-15
Reducing responsibility of BaseRequestManager

carver:
### What was wrong?

- Logic for a new request/response cycle was spread around
- BaseRequestManager had a lot of responsibilities and mutual dependencies with subclasses

### How was it fixed?

A lot going on here, so I'll try to summarize. Everything was kind of connected, so I'm sorry this came in one big dump.

- Consistently refer to `Message` as the decoded data that the p2p protocol returned, and `Result` as the normalized data that is finally returned
- Removed the `BaseRequestManager` in favor of a `MessageManager` which is responsible for returning a stream of messages that match the command of interest
- Added an `ExchangeManager` (where `Exchange` is the IETF word for a paired Request/Response) which is responsible for validating potential messages from the `MessageManager` and handing back the normalized result
- The `Handler` now uses `Exchange` objects
- `Exchange` is meant to be a very small class that ties things together: a normalizing component, a validation component, a `BaseRequest`, the `ExchangeManager`, and other data munging (more details below)
- Changed the `BaseRequest` to only track: the outgoing command type, the response command type, and the parameters that p2p needs to send
- Added the ability to send a command to a peer, with only a `BaseRequest` instance
- Reduced the `Validator` classes so they *only* validate the final result
- Added option for custom lightweight message validator (eg~ for validating LES)
- Updated all the currently implemented commands to work this way

The most important piece to understand when implementing a new command is the `Exchange`. It ties together everything new you need. Here are two annotated examples of `Exchange`, and some related definitions:

```py
class BaseExchange(ABC, Generic[TCommandPayload, TMsg, TResult]):
    """
    TCommandPayload is the data as passed directly to the p2p command
    TMsg is the data as received directly from the p2p command response
    TResult is the response data after normalization
    """
    def __init__(self, manager: ExchangeManager[TCommandPayload, TMsg, TResult]) -> None:
        self._manager = manager
    ...


# This is a core example of an exchange. It will be exposed directly on `peer.requests.get_node_data`
class GetNodeDataExchange(BaseExchange[Tuple[Hash32, ...], Tuple[bytes, ...], NodeDataBundles]):
    _normalizer = GetNodeDataNormalizer()  # the normalizer is responsible for message->result conversion (and counting the number of results)

    # this is the core await-able call, which returns the normalized result
    async def __call__(self, node_hashes: Tuple[Hash32, ...]) -> NodeDataBundles:

        # prepare to validate the result by saving a copy of the request arguments
        validator = GetNodeDataValidator(node_hashes)

        # form the request, which also knows its own command type, and response command type
        request = GetNodeDataRequest(node_hashes)

        # This is a convenience method on the super-class. It mostly just calls out to the 
        # exchange manager, which waits for a result that validates, and returns the normalized version
        return await self.get_result(request, self._normalizer, validator)


#
# The following helper function and exchange shows how to validate the underlying message
#

def _match_message_request_id(payload: Dict[str, Any], message: Dict[str, Any]) -> None:
    if payload['request_id'] != message['request_id']:
        raise ValidationError("Request `id` does not match")


class GetBlockHeadersExchange(BaseExchange[Dict[str, Any], Dict[str, Any], Tuple[BlockHeader, ...]]):
    _normalizer = BlockHeadersNormalizer()
    async def __call__(
            self,
            block_number_or_hash: BlockIdentifier,
            max_headers: int = None,
            skip: int = 0,
            reverse: bool = True) -> Tuple[BlockHeader, ...]:

        validator = GetBlockHeadersValidator(block_number_or_hash, max_headers, skip, reverse)

        # add the request ID so that the request object has everything it needs
        command_args = (block_number_or_hash, max_headers, skip, reverse, gen_request_id())

        request = GetBlockHeadersRequest(*command_args)

        # Wait for a result that matches *both* the message and result validators
        return await self.get_result(
            request,
            self._normalizer,
            validator,
            # the superclass passes the request command arguments to the msg validator:
            _match_message_request_id,  
        )
```

Still TODO, probably in a different PR:
- Not many docs besides this writeup right now :/

#### Cute Animal Picture


trinity/protocol/les/managers.py
```diff
@@ -41,36 +46,58 @@
...
-            request_id,
+            gen_request_id(),
         )
-        return await self._request_and_wait(request, timeout)
+        make_request = partial(self._peer.sub_proto.send_get_block_headers, *request_args)
+        validator = HeaderRequest(*request_args)
+        manager = self._request_manager
+
+        async for message in manager.message_candidates(make_request, self.logger, timeout):
+            async with disconnect_on_misbehavior(self._peer, BlockHeaders, self.logger):
```
cburgdorf: async funkyness! I like :heart:

trinity/protocol/common/managers.py
```diff
@@ -64,6 +70,23 @@ def add(self, time: float, size: int) -> None:
...
 
 
+@asynccontextmanager
+async def disconnect_on_misbehavior(
+        peer: BasePeer,
+        command_class: Type[Command],
+        logger: logging.Logger) -> AsyncGenerator[None, None]:
+    try:
+        yield
+    except MalformedMessage as err:
```
gsalgado: I was surprised to see a MalformedMessage being handled at this level, as I thought that was usually handled at a lower level, closer to where it's actually raised (the decoding of messages on `BasePeer.read_msg()`), and then I found out the manager's `_normalize_response()` methods raise that as well, but that doesn't sound right as that exception should be used to signal a msg we can't decode (`Raised when a p2p command is received with a malformed message`). How would you feel about using a separate exception to indicate an unexpected response at the managers level, @pipermerriam?

pipermerriam: :+1: the use of `MalformedMessage` here isn't appropriate, it was just convenient and felt better than `ValidationError` at the time.

Also worth pointing out.  This exception handling should be extremely unlikely and in the event that it is encountered, it is more likely to be a problem with our code than a problem with the peer since the decoded messages have a fixed and known format since they come directly from the decoded command message.  

IE, if we expected a tuple of block headers and we find we have some other type, something is probably wrong in our code that is allowing us to pass the wrong value into this API.

tests/trinity/core/p2p-proto/test_peer_block_header_request_and_response_api.py
```diff
@@ -94,11 +95,13 @@ def mk_header_chain(length):
 @pytest.mark.asyncio
 async def test_les_peer_get_headers_round_trip(les_peer_and_remote,
                                                params,
+                                               monkeypatch,
                                                headers):
     peer, remote = les_peer_and_remote
+    request_id = 9
+    monkeypatch.setattr(trinity.protocol.les.managers, 'gen_request_id', lambda: request_id)
```
pipermerriam: I just realized that we should be able to get rid of this monkeypatching

In the `send_headers` function we can read the `GetBlockHeaders` request from the `remote` peer and extract the `request_id` from the message itself.  Take a look in the peer `_run` loop for the message handler.  This is probably easiest done using the `with peer.collect_sub_proto_messages` API.

trinity/protocol/common/managers.py
```diff
@@ -214,3 +237,129 @@ def _send_sub_proto_request(self, request: TRequest) -> None:
...
+            self,
+            make_request: Callable[[], None],
+            logger: logging.Logger,
+            timeout: int) -> 'AsyncGenerator[TMsg, None]':
+        """
+        Make a request and iterate through candidates for a valid response.
+
+        To mark a response as valid, use the `complete_if_valid` context manager. Within it,
+        validate the received message. If no ValidationError is raised, then the
+        message will be considered delivered, and the response closed.
```
pipermerriam: I read this twice and I am still not sure what this is telling me to do.  Maybe we could include a code example in the docstring?

trinity/protocol/common/managers.py
```diff
@@ -214,3 +237,129 @@ def _send_sub_proto_request(self, request: TRequest) -> None:
...
+        self.logger.debug("Launching %s for peer %s", self.__class__.__name__, self._peer)
+
+        with self.subscribe_peer(self._peer):
+            while self.is_running:
+                peer, cmd, msg = await self.wait(self.msg_queue.get())
+                if peer != self._peer:
+                    self.logger.error("Unexpected peer: %s  expected: %s", peer, self._peer)
+                    continue
+                elif isinstance(cmd, self._response_msg_type):
+                    await self._handle_msg(cast(TMsg, msg))
```
pipermerriam: The new task API says we should do this using `self.run_task` so that this loop doesn't get blocked by waiting for the message handling to complete.

trinity/protocol/common/managers.py
```diff
@@ -214,3 +237,129 @@ def _send_sub_proto_request(self, request: TRequest) -> None:
...
+        return message
+
+    def _request(self, make_request: Callable[[], None]) -> None:
+        if self.pending_request is not None:
+            self.logger.error(
+                "Already waiting for response to %s for peer: %s",
+                self.response_msg_name,
+                self._peer,
+            )
+            raise AlreadyWaiting(
```
pipermerriam: orthogonal to this pull request, but I've been puzzling over how we can eliminate this failure condition with either a `asyncio.Lock` or `asyncio.Condition` to allow call sites to *aquire* the right to issue a request, and *then* issue it with a strong guarantee that the only errors they are likely to encounter are a `TimeoutError`.  I haven't actually tried anything but I wanted to put that out there in case anyone has any bright ideas.

trinity/protocol/les/managers.py
```diff
@@ -41,36 +46,58 @@
...
+                result = self._normalize_result(message)
+
+            item_count = len(message['headers'])
+
+            with manager.complete_if_valid(item_count):
```
pipermerriam: I am currently :-1: on this API.  It feels very *magic*.  Basic `try/except` handling here seems like it would be significantly easier to read/understand since it should be more explicit, but it shouldn't be that much extra code.

trinity/protocol/common/managers.py
```diff
@@ -214,3 +237,129 @@ def _send_sub_proto_request(self, request: TRequest) -> None:
...
+    async def _handle_msg(self, msg: TMsg) -> None:
+        if self.pending_request is None:
+            self.logger.debug(
+                "Got unexpected %s message from %", self.response_msg_name, self._peer
+            )
+            return
+
+        send_time, future = self.pending_request
+        self.last_response_time = time.perf_counter() - send_time
+        future.set_result(msg)
```
pipermerriam: Do I understand correctly that rather than only setting the result if we find a match, we instead set the result for every message of the correct type and then later we validate the result that was set, and if it doesn't validate we initialize a new future and start the process over?

trinity/protocol/eth/managers.py
```diff
@@ -57,38 +67,47 @@
...
+            with manager.complete_if_valid(len(message)):
+                validator.validate_result(message)
+                return message
 
-    def _get_item_count(self, msg: Tuple[BlockHeader, ...]) -> int:
-        return len(msg)
+        raise ValidationError("Manager is not pending a response, but no valid response received")
+
+    async def _run(self) -> None:
+        await self._request_manager.run()
```
pipermerriam: It'd be nice if we could get rid of the extra service layer here.  See my other comment about removing this class entirely for how we might do that.

trinity/protocol/eth/managers.py
```diff
@@ -57,38 +67,47 @@
...
-    # All `RequestManager` classes are expected to implement the `__call__`
-    # method, including changing the function signature, thus the
-    # `# type: ignore` here is both expected and required.
-    async def __call__(self,  # type: ignore
+    def __init__(self, peer: 'ETHPeer', token: CancelToken) -> None:
+        super().__init__(token)
+        self._peer = peer
+        self._request_manager = BlockHeadersRequestManager(peer, BlockHeaders, token)
+
+    async def __call__(self,
```
pipermerriam: So I think we can get rid of this class entirely by moving this functionality down into the `Handler` class, and rather than having `get_block_headers` be a class with a `__call__` we could just make it an actual method which handles this orchestration.  The `Handler` class would then become responsible for running the `_request_manager` which has the nice effect of removing this extra layer of services.

trinity/protocol/eth/managers.py
```diff
@@ -57,38 +67,47 @@
...
+        manager = self._request_manager
 
-    async def _normalize_response(self,
-                                  msg: Tuple[BlockHeader, ...],
-                                  ) -> Tuple[BlockHeader, ...]:
-        return msg
+        async for message in manager.message_candidates(make_request, self.logger, timeout):
+            with manager.complete_if_valid(len(message)):
+                validator.validate_result(message)
+                return message
```
pipermerriam: This block of context managers and iterators will likely look the same for all command pairs and will need to be repeated in all command pairs.  This strikes me as an un-necessary amount of boilerplate or a violation of DRY or something like that.  I'd like to figure out a way to reduce this down to something like:

```python
return await manager.request_and_wait(make_request, self.logger, timeout, validator)
```

pipermerriam:
:+1: on direction.  There are some things that don't feel intuitive to me (which I commented on) but they all seem reasonably easy to cleanup/fix/whatever.

trinity/protocol/eth/managers.py
carver: In this case, the two manager loops that are implemented are actually quite different. If you try to capture all the differences in components, it starts to look worse than just the loop (to me), like:

```py
return await manager.request_and_wait(make_request, normalizer, item_counter, validator, self.logger, timeout)
```

I'll play around with some options, and implement a few more managers to see which pieces are really repeated.

trinity/protocol/common/managers.py
carver: I think that would be unsafe right now (I think we assume we don't handle a second message before the first is checked for validation), but I'll look how hard it would be to make it safe.

trinity/protocol/common/managers.py
carver: Yes

trinity/protocol/common/managers.py
carver: Yeah, on further inspection, this check appears entirely redundant. The `Command` enforces the correct type when decoding the RLP. I think we can reasonably get rid of the `MalformedMessage` part of the normalization step, or at least drop the bit that disconnects from the peer.

trinity/protocol/eth/managers.py
carver: The first roadblock I ran into is that the normalization step sometimes needs to run in a process pool, so it's easiest to call that from inside a service.

trinity/protocol/common/managers.py
pipermerriam: Leaving it this way is fine by me.  For message loops like this that only process a single message type I think this should be fine.

tests/trinity/core/p2p-proto/test_peer_block_header_request_and_response_api.py
carver: Hm, I thought I could just `await remote.read_msg()` but I'm getting a decryption error. I don't want to chase that down right now.

carver:
BTW, I used the word `Exchange` to refer to a `request/response` pair, which is apparently the word used to describe that pair in the HTTP spec.

Ok, I want to take one last pass at this tomorrow, and then I'm cutting myself off (or rather, submitting for final review :P). It will mostly be focused on how to do a similar simplification as currently, but in a way that doesn't look insane when we make mypy happy.

carver:
@cburgdorf any idea what's happening with mypy here? It *seems* to be reporting errors on lines that don't even exist :ghost: 


```
lint-py36 runtests: commands[6] | mypy --follow-imports=silent --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p trinity
trinity/protocol/common/managers.py:90: error: Name 'AsyncGenerator' is not defined
trinity/nodes/mainnet.py:53: error: Type argument "TCommandPayload`1" of "BaseRequest" must be a subtype of "Union[builtins.dict[builtins.str, Any], builtins.list[Any], builtins.tuple[Any]]"
trinity/nodes/mainnet.py:202: error: Type argument "TCommandPayload`1" of "BaseRequest" must be a subtype of "Union[builtins.dict[builtins.str, Any], builtins.list[Any], builtins.tuple[Any]]"
```

p2p/protocol.py
```diff
@@ -123,6 +126,27 @@ def encode(self, data: _DecodedMsgType) -> Tuple[bytes, bytes]:
         return header, body
 
 
+TCommandPayload = TypeVar('TCommandPayload', bound=_DecodedMsgType)
```
pipermerriam: With this class named `Request` the term payload implies something that we are delivering.  However, I believe that `TCommandPayload` is referring to the response payload.  Thoughts on naming change?

- `TResponsePayload`?

p2p/protocol.py
```diff
@@ -135,11 +159,20 @@ def __init__(self, peer: 'BasePeer', cmd_id_offset: int) -> None:
         self.peer = peer
         self.cmd_id_offset = cmd_id_offset
         self.commands = [cmd_class(cmd_id_offset) for cmd_class in self._commands]
+        self.cmd_by_type = {cmd_class: cmd_class(cmd_id_offset) for cmd_class in self._commands}
         self.cmd_by_id = dict((cmd.cmd_id, cmd) for cmd in self.commands)
 
     def send(self, header: bytes, body: bytes) -> None:
         self.peer.send(header, body)
 
+    def send_request(self, request: BaseRequest[_DecodedMsgType]) -> None:
```
pipermerriam: I :heart: this

tests/trinity/core/p2p-proto/test_block_bodies_request_object.py
```diff
@@ -77,32 +77,29 @@ def mk_headers(*counts):
 def test_block_bodies_request_empty_response_is_valid():
     headers_bundle = mk_headers((2, 3), (8, 4), (0, 1), (0, 0))
     headers, _, _, _, _ = zip(*headers_bundle)
-    request = BlockBodiesRequest(headers)
-    request.validate_response(tuple(), tuple())
+    request = GetBlockBodiesValidator(headers)
```
pipermerriam: maybe change the variable name in these to `validator`?

tests/trinity/core/p2p-proto/test_node_data_request_object.py
```diff
@@ -26,55 +26,51 @@ def mk_node_data(n):
...
-    request.validate_response(nodes[:3], node_data[:3])
-    request.validate_response(
-        (nodes[1], nodes[8], nodes[4]),
-        (node_data[1], node_data[8], node_data[4]),
-    )
+    request.validate_result(node_data[3:])
+
+    request.validate_result(node_data[:3])
+
+    request.validate_result((node_data[1], node_data[8], node_data[4]))
```
pipermerriam: Same here with renaming to `validator`

tests/trinity/core/p2p-proto/test_peer_block_header_validator_api.py
```diff
@@ -94,11 +95,13 @@ def mk_header_chain(length):
 @pytest.mark.asyncio
 async def test_les_peer_get_headers_round_trip(les_peer_and_remote,
                                                params,
+                                               monkeypatch,
                                                headers):
     peer, remote = les_peer_and_remote
+    request_id = 9
+    monkeypatch.setattr(trinity.protocol.les.exchanges, 'gen_request_id', lambda: request_id)
```
pipermerriam: If you go back in to make changes, can you add a `TODO` comment here indicating that we *should* be able to retrieve the `request_id` by reading the request message from the peer.

cburgdorf:
@carver 

>any idea what's happening with mypy here?

The first line is just a missing import of `AsyncGenerator` but the other two lines are really strange. As you said, the lines don't even exist. Clearly, `mypy` seems to be concerned that somewhere the type argument to `BaseRequest` does not qualify as a subtype of `Union[builtins.dict[builtins.str, Any], builtins.list[Any], builtins.tuple[Any]]`.

I spend half an hour trying to find the cause of it but without success so far. I think that the wrong line reporting qualifies as a bug in `mypy` so maybe we should submit it to them. They are usually pretty fast to respond.

trinity/protocol/eth/requests.py
```diff
@@ -7,25 +8,29 @@
...
+    Receipts,
+)
 
 
-class HeaderRequest(BaseHeaderRequest[Tuple[BlockHeader, ...]]):
+class HeaderRequest(BaseHeaderRequest):
     @property
     def max_size(self) -> int:
-        return constants.MAX_HEADERS_FETCH
+        return MAX_HEADERS_FETCH
```
pipermerriam: If you come back through here this could just be turned into a normal attribute of the class.

trinity/protocol/eth/validators.py
```diff
@@ -0,0 +1,107 @@
...
+)
+
+from . import constants
+
+
+class GetBlockHeadersValidator(BaseBlockHeadersValidator):
+    protocol_max_request_size = constants.MAX_HEADERS_FETCH
+
+
+NodeDataBundles = Tuple[Tuple[Hash32, bytes], ...]
```
pipermerriam: These should probably be part of the shared ones from `trinity.protocol.common.types` since the `les` protocol has matching commands.

trinity/protocol/les/exchanges.py
```diff
@@ -0,0 +1,63 @@
...
+)
+from .requests import (
+    GetBlockHeadersRequest,
+)
+from .validators import (
+    GetBlockHeadersValidator,
+)
+
+if TYPE_CHECKING:
+    from .peer import LESPeer  # noqa: #401
```
pipermerriam: This doesn't appear to be used.

trinity/protocol/common/types.py
```diff
@@ -0,0 +1,36 @@
...
+
+from trinity.rlp.block_body import BlockBody
+
+TPeer = TypeVar('TPeer', bound=BasePeer)
+
+# A payload ready to send with a peer command
+TCommandPayload = TypeVar('TCommandPayload')
+
+# A payload delivered by a responding command
+TMsg = TypeVar('TMsg')
```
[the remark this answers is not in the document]
carver: Great, I like Payload better than Message too. The outgoing payload should probably change names a bit too, to distinguish from the response. Something like: `TRequestPayload` and `TResponsePayload`.

p2p/protocol.py
carver: This is actually the outgoing one, but we'll end up with much clearer `TRequestPayload` and `TResponsePayload`.

tests/trinity/core/p2p-proto/test_peer_block_header_validator_api.py
carver: In the process of writing the `TODO` I figured out what I was doing wrong :man_facepalming: (The peer already awaits the commands, I just need to subscribe to the decoded messages). Fixed.

carver:
Yay again @cburgdorf for the integration test that caught a mainnet-only bug :D

PR #1186, opened 2018-08-15
Implement CREATE2 Opcode

cburgdorf:
### What was wrong?

We don't support the `CREATE2` Opcode for Constantinople yet.

### How was it fixed?

Implemented `CREATE2` Opcode based on the existing `CREATE` opcode. Broken down:

- address format is [latest and greatest from last core dev call](https://github.com/ethereum/EIPs/pull/1014#issuecomment-412106601)
- reading items from the stack was abstracted into method `get_stack_data` to allow overwriting and hence read with/without `salt`
- generating address was abstracted into method `generate_contract_address` to allow overwriting and hence implement the different version format

### Known issues

- not tested yet

#### Cute Animal Picture


eth/utils/address.py
```diff
@@ -1,13 +1,29 @@
...
+    return Address(keccak(rlp.encode([address, nonce]))[-20:])
 
 
-def generate_contract_address(address: bytes, nonce: bytes) -> bytes:
-    return keccak(rlp.encode([address, nonce]))[-20:]
+def generate_safe_contract_address(address: Address,
+                                   salt: int,
+                                   call_data: bytes) -> Address:
+    return Address(
+        keccak(decode_hex('0xff') + address + int_to_big_endian(salt) + keccak(call_data))[-20:]
```
pipermerriam: This might be easier to read (and slightly more cleanly typed as) as:

```python
return force_bytes_to_address(keccak(
    decode_hex('0xff') + address + int_to_big_endian(salt) + keccak(call_data)
))
```

Also, the `decode_hex('0xff')` should probably be replaced with a constant `CREATE2_PREFIX` or just a bytes literal.

cburgdorf: Right! Changing it to:

```python
    return force_bytes_to_address(
        keccak(b'\xff' + address + int_to_big_endian(salt) + keccak(call_data))
    )
```

I'll also change `generate_contract_address` to use `force_bytes_to_address` to be aligned with this change.

PR #1188, opened 2018-08-16
unpin web3 version and use a range instead

voith:
### What was wrong?

web3 version was pinned to a specific version in #957. Although pinning is good for `py-evm`'s development, but this causes problems in projects that are dependent on `py-evm`. 
eg. the latest web3 release is `4.5.0`, but updating web3 elsewhere causes version conflicts as py-evm is pinned to `4.4.1`.
IMO, pinning a  dependency in a library that is widely used by others is never a  good solution. 

### How was it fixed?
`web3` version was unpinned and a range was used instead.
cc @cburgdorf 

#### Cute Animal Picture


cburgdorf:
>IMO, pinning a dependency in a library that is widely used by others is never a good solution.

I agree. That said, how would we resolve cases like the one described in #957? Would it also be possible to explicitly blacklist (known) broken versions? Basically, specifying a range + excludes?

voith:
> Basically, specifying a range + excludes?

`setup.py` supports `!=`.

I just tested the following:
`"web3>=4.3.0,!=4.5.0,<5.0.0"`
This installed `4.4.1`. Without `!=`, it installs `4.5.0`

Found this [here](https://stackoverflow.com/questions/29499542/exclude-certain-dependency-version-ranges-in-setuptools-pip).

voith:
@pipermerriam Installing `py-evm` in a fresh virtualenv also installs `web3.py`
`pip install "py-evm==0.2.0-alpha.30"`
```
pip freeze | grep web3
web3==4.4.1
```
```
pipdeptree --reverse --packages web3
web3==4.4.1
  - py-evm==0.2.0a30 [requires: web3==4.4.1]
```
https://github.com/ethereum/py-evm/blob/038d17176ddbd86dc3c7ca3d41d4b64831c42bf8/setup.py#L96
if you don't think that updating `web3` is the correct solution then trinity needs to be removed from install requires.

pipermerriam:
Ah, yes, this should be fixed.

https://github.com/ethereum/py-evm/blob/038d17176ddbd86dc3c7ca3d41d4b64831c42bf8/setup.py#L96

I suspect we did this to make it easier to install and get going, but the `deps['p2p'` and `deps['trinity']` should not be part of install requires anymore.

Issue #679, opened 2018-05-11
p2p.Server must ensure PeerPool is always running


gsalgado:
Now that [PeerPool no longer restarts itself](https://github.com/ethereum/py-evm/pull/678), we need to make sure`p2p.Server` periodically check if it's still running, and possibly restart it if not. This is needed because p2p.Server runs the PeerPool in the background (with `asyncio.ensure_future()`)

gsalgado:
Another alternative (which might make more sense, actually) would be to use something like `asyncio.wait()` to wait for either `FullNodeSyncer.run()` or `PeerPool.run()` to finish, and try to restart whichever one returned. Need to think more about this

PR #1189, opened 2018-08-16
BaseService now terminates itself if a daemon dies unexpectedly

gsalgado:
If a daemon finishes while its parent is still running, the parent terminates itself.

Closes: #679

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -8,6 +8,7 @@ from typing import (
     Callable,
     List,
     Optional,
+    Set,
     cast,
 )
 from weakref import WeakSet
@@ -31,7 +32,7 @@ class ServiceEvents:
 
 class BaseService(ABC, CancellableMixin):
     logger: TraceLogger = None
-    _child_services: List['BaseService']
+    _child_services: Set['BaseService']
     # Use a WeakSet so that we don't have to bother updating it when tasks finish.
     _tasks: 'WeakSet[asyncio.Future[Any]]'
     _finished_callbacks: List[Callable[['BaseService'], None]]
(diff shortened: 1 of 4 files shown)


p2p/service.py
```diff
@@ -136,9 +142,34 @@ def run_task(self, awaitable: Awaitable[Any]) -> None:
...
+        self.run_task(f())
+
+    async def _restart_finished_children(self) -> None:
+        restart_times: Dict[BaseService, List[float]] = {}
+        while self.is_running:
+            child = await self.wait(self._finished_child_services.get())
+            child_restart_times = restart_times.setdefault(child, [])
+            one_min_ago = time.time() - 60
+            if child_restart_times and max(child_restart_times) > one_min_ago:
+                self.logger.warning(
```
gsalgado: It might make sense to `self.cancel()` here, as in most cases I believe a dead child will prevent the parent service from operating at all.

gsalgado:
Yeah, I understand your concern. I think maybe we could restart any given child at most once (with a big fat warning), and then cancel() the parent if it dies again. That way we are a bit more resilient without sweeping real bugs under the rug. I don't have strong feelings about that, though, so happy to change this to simply cancel the parent if any child dies unexpectedly

pipermerriam:
I am going to advocate for full on crashing if any child service crashes.  At this point the stability of the client isn't critical and I'd rather us find these sources of crashes quickly because they take down the whole node than to quietly miss them because they are incredibly intermittent and thus rarely encountered.

carver:
Also, it seems like too strong an assumption to make that any `run_child_service()` call wants that child service to run forever.

pipermerriam:
Hrm, so we need 

- `run_child_service`  (allowed to exit)
- `monitor_child_service` (cascades to local failure if child fails)

carver:
Which sounds just like the conversation we had here (except that was about async methods instead of services): https://github.com/ethereum/py-evm/issues/1118

Maybe `run_daemon()` is the right name for starting a service that should be restarted if it dies or ends.

gsalgado:
Added a `run_daemon()` method and the parent now restarts if any child service executed with that dies

PR #1192, opened 2018-08-17
Make 'persist_header' accept batches

cburgdorf:
### What was wrong?

Noticed [this comment](https://github.com/ethereum/py-evm/pull/1151#discussion_r210757360) and jumped on it.

The `persist_header` API did not allow updating batches. By doing so, we can save some time as there is some work inside that method that can be shared across multiple updates.

### How was it fixed?

- make the method accept a `Union[BlockHeader, Iterable[BlockHeader]]` and internally wrap single values in a list
- update call site in light change syncer to take advantage.

Btw, I think there's a further optimization that we could do. Here's a call site:

https://github.com/ethereum/py-evm/blob/cde40dbd953186aafc2832d915261089ec589002/trinity/sync/light/chain.py#L67-L70

Seems like we are fetching the canonical head right after the usage of `persist_header`. However, we know this internally and could change the return type from `Tuple[BlockHeader, ...]` to `Tuple[BlockHeader, Tuple[BlockHeader, ...]]` (or use a named tuple for readability). Then we could directly use the canonical head from the return instead of doing a second call. Thoughts?

#### Cute Animal Picture

### DIFF
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -34,12 +34,9 @@ from eth_hash.auto import keccak
 
 from eth.constants import (
     EMPTY_UNCLE_HASH,
-    GENESIS_PARENT_HASH,
 )
 from eth.exceptions import (
-    CanonicalHeadNotFound,
     HeaderNotFound,
-    ParentNotFound,
     TransactionNotFound,
 )
 from eth.db.header import BaseHeaderDB, HeaderDB
@@ -186,63 +183,14 @@ class ChainDB(HeaderDB, BaseChainDB):
         else:
             return rlp.decode(encoded_uncles, sedes=rlp.sedes.CountableList(BlockHeader))
 
-    # TODO: This method should take a chain of headers as that's the most common use case
-    # and it'd be much faster than inserting each header individually.
-    def persist_header(self,
-                       header: BlockHeader
-                       ) -> Tuple[Tuple[BlockHeader, ...], Tuple[BlockHeader, ...]]:
-        """
-        Returns iterable of headers newly on the canonical chain
-        """
-        is_genesis = header.parent_hash == GENESIS_PARENT_HASH
-        if not is_genesis and not self.header_exists(header.parent_hash):
-            raise ParentNotFound(
-                "Cannot persist block header ({}) with unknown parent ({})".format(
-                    encode_hex(header.hash), encode_hex(header.parent_hash)))
-
-        self.db.set(
-            header.hash,
-            rlp.encode(header),
-        )
-
-        if is_genesis:
-            score = header.difficulty
-        else:
-            score = self.get_score(header.parent_hash) + header.difficulty
-
-        self.db.set(
-            SchemaV1.make_block_hash_to_score_lookup_key(header.hash),
-            rlp.encode(score, sedes=rlp.sedes.big_endian_int),
-        )
-
-        try:
-            head_score = self.get_score(self.get_canonical_head().hash)
-        except CanonicalHeadNotFound:
-            (
-                new_canonical_headers,
-                old_canonical_headers
-            ) = self._set_as_canonical_chain_head(header)
-        else:
-            if score > head_score:
-                (
-                    new_canonical_headers,
-                    old_canonical_headers
-                ) = self._set_as_canonical_chain_head(header)
-            else:
-                new_canonical_headers = tuple()
-                old_canonical_headers = tuple()
-
-        return new_canonical_headers, old_canonical_headers
-
-    # TODO: update this to take a `hash` rather than a full header object.
     def _set_as_canonical_chain_head(self,
-                                     header: BlockHeader
+                                     block_hash: Hash32
                                      ) -> Tuple[Tuple[BlockHeader, ...], Tuple[BlockHeader, ...]]:
         """
         Returns iterable of headers newly on the canonical head
         """
         try:
-            self.get_block_header_by_hash(header.hash)
+            header = self.get_block_header_by_hash(block_hash)
         except HeaderNotFound:
             raise ValueError("Cannot use unknown block hash as canonical head: {}".format(
                 header.hash))
--- a/eth/db/header.py
+++ b/eth/db/header.py
@@ -1,12 +1,18 @@
 from abc import ABC, abstractmethod
 import functools
-from typing import Tuple, Iterable
+from typing import Iterable, Tuple
 
 import rlp
+
+from cytoolz import (
+    first,
+    sliding_window,
+)
 
 from eth_utils import (
     encode_hex,
     to_tuple,
+    ValidationError,
 )
 
 from eth_typing import (
@@ -71,6 +77,12 @@ class BaseHeaderDB(ABC):
     def persist_header(self,
                        header: BlockHeader
                        ) -> Tuple[Tuple[BlockHeader, ...], Tuple[BlockHeader, ...]]:
+        raise NotImplementedError("ChainDB classes must implement this method")
+
+    @abstractmethod
+    def persist_header_chain(self,
+                             headers: Iterable[BlockHeader]
+                             ) -> Tuple[Tuple[BlockHeader, ...], Tuple[BlockHeader, ...]]:
         raise NotImplementedError("ChainDB classes must implement this method")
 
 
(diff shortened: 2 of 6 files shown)


eth/db/chain.py
```diff
@@ -184,32 +185,36 @@ def get_block_uncles(self, uncles_hash: Hash32) -> List[BlockHeader]:
         else:
             return rlp.decode(encoded_uncles, sedes=rlp.sedes.CountableList(BlockHeader))
 
-    # TODO: This method should take a chain of headers as that's the most common use case
-    # and it'd be much faster than inserting each header individually.
-    def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
+    def persist_header(
```
gsalgado: Should probably rename to `persist_chain()`, no? Also, why does it need to accept either a single header or an iterable?

eth/db/chain.py
```diff
@@ -184,32 +185,36 @@ def get_block_uncles(self, uncles_hash: Hash32) -> List[BlockHeader]:
...
-        )
+            self.db.set(
+                header.hash,
+                rlp.encode(header),
+            )
+
+            if is_genesis:
+                score = header.difficulty
+            else:
+                score = self.get_score(header.parent_hash) + header.difficulty
```
gsalgado: We can probably avoid this `get_score()` call now, as we're processing a chain of headers so we just need to get the score of the first item and then increment that with the difficulty of each subsequent item

eth/db/header.py
```diff
@@ -144,32 +146,38 @@ def header_exists(self, block_hash: Hash32) -> bool:
         validate_word(block_hash, title="Block Hash")
         return block_hash in self.db
 
-    def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
+    def persist_header(
```
gsalgado: Do you think we could refactor some of this code into a separate method that could then be reused in `ChainDB` as well?

cburgdorf:
Oh, actually I think I screwed something up there. Will check back later (afk)

eth/db/chain.py
```diff
@@ -184,32 +185,36 @@ def get_block_uncles(self, uncles_hash: Hash32) -> List[BlockHeader]:
         else:
             return rlp.decode(encoded_uncles, sedes=rlp.sedes.CountableList(BlockHeader))
 
-    # TODO: This method should take a chain of headers as that's the most common use case
-    # and it'd be much faster than inserting each header individually.
-    def persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
+    def persist_header(
+            self,
+            header: Union[BlockHeader, Iterable[BlockHeader]]) -> Tuple[BlockHeader, ...]:
```
pipermerriam: I would update this to be `*headers: Iterable[BlockHeader]` so that it can be called as either.

- `persist_headers(single_header)`
- `persist_headers(header_a, header_b, ...)`
- `persist_headers(*headers_to_be_persisted)`

Seems simplest in terms of typing and API.

eth/db/chain.py
pipermerriam: +1 to alternate name

- `persist_headers`
- `persist_header_chain` (my preference)

eth/db/chain.py
pipermerriam: We will need to do one of the following here:

- pre-validate that the iterable of headers is a contiguous chain.
- handle non-contiguous chains but still retain this optimizations for when it is a chain, i.e. detect when we are in a sequential list of headers and just increment the score, as well as detecting when we are non-sequential and lookup the score in this manner.

eth/db/header.py
pipermerriam: second to this (though lets not do this via inheritance, maybe just a helper utility function.

eth/db/header.py
```diff
@@ -146,46 +146,55 @@ def header_exists(self, block_hash: Hash32) -> bool:
...
+                rlp.encode(score, sedes=rlp.sedes.big_endian_int),
+            )
 
         try:
             head_score = self.get_score(self.get_canonical_head().hash)
         except CanonicalHeadNotFound:
-            new_canonical_headers, old_canonical_headers = self._set_as_canonical_chain_head(
-                header.hash,
-            )
+            (
```
cburgdorf: In case someones wondering why the diff looks more messy then expected with respect to the change, this is because I moved the method from the `ChainDB` here which was just slightly nicer

gsalgado:
AIUI, that's the guarantee we get from having every header in the chain contain a reference to the parent's hash, so you shouldn't need to perform any extra checks.

The `ChainDB` implementation of `persist_header()` is different from its parent, though. In `ChainDB` we need to update the lookup entries for transactions of the new canonical blocks whereas in `HeaderDB` we can't do that

cburgdorf:
>In `ChainDB` we need to update the lookup entries for transactions of the new canonical blocks whereas in `HeaderDB` we can't do that

I think, I partly see what you mean but not fully. 

So `ChainDB` we are removing the transactions for all the blocks of the header that do not belong to the canonical chain anymore. So it seems that the implementation of `_set_as_canonical_chain_head` in `ChainDB` and `HeaderDB` is indeed different.

This part should be in `ChainDB` but not in `HeaderDB`.

https://github.com/ethereum/py-evm/blob/dd304cc4ace18116c8f9fb4bbf9161319c9cb4c3/eth/db/chain.py#L262-L263

However, I don't see where we would update the lookup entries for the transactions for the *new* canonical headers. At least in the current implementation, it seems that we are only dealing with the headers and that `persist_header` of `ChainDB` and `HeaderDB` are identical.

Aren't the transaction hashes only relevant for `persist_block` (not `persist_header_chain`)?

eth/db/header.py
```diff
@@ -146,46 +146,55 @@ def header_exists(self, block_hash: Hash32) -> bool:
...
+        the second containing the old canonical headers
         """
-        if header.parent_hash != GENESIS_PARENT_HASH:
-            try:
-                self.get_block_header_by_hash(header.parent_hash)
-            except HeaderNotFound:
+
+        previous = None
+
+        for header in headers:
```
carver: I think we should check and error out if the headers are not chained. (before starting to do any database writes)

Something like:

```py
consecutive_pairs = sliding_window(2, headers)
if any(parent.hash != child.parent_hash for parent, child in consecutive_pairs):
  raise ValidationError(...)
```

gsalgado:
Right, I seemed to remember that `ChainDB.persist_header()` would add new tx lookup entries but indeed it just removes them in case some blocks are removed from the canonical chain -- it's `persist_block()` that adds the tx lookup entries for blocks that are in the canonical chain.

It looks like we can have a single `persist_header()` implementation, then, with `ChainDB` simply overwriting `_set_as_canonical_chain_head()`?

eth/db/header.py
```diff
@@ -146,46 +146,55 @@ def header_exists(self, block_hash: Hash32) -> bool:
...
+                rlp.encode(header),
+            )
 
-        if header.parent_hash == GENESIS_PARENT_HASH:
-            score = header.difficulty
-        else:
-            score = self.get_score(header.parent_hash) + header.difficulty
+            if is_genesis:
+                score = header.difficulty
+            elif previous is not None and previous[0].hash is header.parent_hash:
```
carver: If we add the validation above, we no longer have to confirm that the parent hash matches here. In fact, I think `previous` can just become `parent_score`, and generated before the loop, which removes this whole if/elif block.

tests/database/test_header_db.py
```diff
@@ -225,12 +224,12 @@ def test_headerdb_header_retrieval_by_hash(headerdb, genesis_header):
...
 
 def test_headerdb_canonical_header_retrieval_by_number(headerdb, genesis_header):
-    headerdb.persist_header(genesis_header)
+    headerdb.persist_header_chain(genesis_header)
 
     headers = mk_header_chain(genesis_header, length=10)
 
     for header in headers:
-        headerdb.persist_header(header)
+        headerdb.persist_header_chain(header)
```
carver: dedent and `headerdb.persist_header_chain(headers)`?

eth/db/header.py
cburgdorf: Yep, that was build with the goal to support non-contiguous chains as well https://github.com/ethereum/py-evm/pull/1192#discussion_r210939381

eth/db/header.py
carver: I read his comment as an either-or, and the first option is what I suggested. :)

I'm inclined to prefer that not-contiguous headers be added with `for h in headers: persist_header(h)`.

eth/db/header.py
carver: I'd say just pick the first option he suggested, and save non-contiguous support for another PR.

eth/db/header.py
pipermerriam: :+1: to only allowing contiguous chains and using `for h in headers: persist_header(header)` style for when we need to do non-contiguous chains.

cburgdorf: >'m inclined to prefer that not-contiguous headers be added with `for h in headers: persist_header(h).`

Yes, I think that makes sense :+1:

eth/db/header.py
```diff
@@ -149,43 +161,60 @@ def header_exists(self, block_hash: Hash32) -> bool:
...
+        the second containing the old canonical headers
         """
-        if header.parent_hash != GENESIS_PARENT_HASH:
-            try:
-                self.get_block_header_by_hash(header.parent_hash)
-            except HeaderNotFound:
+
+        try:
+            first_header = first(headers)
+        except StopIteration:
```
cburgdorf: I guess this could be avoided if we used a `Tuple` instead of accepting all `Iterable` but then, it strikes me as odd to restrict to a concrete type where an abstraction would do it.

eth/db/header.py
```diff
@@ -149,43 +161,60 @@ def header_exists(self, block_hash: Hash32) -> bool:
...
+            return tuple(), tuple()
```
cburgdorf: Also not sure if we should raise instead but in general nothing in/ nothing out strikes me as symmetric and sound

eth/db/header.py
```diff
@@ -149,43 +161,60 @@ def header_exists(self, block_hash: Hash32) -> bool:
...
+        else:
+
+            consecutive_pairs = sliding_window(2, headers)
+            if any(parent.hash != child.parent_hash for parent, child in consecutive_pairs):
+                raise ValidationError("Invalid headers. Headers need to be contiguous")
```
carver: Would be nice to print the consecutive parent and child headers, and the parent_hash that the child expected.

eth/db/header.py
```diff
@@ -294,7 +323,7 @@ class AsyncHeaderDB(HeaderDB):
     async def coro_get_canonical_block_hash(self, block_number: BlockNumber) -> Hash32:
         raise NotImplementedError()
 
-    async def coro_persist_header(self, header: BlockHeader) -> Tuple[BlockHeader, ...]:
+    async def coro_persist_header_chain(self, *headers: BlockHeader) -> Tuple[BlockHeader, ...]:
```
carver: I think this didn't get updated with the latest:

- lose the `*`
- keep the `coro_persist_header`

eth/db/header.py
cburgdorf: Yep, I just have a hard time accepting concrete types as params to an API if the API actually doesn't need that concrete type and could rather work with an abstraction. Another alternative would be to use `Sequence[BlockHeader]` as that allows for a simple `len()` check. But if you are fine with the `try` / `except` I' fine, too. :)

eth/db/header.py
cburgdorf: Yes, I rushed this. Had to decide about finishing this properly vs letting my 6yo waiting in front of the school in tears. Tough decision. I'll finish this before merging.

eth/db/header.py
carver: Personally, I think it's kinda cool to be able to something like:

```py
chaindb.persist_header_chain(reversed(backwards_headers))
```

which wouldn't work with `Sequence`.

eth/db/header.py
cburgdorf: ah yes, eagle eyes :)

Issue #1125, opened 2018-07-31
Unify `ValidationError`


pipermerriam:
### What is wrong?

https://github.com/ethereum/py-evm/pull/1089/files#r206404376

As noted in this comment, there is a valid concern that our duplicate `ValidationError` exceptions could easily be mixed up, causing us to not be catching the exception we think we are catching.

### How can it be fixed

Once https://github.com/ethereum/eth-utils/issues/114 has been done, modify the codebase to use that exception class for all `ValidationError` exceptions.

carver:
It could potentially go in `eth_utils.exceptions.ValidationError`.

pipermerriam:
https://github.com/ethereum/eth-utils/issues/114

glaksmono:
Would like to take on this once https://github.com/ethereum/eth-utils/issues/114 is done. Seems like there is already a PR on the issue @pipermerriam @carver

pipermerriam:
@glaksmono just released and published `eth-utils==1.1.1` which contains the new `ValidationError`.  Go for it :+1:

PR #1194, opened 2018-08-18
Unify `ValidationError`

glaksmono:
### What was wrong?

https://github.com/ethereum/py-evm/pull/1089/files#r206404376

As noted in this comment, there is a valid concern that our duplicate `ValidationError` exceptions could easily be mixed up, causing us to not be catching the exception we think we are catching.

Issue Reference: https://github.com/ethereum/py-evm/issues/1125

### How was it fixed?

Modify the codebase to use that exception class for all `ValidationError` exceptions to use the one from `eth_utils`

#### Cute Animal Picture


setup.py
```diff
@@ -10,7 +10,7 @@
         "eth-bloom>=1.0.0,<2.0.0",
         "eth-keys>=0.2.0b3,<1.0.0",
         "eth-typing>=1.1.0,<2.0.0",
-        "eth-utils>=1.0.1,<2.0.0",
+        "eth-utils==1.1.1",
```
cburgdorf: Why the pin to a specific version here?

eth/precompiles/ecmul.py
```diff
@@ -3,8 +3,10 @@
 )
 
 from eth import constants
-from eth.exceptions import (
+from eth_utils import (
```
pipermerriam: This should move up above the `eth` imports

trinity/sync/common/chain.py
```diff
@@ -14,13 +14,16 @@
 from eth.chains import AsyncChain
 from eth.exceptions import (
     HeaderNotFound,
-    ValidationError as EthValidationError,
+)
+from eth_utils import ValidationError as EthValidationError
```
pipermerriam: You can drop this rename and just change all of the places where `EthValidationError` is raised to just be a `ValidationError`.

Issue #1146, opened 2018-08-06
Refactor gas constants


cburgdorf:
### What is wrong?

I noticed that our gas constants are a bit messy and should be reorganized.

Let's take `GAS_BALANCE` as an example which is defined as such:

https://github.com/ethereum/py-evm/blob/bf1d3c2ad653b4ed6ab9ffb836735863bb5d8148/eth/constants.py#L43

For someone new to the code base this may seem as if retrieving the balance from some account has a fixed gas cost of `20` when in reality it is `400` as specified with the tangerine whistle fork.

https://github.com/ethereum/py-evm/blob/bf1d3c2ad653b4ed6ab9ffb836735863bb5d8148/eth/vm/forks/tangerine_whistle/constants.py#L13

### How can it be fixed

My gut feeling says that only *true constants*™ belong into `eth.constants` and that all *fork specific* constants should live in `evm.vm.forks.<fork_name>.constants`.

pipermerriam:
I'm :+1: on this.  I have been thinking similarly about our opcodes.  That all opcodes in `eth.vm.logic` should be moved into the `FrontierVM`.

cburgdorf:
>I have been thinking similarly about our opcodes. That all opcodes in `eth.vm.logic` should be moved into the `FrontierVM`

Mmh...I'm not so sure about that one. After all, these are logic functions that even private chains may want to use. We already *configure* the opcodes per fork. 

https://github.com/ethereum/py-evm/blob/bf1d3c2ad653b4ed6ab9ffb836735863bb5d8148/eth/vm/forks/constantinople/opcodes.py#L24-L40

But imagine you are implementing a private chain and want to use `add`, `sub`, `mul` etc and have to import it from `eth.vm.forks.frontier`. That would feel weird, no?

I can see the point though, chances are another fork wants to implement `mul` differently (unlikely but anyway) and so we end up with a different implementation of `mul`. But even then, I think this second `mul` implementation would just end up living in `eth.vm.logic` as long as it is *generic* enough to be used across different forks and chains.

pipermerriam:
We could potentially anchor them in *stub* fork like `eth.vm.forks.core` which contains this common stuff and is more *sane* for 3rd parties to import and use.

cburgdorf:
As in `eth.vm.forks.core` contains everything that forks may want to pull in or redefine. Sgtm.

mratsim:
[...]

PR #1195, opened 2018-08-18
[WIP] Refactor gas constants for the forks to use forks specific constants

glaksmono:
### What was wrong?

Our gas constants are a bit messy and should be reorganized.

Issue Reference: https://github.com/ethereum/py-evm/issues/1146

### How was it fixed?

Moving `GAS_BALANCE` to the fork specific constants

#### Cute Animal Picture


carver:
There seems to not be consensus on how/if a new structure is better than the status quo. I think that means we should close this PR and come back to it later if desired. (please reopen if I'm confused)

PR #1196, opened 2018-08-18
Use BaseService.run_daemon() where apropriate

gsalgado:


p2p/DEVELOPMENT.md
```diff
@@ -20,14 +20,17 @@ library.
...
+  unexpectedly.
 
 ```Python
 class Node(BaseService):
     async def _run(self):
         self.discovery = DiscoveryService(token=self.cancel_token)
-        self.run_child_service(self.discovery)
+        self.run_daemon(self.discovery)
         self.run_task(self.discovery.bootstrap())
+        self.run_child_service(UPnPService(token=self.cancel_token))
```
pipermerriam: Might be worth duplicating the comment about this being a bit unreliable and thus not being run as a daemon here as well.

trinity/server.py
```diff
@@ -151,8 +151,10 @@ def _make_peer_pool(self) -> PeerPool:
...
             discovery_proto, self.peer_pool, self.port, self.cancel_token)
-        self.run_child_service(self.peer_pool)
-        self.run_child_service(self.discovery)
+        self.run_daemon(self.peer_pool)
+        self.run_daemon(self.discovery)
+        # UPNP service is still experimental and not essential, so we don't use run_daemon() for
+        # it as that means if it crashes we'd be terminated as well.
         self.run_child_service(self.upnp_service)
         self.syncer = self._make_syncer(self.peer_pool)
         await self.syncer.run()
```
pipermerriam: it should be functionally equivalent but I wonder if it would be more intuitive if this was done as:

```python
self.run_daemon(self.syncer)
```

This seems to make it clearer that syncing is expected to run indefinitely.

gsalgado: But syncing is the main task of this service, so I'd find it weird to run it in the background and then wait for the cancel token to be triggered here. Also, by `await`ing for it here, we ensure any exceptions are propagated directly, whereas bugs in `run_daemon()` may cause them to be silently ignored otherwise

PR #1200, opened 2018-08-20
cleanup from changes to import_block

### LINKED ISSUE #733: Add logging output for chain reorgs
### What is wrong?

Chain re-orgs happen and it would be nice to see some info about them when running trinity.

### How can it be fixed

Whenever trinity is syncing and it encounters a chain reorg we should log something with depth and hashes.  Probably configurable to a certain minimum depth.

For lightchain, this should be relatively easy.

https://github.com/ethereum/py-evm/blob/master/p2p/lightchain.py#L214

Anywhere that `headerdb.coro_persist_header` is called should return an iterable of headers which are *new* to the canonical chain.  If the returned list contains more than 1 entry then the chain just re-organized.

Doing this for full sync is going to require updates to the Py-EVM API so that `chain.import_block` returns an iterable of blocks that are part of the new canonical chain.

pipermerriam:
### What was wrong?

fixes https://github.com/ethereum/py-evm/issues/733

https://github.com/ethereum/py-evm/pull/1151 did most of the heavy lifting but there were some things that needed cleaning up.

- the test case was written using loops which made it a bit hard to parse.
- the return value for block importing was inconsistent, sometimes returning the imported block in the new blocks section even if it wasn't part of the new chain.

### How was it fixed?

- rewrote the test to involve fewer loops and ensured that it used all fork rules for the mainnet VM
- changed `Chain.import_block` to return a 3-tuple
  - `(imported_block, new_canonical_headers, orphaned_canonical_headers)`

I also added logging output to the regular chain syncer which fixes #733 

#### Cute Animal Picture

cute-baby-animal-pictures-002-001


veox:
100% nitpicking: we all know what "orphan" means in this context, and that it's (semantically) incorrect; how about `new_{uncle,ommer}_blocks` instead of `orphaned_canonical_blocks`?..

(It was introduced in PR #1151.)

This might not be a problem now, but _might_ become one come Shasper, where "orphan" may correctly refer to X(n) for which a parent X(n-1) is not known.

See PR #257 for only other mention of term "orphan" in this repo (referencing a now-obsolete sharding implementation).

carver:
>  we all know what "orphan" means in this context, and that it's (semantically) incorrect

Ah, very interesting point. I had the same semantic concern, but `orphan` is used so commonly in other blockchain contexts to mean a non-canonical block, that it seemed reasonable to follow that trend.

> Shasper, where "orphan" may correctly refer to X(n) for which a parent X(n-1) is not known

Agree that since the semantically-appropriate name `orphan` could reasonably be used there, we should reserve the name for that use case.

> how about new_{uncle,ommer}_blocks instead of orphaned_canonical_blocks?

One problem is that `uncle` is used so heavily when the uncle header is included via GHOST. But at this point, we only know that the block is no longer on the canonical chain.

Maybe:
- `decanonicalized_blocks` (mouthful, and spelling problems?)
- `non_canonical_blocks`
- `ejected_blocks`
- `dropped_blocks`
- `loser_blocks`

eth/tools/fixture_tests.py
```diff
@@ -641,8 +641,7 @@ def apply_fixture_block_to_chain(block_fixture, chain):
 
     block = rlp.decode(block_fixture['rlp'], sedes=block_class)
 
-    imported_blocks, _ = chain.import_block(block)
-    mined_block = imported_blocks[-1]
```
carver: I think I like the new API better anyway, but is `mined_block` ever different from `imported_blocks[-1]`?

veox:
> One problem is that `uncle` is used so heavily when the uncle header is included via GHOST. But at this point, we only know that the block is no longer on the canonical chain.

Ah, right. A block becoming non-canonical doesn't mean it's been referenced as an uncle.

There's also the option of `{old,prev}_canonical_block`, similar to how it was called before...

pipermerriam:
I'm gonna go with `non_canonical_blocks`, it's explicit and shouldn't require much of any mental overhead to know what it is.

eth/tools/fixture_tests.py
pipermerriam: Yes, if the imported block is not part of the canonical chain then `imported_blocks` will be an empty tuple.

pipermerriam: Since call-sites should have an exact copy of `mined_block` I considered removing it, but didn't pull the trigger.  Should be easy to do later if we decide to slim the API down.

pipermerriam:
And then I ended up going with `old` because `non` and `new` were too visually similar

- `new_canonical_blocks`
- `non_canonical_blocks`
- `old_canonical_blocks`

Issue #1201, opened 2018-08-21
Convert "Fixture Builder" README to sphinx documentation.


pipermerriam:
### What is wrong?

The file at `eth/tools/fixture_builder/README.md` provides instructions for how to use the `fixture_builder` tools to write state tests.  This is both not easy to find and not the easiest to read docs.

### How can it be fixed

Convert these to be docs in the formal sphinx based docs.

pipermerriam:
Done in #1219

PR #1209, opened 2018-08-22
Chain builder tool

pipermerriam:
### What was wrong?

The `chain_with_block_validation` and `chain_without_block_validation` fixtures are used in a lot of places and are the *standard* way to muck with a chain.  They however are inflexible, requiring users of them to accept things like a common genesis state as well as what VM classes and configuration is used.

Similarly, tests which need to build up chain state prior to making an assertion are often lengthy and hard to read.

### How was it fixed?

A set of functional utilities which can be used to concisely construct chain state or apply procedural operations to a chain.  Still a WIP.

#### Cute Animal Picture


carver:
I think I agree with the overall premise, but this part is no longer true:

> requiring users of them to accept things like a common genesis state

With https://github.com/ethereum/py-evm/blob/master/tests/conftest.py#L151-L165 it's pretty easy to change the genesis state now. :)

> Similarly, tests which need to build up chain state prior to making an assertion are often lengthy and hard to read.

Totally. Ok, time to read the code...

pipermerriam:
> I think I agree with the overall premise, but this part is no longer true:

I consider that mechanism to be more of a hack which results in awkard *implicit* chain state you are correct that we do have a way to inject genesis state.

tests/core/builder-tools/test_chain_initializer.py
```diff
@@ -0,0 +1,118 @@
...
+
+ADDRESS_A = b'a' + b'\0' * 19
+ADDRESS_B = b'b' + b'\0' * 19
+
+
+def test_chain_builder_initialize_chain_with_state_simple(chain_class):
+    chain = pipe(
+        chain_class,
+        genesis(
+            state=((ADDRESS_A, 'balance', 1),),
```
carver: Maybe this would be cleaner as a tuple of objects representing accounts?

Something like:
```py
   chain = pipe(
        chain_class,
        genesis(
            state=(
                  Account(address=ADDRESS_A, balance=1),
            ),
        )
    )

class Account(typing.NamedTuple):
  address: eth_typing.Address
  balance: int = 0
  code: bytes = b''
  ...
```

_(TIL that `NamedTuple` [supports default values](https://docs.python.org/3/library/typing.html#typing.NamedTuple))_

pipermerriam: ooooooh! :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat: :smile_cat:

tests/core/builder-tools/test_chain_builder.py
```diff
@@ -0,0 +1,160 @@
...
+
+    chain = pipe(
+        pre_fork_chain,
+        fork_chain(
+            at_block_number(2),
+            mine_block(extra_data=b'fork-it!'),  # fork 3
+            mine_block(),  # fork 4
+            mine_block(),  # fork 5
+        ),
+    )
```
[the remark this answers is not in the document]
pipermerriam: So, this wasn't intentional but looking at this I think I understand why this way of expressing what should happen in the chain is so much easier to read.

The indentations provide a level of structure that literally gives it physical shape.  I haven't done transactions yet, but I'm thinking about this API.

```python
mine_block(
    txn(sender=account_0),
    txn(senter=account_1, to='0x0123...')
    txn(), # safe defaults
    extra_data=b'heres-some-extra-data',
)
```

It visually contains the transactions within the block.  Above the same thing with the `fork_chain`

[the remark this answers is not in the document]
pipermerriam: For forks, I want to do something like this to try and visually group the forks.

```python
chain = pipe(
    chain,
    mine_block(),
    chain_split(
        (mine_block(), mine_block(), mine_block()),  # fork A
        (mine_block(extra_data=b'...'), mine_block(), mine_block(), mine_block(), # fork B
        # results is a multidimensional array with 
        # the intermediate outputs for each fork
        converge_fn=lambda results: results[1][-1],
    )
    mine_block(), # resumes being *single_threaded*,
)
```

I don't love it but it's a starting point.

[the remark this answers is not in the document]
pipermerriam: Alright, here's a slightly alternate idea:

```python
manage_chain_a = pack(
    mine_block(),
    mine_block(
        tx(),
        tx(),
    ),
)

manage_chain_b = pack(
    mine_block(extra_data=b'...'),
    mine_block(),
    mine_block(),
)


def merge_chains(results):
    chain_a_results, chain_b_results = zip(*results)

    # do the import


def test_chain_split(chain):
    chain = pipe(
        chain,
        mine_block(),
        mine_block(),
        chain_split(
            manage_chain_a,
            manage_chain_b,
            exit_fn=merge_chains,
        )
        mine_block(tx(), tx(), tx()),
    )
```

The `chain_split` function takes `*args` and each of those can be thought of as a thread (but not an actual thread).  For each function provided as a positional argument call it with an identical copy of the chain.  We call each in succession and then pass all of the results in as a multidimensional array to the `exit_fn`.

Something I still want to figure out is cross communication between the functions.

eth/chains/base.py
```diff
@@ -447,14 +449,27 @@ def ensure_header(self, header: BlockHeader=None) -> BlockHeader:
...
         """
-        lower_limit = max(header.block_number - limit, 0)
-        for n in reversed(range(lower_limit, header.block_number)):
-            yield self.get_canonical_block_by_number(BlockNumber(n))
+        ancestor_count = min(header.block_number, limit)
+
+        # We construct a temporary block object
+        vm_class = self.get_vm_class_for_block_number(header.block_number)
+        block_class = vm_class.get_block_class()
+        block = block_class(header=header, uncles=[])
```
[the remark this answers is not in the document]
pipermerriam: IIRC no.  We have to be able to call this function on headers that we are still validating and have not persisted into our database so we can't look it up.

eth/chains/base.py
```diff
@@ -828,7 +843,7 @@ def mine_block(self, *args: Any, **kwargs: Any) -> BaseBlock:
         self.validate_block(mined_block)
 
         self.chaindb.persist_block(mined_block)
-        self.header = self.ensure_header()
```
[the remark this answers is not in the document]
pipermerriam: I don't know yet, I *think* yes, but I won't be surprised if I see a test fail because of it..... :question: :exclamation:

eth/tools/builder/chain/builders.py
```diff
@@ -0,0 +1,312 @@
...
+
+    return pipe(fork_chain, *applicator_fns)
+
+
+def fork_chain(*applicator_fns):
+    """
+    Used for forking the chain.  Should be used in conjunction with
+    `at_block_number` to 'rewind' the chain back to a specific height.
+    """
+    return _fork_chain(applicator_fns)
```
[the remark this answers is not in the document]
pipermerriam: I find long inlined functions visually confusing.  Given there's no preamble or cleanup I suspect this one could look ok.  I'll try it.

cburgdorf:
Great stuff. I guess the benchmarks could also make use of this.

tests/core/builder-tools/test_chain_initializer.py
pipermerriam: Going to save this for a subsequent PR.  That type of named tuple is only supported in python3.6 and this code has to support 3.5.

tests/core/builder-tools/test_chain_builder.py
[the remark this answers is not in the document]
carver: Hm, I think I like the idea of `chain_split` but haven't fallen in :heart: with any of these APIs. Maybe `chain_split` should actually return a tuple of chains (and so is always intended to be used last in the `pipe`):

```py
chaina, chainb = pipe(
    chain,
    mine_block(),
    mine_block(),
    chain_split(
        (mine_block(), ),  # fork A
        (mine_block(extra_data=b'fork-it!'), mine_block(), mine_block(), # fork B
    ),
)
```

---

FWIW, I think the current approach for forking is okay (explicit and clear). Maybe a slightly special version of `pipe` should always be used that never mutates the original chain. Then we wouldn't need a `fork_chain` at all, it would just be:

```py
    pre_fork_chain = build_chain(
        mining_chain,
        mine_block(),  # 1
        mine_block(),  # 2
        mine_block(),  # 3
    )

    fork_chain = build_chain(  # every build_chain automatically forks so the original isn't mutated
        pre_fork_chain,
        at_block_number(2),
        mine_block(extra_data=b'fork-it!'),  # fork 3
        mine_block(),  # fork 4
        mine_block(),  # fork 5
    )
```

I guess the above `chain_split` is easier to read.

pipermerriam:
@carver @cburgdorf seeking another round of feedback.  I've got this to a point where the re-org tests have now been rewritten using these tools.  I also wrote up rudimentary documentation for this as well as fixed #1201 by converting the README docs to sphinx.

After this next round of review I'll update the docs to include examples of how to use all-the-things.

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+
+    This function specifies the state prior to the test execution. Multiple invocations don't override
+    the state but extend it instead.
+
+    In general, the elements of `state_definitions` are nested dictionaries of the following form:
+
+    .. code-block:: python
+
+        {
+            address: {
```
carver: to follow the pattern: `<address>`

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+                "balance": <account balance>,
+                "code": <account code>,
+                "storage": {
+                    <storage slot>: <storage value>
+                }
+            }
+        }
+
+    To avoid unnecessary nesting especially if only few fields per account are specified, the following
+    and similar formats are possible as well:
```
carver: This is a lot of different ways to specify the same thing. It will be harder to spot if a line is a bug, or just one of the 5 different ways to do it correctly. Can we narrow it down a bit?

It doesn't seem that much tougher to always have a `dict` even for the short ones, like:

```py
{address, {'balance': acct_balance}}
{address, {'storage': {slot: val}}
```

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+    +--------------------+------------------------------------------------------------+
+    | ``"vyperLLLCode"`` | the code in Vyper LLL (compiled to bytecode automatically) |
+    +--------------------+------------------------------------------------------------+
+
+
+.. autofunction:: eth.tools.builder.state.expect
+
+    This specifies the expected result of the test.
+
+    For state tests, multiple expectations can be given, differing in the the transaction data, gas
```
carver: > multiple expectations

Worth clarifying that these will be run as separate tests (not sequentially, right?).

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+    limit, and value, in the applicable networks, and as a result also in the post state. VM tests
+    support only a single expectation with no specified network and no transaction (here, its role is
+    played by :func:`~eth.tools.builder.state.execution`).
+
+    * ``post_state`` is a list of state definition in the same form as expected by `pre_state`. State items
```
carver: would be nice to link to 
```
:meth:`pre_state`
```

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+        from cytoolz import pipe
+        from eth.chains.base import MiningChain
+        from eth.tools.builder.chain import fork_at
+
+        FrontierOnlyChain = pipe(MiningChain, fork_at(FrontierVM, 0))
+
+        # these two classes are functionally equivalent.
+        class FrontierOnlyChain(MiningChain):
+            vm_configuration = (
+                (0, FrontierVM),
```
carver: Might be nice to flip the args in `fork_at` just for consistency with this format.

I assume it is in the current order for currying, but we could do a kwarg curry.

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+    * :func:`~eth.tools.builder.chain.constantinople_at`
+
+.. autofunction:: eth.tools.builder.chain.dao_fork_at
+
+    Sets the block number on which the DAO fork will happen.  Requires that a
+    version of the :class:`~eth.vm.forks.homestead.HomesteadVM` is present in
+    the chain's ``vm_configuration``
+    
+
+.. autofunction:: eth.tools.builder.chain.disable_dao_fork
```
carver: If neither of these is called, what's the default?

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+Initializing Chains
+~~~~~~~~~~~~~~~~~~~
+
+The following utilities are provided to assist with initializing a chain into
+the genesis state.
+
+.. autofunction:: eth.tools.builder.chain.genesis
+
+    Initializes the given chain class with the given genesis header parameters
+    and chain state.
```
carver: An example and/or list of parameters would be sweet.

eth/tools/builder/state/builders.py
```diff
@@ -139,21 +140,22 @@ def setup_main_filler(name, environment=None):
...
+        defaults = {address: {
+            "balance": 0,
+            "nonce": 0,
+            "code": b"",
+            "storage": {},
+        } for address in pre_state}
+        new_pre_state = deep_merge(defaults, old_pre_state, pre_state)
 
-    return assoc_in(filler, [test_name, "pre"], new_pre_state)
+        return assoc_in(filler, [test_name, "pre"], new_pre_state)
```
carver: forgot an outer return here?

tests/core/builder-tools/test_chain_builder.py
```diff
@@ -0,0 +1,210 @@
...
+        import_block(block_3),
+    )
+    head = chain.get_canonical_head()
+    assert head == block_3.header
+
+
+def test_chain_import_blocks_many(mining_chain):
+    temp_chain = pipe(
+        mining_chain,
+        copy(),
```
carver: Given how often we want to `copy()`, I'm still liking the idea to have a:

```py
def build_chain(input, *args):
  return pipe(input, copy(), *args)
```

so this reduces to:

```py
temp_chain = build_chain(mining_chain, mine_blocks(3))
```

tests/core/builder-tools/test_chain_construction.py
```diff
@@ -0,0 +1,136 @@
...
+        (tangerine_whistle_at, TangerineWhistleVM),
+        (spurious_dragon_at, SpuriousDragonVM),
+        (byzantium_at, ByzantiumVM),
+        (constantinople_at, ConstantinopleVM),
+    )
+)
+def test_chain_builder_construct_chain_fork_specific_helpers(fork_fn, vm_class):
+    chain = pipe(
+        MiningChain,
+        fork_fn(0),
```
carver: maybe pick a non-zero number here to prove that the VM before the fork is still the old default VM.

tests/core/chain-object/test_chain_reorganization.py
```diff
@@ -1,121 +1,73 @@
...
+    # mine ahead 3 blocks on the fork chain
+    fork_chain = pipe(
+        chain,
+        api.copy(),
+        api.mine_block(extra_data=b'fork-it'),
+        api.mine_blocks(2),
+    )
+    # mine ahead 2 blocks on the main chain
+    pipe(
+        chain,
```
carver: It's weird to me to use `pipe` and rely on mutating the input value.

tests/core/chain-object/test_chain_reorganization.py
```diff
@@ -127,28 +79,42 @@ def test_import_block_with_reorg(chain, fork_chain, funded_address_private_key):
...
+            api.import_block(block_a),
+            api.import_block(block_b),  # the block with the uncle.
+        )
+
+    chain = pipe(
+        chain,
+        api.chain_split(
+            (api.mine_block(),),  # this will be the 'uncle'
+            (api.mine_block(extra_data=b'fork-it'),),
+            exit_fn=converge_split_fn,
```
carver: I'm not feeling this `exit_fn` API still. If you aren't a fan of this: https://github.com/ethereum/py-evm/pull/1209#discussion_r212447203

Another option is to not use the pipe at all when chain splitting, and have it be its own top-level method:

```py
chain_a, chain_b = api.chain_split(
        chain,
        (
            (api.mine_block(),),  # this will be the 'uncle'
            (api.mine_block(extra_data=b'fork-it'),),
         ),
)

## insert the body of convert_split_fn here
```

tests/core/chain-object/test_chain_reorganization.py
```diff
@@ -127,28 +79,42 @@ def test_import_block_with_reorg(chain, fork_chain, funded_address_private_key):
...
+    def converge_split_fn(results):
+        chain_a = results[0][-1]
+        chain_b = results[1][-1]
+
+        # the chain head from the main chain will become an uncle once we
+        # import the fork chain blocks.
+        uncle = chain_a.get_canonical_head()
+
+        pipe(
+            chain_b,
```
carver: more reliance on mutation in `pipe` :/

tests/core/chain-object/test_chain_reorganization.py
pipermerriam: I'm not in love with `exit_fn` either.  I'm going to go with your suggestion to just have `chain_split` return the resulting chain for all of the chains.

This however is an API I want to revisit at some point because I **really** want a way to concisely, explicitly cause a fork, and then adopt the fork blocks as uncles into the main fork and to have it still be easy to read and understand what is going on.  For now, this will do.

tests/core/chain-object/test_chain_reorganization.py
pipermerriam: I agree,  :+1: on a `build(chain, fn(1), fn(1))` function which is equivalent to `pipe(chain, copy(), fn1(), fn2()).

docs/tools/builder/index.rst
pipermerriam: Honestly not sure, this is me porting the README which is a bit out-of-date.  I think I'm going to separate the two since I've realized they do fundamentally different things.

- one builds fixtures for VM and State tests
- one builds chains
- <not yet build thing> builds fixtures for Blockchain tests (on top of the chain builder APIs).

cburgdorf: >differing in ~~the~~ the transaction data

typo

docs/tools/builder/index.rst
```diff
@@ -0,0 +1,329 @@
...
+blocks.
+
+
+.. autofunction:: eth.tools.builder.chain.copy
+
+    Provides a copy of the chain at the given state.  Actions performed on the
+    resulting chain will not be represented on the original chain.
+
+
+.. autofunction:: eth.tools.builder.chain.import_block
```
cburgdorf: All these short descriptions are a great fit for our regular API descriptions that live under the `API` top level inside our docs. I think that should be moved there. Although all API descriptions should be written in simple present tense.

e.g. 

>Import~~s~~ the provided block into the chain

pipermerriam: I'll move this to live under tools

pipermerriam:
(copied from #1210 where I incorrectly posted this)

Dropping an idea here.

The reason that we need the `copy()` command is that we can't rely on reverting all of the canonical based information from the database (number-to-canonical-hash, canonical-head, txn-to-canonical-block).  Also, we don't actually want to revert that chain information because we need both chains to have access to their *canonical-thing* database entries.

The other data is going to *normally* be safe to already have there as I don't believe that we have any code that does pre-checks before storing things like blocks, transactions, receipts, etc.

If we **could** do that, then it'd should be safe to re-use the same database for fork chains.  If we can do this then copying chains just got a lot easier.  So my thought was, maybe we can figure out a scheme where database re-use is safe because we have a reliable way of storing per-chain canonical block information with no chance of cross-leakage.

I've been thinking about different keying schemes, ideally something that lets us key the canonical data to a specific chain.  It could be something as simple as each chain having a unique `salt` that is part of the key we store canonical data under.  Since we already have a schema system for how those keys are generated this should be really simple to implement.

pipermerriam:
@carver and/or @cburgdorf this has now been updated with the PR feedback

PR #1210, opened 2018-08-23
Sync headers in queue instead of batch, add Peer types

### LINKED ISSUE #1215: Subscribers to peer messages shouldn't have to wait for a timeout when the peer disconnects
### What is wrong?

When a peer disconnects, right now, subscribers to messages wait for a timeout before they realize that the peer is gone. This slows down syncing and other things (and in pathological cases with no timeout, can pause forever).

### How can it be fixed

~Maybe cause all subscribers to the peer to receive a TimeoutError immediately on peer disconnect.~

Implement `deregister_peer` in `ResponseCandidateStream` and immediately `raise TimeoutError` on any pending requests.

carver:
### What was wrong?

Step 1 of #868 -- queue headers instead of batching them

Fixes #1215 

### How was it fixed?

Main changes:
- Queue up headers for processing in later stages (instead of waiting for all post-processing to finish before requesting the next batch of headers)

Follow up work:
- Decouple the header sync from the post-processing stages. It doesn't really need to be a subclass I think
- Do the same queue processing for later steps, like downloading block bodies.
- Track one queue per peer, so that a peer is never waiting on another peer to do its own request.

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -135,7 +135,7 @@ class BaseService(ABC, CancellableMixin):
                 pass
             except Exception as e:
                 self.logger.warning("Task %s finished unexpectedly: %s", awaitable, e)
-                self.logger.warning("Task failure traceback", exc_info=True)
+                self.logger.debug("Task failure traceback", exc_info=True)
             else:
                 self.logger.debug("Task %s finished with no errors", awaitable)
         self._tasks.add(asyncio.ensure_future(_run_task_wrapper()))
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -128,7 +128,7 @@ class BaseService(ABC, CancellableMixin):
         If it raises OperationCancelled, that is caught and ignored.
         """
         async def _run_task_wrapper() -> None:
-            self.logger.debug("Running task %s", awaitable)
+            self.logger.trace("Running task %s", awaitable)
             try:
                 await awaitable
             except OperationCancelled:
@@ -137,7 +137,7 @@ class BaseService(ABC, CancellableMixin):
                 self.logger.warning("Task %s finished unexpectedly: %s", awaitable, e)
                 self.logger.debug("Task failure traceback", exc_info=True)
             else:
-                self.logger.debug("Task %s finished with no errors", awaitable)
+                self.logger.trace("Task %s finished with no errors", awaitable)
         self._tasks.add(asyncio.ensure_future(_run_task_wrapper()))
 
     def run_child_service(self, child_service: 'BaseService') -> None:
--- a/tests/p2p/test_service.py
+++ b/tests/p2p/test_service.py
@@ -26,11 +26,16 @@ class WaitService(BaseService):
 async def test_daemon_exit_causes_parent_cancellation():
     service = ParentService()
     asyncio.ensure_future(service.run())
+
     await asyncio.sleep(0.01)
+
     assert service.daemon.is_operational
     assert service.daemon.is_running
+
     await service.daemon.cancel()
     await asyncio.sleep(0.01)
+
     assert not service.is_operational
     assert not service.is_running
-    await service.events.cleaned_up.wait()
+
+    await asyncio.wait_for(service.events.cleaned_up.wait(), timeout=1)
--- a/trinity/sync/common/chain.py
+++ b/trinity/sync/common/chain.py
@@ -1,8 +1,8 @@
 import asyncio
 from abc import abstractmethod
+from operator import attrgetter
 from typing import (
     AsyncGenerator,
-    Set,
     Tuple,
     Union,
     cast,
@@ -81,7 +81,7 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
         # pending queue size should be big enough to avoid starving the processing consumers, but
         # small enough to avoid wasteful over-requests before post-processing can happen
         max_pending_headers = ETHPeer.max_headers_fetch * 8
-        self.header_queue = TaskQueue(max_pending_headers, lambda header: header.block_number)
+        self.header_queue = TaskQueue(max_pending_headers, attrgetter('block_number'))
 
     @property
     def msg_queue_maxsize(self) -> int:
@@ -168,7 +168,7 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
             return
 
         self.logger.info("Starting sync with %s", peer)
-        last_received_header = None
+        last_received_header: BlockHeader = None
         # When we start the sync with a peer, we always request up to MAX_REORG_DEPTH extra
         # headers before our current head's number, in case there were chain reorgs since the last
         # time _sync() was called. All of the extra headers that are already present in our DB
@@ -238,9 +238,20 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
             # Setting the latest header hash for the peer, before queuing header processing tasks
             self._target_header_hash = peer.head_hash
 
-            await self.header_queue.add(headers)
+            unrequested_headers = tuple(h for h in headers if h not in self.header_queue)
+            await self.header_queue.add(unrequested_headers)
             last_received_header = headers[-1]
             start_at = last_received_header.block_number + 1
+
+        # erase any pending tasks, to restart on next _sync() run
+        try:
+            batch_id, pending_tasks = self.header_queue.get_nowait()
+        except asyncio.QueueFull:
+            # nothing pending, continue
+            pass
+        else:
+            # fully remove pending tasks from queue
+            self.header_queue.complete(batch_id, pending_tasks)
 
     async def _fetch_missing_headers(
             self, peer: HeaderRequestingPeer, start_at: int) -> Tuple[BlockHeader, ...]:
--- a/trinity/sync/common/chain.py
+++ b/trinity/sync/common/chain.py
@@ -2,6 +2,7 @@ import asyncio
 from abc import abstractmethod
 from typing import (
     AsyncGenerator,
+    Set,
     Tuple,
     Union,
     cast,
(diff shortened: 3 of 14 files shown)


p2p/DEVELOPMENT.md
```diff
@@ -18,7 +18,7 @@ library.
 ## BaseService
 
 - If your service needs to run coroutines in the background, you should use the `BaseService.run_task()` method and
-  ensure they exit when `is_running` is False or when the cancel token is triggered.
+  ensure they exit when `is_operational` is False or when the cancel token is triggered.
```
pipermerriam: Just making a note that the `BaseService` class is probably at a stage where we could document it or potentially start the process of extracting it into a stand-alone `asyncio-service` library or something like that.

p2p/peer.py
```diff
@@ -106,14 +108,16 @@
...
 
 async def handshake(remote: Node,
                     privkey: datatypes.PrivateKey,
-                    peer_class: 'Type[BasePeer]',
+                    peer_class: 'Type[TPeer]',
                     headerdb: 'BaseAsyncHeaderDB',
                     network_id: int,
                     token: CancelToken,
-                    ) -> 'BasePeer':
+                    ) -> 'TPeer':
```
pipermerriam: This `TPeer` doesn't need to be a string does it?

p2p/peer.py
```diff
@@ -106,14 +108,16 @@
     from trinity.protocol.eth.requests import HeaderRequest  # noqa: F401
     from trinity.protocol.base_request import BaseRequest  # noqa: F401
 
+TPeer = TypeVar('TPeer', bound='BasePeer')
+
 
 async def handshake(remote: Node,
                     privkey: datatypes.PrivateKey,
-                    peer_class: 'Type[BasePeer]',
+                    peer_class: 'Type[TPeer]',
```
pipermerriam: This doesn't need to be in string form does it?

p2p/service.py
```diff
@@ -109,11 +109,14 @@ def get_event_loop(self) -> asyncio.AbstractEventLoop:
...
             self.cancel_token.trigger()
 
+            print('waiting for', self, ',', self._tasks, 'and', self._child_services, 'to cleanup')
             await self.cleanup()
 
+            print('waiting for', self, 'to callback')
             for callback in self._finished_callbacks:
                 callback(self)
 
+            print(self, 'closed')
```
pipermerriam: Just noting print statements here for cleanup before merge.

p2p/service.py
```diff
@@ -135,6 +138,7 @@ def run_task(self, awaitable: Awaitable[Any]) -> None:
                 pass
             except Exception as e:
                 self.logger.warning("Task %s finished unexpectedly: %s", awaitable, e)
+                self.logger.warning("Task failure traceback", exc_info=True)
```
pipermerriam: Maybe log at `DEBUG` level?

tests/p2p/test_service.py
```diff
@@ -27,8 +27,10 @@ class WaitService(BaseService):
     service = ParentService()
     asyncio.ensure_future(service.run())
     await asyncio.sleep(0.01)
+    assert service.daemon.is_operational
```
pipermerriam: Maybe it's just the diff but the asserts and awaits are all running together visually for me, maybe add some line breaks if you get back in here for other things?

trinity/protocol/eth/exchanges.py
```diff
@@ -45,7 +49,14 @@
     ReceiptsValidator,
 )
 
-BaseGetBlockHeadersExchange = BaseExchange[
+if TYPE_CHECKING:
+    from .peer import ETHPeer
+else:
+    ETHPeer = None
```
pipermerriam: This is confusing me

trinity/sync/common/chain.py
```diff
@@ -99,19 +118,18 @@ def register_peer(self, peer: BasePeer) -> None:
...
-                    self._sync_requests.get(),
-                    self._sync_complete.wait()
-                )
-
-                # In the case of a fast sync, we return once the sync is completed, and our caller
-                # must then run the StateDownloader.
-                if self._sync_complete.is_set():
+                try:
+                    print('waiting for peer to sync')
+                    peer = await self.cancel_token.cancellable_wait(self._sync_requests.get())
```
pipermerriam: Is there a reason not to use `self.wait`?

p2p/service.py
carver: Ah yes, that's exactly what I meant to do.

trinity/sync/common/chain.py
carver: :+1: There used to be, but it's gone.

carver:
Ok, the strict peer typing changes were a rabbit hole that never seemed to end. Turns out they weren't strictly necessary, so I dropped them. Posting the changes shortly...

carver:
This still has at least one blocker: the search for the common ancestor assumes that the headers have finished processing. Solution in progress...

carver:
> the search for the common ancestor assumes that the headers have finished processing

:man_facepalming: okay, so this was a little less obvious to solve cleanly, and sent me on another quest. Feel free to skip reviewing until I *really* finish...

pipermerriam:
Dropping an idea here.

The reason that we need the `copy()` command is that we can't rely on reverting all of the canonical based information from the database (number-to-canonical-hash, canonical-head, txn-to-canonical-block).  Also, we don't actually want to revert that chain information because we need both chains to have access to their *canonical-thing* database entries.

The other data is going to *normally* be safe to already have there as I don't believe that we have any code that does pre-checks before storing things like blocks, transactions, receipts, etc.

If we **could** do that, then it'd should be safe to re-use the same database for fork chains.  If we can do this then copying chains just got a lot easier.  So my thought was, maybe we can figure out a scheme where database re-use is safe because we have a reliable way of storing per-chain canonical block information with no chance of cross-leakage.

I've been thinking about different keying schemes, ideally something that lets us key the canonical data to a specific chain.  It could be something as simple as each chain having a unique `salt` that is part of the key we store canonical data under.  Since we already have a schema system for how those keys are generated this should be really simple to implement.

carver:
> The reason that we need the `copy()` command

Sorry, which `copy()` usage are you talking about?

> should be safe to re-use the same database for fork chains

Hm, wo do store multiple forks in the same database. -- I think I'm missing some preamble to this conversation.

pipermerriam:
I posted that comment in the wrong PR, meant to put it in #1209

trinity/sync/common/chain.py
```diff
@@ -245,12 +262,18 @@ def register_peer(self, peer: BasePeer) -> None:
...
+    async def pop_all_pending_headers(self) -> Tuple[BlockHeader, ...]:
+        """Get all the currently pending headers. If no headers pending, wait until one is"""
+        queue = self.pending_headers
+        if queue.empty():
+            first_header = await queue.get()
+        else:
+            first_header = queue.get_nowait()
+
+        available = queue.qsize()
+        return (first_header, ) + tuple(queue.get_nowait() for _ in range(available))
```
pipermerriam: This function is clever enough that I think it deserves an initial comment.  It wasn't clear to me why it was written this way on my first read.

I.E. why get `first_header` and then pull the remaining tail headers...... Oh! so that we know there *was* something in the queue and still get the benefit of not doing a bunch of await calls.

trinity/sync/common/chain.py
```diff
@@ -162,7 +176,11 @@ def register_peer(self, peer: BasePeer) -> None:
                 break
 
             try:
-                headers = await self._fetch_missing_headers(peer, start_at)
+                fetch_headers_coro = self._fetch_missing_headers(peer, start_at)
+                headers = await self.complete_token.cancellable_wait(fetch_headers_coro)
+            except OperationCancelled:
+                self.logger.info("Sync with %s completed", peer)
```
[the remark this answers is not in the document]
carver: > Will this logging statement be reliably accurate?

Mostly, except it's not perfectly accurate when it's canceled at shutdown.

> can it be triggered by another mechanism?

I guess someone misusing it, like you alluded to.

> can you think of a way that we can make it harder for someone to come along and make a mistake triggering the complete_token for the wrong reason?

Maybe hiding `complete_token` into `_complete_token` behind a `complete_sync()`.

trinity/sync/common/chain.py
```diff
@@ -192,22 +210,21 @@ def register_peer(self, peer: BasePeer) -> None:
...
-                    self._sync_complete.set()
-                break
+            # Setting the latest header hash for the peer, before queuing header processing tasks
+            self._target_header_hash = peer.head_hash
+
+            await self._queue_headers_for_processing(headers)
+            start_at = headers[-1].block_number + 1
+
+    async def _queue_headers_for_processing(self, headers: Tuple[BlockHeader, ...]) -> None:
+        # this block is an optimization to avoid lots of await calls
```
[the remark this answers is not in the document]
carver: Did this by moving to a new data structure, since it was getting complicated to solve the other problem.

carver:
> As it stands I think that we may start syncing with peer A and then connect to peer B which has a higher TD.

Right, we would only fast sync to B's head AFAICT.

---

Final TODO on this PR: hunt down the `QueueFull` bug that's trying to push tasks into the queue when it's full.

carver:
Hm looks like a library download failure, and circle won't let me restart it. If I don't hear from anyone by tomorrow, I'll go ahead and merge. I've run `trinity --ropsten` and `trinity --ropsten --light` for at least a few hours locally, and things look ok. (more performance improvements to come)

pipermerriam:
I was able to kick the job

pipermerriam:
So up next does this get applied to the trie nodes?

carver:
Up next, I had applying this to the receipts/blocks in the full syncer. Right now the batching approach there is definitely a cause of slowdowns.

PR #1211, opened 2018-08-23
PoC: Communication between PeerPool and other process via event bus

cburgdorf:
### Abstract?

This is a follow up to the (now closed) #1202 PR. After substantial progress, the description became outdated so I decided to open up a new PR. 

This is a PoC to show case how we can communicate between processes with the following properties.

- Messaging rather than RPC ([related article](http://www.inspirel.com/articles/RPC_vs_Messaging.html), [related comment](https://github.com/ethereum/py-evm/issues/1075#issuecomment-413489849))
- no `Proxy` or `Manager` classes were used. Only `Queue`
- all APIs are non-blocking and are built on top of `asyncio`

### What does this do?

1. The `PeerPool` got a handle to the event bus and is listening to `PeerCountRequest`s and answers them with a `PeerCountResponse`

```python
    self.event_bus.broadcast(PeerCountResponse(len(self)), req.broadcast_config())
```
Notice that while the event bus enables us to raise events to be consumed by many independent players, in this case the `broadcast` is done in a fashion to only broadcast an answer to the call site that raised the event we are listening for. This is done by passing a `BroadcastConfig` as a second parameter which we derive from the *incoming* event by calling `broadcast_config()` on it.

2. We set up another process to communicate with the `PeerPool` via these events.

```
test_proc = ctx.Process(
        target=run_test_proc,
        args=(
            test_proc_endpoint,
        ),
    )
     test_proc.start()
```

Notice that the process gets a handle to the event bus injected (similar to how we inject a handle in the networking process that hosts the `PeerPool`)

Then, inside that process we are raising a `PeerCountRequest` on the event bus every second.

```python
async def request_receive_peer_count(event_bus):
    while True:
        response = await event_bus.request(PeerCountRequest())
        print("Peer Count: " + str(response.payload))
        await asyncio.sleep(1)
```

### Where to go from here?

There are many things we can do in the event bus to improve ergonomics and efficiency. However, I feel we are nearing a state where this can actually be used for real. I see the next steps as follows:

#### Support plugins to run in their own process

Here's roughly what I think needs to be done:

1. ~~Allow plugins to express that they want to run in their own process. That in turn will cause the plugin manger that lives in the main process (not  to be confused with the instance that lives in the network process) to setup an eventbus endpoint for the plugin. This *needs* to happen in the main process for now.~~ This is now done.

2. Let the plugin handle the launch of the process (not sure, needs more thought)

3. Let the plugin do it's thing and communicate with the rest of Trinity via the event bus. This also means, depending on the first use case to spearhead this, that probably the `PeerPool` needs to expose some APIs built on top of the eventbus (similar to the demoed `PeerCountRequest` / `PeerCountResponse` in this PR)

I think that the `TxPool` is a great candidate to spearhead this.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -42,6 +42,7 @@ deps = {
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.5",
         "web3==4.4.1",
+        "lahja==0.4.0",
     ],
     'test': [
         "hypothesis==3.44.26",
--- a/trinity/events.py
+++ b/trinity/events.py
@@ -1,0 +1,13 @@
+from lahja import (
+    BaseEvent
+)
+
+
+class PeerCountRequest(BaseEvent):
+
+    def __init__(self) -> None:
+        super().__init__(None)
+
+
+class PeerCountResponse(BaseEvent):
+    pass
--- a/trinity/extensibility/plugin.py
+++ b/trinity/extensibility/plugin.py
@@ -6,17 +6,43 @@ from argparse import (
     ArgumentParser,
     _SubParsersAction,
 )
+from enum import (
+    auto,
+    Enum,
+)
 import logging
 from typing import (
-    Any
+    Any,
+    Dict,
+)
+
+from lahja import (
+    Endpoint
 )
 
 from trinity.extensibility.events import (
     BaseEvent
 )
+from trinity.utils.mp import (
+    ctx,
+)
+from trinity.utils.logging import (
+    setup_queue_logging,
+)
 
-# TODO: we spec this out later
-PluginContext = Any
+
+class PluginProcessScope(Enum):
+
+    SHARED = auto()
+    MAIN = auto()
+    OWN = auto()
+
+
+class PluginContext:
+
+    def __init__(self, endpoint: Endpoint, boot_kwargs: Dict[str, Any] = None) -> None:
+        self.event_bus = endpoint
+        self.boot_kwargs = boot_kwargs
 
 
 class BasePlugin(ABC):
@@ -30,6 +56,10 @@ class BasePlugin(ABC):
         raise NotImplementedError(
             "Must be implemented by subclasses"
         )
+
+    @property
+    def process_scope(self) -> PluginProcessScope:
+        return PluginProcessScope.SHARED
 
     @property
     def logger(self) -> logging.Logger:
@@ -67,6 +97,37 @@ class BasePlugin(ABC):
         Called when the plugin gets stopped. Should be overwritten to perform cleanup
         work in case the plugin set up external resources.
         """
+        pass
+
+
+class BaseOwnProcessPlugin(BasePlugin):
+
+    @property
+    def process_scope(self) -> PluginProcessScope:
+        return PluginProcessScope.OWN
+
+    def start(self, context: PluginContext) -> None:
+        proc = ctx.Process(
+            target=self._launch_process,
+            args=(
+                context.event_bus,
+            ),
+            kwargs=context.boot_kwargs
+        )
+
+        proc.start()
+
+    @classmethod
+    def _launch_process(cls, event_bus: Endpoint, **kwargs: Any) -> None:
+        log_queue = kwargs['log_queue']
+        level = kwargs.get('log_level', logging.INFO)
+        setup_queue_logging(log_queue, level)
+
+        cls.launch_process(event_bus)
+
+    @staticmethod
+    @abstractmethod
+    def launch_process(event_bus: Endpoint, **kwargs: Any) -> None:
         pass
 
 
--- a/trinity/extensibility/plugin_manager.py
+++ b/trinity/extensibility/plugin_manager.py
@@ -4,9 +4,16 @@ from argparse import (
 )
 import logging
 from typing import (
+    Any,
+    Dict,
     Iterable,
     List,
     Union,
+)
+
+from lahja import (
+    Endpoint,
+    EventBus,
 )
 
 from trinity.extensibility.events import (
@@ -15,7 +22,25 @@ from trinity.extensibility.events import (
 )
 from trinity.extensibility.plugin import (
     BasePlugin,
+    PluginContext,
+    PluginProcessScope,
 )
+
+
+class MainAndOwnProcessScope():
+
+    def __init__(self, event_bus: EventBus, main_proc_endpoint: Endpoint) -> None:
+        self.event_bus = event_bus
+        self.endpoint = main_proc_endpoint
+
+
+class SharedProcessScope():
+
+    def __init__(self, shared_proc_endpoint: Endpoint) -> None:
+        self.endpoint = shared_proc_endpoint
+
+
+ManagerProcessScope = Union[SharedProcessScope, MainAndOwnProcessScope]
 
 
 class PluginManager:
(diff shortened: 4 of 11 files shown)


cburgdorf:
@pipermerriam I pushed this another step forward. The new process is now wrapped as a plugin. There are still a lot of rough edges but hey, this is a plugin, running in its own process, talking to the `PeerPool` via the event bus with zero plugin specific changes to the core code base :)

setup.py
```diff
@@ -42,6 +42,7 @@
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.5",
         "web3==4.4.1",
+        "lahja==0.4.0",
```
pipermerriam: I believe that you need to cut another release to remove `aioprocessing` from the `lahja` dependencies as I think it is no longer needed.

trinity/events.py
```diff
@@ -0,0 +1,13 @@
+from lahja import (
+    BaseEvent
+)
+
+
+class PeerCountRequest(BaseEvent):
+
+    def __init__(self) -> None:
+        super().__init__(None)
```
pipermerriam: I think that `BaseEvent` should be updated so that it doesn't require this hack.  I'd suggest using a *sentinal* empty object to allow distinguishing between `None` and no-payload-provided.

```python
class EmptyPayload:
    pass

empty_payload = EmptyPayload()


class BaseEvent:
    def __init__(self, payload=empty_payload):
        ...
```

*or*

It may just make sense to have payload only accessible via a method like `get_payload()` and raise an exception if no payload was provided so that events with no payload cannot accidentally have their payload used for something when it is empty.

Honestly it's not clear what the right thing to do here is wrt payloads but I think this hack should be removed.

trinity/extensibility/plugin.py
```diff
@@ -6,17 +6,43 @@
...
+)
+
+
+class PluginProcessScope(Enum):
 
-# TODO: we spec this out later
-PluginContext = Any
+    SHARED = auto()
+    MAIN = auto()
+    OWN = auto()
```
pipermerriam: Maybe 

-`PRIVATE` 
-`ISOLATED`

Also, these probably deserve comments indicating what they mean here in the code.

trinity/extensibility/plugin_manager.py
```diff
@@ -63,7 +96,7 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
         """
         for plugin in self._plugin_store:
 
-            if plugin is exclude:
+            if plugin is exclude or not self._is_responsible_for_plugin(plugin):
```
pipermerriam: This `exclude` API stands out as something that probably needs an iteration to cleanup.

- it seems to only be used by the `PluginManager` itself, indicating that the API shouldn't be public.
- it only supports exclusion of a single plugin which seems arbitrarily inflexible.

Not for this PR necessarily but probably something worth cleaning up.

trinity/extensibility/plugin_manager.py
```diff
@@ -74,7 +107,53 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
             if not plugin.should_start():
                 continue
 
-            plugin.start(None)
+            context = self._create_context_for_plugin(plugin)
+            plugin.start(context)
```
pipermerriam: I'm thinking that we should be really intentional/careful about this lifecycle.  Something about starting the next plugin mid-loop stands out to me as potentially problematic.

Consider this.

- `A` broadcasts event `x`
- Event `x` triggers `B` to start, broadcasting the `started` event.
- The `started` event triggers `C` to start.
- Finally `x` resumes being proadcast to the remaining plugins.

I see you have a comment about `start` being a temporary API so maybe this is a non issue.  Just wanted to surface it.

trinity/extensibility/plugin_manager.py
```diff
@@ -74,7 +107,53 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+
+            if plugin.process_scope is not PluginProcessScope.OWN:
+                continue
+
+            context = self._create_context_for_plugin(plugin, boot_kwargs)
+            plugin.start(context)
+
+    def _is_responsible_for_plugin(self, plugin: BasePlugin) -> bool:
+
+        main_and_own_scopes = [PluginProcessScope.MAIN, PluginProcessScope.OWN]
```
pipermerriam: You should use `set` notation when creating a data structure for *inclusion* checks.

```python
{PluginProcessScope.Main, PluginProcessScope.OWN}
```

Also, maybe store this as a top level module variable so we don't incure the minor but additional runtime overhead of constructing this set inline.

trinity/extensibility/plugin_manager.py
```diff
@@ -74,7 +107,53 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+        plugin_for_main_or_own = plugin.process_scope in main_and_own_scopes
+        plugin_for_shared = not plugin_for_main_or_own
+
+        manager_for_main_or_own = isinstance(self._scope, MainAndOwnProcessScope)
+        manager_for_shared = not manager_for_main_or_own
+
+        return (plugin_for_main_or_own and manager_for_main_or_own or
```
pipermerriam: Can you add one extra layer of parenthesis to group these statements in a way that doesn't require the reader to think about what order they will be executed in?  This is hard to read for me, though I do think the variable names help.

trinity/plugins/builtin/proc_test/plugin.py
```diff
@@ -0,0 +1,47 @@
...
+)
+from trinity.extensibility import (
+    BaseOwnProcessPlugin,
+)
+
+
+async def request_receive_peer_count(event_bus: Endpoint) -> None:
+    while True:
+        response = await event_bus.request(PeerCountRequest())
+        print("Peer Count: " + str(response.payload))
```
pipermerriam: Can this be a logging statement?  This also implies that we need to have the plugin API correctly configure logging to go over the queued logging API.

setup.py
cburgdorf: Ah! Probably! How do you feel about moving that under the ethereum umbrella?

trinity/extensibility/plugin_manager.py
cburgdorf: I agree. I think that whole thing will change.

trinity/extensibility/plugin_manager.py
cburgdorf: Agreed. That whole API is entirely in flux. I'll get back to that as soon as the dust settles.

trinity/extensibility/plugin.py
cburgdorf: Agreed. I like `ISOLATED` for plugins that run in their own process. `SHARED` means, the plugin runs in a process that is shared among other plugins (currently `networking` but long term I see them running in a shared process that is exclusively for plugins)

trinity/events.py
cburgdorf: I was even considering getting rid of `payload` entirely and let the events specify whatever fields they need.

setup.py
cburgdorf: Ah! Probably! How do you feel about moving that under the ethereum umbrella?

trinity/plugins/builtin/proc_test/plugin.py
cburgdorf: Yes. In fact, the queued logging is working for the isolated plugins already. This is really just dummy code. I have half of the work done to move the JSON-RPC server into a plugin that runs in an isolated process and communicates via event bus with the peer pool. I think that will be the first useful plugin.

setup.py
pipermerriam: Can you coordinate with devops team.  I think someone will have to click "accept" to migrate the repository.  Have them put it under the "Python-Admin` group for administration.

cburgdorf:
This can be closed. It was only meant as a PoC and #1212 aims to move this over the finish line.

PR #1212, opened 2018-08-23
Move JSON-RPC server into an isolated plugin

cburgdorf:
### What was wrong?

Currently the `JSON-RPC` server runs in the `networking` process together the syncing and other messaging. Now that isolated plugins are a thing, it's a good idea to move the `JSON-RPC` server because:

- a separate process provides better isolation
  - DDOS attacks
  - crashes in the JSON-RPC can't bring the syncing down
- better code organization

### How was it fixed?

- provide a `LightPeerChainBridge` plugin that brings an `EventBusLightPeerChain` to be consumed from other processes
- enforce all RPC modules to be async
- turn existing JSON-RPC server into an isolated plugin

In addition to the current functionality of the `JSON-RPC` server, the new plugin-based version can be disabled by passing `--disable-rpc` at startup

#### Cute Animal Picture

### DIFF
--- a/p2p/events.py
+++ b/p2p/events.py
@@ -1,0 +1,21 @@
+from typing import (
+    Type,
+)
+
+from lahja import (
+    BaseEvent,
+    BaseRequestResponseEvent,
+)
+
+
+class PeerCountResponse(BaseEvent):
+
+    def __init__(self, peer_count: int) -> None:
+        self.peer_count = peer_count
+
+
+class PeerCountRequest(BaseRequestResponseEvent[PeerCountResponse]):
+
+    @staticmethod
+    def expected_response_type() -> Type[PeerCountResponse]:
+        return PeerCountResponse
--- a/tests/trinity/conftest.py
+++ b/tests/trinity/conftest.py
@@ -5,11 +5,17 @@ import pytest
 import tempfile
 import uuid
 
-from p2p.tools.paragon import (
-    ParagonContext,
-    ParagonPeerPool,
+from lahja import (
+    EventBus,
 )
 
+from eth.chains import (
+    Chain,
+)
+
+from trinity.chains.coro import (
+    AsyncChainMixin,
+)
 from trinity.rpc.main import (
     RPCServer,
 )
@@ -22,6 +28,13 @@ from trinity.utils.xdg import (
 from trinity.utils.filesystem import (
     is_under_path,
 )
+from tests.conftest import (
+    _chain_with_block_validation,
+)
+
+
+class TestAsyncChain(Chain, AsyncChainMixin):
+    pass
 
 
 def pytest_addoption(parser):
@@ -51,16 +64,35 @@ def event_loop():
         loop.close()
 
 
+@pytest.fixture(scope='module')
+def event_bus(event_loop):
+    bus = EventBus()
+    endpoint = bus.create_endpoint('test')
+    bus.start(event_loop)
+    endpoint.connect(event_loop)
+    try:
+        yield endpoint
+    finally:
+        endpoint.stop()
+        bus.stop()
+
+
 @pytest.fixture(scope='session')
 def jsonrpc_ipc_pipe_path():
     with tempfile.TemporaryDirectory() as temp_dir:
         yield Path(temp_dir) / '{0}.ipc'.format(uuid.uuid4())
 
 
+@pytest.fixture
+def chain_with_block_validation(base_db, genesis_state):
+    return _chain_with_block_validation(base_db, genesis_state, TestAsyncChain)
+
+
 @pytest.mark.asyncio
 @pytest.fixture
 async def ipc_server(
         monkeypatch,
+        event_bus,
         jsonrpc_ipc_pipe_path,
         event_loop,
         chain_with_block_validation):
@@ -68,9 +100,7 @@ async def ipc_server(
     This fixture runs a single RPC server over IPC over
     the course of all tests. It yields the IPC server only for monkeypatching purposes
     '''
-    peer_pool = ParagonPeerPool(b'unicornsrainbows' * 2, ParagonContext())
-
-    rpc = RPCServer(chain_with_block_validation, peer_pool)
+    rpc = RPCServer(chain_with_block_validation, event_bus)
     ipc_server = IPCServer(rpc, jsonrpc_ipc_pipe_path, loop=event_loop)
 
     asyncio.ensure_future(ipc_server.run(), loop=event_loop)
--- a/tests/trinity/core/json-rpc/test_ipc.py
+++ b/tests/trinity/core/json-rpc/test_ipc.py
@@ -14,6 +14,11 @@ from eth_utils import (
     to_hex,
 )
 
+from p2p.events import (
+    PeerCountRequest,
+    PeerCountResponse,
+)
+
 from trinity.utils.version import construct_trinity_client_identifier
 
 
@@ -33,15 +38,6 @@ def build_request(method, params=[]):
         'params': params,
     }
     return json.dumps(request).encode()
-
-
-class MockPeerPool:
-
-    def __init__(self, peer_count=0):
-        self.peer_count = peer_count
-
-    def __len__(self):
-        return self.peer_count
 
 
 def id_from_rpc_request(param):
@@ -68,6 +64,7 @@ async def get_ipc_response(
         jsonrpc_ipc_pipe_path,
         request_msg,
         event_loop):
+
     assert wait_for(jsonrpc_ipc_pipe_path), "IPC server did not successfully start with IPC file"
 
     reader, writer = await asyncio.open_unix_connection(str(jsonrpc_ipc_pipe_path), loop=event_loop)
@@ -399,18 +396,27 @@ async def test_eth_call_with_contract_on_ipc(
     assert result == expected
 
 
+def mock_peer_count(count):
+    async def mock_event_bus_interaction(bus):
+        async for req in bus.stream(PeerCountRequest):
+            bus.broadcast(PeerCountResponse(count), req.broadcast_config())
+            break
+
+    return mock_event_bus_interaction
+
+
 @pytest.mark.asyncio
 @pytest.mark.parametrize(
-    'request_msg, mock_peer_pool, expected',
+    'request_msg, event_bus_setup_fn, expected',
     (
         (
             build_request('net_peerCount'),
-            MockPeerPool(peer_count=1),
+            mock_peer_count(1),
             {'result': '0x1', 'id': 3, 'jsonrpc': '2.0'},
         ),
         (
             build_request('net_peerCount'),
-            MockPeerPool(peer_count=0),
+            mock_peer_count(0),
             {'result': '0x0', 'id': 3, 'jsonrpc': '2.0'},
         ),
     ),
@@ -422,10 +428,17 @@ async def test_peer_pool_over_ipc(
         monkeypatch,
         jsonrpc_ipc_pipe_path,
         request_msg,
-        mock_peer_pool,
+        event_bus_setup_fn,
+        event_bus,
         expected,
         event_loop,
         ipc_server):
-    monkeypatch.setattr(ipc_server.rpc.modules['net'], '_peer_pool', mock_peer_pool)
-    result = await get_ipc_response(jsonrpc_ipc_pipe_path, request_msg, event_loop)
+
+    asyncio.ensure_future(event_bus_setup_fn(event_bus))
+
+    result = await get_ipc_response(
+        jsonrpc_ipc_pipe_path,
+        request_msg,
+        event_loop
+    )
     assert result == expected
--- a/tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -25,6 +25,9 @@ from eth.tools.fixtures import (
     should_run_slow_tests,
 )
 
+from trinity.chains.mainnet import (
+    MainnetFullChain
+)
 from trinity.rpc import RPCServer
 from trinity.rpc.format import (
     empty_to_0x,
@@ -383,7 +386,7 @@ def chain(chain_without_block_validation):
 
 @pytest.mark.asyncio
 async def test_rpc_against_fixtures(chain, ipc_server, chain_fixture, fixture_data):
-    rpc = RPCServer(None)
+    rpc = RPCServer(MainnetFullChain(None))
 
     setup_result, setup_error = await call_rpc(rpc, 'evm_resetToGenesisFixture', [chain_fixture])
     assert setup_error is None and setup_result is True, "cannot load chain for %r" % fixture_data
--- a/tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py
+++ b/tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py
@@ -199,21 +199,21 @@ def result_from_response(response_str):
     return (response.get('result', None), response.get('error', None))
 
 
-def call_rpc(rpc, method, params):
+async def call_rpc(rpc, method, params):
     request = build_request(method, params)
-    response = rpc.execute(request)
+    response = await rpc.execute(request)
     return result_from_response(response)
 
 
-def assert_rpc_result(rpc, method, params, expected):
-    result, error = call_rpc(rpc, method, params)
+async def assert_rpc_result(rpc, method, params, expected):
+    result, error = await call_rpc(rpc, method, params)
     assert error is None
     assert result == expected
     return result
 
 
-def validate_account_attribute(fixture_key, rpc_method, rpc, state, addr, at_block):
-    state_result, state_error = call_rpc(rpc, rpc_method, [addr, at_block])
+async def validate_account_attribute(fixture_key, rpc_method, rpc, state, addr, at_block):
+    state_result, state_error = await call_rpc(rpc, rpc_method, [addr, at_block])
     assert state_result == state[fixture_key], "Invalid state - %s" % state_error
 
 
@@ -224,19 +224,31 @@ RPC_STATE_LOOKUPS = (
 )
 
 
-def validate_account_state(rpc, state, addr, at_block):
+async def validate_account_state(rpc, state, addr, at_block):
     standardized_state = fixture_state_in_rpc_format(state)
     for fixture_key, rpc_method in RPC_STATE_LOOKUPS:
-        validate_account_attribute(fixture_key, rpc_method, rpc, standardized_state, addr, at_block)
+        await validate_account_attribute(
+            fixture_key,
+            rpc_method,
+            rpc,
+            standardized_state,
+            addr,
+            at_block
+        )
     for key in state['storage']:
         position = '0x0' if key == '0x' else key
         expected_storage = state['storage'][key]
-        assert_rpc_result(rpc, 'eth_getStorageAt', [addr, position, at_block], expected_storage)
+        await assert_rpc_result(
+            rpc,
+            'eth_getStorageAt',
+            [addr, position, at_block],
+            expected_storage
+        )
 
 
-def validate_accounts(rpc, states, at_block='latest'):
+async def validate_accounts(rpc, states, at_block='latest'):
     for addr in states:
-        validate_account_state(rpc, states[addr], addr, at_block)
+        await validate_account_state(rpc, states[addr], addr, at_block)
 
 
 def validate_rpc_block_vs_fixture(block, block_fixture):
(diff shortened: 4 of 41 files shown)


trinity/extensibility/plugin.py
```diff
@@ -6,17 +6,43 @@
...
+class PluginProcessScope(Enum):
 
-# TODO: we spec this out later
-PluginContext = Any
+    SHARED = auto()
+    MAIN = auto()
+    OWN = auto()
+
+
+class PluginContext:
```
veox: I vaguely remember at least `PluginContext(object)` being recommended somewhere...

pipermerriam: I believe that since python3 the two are equivalent.

cburgdorf:
>THis looks the same as something else I just reviewed....

Yep, Will probably have something more interesting here tomorrow ;)

cburgdorf:
I moved this to the Ada Lovelace milestone because I think we shouldn't rush it. There's a lot of code to review (not only in this PR but also in the new [lahja event bus library](https://github.com/cburgdorf/lahja))

cburgdorf:
Ok, I think I need a break and I'm reaching out for help. I've been trying whole day to work out the remaining issues but there are two things that stand out that I think are related.

1. Trinity hangs on shutdown. This is also reproducible with a minimal test case that basically just runs `event_bus.start()`. This will prevent a graceful shutdown.

2. The tests fail pretty much for the same reason. The event bus is preventing the tests from closing, so while they succeed in principle, the test runner hangs and is shut down after 10 minutes of inactivity.

Both issues indicate that there's a problem with the event bus in general. Even the [simple test in the lahja repo](https://github.com/cburgdorf/lahja/blob/c2f8811e0dc048255aec1f06fe17b7f525c665a9/tests/core/test_basics.py) shows that behaviour. It goes green but never closes.

I guess it has to do with the `Queue` but I tried pretty much every thing I could think of (and what I found in the docs and StackOverflow). Really happy for any help on this.

cburgdorf:
Ok, after lots blood, sweat and pain, the event bus isn't preventing the processes to shut down anymore. :sparkles: That said, for isolated plugins there's still messy error output on shutdown. I think I do know why. The shutdown routine is anything but clean there. But still, it's not *that* bad anymore ;)

So, this is finally passing CI :sparkling_heart: There's still at least one thing that I need to double check regarding `light` mode before this can get another round of review. I can see the light at the end of the tunnel though ;)

cburgdorf:
@pipermerriam I think this can be given a full review now. I'll put in some inline comments.

tests/trinity/core/json-rpc/test_ipc.py
```diff
@@ -399,18 +405,27 @@ def uint256_to_bytes(uint):
     assert result == expected
 
 
+def mock_peer_count(count):
```
cburgdorf: This is an ad-hoc way to EventBus interactions. I guess we can do much better :)

tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py
```diff
@@ -184,21 +184,21 @@ def result_from_response(response_str):
     return (response.get('result', None), response.get('error', None))
 
 
-def call_rpc(rpc, method, params):
+async def call_rpc(rpc, method, params):
```
cburgdorf: All these became `async` because the `execute` of the RPC Server became async (due to the event-bused interaction and also because I think it's the right thing to do)

trinity/plugins/builtin/json_rpc/plugin.py
```diff
@@ -0,0 +1,122 @@
...
+    def start(self) -> None:
+        self.logger.info('JSON-RPC Server started')
+
+        rpc = RPCServer(self.chain, self.context.event_bus)
+        # The IPCServer used to have its own event loop because a comment indicated
+        # it may otherwise run into deadlocks. Not sure yet what to do about that
+        # because having two distinct event loops seems to be problematic now that
+        # the RPCServer uses the eventbus to retrieve the peer count
+        ipc = IPCServer(rpc, self.context.chain_config.jsonrpc_ipc_path)
+        asyncio.ensure_future(ipc.run())
```
cburgdorf: This is the part that I'm most unsure about. This is in the `light` plugin - the one that remains in the `networking` process. Previously, the IPC server had its own event loop running in another thread. One way we could use to make that work is to have another endpoint (shared plugins only have one single `networking` endpoint so far) end then call `another_endpoint.connect(ipc_loop))`. I tried that end it works in theory but I wasn't sure if this is a real issue to invest time in.

I'd rather invest time into making sure we can get rid of this extra plugin sooner than later. Not sure.

trinity/rpc/main.py
```diff
@@ -116,7 +117,10 @@ def _get_result(self,
 
             method = self._lookup_method(request['method'])
             params = request.get('params', [])
-            result = method(*params)
+            if asyncio.iscoroutinefunction(method):
```
cburgdorf: All the RPC methods were synchronous so far. I changed this to support both.

trinity/utils/db_proxy.py
```diff
@@ -0,0 +1,34 @@
+from multiprocessing.managers import (
```
cburgdorf: This just moved here because we need it in both the node and the plugin

trinity/rpc/main.py
pipermerriam: I think all JSON-RPC methods should be changed to be async.

tests/trinity/conftest.py
```diff
@@ -51,6 +55,19 @@ def event_loop():
         loop.close()
 
 
+@pytest.fixture(scope='session')
```
pipermerriam: This seems potentially problematic to run at the session level due to leakage between tests?  Seems like it should be function scoped (also I'm surprised this works because I assumed the `event_loop` fixture was function scoped....

trinity/extensibility/plugin_manager.py
```diff
@@ -74,7 +113,55 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+        for plugin in self._plugin_store:
+
+            if not self._is_responsible_for_plugin(plugin):
+                continue
+
+            context = self._create_context_for_plugin(plugin)
+            context.args = args
+            context.boot_kwargs = boot_kwargs
+            context.chain_config = chain_config
+            plugin.context = context
```
pipermerriam: Might be good to use a method for this `Plugin.set_context()` to be 1) explicit about the mutation and 2) allow the plugin to respond if needed.

trinity/plugins/builtin/json_rpc/plugin.py
pipermerriam: We can do dirty intermediate stuff to keep this working but I'm still in the camp that it needs to remain functional.

trinity/rpc/main.py
cburgdorf: We can do that but it may look odd if the only async thing in these functions is the `async` keyword, no?

trinity/plugins/builtin/json_rpc/plugin.py
cburgdorf: I'm totally on board. It is functional and I'm not sure how I would prove that the extra thread is needed. Seems like something that may only become obvious in longer real world testing?

trinity/main.py
```diff
@@ -220,6 +256,10 @@ def trinity_boot(args: Namespace,
         # simply uses 'kill' to send a signal to the main process, but also because they will
         # perform a non-gracefull shutdown if the process takes too long to terminate.
         logger.info('Keyboard Interrupt: Stopping')
+        plugin_manager.shutdown()
+        networking_endpoint.stop()
+        main_endpoint.stop()
+        event_bus.stop()
```
cburgdorf: Note to self: `event_bus.stop()` should probably just call `stop()` on every `endpoint`

cburgdorf:
I rebased this in preparation of a merge after the next trinity release. However, I just noticed that stopping Trinity (`CTRL +C`) may still leave us with an ugly hanging process.

Will dive in.

```
^C    INFO  09-03 18:06:15        main  Keyboard Interrupt: Stopping
    INFO  09-03 18:06:15      server  Closing server...
  File "/usr/lib/python3.6/multiprocessing/queues.py", line 240, in _feed
    send_bytes(obj)
  File "/usr/lib/python3.6/multiprocessing/connection.py", line 200, in send_bytes
    self._send_bytes(m[offset:offset + size])
  File "/usr/lib/python3.6/multiprocessing/connection.py", line 404, in _send_bytes
    self._send(header + buf)
  File "/usr/lib/python3.6/multiprocessing/connection.py", line 368, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pipe
    INFO  09-03 18:06:15         ipc  Sent SIGINT to process 26308, waiting 10 seconds for it to terminate
    INFO  09-03 18:06:15         ipc  Process 26308 has already terminated
    INFO  09-03 18:06:15        main  DB server process (pid=26308) terminated
Error in atexit._run_exitfuncs:
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 40, in _python_exit
    INFO  09-03 18:06:16         ipc  Sent SIGINT to process 26312, waiting 10 seconds for it to terminate
    t.join()
  File "/usr/lib/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
Exception ignored in: <coroutine object CancelToken.wait.<locals>._wait_for_first at 0x7ff347d0bbf8>
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venvx/lib/python3.6/site-packages/cancel_token/token.py", line 106, in _wait_for_first
    await cast(Awaitable[Any], future)
  File "/usr/lib/python3.6/asyncio/tasks.py", line 454, in _wait_for_one
    f = yield from done.get()
  File "/usr/lib/python3.6/asyncio/queues.py", line 169, in get
    getter.cancel()  # Just in case getter is not done yet.
  File "/usr/lib/python3.6/asyncio/base_events.py", line 575, in call_soon
    self._check_closed()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 358, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Exception ignored in: <coroutine object BaseService.run at 0x7ff347d0b8e0>
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/p2p/service.py", line 112, in run
    await self.cleanup()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/p2p/service.py", line 185, in cleanup
    self._cleanup()
  File "/usr/lib/python3.6/asyncio/tasks.py", line 596, in gather
    fut = ensure_future(arg, loop=loop)
  File "/usr/lib/python3.6/asyncio/tasks.py", line 519, in ensure_future
    task = loop.create_task(coro_or_future)
  File "/usr/lib/python3.6/asyncio/base_events.py", line 283, in create_task
    self._check_closed()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 358, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
sys:1: RuntimeWarning: coroutine 'IPCServer._cleanup' was never awaited
    INFO  09-03 18:06:16         ipc  Process 26312 has already terminated
    INFO  09-03 18:06:16        main  Networking process (pid=26312) terminated
```

p2p/events.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class PeerCountRequest(BaseEvent):
+    pass
+
+
+class PeerCountResponse(BaseEvent):
+
+    def __init__(self, peer_count: int) -> None:
+        super().__init__()
```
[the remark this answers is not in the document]
cburgdorf: After learning https://github.com/ethereum/py-evm/pull/1079#discussion_r214506684 I realize the `__init__` in `BaseEvent` isn't even needed as all it does is initializing things to `None`

tests/trinity/conftest.py
cburgdorf: Just to follow up on this ´function` scope works fine for the `event_bus` fixture but for the (existing) `event_loop` fixture it seems to need `session` scope or otherwise quite a bunch of tests fall apart. Not going to dig deeper as part of this PR.

trinity/rpc/main.py
cburgdorf: Additional question here. If we make all the sync methods `async` but then inside do synchronous operations doesn't that mean we are doing the cardinal sin of blocking the event loop?

cburgdorf:
Problems with the unclean shutdown are resolved :sparkles:

trinity/rpc/main.py
```diff
@@ -135,11 +139,11 @@ def _get_result(self,
         else:
             return result, None
 
-    def execute(self, request: Dict[str, Any]) -> str:
+    async def execute(self, request: Dict[str, Any]) -> str:
```
carver: It really makes me uneasy how async spreads like a virus. Now any package that wants to use RPC must be async also. It also makes REPL interaction a PITA.

Imagine if the file object returned by `open()` _only_ had async methods. That would be a net negative, right?

I'm not going to advocate to change this back, but... my spider sense is tingling that we will have to do more work in the future to either unravel the mess or support multiple entry points (async and not async).

trinity/rpc/main.py
carver: Either way (async or not) the synchronous block bodies will block the event loop, unless you farm it out to an executor. Not `await`ing here won't change that.

trinity/rpc/main.py
cburgdorf: I kinda feel your pain but here's my personal perspective on it.

There's a difference between *event loop based async I/O* in general and `async` / `await` as a pattern for consumption of the former. I feel your pain with `async` / `await` as a pattern but I don't think there is a way to get around *event loop based async I/O* in general.

I think event loop based async I/O is the best way to build heavy I/O dependent software like this especially in a language like Python. Because what else could we do?

1. If we just go synchronous it basically means all I/O is going to run sequentially -> slow

2. Threading out doesn't help much because of the GIL.

Hence, I think that the pattern of event loop based async I/O is fine by itself.

It is the pattern that browsers (as well as nodejs) use since forever and pretty much every browser API (apart from `prompt` which really locks the entire screen) is async because everything else will ruin the user experience. If you write blocking code in the browser, the browser will tell you that "this script is "hanging" and asks you to kill it".

However, **how** we deal with event loop based async I/O is a different story and for most of my life Futures (or Promises for that matter) as well as Observables have been my goto  answer to deal with it. Now I'm slowly adapting to `async` / `await` and part of me likes it and part of me doesn't like it.

However, in a language such as Python working with Futures doesn't feel very welcoming either (`future.add_done_callback(some_fn_that_I_wish_I_could_write_as_a_lambda_but_couldnt`). I guess Observables (RxPy) make it a little nicer (I'm biased I worked with Rx and contributed to various parts since 2010) but still suffers a bit from the fact that lambdas just can't reveal their full potential in Python.

But as I said, I hate the fact that `async` methods spread like a virus as much as you do. That's also why in `lahja` I made it so that you can consume events both ways. Using `async` or callback based.

Because the latter has its place e.g. in situations where you want to consume events without having to `await` it and forcing the function that consumes it to be `async` as well. For instance, we use that in `main.py`.

https://github.com/ethereum/py-evm/blob/dd304cc4ace18116c8f9fb4bbf9161319c9cb4c3/trinity/main.py#L245-L254

trinity/rpc/main.py
cburgdorf: Yes, but that is assuming that synchronous and asynchronous functions run in the same thread. We *could* run an event loop in a separate thread (as it was before for the IPCServer) and then I believe all functions that are consumed via `async` / `await` would run on the event loop whereas all others would run blocking in the other thread. However, if the `async` functions do blocking I/O you are still blocking the event loop.

cburgdorf:
I've seen `py36-rpc-state-byzantium` fail 50 % of the times in this PR. So this makes me think there is still something about it that is causing this. No clue what that would be though.

tests/trinity/core/json-rpc/test_ipc.py
pipermerriam: :+1: I suspect a pattern will emerge once this gets some attention.

cburgdorf:
>There's a lot of `Task was destroyed but it is pending!`

Yep, and I'm *trying to believe* that's okay for these tests. But the truth is, I don't really know :sweat_smile: 

>and it's getting OOM-killed

Yes, I think so, too. Now the question is: Why am I seeing this more often on this PR than on other PRs. This makes me uneasy and I'm starting to believe there may actually be an issue in terms of memory usage introduced by this PR. I find this quite hard to debug.

cburgdorf:
@veox Oh actually, seeing these event bus related `Task was destroyed but it is pending!` messages as part of the RPC tests is suspicious. I think the eventbus shouldn't even be involved in these test. Investigating.

pipermerriam:
This *might* be unrelated to this PR but, I'm curious if we should move the `connection_handler` and `connection_loop` inside of the `IPCServer` class.  I think I understand why they were initially written separately, but with the advent of the `BaseService.run_task` and the `BaseService.wait` API I think that reading them would be a lot cleaner if they were integrated as part of the actual service class.

Main reasons:

1. Better readability since `self.wait` is easier to read than `cancel_token.cancellable_wait`
2. Task cancellation when the service exits.

cburgdorf:
@pipermerriam I can look into that. However, I found the cause for the failed RPC tests but its only the beginning of a broader hunt. 

If I change the `event_bus` fixture from `function` scope back to `session` scope, everything works out well. Now the thing is, the `event_bus` fixture isn't even used for any of these RPC test. But what seems to happen with `function` scope is that it is running for every single test even if none of these tests should use that fixture (and it isn't `autouse` either). Not sure why that is. But maybe that is the same reason why our `event_loop` fixture uses `session` scope?

cburgdorf:
Ah, no! I was wrong. It is indirectly used through the `ipc_server` fixture.

cburgdorf:
So either `session` or `module` seems to work where `module` at least provides a bit more isolation than `session`.

Anyway, I started some PoC work on top of this that will maybe allow us to also use an isolated plugin for the JSON-RPC Server in light mode. First results look promising so far. Will explore further tomorrow.

cburgdorf:
@pipermerriam @carver This is pretty much ready but tomorrow I'll write an alternative "less magical"  version of the `LightPeerChainBridgePlugin` to have us look at the trade offs.

cburgdorf:
@pipermerriam I think the main point left open here is the proper handling of cancellations (also, I moved it to a `BaseService` and use `run_daemon_task` already). I feel that can be addressed as part of https://github.com/ethereum/py-evm/issues/1284 that I would soon start looking into.

How do you feel about taking it as is?

cburgdorf:
@pipermerriam this looks weird. I just rebased this and now the integration test caught this:

```
"p2p.exceptions.DecryptionError: Invalid frame mac: expected b'\\xde\\xcag\\x0e\\x94\\xdbkC\\xa8\\xc4\\xb8\\x10\\xdb@\\xf8o', got b'\\xdd\\x16\\xcea]B\\xbd1\\x8a\\x0f\\xa7\\xe8B\\x87\\xf9\\xce'\n"
```

Looks totally random to me. https://circleci.com/gh/ethereum/py-evm/77685?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I'm out of town tomorrow until evening. Guess I'll investigate that when I come home.

cburgdorf:
Ok, turns out that `DecryptionError` was indeed random. After all, the integration test boots the node in an environment with internet access which means these things can happen. Forgot that for a moment.

pipermerriam:
two things, I never finished reviewing this (will try to do it tonight so you're clear to move forward tomorrow)

second, make an issue for that decryption error as thoseshouldn't escape the trinity process.

trinity/rpc/format.py
```diff
@@ -125,13 +126,22 @@ def block_to_dict(block: BaseBlock,
...
+            @functools.wraps(func)
+            async def async_formatted_func(self: Any, *args: Any) -> Callable[..., Any]:
+                if len(formatters) != len(args):
+                    raise TypeError("could not apply %d formatters to %r" % (len(formatters), args))
+                formatted = (formatter(arg) for formatter, arg in zip(formatters, args))
+                return await func(self, *formatted)
+            return async_formatted_func
+        else:
+            @functools.wraps(func)
+            def formatted_func(self: Any, *args: Any) -> Callable[..., Any]:
```
pipermerriam: Is this still necessary,  quick glance suggests that all of the wrapped methods are now coroutines.

trinity/chains/light.py
```diff
@@ -135,14 +135,19 @@ def get_block(self) -> BaseBlock:
         raise NotImplementedError("Chain classes must implement " + inspect.stack()[0][3])
 
     def get_block_by_hash(self, block_hash: Hash32) -> BaseBlock:
+        raise NotImplementedError("Use coro_get_block_by_hash")
+
+    async def coro_get_block_by_hash(self, block_hash: Hash32) -> BaseBlock:
         header = self._headerdb.get_block_header_by_hash(block_hash)
-        return self.get_block_by_header(header)
+        return await self.get_block_by_header(header)
```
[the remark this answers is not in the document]
cburgdorf: yep, that was meant to go to `coro_get_block_by_header`

trinity/rpc/format.py
cburgdorf: yep, will change that.

PR #1216, opened 2018-08-26
WIP: implementation of ethstats plugin

evgeniuz:


evgeniuz:
Started implementation of ethstats plugin for trinity, added basic stuff for plugin, implemented some websockets interaction with fake data for now. Plan to add stuff that doesn't depend on py-evm/blockchain like latency reporting, OS, client version etc., adding more logging and some documentation. After that will proceed with addition of blockchain-dependendent stats (most complicated part).

setup.py
```diff
@@ -42,6 +42,7 @@
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.5",
         "web3==4.4.1",
+        "websockets==5.0.1",
```
veox: IMO better specify `"websockets>=5.0.0,<6.0.0"` - unless there's a specific reason for `5.0.1`?..

`websockets` is already pulled in as a nested dependency, so this'll make it easier to avoid breaks.

trinity/plugins/registry.py
```diff
@@ -28,4 +31,5 @@ def is_ipython_available() -> bool:
     AttachPlugin() if is_ipython_available() else AttachPlugin(use_ipython=False),
     FixUncleanShutdownPlugin(),
     TxPlugin(),
+    EthstatsPlugin(),
```
veox: Alphabetical order would likely be better. Here and when importing the plugin above.

setup.py
pipermerriam: Hrm, as long as this is within the range that `web3.py` uses, specific version pinning is actually appropriate.  We consider trinity to be an *Application* and thus, we've chosen to pin to specific versions to avoid odd errors from changes to other libraries which would otherwise cause breakage.

I'm open to discussion on the validity of this tactic since one *could* make the argument that pinning between major versions should be sufficient.  We've also talked about pinning within `minor` versions as a middle ground.

voith: FYI, python 3.7 support for `websockets` was added in `6.0.0`.

pipermerriam: Looks like we'll need to bump `web3.py` to something `>4.5.0` (but still pin at exact version)

pipermerriam: If any test failures show up from that version bump, lets address them separate from this PR.

trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def handle_event(self, activation_event: BaseEvent) -> None:
+        if isinstance(activation_event, TrinityStartupEvent):
+            self.is_enabled = activation_event.args.ethstats
+
+    def should_start(self) -> bool:
+        return self.is_enabled
+
+    def start(self, context: PluginContext) -> None:
+        asyncio.ensure_future(self.stats_client.run())
```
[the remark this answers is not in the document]
cburgdorf: There's nothing yet. It's probably worth opening up an issue to collect specific use cases that need to be addressed. In this case, are you referring to how we would stop/restart this plugin or how it would recover from failure to auto-restart or something completely different?

setup.py
evgeniuz: Will leave this for now as is (pinned to 5.0.1). Can be bumped later along with web3.py.

Bumping web3.py seems out of scope of this issue, but I can do it if needed.

trinity/plugins/registry.py
evgeniuz: Fixed, changes will visible be when I update branch.

trinity/plugins/builtin/ethstats/stats_client.py
```diff
@@ -0,0 +1,132 @@
+import asyncio
+import json
+
+from datetime import datetime
```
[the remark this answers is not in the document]
evgeniuz: Fixed, changes will visible be when I update branch.

trinity/plugins/builtin/ethstats/stats_client.py
```diff
@@ -0,0 +1,132 @@
...
+        self.stats_server_url = stats_server_url
+        self.stats_server_secret = stats_server_secret
+
+        self.node_id = '#some_id'
+
+    async def _run(self) -> None:
+        await self.connection_loop()
+
+    async def connection_loop(self) -> None:
+        while True:
```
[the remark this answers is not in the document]
evgeniuz: Fixed, changes will visible be when I update branch.

trinity/plugins/builtin/ethstats/stats_client.py
```diff
@@ -0,0 +1,132 @@
...
+            try:
+                self.logger.info('Connecting...')
```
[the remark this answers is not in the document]
evgeniuz: Fixed, changes will visible be when I update branch.

trinity/plugins/builtin/ethstats/stats_client.py
```diff
@@ -0,0 +1,132 @@
...
+                async with websockets.connect(self.stats_server_url) as websocket:
+                    self.websocket = websocket
+                    await self.connection_handler()
+            except websockets.ConnectionClosed as e:
+                self.logger.warning(f'Connection is closed - code: {e.code}, reason: {e.reason}.')
+
+            self.logger.info('Reconnecting in 5s...')
+            await asyncio.sleep(5)
```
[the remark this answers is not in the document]
evgeniuz: Fixed, changes will visible be when I update branch.

trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -0,0 +1,48 @@
...
+from trinity.plugins.builtin.ethstats.stats_client import (
+    StatsClient,
+)
+
+
+class EthstatsPlugin(BasePlugin):
+
+    def __init__(self) -> None:
+        self.is_enabled: bool = False
+        # TODO: make server/secret configurable
```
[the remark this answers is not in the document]
evgeniuz: There are multiple things that need to be configured:

- node ID;
- contact information;
- API endpoint (required);
- API secret (required).

Maybe something else I haven't noticed yet. It seems impractical to have CLI argument for all of them, I think maybe registering these in configuration is an option? (not sure if plugin can inject configuration yet). Or should I add CLI for required ones and allow configuration of others through environment variables?

trinity/plugins/builtin/ethstats/stats_client.py
```diff
@@ -0,0 +1,132 @@
...
+    async def _run(self) -> None:
+        await self.connection_loop()
+
+    async def connection_loop(self) -> None:
+        while True:
+            try:
+                self.logger.info('Connecting...')
+                async with websockets.connect(self.stats_server_url) as websocket:
+                    self.websocket = websocket
+                    await self.connection_handler()
```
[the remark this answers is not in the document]
evgeniuz: Fixed, changes will visible be when I update branch.

evgeniuz:
I've fixed some of the comments, also added some basic support for static things (like versions etc.) and proceeded with adding some basic blockchain stats (latest block, peer count) by getting them from `PeerPool` and `BaseChain` resources received through events similar to `TxPool` plugin. I'll add these changes to branch when they will be minimally working.

pipermerriam:
@evgeniuz lemme know when you'd like another round of review.

trinity/plugins/builtin/ethstats/plugin.py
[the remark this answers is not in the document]
pipermerriam: CLI args for all is fine.  It's not that much configuration.  Eventually we'll have a better way to do heavier config like this, but for now, CLI is the way to go.

trinity/plugins/builtin/ethstats/plugin.py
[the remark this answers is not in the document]
carver: I'm interested in the clean-shutdown aspect (which might be what he meant too). In this case, there is no cancel token being passed to the stats client from any parent service, so nothing causes it to shut down smoothly.

I guess the plugin could listen for the shutdown event and call `await self.stats_client.cancel()`. Any other good options?

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -0,0 +1,126 @@
...
+
+        self.client = construct_trinity_client_identifier()
+        self.os = platform.system()
+        self.os_v = platform.release()
+
+    async def response_handler(self) -> None:
+        async for message in self.websocket:
+            command, data = self.stat_recv(message)
+            if command == 'node-pong':
+                timestamp = round(datetime.datetime.now().timestamp() * 1000)
```
pipermerriam: I assume you want `UTC` time which is `datetime.datetime.utcnow()`

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -0,0 +1,126 @@
...
+            },
+        }])
+
+    async def send_stats(self, stats) -> None:
+        await self.send_queue.put(['stats', {
+            'stats': stats,
+        }])
+
+    async def send_node_ping(self) -> None:
+        timestamp = round(datetime.datetime.now().timestamp() * 1000)
```
pipermerriam: Same here, also, you might wrap this in a small utility since it's repeated in two places.

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -0,0 +1,126 @@
...
+                latency = (timestamp - data['clientTime']) // 2
+                await self.send_latency(latency)
+    
+    async def request_handler(self) -> None:
+        while True:
+            message = await self.send_queue.get()
+            [command, data] = message
+            await self.stat_send(command, data)
+
+    async def connection_handler(self) -> None:
```
pipermerriam: It might be good to move this into the `EthStatsService` so that you can use `BaseService.run
_task()` to run the `request_handler` and `response_handler`.  Otherwise, they won't be properly shutdown.

trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -0,0 +1,86 @@
...
+        super(EthstatsService, self).__init__(*args, **kwargs)
+
+        self.server_url: str = server_url
+        self.server_secret: str = server_secret
+
+        self.chain: BaseChain = chain
+        self.peer_pool: PeerPool = peer_pool
+
+    async def _run(self) -> None:
+        await self.connection_loop()
```
pipermerriam: minor but maybe just drop the `connection_loop` function in favor of putting the logic here in the `_run` method?

trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -0,0 +1,86 @@
...
+                            asyncio.ensure_future(self.statistics_handler(client)),
+                        ],
+                        return_when=asyncio.FIRST_COMPLETED,
+                    )
+
+                    for task in pending:
+                        task.cancel()
+
+            except websockets.ConnectionClosed as e:
+                self.logger.warning(f'Connection to {self.server_url} is closed - code: {e.code}, reason: {e.reason}.')
```
pipermerriam: Logging statements need to still use `%` formatting.  The reason is that logging aggregators cannot group the same logging statements if the text is pre-formatted.

trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -0,0 +1,86 @@
...
+    async def statistics_handler(self, client) -> None:
+        await client.send_hello()
+        # await client.send_history()
+        await client.send_pending()
+
+        while self.is_operational:
+            block = self.chain.get_canonical_head()
+            peers = len(self.peer_pool)
+
+            await client.send_node_ping()
```
pipermerriam: Each of these `await` statements need to be wrapped in `self.wait` so that they will be properly cancelled.

trinity/plugins/builtin/ethstats/ethstats_client.py
cburgdorf: Yes, either that or derive the `EthstatsClient` from `BaseService` as well and pass the token from the `EthStatsService` so that it will shutdown in tandem.

trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -0,0 +1,63 @@
...
+)
+
+
+class EthstatsPlugin(BasePlugin):
+
+    def __init__(self) -> None:
+        self.is_enabled: bool = False
+
+        self.chain: BaseChain = None
+        self.peer_pool: PeerPool = None
```
cburgdorf: It seems that getting the `len()` of the `PeerPool` is currently the only thing that this plugin needs from the `PeerPool`. There's an alternative way of retrieving the number of peers that works cross process.

https://github.com/ethereum/py-evm/blob/d225f4a26811aa15f450378aa8d618caa573b158/trinity/rpc/modules/net.py#L20-L21

I think we should aim for this plugin to become a `BaseIsolatedPlugin` similar to how the `JSON-RPC Server` works. I don't see that as a definitive requirement to be addressed before merging as the plugin doesn't run by default anyway but I *think* this is something that we should do. There will probably be many things that isolated plugins can't do right now but our cross process support is improving and in the long term most things including stuff like the transaction  pool will most likely end up as isolated plugins.

Here's the `JSON-RPC` server to get you started:

https://github.com/ethereum/py-evm/blob/d225f4a26811aa15f450378aa8d618caa573b158/trinity/plugins/builtin/json_rpc/plugin.py#L30-L70

trinity/plugins/builtin/ethstats/plugin.py
```diff
@@ -0,0 +1,63 @@
...
+            elif activation_event.resource_type is BaseChain:
+                self.chain = activation_event.resource
+
+    def should_start(self) -> bool:
+        return all((self.is_enabled, self.chain is not None, self.peer_pool is not None))
+
+    def start(self) -> None:
+        service = EthstatsService('ws://localhost:3000/api', 'SECRET', self.chain, self.peer_pool)
+
+        asyncio.ensure_future(service.run())
```
cburgdorf: Plugins that run in the networking process do currently not receive any notification about cancellation. In the case of e.g. the `TxPool` plugin that isn't a real issue as that one chains with the `CancelToken` of the node and hence at least gets shut down cleanly with the node.

https://github.com/ethereum/py-evm/blob/d225f4a26811aa15f450378aa8d618caa573b158/trinity/plugins/builtin/tx_pool/plugin.py#L84

[There's a PR coming](https://github.com/ethereum/py-evm/pull/1309) that will guarantee a cleaner teardown story for networking plugins, however, if this would move into a `BaseIsolatedPlugin` the teardown story would change a bit as these plugins run in their own process and receive `SIGINT` / `SIGTERM` that need to be handled.

https://github.com/ethereum/py-evm/blob/d225f4a26811aa15f450378aa8d618caa573b158/trinity/plugins/builtin/json_rpc/plugin.py#L67

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: @cburgdorf That's approach I've working on now, it seems a bit better to handle responses/requests in client completely and EthstatsService will handle actually getting stats and sending them to client.

trinity/plugins/builtin/ethstats/plugin.py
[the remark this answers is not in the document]
evgeniuz: Added CLI/environment configuration as suggested.

trinity/plugins/builtin/ethstats/plugin.py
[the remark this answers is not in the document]
evgeniuz: Reimplemented whole plugin as BaseIsolatedPlugin, hopefully this should resolve issues with graceful shutdown.

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: Fixed to use `utcnow()` (this actually doesn't matter, as it calculates delta between two calls, but it's good idea to use UTC here to avoid potential issues).

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: Wrapped in utility.

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: Reimplemented `EthstatsClient` as child service.

trinity/plugins/builtin/ethstats/ethstats_service.py
evgeniuz: Moved `connection_loop` into `_run`.

trinity/plugins/builtin/ethstats/ethstats_service.py
evgeniuz: Changed all logging to use `%` formatting.

trinity/plugins/builtin/ethstats/ethstats_service.py
evgeniuz: Not sure if I should wrap every async call in `self.wait`, but now `server_handler` and `statistics_handler` are wrapped in `self.wait_first` and seem to cancel properly (or maybe I missed something :).

trinity/plugins/builtin/ethstats/plugin.py
evgeniuz: Refactored plugin to be `BaseIsolatedPlugin`, used suggested method to get peer count.

trinity/plugins/builtin/ethstats/plugin.py
evgeniuz: Refactored plugin to be `BaseIsolatedPlugin`.

trinity/plugins/builtin/ethstats/ethstats_client.py
pipermerriam: If all that is needed is a whole number delta in seconds, then this should be updated to use [`time.monotonic()`](https://docs.python.org/3.5/library/time.html#time.monotonic)

pipermerriam: I think you can ignore my comment here, I mis-interpreted your comment.  `utcnow()` is correct.

cburgdorf:
>Since now plugin is isolated I cannot use `BaseChain` resource like `TxPool`. Should I get it same way as in `JsonRpcServerPlugin`?

Yes, that's the right way to do it. Will take a closer look at the PR tomorrow

trinity/plugins/builtin/ethstats/ethstats_service.py
cburgdorf: I *think* that should be fine.

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+# Returns UTC timestamp in ms, used for latency calculation
+def timestamp_ms():
+    return round(datetime.datetime.utcnow().timestamp() * 1000)
+
+
+class EthstatsMessage(typing.NamedTuple):
+    command: str
+    data: dict
```
cburgdorf: This probably needs to be `Dict[str, Any]`

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -0,0 +1,141 @@
...
+        node_id: str,
+        *args,
+        **kwargs,
+    ) -> None:
+        super().__init__(*args, **kwargs)
+
+        self.websocket = websocket
+        self.node_id = node_id
+
+        self.send_queue: asyncio.Queue = asyncio.Queue()
```
cburgdorf: I think this needs to be typed as `'asyncio.Queue[EthstatsMessage]'`

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: Fixed this.

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: Fixed this.

evgeniuz:
Fixed suggested type hinting and some more. I would like to add some basic chain stats before merging, though (block, hash, difficulty).

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -29,16 +36,15 @@ def __init__(
...
-        super().__init__(*args, **kwargs)
+        super().__init__(token)
 
         self.websocket = websocket
         self.node_id = node_id
 
-        self.send_queue: asyncio.Queue = asyncio.Queue()
-        self.recv_queue: asyncio.Queue = asyncio.Queue()
+        self.send_queue: asyncio.Queue[EthstatsMessage] = asyncio.Queue()
+        self.recv_queue: asyncio.Queue[EthstatsMessage] = asyncio.Queue()
```
cburgdorf: Did you try running the plugin after that change? Not sure if this works with Python 3.7 but at least with Python 3.6 you need to put that in quotes `asyncio.Queue[EthstatsMessage]'` which is an annoying limitation that applies to some generic builtin types.

cburgdorf:
>I would like to add some basic chain stats before merging, though (block, hash, difficulty).

Fine by me

trinity/plugins/builtin/ethstats/ethstats_client.py
evgeniuz: Working correctly for me in Python 3.6.6, but will add quotes around `EthstatsMessage` just in case.

evgeniuz:
Wrapped type parameter in quotes, added blockchain stats (block number, hash, difficulties), some version-based stats. Can be merged if everything is ok.

trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -43,8 +43,8 @@ def __init__(
         self.websocket = websocket
         self.node_id = node_id
 
-        self.send_queue: asyncio.Queue[EthstatsMessage] = asyncio.Queue()
-        self.recv_queue: asyncio.Queue[EthstatsMessage] = asyncio.Queue()
+        self.send_queue: asyncio.Queue['EthstatsMessage'] = asyncio.Queue()
+        self.recv_queue: asyncio.Queue['EthstatsMessage'] = asyncio.Queue()
```
cburgdorf: It is the entire `'asyncio.Queue[EthstatsMessage]'` that would need quotes. But if you say that it works without those maybe just leave them out. That said, I gotta double check that because everywhere else we had to quote `asyncio.Queue[T]` so far.

evgeniuz: Was curious about why that happened, found following:

https://www.python.org/dev/peps/pep-0526/#runtime-effects-of-type-annotations

Since that declaration is inside the method it gets treatment as a local variable and type definition is not evaluated, so can be any value (even undefined). `mypy` gets type definitions from `typeshed` where `asyncio.Queue` is defined, so both runtime and `mypy` check work without quotes. If I move that declaration to a class level, it will get evaluated and will fail, so in that context it will need quotes.

cburgdorf: Ha! Great find! Thanks for investigating. Let's do it without quotes then.

cburgdorf:
Hey @evgeniuz 

Thanks for all the work you put into this. I'm going to squash this and force push it here for another CI run as this one went red for unrelated reasons.

From here, I think it would be nice to add some test coverage with another PR.

veox:
@evgeniuz or @cburgdorf - did you happen to check if the plug-in actually works? I've detailed my difficulties in issue #1337.

If you did get it to work: against which remote server? A locally-run instance of the JS `ethstats` server, perhaps?..

cburgdorf:
I didn't but I'd expect @evgeniuz developed against *something*. However, I expect the implementation to have bugs and in general be an early start. I'm really looking forward to see some added tests (+ @evgeniuz jumping in to provide clarification on how to use it.)

In addition to that, it would be great to have some coverage in the docs about how to set this up. Notice that merging this was relatively low impact as the plugin is disabled by default and having it in the code base kinda exposes it more broadly leading to early adopters like you running into bugs and creating issues such as #1337 which in general is leading to further improvement :)

evgeniuz:
> If you did get it to work: against which remote server? A locally-run instance of the JS `ethstats` server, perhaps?..

Yes, that's what I was using, locally run `eth-netstats` instance. I think that your initial guess about SSL is correct, will try to look more closely into this and fix it.

PR #1217, opened 2018-08-28
eth/vm/logic/arithmetic: handle exp(X,0) == 1.

veox:
### What was wrong?

0^0 == 1, but this wasn't tested for.

### How was it fixed?

Caught by [`exp8.json` test](https://github.com/ethereum/tests/blob/10ab37c095bb87d2e781bcf112b6104912fccb44/VMTests/vmArithmeticTest/exp8.json) (not yet tracked in tests fixture on `master`).

Semi-accidentally caught in PR #1181, which updated the JSON test fixtures to a newer version, which does contain the `exp8.json` test. The failure got lost among others.

Can be seen in `py3{5,6}-vm` CI tests: for the EVM bytecode `0x600060000a600055`, which is:

```
000000: PUSH1 0x00
000002: PUSH1 0x00
000004: EXP
000005: PUSH1 0x00
000007: SSTORE
```

... there was a `15000` gas use delta (to that of expected): as if the storage was (re-)set to 0, not 1. Mentioned CI runs:

* [py35-vm](https://circleci.com/gh/ethereum/py-evm/59992?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)
* [py36-vm](https://circleci.com/gh/ethereum/py-evm/59987?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

#### Cute Animal Picture

Source: not sure; via [`@ano` at peepeth](https://peepeth.com/ano/peeps/QmfQHVNqCHzj73e9hXiiEC5jdxr2yJHS1er4mFBvnegu1u)

### DIFF
--- a/tests/core/opcodes/test_opcodes.py
+++ b/tests/core/opcodes/test_opcodes.py
@@ -124,6 +124,32 @@ def test_mul(vm_class, val1, val2, expected):
 
 
 @pytest.mark.parametrize(
+    'vm_class, base, exponent, expected',
+    (
+        (ByzantiumVM, 0, 1, 0,),
+        (ByzantiumVM, 0, 0, 1,),
+        (SpuriousDragonVM, 0, 1, 0,),
+        (SpuriousDragonVM, 0, 0, 1,),
+        (TangerineWhistleVM, 0, 1, 0,),
+        (TangerineWhistleVM, 0, 0, 1,),
+        (HomesteadVM, 0, 1, 0,),
+        (HomesteadVM, 0, 0, 1,),
+        (FrontierVM, 0, 1, 0,),
+        (FrontierVM, 0, 0, 1,),
+    )
+)
+def test_exp(vm_class, base, exponent, expected):
+    computation = prepare_computation(vm_class)
+    computation.stack_push(exponent)
+    computation.stack_push(base)
+    computation.opcodes[opcode_values.EXP](computation)
+
+    result = computation.stack_pop(type_hint=constants.UINT256)
+
+    assert result == expected
+
+
+@pytest.mark.parametrize(
     # Testcases from https://github.com/ethereum/EIPs/blob/master/EIPS/eip-145.md#shl-shift-left
     'vm_class, val1, val2, expected',
     (
--- a/eth/vm/logic/arithmetic.py
+++ b/eth/vm/logic/arithmetic.py
@@ -147,7 +147,9 @@ def exp(computation, gas_per_byte):
     bit_size = exponent.bit_length()
     byte_size = ceil8(bit_size) // 8
 
-    if base == 0:
+    if exponent == 0:
+        result = 1
+    elif base == 0:
         result = 0
     else:
         result = pow(base, exponent, constants.UINT_256_CEILING)


voith:
I have an unrelated question to this PR:
I was trying to understand this line:
https://github.com/ethereum/py-evm/blob/08674f611f4f692e76bbcd7ea1cefcdfd15c723a/eth/vm/logic/arithmetic.py#L155
In my understanding `mod=constants.UINT_256_CEILING` is used to prevent an overflow.
I was trying to see what would happen when base would be negative.
```python
In [0]: pow(-1, 3)
Out[0]: -1

In [1]: pow(-1, 3, 2**256)
Out[1]: 115792089237316195423570985008687907853269984665640564039457584007913129639935
```
is this intended behaviour?

veox:
@voith I think so?..

The yellowpaper is a bit terse on this, but it does specify _some_ signed arithmetic operations (e.g. `SDIV` in addition to `DIV` (division, `/`), or `SGT` in addition to `GT` (greater than, `>`); but not `SEXP` in addition to `EXP`).

It could be _assumed_ that the stack values are to be interpreted as unsigned integers, but I haven't found this stated explicitly anywhere.

There _is_ a section on precompile 5 ("arbitrary-precision exponentiation under modulo"), for which one could _assume_ the logic follows that of `EXP`. For this precompile, inputs are non-negative integers.

pipermerriam:
@voith I am pretty sure @veox is correct that arguments are treated as unsigned integers.

veox:
@pipermerriam Should be able to...

voith:
> For this precompile, inputs are non-negative integers.

Yeah, just tested some solidity code. solidity doesn't allow exponentiation when the base is negative!

veox:
@pipermerriam See commit e7b0af9 - 0^0==1, 0^1==0. Is this enough? Too much?..

tests/core/opcodes/test_opcodes.py
```diff
@@ -123,6 +123,32 @@ def test_mul(vm_class, val1, val2, expected):
...
+        (ByzantiumVM, 0, 1, 0,),
+        (ByzantiumVM, 0, 0, 1,),
+        (SpuriousDragonVM, 0, 1, 0,),
+        (SpuriousDragonVM, 0, 0, 1,),
+        (TangerineWhistleVM, 0, 1, 0,),
+        (TangerineWhistleVM, 0, 0, 1,),
+        (HomesteadVM, 0, 1, 0,),
+        (HomesteadVM, 0, 0, 1,),
+        (FrontierVM, 0, 1, 0,),
+        (FrontierVM, 0, 0, 1,),
```
carver: Side note: looks like it would be valuable to have a parametrized fixture in conftest that returns all mainnet VMs. Then this would reduce to two cases, and we wouldn't have to remember to update the test on every new fork, like:

```py
@pytest.mark.parametrize(
    'base, exponent, expected',
    (
        (0, 1, 0,),
        (0, 0, 1,),
    )
)
def test_exp(mainnet_vm, base, exponent, expected):
    computation = prepare_computation(mainnet_vm)
    computation.stack_push(exponent)
    computation.stack_push(base)
    computation.opcodes[opcode_values.EXP](computation)
    result = computation.stack_pop(type_hint=constants.UINT256)
    assert result == expected
```

pipermerriam: Lets wait for the chain-builder to land as it will make this a lot easier to do

pipermerriam: chain builder has landed so this should be pretty easy.

PR #1220, opened 2018-08-28
Extract tracker concept from normalizers

pipermerriam:
### What was wrong?

In the peer Exchange api, currently the normalizer classes are required to implement the logic for computing the item count.  This seems wrong.

### How was it fixed?

New tracker classes that handle this logic.

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/trackers.py
+++ b/trinity/protocol/common/trackers.py
@@ -1,0 +1,115 @@
+from abc import ABC, abstractmethod
+import logging
+from typing import (
+    Any,
+    cast,
+    Generic,
+    Optional,
+    TypeVar,
+)
+
+from eth.tools.logging import TraceLogger
+
+from p2p.protocol import (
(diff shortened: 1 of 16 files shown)


trinity/protocol/common/trackers.py
```diff
@@ -0,0 +1,112 @@
...
+
+        if response_size > request_size:
+            self.logger.warning(
+                "%s got oversized response.  requested: %d  received: %d",
+                type(self).__name__,
+                request_size,
+                response_size,
+            )
+        else:
+            self.total_missing += (request_size - response_size)
```
pipermerriam: Pointing out that we are now also tracking the number of missing items from their responses.

carver:
I think this is all in #1225 -- my review is there

pipermerriam:
I'll it all up in #1225

Issue #429, opened 2018-03-02
Implement p2p discovery v5


gsalgado:
The discovery v4 protocol is really bad at finding LES peers, so we had to resort to hard-coding known LES peers in the `PeerPool` we use with our `LightChain`. The latest release of geth (1.8.x) implements the [discovery v5 protocol](https://github.com/fjl/p2p-drafts), though, which is much better at finding peers that support a specific sub-protocol (e.g. LES), so we should implement that (probably even dropping v4) and change our `PeerPool` to use `p2p.DiscoveryProtocol` instead of hard-coded nodes.

gsalgado:
FWIW, I've started experimenting with it in https://github.com/gsalgado/py-evm/tree/discv5 a while ago

gsalgado:
I'm planning to resume work on this, and am trying to think of any reasons why we might want to have the option to run a node that supports only the discovery v4 proto. Given that v5 is (AIUI) a backwards-compatible extension of v4 (meaning a v5 node can still talk to nodes that only support v4), the only reason I can think of is that the new code written to support v5 may be more fragile/buggy than the existing v4 code, but then all of trinity is still in early alpha and definitely not stable/reliable, specially the current discv4 implementation, which I wrote in my early ethereum/asyncio days. I think it'd make more sense for us to simply extend our current implementation to match the v5 spec, instead of maintaining two separate implementations. @pipermerriam, @carver, any thoughts on this?

pipermerriam:
I'm fine with v5 only approach, we can always dig the v4 one back out of history if things don't go as planned.

carver:
Dropping v4-only sounds good to me :+1:

gsalgado:
They actually changed things and now support both v4 and v5 simultaneously, with a prefix to distinguish v5 messages. Also, v5 is not enabled by default, so if we would support only v5 (as I'd originally suggested), we wouldn't be able to talk to most of the nodes.

gsalgado:
We now have a basic implementation of discv5 that can be enabled via a command line flag, but [since they plan to rework geth's implementation](https://github.com/ethereum/go-ethereum/issues/17468#issuecomment-422341018) soon, it doesn't make sense to invest more effort on this until then

fjl:
Hi guys, I've created a tracking issue for the v5 spec over in the devp2p repo. We can discuss the protocol there :).

EBGToo:
[...]

PR #1221, opened 2018-08-29
Implement the discv5 wire protocol

gsalgado:
Basic handling of incoming v5 messages as well as methods to send
ping, pong, find_node and topic_query messages.

I'll soon submit separate PRs to keep track of tickets and topics
advertised by other peers, but wanted to get this out of the way
so that I can do some refactorings

Part of #429


p2p/discovery.py
```diff
@@ -419,15 +569,15 @@ def _pack(cmd_id: int, payload: List[Any], privkey: datatypes.PrivateKey) -> byt
...
+    expiration = rlp.sedes.big_endian_int.serialize(_get_msg_expiration())
     encoded_data = cmd_id + rlp.encode(payload + [expiration])
     signature = privkey.sign_msg(encoded_data)
     message_hash = keccak(signature.to_bytes() + encoded_data)
     return message_hash + signature.to_bytes() + encoded_data
 
 
-def _unpack(message: bytes) -> Tuple[datatypes.PublicKey, int, List[Any], bytes]:
-    """Unpack a UDP message received from a remote node.
+def _unpack(message: bytes) -> Tuple[datatypes.PublicKey, int, List[Any], Hash32]:
```
pipermerriam: Maybe go ahead and include the `_v4` suffix for these so that there's no ambiguity or accidental mis-use of v4/v5?  We can easily drop the suffixes once we're fully onto v5.

p2p/discovery.py
```diff
@@ -445,6 +595,37 @@ def _unpack(message: bytes) -> Tuple[datatypes.PublicKey, int, List[Any], bytes]
...
+
+def _pack_v5(cmd_id: int, payload: List[Any], privkey: datatypes.PrivateKey) -> bytes:
+    """Create and sign a discovery v5 UDP message to be sent to a remote node."""
+    cmd_id = to_bytes(cmd_id)
+    encoded_data = cmd_id + rlp.encode(payload)
+    signature = privkey.sign_msg(encoded_data)
+    return signature.to_bytes() + encoded_data
+
+
+def _unpack_v5(message: bytes) -> Tuple[datatypes.PublicKey, int, List[Any], Hash32]:
```
pipermerriam: Side note.  We should document the following terminology

- message: the raw bytes that are sent over the wire between peers.
- payload: the data embedded in the message
- command: the abstraction for a message which includes decoding/encoding of payloads.

In another issue @carver and I talked about trying to normalize our naming since I think we currently use message in place of payload in many places.

Issue #1222, opened 2018-08-29
Documentation with autofunction doesn't work on life docs


cburgdorf:
### What is wrong?

See live:

![image](https://user-images.githubusercontent.com/521109/44808114-ac2ad880-abcb-11e8-9320-79a894b6d061.png)

vs local

![image](https://user-images.githubusercontent.com/521109/44808133-b947c780-abcb-11e8-8d30-26d08d2f5eef.png)

### How can it be fixed

:hear_no_evil: :see_no_evil:

cburgdorf:
Ok, turns out, the live site is still broken. It's something else. 

![image](https://user-images.githubusercontent.com/521109/44812470-8b688000-abd7-11e8-9876-b92add74cce9.png)

Also fails for a bunch of other classes. I wonder why that didn't fail in our CI docs job. I'll investigate tomorrow.

PR #1223, opened 2018-08-29
Docs: Use automodule instead of autofunction to workaround RTD issue

cburgdorf:
### What was wrong?

Live docs are broken (#1222)

### How was it fixed?

This achieves the same visuals but doesn't use `autofunction` which I believe is causing the issue with the live docs.

#### Cute Animal Picture

### DIFF
--- a/docs/api/eth/tools/api.tools.fixtures.rst
+++ b/docs/api/eth/tools/api.tools.fixtures.rst
@@ -60,129 +60,138 @@ sequence of functions.
     :func:`~eth.tools.fixtures.expect`, expect to be passed a dictionary
     as their single argument and return an updated version of the dictionary.
 
-
-.. autofunction:: eth.tools.fixtures.fillers.common.setup_main_filler
-
-    This function kicks off the filler generation process by creating the general filler scaffold with
-    a test name and general information about the testing environment.
-
-    For tests for the main chain, the `environment` parameter is expected to be a dictionary with some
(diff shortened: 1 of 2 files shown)


docs/api/eth/tools/api.tools.fixtures.rst
```diff
@@ -60,129 +60,138 @@ sequence of functions.
...
+    (address, "storage", <storage slot>, <storage value>)
+    (address, "storage", {<storage slot>: <storage value>})
+    (address, {"balance", <account balance>})
+
+.. automodule:: eth.tools.fixtures.fillers
+  :noindex:
+  :members: execution
+
+
+For VM tests, this function specifies the code that is being run as well as the current state of
```
cburgdorf: @pipermerriam I actually think we should move these docstrings *into* the function because as we do with the other API docs. That ensures they stay more up to date and is more in line with the rest of the API section which all has the docstrings right at the implementation.

Didn't want to change that in this PR though as I just wanted to get the live docs fixed without changing the visuals.

pipermerriam: :+1: I actually did that with the builder ones but forgot to apply it here.  Are you good doing that cleanup in a follow up PR?

cburgdorf: Sure thing :+1:

Issue #1152, opened 2018-08-07
Enable state tests for constantinople


cburgdorf:
### What is wrong?

We currently do not run state tests against Constantinople. 

### How can it be fixed

- Pin the `fixtures` submodule to a newer version that contains the Constantinople tests (lookup which version geth is pinned against)
- Configure tests to run
- Make tests green

cburgdorf:
I think I pushed this one as far as I could (see https://github.com/ethereum/py-evm/pull/1181) Both, the Py-EVM and ethereum/tests need to do some work before we can finish that.

veox:
FTR, I'm trying to tease apart the issues, by first updating the `fixtures`, in PR #1224.

I've determined the last successful merge commit in `ethereum/tests` that `py-evm` still passes (`test_state_fixtures` in particular, since that's what's failing in PR #1181). Used `git bisect run` for that, with a [script](https://gist.github.com/veox/e626ccfe5702630d64716f719ca4b1b2); but it basically amounts to:

``` sh
source .virtualenv/*/bin/activate
cd fixtures
git checkout develop
git bisect start develop 47b09f42c0681548a00da5ab1c98808b368af49a --
git bisect run sh -c "cd .. && pytest --cache-clear --quiet --exitfirst --show-capture=no -k 'test_state_fixtures'"
git bisect log
```

I also tried every merge commit from current `develop` to `db8ae9de`, and they all have at least one failure.

-----

I think it'd be best to update the fixtures first before enabling `ConstantinopleVM`. Will try to step through the state failures one-by-one, to see which end (`py-evm` or `ethereum/tests`) is the culprit.

veox:
See https://github.com/ethereum/py-evm/pull/1224#issuecomment-419714693 for status of the approach after PR #1224 was merged.

There were seemingly changes in the test specification format, and it might not be prudent to try and "break through" those `ethereum/tests` commits without understanding the changes first; and whether they're still relevant, or can be "jumped over".

(FTR, I've layed off working on this.)

carver:
Confirmed finished in #1579

PR #1224, opened 2018-08-29
Update ethereum/tests fixtures

veox:
### What was wrong?

See issue #1152, in particular https://github.com/ethereum/py-evm/issues/1152#issuecomment-417284202.

* The `ethereum/tests` git submodule in `fixtures` is rather old:
  * 2017-12-08 - https://github.com/ethereum/tests/commit/47b09f42c0681548a00da5ab1c98808b368af49a - merge of PR https://github.com/ethereum/tests/pull/374.
* `py-evm` PR #1181 enables `ConstantinopleVM` testing with a much newer test set; having these changes "lumped together" makes it harder to attribute CI run failures to one or the other, at least at a cursory look.

### How was it fixed?

Incrementally:

* [Updates the submodule](https://github.com/ethereum/py-evm/pull/1224/commits/3fef4bd3d596eb4a16fc63ae4c6795bfaa03b641) to:
  * 2018-01-22 - https://github.com/ethereum/tests/commit/db8ae9de91c303caad3eb9354015c07a0ec3adc3 - merge of PR https://github.com/ethereum/tests/pull/401
  * all tests pass without `py-evm` modifications.
* [Updates the submodule](https://github.com/ethereum/py-evm/pull/1224/commits/b11a7b8887828d397aeb67dbcbc9e549f4ee1e81) to:
  * 2018-01-30 - https://github.com/ethereum/tests/commit/9b1f07c58a70d1b17c4489c49eb9bebf4a27d290 - merge of PR https://github.com/ethereum/tests/pull/405
  * [applies fix for complexity calculation](https://github.com/ethereum/py-evm/pull/1224/commits/7752f05d1df95e519d2daab3b971678d7244b3c6) - calling the `MODEXP` precompile was too cheap.
* [Updates the submodule](https://github.com/ethereum/py-evm/pull/1224/commits/11d61b005f73eef3f6ee441228b13ff9b97fa794) to:
  * 2018-03-01 - https://github.com/ethereum/tests/commit/61185fe4b8762118fe9ee318539683b47cb04ed6 - merge of PR https://github.com/ethereum/tests/pull/419
  * marks the upstream test `RevertInCreateInInit` as `xfail`, for both `StateTests` and `BlockchainTests`.
* [Updates the submodule](https://github.com/ethereum/py-evm/pull/1224/commits/fd26ec65c725d40aaab11e8d87ca8c14a56ef45d) to:
  * 2018-05-11 - https://github.com/ethereum/tests/commit/f4faae91c5ba192c3fd9b8cf418c24e627786312 - merge of PR https://github.com/ethereum/tests/pull/444
  * all tests still pass.

### WIP: current state

- [x] Evaluated PR https://github.com/ethereum/tests/pull/405: a particularly large modulus length in data to the `MODEXP` precompile. `py-evm` calculated the gas price wrong (this is _not_ the same bug Parity had). Discussion starts at https://github.com/ethereum/py-evm/pull/1224#issuecomment-417351843.
- [x] Evaluated PR https://github.com/ethereum/tests/pull/419: "revert in `CREATE` in init code, followed by returndatasize/copy". Discussion starts at https://github.com/ethereum/py-evm/pull/1224#issuecomment-418408696.
- [x] Took a glance at where the remaining 74 failures originate from - seemingly https://github.com/ethereum/tests/pull/437. Looks like it changed the fixture JSON generator; perhaps the file format, too. Best tackled in a follow-up PR.

### Cute Animal Picture

"Love this fixture..."

Cat hugs lamp shade

Source: [eleejhon @ imgur](https://imgur.com/gallery/E6jbC3G)

### DIFF
--- a/eth/precompiles/modexp.py
+++ b/eth/precompiles/modexp.py
@@ -35,7 +35,7 @@ def _compute_complexity(length):
             length ** 2 // 4 + 96 * length - 3072
         )
     else:
-        return 2 ** 2 // 16 + 480 * length - 199680
+        return length ** 2 // 16 + 480 * length - 199680
 
 
 def _extract_lengths(data):
--- a/tests/core/vm/test_modexp_precompile.py
+++ b/tests/core/vm/test_modexp_precompile.py
@@ -41,11 +41,11 @@ EIP198_VECTOR_C = decode_hex(
         (EIP198_VECTOR_A, 13056),
         (
             EIP198_VECTOR_C,
-            708647586132375115992254428253169996062012306153720251921480414128428353393856280,
+            10684346944173007063723051170445283632835119638284563472873463025465780712173320789629146724657549280936306536701227228889744512638312451529980055895215896,  # noqa: E501
         ),
     ),
 )
-def test_modexp_gas_fee_calcultation(data, expected):
+def test_modexp_gas_fee_calculation(data, expected):
     actual = _compute_modexp_gas_fee(data)
     assert actual == expected
 
--- a/tests/json-fixtures/test_blockchain.py
+++ b/tests/json-fixtures/test_blockchain.py
@@ -32,17 +32,33 @@ ROOT_PROJECT_DIR = os.path.dirname(os.path.dirname(os.path.dirname(__file__)))
 BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, 'fixtures', 'BlockchainTests')
 
 
+# These are tests that are thought to be incorrect or buggy upstream,
+# at the commit currently checked out in submodule `fixtures`.
+# Ideally, this list should be empty.
+# WHEN ADDING ENTRIES, ALWAYS PROVIDE AN EXPLANATION!
+INCORRECT_UPSTREAM_TESTS = {
+    # The test considers a "synthetic" scenario (the state described there can't
+    # be arrived at using regular consensus rules).
+    # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418775512
+    # The result is in conflict with the yellow-paper:
+    # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418800369
+    ('GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json', 'RevertInCreateInInit_d0g0v0_Byzantium'),  # noqa: E501
+}
+
+
 def blockchain_fixture_mark_fn(fixture_path, fixture_name):
     if fixture_path.startswith('bcExploitTest'):
         return pytest.mark.skip("Exploit tests are slow")
     elif fixture_path == 'bcWalletTest/walletReorganizeOwners.json':
         return pytest.mark.skip("Wallet owner reorganization tests are slow")
+    elif (fixture_path, fixture_name) in INCORRECT_UPSTREAM_TESTS:
+        return pytest.mark.xfail(reason="Listed in INCORRECT_UPSTREAM_TESTS.")
 
 
 def blockchain_fixture_ignore_fn(fixture_path, fixture_name):
     if fixture_path.startswith('GeneralStateTests'):
         # General state tests are also exported as blockchain tests.  We
-        # skip them here so we don't run them twice"
+        # skip them here so we don't run them twice
         return True
 
 
--- a/tests/json-fixtures/test_state.py
+++ b/tests/json-fixtures/test_state.py
@@ -138,6 +138,12 @@ SLOWEST_TESTS = {
 # Ideally, this list should be empty.
 # WHEN ADDING ENTRIES, ALWAYS PROVIDE AN EXPLANATION!
 INCORRECT_UPSTREAM_TESTS = {
+    # The test considers a "synthetic" scenario (the state described there can't
+    # be arrived at using regular consensus rules).
+    # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418775512
+    # The result is in conflict with the yellow-paper:
+    # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-418800369
+    ('stRevertTest/RevertInCreateInInit.json', 'RevertInCreateInInit', 'Byzantium', 0),
 }
 
 
--- a/tests/json-fixtures/test_state.py
+++ b/tests/json-fixtures/test_state.py
@@ -133,6 +133,19 @@ SLOWEST_TESTS = {
 }
 
 
+# These are tests that are thought to be incorrect or buggy upstream,
+# at the commit currently checked out in submodule `fixtures`.
+# Ideally, this list should be empty.
+# WHEN ADDING ENTRIES, ALWAYS PROVIDE AN EXPLANATION!
+INCORRECT_UPSTREAM_TESTS = {
+    # Upstream seems to specify that the precompile call fails, but `py-evm`
+    # handles it just fine.
+    # * https://github.com/ethereum/py-evm/pull/1224#issuecomment-417351843
+    # * https://github.com/ethereum/tests/pull/405#issuecomment-417855812
+    ('stReturnDataTest/modexp_modsize0_returndatasize.json', 'modexp_modsize0_returndatasize', 'Byzantium', 4),  # noqa: E501
+}
+
+
 def mark_statetest_fixtures(fixture_path, fixture_key, fixture_fork, fixture_index):
     fixture_id = (fixture_path, fixture_key, fixture_fork, fixture_index)
     if fixture_path.startswith('stTransactionTest/zeroSigTransa'):
@@ -144,6 +157,8 @@ def mark_statetest_fixtures(fixture_path, fixture_key, fixture_fork, fixture_ind
             return pytest.mark.skip("Skipping slow test")
     elif fixture_path.startswith('stQuadraticComplexityTest'):
         return pytest.mark.skip("Skipping slow test")
+    elif fixture_id in INCORRECT_UPSTREAM_TESTS:
+        return pytest.mark.xfail(reason="Listed in INCORRECT_UPSTREAM_TESTS.")
 
 
 def pytest_generate_tests(metafunc):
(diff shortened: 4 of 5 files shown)


veox:
The test I'm evaluating currently was introduced in https://github.com/ethereum/tests/pull/405 - you can try it by `git checkout 9b1f07c58a70d1b17c4489c49eb9bebf4a27d290` in `fixtures`.

I couldn't find a rationale for it, but looks like it deals with the `MODEXP` precompile ("contact" `0x5`, introduced in Byzantium, [EIP-198](https://eips.ethereum.org/EIPS/eip-198)), which computes `(B**E)%M`; and the particular case where `B==1`, `E==1`, `M==999188`.

Running

``` sh
pytest --cache-clear --capture=no -k "test_state_fixtures[/home/veox/src/py-evm/fixtures/GeneralStateTests/stReturnDataTest/modexp_modsize0_returndatasize.json:modexp_modsize0_returndatasize:Byzantium:4]"
```

fails on

https://github.com/ethereum/py-evm/blob/3fef4bd3d596eb4a16fc63ae4c6795bfaa03b641/tests/json-fixtures/test_state.py#L339

The helper contract puts some results of the call to the precompile into storage, and that produces a mismatch.

ATM, still looking for the culprit.

pipermerriam:
Maybe we can uncover what is wrong with opcode level test cases assuming we know what the expected result is for the inputs you mentioned?

pipermerriam:
FYI, the mismatched state root failure case is hard to debug.  It normally requires enabling `TRACE` level output, running the single test and inspecting the logging output, but in this case, it's likely different because of the *result* being wrong and thus the storage entry that the contract is storing being wrong.

Also maybe look to see if there are VM level tests for this as they would provide explicit output for the call.

veox:
(Currently on mobile.)

Took a glance y-day, there is a modexp test file, but no test case for this is particular.

Tried inspecting the `computation`, the result seemed wrong - 0, not 1; but does `computation` have the memory as it existed when exiting the EVM call frame?..

veox:
I think the computation fails to allocate 999188 bytes in memory. That is, the failure might be coming from the precompile impl-n.

The contract in the test `SSTORE`s three 32-byte values: whatever is returned from `CALLCODE`ing the precompile, whatever is at memory location `0x03e8` (hex for decimal `1000`), and value of `RETURNDATASIZE`. These are written to storage locations `0x01` to `0x03`.

I haven't gotten to inspecting actual storage post-computation - trying to find out how...

pipermerriam:
Have you looked at the `TRACE` level logging output.  my guess is that it'll go OOG when it tries to allocate such a large memory size.

veox:
OK, changed `logging` level in `tests/conftest.py`; gas consumption for calling the precompile is 23970528 (that's 23.97 million); tx gas allowance is 1 billion, and block limit is set at 10 billion for that test case - so no OOG.

I haven't checked yet if ~24 million is the precise expected gas consumption (the test case seems to be under-specified in this regard); the values put to storage, according to the `TRACE`:

* 1
* 0
* 999188

(Will link gist in a sec.)

veox:
What's more confusing to me is, doing this printf-style debugging, I can see that the return value length from `modexp()` (the precompile impl-n) is indeed 999188, but

``` python
print('computation.should_erase_return_data:', computation.should_erase_return_data)
print('len(computation._output):', len(computation._output))
```

at the end of `test_state_fixtures()`, which gives `False` and `0`.

So, seems like somewhere between the result being `return`ed from `modexp()` and the `computation` being returned from `vm.apply_transaction()`, `computation._output` gets truncated...

veox:
Maybe [found it](https://gist.github.com/veox/81dd5500946f5ce07b670220447195b0#file-pytest-out-py-L254)?..

When writing `999188` to storage slot 3 of the contract, there seems to be a value `4294967295` already there; so the gas cost is `5000` instead of `20000`.

veox:
I've been stuck on this single test case for too long, and, to the best of my understanding, `py-evm`'s operation is correct.

My inclination is to mark this consensus test `xfail` (`¯\_(ツ)_/¯`), and move on.

I've brought up my findings in https://github.com/ethereum/tests/pull/405 (where it was incuded).

pipermerriam:
Thanks for all the work you've put into this.  I think `xfail` is fine for now but we'll want to do a full re-visit of any `xfail` tests before constantinople because it's looking like we'll have a minimally viable client by around that time.

veox:
Squashed/rebased previous linting failure.

Also, accidentally found a typo in `py-evm` that caused the failure. After all, upstream was correct.

Running full state tests locally while CircleCI checks linting, will submit afterwards.

veox:
Possible candidate:

Narrowed it down to a difference in gas usage:

* in ["complementary" JSON from `BlockchainTests`](https://github.com/ethereum/tests/blob/61185fe4b8762118fe9ee318539683b47cb04ed6/BlockchainTests/GeneralStateTests/stRevertTest/RevertInCreateInInit_d0g0v0.json#L18), `gasUsed` is `0x01f250` (decimal: 127568); whereas
* `py-evm` shows `gas_used` for the block at 112568.

The difference is 15000, which would suggest a miscalculation of gas use on `SSTORE` by one of the implementations.

pipermerriam:
@veox not sure what commit you are on, but the net gas metering for SSTORE hasn't been implemented yet so if those are constantinople tests that could be the case.

If not, the simple case is a small miscalculation, but I've found that often the gas usage difference is actually indicative of a previous error where either a value should not have been written to storage (but it was) and thus a subsequent write to that storage slot costs less than it should, or some other variant of the same thing.

veox:
@pipermerriam Thanks for the suggestion! I'm trying not to assume things; this is a Byzantium test, so net gas metering shouldn't be in play yet (I hope so!).

I've edited the previous comment to indicate the commit used, for the sake of easier reproduction.

veox:
OK, here's what's going on in this test:

There is an account `0x6295ee1b4f6dd65047762f924ecd367c17eabf8f` with a storage location `0x00` already set to a non-zero value `0x01`, but with no code. How? Test case magic.

A call is made by another account, `0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b`, with a `CREATE` that _would_ place code at `0x6295..bf8f`... except it always `REVERT`s.

At this point - question: what is the value at storage location `0x00` of `0x6295..bf8f`?

* According to the test case (i.e. then-`cpp-ethereum`, now-`aleth` that generated the test); as well as `geth` and `parity`, - the storage is empty. So the two writes to empty storage consume 20000 gas each.
* According to `py-evm`, the storage looks as it has at the beginning of the test - that is, `0x00` has value `0x01`. So, the writes consume 5000 and 20000 gas for locations `0x00` and `0x01` respectively.

carver:
Is the tl;dr that a `CREATE` call should always start by wiping out the storage at the address (and that emptying the storage is *not* reverted on a `REVERT`)?

holiman:
See https://github.com/ethereum/EIPs/issues/684

veox:
@carver 

> Is the tl;dr that a `CREATE` call should always start by wiping out the storage at the address (and that emptying the storage is not reverted on a `REVERT`)?

It would seem so; except the yellowpaper (Section 7, "Contract Creation") either explicitly states otherwise, is contradictory, or is under-specified:

> The account’s nonce is initially defined as one, the balance as the value passed, **the storage as empty** (...)

But then later:

> execution may exit before the code has come to a natural halting state. In this (and several other) exceptional cases we say an out-of-gas (OOG) exception has occurred: The evaluated state is defined as being the empty set, ∅, and **the entire create operation should have no effect on the state, effectively leaving it as it was immediately prior to attempting the creation**.

The YP does not consider a "synthetic" case, like in this test.

-----

@holiman (near-copy from [gitter chat](https://gitter.im/ethereum/tests?at=5b900469ac25fd11b5e5f503))

It seems that https://github.com/ethereum/EIPs/issues/684 _also_ does not consider no-`nonce`, no-`code`, some-`storage` accounts. The proposition is somewhat ludicrous, but here we are!

I assume the test case is being run, and passes against all the implementations handled by Hive?..

pipermerriam:
@veox it appears you've got a handle on the why (your explanation makes sense to me).  Let me know if you need guidance on how to proceed with fixing things.

veox:
@pipermerriam Cool!

Definitely confused here, might take a day or two to formulate something concise.

-----

It seems to me that the test case is in contradiction to the yellowpaper. _If so_, then all the major implementations may be as well (need to verify). _If so_, then an exception/clarification to the yellowpaper might be inevitable; possibly complemented by a "correcting" EIP for Constantinople/Istanbul...

If my _guesses_ above are correct, and `aleth`/`geth`/`parity` implement it this way, then there is no recourse but to implement it the same way also - at least for the Byzantium block dynasty. :D (Would be nice to get confirmation from these other implementations...)

veox:
TL;DR: I won't be fixing this one. q:D

-----

The "correct" approach from the very beginning would have been immediately marking the test `xfail` when bumping `fixtures`. This highlights that a specific upstream test is failing, and makes sense no matter where the breakage originates from. I've done this now in https://github.com/ethereum/py-evm/pull/1224/commits/e02123e508368efd27c975f38646baeb4acf7978.

As to the test itself - IMO this _really_ should be fixed upstream. Leaving a newly-`CREATE`d contract's state wiped after a `REVERT` contradicts the yellowpaper.

This difference between `py-evm` and (probably) every other implementation is unfortunate. However, I think it doesn't present a danger of consensus failure on its own, as the initial state that the test starts with can't (currently) be arrived at "naturally", using regular consensus rules. (This is why I called it a "synthetic" test.)

However, it seems that this particular detail is a quirk in the test definition; it's probably testing for _something else_. By marking the test `xfail`, `py-evm` is no longer testing for _that_, too (but, note: only on Byzantium; the earlier chains are still being tested).

If someone's inclined, they can push `py-evm`'s behaviour towards that in the test, but I think this is _wrong_.

sorpaas:
We recently fixed a bug for checkpointing in https://github.com/paritytech/parity-ethereum/pull/9319, and I believe right now, parity's behavior on this is the same as py-evm -- on `REVERT` or out-of-gas, old storage values should re-appear, if that account ever existed.

veox:
Current state might be where the "simple" fixes end.

The next merge commit in `ethereum/fixtures` gives 74 failures for `test_state_fixtures()` alone, IMO better handled in a separate PR. Also, the discussion here is getting a bit too long, and hard to navigate.

Unmarking as `[WIP`].

Should be OK to squash-merge, especially since there's an interim commit that failed CI.

Nathapon8899:
#1181

Nathapon8899:
#9532

Nathapon8899:
veox : update - ethereum - test - fixtures

Issue #1218, opened 2018-08-28
Switching to a better peer mid-sync


carver:
### What is wrong?

@pipermerriam 's comment:

> As it stands I think that we may start syncing with peer A and then connect to peer B which has a higher TD. At this point we are locked into syncing with A even though B is the better peer.

(where better := higher total difficulty)

### How can it be fixed

Each time, after adding headers to the task queue (in `trinity.sync.common.chain`), check if the current peer is the highest td peer. If so, continue as before. Otherwise, break out of `_sync` and start again with the better peer.

carver:
Additionally, we should consider the throughput that the peer provides us, when we pick the best peer to sync with. The header-syncing peer can easily become our syncing bottleneck.

carver:
So I think this might be 3 issues, with 2 that depend on the third:

1. Syncing to the peer with the head with the best total difficulty
2. (fast) syncing with the peer that will give us the best header throughput
3. Both of those depend on: the ability to choose the best peer to sync with mid-sync

Maybe a fourth: something that intelligently trades off the first two goals. For example: we may not care about the absolute best total difficulty during fast sync, preferring header throughput instead. Unless we've caught up with the peer that has the best throughput, then we want the one with next-best throughput. etc. Then during regular sync, we pretty much only care about best TD.

PR #1226, opened 2018-08-30
Download all data in queues (instead of batches) while fast & regular syncing

carver:
### What was wrong?

Requests for block data and receipts were batched together. A slow peer can dramatically slow down this process.

### How was it fixed?

Roughly, put everything into queues so they can run independently, and one peer isn't waiting on another to finish.

Numbers vary widely, but I've seen as high as 1500 blocks imported in 5 s.

TODO:
- [x] hunt down an occassional `RuntimeError: Cannot restart a service that has already been cancelled`
- [x] manually test
- [ ] overnight manual test
- [x] make code more readable
- [x] linting
- [x] clean commit history

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -142,6 +142,25 @@ class BaseService(ABC, CancellableMixin):
                 self.logger.trace("Task %s finished with no errors", awaitable)
         self._tasks.add(asyncio.ensure_future(_run_task_wrapper()))
 
+    def run_daemon_task(self, awaitable: Awaitable[Any]) -> None:
+        """Run the given awaitable in the background.
+
+        Like :meth:`run_task` but if the task ends without cancelling, then this
+        this service will terminate as well.
+        """
+        async def _run_daemon_task_wrapper() -> None:
+            try:
+                await awaitable
+            finally:
+                if not self.is_cancelled:
+                    self.logger.debug(
+                        "%s finished while %s is still running, terminating as well",
+                        awaitable,
+                        self,
+                    )
+                    self.cancel_token.trigger()
+        self.run_task(_run_daemon_task_wrapper())
+
     def run_child_service(self, child_service: 'BaseService') -> None:
         """
         Run a child service and keep a reference to it to be considered during the cleanup.
@@ -178,10 +197,18 @@ class BaseService(ABC, CancellableMixin):
         async def _run_daemon_wrapper() -> None:
             try:
                 await service.run()
+            except OperationCancelled:
+                pass
+            except Exception as e:
+                self.logger.warning("Daemon Service %s finished unexpectedly: %s", service, e)
+                self.logger.debug("Daemon Service failure traceback", exc_info=True)
             finally:
                 if not self.is_cancelled:
                     self.logger.debug(
-                        "%s finished while we're still running, terminating as well", service)
+                        "%s finished while %s is still running, terminating as well",
+                        service,
+                        self,
+                    )
                     self.cancel_token.trigger()
 
         self.run_task(_run_daemon_wrapper())
--- a/trinity/sync/full/chain.py
+++ b/trinity/sync/full/chain.py
@@ -4,10 +4,12 @@ from asyncio import (
 )
 from concurrent.futures import CancelledError
 import enum
+from functools import (
+    partial,
+)
 from operator import attrgetter
 from typing import (
     Any,
-    Callable,
     Dict,
     List,
     Set,
@@ -74,6 +76,9 @@ BlockBodyBundle = Tuple[
     Tuple[Hash32, Dict[Hash32, bytes]],
     Hash32,
 ]
+
+# How big should the pending request queue get, as a multiple of the largest request size
+REQUEST_BUFFER_MULTIPLIER = 8
 
 
 class WaitingPeers:
@@ -157,19 +162,10 @@ class BaseBodyChainSyncer(BaseHeaderChainSyncer):
         self._body_peers = WaitingPeers(commands.BlockBodies)
 
         # Track incomplete block body download tasks
-        # - arbitrarily allow up to 4 requests-worth of headers queued up
+        # - arbitrarily allow several requests-worth of headers queued up
         # - try to get bodies from lower block numbers first
-        self._block_body_tasks = TaskQueue(MAX_BODIES_FETCH * 4, attrgetter('block_number'))
-
-    # TODO move this to BaseService if it gets broader usage and/or stays stable for long enough
-    def delayed_run_func(self, func: Callable[[], None], delay: float) -> None:
-        """
-        Run function `func` after `delay` seconds
-        """
-        async def delayed_run() -> None:
-            await self.sleep(delay)
-            func()
-        self.run_task(delayed_run())
+        buffer_size = MAX_BODIES_FETCH * REQUEST_BUFFER_MULTIPLIER
+        self._block_body_tasks = TaskQueue(buffer_size, attrgetter('block_number'))
 
     async def _assign_body_download_to_peers(self) -> None:
         """
@@ -207,26 +203,27 @@ class BaseBodyChainSyncer(BaseHeaderChainSyncer):
         non_trivial_headers = tuple(header for header in all_headers if not _is_body_empty(header))
         trivial_headers = tuple(header for header in all_headers if _is_body_empty(header))
 
-        try:
-            if len(non_trivial_headers) == 0:
-                self.logger.debug(
-                    "Block bodies for all %d headers were trivial, from %r..%r",
-                    len(all_headers),
-                    all_headers[0],
-                    all_headers[-1],
-                )
-                completed_headers = non_trivial_headers
-            else:
-                bundles, completed_headers = await self._get_block_bodies(peer, non_trivial_headers)
-                await self._block_body_bundle_processing(bundles)
+        if trivial_headers:
+            self.logger.debug(
+                "Found %d/%d trivial block bodies, skipping those requests",
+                len(trivial_headers),
+                len(all_headers),
+            )
 
-            self._mark_body_download_complete(batch_id, completed_headers + trivial_headers)
-        except BaseP2PError:
-            self._block_body_tasks.complete(batch_id, trivial_headers)
+        # even if trivial_headers is (), assign it so the finally block can run, in case of error
+        completed_headers = trivial_headers
+
+        try:
+            if non_trivial_headers:
+                bundles, received_headers = await peer.wait(
+                    self._get_block_bodies(peer, non_trivial_headers)
+                )
+                await self._block_body_bundle_processing(bundles)
+                completed_headers = trivial_headers + received_headers
+
+        except BaseP2PError as exc:
+            self.logger.info("Unexpected p2p perror while downloading body from peer: %s", exc)
             self.logger.debug("Problem downloading body from peer, dropping...", exc_info=True)
-        except Exception:
-            self._block_body_tasks.complete(batch_id, trivial_headers)
-            raise
         else:
             if len(non_trivial_headers) == 0:
                 # peer had nothing to do, so have it get back in line for processing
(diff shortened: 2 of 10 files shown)


tests/trinity/core/p2p-proto/test_stats.py
```diff
@@ -90,8 +90,10 @@ def mk_header_chain(length):
 
         stats = peer.requests.get_stats()
 
-        assert stats['BlockHeaders'].startswith('count={0}, items={0}, avg_rtt='.format(idx))
-        assert stats['BlockHeaders'].endswith(', timeouts=0')
+        header_stats = stats['BlockHeaders']
+        assert header_stats['count'] == idx
+        assert header_stats['items'] == idx
+        assert header_stats['timeouts'] == 0
```
pipermerriam: You might rebase this on top of #1225 so that we don't conflict here and also the stats available in that branch are easier to get at and probably better.

carver: > You might rebase this on top of #1225

Yeah, I just noticed that conflict while reviewing.

> also the stats available in that branch are easier to get at and probably better.

At first glance, I agree. I'll review it now.

carver:
Ok, getting closer. I *think* most/all of what's left is linting. Plus whatever the full CI run digs up.

carver:
I'm in the process of pulling most of these pieces out into other PRs. I'll rebase this after the others are merged:

- #1248 
- #1254

carver:
I rewrote most of `trinity/sync/full/chain.py` to fit the new model. So much has changed that it's not really worth looking at the diff on that file, just read the whole new file at: https://github.com/carver/py-evm/blob/block-data-queues/trinity/sync/full/chain.py

In my tests, we seem to be able to download bodies and receipts so fast that the new bottleneck is downloading the headers. You may see this yourself if you find something like this in the debug logs:

```
(in progress, queued, max size) of headers, bodies, receipts: [(0, 0, 1536), (320, 320, 512), (576, 576, 1024)]
```

That means that there are no new headers from which to start downloading bodies and receipts. All of the current download tasks (320 and 576, for bodies and receipts respectively) have current outstanding requests to peers.

This problem is similar to #1218 -- we probably have a peer with low header throughput, and we are stuck syncing on them until we give-up/time-out/catch-up. Some changes to make header sync look even more like body/receipt downloads might help (always getting the peer with the best throughput, on every new request).

trinity/sync/full/chain.py
```diff
@@ -61,16 +76,59 @@
...
+
+    async def get_fastest(self) -> ETHPeer:
+        wrapped_peer = await self._waiting_peers.get()
+        peer = wrapped_peer.original
+
+        # make sure the peer has not gone offline while waiting in the queue
+        while not peer.is_operational:
+            # if so, look for the next best peer
+            wrapped_peer = await self._waiting_peers.get()
+            peer = wrapped_peer.original
```
pipermerriam: This suffers from the same problem you have with the previous `peer.is_operational` check. I see the following options.

- remove the check entirely and rely on call sites to do the check (not ideal, probably error prone)
- infinite loop or recursive call (not ideal due to uncertain reliability of exit conditions)
- infinite loop or recursive call **with timeout**  (provides certainty around exit conditions and *guarantees* operation peer)

So maybe change to something like:

```python
async def get_fastest(self, timeout=DEFAULT_TIMEOUT):
    start_at = time.perf_counter()
    while True:
        remaining = timeout - (time.perf_counter() - start_at)
        peer = self.cancel_token.cancellable_wait(self._waiting_peers.get(), timeout=remaining)
        if peer.is_operational:
            return peer
```

I'm sure there is a way to do this without the cancel token....

trinity/sync/full/chain.py
```diff
@@ -85,314 +143,194 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+        # queue up any idle peers, in order of how fast they return block bodies
+        self._body_peers = WaitingPeers(commands.BlockBodies)
+
+        # Track incomplete block body download tasks
+        # - arbitrarily allow up to 4 requests-worth of headers queued up
+        # - try to get bodies from lower block numbers first
+        self._block_body_tasks = TaskQueue(MAX_BODIES_FETCH * 4, attrgetter('block_number'))
+
+    # TODO move this to BaseService if it gets broader usage and/or stays stable for long enough
+    def delayed_run_func(self, func: Callable[[], None], delay: float) -> None:
```
pipermerriam: This seems like it should be using [`loop.call_later()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.call_later) or `call_at`

trinity/sync/full/chain.py
```diff
@@ -85,314 +143,194 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
 
-            block_bodies_by_key = merge(block_bodies_by_key, {
-                (transaction_root, uncles_hash): block_body
-                for block_body, (transaction_root, trie_dict_data), uncles_hash
-                in concat(all_block_body_bundles)
-            })
-            headers = tuple(concat(all_missing_headers))
+            self._mark_body_download_complete(batch_id, completed_headers + trivial_headers)
+        except BaseP2PError:
+            self._block_body_tasks.complete(batch_id, trivial_headers)
```
pipermerriam: Could this go in a `finally` statement so that it doesn't have to be present in all `except` blocks?

trinity/sync/full/chain.py
```diff
@@ -85,314 +143,194 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+            self.logger.debug("Problem downloading body from peer, dropping...", exc_info=True)
```
pipermerriam: This blanket catch of any `p2p` exception stands out to me as potentially problematic. What do you think about adding an `INFO` level statement so that we can zero in on being more explicit about what we catch?

trinity/sync/full/chain.py
```diff
@@ -85,314 +143,194 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+                self._body_peers.put_nowait(peer)
+            else:
+                # peer returned no results, wait a while before trying again
+                delay = self.EMPTY_PEER_RESPONSE_PENALTY
+                self.logger.debug("Pausing %s for %.1fs, for sending 0 block bodies", peer, delay)
+                self.delayed_run_func(lambda: self._body_peers.put_nowait(peer), delay)
 
-        self.logger.debug("Got block bodies batch for %d headers", len(all_headers))
-        return block_bodies_by_key
+    def _mark_body_download_complete(
```
pipermerriam: Is this passthrough function just for readability?

trinity/sync/full/chain.py
```diff
@@ -85,314 +143,194 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
-                for block_body, (transaction_root, trie_dict_data), uncles_hash
-                in concat(all_block_body_bundles)
-            })
-            headers = tuple(concat(all_missing_headers))
+            self._mark_body_download_complete(batch_id, completed_headers + trivial_headers)
+        except BaseP2PError:
+            self._block_body_tasks.complete(batch_id, trivial_headers)
+            self.logger.debug("Problem downloading body from peer, dropping...", exc_info=True)
+        except Exception:
+            self._block_body_tasks.complete(batch_id, trivial_headers)
```
pipermerriam: This could/should use `self._mark_body_download_complete`?

trinity/sync/full/chain.py
```diff
@@ -85,314 +143,194 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
-            self.logger.trace("Pending receipts call to %r operation cancelled", peer)
-            return tuple(), batch
+            self.logger.trace("Pending block bodies call to %r operation cancelled", peer)
+            return tuple()
         except PeerConnectionLost:
-            self.logger.debug("Peer went away, cancelling the receipts request and moving on...")
-            return tuple(), batch
+            self.logger.debug("Peer went away, cancelling the block body request and moving on...")
+            return tuple()
         except Exception:
```
pipermerriam: I guess we were already doing this but whatcha think about removing the bare `Exception` catching?  At this stage in the software I'd rather the node crash hard so that we're forced to address anything that bubbles out of there.

trinity/sync/full/chain.py
```diff
@@ -442,29 +380,444 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+            dependency_extractor=attrgetter('parent_hash'),
+        )
+
+    async def _run(self) -> None:
+        head = await self.wait(self.db.coro_get_canonical_head())
+        self._block_persist_tracker.set_finished_dependency(head)
+        self.run_task(self._launch_prerequisite_tasks())
+        self.run_task(self._assign_receipt_download_to_peers())
+        self.run_task(self._assign_body_download_to_peers())
+        self.run_task(self._persist_ready_blocks())
```
pipermerriam: So in theory, most of these background tasks are required for the service to be operational/functional.  Using `run_task` allows for them to exit at which point the client will keep running but sync will stop working.  Thoughts on how we can address this?  Something like `BaseService.run_daemon_task()`?

trinity/sync/full/chain.py
```diff
@@ -442,29 +380,444 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+            if header.uncles_hash != EMPTY_UNCLE_HASH:
+                body = self._pending_bodies.pop(header)
+                uncles = body.uncles
+            else:
+                uncles = tuple()
+
+            vm_class = self.chain.get_vm_class(header)
+            block_class = vm_class.get_block_class()
+            # We don't need to use our block transactions here because persist_block() doesn't do
+            # anything with them as it expects them to have been persisted already.
```
pipermerriam: I don't think this comment is accurate:

https://github.com/ethereum/py-evm/blob/de4809fe77f9e7502d51478a30f443f007fd0923/eth/db/chain.py#L233-L243

By not including the transactions, we are failing to populate the canonical transaction hash to block lookups right?

trinity/sync/full/chain.py
carver: > This suffers from the same problem you have with the previous peer.is_operational check. I see the following options.

Hm, I think it's working as intended. Maybe I'll add a comment to clarify, like:

```py
        while not peer.is_operational:
            # if so, look for the next best peer
            wrapped_peer = await self._waiting_peers.get()
            peer = wrapped_peer.original
            # assign next peer to `peer` so that we keep looping until we find an operational one
```

pipermerriam: Blah, somehow I read the original loop as an `if` statement.  Ignore me.

trinity/sync/full/chain.py
carver: Maybe, let's see if I can work it out (right now, it's being called separately in `_mark_body_download_complete()`)

trinity/sync/full/chain.py
carver: It's an unfortunate side-effect of the two classes being built with inheritance. They have to register that the download is complete in a different `OrderedTaskPreparation` object, with a different prerequisite.

trinity/sync/full/chain.py
carver: Hm, I haven't looked deeply (it was already there), but at first glance I think you're right.

p2p/service.py
```diff
@@ -142,6 +142,25 @@ def run_task(self, awaitable: Awaitable[Any]) -> None:
                 self.logger.trace("Task %s finished with no errors", awaitable)
         self._tasks.add(asyncio.ensure_future(_run_task_wrapper()))
 
+    def run_daemon_task(self, awaitable: Awaitable[Any]) -> None:
```
pipermerriam: I was looking at this and the `run_task` api and I think we should update them as follows.

```python
def run_task(self, fn, *args):
    @functools.wraps(fn)
    async def _run_task_wrapper():
        ...
    ...
```

By doing it this way we'll get intelligible stacktraces when tasks are not properly cleaned up which should make finding the offending code a lot easier.

trinity/sync/full/chain.py
```diff
@@ -238,7 +235,10 @@ def delayed_run_func(self, func: Callable[[], None], delay: float) -> None:
                 # peer returned no results, wait a while before trying again
                 delay = self.EMPTY_PEER_RESPONSE_PENALTY
                 self.logger.debug("Pausing %s for %.1fs, for sending 0 block bodies", peer, delay)
-                self.delayed_run_func(lambda: self._body_peers.put_nowait(peer), delay)
+                loop = self.get_event_loop()
+                loop.call_later(delay, partial(self._body_peers.put_nowait, peer))
```
pipermerriam: I think that we need to wrap `call_later` in a `BaseService.call_later` which retains the return value which I believe is the only way we can issue a cancellation when triggered.

PR #1228, opened 2018-08-30
Merge KademliaProtocol into DiscoveryProtocol

gsalgado:
KademliaProtocol has no reason to exist as both classes just keep references
to each other and go back and forth, e.g.:

  disc.recv_ping() -> kademlia.recv_ping() - > disc.send_pong()

Now everything related to handling of discovery messages is in
DiscoveryProtocol, and kademlia.py contains only the RoutingTable and
code related to calculating the distance between peers.


gsalgado:
That's right, no behaviour changes

PR #1231, opened 2018-08-31
WIP: sync chain and state data concurrently

pipermerriam:
### What was wrong?

Work to fix multiple issues.

- speed up sync by syncing chain and state data concurrently
- periodically update `root_hash` that state sync is syncing against
- allow for independent and concurrent calls to the `peer.requests` apis.

#### Cute Animal Picture


pipermerriam:
Some learnings were uncovered in this effort. closing as there are some things that need to be done before this can be done.

PR #1232, opened 2018-08-31
Enforce timeout correctly and add lock for concurrency handling

pipermerriam:
### What was wrong?

- The `timeout` parameter for the Exchange API was not being enforced correctly
- Two independent processes which issue requests over the exchange API consistently encounter the `AlreadyWaiting` exception.

### How was it fixed?

For the timeout, we keep track of when the request started within the response candidate stream and enforce the *time remaining* for each payload candidate.

For the concurrency issue, the `Manager` class now has an `asyncio.Lock` on it which is acquired and released as part of the request/response cycle.  Concurrent requests will run up against this lock.  In order to differentiate between a peer `Timeout` and a timeout in acquiring the request lock, when lock acquisition times out the `AlreadyWaiting` exception is raised instead.

#### Cute Animal Picture

monkey-riding-a-dog

### DIFF
--- a/trinity/protocol/common/managers.py
+++ b/trinity/protocol/common/managers.py
@@ -65,6 +65,7 @@ class ResponseCandidateStream(
         super().__init__(token)
         self._peer = peer
         self.response_msg_type = response_msg_type
+        self._lock = asyncio.Lock()
 
     async def payload_candidates(
             self,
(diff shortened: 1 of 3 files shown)


cburgdorf:
>Two independent processes which issue requests

Are we talking about actual OS processes? Isn't this all running in one process (networking) for now?

cburgdorf:
I had the same random CI fail (`py37-core`) on a unrelated PR earlier today.

pipermerriam:
> Are we talking about actual OS processes? Isn't this all running in one process (networking) for now?

No, two things in the same event loop.  Current state of affairs is that if one process is *awaiting* a response and another tries to issue a request, then the requesting code will raise an `AlreadyWaiting`.  This code changes that so that it properly waits for the first to finish and then the next will proceed.

cburgdorf:
Thanks for clarifying. The term process is a bit overloaded ;)

trinity/protocol/common/managers.py
```diff
@@ -81,13 +82,31 @@ def __init__(
...
+            self._request(request)
+            while self._is_pending():
+                timeout_remaining = max(0, timeout - (time.perf_counter() - start_at))
+                try:
+                    yield await self._get_payload(timeout_remaining)
+                except TimeoutError:
+                    tracker.record_timeout(timeout)
+                    raise
+        finally:
+            self._lock.release()
```
cburgdorf: Wait, does this `finally` account for both `try` blocks? If not, don't we need to free the lock in another `finally` for the first `try` as well?

carver: You mean the inner `try`? Yeah, the exception will bubble up and this outer `finally` will run.

cburgdorf: Not the inner, the one above. This is where we acquire the lock.

https://github.com/ethereum/py-evm/pull/1232/files/d3a6d95b5ca6663e35e56a36f7b7bdf70a91cb35#diff-31d44fca310c7755ffc8e946f659e7c3R89

And then we raise `AlreadyWaiting` and jump out of this method. So are we releasing the lock with that `finally` at the bottom?

cburgdorf: Ah, scratch that. We do not acquire the lock there, hence the exception in the first place.

PR #1238, opened 2018-09-03
Move ShutdownRequest to trinity.extensibility.events

cburgdorf:
### What was wrong?

This is a small architectural cleanup that I just became aware of. The `ShutdownRequest` was introduced at `trinity.events`. However, I believe it is more appropriate for that to live under `trinity.extensibility.events`.

Reasoning:

Imagine we would rip the whole `trinity.extensibility` system out of Trinity to be used in other projects (No plans for that, just hypothetically!). A plugin would still be able to call `self.context.shutdown_host()` as it pretty much has the same meaning no matter if we are talking about the plugin system within Trinity or within a different app that uses that plugin system.

Hence, it makes more sense to have the `ShutdownRequest` event be part of `trinity.extensibility.events` as opposed to `trinity.events`.

In general, the less `trinity.extensibility` imports from outside of that namespace, the more generic it remains. While I'm not advocating for making our live harder by trying to build a *generic* plugin mechanism aimed to become a standalone project, at the same time I see no reason to artificially increase coupling when it isn't needed.

### How was it fixed?

Moved class.

#### Cute Animal Picture

### DIFF
--- a/trinity/events.py
+++ b/trinity/events.py
@@ -1,7 +1,0 @@
-from lahja import (
-    BaseEvent
-)
-
-
-class ShutdownRequest(BaseEvent):
-    pass
--- a/trinity/extensibility/events.py
+++ b/trinity/extensibility/events.py
@@ -7,6 +7,8 @@ from argparse import (
     Namespace,
 )
 
+import lahja
+
 from trinity.config import (
     ChainConfig,
 )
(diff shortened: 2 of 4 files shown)


cburgdorf:
Mmh..the other events that are related to the plugin manager do also live in `trinity.extensibility.events` e.g. `TrinityStartupEvent`. Today, these have different base events but long term they will probably all come through the event bus.

Anyway, if nobody is convinced...BOOOO....closing.

PR #1240, opened 2018-09-03
Implement node discovery by topic, using discv5

gsalgado:
New, highly experimental, DiscoveryByTopicProtocol class that looks
up nodes by topic and falls back to kademlia's random node ID lookups
when not enough topic nodes are found.

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -280,7 +280,7 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
         else:
             self.send_find_node_v4(node, target_node_id)
 
-    async def lookup(self, node_id: int) -> List[kademlia.Node]:
+    async def lookup(self, node_id: int) -> Tuple[kademlia.Node, ...]:
         """Lookup performs a network search for nodes close to the given target.
 
         It approaches the target by querying nodes that are closer to it on each iteration.  The
@@ -333,9 +333,9 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
             nodes_to_ask = _exclude_if_asked(closest)
 
         self.logger.debug("lookup finished for %s: %s", node_id, closest)
-        return closest
+        return tuple(closest)
 
-    async def lookup_random(self) -> List[kademlia.Node]:
+    async def lookup_random(self) -> Tuple[kademlia.Node, ...]:
         return await self.lookup(random.randint(0, kademlia.k_max_node_id))
 
     def get_random_bootnode(self) -> Iterator[kademlia.Node]:
@@ -644,10 +644,12 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
 
     def recv_find_node_v5(
             self, node: kademlia.Node, payload: Tuple[Any, ...], msg_hash: Hash32) -> None:
+        # FIND_NODE in v5 is identical to v4, so just call the v4 handler.
         self.recv_find_node_v4(node, payload, msg_hash)
 
     def recv_neighbours_v5(
             self, node: kademlia.Node, payload: Tuple[Any, ...], msg_hash: Hash32) -> None:
+        # NEIGHBOURS in v5 is identical to v4, so just call the v4 handler.
         self.recv_neighbours_v4(node, payload, msg_hash)
 
     def recv_find_nodehash(self, node: kademlia.Node, payload: Tuple[Any, ...], _: Hash32) -> None:
@@ -717,17 +719,13 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
 
     def send_topic_nodes(
             self, node: kademlia.Node, echo: Hash32, nodes: Tuple[kademlia.Node, ...]) -> None:
-        encoded_nodes = []
-        for n in nodes:
-            encoded_nodes.append(n.address.to_endpoint() + [n.pubkey.to_bytes()])
-
+        encoded_nodes = tuple(
+            n.address.to_endpoint() + [n.pubkey.to_bytes()]
+            for n in nodes)
         max_neighbours = self._get_max_neighbours_per_packet()
-        for i in range(0, len(encoded_nodes), max_neighbours):
-            message = _pack_v5(
-                CMD_TOPIC_NODES.id,
-                (echo, encoded_nodes[i:i + max_neighbours]),
-                self.privkey)
-            self.logger.trace('>>> topic_nodes to %s: %s', node, nodes[i:i + max_neighbours])
+        for batch in cytoolz.partition_all(max_neighbours, encoded_nodes):
+            message = _pack_v5(CMD_TOPIC_NODES.id, (echo, batch), self.privkey)
+            self.logger.trace('>>> topic_nodes to %s: %s', node, batch)
             self.send_v5(node, message)
 
     async def wait_topic_nodes(
@@ -857,21 +855,16 @@ class DiscoveryByTopicProtocol(DiscoveryProtocol):
         num_nodes_needed = max(0, count - len(topic_nodes))
         yield from super().get_nodes_to_connect(num_nodes_needed)
 
-    async def lookup_random(self) -> List[kademlia.Node]:
+    async def lookup_random(self) -> Tuple[kademlia.Node, ...]:
         query_nodes = list(self.routing.get_random_nodes(self._concurrent_topic_nodes_requests))
-        expected_echoes = []
-        for n in query_nodes:
-            echo = self.send_topic_query(n, self.topic)
-            expected_echoes.append((n, echo))
+        expected_echoes = tuple(
+            (n, self.send_topic_query(n, self.topic))
+            for n in query_nodes)
         replies = await asyncio.gather(
             *[self.wait_topic_nodes(n, echo) for n, echo in expected_echoes])
-        seen_nodes: Set[kademlia.Node] = set()
-        for reply in replies:
-            for node in reply:
-                if node in seen_nodes:
-                    continue
-                seen_nodes.add(node)
-                self.topic_table.add_node(node, self.topic)
+        seen_nodes = set(cytoolz.concat(replies))
+        for node in seen_nodes:
+            self.topic_table.add_node(node, self.topic)
 
         if len(seen_nodes) < kademlia.k_bucket_size:
             # Not enough nodes were found for our topic, so perform a regular kademlia lookup for
@@ -879,7 +872,7 @@ class DiscoveryByTopicProtocol(DiscoveryProtocol):
             extra_nodes = await super().lookup_random()
             seen_nodes.update(extra_nodes)
 
-        return list(seen_nodes)
+        return tuple(seen_nodes)
 
 
 class DiscoveryService(BaseService):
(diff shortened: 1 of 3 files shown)


p2p/discovery.py
```diff
@@ -617,40 +630,56 @@ def recv_ping_v5(
...
-        # TODO: Create/store ticket(s)
+        # TODO: Create/store ticket in ticket store.
+
+    def recv_find_node_v5(
+            self, node: kademlia.Node, payload: Tuple[Any, ...], msg_hash: Hash32) -> None:
+        self.recv_find_node_v4(node, payload, msg_hash)
+
+    def recv_neighbours_v5(
+            self, node: kademlia.Node, payload: Tuple[Any, ...], msg_hash: Hash32) -> None:
+        self.recv_neighbours_v4(node, payload, msg_hash)
```
pipermerriam: double checking that this isn't a typo (v5 in function name, v4 in this call)

p2p/discovery.py
```diff
@@ -617,40 +630,56 @@ def recv_ping_v5(
...
         # to, token, expiration, topic_hash, ticket_serial, wait_periods
         _, token, _, _, _, _ = payload
         self.logger.trace('<<< pong (v5) from %s (token == %s)', node, encode_hex(token))
         self.process_pong(node, token)
-        # TODO: Create/store ticket(s)
+        # TODO: Create/store ticket in ticket store.
+
+    def recv_find_node_v5(
+            self, node: kademlia.Node, payload: Tuple[Any, ...], msg_hash: Hash32) -> None:
+        self.recv_find_node_v4(node, payload, msg_hash)
```
pipermerriam: double checking that this isn't a typo (v5 in function name, v4 in this call)

p2p/discovery.py
```diff
@@ -680,13 +709,63 @@ def send_find_node_v5(self, node: kademlia.Node, target_node_id: int) -> None:
...
-        # TODO: Get the msg hash and store it in the ticket store to match against received
-        # topic_node msgs.
-        self.send_v5(node, message)
+        return self.send_v5(node, message)
+
+    def send_topic_nodes(
+            self, node: kademlia.Node, echo: Hash32, nodes: Tuple[kademlia.Node, ...]) -> None:
+        encoded_nodes = []
+        for n in nodes:
+            encoded_nodes.append(n.address.to_endpoint() + [n.pubkey.to_bytes()])
```
pipermerriam: Any reason not to use a comprehension to do this.  I think it's more efficient and maybe reads a bit easier.

```python
encoded_nodes = tuple(
    n.address.to_endpoint() + [n.pubkey.to_bytes()]
    for n in nodes
)

File: p2p/discovery.py [hunk 4]
```diff
@@ -680,13 +709,63 @@ def send_find_node_v5(self, node: kademlia.Node, target_node_id: int) -> None:
...
+
+        max_neighbours = self._get_max_neighbours_per_packet()
+        for i in range(0, len(encoded_nodes), max_neighbours):
+            message = _pack_v5(
+                CMD_TOPIC_NODES.id,
+                (echo, encoded_nodes[i:i + max_neighbours]),
```
pipermerriam: This looks likes the right case for `cytoolz.partition_all`

```python
for batch in cytoolz.partition_all(max_neighbors, encoded_nodes):
    ....
```

p2p/discovery.py
```diff
@@ -756,6 +835,53 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+
+        num_nodes_needed = max(0, count - len(topic_nodes))
+        yield from super().get_nodes_to_connect(num_nodes_needed)
+
+    async def lookup_random(self) -> List[kademlia.Node]:
+        query_nodes = list(self.routing.get_random_nodes(self._concurrent_topic_nodes_requests))
+        expected_echoes = []
+        for n in query_nodes:
+            echo = self.send_topic_query(n, self.topic)
+            expected_echoes.append((n, echo))
```
pipermerriam: I think this might be both more efficient and maybe easier to read as a comprehension.

expected_echos = tuple(
    (n, self.send_topic_query(n, self.topic)),
    for n in query_nodes
)
```

The inlining of the `self.send_topic_query` call is the only part that is a little questionable in terms of readability for me but I think it should e fine.

File: p2p/discovery.py [hunk 6]
```diff
@@ -756,6 +835,53 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+        replies = await asyncio.gather(
+            *[self.wait_topic_nodes(n, echo) for n, echo in expected_echoes])
+        seen_nodes: Set[kademlia.Node] = set()
+        for reply in replies:
+            for node in reply:
+                if node in seen_nodes:
+                    continue
+                seen_nodes.add(node)
```
pipermerriam: Maybe cleanup and remove some of the nesting?  I think the following is equivalent.

```
seen_nodes = set(cytoolz.concat(replies))
for node in seen_nodes:
    self.topic_table.add_node(node, self.topic)
```

p2p/discovery.py
```diff
@@ -756,6 +835,53 @@ def get_nodes_to_connect(self, count: int) -> Iterator[kademlia.Node]:
...
+                self.topic_table.add_node(node, self.topic)
+
+        if len(seen_nodes) < kademlia.k_bucket_size:
+            # Not enough nodes were found for our topic, so perform a regular kademlia lookup for
+            # a random node ID.
+            extra_nodes = await super().lookup_random()
+            seen_nodes.update(extra_nodes)
+
+        return list(seen_nodes)
```
pipermerriam: Is there a reason to use a list over a tuple here?

p2p/discovery.py
```diff
@@ -820,6 +946,63 @@ def __init__(self, proto: DiscoveryProtocol, peer_pool: PeerPool,
...
+
+    def __init__(self, logger: TraceLogger) -> None:
+        self.logger = logger
+        # A per-topic FIFO list of nodes.
+        self.topics: Dict[bytes, List[TopicEntry]] = {}
+        # The IDs of the last issued/used tickets for any given node.
+        self.node_tickets: Dict[kademlia.Node, NodeTicketInfo] = {}
+
+    def add_node(self, node: kademlia.Node, topic: bytes) -> None:
+        entries = self.topics.setdefault(topic, [])
```
pipermerriam: Did you consider using a `collections.defaultdict`?  That would allow you to remove this initialization step.

p2p/discovery.py
```diff
@@ -820,6 +946,63 @@ def __init__(self, proto: DiscoveryProtocol, peer_pool: PeerPool,
...
+        if len(entries) >= MAX_ENTRIES_PER_TOPIC:
+            entries.pop(0)
```
pipermerriam: Thoughts on changing this to be `entries = entries[-1 * MAX_ENTRIES_PER_TOPIC:]` so that this will handle the case where it exceeds the maximum by more than 1?

p2p/discovery.py
```diff
@@ -820,6 +946,63 @@ def __init__(self, proto: DiscoveryProtocol, peer_pool: PeerPool,
...
+            return tuple()
+        else:
+            now = time.time()
+            entries = [entry for entry in self.topics[topic] if entry.expiry > now]
+            self.topics[topic] = entries
+            return tuple(entry.node for entry in entries)
+
+    def use_ticket(self, node: kademlia.Node, ticket_serial: int, topic: bytes) -> None:
+        ticket_info = self.node_tickets.get(node)
+        if ticket_info is None:
```
pipermerriam: Thoughts on using a `try/except` for this and doing the lookup as `ticket_info = self.node_tickets[node]`.  It'll perform worse in the failure case but reduce some overhead in the success case.

p2p/discovery.py
```diff
@@ -820,6 +946,63 @@ def __init__(self, proto: DiscoveryProtocol, peer_pool: PeerPool,
         await self.proto.stop()
 
 
+class NodeTicketInfo:
+    last_issued: int = 0
+    last_used: int = 0
```
pipermerriam: These could potentially use some comments explaining the purpose of both of these fields.  Reading the code I haven't figured out exactly what they are yet.

p2p/discovery.py
```diff
@@ -993,21 +1176,21 @@ def _test() -> None:
     else:
         bootstrap_nodes = tuple(
             kademlia.Node.from_uri(enode) for enode in constants.ROPSTEN_BOOTNODES)
-    discovery = DiscoveryProtocol(privkey, addr, bootstrap_nodes, CancelToken("discovery"))
-    loop.run_until_complete(
-        loop.create_datagram_endpoint(lambda: discovery, local_addr=('0.0.0.0', listen_port)))
+
+    if args.v5:
+        # topic = b'LES2@41941023680923e0'  # LES2/ropsten
+        topic = b'LES2@d4e56740f876aef8'  # LES2/mainnet
```
pipermerriam: Maybe move the strings themselves to a `constants.py` file somewhere?

gsalgado: This is in the `_test()` function. In actual production code the topics will be generated from the supported protocols and the hash of the genesis block

p2p/discovery.py
gsalgado: I did that and got an error from mypy:

> No overload variant of "defaultdict" matches argument type "List[<nothing>]"

Any idea how to make that go away?

pipermerriam: A `cast` seems easiest.

```
self.topics = cast(Dict[bytes, List[TopicEntry]], defaultdict(list))
```

I haven't tried it but sometimes `mypy` knows how to handle the string versions:

```python
self.topics: "defaultdict[bytes, List[TopicEntry]]" = defaultdict(list)
```

p2p/discovery.py
gsalgado:
```
class NodeTicketInfo:
    # The serial number of the last ticket we issued for a given remote node. New tickets are
    # issued when a node sends a PING msg containing one or more topics.
    last_issued: int = 0
    # The serial number of the last ticket used by a given remote node. Tickets are marked as used
    # when a node sends a REGISTER_TICKET msg using a ticket previously issued.
    last_used: int = 0

```

p2p/discovery.py
gsalgado: I think this is more readable and I'm not sure it's worth worrying about that kind of overhead here. Can change if you feel strong about it, though

p2p/discovery.py
gsalgado: Nope, it was just what the existing API used. Changed to tuple here and in the type annotations

p2p/discovery.py
gsalgado: This is actually buggy as it will not pop one item when we're at max capacity, right? Went with this instead:
```
        if len(entries) >= MAX_ENTRIES_PER_TOPIC:
            # We add 1 to :to_drop: here because we want to have room for the new item we're
            # adding.
            to_drop = (len(entries) - MAX_ENTRIES_PER_TOPIC) + 1
            entries = cytoolz.drop(entries, to_drop)
```

gsalgado:
@pipermerriam you happy for this to go in now?

p2p/discovery.py
carver: What are the challenges for moving all the p2p `_test()` methods somewhere into the `tests` folder? Is it just a straight cut/paste job into a `def test_...` method, or is there something tougher?

gsalgado: This is not an automated test -- it's meant to be executed manually as it will connect to remote nodes, like the one in, say, `trinity/sync/__main__.py`

PR #1247, opened 2018-09-03
WIP: Use private genesis config for private network

### LINKED ISSUE #722: Make trinity conform to standard genesis.json standard from EIP-1085
### What is wrong?

https://github.com/ethereum/EIPs/issues/1085

### How can it be fixed

Implement the spec

cpurta:
Will resolve #722 when merged.

### What was wrong?

Currently trinity does not allow for the use of a private chain by specifying a custom genesis block. Trinity only allows the use of the mainnet and ropsten test-net.

### How was it fixed?

Allow trinity to use a custom genesis configuration to allow for private
networks to be specified. The changes specify that a genesis flag be
supplied with a valid filepath to the genesis config. It will also
require that a datadir flag is provided.

If using the custom genesis config if it does not contain the required
parameters trinity will not continue to try and create a new network.
Testing is still yet to be implemented.


trinity/chains/__init__.py
```diff
@@ -56,6 +65,50 @@
 )
 
 
+def validate_genesis(chain_config: ChainConfig) -> (BlockHeader, int):
```
pipermerriam: Can we change this to `validate_EIPXXX_genesis` (with the `XXX` replaced by the EIP number), as well as maybe linking to the EIP in the docstring for easy reference.

trinity/chains/__init__.py
```diff
@@ -56,6 +65,50 @@
...
+    """
+    Checks the genesis configuration file provided and ensures that it is valid.
+    """
+    if not os.path.exists(chain_config.genesis):
+        raise MissingPath(
```
pipermerriam: I think the built-in `FileNotFoundError` would be sufficient here.

trinity/chains/__init__.py
```diff
@@ -56,6 +65,50 @@
...
+            "The base chain genesis configuration file does not exist: `{0}`".format(
+                chain_config.genesis,
+            ),
+        )
+
+    with open(chain_config.genesis, 'r') as genesis_config:
+        raw_genesis = genesis_config.read()
+        genesis = json.load(raw_genesis)
```
pipermerriam: These two lines should be replaced with `genesis = json.load(genesis_config)`.  No need to `read()` the file object.

trinity/chains/__init__.py
```diff
@@ -56,6 +65,50 @@
...
+
+        if not 'chainId' in genesis.keys():
```
pipermerriam: Once we've parsed the JSON, we can exit the `with` block above, dedenting these lines a level.

trinity/chains/__init__.py
```diff
@@ -56,6 +65,50 @@
...
+        if not 'difficulty' in genesis.keys():
+            raise ValidationError("genesis config missing required 'difficulty'")
+        if not 'gasLimit' in genesis.keys():
+            raise ValidationError("genesis config missing required 'gasLimit'")
+        if not 'nonce' in genesis.keys():
+            raise ValidationError("genesis config missing required 'nonce'")
+        if not 'extraData' in genesis.keys():
+            raise ValidationError("genesis config missing required 'extraData'")
+
+        return BlockHeader(
```
pipermerriam: Our general pattern is that `validate_` functions don't return data, so this should likely be re-written to three independent things.

1. load the JSON
2. validate the JSON
3. extraction (block header, chain configuration, etc)

pipermerriam:
@cpurta let me know when you'd like me to take another look.

cpurta:
@pipermerriam feel free to take another look. Most recent commits have been just trying to get tests to pass.

cpurta:
Got this to a state to where it is using the genesis config correctly and is now running using that config. But it seems to be wanting to connect to the discovery peers to sync with them. Also will be addresses some of the failing CI tests.

cpurta:
@pipermerriam I believe that this could use another review before I address the issues in CI.

pipermerriam:
@cpurta :eyes:

cpurta:
@pipermerriam any comments on the changes? Going to try and tackle the CI issues today.

pipermerriam:
@cpurta I started looking but got pulled away last time.  Going to try and close this out this week.  Likely late today or tomorrow before I get to reviewing this.

eth/chains/base.py
```diff
@@ -125,6 +125,10 @@ class BaseChain(Configurable, ABC):
     def __init__(self) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
+    @property
+    def vm_configuration(self) -> Tuple[Tuple[int, Type['BaseVM']], ...]:
+        return self.vm_configuration
```
pipermerriam: This is going to cause infinite recursion.

pipermerriam: Also, I believe we already have a `get_vm_configuration` method that you can use.

pipermerriam:
@cpurta I'm going to take this PR over to get it across the line and merged.  thanks for getting it this far :joy_cat:

Issue #1234, opened 2018-09-01
Trinity crash


pipermerriam:
### What is wrong?

Trinity crashed, logs are unclear as to why it shutdown.

https://gist.githubusercontent.com/pipermerriam/f33802d8d8890af3c6c7a128e2745e75/raw/a272bab381ed7b2c8dc68573cc3ac53058ca4c8b/trinity-strange-crash.log

### How can it be fixed

Still need to try and diagnose...

carver:
I don't have the answer, but this seems like the critical bit:

```
   DEBUG  08-31 17:48:55        peer  Malformed message from peer ETHPeer <Node(0xcd7c@167.114.230.91)>: CMD:BlockHeaders Error: MalformedMessage('Malformed %s message: %r',)
   DEBUG  08-31 17:48:55        peer  Disconnecting from remote peer; reason: bad_protocol
   DEBUG  08-31 17:48:55     service  Cancelling ETHPeer <Node(0xcd7c@167.114.230.91)>
   DEBUG  08-31 17:48:55     service  <ResponseCandidateStream(ETHPeer <Node(0xcd7c@167.114.230.91)>, <class 'trinity.protocol.eth.commands.BlockHeaders'>)> finished: Cancellation requested by ETHPeer:PeerPool:Server:MainnetFullNode token
```

That first debug log happens right before re-raising the exception. So I guess the exception is somehow bubbling all the way up to somewhere it shouldn't, and causing a shutdown. If we find where that catch-all exception is, we should definitely log at the `exception` level to get the full stack trace.

carver:
BTW, I see the message is not formatted correctly, I'll fix that in the next PR I'm working on.

gsalgado:
This caught my attention
```
   DEBUG  08-31 17:48:55     service  Cancelling ETHPeer <Node(0xcd7c@167.114.230.91)>
   DEBUG  08-31 17:48:55     service  <ResponseCandidateStream(ETHPeer <Node(0xcd7c@167.114.230.91)>, <class 'trinity.protocol.eth.commands.BlockHeaders'>)> finished: Cancellation requested by ETHPeer:PeerPool:Server:MainnetFullNode token
    INFO  08-31 17:48:55   discovery  stopping discovery
   DEBUG  08-31 17:48:55     service  ETHPeer <Node(0x473b@167.99.45.137)> finished: Cancellation requested by Server:MainnetFullNode token
```

[...]

PR #1249, opened 2018-09-03
Improve fast-sync reliability & clean up messages

carver:
### What was wrong?

Some sync cases were crashing:
- when no peers available for sync download
- when no receipts are received in the entire batch

Some messages are not formatted correctly, or unclear

Some unnecessary cruft remained from exploratory implementations.

### How was it fixed?

Crash fixes:
- early escape from no receipts
- catch `NoEligiblePeers`, log, and pause

Improve & fix formatting on exception messages and logs

Remove cruft (like the `complete_token`, in favor of just canceling the service)

Add a new `BaseService.cancel_nowait()` that can be called from inside the service without hanging.

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/common/chain.py
+++ b/trinity/sync/common/chain.py
@@ -36,7 +36,6 @@ from trinity.protocol.eth.peer import ETHPeer
 from trinity.protocol.les.peer import LESPeer
 from trinity.utils.datastructures import TaskQueue
 
-
 HeaderRequestingPeer = Union[LESPeer, ETHPeer]
 
 
@@ -49,8 +48,6 @@ class BaseHeaderChainSyncer(BaseService, PeerSubscriber):
     """
     # We'll only sync if we are connected to at least min_peers_to_sync.
     min_peers_to_sync = 1
-    # Post-processing steps can exit out of sync (for example, fast sync) by triggering this token:
-    complete_token = None
     # TODO: Instead of a fixed timeout, we should use a variable one that gets adjusted based on
     # the round-trip times from our download requests.
     _reply_timeout = 60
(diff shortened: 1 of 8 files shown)


carver:
Fixes the unformatted message mentioned in #1234

trinity/sync/common/chain.py
```diff
@@ -64,12 +61,7 @@ def __init__(self,
                  db: AsyncHeaderDB,
                  peer_pool: PeerPool,
                  token: CancelToken = None) -> None:
-        self.complete_token = CancelToken('trinity.sync.common.BaseHeaderChainSyncer.SyncCompleted')
-        if token is None:
-            master_service_token = self.complete_token
-        else:
-            master_service_token = token.chain(self.complete_token)
-        super().__init__(master_service_token)
+        super().__init__(token)
```
pipermerriam: cc @gsalgado ^ I'm super distracted right now but pointing you at this location as a possible candidate for how #1234 is happening.

gsalgado: I'd came across that yesterday when trying to find out if we were wrongly chaining tokens somewhere, but this actually looked correct to me

trinity/sync/full/chain.py
```diff
@@ -94,8 +94,16 @@ class FastChainSyncer(BaseHeaderChainSyncer):
             # TODO implement the maximum task size at each step instead of this magic number
             max_headers = min((MAX_BODIES_FETCH, MAX_RECEIPTS_FETCH)) * 4
             batch_id, headers = await self.header_queue.get(max_headers)
-            await self._process_headers(headers)
-            self.header_queue.complete(batch_id, headers)
+            try:
+                await self._process_headers(headers)
+            except NoEligiblePeers:
+                self.logger.info(
+                    f"No available peers to sync with, retrying in {self.NO_PEER_RETRY_PAUSE}s"
```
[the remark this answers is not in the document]
carver: Yup, starting in Py3.6. So we can't use it in py-evm yet, only p2p and trinity.

carver:
Tests are green, Circle is just waiting for the SSH session to time out. Merging...

PR #1254, opened 2018-09-04
Add TaskIntegrator tracking class

carver:
### What was wrong?

In the process of #1226 it became necessary to track a bunch of asynchronous tasks that happen out of order, and then do follow-up post-processing only after they have been completed in sequence.

For example, we can download block bodies and receipt bodies in any order, but can't actually persist them to disk until we have each subtask completed (block download & receipt download), and in order.

But that PR is too big, so I'm pulling this into a separate one.

### How was it fixed?

From the new class (`OrderedTaskPreparation`) docstring:

This class is useful when a series of tasks with prerequisites must be run
sequentially. The prerequisites may be finished in any order, but the
tasks may only be run when all prerequisites are complete, and the
dependent task is also complete. Tasks may only depend on one other task.

For example, you might want to download block bodies and receipts at
random, but need to import them sequentially. Importing blocks is the ``task``,
downloading the parts is the ``prerequisite``, and a block's parent is its
``dependency``.

Below is a sketch of how to do that:

```py
    # The complete list of prerequisites to complete
    class BlockDownloads(Enum):
        receipts = auto()
        bodies = auto()

    block_import_tasks = OrderedTaskPreparation(
        BlockDownloads,

        # we use this method to extract an ID from the header:
        lambda header: header.hash,

        # we use this method to extract the ID of the dependency,
        # so that we can guarantee that the parent block gets imported first
        lambda header: header.parent_hash,
    ) 

    # We mark the genesis block as already imported, so header1 is ready
    # as soon as its prerequisites are complete.
    block_import_tasks.set_finished_dependency(header0)

    # We register the tasks before completing any prerequisites
    block_import_tasks.register_tasks((header1, header2, header3))

    # Start download of bodies & receipts...

    # They complete in random order

    # We notify this class which prerequisites are complete:
    block_import_tasks.finish_prereq(BlockDownloads.receipts, (header2, header3))
    block_import_tasks.finish_prereq(BlockDownloads.bodies, (header1, header2))

    # This await would hang, waiting on the receipt from header1:
    # await block_import_tasks.ready_tasks()

    block_import_tasks.finish_prereq(BlockDownloads.receipts, (header1, ))

    # Now we have all the necessary info to import blocks 1 and 2 
    headers_ready_to_import = await block_import_tasks.ready_tasks()

    # these will always return in sequential order:
    assert headers_ready_to_import == (header1, header2)
```

In a real implementation, you would have a loop waiting on
:meth:`ready_tasks` and import them, rather than interleaving them like
the above example.

Note that this class does *not* track when the main tasks are
complete. It is assumed that the caller will complete the tasks in the
order they are returned by ready_tasks().

The memory needs of this class would naively be unbounded. Any newly
registered task might depend on any other task in history. To prevent
unbounded memory usage, old tasks are pruned after a configurable depth.

Vocab:

- prerequisites: all these must be completed for a task to be ready
    (a necessary but not sufficient condition)
- ready: a task is ready after all its prereqs are completed, and the task it depends on is
    also ready

#### Cute Animal Picture


trinity/utils/datastructures.py
```diff
@@ -286,3 +297,267 @@ def complete(self, batch_id: int, completed: Tuple[TTask, ...]) -> None:
...
+    """
+    This class is useful when sequential follow-up processing is necessary after a non-sequential
+    task runs. For example, you might be able to download block bodies at random, but need to
+    import them sequentially. So you mark the block download task as complete and
+    await next_completion() before importing the block.
+
+    By combining with BaseTaskCompletion, you might have multiple steps required before sequential
+    completion, like downloading the block *and* receipt.
+
+    Tasks may only depend on one task, but there may be many tasks dependent on one task.
```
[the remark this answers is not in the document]
carver: Ok, this gave me an idea for how to improve the language (ie~ focus on the subtasks). I'll flesh it out and push it.

I checked [the celery primitives](http://docs.celeryproject.org/en/latest/userguide/canvas.html#the-primitives), but none of them are quite right. :/

---

(BTW, the answer is that downloading block bodies and receipts are subtasks, which can be many. The requirement that a task "may only depend on one task" is referring to the parent of the header)

carver:
Ok, I changed a bunch of the lingo used, and updated/added comments. I hope it's a bit easier to pick up. I also "cheated" a bit by showing an example in the class docstring rather than try to explain it more/better.

trinity/utils/datastructures.py
```diff
@@ -286,3 +297,315 @@ def complete(self, batch_id: int, completed: Tuple[TTask, ...]) -> None:
...
+        # all of the tasks that have been completed, and not pruned
+        self._tasks: Dict[TTaskID, BaseTaskPrerequisites[TTask, TSubtask]] = {}
+
+        # In self._dependencies, when the key becomes ready, the task ids in the
+        # value set *might* also become ready
+        # (they only become ready if their prerequisites are complete)
+        self._dependencies: Dict[TTaskID, Set[TTaskID]] = defaultdict(set)
+
+        # task ids are in this set if they either:
+        # - are missing a prerequisite OR
```
pipermerriam: missing is confusing for me here, would saying "all of their prerequisites are not completed" be accurate?

trinity/utils/datastructures.py
```diff
@@ -286,3 +297,315 @@ def complete(self, batch_id: int, completed: Tuple[TTask, ...]) -> None:
...
+        # - their dependent task is not ready
```
pipermerriam: Would saying, "The task that it depends on is not complete" be accurate?  I'm having trouble with the ready language.

trinity/utils/datastructures.py
```diff
@@ -286,3 +297,315 @@ def complete(self, batch_id: int, completed: Tuple[TTask, ...]) -> None:
...
+    _completed_subtasks: Set[TSubtask]
+    _task: TTask
+
+    @classmethod
+    def from_enum(cls, subtasks: Type[TSubtask]) -> 'Type[BaseTaskPrerequisites[Any, Any]]':
+
+        if len(subtasks) < 1:
+            raise ValidationError("There must be at least one subtask to track completions")
+
+        return type('CompletionFor' + subtasks.__name__, (cls, ), dict(_subtasks=subtasks))
```
pipermerriam: Alright, I think I found one source of confusion for me.  The term subtask here doesn't refer to actual [`asyncio.Task`] objects does it.  Rather, it's just something to represent the "jobs" that need to be done before the master task can be done.  Thoughts on using "job"?

trinity/utils/datastructures.py
```diff
@@ -286,3 +297,315 @@ def complete(self, batch_id: int, completed: Tuple[TTask, ...]) -> None:
...
+        else:
+            first_task = queue.get_nowait()
+
+        # In order to return from get() as soon as possible, never await again.
+        # Instead, take only the tasks that are already available.
+        available = queue.qsize()
+
+        available_tasks = tuple(queue.get_nowait() for _ in range(available))
+
+        completed = (first_task, ) + available_tasks
```
pipermerriam: I feel like this pattern is present in multiple places,  seems like a good candidate for a utility function.

trinity/utils/datastructures.py
carver: Yeah, the tricky bit is that we're not measuring task completion in this class, so it's not exactly right to say completion, either. All we know is that the task (say, block import) is `ready` to be run. And it's only ready after its prerequisites are done, and the previous block has also been marked ready.

trinity/utils/datastructures.py
carver: Oh yeah, this name is a bug (stayed the old name). Every name subtask should have been changed to prerequisite.

PR #1255, opened 2018-09-05
Catch operation cancelled and re-raise as PeerConnectionLost in exchange API

### LINKED ISSUE #1234: Trinity crash

pipermerriam:
fixes #1234 

### What was wrong?

The exchange API has a number of `self.wait` calls which will raise `OperationCancelled` if the peer's cancel token is triggered.  If this occurs during the dao fork check then the exception was bubbling all the way up to the `DiscoveryService` which orchestrates outbound peer connections.  This caused the `DiscoveryService` to *think* a cancellation had been triggered from it's token chain and it *gracefully* shut itself down.  Since the discovery service is run using the `run_daemon` API, this caused the `Server` to shutdown, which inturn, caused the master `Node` to shutdown.

### How was it fixed?

Addded a `try/catch` at the outermost common layer of the peer exchange API which will turn the `OperationCancelled` exception into a `p2p.PeerConnectionLost` exception.  This is already handled in the DAO fork code.  It probably needs to be handled in some other places as well.

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/exchanges.py
+++ b/trinity/protocol/common/exchanges.py
@@ -7,6 +7,9 @@ from typing import (
     Type,
 )
 
+from cancel_token import OperationCancelled
+
+from p2p.exceptions import PeerConnectionLost
 from p2p.protocol import (
     BaseRequest,
     Command,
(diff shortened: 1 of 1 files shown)


gsalgado:
I think we should either move the `ensure_same_side_on_dao_fork()` check inside the `handshake()` functions (annoying because it'd need to keep and return the messages received during the DAO fork check), or add a `except OperationCancelled` (with a big fat warning) around the `peer.requests.get_block_headers()` call in `PeerPool.ensure_same_side_on_dao_fork()`

pipermerriam:
Closing.  I agree with @gsalgado comment about trying to get the `ensure_same_side_on_dao_fork()` to live inside the `_run` and to try and reduce the places where we catch `OperationCancelled` to be inside of `_run` methods (ideally)

PR #1256, opened 2018-09-05
Move receipt validation to Chain and VM methods

### LINKED ISSUE #1246: ValidationError exception escaping during decoding of receipts.
### What is wrong?

Since `Receips` do some validation during initialization, the following error escapes while decoding a payload of receipts.

```
   ERROR  09-03 12:21:43     service  Unexpected error in ETHPeer <Node(0xdbda476f64495a02b3b07a0869bcd88e90cde016e988608e73f74ce96062cea7417d00ea4f6ca3b4fb2b996eaf20fe41fa7789ec737b222725288a9b652f627a@86.140.137.123:40303)>, exiting
  File "/home/piper/projects/py-evm/p2p/service.py", line 101, in run
    await self._run()
  File "/home/piper/projects/py-evm/p2p/peer.py", line 347, in _run
  File "/home/piper/projects/py-evm/p2p/peer.py", line 384, in read_msg
    decoded_msg = cast(Dict[str, Any], cmd.decode(msg))
  File "/home/piper/projects/py-evm/p2p/protocol.py", line 111, in decode
    return self.decode_payload(data[1:])
  File "/home/piper/projects/py-evm/p2p/protocol.py", line 93, in decode_payload
    data = rlp.decode(rlp_data, sedes=decoder, recursive_cache=True)
  File "/home/piper/python-environments/evm/lib/python3.6/site-packages/rlp/codec.py", line 235, in decode
    obj = sedes.deserialize(item, **kwargs)
  File "/home/piper/python-environments/evm/lib/python3.6/site-packages/eth_utils/functional.py", line 22, in inner
    return callback(fn(*args, **kwargs))
  File "/home/piper/python-environments/evm/lib/python3.6/site-packages/rlp/sedes/lists.py", line 138, in deserialize
    yield self.element_sedes.deserialize(element)
  File "/home/piper/python-environments/evm/lib/python3.6/site-packages/eth_utils/functional.py", line 22, in inner
    return callback(fn(*args, **kwargs))
  File "/home/piper/python-environments/evm/lib/python3.6/site-packages/rlp/sedes/lists.py", line 138, in deserialize
    yield self.element_sedes.deserialize(element)
  File "/home/piper/python-environments/evm/lib/python3.6/site-packages/rlp/sedes/serializable.py", line 262, in deserialize
    return cls(**args_as_kwargs, **extra_kwargs)
  File "/home/piper/projects/py-evm/eth/rlp/receipts.py", line 55, in __init__
    "present in the provided bloom filter.".format(log_idx)
eth_utils.exceptions.ValidationError: The address from the log entry at position 0 is not present in the provided bloom filter.
```

### How can it be fixed

Not entirely clear 

- Maybe handled in the `decode` of the `Receipts` command?

Also, this probably means we should disconnect from this peer since they sent us provably invalid data.

pipermerriam:
fixes #1246

### What was wrong?

See #1246 

### How was it fixed?

Both the `Chain` and `VM` classes now expose a `validate_receipt` method.  This validation is called during receipt creation by the VM class as well as during chain sync.

#### Cute Animal Picture

x26noborder

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -281,7 +281,7 @@ class BaseChain(Configurable, ABC):
     # Validation API
     #
     @abstractmethod
-    def validate_receipt(self, block_number: BlockNumber, receipt: Receipt) -> None:
+    def validate_receipt(self, receipt: Receipt, at_header: BlockHeader) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
@@ -672,8 +672,8 @@ class Chain(BaseChain):
     #
     # Validation API
     #
-    def validate_receipt(self, block_number: BlockNumber, receipt: Receipt) -> None:
-        VM = self.get_vm_class_for_block_number(block_number)
+    def validate_receipt(self, receipt: Receipt, at_header: BlockHeader) -> None:
+        VM = self.get_vm_class(at_header)
         VM.validate_receipt(receipt)
 
     def validate_block(self, block: BaseBlock) -> None:
@@ -889,6 +889,6 @@ class AsyncChain(Chain):
         raise NotImplementedError()
 
     async def coro_validate_receipt(self,
-                                    block_number: BlockNumber,
-                                    receipt: Receipt) -> None:
+                                    receipt: Receipt,
+                                    at_header: BlockHeader) -> None:
         raise NotImplementedError()
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -78,6 +78,9 @@ from eth.rlp.headers import (
     BlockHeader,
     HeaderParams,
 )
+from eth.rlp.receipts import (
+    Receipt,
+)
 from eth.rlp.transactions import (
     BaseTransaction,
     BaseUnsignedTransaction,
@@ -277,6 +280,10 @@ class BaseChain(Configurable, ABC):
     #
     # Validation API
     #
+    @abstractmethod
+    def validate_receipt(self, block_number: BlockNumber, receipt: Receipt) -> None:
+        raise NotImplementedError("Chain classes must implement this method")
+
     @abstractmethod
     def validate_block(self, block: BaseBlock) -> None:
         raise NotImplementedError("Chain classes must implement this method")
@@ -665,6 +672,10 @@ class Chain(BaseChain):
     #
     # Validation API
     #
+    def validate_receipt(self, block_number: BlockNumber, receipt: Receipt) -> None:
+        VM = self.get_vm_class_for_block_number(block_number)
+        VM.validate_receipt(receipt)
+
     def validate_block(self, block: BaseBlock) -> None:
         """
         Performs validation on a block that is either being mined or imported.
@@ -875,4 +886,9 @@ class AsyncChain(Chain):
             parent: BlockHeader,
             chain: Tuple[BlockHeader, ...],
             seal_check_random_sample_rate: int = 1) -> None:
+        raise NotImplementedError()
+
+    async def coro_validate_receipt(self,
+                                    block_number: BlockNumber,
+                                    receipt: Receipt) -> None:
         raise NotImplementedError()
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -6,8 +6,7 @@ from abc import (
 import contextlib
 import functools
 import logging
-from typing import (  # noqa: F401
-    List,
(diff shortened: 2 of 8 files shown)


carver:
Seems like a reasonable direction :+1:

carver:
Weird that the header timestamp in the chain builder was 17s different from `time.time()` in that last CI run.

eth/chains/base.py
```diff
@@ -277,6 +280,10 @@ def import_block(self,
     #
     # Validation API
     #
+    @abstractmethod
+    def validate_receipt(self, block_number: BlockNumber, receipt: Receipt) -> None:
```
carver: The arguments felt in the wrong order to me. I checked some other methods that are basically VM passthroughs, and they flip the order of arguments, so it would be nice to keep consistency if possible. Also, most other places are using the header instead of the block_number, which looks like it would be okay.

I'm looking at:
- `get_transaction_result()`
- `estimate_gas()`
- `build_block_with_transactions()`

So I think an ideal option here is: 

```py
    def validate_receipt(self, receipt: Receipt, at_header: BlockHeader) -> None:
```

eth/vm/base.py
```diff
@@ -189,7 +191,7 @@ def generate_block_from_parent_header_and_coinbase(cls, parent_header, coinbase)
 
     @classmethod
     @abstractmethod
-    def get_block_class(cls) -> Type['BaseBlock']:
+    def get_block_class(cls) -> Type[BaseBlock]:
```
carver: :+1:

trinity/sync/full/chain.py
```diff
@@ -354,7 +357,38 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+            receipt_root: receipt
+            for (receipt, (receipt_root, _))
+            in receipt_bundles
+            if receipt_root != BLANK_ROOT_HASH
+        }
+        receipts_by_header = {
+            headers_by_root[receipt_root]: receipts
+            for receipt_root, receipts
+            in receipts_by_root.items()
+            if receipt_root != BLANK_ROOT_HASH
```
carver: This `if` looks redundant to the one in 370.

trinity/sync/full/chain.py
```diff
@@ -354,7 +357,38 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
+        }
+        for header, receipts in receipts_by_header.items():
+            for receipt in receipts:
```
carver: Maybe just skip that last dict comprehension and:

```py
        for receipt_root, header in header_by_root.items():
            for receipt in receipts_by_root[receipt_root]:
```

trinity/sync/full/chain.py
```diff
@@ -354,7 +357,38 @@ class FastChainSyncer(BaseHeaderChainSyncer):
...
-        receipts, trie_roots_and_data_dicts = zip(*receipt_bundles)
+        # First validate the receipts.
+        headers_by_root = {
+            header.receipt_root: header
+            for header in batch
+            if not _is_receipts_empty(header)
+        }
+        receipts_by_root = {
+            receipt_root: receipt
+            for (receipt, (receipt_root, _))
```
carver: I think these should be plural `receipts`:

```py
            receipt_root: receipts
            for (receipts, (receipt_root, _))
```

trinity/sync/full/chain.py
```diff
@@ -354,7 +357,38 @@ class FastChainSyncer(BaseHeaderChainSyncer):
         if not receipt_bundles:
             return tuple(), batch
 
-        receipts, trie_roots_and_data_dicts = zip(*receipt_bundles)
+        # First validate the receipts.
+        headers_by_root = {
```
carver: To distinguish between single header values here, and multiple receipts values in the next dict, maybe call this: `header_by_root`?

eth/chains/base.py
pipermerriam: I originally chose `block_number` because I wasn't sure I'd have the full header, but it looks like the header is available everywhere I need it.

pipermerriam:
> Weird that the header timestamp in the chain builder was 17s different from time.time() in that last CI run.

My gut is telling me that we are approaching a pitt filled with sharp spiky non-deterministic test failures lining the bottom.  The RPC fixtures OOM-ing, this one have weirdly long runtimes, (i think there are others).... sketchy.

PR #1258, opened 2018-09-06
Handle OperationCancelled when peer disconnects during DAO fork check

gsalgado:
Closes: #1234

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -50,7 +50,7 @@ from eth_keys import (
     keys,
 )
 
-from cancel_token import CancelToken, OperationCancelled
+from cancel_token import CancelToken
 
 from eth.chains.mainnet import MAINNET_NETWORK_ID
 from eth.chains.ropsten import ROPSTEN_NETWORK_ID
@@ -813,9 +813,6 @@ class PeerPool(BaseService, AsyncIterable[BasePeer]):
                     self.cancel_token))
 
             return peer
-        except OperationCancelled:
-            # Pass it on to instruct our main loop to stop.
-            raise
         except BadAckMessage:
             # This is kept separate from the `expected_exceptions` to be sure that we aren't
             # silencing an error in our authentication code.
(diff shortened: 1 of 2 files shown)


gsalgado:
Alternative solutions I considered:

In order to perform the dao check in Peer._run() we'd need to run it in the background
and use a notification mechanism other than an Event so that we can tell the PeerPool
whether the check was successful or not. That is very error prone, though, because a
callsite may very easily forget to wait on that condition and start to use the peer straight
away. And we have no way of putting checks on peer methods to avoid them being called before
the DAO fork check has completed because we need to use those methods in order to perform
the DAO fork check.

Also thought about having the DAO fork check in the handshake() func, but then we'd have to
duplicate some code from _run() to read messages, and it'd also have to return the messages
received during the DAO check so that the PeerPool can re-add them to its subscribers. In
fact it'd require a lot of changes to make that work because the peer is not running at that
time and the new ExchangeManager (which is used during the DAO check) stuff requires the
peer to be running.

p2p/peer.py
```diff
@@ -956,10 +963,7 @@ def get_peers(self, min_td: int) -> List[BasePeer]:
                 for line in peer.get_extra_stats():
                     self.logger.debug("    %s", line)
             self.logger.debug("== End peer details == ")
-            try:
-                await self.sleep(self._report_interval)
-            except OperationCancelled:
-                break
```
gsalgado: This was no longer necessary as this coroutine now runs via `BaseService.run_task()`

pipermerriam:
Alright, seeing this I want to take a stab at an alternate idea.

pipermerriam:
Alright, it's really messy but I think it actually attacks the root of the problem as well as doing some cleanup migrating things that still live in the `p2p` module to be under `trinity`.

see https://github.com/ethereum/py-evm/pull/1260 as my proposed alternative to this PR

carver:
~~#1260~~
#1259 I think

pipermerriam:
#1265 is now the correct link.

PR #1260, opened 2018-09-06
Remove sharding code

pipermerriam:
### What was wrong?

The spec for sharding is no longer sufficiently defined to justify the sharding related code in this codebase.

### How was it fixed?

Removed it all

#### Cute Animal Picture

proxy duckduckgo com


hwwhww:
The SMC-related stuff is no longer used, but the blob serialization is still on the track. Agree with removing all from `master` branch now. See if @jannikluhn would like to add another branch (`sharding-p2p`?) for these code, or leave it for now, we can look back when we need it.

jannikluhn:
I'm fine with removing it all. I don't think a branch is necessary, we can easily find and restore the code from history if necessary.

Issue #1243, opened 2018-09-03
Cleanup trinity log level setting to include defaults and make them overridable


pipermerriam:
From: https://github.com/ethereum/py-evm/issues/806#issuecomment-417987842

### What is wrong?

We have some defaults that are used to quiet down some noisy loggers:

https://github.com/ethereum/py-evm/blob/v0.2.0-alpha.31/trinity/utils/logging.py#L111-L114

Were the use to supply overrides to these via the command line they would not be respected

### How can it be fixed

This lines should be relocated such that they are merged with the user supplied logging levels before being passed into `setup_log_levels`.

https://github.com/ethereum/py-evm/blob/v0.2.0-alpha.31/trinity/main.py#L111-L112

glaksmono:
@pipermerriam just want to align. Do we want [1] to keep those noisy logs at `INFO` level (i.e as current, less noisy) then overrides it when log level is defined from the command line?

OR

[2] We want to take out the `INFO` level config (i.e will become more noisy), then override when log level defined in command line?

PR #1261, opened 2018-09-07
Make the noisy logger to be overridable, but default it to INFO

glaksmono:
### What was wrong?
We have some defaults that are used to quiet down some noisy loggers:

https://github.com/ethereum/py-evm/blob/v0.2.0-alpha.31/trinity/utils/logging.py#L111-L114

Were the use to supply overrides to these via the command line they would not be respected

Issue Reference: https://github.com/ethereum/py-evm/issues/1243

### How was it fixed?
Default the noisy logger to `INFO`, but make it overridable when the `log_level` argument is supplied

#### Cute Animal Picture

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -143,7 +143,10 @@ def main() -> None:
     )
 
     if args.log_levels:
-        setup_log_levels(args.log_levels)
+        log_levels = args.log_levels.copy()  # use a copy so we don't mutate the input values.
+        log_levels.setdefault('p2p.kademlia', logging.INFO)
+        log_levels.setdefault('p2p.discovery', logging.INFO)
+        setup_log_levels(log_levels)
 
     try:
         chain_config = ChainConfig.from_parser_args(args)
--- a/trinity/utils/logging.py
+++ b/trinity/utils/logging.py
@@ -125,10 +125,6 @@ def setup_queue_logging(log_queue: 'Queue[str]', level: int) -> None:
     logger.addHandler(queue_handler)
     logger.setLevel(level)
 
-    # These loggers generates too much DEBUG noise, drowning out the important things, so force
-    # the INFO level for it until https://github.com/ethereum/py-evm/issues/806 is fixed.
-    logging.getLogger('p2p.kademlia').setLevel(logging.INFO)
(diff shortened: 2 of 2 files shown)


trinity/main.py
```diff
@@ -196,6 +199,17 @@ def main() -> None:
...
 
+def config_noisy_logger(args: Namespace) -> None:
+    # These loggers generates too much DEBUG noise, drowning out the important things
+    # We will use the log_level if it's set, otherwise default to INFO
+    if args.log_levels:
+        logging.getLogger('p2p.kademlia').setLevel(args.log_level)
+        logging.getLogger('p2p.discovery').setLevel(args.log_level)
+    else:
+        logging.getLogger('p2p.kademlia').setLevel(logging.INFO)
+        logging.getLogger('p2p.discovery').setLevel(logging.INFO)
```
cburgdorf: @glaksmono I think we can do this more generic by creating a constant `LOG_LEVEL_DEFAULTS` which is basically a `Dict[str, int]` to provide the default log levels and then merge those with the `log_level` that the user may provide via the command line (where the command line arguments take precedence over the defaults).

trinity/main.py
```diff
@@ -129,6 +129,9 @@ def main() -> None:
     logger, formatter, handler_stream = setup_trinity_stderr_logging(
         args.stderr_log_level
     )
+
+    config_noisy_logger(args)
```
pipermerriam: I think we can simplify this, removing any need for the `config_noisy_loggers`

```python
log_levels = args.log_levels.copy()  # use a copy so we don't mutate the input values.
log_levels.setdefault('p2p.kademlia', logging.INFO)
log_levels.setdefault('p2p.discovery', logging.INFO)
setup_log_levels(log_levels)
```

carver:
Another option is to move some of the noisiest logs in kademlia and discovery to `trace` level, in case some of the less noisy debug logs are useful during debugging.

gsalgado:
While implementing discv5 I'd changed most discovery/kademlia logs to use the TRACE level, so in #1310 I've included a change that removes the hardcoding of their level to INFO

PR #1262, opened 2018-09-07
POC: Introduce a bridge to the LightPeerChain to be consumed from other processes

cburgdorf:
**DISCLAIMER: This is PoC WIP code so no need to comment on naming, imports or whatever is just *messy***

### What is this?

This is a PoC sitting on top of #1212 

### Motivation

#1212 Moves the `JSON-RPC` service into an isolated plugin to run in its own process. However, that approach didn't work for `--light` mode so far because in that mode the chain is heavily depending on the `PeerPool`. So what #1212 does so far is that it provides a second plugin only for `--light` mode which isn't isolated but continues to run in the networking process instead.

That however isn't ideal because:

a.) we want the `JSON-RPC` plugin to run in a separate process to provide isolation from the rest of the networking stuff (e.g. reduce DDOS vector etc)

b.) Before #1212 (current master) the IPC Server created a separate thread to provide some isolation but I had problems reusing that approach together with the event bus. Leaving the `JSON-RPC` server inside the `networking` process but altering the process model to remove that extra thread scares me :sweat_smile: 

This PoC tries to pave the way for an isolated `JSON-RPC` server plugin that works for both `full` and `light` mode.

### How does it work

First, we create another plugin `LightPeerChainBridgePlugin` which does the following:

- Provide a `EventBusLightPeerChain` which is basically a version the `LightPeerChain` that can be used in any other process
- The `EventBusLightPeerChain` basically issues events via the event bus that will be answered by the plugin that is running inside the `networking` process and hence has access to the `LightPeerChain`. @pipermerriam That is basically no different from what #1212 does with the `peer_count` already and what I think we both agreed on would be the way forward for the general interaction between separate processes.
- I believe the `EventBusLightPeerChain` can become zero maintenance in the sense that it will magically provide all APIs that the `LightPeerChain` has which means whenever APIs change or are added, the `EventBusLightPeerChain` will automatically have them. (In contrast to our existing `Proxy` types that need to be altered whenever the remote type changes. Though, to be fair, I guess we could auto generated the methods for them as well). That isn't currently implemented in this PoC though.
- Notice that the `LightPeerChainBridgePlugin` isn't coupled to this particular use case and can be reused by other plugins that want to run in their own isolated process but need access to the `LightPeerChain`. Also notice that while this *could* equally be implemented in the core code, I believe there is no reason to do so and providing the functionality as a plugin makes it easy to drop it later e.g. when the current networking code becomes more event bus friendly in general (aka exposes thins on the event bus for other processes to consume)

Also in current master, since the IPC Server uses another thread, we are running this async call in another thread and block to make it synchronous.

https://github.com/ethereum/py-evm/blob/55613b7e3ef0a7ca42a981d014911eef2c07874c/trinity/chains/light.py#L143-L145

With the new approach the `JSON-RPC` server that now runs in its own process doesn't have this second thread anymore and hence we schedule that async method on the event loop as we normally would. I *believe* that this would be a win in general.

Would love to receive feedback on the approach in general and also on the idea to make the `EventBusLightPeerChain` zero maintenance through some meta programming magic if we choose to go down the path of this PR.

### DIFF
--- a/trinity/plugins/builtin/light_peer_chain_bridge/light_peer_chain_bridge.py
+++ b/trinity/plugins/builtin/light_peer_chain_bridge/light_peer_chain_bridge.py
@@ -1,0 +1,154 @@
+import asyncio
+import inspect
+from types import (
+    FrameType
+)
+from typing import (
+    Any,
+    Iterable,
+    List,
+)
+
+from eth_typing import (
+    Address,
+    Hash32,
+)
+
+from eth_utils import (
+    to_tuple,
+)
+
+from eth.rlp.accounts import (
+    Account,
+)
+from eth.rlp.headers import (
+    BlockHeader,
+)
+from eth.rlp.receipts import (
+    Receipt,
+)
+
+from lahja import (
+    BaseEvent,
+    Endpoint,
+)
+
+from trinity.rlp.block_body import BlockBody
+from trinity.sync.light.service import (
+    BaseLightPeerChain,
+)
+
+
+class LightPeerChainRequest(BaseEvent):
+
+    def __init__(self, method_name: str, payload: Any = None) -> None:
+        self.method_name = method_name
+        self.payload = payload
+
+
+class LightPeerChainResponse(BaseEvent):
+
+    def __init__(self, payload: Any = None) -> None:
+        self.payload = payload
+
+
+class LightPeerChainEventBusResponder:
+    """
+    The ``LightPeerChainEventBusResponder`` listens for certain events on the eventbus and
+    delegates them to the ``LightPeerChain`` to get answers. It then propagates responses
+    back to the caller.
+    """
+
+    def __init__(self, chain: BaseLightPeerChain, event_bus: Endpoint) -> None:
+        self.chain = chain
+        self.event_bus = event_bus
+        asyncio.ensure_future(self.answer_requests())
+
+    async def answer_requests(self) -> None:
+        async for event in self.event_bus.stream(LightPeerChainRequest):
+
+            method = getattr(self.chain, event.method_name)
+
+            if not callable(method):
+                self.event_bus.broadcast(
+                    LightPeerChainResponse(Exception(f"Not a method: {event.method_name}")),
+                    event.broadcast_config()
+                )
+                continue
+
+            try:
+                if event.payload is not None:
+                    response = await method(*event.payload)
+                else:
+                    response = await method()
+            except Exception as e:
+                # we send the exception over to re-raise it on the consumer side (other process)
+                self.event_bus.broadcast(
+                    LightPeerChainResponse(e),
+                    event.broadcast_config()
+                )
+            else:
+                self.event_bus.broadcast(
+                    LightPeerChainResponse(response),
+                    event.broadcast_config()
+                )
+
+
+class EventBusLightPeerChain(BaseLightPeerChain):
+    """
+    The ``EventBusLightPeerChain`` is an implementation of the ``BaseLightPeerChain`` that can
+    be used from within any process.
+    """
+
+    def __init__(self, event_bus: Endpoint) -> None:
+        self.event_bus = event_bus
+
+    async def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    async def get_block_body_by_hash(self, block_hash: Hash32) -> BlockBody:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    async def get_receipts(self, block_hash: Hash32) -> List[Receipt]:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    async def get_account(self, block_hash: Hash32, address: Address) -> Account:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    async def get_contract_code(self, block_hash: Hash32, address: Address) -> bytes:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    async def get_foo(self) -> BlockHeader:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    async def get_foobar(self, foo: int, bar: int) -> BlockHeader:
+        event = self._prepare_event(inspect.currentframe())
+        return self._return_or_raise(await self.event_bus.request(event))
+
+    def _return_or_raise(self, response: LightPeerChainResponse) -> Any:
+        if issubclass(type(response.payload), Exception):
+            raise Exception(
+                "Exception raised in LightPeerChain", response.payload
+            ) from response.payload
+
+        return response.payload
+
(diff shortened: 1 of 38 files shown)


carver:
Direction looks okay to me. What happens to mypy if we use inferred methods in `EventBusLightPeerChain`?

cburgdorf:
>What happens to mypy if we use inferred methods in `EventBusLightPeerChain`

Oh good catch. Mypy will hate that. As an alternative we could ditch the `EventBusLightPeerChain` and use the bare event bus directly.

Of course, in that case we are back to either:

**1. Maintaining something like an `RequestType` enum manually**

```python
block = await eventbus.request(LightPeerChainRequest(RequestType.GET_BLOCK_BY_HASH, (b'foo',)))
```

**2. Use stringly typed events**

```python
block = await eventbus.request(LightPeerChainRequest("get_block_by_hash", (b'foo',)))
```

**3. Or, if that kind of lamdba introspection is possible in Python, use a lamdba that acts on a `LightPeerChain` just to infer the right event in a type safe way (something not too uncommon in C#).**

```python
block = await eventbus.request(
    LightPeerChainRequest(lamdba chain: chain.get_block_by_hash(b'hash)
)
```
Just to be clear about this last pattern, the `lambda` is never executed it is only used to introspect it and derive the `method` as well as the params to create the right event in a type safe way. Not sure if that is possible in Python though.

carver:
> Not sure if that is possible in Python though.

I'm not sure either. I certainly haven't tried it before. It's weird enough that I'd say we should probably look for an alternative.

---

Given the options you laid out, the option of just having a class that explicitly lists pass-through methods (like how we do with the async ones) doesn't sound so bad.

trinity/plugins/builtin/light_peer_chain_bridge/light_peer_chain_bridge.py
```diff
@@ -0,0 +1,114 @@
...
+            try:
+                response = await method(*event.payload)
+            except Exception as e:
+                # we send the exception over to raise it on the other end
+                self.event_bus.broadcast(LightPeerChainResponse(e), event.broadcast_config())
+            else:
+                self.event_bus.broadcast(LightPeerChainResponse(response), event.broadcast_config())
+
+
+class EventBusLightPeerChain:
```
pipermerriam: I don't know if it actually makes sense, but looks like inheriting this from `BaseChain` the same way that `LightDispatchChain` does would provide some glue to ensure that the types and return types match up.

cburgdorf: Yes, I think so, too. Will comment a bit more on that further down.

cburgdorf:
Thank you both for your reviews and approval of the general direction. Here are some thoughts:

I think there's one major problem today that we probably all agree we don't want to repeat / spread. That is, today, if we introduce a chain method it is easy to miss to add the same method to the proxies and run into a runtime error. We had that several times in the past.

With that in mind I think the direction we are taking here should either:

1. Be maintenance free in the sense that every API call that works on `LightPeerChain` will automagically work on the event bus based sister chain. This would cover all API additions or changes.

or

2. Changes / additions to the `LightPeerChain` that aren't reflected on the event bus bases sister will cause the type system to yell.

From your comments I feel that you are both in favor of the second option.

>That is roughly an event pair for each request/response and explicit handling for each event type on the bridge

Yes, I agree that the `event_bus.request(...)` API should dictate that if the request is type `TRequest` then the return must be type `TResponse` with `TResponse` being automatically derived from `TRequest`. Notice though that afaik currently all 3rd party library type information was lost unless we have a stub file for it. That may have changed with [PEP561](https://www.python.org/dev/peps/pep-0561/) though. I still need to [read more about that](https://mypy.readthedocs.io/en/latest/installed_packages.html).

I think I'm going to experiment with both directions to continue the discussion with more concrete examples at hand.

cburgdorf:
@carver @pipermerriam I spend the last hours coming up with https://github.com/ethereum/py-evm/pull/1262/commits/2e9727fa867f3120ea523452db57b4fe56b7c542 but I guess I'll abandon it and try out something more standard now :grimacing:

cburgdorf:
Ok, here's another update.

First, let me cite my previous comment because it really is I think is most important here.

>I think there's one major problem today that we probably all agree we don't want to repeat / spread. That is, today, if we introduce a chain method it is easy to miss to add the same method to the proxies and run into a runtime error. We had that several times in the past.

With that in mind, here is what I have done:

1. I created a `BaseLightPeerChain` that both `LightPeerChain` and `EventBusLightPeerChain` derive from. I also created tests that check both can be instantiated. Also all consuming sides use `BaseLightPeerChain`. That ensures that new APIs need to be added to the `BaseLightPeerChain` and the test ensure that these APIs need to be implemented on both derived chains. So we catch these kinds of errors before runtime.

2.The `EventBusLightPeerChain` currently looks like this.

```python
class EventBusLightPeerChain(BaseLightPeerChain):

    def __init__(self, event_bus: Endpoint) -> None:
        self.event_bus = event_bus

    async def get_block_header_by_hash(self, block_hash: Hash32) -> BlockHeader:
        event = self._prepare_event(inspect.currentframe())
        return self._return_or_raise(await self.event_bus.request(event))

    async def get_block_body_by_hash(self, block_hash: Hash32) -> BlockBody:
        event = self._prepare_event(inspect.currentframe())
        return self._return_or_raise(await self.event_bus.request(event))

   ...

    def _return_or_raise(self, response: LightPeerChainResponse) -> Any:
        if issubclass(type(response.payload), Exception):
            raise Exception(
                "Exception raised in LightPeerChain", response.payload
            ) from response.payload

        return response.payload

    def _prepare_event(self, frame: FrameType) -> LightPeerChainRequest:
        args = self._extract_args(frame.f_locals.values())
        fn_name = frame.f_code.co_name

        payload = None if len(args) == 0 else args

        return LightPeerChainRequest(fn_name, payload)

    @to_tuple
    def _extract_args(self, args: Iterable[Any]) -> Iterable[Any]:
        for arg in args:
            if arg is not self:
                yield arg

```

The reason I went with this design for now is because it seems to do the best job to prevent human error.  The event that is propagated over the event bus is automatically derived from the signature of the methods and the signature of the methods must match `BaseLightPeerChain` which eliminates a wide range of problems imho.

However, let's not focus too much on the concrete implementation of `EventBusLightPeerChain` for now because in this never ending rabbit whole there's another "problem" that I'd like to discuss. Remember that previously the `LightDispatchChain` used to have another thread and turned `async` APIs into blocking APIs using that other thread. Now that this is gone and I introduced a couple of `coro_*` methods obviously the `eth` module doesn't work anymore in regular mode :see_no_evil: 

So, I'm beginning to think we really need an async version of our chain that we can use here. My idea is to just create a super thin wrapper similar to what we did here:

https://github.com/ethereum/py-evm/blob/2ce991b9bbb42c1042f65377de10e761962e67ec/trinity/db/chain.py#L56-L80

Does that sound alright to you?

While I fell deeper and deeper into this rabbit whole I started asking my self from time to time if it would be easier to keep this sync but I keep coming back to this: Asyncio seems like the most sane choice for the RPC server basically [because of what I wrote here](https://github.com/ethereum/py-evm/pull/1212#discussion_r215151984).

cburgdorf:
Alright, I created an `AsyncChainMixin` as a short term solution. Seems to work just fine. If I don't find any further big blockers, then it seems I just have tons of broken tests and cleanup work to get done left :tada:

cburgdorf:
I'm going to close this one as it was a spin off from #1212 and that other PR already has more info and discussion.

PR #1265, opened 2018-09-07
Move DAO fork check into Peer class

### LINKED ISSUE #1234: Trinity crash

pipermerriam:
monkeyridingdog
### What was wrong?

fixes #1234 

### How was it fixed?

Added a new *boot* sequence to the `Peer` class which performs the DAO fork check.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -190,6 +190,7 @@ class BasePeer(BaseService):
         self._subscribers: List[PeerSubscriber] = []
         self.start_time = datetime.datetime.now()
         self.received_msgs: Dict[protocol.Command, int] = collections.defaultdict(int)
+        self.booted = asyncio.Event()
 
         self.egress_mac = egress_mac
         self.ingress_mac = ingress_mac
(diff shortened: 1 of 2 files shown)


p2p/peer.py
```diff
@@ -342,7 +343,81 @@ def is_closing(self) -> bool:
...
+        except OperationCancelled:
+            # If a cancellation happens during boot we suppress it here and
+            # simply exit without setting the `booted` event.
+            return
+        else:
+            self.booted.set()
+
+    async def _boot(self) -> None:
+        try:
+            await self.ensure_same_side_on_dao_fork()
```
cburgdorf: How about we would do something like:

```python

if asyncio.isCoroutineFunction(self.boot_check):
    try:
       await self.boot_check()
    except OperationCancelled:
...
```
And then either pass the `boot_check` on creation or use an inheritence approach to set it.
That way we would keep this concrete DAO check out of the `BasePeer` which strikes me as being more generic in respect to other networks.

p2p/peer.py
```diff
@@ -342,7 +343,81 @@ def is_closing(self) -> bool:
...
+            self.logger.debug("DAO fork check with %s failed: %s", self, err)
+            await self.disconnect(DisconnectReason.useless_peer)
+            raise
+        except Exception as err:
+            self.logger.exception('ERROR BOOTING')
+            raise
+
+    vm_configuration: Tuple[Tuple[int, Type[BaseVM]], ...] = None
+
+    async def ensure_same_side_on_dao_fork(self) -> None:
```
cburgdorf: With https://github.com/ethereum/py-evm/pull/1265/files#r216126130, this whole function could live outside of `BasePeer` to make it more generic.

p2p/peer.py
```diff
@@ -342,7 +343,81 @@ def is_closing(self) -> bool:
...
+            await self.ensure_same_side_on_dao_fork()
+        except DAOForkCheckFailure as err:
+            self.logger.debug("DAO fork check with %s failed: %s", self, err)
+            await self.disconnect(DisconnectReason.useless_peer)
+            raise
+        except Exception as err:
+            self.logger.exception('ERROR BOOTING')
+            raise
+
+    vm_configuration: Tuple[Tuple[int, Type[BaseVM]], ...] = None
```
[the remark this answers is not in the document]
pipermerriam: I intentionally placed it here to keep it close to the line where it's assigned to the peer class.  This is a temporary hack which is being cleaned up in #1270

p2p/peer.py
```diff
@@ -342,7 +343,81 @@ def is_closing(self) -> bool:
...
+            # simply exit without setting the `booted` event.
+            return
+        else:
+            self.booted.set()
+
+    async def _boot(self) -> None:
+        try:
+            await self.ensure_same_side_on_dao_fork()
+        except DAOForkCheckFailure as err:
+            self.logger.debug("DAO fork check with %s failed: %s", self, err)
```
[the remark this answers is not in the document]
pipermerriam: Not for logging statements.  We *alway* do `%` style formatting for logging statements because:

1. The logging module does the formatting for us **and** there are APIs to hook into this formatting to do extra *things* (very minor reason)
2. When you hook up logging to something like rollbar or other logging aggregation services, they can only group messages if we do formatting this way.

p2p/peer.py
pipermerriam: This is another half-step towards the full solution.  Immediately following this PR is #1270 which completely moves this and other EVM stuff out of this module and into the `trinity` codebase.

p2p/peer.py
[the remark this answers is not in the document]
cburgdorf: Ah didn't know that!

p2p/service.py
```diff
@@ -39,7 +39,7 @@ class BaseService(ABC, CancellableMixin):
     _tasks: 'WeakSet[asyncio.Future[Any]]'
     _finished_callbacks: List[Callable[['BaseService'], None]]
     # Number of seconds cancel() will wait for run() to finish.
-    _wait_until_finished_timeout = 5
+    _wait_until_finished_timeout = 10
```
gsalgado: Hmm, we shouldn't have to wait this long... even 5s is already too much, really. If we're seeing timeouts when terminating we must have something that is failing to cleanup

p2p/peer.py
```diff
@@ -342,7 +343,81 @@ def is_closing(self) -> bool:
     async def _cleanup(self) -> None:
         self.close()
 
+    async def boot(self) -> None:
```
gsalgado: I think we should wait for the `started` event or for `self.is_operational` here, no?

p2p/peer.py
```diff
@@ -190,6 +190,7 @@ def __init__(self,
         self._subscribers: List[PeerSubscriber] = []
         self.start_time = datetime.datetime.now()
         self.received_msgs: Dict[protocol.Command, int] = collections.defaultdict(int)
+        self.booted = asyncio.Event()
```
gsalgado: As I mentioned in https://github.com/ethereum/py-evm/issues/1234#issuecomment-418798308 and https://github.com/ethereum/py-evm/pull/1258#issuecomment-419061543, I'm not super keen on this approach, but if you think it's ok we should at least have docstrings making it clear callsites have to wait for `self.booted` before using the peer. And ideally that should be moved into `Peer.events`, no?

p2p/peer.py
pipermerriam: Since `boot` is called from within the `_run`, `started` would always be set here.  That could change if code gets refactored, so I added a check to see if `started` was already set.

p2p/service.py
pipermerriam: Didn't mean to include this change, reverted.

p2p/peer.py
pipermerriam: Will get this updated.  I'm not inclined to add it to the `Peer.events` yet because that requires a decent bit of boilerplate to get the type checks right *or* to just push the event onto the object at runtime and I don't love either of those yet.  (also, it muddles the Peer API which is special with the Service API which is generic).

pipermerriam:
@gsalgado merged with the plan to iterate on the `booted` API a bit if it seems to be problematic or not fix the core issue.

PR #1266, opened 2018-09-07
Fix logging configuration

pipermerriam:
### What was wrong?

Using just `--log-level DEBUG` didn't correctly effect the stderr log level.

### How was it fixed?

Removed the defaults for `--stderr-log-level` so that we can detect when the value has not been provided and to default it to the global log level if provided by `--log-level`.

#### Cute Animal Picture

qwdasx4

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -126,16 +126,32 @@ def main() -> None:
             "networks are supported.".format(args.network_id)
         )
 
-    logger, formatter, handler_stream = setup_trinity_stderr_logging(
-        args.stderr_log_level
+    has_ambigous_logging_config = (
+        args.log_levels is not None and
+        None in args.log_levels and
+        args.stderr_log_level is not None
     )
+    if has_ambigous_logging_config:
+        # setup the logger so we can output the error
+        logger, _, _ = setup_trinity_stderr_logging(logging.ERROR)
+        logger.error(
+            "Ambiguous logging configuration: The logging level for stderr was "
+            "configured with both `--stderr-log-level` and `--log-level`. "
+            "Please remove one of these flags",
+        )
+        sys.exit(1)
+
+    stderr_logger, formatter, handler_stream = setup_trinity_stderr_logging(
+        args.stderr_log_level or args.log_levels.get(None)
+    )
+
     if args.log_levels:
         setup_log_levels(args.log_levels)
 
     try:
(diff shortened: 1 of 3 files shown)


trinity/main.py
```diff
@@ -144,20 +157,18 @@ def main() -> None:
         try:
             initialize_data_dir(chain_config)
         except AmbigiousFileSystem:
-            exit_because_ambigious_filesystem(logger)
+            parser.error(TRINITY_AMBIGIOUS_FILESYSTEM_INFO)
         except MissingPath as e:
-            msg = (
+            parser.error((
```
cburgdorf: I think this would be nicer with a multiline f-string as it:

- doesn't need the `\n`
- just needs one pair of triple quotes at start / end
- can interpolate the `e.path` and get rid of the `format(e.path)` 

```python
parser.error(
    f"""
    It appears that {e.path} does not exist.
    Trinity does not attempt to create directories outside of its root path
    Either manually create the path or ensure you are using a data directory
    inside the XDG_TRINITY_ROOT path"""
)
```

trinity/main.py
```diff
@@ -126,16 +125,30 @@ def main() -> None:
...
+    )
+
     if args.log_levels:
         setup_log_levels(args.log_levels)
 
     try:
         chain_config = ChainConfig.from_parser_args(args)
     except AmbigiousFileSystem:
-        exit_because_ambigious_filesystem(logger)
+        parser.error(TRINITY_AMBIGIOUS_FILESYSTEM_INFO)
```
cburgdorf: Oh `parser.error(...)`. Sweet!

trinity/main.py
```diff
@@ -297,7 +309,6 @@ def kill_trinity_gracefully(logger: logging.Logger,
     time.sleep(0.2)
     kill_process_gracefully(networking_process, logger)
     logger.info('Networking process (pid=%d) terminated', networking_process.pid)
-    sys.exit()
```
cburgdorf: Caution! That one was introduced to support programatically calling `killl_trinity_gracefully` (as opposed to hitting it through `CTRL+C`).

https://github.com/ethereum/py-evm/blob/ca054545c4afea5b0961c849486b189daa750fff/trinity/main.py#L245-L254

If you remove this, you'll notice that `trinity --tx-pool --light` will leave you with a hanging process.

PR #1267, opened 2018-09-07
Remove no-longer-needed catch of OperationCancelled

pipermerriam:
extracted from https://github.com/ethereum/py-evm/pull/1258/files#r215635281

### What was wrong?

We no longer need to handle `OperationCancelled` within the stats reporting task since it is run with `BaseService.run_task()`

### How was it fixed?

Removed the `try/except`

#### Cute Animal Picture


pipermerriam:
Test failure was the OOM issue we've seen some lately.  Might just mean we need to split the byzantium tests into two runs or something.

PR #1270, opened 2018-09-07
WIP: Extract trinity code from p2p module

pipermerriam:
### What was wrong?

WIP

#### Cute Animal Picture


pipermerriam:
@carver 

> `TrinityPeer` doesn't feel like the right name. Would a swarm peer be a subclass of `TrinityPeer`? Maybe it's an `EVMPeer` that is the superclass of eth and les peers.

at first I was going to push back but then I realized that `TrinityPeer` only shows up later **if** there are common carry-over things we need in both `SwarmPeer` and `ETHPeer` but even then we should probably be looking at a composition based approach since the peer classes are already getting a bit unweildy in my opinion.

trinity/protocol/common/peer.py
```diff
@@ -0,0 +1,253 @@
...
+                 privkey: eth_keys_datatypes.PrivateKey,
+                 max_peers: int = DEFAULT_MAX_PEERS,
+                 token: CancelToken = None,
+                 ) -> None:
+        super().__init__(privkey=privkey, max_peers=max_peers, token=token)
+        self.headerdb = headerdb
+        self.network_id = network_id
+        self.vm_configuration = vm_configuration
+
+    def create_peer(self,
```
pipermerriam: I'm going to try updating this to remove the `create_peer` in favor of a composition based approach, adding a `PeerFactory` which is responsible for instantiating peers.  I think it'll make the types cleaner and I'm hoping it makes some of the code here cleaner.

trinity/protocol/common/peer.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def get_peers(self, min_td: int) -> List[EVMPeer]:
+        # TODO: Consider turning this into a method that returns an AsyncIterator, to make it
+        # harder for callsites to get a list of peers while making blocking calls, as those peers
+        # might disconnect in the meantime.
+        peers = tuple(self.connected_nodes.values())
+        return [peer for peer in peers if peer.head_td >= min_td]
+
+
+DEFAULT_PREFERRED_NODES: Dict[int, Tuple[Node, ...]] = {
```
pipermerriam: TODO, pull this out into a `constants.py` file somewhere.

trinity/protocol/common/peer.py
```diff
@@ -0,0 +1,253 @@
...
+    DAOForkCheckFailure,
+)
+
+from .constants import (
+    CHAIN_SPLIT_CHECK_TIMEOUT,
+)
+from .proto import ChainInfo
+
+
+class EVMPeer(BasePeer):
```
pipermerriam: I really dislike this name.  If anyone has any better ideas I'm all :ear: :ear: 

I'm also looking into whether we can do this with composition but I haven't quite seen how that would work.  Best stab looks roughly like:

- something for dealing with the boot sequence (dao check)
- container for metadata (`head_td`, `head_hash`).  Might be a service if it ends up responsible for tracking peer head.
- request/response API

I'm still wrapping my head around this...

pipermerriam:
Alright, I think I have a handle on how I can pick this apart.  Probably won't get to act on it till late tomorrow or Thursday.

PR #1271, opened 2018-09-10
Implement discv5 topic registration

gsalgado:
This basically adds two new APIs that allow us to ask a remote node for
a discv5 ticket (DiscoveryProtocol.get_ticket()) and then use that to
register ourselves under a given topic (DiscoveryProtocol.send_topic_register())

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -54,7 +54,7 @@ from eth.tools.logging import TraceLogger, TRACE_LEVEL_NUM
 
 from cancel_token import CancelToken, OperationCancelled
 
-from p2p.exceptions import AlreadyWaitingDiscoveryResponse, NoEligibleNodes
+from p2p.exceptions import AlreadyWaitingDiscoveryResponse, NoEligibleNodes, UnableToGetDiscV5Ticket
 from p2p import kademlia
 from p2p.peer import PeerPool
 from p2p.service import BaseService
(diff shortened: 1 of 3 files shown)


p2p/discovery.py
```diff
@@ -765,6 +791,45 @@ def process(received_echo: Hash32, response: List[kademlia.Node]) -> None:
 
         return tuple(n for n in nodes if n != self.this_node)
 
+    async def get_ticket(self, node: kademlia.Node, topics: List[bytes]) -> 'Ticket':
+        token = self.send_ping_v5(node, topics)
+        try:
+            got_pong, raw_pong, wait_periods = await self.wait_pong_v5(node, token)
+        except AlreadyWaitingDiscoveryResponse:
+            self.logger.debug("failed to get ticket from %s, already waiting for pong", node)
+            return None
```
pipermerriam: I'm generally :+1: on APIs that require return value checking as they are prone to errors when someone fails to actually check.

p2p/discovery.py
```diff
@@ -765,6 +791,45 @@ def process(received_echo: Hash32, response: List[kademlia.Node]) -> None:
 
         return tuple(n for n in nodes if n != self.this_node)
 
+    async def get_ticket(self, node: kademlia.Node, topics: List[bytes]) -> 'Ticket':
```
pipermerriam: See my comment below about return value checking but.... the signature for this function says it returns a `Ticket` when it should really be `Optional[Ticket]` right?

tests/p2p/test_discovery.py
```diff
@@ -431,6 +450,14 @@ def get_discovery_protocol(seed=b"seed", address=None):
     return discovery.DiscoveryProtocol(privkey, address, [], CancelToken("discovery-test"))
 
 
+async def get_listening_discovery_protocol(event_loop):
+    addr = kademlia.Address('127.0.0.1', random.randint(1024, 9999))
```
pipermerriam: Just noting that I looked (briefly) and it appears that there are no obvious ways to get a random open UDP port

p2p/discovery.py
gsalgado: I guess you mean :-1:?

gsalgado:
I agree that `DiscoveryProtocol` may be getting unnecessarily complex; I'll have a look into ways to simplify it

p2p/discovery.py
gsalgado: That's right

p2p/discovery.py
gsalgado: Anyway, I've changed it to raise an exception instead

pipermerriam: :joy_cat: yup, :-1:

PR #1276, opened 2018-09-11
New cmdline flag to enable v5 topic discovery

gsalgado:
With this we should be able to find LES nodes much faster. Still experimental, but eventually we
should turn it on by default when running in light mode, like geth does.

### DIFF
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -15,6 +15,10 @@ from cancel_token import CancelToken, OperationCancelled
 
 from eth.chains import AsyncChain
 
+from eth_typing import BlockNumber
+
+from eth.constants import GENESIS_BLOCK_NUMBER
+
 from p2p.auth import (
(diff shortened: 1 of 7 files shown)


trinity/server.py
```diff
@@ -163,6 +178,17 @@ def _make_peer_pool(self) -> PeerPool:
         self.logger.info("Closing server...")
         await self._close_tcp_listener()
 
+    def _get_discv5_topic(self) -> bytes:
```
pipermerriam: This looks like a good candidate to 

- extract to a utility function (probably passing in the `genesis_hash` as an argument) 
- write explicit tests against the `LESProtocol` classes and known genesis blocks (ropsten/mainnet) that we support.

gsalgado: Indeed, I've done that

PR #1277, opened 2018-09-11
BaseService extensions

carver:
### What was wrong?

Using `loop.call_later` directly won't get cancelled with the service.

`BaseService` wraps a lot of coroutines, which makes for bad messages and stack traces.

### How was it fixed?

- Added a new `BaseService.call_later()` that cancels with `BaseService`.
- Decorate the wrappers with `@functools.wraps(awaitable)`

#### Cute Animal Picture

### DIFF
--- a/p2p/service.py
+++ b/p2p/service.py
@@ -129,7 +129,7 @@ class BaseService(ABC, CancellableMixin):
 
         If it raises OperationCancelled, that is caught and ignored.
         """
-        @functools.wraps(awaitable)
+        @functools.wraps(awaitable)  # type: ignore
         async def _run_task_wrapper() -> None:
             self.logger.trace("Running task %s", awaitable)
             try:
@@ -149,7 +149,7 @@ class BaseService(ABC, CancellableMixin):
         Like :meth:`run_task` but if the task ends without cancelling, then this
         this service will terminate as well.
         """
-        @functools.wraps(awaitable)
+        @functools.wraps(awaitable)  # type: ignore
         async def _run_daemon_task_wrapper() -> None:
             try:
                 await awaitable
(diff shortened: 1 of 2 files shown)


p2p/service.py
```diff
@@ -148,6 +149,7 @@ def run_daemon_task(self, awaitable: Awaitable[Any]) -> None:
         Like :meth:`run_task` but if the task ends without cancelling, then this
         this service will terminate as well.
         """
+        @functools.wraps(awaitable)  # type: ignore
```
pipermerriam: I think this is sufficient since it appears that the `__name__` is preserved on the couroutine object.  It however doesn't preserve things like `__doc__` but I think that is acceptable.

p2p/service.py
```diff
@@ -213,6 +216,15 @@ def run_daemon(self, service: 'BaseService') -> None:
 
         self.run_task(_run_daemon_wrapper())
 
+    def call_later(self, delay: float, callback: 'Callable[..., None]', *args: Any) -> None:
+
+        @functools.wraps(callback)
+        async def _call_later_wrapped() -> None:
+            self.sleep(delay)
```
pipermerriam: Is there a reason why you are not using the [`AbstractEventLoop.call_later`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.AbstractEventLoop.call_later) API?

pipermerriam: Reason I ask is that if we use that, we shouldn't have to wrap the function if we use that API, though we will need to add another container to store the `TimeHandle` return values and cancel them upon cancellation.

carver: > we will need to add another container to store the TimeHandle return values and cancel them upon cancellation.

Yup, this is why. I didn't want to add yet another set that we had to remember to update in all the places. I started writing a `WeakSet[asyncio.TimerHandle]` and it just felt wrong to keep adding complexity to `BaseService` as it's already getting heavy. We can't treat it like the other tasks, and would explicitly have to cancel them right away at cancel time, which is fairly different from how the other tasks are dealt with (since they all usually die on their own with the cancel token).

pipermerriam: I guess it seems like a relatively simple addition, though I agree our `BaseService` API has grown substantially.

One way we might be able to battle some of that complexity would be beefing up the test suite, and looking into extracting into a dedicated library.  Forces us to document it, and think about breaking API changes which we currently don't do.

PR #1279, opened 2018-09-12
batch block persist to avoid inconsistent state

carver:
### What was wrong?

We can end up with inconsistent state when shutting down trinity (say, a header persisted, but the canonical head not updated, or missing the transaction or uncle data, etc). One thing that contributes to inconsistent state is that we persist in parts.

### How was it fixed?

Batch together all the writes into a single commit. It is still possible for the writes to be interrupted mid-stream, but it's much less likely. At some point, it's worth changing this to a proper LevelDB transaction, but that is a bigger change that's worth doing broader work on. (and I don't want to get distracted from dealing with the current inconsistent state issues during sync)

#### Cute Animal Picture


carver:
I decided it's better to just do this the right way, and get real levelDB transaction atomic commits. So this is still a WIP.

carver:
Wrapping up for the day. Last thing I'd want to do is detect and blow up if the `batch_write()` context was entered twice (unless leveldb supports it, in which case, I might try to add support into `BatchableDB`)

carver:
All of the prep work, and type changing is really just to support 13211f0 and 0f42713, which is relatively small, but a huge win IMO.

carver:
This is a little big. I'll peel it off into parts, starting with #1295

carver:
... and onto #1296

carver:
and #1304 -- which fixes some conceptual issues that will slightly change the approach in this PR: it will require that header and chain db can write to an explicitly provided database handle, instead of their internal `self.db`, when desired.

cburgdorf:
Looks good to me :+1: It's a shame that our benchmark suite still entirely relies on the `MemoryDB` which pretty much eliminates any chance to see the perf impact of this.

Issue #1282, opened 2018-09-12
Remote peers disconnecting with a bad_protocol reason immediately after handshake


gsalgado:
I'm testing discv5 in light mode and am seeing lots of peers disconnecting from us with a bad_protocol reason immediately after the handshake is successfully completed. Need to investigate but I guess we're doing something wrong

```
   DEBUG  09-04 22:28:38      server  Receiving handshake from Address(80.111.20.180:udp:54907|tcp:54907)
   DEBUG  09-04 22:28:39        peer  Finished P2P handshake with <Node(0xc79e@80.111.20.180)>, using sub-protocol (les, 2)
   DEBUG  09-04 22:28:39       proto  Sending LES/Status msg: {'announceType': 1, 'protocolVersion': 2, 'networkId': 1, 'headTd': 169726491918314, 'headHash': b'R\xf4FQg\xa5vw\xd1\xca!                           \x0f=s\xbay\xca8\xe3\xbe\xd7#/\x02l\x05\x8a\xf2\x81\xac\xc8+', 'headNum': 3629, 'genesisHash': b'\xd4\xe5g@\xf8v\xae\xf8\xc0\x10\xb8j@\xd5\xf5gE\xa1\x18\xd0\x90j4\xe6\x9a\xec\x8c\r\xb1\xcb\x8f\xa3',             'serveHeaders': None, 'serveChainSince': 0, 'txRelay': None}
   DEBUG  09-04 22:28:39        peer  Finished (les, 2) handshake with <Node(0xc79e@80.111.20.180)>
   DEBUG  09-04 22:28:39    managers  Launching ResponseCandidateStream for peer LESPeer <Node(0xc79e@80.111.20.180)>
   DEBUG  09-04 22:28:39        peer  LESPeer <Node(0xc79e@80.111.20.180)> disconnected: subprotocol_error
   DEBUG  09-04 22:28:39     service  <ResponseCandidateStream(LESPeer <Node(0xc79e@80.111.20.180)>, <class 'trinity.protocol.les.commands.BlockHeaders'>)> finished: Cancellation requested by LESPeer token
   DEBUG  09-04 22:28:39     service  <ResponseCandidateStream(LESPeer <Node(0xc79e@80.111.20.180)>, <class 'trinity.protocol.les.commands.BlockHeaders'>)> halted cleanly
   DEBUG  09-04 22:28:39     service  LESPeer <Node(0xc79e@80.111.20.180)> halted cleanly
```

PR #1283, opened 2018-09-12
Fail handshake when LES peer doesn't serve headers

gsalgado:
Closes: #1282

### DIFF
--- a/trinity/protocol/les/peer.py
+++ b/trinity/protocol/les/peer.py
@@ -87,7 +87,12 @@ class LESPeer(BasePeer):
             raise HandshakeFailure(
                 "{} genesis ({}) does not match ours ({}), disconnecting".format(
                     self, encode_hex(msg['genesisHash']), genesis.hex_hash))
-        # TODO: Disconnect if the remote doesn't serve headers.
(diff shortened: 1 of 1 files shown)


trinity/protocol/les/peer.py
```diff
@@ -87,7 +87,12 @@ def handle_sub_proto_msg(self, cmd: Command, msg: _DecodedMsgType) -> None:
             raise HandshakeFailure(
                 "{} genesis ({}) does not match ours ({}), disconnecting".format(
                     self, encode_hex(msg['genesisHash']), genesis.hex_hash))
-        # TODO: Disconnect if the remote doesn't serve headers.
+        # Eventually we might want to keep connections to peers where we are the only side serving
+        # data, but right now both our chain syncer and the Peer.boot() method expect the remote
+        # to reply to header requests, so if they don't we simply disconnect here.
+        if 'serveHeaders' not in msg:
+            await self.disconnect(DisconnectReason.useless_peer)
+            raise HandshakeFailure("{} doesn't serve headers, disconnecting".format(self))
```
pipermerriam: This should use `f` string formatting

```python
raise HandshakeFailure(f"{self} doesn't serve headers, disconnecting")
```

PR #1289, opened 2018-09-12
Fix sphinx doctest issue

pipermerriam:
### What was wrong?

Sphinx has a bug:

https://github.com/sphinx-doc/sphinx/issues/3494

### How was it fixed?

Pinned to earlier version until it is resolved.

#### Cute Animal Picture


carver:
> Pinned to earlier version until it is resolved.

Bug is 1.5 yrs old, may be pinned a while... X)

pipermerriam:
Or until we drop 3.5 support.

PR #1290, opened 2018-09-12
Fast sync start forked

carver:
### What was wrong?

When running fast sync, it might start on a fork from the end of the last sync. This was causing a crash.

Other similar problems are currently possible (but [shouldn't be for long](https://github.com/ethereum/py-evm/pull/1279)), like starting later than the canonical head, because headers were saved into an inconsistent state.

### How was it fixed?

Catch the `ValidationError` when trying to register the block/receipt download tasks. Confirm that the parent header already exists (which the common sync should have already done). Then allow the sync to continue, with a log.

This required a change in `OrderedTaskPreparation`: it can now accept multiple "already finished" tasks.

Some other bonus fixes:

- speed up uncle validation if a block has no uncles
- logging a weird RLP bug from an overnight ropsten run
- await a sleep from an old PR :man_facepalming: 

#### Cute Animal Picture

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -52,6 +52,7 @@ from eth.db.header import (
 )
 from eth.constants import (
     BLANK_ROOT_HASH,
+    EMPTY_UNCLE_HASH,
     MAX_UNCLE_DEPTH,
 )
 from eth.estimators import (
@@ -720,6 +721,10 @@ class Chain(BaseChain):
         """
         Validate the uncles for the given block.
         """
+        if block.header.uncles_hash == EMPTY_UNCLE_HASH and len(block.uncles) == 0:
+            # optimization to avoid checking ancestors if the block has no uncles
+            return
+
         # Check for duplicates
         uncle_groups = groupby(operator.attrgetter('hash'), block.uncles)
         duplicate_uncles = tuple(sorted(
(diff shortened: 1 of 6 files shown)


eth/chains/base.py
```diff
@@ -720,6 +721,10 @@ def validate_uncles(self, block: BaseBlock) -> None:
         """
         Validate the uncles for the given block.
         """
+        if block.header.uncles_hash == EMPTY_UNCLE_HASH and len(block.uncles) == 0:
```
pipermerriam: If you get back in here, you might expand this to:

```python
has_uncles = len(block.uncles) > 0
should_have_uncles = block.header.uncles_hash != EMPTY_UNCLE_HASH

if not has_uncles and not should_have_uncles:
    return #otpimization
elif has_uncles and not should_have_uncles:
    raise ValidationError("Block has uncles but header suggests uncles should be empty")
elif should_have_uncles and not has_uncles:
    raise ValidationError("Header suggests block should have uncles but block has no uncles")
```

p2p/service.py
```diff
@@ -220,7 +220,7 @@ def call_later(self, delay: float, callback: 'Callable[..., None]', *args: Any)
 
         @functools.wraps(callback)
         async def _call_later_wrapped() -> None:
-            self.sleep(delay)
+            await self.sleep(delay)
```
pipermerriam: :open_mouth:

PR #1292, opened 2018-09-13
RFC: Run the Node service in trinity's main process

gsalgado:
How many times have you wished you could stick some pdb.set_trace() calls in the chain-syncer code
to debug a trinity run? This is something that always frustrated me, but the fact that we run the
chain syncer in a sub process meant it was not easy (if possible at all?) to do so.

Well, it recently ocurred to me that trinity's main process is actually not doing much work and
if we run the Node service in the main process it will be much easier to debug and we'd significantly
reduce the amount of IPC calls we make (since we'd no longer use the queued logger there).

How would you guys feel about such a change? Am I overlooking something? Is there anything else that
warrants running the Node service in a separate process?


pipermerriam:
I have no objections that come to mind. :+1:

gsalgado:
@cburgdorf how should I go about combining the two plugin_managers that would now live in the main process?

cburgdorf:
So, one thing that comes to my mind is that currently the central event bus is running in the main process which was convenient to begin with because:

1. the main process naturally holds handles to each other process to stick `Endpoint`s in
2. As @gsalgado it doesn't do much which is perfect for the event bus because it really just runs an event loop to pass messages on to other processes.

This would change quite a bit when the node itself runs in the main process. I noticed that the syncing process can block quite a lot which would in turn slow down message passing to all other processes.

However, I'd actually love to have the central event bus (that piece that creates endpoints for each other process) to live in its own process so that we never have to worry about anything blocking it and slowing message delegation down. I'm just not sure how that would look like yet.

So, in general, no objections we just should make sure to figure out how to deal with that.

Also, I'd hope we could get #1212 merged in *before* this.

>how should I go about combining the two plugin_managers that would now live in the main process?

I guess we need to redesign the things around the process scopes a bit. We would basically go from our current 3 different scopes to just 2. https://github.com/ethereum/py-evm/blob/master/trinity/extensibility/plugin.py#L51-L62

gsalgado:
@cburgdorf Any idea what is causing trinity to hang after the main() process has successfully killed the DB subprocess and returned? As shown below, after I hit Ctrl+C, trinity.main() returns, but it hangs there until I Ctrl+C again, at which point I get those errors from lahja

```
    INFO  09-13 17:33:38     service  Starting fast-sync; current head: #35668
^C    INFO  09-13 17:33:40      server  Closing server...
    INFO  09-13 17:33:45     service  Timed out waiting for <trinity.nodes.mainnet.MainnetFullNode object at 0x7fec9d55bc18> to finish its cleanup, forcibly cancelling pending tasks and exiting anyway
    INFO  09-13 17:33:46        main  Keyboard Interrupt: Stopping
    INFO  09-13 17:33:46         ipc  Sent SIGINT to process 8960, waiting 10 seconds for it to terminate
    INFO  09-13 17:33:46         ipc  Process 8960 has already terminated
    INFO  09-13 17:33:46        main  DB server process (pid=8960) terminated
    INFO  09-13 17:33:46        main  Reached the end of trinity.main()
^CException ignored in: <module 'threading' from '/usr/lib/python3.6/threading.py'>
  File "/usr/lib/python3.6/threading.py", line 1294, in _shutdown
    t.join()
  File "/usr/lib/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
Task was destroyed but it is pending!
task: <Task pending coro=<Endpoint._try_connect() running at /home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/lahja/endpoint.py:70> wait_for=<Task pending coro=<Endpoint._connect() running at /home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/lahja/endpoint.py:78> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /usr/lib/python3.6/asyncio/futures.py:403, <TaskWakeupMethWrapper object at 0x7fec97d350d8>()]> cb=[<TaskWakeupMethWrapper object at 0x7fec97d79738>()]>>
Task was destroyed but it is pending!
task: <Task pending coro=<EventBus._start() running at /home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/lahja/eventbus.py:59> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /usr/lib/python3.6/asyncio/futures.py:403, <TaskWakeupMethWrapper object at 0x7fec97d798b8>()]>>
Task was destroyed but it is pending!
task: <Task pending coro=<Endpoint._connect() running at /home/salgado/virtualenvs/pyevm/lib/python3.6/site-packages/lahja/endpoint.py:78> wait_for=<Future pending cb=[_chain_future.<locals>._call_check_cancel() at /usr/lib/python3.6/asyncio/futures.py:403, <TaskWakeupMethWrapper object at 0x7fec97d350d8>()]> cb=[<TaskWakeupMethWrapper object at 0x7fec97d79738>()]>
```

gsalgado:
As @pipermerriam pointed out, both https://pypi.org/project/rpdb/#description and https://pypi.org/project/remote-pdb/#description should allow us to interactively debug any trinity subprocess, so I'm no longer sure it's worth pursuing this, although it'd certainly simplify trinity.main() a bit

pipermerriam:
@gsalgado I'm now inclined to close if you think you can get what you want from those libraries.  I do like the main process being simple and would like to keep that if possible, though my reasons are thin (just liking the simplicity of the main process doing so little).

cburgdorf:
>Any idea what is causing trinity to hang after the main() process has successfully killed the DB subprocess and returned

I think that is partly because of a bug in lahja that is already fixed in a newer version and coming with [this commit](https://github.com/ethereum/py-evm/pull/1212/commits/0dc29c353cf9d65cda215fc866e51e9aa69142ba).

However, I also have a slight preference of keeping the main process as is mostly because:

- the current event bus topology benefits from a (mostly) idle main process
- we will continue to refactor things out into isolated plugins that spawn their own processes (e.g. #1212 moves the JSON-RPC Server but I guess things like the transaction pool will follow)
- that previous point basically means that having a better way of debugging other processes still seems like something we would benefit more from

PR #1294, opened 2018-09-13
Piper/finish extracting trinity logic from p2p

pipermerriam:
Second part of #1286 

### What was wrong?

PR #1286 setup the structure to get much of the remaining trinity logic out of the `p2p` module, but didn't move all of it in favor of a more readable PR.

### How was it fixed?

Finished moving things out of the `p2p` module.

#### Cute Animal Picture

a-monkey-riding-a-dog-a-monkey-riding-a-dog11

### DIFF
--- a/trinity/protocol/common/peer.py
+++ b/trinity/protocol/common/peer.py
@@ -1,0 +1,102 @@
+import operator
+import random
+from typing import (
+    Dict,
+    List,
+    NamedTuple,
+    Tuple,
+    Type,
+)
+
+from eth_typing import (
+    BlockNumber,
+    Hash32,
+)
+
+from eth_utils.toolz import groupby
+
+from eth.constants import GENESIS_BLOCK_NUMBER
+from eth.rlp.headers import BlockHeader
+from eth.vm.base import BaseVM
+
+from p2p.exceptions import NoConnectedPeers
+from p2p.kademlia import Node
+from p2p.peer import (
+    BasePeer,
+    BasePeerFactory,
+    BasePeerPool,
+)
+
+from trinity.db.header import BaseAsyncHeaderDB
+
+from .boot import DAOCheckBootManager
+from .context import ChainContext
+
+
+class ChainInfo(NamedTuple):
+    block_number: BlockNumber
+    block_hash: Hash32
+    total_difficulty: int
+    genesis_hash: Hash32
+
+
+class BaseChainPeer(BasePeer):
+    boot_manager_class = DAOCheckBootManager
+    context: ChainContext
+
+    head_td: int = None
+    head_hash: Hash32 = None
+
+    @property
+    def headerdb(self) -> BaseAsyncHeaderDB:
+        return self.context.headerdb
+
+    @property
+    def network_id(self) -> int:
+        return self.context.network_id
+
+    @property
+    def vm_configuration(self) -> Tuple[Tuple[int, Type[BaseVM]], ...]:
+        return self.context.vm_configuration
+
+    @property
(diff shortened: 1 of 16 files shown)


trinity/protocol/common/peer.py
```diff
@@ -0,0 +1,102 @@
...
+class ChainInfo(NamedTuple):
+    block_number: BlockNumber
+    block_hash: Hash32
+    total_difficulty: int
+    genesis_hash: Hash32
+
+
+class BaseChainPeer(BasePeer):
+    boot_manager_class = DAOCheckBootManager
+    context: ChainContext
```
carver: Ok, so I thought about `ChainContext` for a bit longer than I care to admit. But I think I found a good direction:

If you look at where the properties derived from `context` are used, it's mostly in `ETHProtocol`. That's actually great, and makes a lot of sense: these protocol specific values are only used where we know they should have been set, inside `ETHProtocol` (as opposed to the base peer kind of knowing about them, but not really because it's this opaque type-busting `context` thing).

So the idea is `ETHProtocol` should actually be the thing that receives these protocol specific connection data on init, as part of the peer factory, which would set curried versions of ETHProtocol on `ETHPeer`s `_supported_sub_protocols`. If you follow that rabbit hole, I think that means you want to break the dependency that protocol has on peer. The subprotocols would expose one unifying API to the peer: `def encode_request(request: BaseRequest) -> (header: bytes, body: bytes)`. Then the peer would use that to expose `peer.send_request(request)`, which uses the protocol encoding under the hood.

So it's potentially a big-ish change. Also, my thoughts/explanation is probably too messy to really convey what I'm thinking. Luckily, I think most of this PR wouldn't change (or at least, not any more than the pre-PR code would change). So although I want to keep exploring the change, I don't think it needs to hold up the PR. Probably the easiest way forward would be to merge this PR and then for me to make a rough WIP version of what I'm thinking, for us to talk about.

trinity/protocol/eth/peer.py
```diff
@@ -83,9 +84,9 @@ def handle_sub_proto_msg(self, cmd: Command, msg: _DecodedMsgType) -> None:
         self.head_hash = msg['best_hash']
 
 
-class ETHPeerFactory(BasePeerFactory):
+class ETHPeerFactory(BaseChainPeerFactory):
     peer_class = ETHPeer
 
 
-class ETHPeerPool(BasePeerPool):
+class ETHPeerPool(BaseChainPoolPool):
```
carver: `BaseChainPeerPool`

trinity/protocol/les/peer.py
```diff
@@ -103,9 +104,9 @@ def handle_sub_proto_msg(self, cmd: Command, msg: _DecodedMsgType) -> None:
         self.head_number = msg['headNum']
 
 
-class LESPeerFactory(BasePeerFactory):
+class LESPeerFactory(BaseChainPeerFactory):
     peer_class = LESPeer
 
 
-class LESPeerPool(BasePeerPool):
+class LESPeerPool(BaseChainPoolPool):
```
carver: `BaseChainPeerPool`

trinity/protocol/common/peer.py
pipermerriam: I think you're onto something.  Do you mind writing up an issue for this. I am pretty sure I see where you're going but I think maybe we start with an issue write-up and we can hash out any questions there.  This feels really *right* in terms of getting our abstractions in the right place.

coreyandersonbuilders:
My account was hi-jacked and fraud committed against me

PR #1295, opened 2018-09-13
AtomicDB

carver:
### What was wrong?

There was no way to use LevelDB's built-in write batching.

### How was it fixed?

- Added a new `BaseBatchableDB` which extends the BaseDB API, for write batching
- Added a new `BatchableDB` as a parallel to `MemoryDB` that supports batching. (Note that it's different from `BatchDB`, which *always* batches requests until commit. `BatchableDB` will immediately commit any changes outside of a `batch_write` context.)
- Added support for `batch_write` to the `LevelDB` backend

This is not yet wired up to anything, it's a few starting commits PR stripped off of #1279 

#### Cute Animal Picture

### DIFF
--- a/eth/db/batch.py
+++ b/eth/db/batch.py
@@ -1,5 +1,4 @@
 import logging
-from typing import Type, Dict  # noqa: F401
 
 from eth.db.diff import (
     DBDiff,
@@ -15,7 +14,7 @@ class BatchDB(BaseDB):
     which represents as a dictionary of database keys and values.
     This class should be usable as a context manager, the changes either all fail or all succeed.
     Upon exiting the context, it writes all of the key value pairs from the cache into
-    the underlying database. If anyerror occurred before committing phase,
+    the underlying database. If any error occurred before committing phase,
     we would not apply commits at all.
     """
     logger = logging.getLogger("eth.db.BatchDB")
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -721,9 +721,16 @@ class Chain(BaseChain):
         """
         Validate the uncles for the given block.
         """
-        if block.header.uncles_hash == EMPTY_UNCLE_HASH and len(block.uncles) == 0:
-            # optimization to avoid checking ancestors if the block has no uncles
+        has_uncles = len(block.uncles) > 0
+        should_have_uncles = block.header.uncles_hash != EMPTY_UNCLE_HASH
+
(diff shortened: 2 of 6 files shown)


eth/db/batch.py
```diff
@@ -74,3 +80,88 @@ def __delitem__(self, key: bytes) -> None:
...
+    wrapped_db = None  # type: BaseDB
+    _track_diff = None  # type: DBDiffTracker
+
+    def __init__(self, wrapped_db: BaseDB = None) -> None:
+        if wrapped_db is None:
+            self.wrapped_db = MemoryDB()
+        else:
+            self.wrapped_db = wrapped_db
+        self._track_diff = DBDiffTracker()
+        self._is_batching = False
```
pipermerriam: I think this will be cleaner if you use a `threading.Lock` since they support context manager APIs which should cleanup the internals of `write_batch`.

tests/database/test_batchable_db.py
```diff
@@ -0,0 +1,131 @@
...
+    batch_db.set(b'key-2', b'value-2')
+    assert batch_db.get(b'key-2') == b'value-2'
+
+    # keys should be immediately set in base db.
+    assert base_db.get(b'key-1') == b'value-1'
+    assert base_db.get(b'key-2') == b'value-2'
+
+
+def test_batchable_db_cannot_recursively_batch(base_db, batch_db):
+    with pytest.raises(ValidationError):
```
pipermerriam: To accurately know where the exception comes from, wouldn't it be cleaner to have this inside the first `with batch_db.write_batch()` and above the inner `with batch_db.write_batch`?

The way this is currently written, the outer call to `write_batch` could be throwing the `ValidationError` and the test would pass.

pipermerriam:
Naming wise, this might be better called `AtomicDB` which helps differentiate it from `BatchDB`, and the context method could be `with atomicdb.atomic_batch()` which I think communicates what is happening a little clearer.

tests/database/test_batchable_db.py
carver: Yeah, I got stuck on this.

If `pytest.raises` is inside the first context, then the it catches the exception, and the outer writes will succeed. We know the outer one works from other tests. I wanted to test that the outer writes will be reverted. But I guess we know that works from other tests, too.

I guess I'll move it inside.

Issue #821, opened 2018-05-30
Trying to peer with self?


MysticRyuujin:
* py-evm Version: `Trinity v0.1.0-alpha.1 "Valentina Tereshkova"`
* OS: `Ubuntu 17.10`
* Python Version (python --version): `Python 3.6.3`
* Environment (output of `pip freeze`):
```
aiohttp==2.3.10
asn1crypto==0.24.0
async-lru==0.1.0
async-timeout==3.0.0
attrdict==2.0.0
backcall==0.1.0
certifi==2018.4.16
cffi==1.11.5
chardet==3.0.4
coincurve==7.1.0
command-not-found==0.3
cryptography==2.2.2
cytoolz==0.9.0.1
decorator==4.3.0
distro-info==0.17
eth-abi==1.1.1
eth-account==0.2.3
eth-bloom==1.0.0
eth-hash==0.1.4
eth-keyfile==0.5.1
eth-keys==0.2.0b3
eth-rlp==0.1.2
eth-utils==1.0.3
hexbytes==0.1.0
idna==2.6
idna-ssl==1.0.1
ipython==6.4.0
ipython-genutils==0.2.0
jedi==0.12.0
keyring==10.4.0
keyrings.alt==2.2
language-selector==0.1
lru-dict==1.1.6
lxml==4.2.1
multidict==4.3.1
netdisco==1.4.1
netifaces==0.10.7
parsimonious==0.8.0
parso==0.2.1
pexpect==4.6.0
pickleshare==0.7.4
plyvel==1.0.4
prompt-toolkit==1.0.15
ptyprocess==0.5.2
py-ecc==1.4.2
py-evm==0.2.0a19
pycparser==2.18
pycrypto==2.6.1
pycryptodome==3.6.1
pyethash==0.1.27
Pygments==2.2.0
pygobject==3.24.1
pysha3==1.0.2
python-apt==1.4.0b3
python-dateutil==2.7.3
pyxdg==0.25
PyYAML==3.12
requests==2.18.4
rlp==1.0.1
SecretStorage==2.3.1
simplegeneric==0.8.1
six==1.11.0
ssh-import-id==5.7
toolz==0.9.0
traitlets==4.3.2
trie==1.3.7
trinity==0.1.0a3
ufw==0.35
unattended-upgrades==0.1
uPnPClient==0.0.8
urllib3==1.22
wcwidth==0.1.7
web3==4.2.1
websockets==5.0.1
yarl==1.2.4
[...]

PR #1297, opened 2018-09-14
During node discovery, skip those with invalid addresses

gsalgado:
Sometimes a NEIGHBOURS response from remote peers will include nodes
with a loopback or private network address that is not reachable to us,
so we must skip those. We must also skip those on reserved networks or
with an unspecified address (0.0.0.0).

Closes: #821


p2p/discovery.py
```diff
@@ -1083,11 +1083,16 @@ def __repr__(self) -> str:
 
 @to_list
 def _extract_nodes_from_payload(
-        payload: List[Tuple[str, str, str, str]]) -> Iterator[kademlia.Node]:
+        sender: kademlia.Address,
+        payload: List[Tuple[str, str, str, str]],
+        logger: TraceLogger) -> Iterator[kademlia.Node]:
```
pipermerriam: I am thinking I'd like to move away from this pattern of passing loggers around, in favor of just either a function specific logger or a module level logger.

PR #1299, opened 2018-09-17
Support for custom genesis via EIP1085

pipermerriam:
Continuation of https://github.com/ethereum/py-evm/pull/1247

fixes https://github.com/ethereum/py-evm/issues/722

### What was wrong?

Trinity should be able to run custom chains beyond just mainnet and ropsten.

### How was it fixed?

Now mainnet and ropsten are done in the same way that a custom chain would be, via an EIP1085 genesis file.  There is now a new cli option `--genesis` which can be used to specify either the raw JSON string or a filesystem path to a JSON document.

#### Cute Animal Picture


pipermerriam:
Don't bother looking at this.  It's piling up into 3-4 separate PRs that need to be untangled.

voith:
@pipermerriam from what I read, [EIP1085](https://github.com/ethereum/EIPs/issues/1085) doesn't seem to be finalised. 
Do you have any timeline for this PR to land. And even if it lands do you think the spec will remian unchanged? I'm asking because I'd like to integrate this into `eth-teser` as part of https://github.com/ethereum/eth-tester/issues/129.

pipermerriam:
@voith no idea on either (this PR or finalization of spec).  I don't think that should block you however from moving forward with using it in `eth-tester`.  Make sure the API is marked in some way as experimental and subject to breaking changes.

trinity/assets/eip1085.schema.json
```diff
@@ -0,0 +1,153 @@
...
+          "pattern": "^0x([0-9a-fA-F]{2}){32}$"
+        },
+        "gasLimit": {
+          "type": "string",
+          "pattern": "^0x([0-9a-fA-F]{2})+$"
+        }
+      }
+    },
+    "Accounts": {
+      "title": "Genesis block parameters",
```
veox: Minor: looks like copy-paste left-over, here and below, in `Account`.

pipermerriam:
More updates on this:

@carver and @cburgdorf

This is now yielding some interesting and fundamental flaws with how we currently do multiprocess communication, but I think it's a good thing.  The concise issue is that with the introduction of custom genesis configurations, we end up with un-pickleable VM classes since they must be dynamically constructed based on the genesis configuration (this is primaily due to the dao fork, but even without this we'd have the same problem with POA networks since we'll have to inject the proper consensus engine).

That means all of `Chain.get_vm_configuration()`, `Chain.get_vm`, `Chain.get_vm_class_for_block_number` and probably a number of others can no longer be called across the multiprocessing boundary.  I think the *proper* way to fix this is to do some or all of the following.

1. Remove the `chain` object from the things served up by the database process and require local chain instances in each process (thought the underlying database connection may still be across process boundaries)
2. Switch to `RocksDB` which allows for multiple database connections at which point we can potentially get rid of the database process entirely.

I see the potential for this to break some things, specifically surrounding our assumptions about where processing will happen and thus which calls are *safe* to make within the networking process without risking blocking too long and timing out with all of our peers.

carver:
> require local chain instances in each process

This sounds right to me. I believe I've done all the necessary work on Chain to make this fairly painless (no local state, all state depends on the database). The only necessary change was removing the canonical head attribute into the MiningChain.

trinity/initialization.py
```diff
@@ -106,8 +106,14 @@ def initialize_database(chain_config: ChainConfig, chaindb: AsyncChainDB, base_d
...
+            address: account.to_dict()
+            for address, account
+            in chain_config.genesis_state.items()
+        }
         chain_class.from_genesis(
             base_db=base_db,
-            genesis_params=chain_config.genesis_params,
-            genesis_state=chain_config.genesis_state,
+            genesis_params=genesis_params,
+            genesis_state=genesis_state,
```
carver: Maybe some of this logic that is just dumping in data from `chain_config` could be pushed inside `chain_config`, so the API could be:

```py
chain = chain_config.launch_genesis(base_db)
```

If you like it

tests/trinity/core/configuration/test_chain_config_object.py
```diff
@@ -51,17 +47,20 @@ def test_chain_config_one_of_header_or_params_is_required():
 )
 def test_chain_config_from_preconfigured_network(network_id):
     chain_config = ChainConfig.from_preconfigured_network(network_id)
-
-    assert chain_config.is_preconfigured_chain is True
+    chain = chain_config.full_chain_class.from_genesis(
+        AtomicDB(MemoryDB()),
+        chain_config.genesis_params.to_dict(),
+        {address: account.to_dict() for address, account in chain_config.genesis_state.items()}
```
carver: Here too.

pipermerriam:
Working on rebasing and splitting this up but also pleased to report that a basic test of this code seems to be working right  now.

pipermerriam:
Pulled out 6 things from this. will rebase when they are merged:

- https://github.com/ethereum/py-evm/pull/1433
- https://github.com/ethereum/py-evm/pull/1434
- https://github.com/ethereum/py-evm/pull/1435
- https://github.com/ethereum/py-evm/pull/1436
- https://github.com/ethereum/py-evm/pull/1437
- https://github.com/ethereum/py-evm/pull/1438

pipermerriam:
@carver looking for feedback, still not sure this is really "done" but I'd like to get some review before I take it any further.  I think I can still yank out maybe 100-200 lines from the diff into smaller pull requests but it's getting harder.

pipermerriam:
grumble,  There is no reason I can tell for the integration test to be failing.  It passes locally and [passed in python3.7](https://circleci.com/gh/ethereum/py-evm/98101?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

carver:
> grumble, There is no reason I can tell for the integration test to be failing. It passes locally and passed in python3.7

Huh, no idea. It looks like it's just hanging, huh?

tests/trinity/core/configuration/test_chain_config_object.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.parametrize(
+    'network_id',
+    (MAINNET_NETWORK_ID, ROPSTEN_NETWORK_ID),
+)
+def test_chain_config_from_preconfigured_network(network_id):
+    chain_config = ChainConfig.from_preconfigured_network(network_id)
+    chain = chain_config.initialize_chain(AtomicDB(MemoryDB()))
+
+    if network_id == MAINNET_NETWORK_ID:
+        assert chain_config.chain_id == MainnetChain.chain_id
```
carver: I guess it's super trivial, but maybe compare network IDs too.

tests/trinity/core/configuration/test_chain_config_object.py
```diff
@@ -0,0 +1,105 @@
...
+        assert_vm_configuration_equal(chain_config.vm_configuration, MainnetChain.vm_configuration)
+        assert chain.get_canonical_head() == MAINNET_GENESIS_HEADER
+    elif network_id == ROPSTEN_NETWORK_ID:
+        assert chain_config.chain_id == RopstenChain.chain_id
```
carver: It looks like this whole if/elif clause could be replaced with a few lines if the expected data was also parametrized.

trinity/utils/chains.py
```diff
@@ -132,6 +132,11 @@ def construct_trinity_config_params(
     if args.trinity_root_dir is not None:
         yield 'trinity_root_dir', args.trinity_root_dir
 
+    if args.genesis is not None:
+        if args.data_dir is None:
+            raise ValueError("Must provide both genesis and data-dir")
```
carver: Maybe "When providing a custom genesis, must also provide a data-dir"?

veox:
I can reproduce the integration failure reliably, although I'm using Python 3.7 (not 3.6 as in current failure).

The reason seems to me as follows. The explanation may be bogus. If correct, though, then it's not strictly caused by changes in this PR.

-----

During a regular (non-test) run of

``` sh
trinity --light --tx-pool
````

the error message

```
   ERROR  11-11 17:15:46   BasePlugin#TxPlugin  Transaction pool not available in light mode
```

is printed after being caught as an exception; the fact that it's an exception allows `trinity` to back-track the start-up procedure, and shut down gracefully.

-----

During a test run of

``` sh
pytest "tests/trinity/integration/test_trinity_cli.py::test_txpool_deactivated[command0]"
```

the exception seems to be "suppressed" by `pytest` somehow, which is subsequently reflected in

```
ERROR  11-11 17:17:46             plugin.py  Task exception was never retrieved
```

The fact that the exception doesn't reach where it's supposed to prevents graceful shutdown.

veox:
Rebasing against `master` to resolve current conflict with `trinity/plugins/builtin/tx_pool/plugin.py` gives me the same result as before rebasing.

veox:
The weird thing is that running `command1` instead of `command0` with `pytest`, which amounts to

``` sh
trinity --light --ropsten --tx-pool
````

passes successfully.

On Ropsten, my `geth` `lightserv` always has free capacity, so I'm assuming there's no problem finding a peer; and the server is in `trinity`'s bootnode list...

On main-net, that is not the case...

However, I find it extremely weird that finding a peer is a prerequisite of determining that both `--light` and `--tx-pool` have been specified, and shutting down. X_X

**Comment by veox** [comment]
Found culprit. See https://github.com/pipermerriam/py-evm/pull/2

**Comment by pipermerriam** [comment]
I'm unlikely to be able to pick this back up for at least another 1-2 weeks.  Anyone else should feel free to take this over if they choose, otherwise, I'll pick it back up when I can.

File: tests/trinity/core/configuration/test_chain_config_object.py (hunk 4 above)
Comment (cburgdorf) [reply tests/trinity/core/configuration/test_chain_config_object.py]: The `network_id` does not exist on the `ChainConfig` and it's already doing `if network_id == MAINNET_NETWORK_ID`. Do I misread what you are suggesting?

File: trinity/utils/chains.py (hunk 6 above)
Comment (cburgdorf) [reply trinity/utils/chains.py]: Updated!

**Comment by cburgdorf** [comment]
I picked this up today to move it forward. Great work @pipermerriam !
I just applied some basic cleanup so far. 

More importantly though, I took a deep dive into the changes to understand them better. I think the most important thing to understand is what @pipermerriam pointed out already in a comment.

>I see the potential for this to break some things, specifically surrounding our assumptions about where processing will happen and thus which calls are safe to make within the networking process without risking blocking too long and timing out with all of our peers.

To make it clear, this PR makes quite a drastic change in which the `Chain` object that is passed into the `Server` does not continue to be a proxy anymore but instead is a local chain of that process. That means that **all** called APIs (importing blocks, executing transactions (regular sync)) will run in the same process as the rest of the networking code.

Previously, all this work happened outside of the `networking` process in the `database` process.

There are two possible scenarios now:

1. Turns out, doing this work in the `networking` process isn't a problem

2. Doing this work in the `networking` process is a problem because and we can not keep up with the networking traffic anymore eventually resulting in a loss of connections.

It's probably a good idea to dig deeper to gain clarity about this but my gut feeling says that it will most likely be a problem (after all, there's a reason why these things were separated in the first place)

Assuming that it **will be a problem**, here are some things to try out.

1. Try to achieve the previous architecture again that relied on proxies. This has become problematic because the `ChainConfig` is now dynamic and non-pickable. I believe it should be possible to hook into the serialization to simply teach it that the `ChainConfig` does not need to be serialized. After all, we do not care about the serialization of the  `ChainConfig` (as it can be reconstructed in every process) and only want to continue to have a `ProxyChain` to call methods on remotely. I think that should be doable.

2. Try to build a new isolated plugin that serves as process for block processing and gets it work items feed through the `EventBus`. This should be fairly straight forward (the `JSON-RPC Server` does something roughly related when in `light` mode)

3. Extract the entire syncing logic to become an isolated plugin. I believe this is a long term goal anyway but it I think it's not really a relaxed afternoon task as it basically means we would need to detangle, node, peer pool and syncing logic to all go into separate processes (isolated plugins) that use the `EventBus` as a first class communication pattern.

I'm intrigued to spend some time to look into the second option as it seems to be a nice stepping stone towards the third option (aka the *end game*) but less frightening :)

@carver @pipermerriam thoughts?

**Comment by cburgdorf** [comment]
So it seems `import_block` and maybe `validate_receipt` are the only potentialy costly methods that are called as part of syncing.

I started working on an isolated block importer and I think I like it.

https://github.com/ethereum/py-evm/pull/1299/commits/eb9e208ca66a1fa96be1d73a01cf5e9f078282f9

**Comment by pipermerriam** [comment]
@cburgdorf I don't have strong enough intuition about this to know what is best so use your best judgement.  Feel free to consider anything that steps us closer towards the goals outlined in #1485 if it seems to fit.

**Comment by cburgdorf** [comment]
@pipermerriam ok, I'm currently fiddling with the second approach where we move the expensive chain interaction of the syncing in a dedicated process provided by an isolated plugin. That said, the really costly stuff doesn't affect fast sync so much. So, if we move forward without that, we would most likely only see things becoming a problem when we go into regular sync (which appears to be broken currently anyway)

I will fiddle with that a bit longer to form a better opinion.

**Comment by cburgdorf** [comment]
@pipermerriam I moved that exploratory work out of this into https://github.com/ethereum/py-evm/pull/1497

I think we should take this as is because the performance issues will most likely not hit us too hard during fast-sync and we aren't quite at a point where regular sync is our main prio yet. So, I think we should move this forward so that hive integration / görli can advance forward.

I will still follow up on the performance issues as (hence #1479) but I don't think they should be blocking this from getting merged.

**Comment by cburgdorf** [comment]
(CI is green, rate limit issue as usual)

**Comment by cburgdorf** [comment]
@pipermerriam Apparently I can't ping you for a review since it is your PR. I approved it (you've written most of it after all) but I would wait on another :+1: before I merge.

**Comment by veox** [comment]
PR still missing Cute Animal Pic.

**Comment by cburgdorf** [comment]
@veox fixed!

File: tests/trinity/core/configuration/test_chain_config_object.py (hunk 4 above)
Comment (cburgdorf) [reply tests/trinity/core/configuration/test_chain_config_object.py]: But there is no `network_id` on the `ChainConfig`. The `network_id` exists on the `TrinityConfig`.

File: trinity/plugins/builtin/json_rpc/plugin.py [hunk 7]
```diff
@@ -48,15 +46,20 @@ def do_start(self) -> None:
...
+        chain: BaseAsyncChain
 
-        if self.context.trinity_config.sync_mode == SYNC_LIGHT:
+        if self.context.trinity_config.is_light_mode:
             header_db = db_manager.get_headerdb()  # type: ignore
             event_bus_light_peer_chain = EventBusLightPeerChain(self.context.event_bus)
-            chain = chain_class(header_db, peer_chain=event_bus_light_peer_chain)
-        else:
+            chain = chain_config.light_chain_class(header_db, peer_chain=event_bus_light_peer_chain)
+        elif trinity_config.is_full_mode:
```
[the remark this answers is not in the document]
cburgdorf: >If we're looking for more type-safety, sync_mode could be an `Enum` value instead of a `str`.

With just two options, I don't feel strongly here but in general I agree that an `Enum` is nicer. Will change that.

tests/trinity/core/configuration/test_chain_config_object.py
carver: Yup, you're right, I was thinking of the TrinityConfig. Did that used to be called ChainConfig? Anyway, nothing to do here AFAICT.

trinity/plugins/builtin/tx_pool/validators.py
```diff
@@ -30,16 +28,35 @@ class DefaultTransactionValidator():
...
         self._ordered_tx_classes = [
             vm_class.get_transaction_class()
-            for _, vm_class in chain.get_vm_configuration()
+            for _, vm_class in self.vm_configuration
         ]
-        self._initial_tx_class_index = self._ordered_tx_classes.index(self._initial_tx_class)
+
+        if initial_tx_validation_block_number is None:
+            self._initial_tx_class = self._get_tx_class_for_block_number(
+                initial_tx_validation_block_number
```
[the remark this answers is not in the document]
cburgdorf: I will add test for this. I think you are right with it being broken currently.

cburgdorf:
@carver Thanks for reviewing this!

>Obviously needs a rebase and/or squashing since there are dirty commits.

Yep, will just squash this all together into one commit that adds custom genesis support

>The only thing that might need more discussion is: what's happening in the transaction validator? At best, something is unclear, but it appears broken.

I *think* you are right. Will add tests for this (tomorrow)

trinity/plugins/builtin/json_rpc/plugin.py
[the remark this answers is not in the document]
pipermerriam: @carver I found myself needing to import the enum/constant anywhere that I wanted to check sync mode, thus arose the creation of the `is_x_mode`.  I don't feel strongly, but it was done to clean up the call sites.

eth/chains/mainnet/constants.py
```diff
@@ -24,16 +24,16 @@
...
 #
-TANGERINE_WHISTLE_MAINNET_BLOCK = 2463000
+TANGERINE_WHISTLE_MAINNET_BLOCK = BlockNumber(2463000)
 
 
 #
 # Spurious Dragon Block
 #
-SPURIOUS_DRAGON_MAINNET_BLOCK = 2675000
+SPURIOUS_DRAGON_MAINNET_BLOCK = BlockNumber(2675000)
```
pipermerriam: These could be extracted to make this PR smaller.

eth/chains/ropsten/constants.py
```diff
@@ -6,34 +9,34 @@
...
 #
-TANGERINE_WHISTLE_ROPSTEN_BLOCK = 0
+TANGERINE_WHISTLE_ROPSTEN_BLOCK = BlockNumber(0)
 
 
 #
 # Homestead Block
 #
-HOMESTEAD_ROPSTEN_BLOCK = 0
+HOMESTEAD_ROPSTEN_BLOCK = BlockNumber(0)
```
pipermerriam: Same, extractable into another small PR.

eth/tools/builder/chain/builders.py
```diff
@@ -159,7 +158,7 @@ def _is_homestead(vm_class: Type[BaseVM]) -> bool:
 
 
 @to_tuple
-def _set_vm_dao_support_false(vm_configuration: VMConfiguration) -> VMConfiguration:
+def _set_vm_dao_support_false(vm_configuration: VMConfiguration) -> Iterable[VMFork]:
```
pipermerriam: This also looks extractable into a separate PR.

tests/trinity/core/p2p-proto/test_v5_les_protocol_topics.py
```diff
@@ -0,0 +1,12 @@
+from eth.chains.ropsten import ROPSTEN_GENESIS_HEADER
+
+from p2p import discovery
+
+from trinity.protocol.les.proto import LESProtocol, LESProtocolV2
+
+
+def test_get_v5_topic():
```
pipermerriam: This looks like it could get extracted into a stand alone along with these changes:

https://github.com/ethereum/py-evm/pull/1299/files#diff-068a4555749b83cfb235f8ea0457fab2L373

tests/trinity/core/config/test_trinity_config_object.py
```diff
@@ -1,92 +0,0 @@
-from pathlib import Path
```
pipermerriam: Extracting this into a small PR that only moves this test file into the new location under `tests/trinity/core/configuration/test_trinity_config_object.py`  will help reduce the overall size of this diff.

trinity/nodes/base.py
```diff
@@ -42,16 +40,16 @@ class Node(BaseService):
...
     def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
         super().__init__()
+        self.trinity_config = trinity_config
         self._db_manager = create_db_manager(trinity_config.database_ipc_path)
         self._db_manager.connect()
         self._headerdb = self._db_manager.get_headerdb()  # type: ignore
 
         self._jsonrpc_ipc_path: Path = trinity_config.jsonrpc_ipc_path
         self._network_id = trinity_config.network_id
+        self._use_discv5 = trinity_config.use_discv5
```
pipermerriam: This could get extracted into it's own PR.

FrankSzendzielarz:
Ping me please when custom genesis / config is ready to play with. I am adding Trinity to Hive.

tests/trinity/core/configuration/test_chain_config_object.py
cburgdorf: They got split into `ChainConfig` and `TrinityConfig`

cburgdorf:
I didn't branch things out into multiple PRs but I sliced semantic commits that can be reviewed in isolation and should make it easier to follow.

I also added tests for the transaction validation stuff and corrected the functionality (it was indeed buggy).

I did not change the `is_full_mode` / `is_light_mode` stuff yet because the `SYNC_LIGHT` / `SYNC_FULL` constants are used at multiple places across the code base and if we refactor that to use an enum (I think we should) I would bloat this PR even more. I can send a follow up PR to refactor that.

@carver do you like to give this another review?

carver:
> I did not change the is_full_mode / is_light_mode stuff yet ...  would bloat this PR even more

Yup, totally reasonable.

It's Thanksgiving, so I'll be mostly offline. I'll give a optimistic :+1: since you mentioned that this merge is needed to let master run a regular sync.

cburgdorf:
@FrankSzendzielarz This is now merged. Let me know if you run into any issues trying it out :tada:

trinity/plugins/builtin/tx_pool/validators.py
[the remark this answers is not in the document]
zaman10d: Done. Setup[ Robot|  2021 coming up/ Countdown.

PR #1300, opened 2018-09-18
[conceptual] configured sub-protocols

carver:
### What was wrong?

Conversation from here was hard to continue without code: https://github.com/ethereum/py-evm/pull/1294#discussion_r218194387

### How was it fixed?

Made a few changes demonstrating the idea of pre-configuring subprotocols, as a way to get rid of the chain context. I think it also starts to send us in the nice direction of removing a bunch of sub-protocol specific APIs from peer, and putting them down in the `sub_proto` attribute where it makes a bit more sense.

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/proto.py
+++ b/trinity/protocol/common/proto.py
@@ -1,0 +1,53 @@
+
+
+class ChainProtocol(Protocol, ABC):
+
+    @classmethod
+    def configure_protocol(
+            cls,
+            headerdb: BaseAsyncHeaderDB,
+            network_id: int,
+            vm_configuration: Tuple[Tuple[int, Type[BaseVM]], ...]) -> Type[ChainProtocol]:
+        return type(
+            f'{cls.__name__}Network{network_id}',
+            (cls, ),
+            dict(
+                headerdb=headerdb,
+                network_id=network_id,
+                vm_configuration=vm_configuration,
+            ),
+        )
+
+    @abstractmethod
+    @property
+    def headerdb(self):
+        pass
+
+    @abstractmethod
+    @property
+    def network_id(self):
+        pass
+
+    @abstractmethod
+    @property
+    def vm_configuration(self):
+        pass
+
+    @property
+    async def genesis(self) -> BlockHeader:
+        genesis_hash = await self.wait(
+            self.headerdb.coro_get_canonical_block_hash(BlockNumber(GENESIS_BLOCK_NUMBER)))
+        return await self.wait(self.headerdb.coro_get_block_header_by_hash(genesis_hash))
+
+    @property
+    async def _local_chain_info(self) -> ChainInfo:
+        genesis = await self.genesis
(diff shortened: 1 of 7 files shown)


carver:
@pipermerriam obviously I haven't even started to actually make this work, I just thought it was easier to explore the idea in code than text. What do you think of this direction?

trinity/protocol/common/proto.py
```diff
@@ -0,0 +1,53 @@
...
+            ),
+        )
+
+    @abstractmethod
+    @property
+    def headerdb(self):
+        pass
+
+    @abstractmethod
+    @property
```
pipermerriam: noting that the order of these decorators is backwards,  `abstractmethod` should always be the innermost decorator.

trinity/protocol/common/proto.py
```diff
@@ -0,0 +1,53 @@
...
+    def network_id(self):
+        pass
+
+    @abstractmethod
+    @property
+    def vm_configuration(self):
+        pass
+
+    @property
+    async def genesis(self) -> BlockHeader:
```
pipermerriam: Just noting that I'd like to change this to a method as I find awaitable properties to be confusing and thus prone to errors.

trinity/protocol/common/proto.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ChainProtocol(Protocol, ABC):
+
+    @classmethod
+    def configure_protocol(
+            cls,
+            headerdb: BaseAsyncHeaderDB,
+            network_id: int,
+            vm_configuration: Tuple[Tuple[int, Type[BaseVM]], ...]) -> Type[ChainProtocol]:
+        return type(
```
pipermerriam: I wouldn't say it is definitive, but I'm closing in on this being a pattern we try to avoid.  However.... I did struggle a decent bit to avoid this in my previous refactor of the peer class and repeatedly almost went this route as it made things easier....

That said, any way we can avoid dynamic class creation is probably going to save us some type hinting headaches.

pipermerriam:
Big :+1: in this direction.  Do you want to run with it or should I? (no preference on my end)

pipermerriam:
Also, I don't mind pushing this out a release or two since it should probably be treated as lower priority than concurrent chain and state sync.

carver:
> Also, I don't mind pushing this out a release or two since it should probably be treated as lower priority than concurrent chain and state sync.

Yup, agreed on priority, and potentially pushing it into future releases.

> Big +1 in this direction. Do you want to run with it or should I? (no preference on my end)

Great! No preference either. I will be focusing on atomic write batches for the next couple days. I suspect this won't be the next most important thing after that, so it may sit in the queue for a bit. I'm fine if you end up picking it up before I do.

pipermerriam:

carver:
migrated

PR #1302, opened 2018-09-18
Make docker images smaller

jin10086:
- the images is base python:3.6
   so i think don't create new virtual environment

- use --no-cache-dir

The following is a comparison of space

```bash
➜  pyevm git:(master) ✗ docker images ethereum/trinity
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
ethereum/trinity    nocache             adfa8f214bfa        4 minutes ago       1.09GB
ethereum/trinity    latest              8798f7795bc6        9 minutes ago       1.32GB
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,26 +3,12 @@ FROM python:3.6
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-RUN apt-get update && apt-get install -y libssl-dev
-RUN apt-get update && apt-get install -y python3-pip
-RUN apt-get update && apt-get install -y pandoc
-
-#Copy the application py-evm to the /usr/src/app folder
 COPY . /usr/src/app
 
-# Remove existing virtualenv directory
-RUN rm -rf venv_docker_build
-# Install python dependencies
-RUN pip install virtualenv
-RUN cd /usr/src/app
-RUN virtualenv -p python3 venv_docker_build
-RUN  . venv_docker_build/bin/activate
-RUN pip3 install -e .[dev]
-RUN pip3 install -U trinity
+RUN pip install -e .[dev]  --no-cache-dir
+RUN pip install -U trinity --no-cache-dir
 
 RUN echo "Type \`trinity\` to boot or \`trinity --help\` for an overview of commands"
 
 EXPOSE 30303 30303/udp
 ENTRYPOINT ["trinity"]
-
-


Dockerfile
```diff
@@ -3,26 +3,12 @@ FROM python:3.6
...
-RUN apt-get update && apt-get install -y python3-pip
-RUN apt-get update && apt-get install -y pandoc
-
-#Copy the application py-evm to the /usr/src/app folder
 COPY . /usr/src/app
 
-# Remove existing virtualenv directory
-RUN rm -rf venv_docker_build
-# Install python dependencies
-RUN pip install virtualenv
```
cburgdorf: I'm far out of my comfort zone regarding judging of ditching `virtualenv` is appropriate or not. @carver @pipermerriam any thoughts on this? Can we assume that a virtual directory isn't needed when we are based on a fresh system (Debian I think that is)

Dockerfile
```diff
@@ -3,26 +3,12 @@ FROM python:3.6
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-RUN apt-get update && apt-get install -y libssl-dev
```
cburgdorf: I'm also not really sure where that was coming from but assumed it is there for a reason

Dockerfile
cburgdorf: Ok, judging by what web3 does, it really seems that a virtualenv can be ditched here:

https://github.com/ethereum/web3.py/blob/d436b0a799d92244ac8a419bf6da03f32a4feb4f/Dockerfile#L8

Dockerfile
jin10086: It is because this sentence is deleted and it will run normally. So I deleted it.

cburgdorf: I can confirm that it works and also the web3 interaction seems to do its job.

```
$ docker exec -i -t 8f5f6e2c80c5 trinity attach
    INFO  09-18 10:01:18        main  
      ______     _       _ __       
                           /____/   
    INFO  09-18 10:01:18        main  Trinity/0.1.0a15/linux/cpython3.6.4
    INFO  09-18 10:01:18        main  Trinity DEBUG log file is created at /root/.local/share/trinity/mainnet/logs/trinity.log
Trinity Console
---------------
An instance of Web3 connected to the running chain is available as the `w3` variable

In [1]: w3.eth.getBlock('latest')
Out[1]: 
AttributeDict({'difficulty': 17179869184,
 'extraData': HexBytes('0x11bbe8db4e347b4e8c937c1c8370e4b5ed33adb3db69cbdb7a38e1e50b1b82fa'),
 'gasLimit': 5000,
 'gasUsed': 0,
 'hash': HexBytes('0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3'),
 'logsBloom': HexBytes('0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'),
 'mixHash': HexBytes('0x0000000000000000000000000000000000000000000000000000000000000000'),
 'nonce': HexBytes('0x0000000000000042'),
 'number': 0,
 'parentHash': HexBytes('0x0000000000000000000000000000000000000000000000000000000000000000'),
 'receiptsRoot': HexBytes('0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'),
 'sha3Uncles': HexBytes('0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347'),
 'stateRoot': HexBytes('0xd7f8974fb5ac78d9ac099b9ad5018bedc2ce0a72dad1827a1709da30580f0544'),
 'timestamp': 0,
 'transactionsRoot': HexBytes('0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421'),
 'miner': '0x0000000000000000000000000000000000000000',
 'totalDifficulty': 17179869184,
 'uncles': [],
 'size': 540,
 'transactions': []})

In [2]: 
```

However, sincec the `Dockerfile` in web3 has the same dependency I'd like to wait for @carver to comment on that because we may be overlooking something important.

Dockerfile
pipermerriam: Yeah, I think the premise of docker allows us to treat the main system python as a sort-of venv.

Dockerfile
carver: I didn't remember why, so I did some git spelunking. Turns out it's been in the docs for a couple years, so maybe no one remembers why! :) https://github.com/ethereum/web3.py/commit/d0da4ac4e8e6

I think at this point, it's reasonable to remove it and re-address if something comes up again. It seems likely that it's just cruft for support of old linux distros.

cburgdorf:
Landed! Thank you for your contribution!

PR #1303, opened 2018-09-18
Implement EIP 1234

cburgdorf:
### What was wrong?

We need to implement [EIP1234](https://eips.ethereum.org/EIPS/eip-1234)

### How was it fixed?

Refactored Byzantium implementation to be more reusable for Constantinople. Left a comment inline for an alternative implementation idea.

#### Cute Animal Picture

### DIFF
--- a/eth/vm/forks/byzantium/__init__.py
+++ b/eth/vm/forks/byzantium/__init__.py
@@ -2,6 +2,10 @@ from typing import (  # noqa: F401
     Type,
 )
 
+from cytoolz import (
+    curry,
+)
+
 from eth_utils import (
     encode_hex,
     ValidationError,
@@ -44,6 +48,13 @@ def make_byzantium_receipt(base_header, transaction, computation, state):
     return frontier_receipt.copy(state_root=status_code)
 
 
+@curry
+def get_uncle_reward(block_reward, block_number, uncle):
+    block_number_delta = block_number - uncle.block_number
+    validate_lte(block_number_delta, MAX_UNCLE_DEPTH)
+    return (8 - block_number_delta) * block_reward // 8
+
+
 EIP658_STATUS_CODES = {
     EIP658_TRANSACTION_STATUS_CODE_SUCCESS,
     EIP658_TRANSACTION_STATUS_CODE_FAILURE,
@@ -63,6 +74,7 @@ class ByzantiumVM(SpuriousDragonVM):
     compute_difficulty = staticmethod(compute_byzantium_difficulty)
     configure_header = configure_byzantium_header
     make_receipt = staticmethod(make_byzantium_receipt)
+    get_uncle_reward = staticmethod(get_uncle_reward(EIP649_BLOCK_REWARD))
 
     @classmethod
     def validate_receipt(cls, receipt: Receipt) -> None:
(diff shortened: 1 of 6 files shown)


cburgdorf:
>Just noting that I think you'll need to also implement get_uncle_reward as well (but I'm not 100% sure about this)

You are right. Should be fixed now. I'll make an inline comment to get feedback on an alternative implementation idea.

eth/vm/forks/byzantium/__init__.py
```diff
@@ -80,9 +92,3 @@ def validate_receipt(cls, receipt: Receipt) -> None:
     @staticmethod
     def get_block_reward():
         return EIP649_BLOCK_REWARD
-
-    @staticmethod
-    def get_uncle_reward(block_number, uncle):
```
cburgdorf: Alternatively, we could make this a `@classmethod` and then access the block reward on `cls.block_reward` which would allow us drop any uncle reward refinement in `ConstantinopleVM`. Thoughts?

pipermerriam: Conceptually I'm fine with that but I think it goes against the direction we're trying to go with composition > inheritance.  I like the mechanism you've got in place now with the curried re-usable utilities so if you're happy with that I say keep it (and maybe backport it to the `ByzantiumVM`)

cburgdorf:
@pipermerriam I was leaving out tests because I thought we'll have plenty of shared `ethereum/tests` that will cover that. Will add some independent tests as well.

pipermerriam:
> I was leaving out tests because I thought we'll have plenty of shared ethereum/tests that will cover that.

I figured as much.  If it proves too big an issue to write them I could be convinced to leave them out but I think it's a good practice for us to start.  I know of a few things that I fully expected would be covered by `ethereum/tests` but aren't so getting better at writing our own (and subsequently being able to contribute them back upstream) would be a big win.

eth/vm/forks/byzantium/__init__.py
carver: Yup, I like it as-is, too. I kind of like the idea of pushing the `get_uncle_reward(...)` definition as far back as possible, so that even the import carries semantic meaning, like:

```py
from eth.vm.forks.frontier import get_uncle_reward
```

Would mean that the uncle mechanism has stayed the same since frontier, with only the block reward changing in: `get_uncle_reward = staticmethod(get_uncle_reward(EIP649_BLOCK_REWARD))`.

(I don't remember when/if the uncle mechanism changed, but the general concept stands.)

cburgdorf: SGTM :+1:

cburgdorf:
> I think it's a good practice for us to start. I know of a few things that I fully expected would be covered by ethereum/tests but aren't so getting better at writing our own (and subsequently being able to contribute them back upstream) would be a big win.

Totally agree to that. We can leave this PR open a bit longer if we aren't in a rush. I first want to finish up #1309 if possible.

eth/vm/forks/byzantium/__init__.py
cburgdorf: >Yup, I like it as-is, too. I kind of like the idea of pushing the get_uncle_reward(...) definition as far back as possible

It was last changed in `byzantium` so I think it is already pushed as far back as possible.

cburgdorf:
Interesting test cases! I'll create a good first starter issue for someone to grab to add expand the reward tests to cover these cases.

PR #1304, opened 2018-09-18
Switch to explicit atomic_batch() database handle

carver:
### What was wrong?

The old approach of batching all changes made while leveldb was locked was flawed in a multi-threading context: it permits unintended writes to be batched with the ones you want.

### How was it fixed?

Return an explicit handle when making an atomic batch, so that all the batched changes you want to make are explicitly part of that atomic batch. eg~

```py
atomic_db = AtomicDB()
with atomic_db.atomic_batch() as db:
    # changes are not immediately saved to the db, inside this context
    db[key] = val

    # changes are still locally visible even though they are not yet committed to the db
    assert db[key] == val

    # changes using the original db happen instantly, and are visible in the batch db
    atomic_db[key2] = val2
    assert db[key2] == val2
```

#### Cute Animal Picture


carver:
We should expect performance gains once we start using write batches, as well.

eth/db/backends/level.py
```diff
@@ -39,10 +53,73 @@ def __delitem__(self, key: bytes) -> None:
...
+            return True
+
+    def __delitem__(self, key: bytes) -> None:
+        if self._track_diff is None:
+            raise ValidationError("Cannot delete data from a write batch, out of context")
+
+        self._write_batch.delete(key)
+        del self._track_diff[key]
+
+    def shutdown(self) -> None:
```
pipermerriam: Maybe rename to `decomission`

tests/database/test_leveldb_db_backend.py
```diff
@@ -37,6 +46,136 @@ def test_set_and_get(memory_db, level_db):
     assert level_db.get(b'1') == memory_db.get(b'1')
 
 
+def test_atomic_batch(level_db, atomic_db):
```
pipermerriam: Rather than duplicating these here, do you think it makes sense to make this test suite generic or parametrized so that we can confirm feature parity between our in-memory databases and the leveldb ones?

eth/db/atomic.py
```diff
@@ -31,69 +30,98 @@ def __init__(self, wrapped_db: BaseDB = None) -> None:
...
         try:
             self[key]
         except KeyError:
             return False
         else:
             return True
+
+    @classmethod
+    @contextmanager
+    def commit_unexceptional(cls, write_target_db):
```
pipermerriam: This name is odd, but am I correct that it's an internal API.  If so, then I don't really care that much other than maybe prefixing with an `_` to make that clear.

PR #1305, opened 2018-09-19
New script to access the chain DB via an interactive python shell

gsalgado:
Also move a couple test scripts into the scripts/ dir

### DIFF
--- a/scripts/chainsync.py
+++ b/scripts/chainsync.py
@@ -1,3 +1,7 @@
+"""Run the chain syncer, storing data in the given db dir.
+
+Run with `python -m scripts.chainsync -db <path>`.
+"""
 import asyncio
 import logging
 from typing import (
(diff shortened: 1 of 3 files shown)


cburgdorf:
Looks useful. How about making this a plugin (pretty much compareable to `trinity attach`) or alternatively directly incorporate the functionality in the `trinity attach` plugin (my preference).

Benefits:

- more in line with the rest of the code base
- greater exposure (will be listed as part of `trinity --help` and people find it in the general list of plugins)

gsalgado:
I don't think we want to expose raw DB access to users in general. This is meant as a tool for devs

cburgdorf:
>I don't think we want to expose raw DB access to users in general. This is meant as a tool for devs

I think it is fair to assume that a big share of our targeted users are devs in general. We want developers to build solutions on top of Trinity so I think there's not much to lose and more to win by having this a a plugin or directly integrated in `trinity attach`.

gsalgado:
By devs I meant protocol devs or anybody who understands the risks of messing with the chain DB directly. I don't expect most devs building solutions _on top of trinity_ (i.e. trinity users) to have that knowledge, hence my arguing against exposing it

gsalgado:
If there's demand for such a thing from users, it can always be exposed later on

pipermerriam:
I'm somewhat in the same camp as @cburgdorf but I'm ok with these going in as-is if we can get at minimum a single smoke test in place which runs the `db-shell` script and does something like a sanity test that the printed output in the shell matches the expected chain state.

gsalgado:
So, this is just a tool I've used when debugging and I thought it could be useful to others so decided to check it in. Although ideally we'd want to prevent regressions in it, I'm not convinced writing an automated test for it is worth the effort, as it's not production code, is not required by trinity in any way, and if the smoke test is fragile we'll end up having CI failures that will just slow down development

Issue #1047, opened 2018-07-19
ChainSyncer should ignore data-request messages when peer is no longer connected


gsalgado:
### What is wrong?

All peer messages are put into a queue for later processing by `ChainSyncer`, and it is possible that by the time it reads a given message, the peer which sent it has already disconnected. When the msg contains data we requested, it makes sense for us to still process it normally, but when it is a data-request msg (e.g. `GetBlockHeaders`), we should simply ignore it.

### How can it be fixed

Currently we have separate methods to handle each msg type, so we'd need to modify them all to return early when the peer is already disconnected. A better approach may be to add a flag (e.g. `is_data_request`) to the `Command` class, and have our dispatcher method (`_handle_msg()`) use that to ignore the data requests from peers that have already disconnected

PR #1307, opened 2018-09-19
PeerRequestHandler: ignore requests dispatched after peer disconnected

gsalgado:
Closes: #1047

### DIFF
--- a/trinity/p2p/handlers.py
+++ b/trinity/p2p/handlers.py
@@ -33,6 +33,8 @@ class PeerRequestHandler(CancellableMixin):
         self.cancel_token = token
 
     async def handle_get_block_bodies(self, peer: ETHPeer, block_hashes: List[Hash32]) -> None:
+        if not peer.is_operational:
+            return
         self.logger.trace("%s requested bodies for %d blocks", peer, len(block_hashes))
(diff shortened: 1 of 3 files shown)


gsalgado:
Your interpretation is correct, but the `await` calls are for DB operations, which in general are orders of magnitude faster than network roundtrips, so it's unlikely that the peer's disconnection msg will be processed while we wait for those. The most common scenario that will cause the `PeerRequestHandler` to be called with a request from a peer that has disconnected is when there's a backlog in the syncer's msg queue, and by the time it gets to processing those data requests a peer may have been disconnected for a long time already.

So, IOW, this is indeed not addressing all possible cases, but to do that we'd have to introduce significant extra complexities and I believe it's not worth it as those missing cases should almost never happen.

FWIW, we can't simply use `peer.wait()` because then the operations wouldn't be cancelled when, say, the syncer's token is triggered -- we'd need to chain them, but if we do that we wouldn't know whether to catch and silence it or catch and re-raise it

pipermerriam:
I'm not opposed to this as a stop-gap solution, however, I think it hints at a larger architectural issue which deserves to be addressed at some point (for which I'll go make an issue to try and capture the problem).

:+1: :shipit:

Issue #1284, opened 2018-09-12
Handle shutdown of plugins that run in the network process more gracefully


cburgdorf:
### What is wrong?

Plugins that run in the network process don't really shut down cleanly yet:

Most info about this can be found here: https://github.com/ethereum/py-evm/pull/1216#discussion_r216047292

### How can it be fixed

:see_no_evil: :hear_no_evil:

PR #1309, opened 2018-09-20
Ensure shared process plugins shutdown cleanly

cburgdorf:
### What was wrong?

Currently, plugins that run in the shared network process do not shutdown cleanly in all scenarios (see #1284). To be more precise, they only shutdown if they happen to be chained with some `CancelToken` which is triggered. This means that for all current plugins that run in the networking process (`TxPool`, `LightPeerChainBridge`) there isn't a real problem as these will simply shutdown based the fact that e.g the `PeerPool` shuts down or the `LightPeerChain` shuts down. However, this is still leaving us with two problems:

1. Plugins that do not have a logical dependency that they can rely on to initiate cancellation are left out
2. Plugins can't stopped individually (think, future command `trinity stop-plugin <plugin-name>`

#### Understanding the sync vs async scenario

Before we dive into how exactly it was fixed, let's first take a moment to reflect on the three different kind of operational models that we currently have for plugins.

**1. Plugins that overtake the whole main process**

This category of plugin doesn't need any special tear down treatment as it is basically breaking out of the standard trinity boot sequence early and entirely redefines what the trinity process even means.

**2. Plugins that spawn there own process**

The teardown for these plugins is similar to how we shut down other non-plugin processes.

https://github.com/ethereum/py-evm/blob/d225f4a26811aa15f450378aa8d618caa573b158/trinity/extensibility/plugin.py#L180-L182 

Notice that `kill_process_gracefully` is a blocking operation.

**3. Plugins that run in a shared process (currently `networking`)**

Plugins that run in the shared networking process are required to be written in a non-blocking fashion using `asyncio` (This is a **MUST** for builtin plugins and at least a strong recommendation for any user defined plugin).

>**Sidenote:** That doesn't mean that plugins that spawn their own process can't be written in the same async fashion (in fact, most should) but since they run in an isolated space they can do pretty much everything they want including spinning up a virtual system to run a PHP server on Windows 95)

Means, since these plugins run asynchronously, they need to be stopped asynchronously as well.

#### Enter `shutdown` vs `shutdown_nowait`

Initially, I tried to streamline the shutdown of the `PluginManager` to either enforce being:

1. Always synchronous (turning non blocking operations into blocking ones)
2. Always asynchronous (turning blocking operations into non-blocking ones)

This has proven to be problematic for various reasons (I can elaborate if needed). So what I settled with instead is to come up with `shutdown` and `shutdown_nowait` (naming is inline with other places of our code base (and python stdlib) that use that pattern ).

Then inside the `kill_trinity_gracefully()` function, we continue to call the synchronous version (only isolated plugins are managed by this plugin manager instance) and in the unwind routine of the `networking` process we await the async version of shutdown (only async plugins run in this plugin manager instance.

### Refactorings

I refactored a bunch of stuff around plugins to make this possible:

1. Because some plugins need no `stop()` at all, some need `async def stop()` and some need `def stop()`, this method was removed from the `BasePlugin` and new derived plugins `BaseAsyncStopPlugin` as well as `BaseSyncStopPlugin` where defined accordingly. 

2. `PluginProcessScope` is no more. Otherwise it would be possible to e.g. derive from `BaseAsyncStopPlugin` but set `PluginProcessScope.MAIN` which is a contradicting configuration.

3. The `PluginManager` itself was refactored to rely more on composition. It now delegates to `scope.is_responsible_for_plugin(...)` and `scope.create_plugin_context(...)` instead of hardcoding if/else rules in methods of the `PluginManager` directly.

4. A new `AsyncContextManager` (backport from Python 3.7) was created to allow more flexible teardown of event loops compared to what we previously had with `exit_on_signal`. The `exit_on_signal` helper does still exist but it is now built on top of said `AsyncContextManager`-based API.

#### Cute Animal Picture

### DIFF
--- a/trinity/extensibility/plugin_manager.py
+++ b/trinity/extensibility/plugin_manager.py
@@ -1,14 +1,21 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
 from argparse import (
     ArgumentParser,
     Namespace,
     _SubParsersAction,
 )
+import asyncio
 import logging
 from typing import (
     Any,
+    Awaitable,
     Dict,
     Iterable,
     List,
+    Optional,
     Union,
 )
 
@@ -24,27 +31,96 @@ from trinity.extensibility.events import (
     BaseEvent,
     PluginStartedEvent,
 )
+from trinity.extensibility.exceptions import (
+    UnsuitableShutdownError,
+)
 from trinity.extensibility.plugin import (
+    BaseAsyncStopPlugin,
+    BaseIsolatedPlugin,
+    BaseMainProcessPlugin,
     BasePlugin,
+    BaseSyncStopPlugin,
     PluginContext,
-    PluginProcessScope,
 )
 
 
-class MainAndIsolatedProcessScope():
+class BaseManagerProcessScope(ABC):
+    """
+    Define the operational model under which a ``PluginManager`` runs.
+    """
+
+    endpoint: Endpoint
+
+    @abstractmethod
+    def is_responsible_for_plugin(self, plugin: BasePlugin) -> bool:
+        """
+        Define whether a ``PluginManager`` operating under this scope is responsible
+        for a given plugin or not.
+        """
+        raise NotImplementedError("Must be implemented by subclasses")
+
+    @abstractmethod
+    def create_plugin_context(self,
+                              plugin: BasePlugin,
+                              args: Namespace,
+                              chain_config: ChainConfig,
+                              boot_kwargs: Dict[str, Any]) -> PluginContext:
+        """
+        Create the ``PluginContext`` for a given plugin.
+        """
+        raise NotImplementedError("Must be implemented by subclasses")
+
+
+class MainAndIsolatedProcessScope(BaseManagerProcessScope):
 
     def __init__(self, event_bus: EventBus, main_proc_endpoint: Endpoint) -> None:
         self.event_bus = event_bus
         self.endpoint = main_proc_endpoint
 
+    def is_responsible_for_plugin(self, plugin: BasePlugin) -> bool:
+        return isinstance(plugin, BaseIsolatedPlugin) or isinstance(plugin, BaseMainProcessPlugin)
 
-class SharedProcessScope():
+    def create_plugin_context(self,
+                              plugin: BasePlugin,
+                              args: Namespace,
+                              chain_config: ChainConfig,
+                              boot_kwargs: Dict[str, Any]) -> PluginContext:
+
+        if isinstance(plugin, BaseIsolatedPlugin):
+            # Isolated plugins get an entirely new endpoint to be passed into that new process
+            context = PluginContext(
+                self.event_bus.create_endpoint(plugin.name)
+            )
+            context.args = args
+            context.chain_config = chain_config
+            context.boot_kwargs = boot_kwargs
+            return context
+
+        # A plugin that overtakes the main process never gets far enough to even get a context.
+        # For now it should be safe to just return `None`. Maybe reconsider in the future.
+        return None
+
+
+class SharedProcessScope(BaseManagerProcessScope):
 
     def __init__(self, shared_proc_endpoint: Endpoint) -> None:
         self.endpoint = shared_proc_endpoint
 
+    def is_responsible_for_plugin(self, plugin: BasePlugin) -> bool:
+        return isinstance(plugin, BaseAsyncStopPlugin)
 
-ManagerProcessScope = Union[SharedProcessScope, MainAndIsolatedProcessScope]
+    def create_plugin_context(self,
+                              plugin: BasePlugin,
+                              args: Namespace,
+                              chain_config: ChainConfig,
+                              boot_kwargs: Dict[str, Any]) -> PluginContext:
+
+        # Plugins that run in a shared process all share the endpoint of the plugin manager
+        context = PluginContext(self.endpoint)
+        context.args = args
+        context.chain_config = chain_config
+        context.boot_kwargs = boot_kwargs
+        return context
 
 
 class PluginManager:
@@ -57,10 +133,7 @@ class PluginManager:
         This API is very much in flux and is expected to change heavily.
     """
 
-    MAIN_AND_ISOLATED_SCOPES = {PluginProcessScope.MAIN, PluginProcessScope.ISOLATED}
-    MAIN_AND_SHARED_SCOPES = {PluginProcessScope.MAIN, PluginProcessScope.SHARED}
-
-    def __init__(self, scope: ManagerProcessScope) -> None:
+    def __init__(self, scope: BaseManagerProcessScope) -> None:
         self._scope = scope
         self._plugin_store: List[BasePlugin] = []
         self._started_plugins: List[BasePlugin] = []
@@ -103,7 +176,8 @@ class PluginManager:
         """
         for plugin in self._plugin_store:
 
-            if plugin is exclude or not self._is_responsible_for_plugin(plugin):
+            if plugin is exclude or not self._scope.is_responsible_for_plugin(plugin):
+                self._logger.debug("Skipping plugin %s (not responsible)", plugin.name)
                 continue
 
             plugin.handle_event(event)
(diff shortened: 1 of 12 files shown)


trinity/extensibility/plugin_manager.py
```diff
@@ -116,65 +187,64 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
 
-        return context
+        self._logger.info("Shutting down PluginManager with scope %s", type(self._scope))
+
+        for plugin in self._started_plugins:
+
+            if not isinstance(plugin, BaseAsyncStopPlugin):
+                continue
+
+            # TODO: This should run concurrently to not waste time
```
carver: A `await asyncio.gather()` seems pretty doable in this PR. If there's some significant complication I'm not seeing, an issue is preferable to a TODO.

trinity/extensibility/plugin_manager.py
```diff
@@ -116,65 +187,64 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
-            if not self._is_responsible_for_plugin(plugin):
+            if not self._scope.is_responsible_for_plugin(plugin):
                 continue
 
-            context = self._create_context_for_plugin(plugin, args, chain_config, boot_kwargs)
+            context = self._scope.create_plugin_context(plugin, args, chain_config, boot_kwargs)
             plugin.set_context(context)
 
-    def shutdown(self) -> None:
+    def shutdown_nowait(self) -> None:
```
carver: *minor naming nitpick*

I'm not sure the `_nowait` suffix is being used here in quite the same way as other `asyncio` library usages. This is potentially a long blocking wait for each plugin to shut down, right? Seems like it's more like a `shutdown_blocking` to me.

trinity/utils/shutdown.py
```diff
@@ -12,16 +18,23 @@
...
     loop = service_to_exit.get_event_loop()
+    async with wait_for_exit(loop):
+        await service_to_exit.cancel()
+        if endpoint is not None:
+            endpoint.stop()
+        service_to_exit._executor.shutdown(wait=True)
+
+
+@asynccontextmanager
+async def wait_for_exit(loop: asyncio.AbstractEventLoop = asyncio.get_event_loop()
```
carver: I think this default value is going to be problematic. It will be evaluated once at import time, which is probably not be desired. So far all method calls explicitly set the loop, anyway.

How about we require the loop to be passed for now. If leaving it out becomes convenient at some point in the future, we can re-add it. (The default lookup would go in the method body, to detect the loop at runtime).

trinity/utils/shutdown.py
```diff
@@ -12,16 +18,23 @@
 
 async def exit_on_signal(service_to_exit: BaseService, endpoint: Endpoint = None) -> None:
     loop = service_to_exit.get_event_loop()
+    async with wait_for_exit(loop):
+        await service_to_exit.cancel()
+        if endpoint is not None:
+            endpoint.stop()
+        service_to_exit._executor.shutdown(wait=True)
```
carver: Maybe add another one like:

```py
@asynccontextmanager
async def wait_for_service_exit(service_to_exit: BaseService) -> None:
    loop = service_to_exit.get_event_loop()
    async with wait_for_exit(loop):
        await service_to_exit.cancel()
        yield
        service_to_exit._executor.shutdown(wait=True)
```

`exit_on_signal` might go away or be pushed elsewhere, into `exit_service_with_plugin_endpoint(service, endpoint)` which just stops the endpoint, and `simple_service_exit(service)` which does nothing at all in `wait_for_service_exit`.

trinity/main.py
```diff
@@ -376,15 +379,18 @@ def display_launch_logs(chain_config: ChainConfig) -> None:
...
-    loop.run_forever()
-    loop.close()
+async def handle_networking_exit(service: BaseService,
+                                 plugin_manager: PluginManager,
+                                 endpoint: Endpoint) -> None:
+
+    async with wait_for_exit(service.get_event_loop()):
+        await service.cancel()
+        await plugin_manager.shutdown()
+        endpoint.stop()
```
carver: With a new `wait_for_service_exit` this could be:

```py
    async with wait_for_service_exit(service):
        await plugin_manager.shutdown()
        endpoint.stop()
```

trinity/extensibility/plugin_manager.py
cburgdorf: Yep, that was just me slacking :palm_tree: :sunny:

trinity/extensibility/plugin_manager.py
cburgdorf: Ah, *that* seems like the right name for it :+1:

trinity/utils/shutdown.py
cburgdorf: > It will be evaluated once at import time

Oh, I assumed that would only be evaluated in the case it is called without that parameter. Didn't know these things are evaluated eagerly in Python (Still a Python noob after all).

>How about we require the loop to be passed for now

Yes, let's do that.

trinity/utils/shutdown.py
cburgdorf: I went down that way but slightly adjusted naming scheme:

```python
async def exit_with_service_and_endpoint(service_to_exit: BaseService, endpoint: Endpoint) -> None:
    async with exit_signal_with_service(service_to_exit):
        endpoint.stop()


async def exit_with_service(service_to_exit: BaseService) -> None:
    async with exit_signal_with_service(service_to_exit):
        pass


@asynccontextmanager
async def exit_signal_with_service(service_to_exit: BaseService) -> AsyncGenerator[None, None]:
    loop = service_to_exit.get_event_loop()
    async with exit_signal(loop):
        await service_to_exit.cancel()
        yield
        service_to_exit._executor.shutdown(wait=True)


@asynccontextmanager
async def exit_signal(loop: asyncio.AbstractEventLoop) -> AsyncGenerator[None, None]:
    sigint_received = asyncio.Event()
    for sig in [signal.SIGINT, signal.SIGTERM]:
        # TODO also support Windows
        loop.add_signal_handler(sig, sigint_received.set)

    await sigint_received.wait()
    try:
        yield
    finally:
        loop.stop()
```

I think this reads more natural:

```python
async with exit_signal(loop):
    # do something
```

Same with:

```python

asyncio.ensure_future(exit_with_service_and_endpoint(service, endpoint))
```

trinity/extensibility/plugin_manager.py
```diff
@@ -116,65 +190,72 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+
+        for plugin, result in zip(self._started_plugins, stop_results):
+            if isinstance(result, Exception):
+                self._logger.error(
+                    'Exception thrown while stopping plugin %s: %s', plugin.name, result
+                )
+            else:
+                self._logger.info("Successfully stopped plugin: %s", plugin.name)
+
+    def _stop_async_plugins(self) -> Iterable[Awaitable[Optional[Exception]]]:
```
cburgdorf: I pushed this out into a private method rather than using an inline comprehension for two reasons:

1. I wanted to do `self._logger.info("Stopping plugin: %s", plugin.name)` to have the same logs as with sync stops
2. We need an `isinstance` check to avoid a cast

trinity/extensibility/plugin_manager.py
```diff
@@ -116,65 +190,72 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
 
-        context.args = args
-        context.chain_config = chain_config
-        context.boot_kwargs = boot_kwargs
+        self._logger.info("Shutting down PluginManager with scope %s", type(self._scope))
 
-        return context
+        stop_results = await asyncio.gather(*self._stop_async_plugins(), return_exceptions=True)
+
+        for plugin, result in zip(self._started_plugins, stop_results):
```
carver: There isn't a result for sync plugins in `self._started_plugins`, so this will emit the wrong logs below, when a sync plugin comes before an async one.

Maybe add a private method that returns all async plugins, so you can do:

```py
async_plugins = self._get_async_plugins()
stop_results = await asyncio.gather(*self._stop_plugins(async_plugins), return_exceptions=True)
for plugin, result in zip(async_plugins, stop_results):
  ...
```

cburgdorf: Oh, indeed! 🙉👍

trinity/extensibility/plugin_manager.py
```diff
@@ -116,65 +190,77 @@ def broadcast(self, event: BaseEvent, exclude: BasePlugin = None) -> None:
...
+        Asynchronously shut down all started plugins.
+        """
 
-        return context
+        if isinstance(self._scope, MainAndIsolatedProcessScope):
+            raise UnsuitableShutdownError("Use `shutdown_blocking` for instances of this scope")
+
+        self._logger.info("Shutting down PluginManager with scope %s", type(self._scope))
+
+        async_plugins = [
```
cburgdorf: @carver doing this inline here to not clutter the class with methods that are only used by this method too much. For a moment I thought we could reuse `_stop_plugins()` for both async / sync but since the sync `stop()` needs to be surrounded with a try/catch and is considered a success as soon as `stop()` returns, that isn't really reusable for the async version. I guess its cleaner to handle these separately.

PR #1310, opened 2018-09-20
Fix a never-ending subtask in BasePeerPool

gsalgado:
The handle_peer_count_requests() coroutine runs as a background task but
did not exit when the pool's cancel token was triggered.

Also introduce a custom log formatter for trinity that uses the last
element of the logger's name instead of the module, otherwise it's not
clear where they're coming from (specially for those generated by
BaseService subclasses).

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -768,12 +768,17 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             self.run_task(self.handle_peer_count_requests())
 
     async def handle_peer_count_requests(self) -> None:
-        async for req in self.event_bus.stream(PeerCountRequest):
-            # We are listening for all `PeerCountRequest` events but we ensure to
-            # only send a `PeerCountResponse` to the callsite that made the request.
-            # We do that by retrieving a `BroadcastConfig` from the request via the
-            # `event.broadcast_config()` API.
-            self.event_bus.broadcast(PeerCountResponse(len(self)), req.broadcast_config())
+        async def f() -> None:
+            # FIXME: There must be a way to cancel event_bus.stream() when our token is triggered,
+            # but for the time being we just wrap everything in self.wait().
+            async for req in self.event_bus.stream(PeerCountRequest):
+                # We are listening for all `PeerCountRequest` events but we ensure to only send a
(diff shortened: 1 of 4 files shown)


p2p/peer.py
```diff
@@ -768,12 +768,17 @@ def __init__(self,
...
-        async for req in self.event_bus.stream(PeerCountRequest):
-            # We are listening for all `PeerCountRequest` events but we ensure to
-            # only send a `PeerCountResponse` to the callsite that made the request.
-            # We do that by retrieving a `BroadcastConfig` from the request via the
-            # `event.broadcast_config()` API.
-            self.event_bus.broadcast(PeerCountResponse(len(self)), req.broadcast_config())
+        async def f() -> None:
+            # FIXME: There must be a way to cancel event_bus.stream() when our token is triggered,
+            # but for the time being we just wrap everything in self.wait().
+            async for req in self.event_bus.stream(PeerCountRequest):
```
gsalgado: @cburgdorf is there a way to do this already?

cburgdorf: Oh..so one thing you can do with the `stream` API is to break out e.g.

```python
    if not self.is_operational:
        break
```

This will then be taken care of to handle the unsubscribing.

https://github.com/ethereum/lahja/blob/4c2bb9c6f038f81566f7852ef9abd0d3928d0ca5/lahja/endpoint.py#L191-L192

```python
except GeneratorExit:
    self._queues[event_type].remove(queue)
```

However, as I'm writing this, I realize this is probably not what you want because it only allows you to break out at the moment where another event comes through.

I think what we really want is to be able to pass a `CancelToken` as in.

```python
async for req in self.event_bus.stream(PeerCountRequest, cancel_token=token):
    ...
```

Does that make sense?

gsalgado: Yep, passing a CancelToken is what I had in mind. Also, as you pointed out, doing the is_operational check and breaking out of the loop is not going to fix it

cburgdorf: Alright, there are a bunch of improvements that I'd like to get in for a next release soon. If we can live with the stopgap approach of this PR for a few days I'll come up with a new release soon that will have this API change.

I created an issue for this particular problem. https://github.com/ethereum/lahja/issues/9 and plan

Issue #930, opened 2018-06-19
Beacon chain loves Trinity


hwwhww:
Recently, research team is working on refining [the new spec](https://notes.ethereum.org/s/Syj3QZSxm) and the [**beacon chain** PoC codebase](https://github.com/ethereum/beacon_chain)  (**big shout out to V!**). Basically, there’s a beacon chain that will produce random numbers with RANDAO-based scheme, and we can build Casper logic in the beacon chain to support both beacon chain and shard chains staking.

This issue is for discussing (1) what feature/components are required and (2) what's the blueprint looks like *if we want to implement the beacon chain in the Trinity pyeth world*?

cc @djrtwo , please correct me if I'm wrong. :)

## The certain differences between the main PoW chain and beacon chain
* They both have *Block*, but different fields.
* They both have *State*, but totally different content.
* The main chain uses EVM; beacon chain doesn’t.
* The main chain is still the PoW chain; beacon chain is the full PoS chain (and can implicitly finalize the main PoW chain).

## The potential differences between the main PoW chain and beacon chain
* Beacon chain may use a different serialization scheme instead of RLP.
* Beacon chain may use a different p2p layer for integrating with shard chain networks.
* The main chain has a one-way burns contract; the Casper logic is implemented in beacon chain protocol.
* Main PoW chain, beacon chain, and shards relationship:
     * **Main PoW chain <> beacon**: To add and verify the `main_chain_ref` in beacon block, **the beacon chain validators have to follow the actual canonical main PoW chain**. (see [Beacon-chain-processing](https://notes.ethereum.org/s/Syj3QZSxm#Beacon-chain-processing for more details)
     * **beacon <> shards**: One active validators queue for the validators serve different roles; the validator could be sampled as beacon/shards proposer, notary, attester... etc., so they will have to be both beacon client and shards client.

## Discussion
1. Given those differences, it probably makes sense to have a `beacon` module besides `evm` to put the chain logic?
2. If (1), since there’s no EVM in beacon chain, it will be much simpler here. How many components do we need?
	1. Start by mirroring the reusable `evm` architecture.
        ```
        ├── chains
        ├── constants.py
        ├── db
        ├── exceptions.py
        ├── tools
        ├── utils
        └── validation.py
        ```
	2. Move and reorganize the beacon logics to here.
3. Is there anything we need to be aware of?

pipermerriam:
This all looks generally correct.  

[...]

PR #1311, opened 2018-09-20
[Draft] Add `GRPCServer` and `GRPCClient` with its deps

mhchia:
Note: this PR is only a draft, to demonstrate how to communicate with the p2p layer written in Go. Apology if the structure or code is inappropriate.

### What was wrong?
For later sharding design, we need a P2P layer for sharding clients. Currently, we use libp2p to build this layer, which is written in Go. For the Python client, we need a way to control and communicate with the Go code. 

### How was it fixed?
Several approaches can be used to achieve the communication. Options that are in our mind were posted in https://github.com/ethereum/py-evm/issues/930#issuecomment-420013619. Finally, we decided to do it with gRPC for now, since it is easier to debug/test.

The structure in this PR is similar to the gRPC structure in the link, but we move more logics from Go to Python, to allow Python side to define the types of messages without modifying Go code.

The design details of this draft can found [here](https://notes.ethereum.org/mHOKQ0o_Svymt3_LZNjtTA?both). This PR only includeds the `GRPCServer` and `GRPCClient` part. The rough structure currently looks like:

current structure

We define message types on Python side, serialize messages to bytes, and broadcast them. On receiving anything from the P2P layer on Python side, we just deserialize them back to the message types according to the paramter `msgType`.

The main logics are in `grpc_server.py` with `GRPCServer` in Python, called by Go. And `grpc_client.py` with `GRPCClient` for Python to call the RPC server in Go. This PR works well with the Go code in PR https://github.com/ethresearch/sharding-p2p-poc/pull/65. To run the test `test_grpc.py`, we need to spin up a python grpc server for each P2P node, through `python grpc_server.py`, with correct configs.

Really appreciate if there is any feedback on this draft of design:)

#### Cute Animal Picture


pipermerriam:
@carver flagged you to go ahead and take a look at this.  In case it's foreign to you, just ignore all of the generated files.  That's how protobuf works.

mhchia:
@pipermerriam Big thanks for the review!

> I think we'll need to spin up some robust tooling around spinning up the server, as well as good documentation on how to do the base installation of all of the dependencies.

Will do. For me, the simplest tooling will be having a Makefile or shell script to do it. However, I'm considering the options to build/install the go code:
1. Assuming users have installed `go`, clone the repo, and compile the code.
2. Assuming users have installed docker, and let them build the docker image with our [dockerfile](https://github.com/ethresearch/sharding-p2p-poc/blob/master/docker/dev.Dockerfile) and run the container.
Personally, I prefer the 2nd option since it doesn't need `go` to be installed. But I don't have much experience in cross-language setups, do you have any thoughts on which way to go?

carver:
> 1.Assuming users have installed go, clone the repo, and compile the code.
> 2. Assuming users have installed docker, and let them build the docker image with our dockerfile and run the container.

We're talking about end users, right? We can't assume they have `go` or `docker`. It sounds like we would have to install one of those as part of the trinity install. (part of the cost of a non-python implementation)

carver:
Well, I guess if the idea is that this implementation would be replaced by a pure python one before leaving alpha, then I guess # 2 is ok (documenting that docker is required, and how to get it).

pipermerriam:
On the Eth2.0 implementers call, one of the protocol labs guys gave an overview of the libp2p daemon which sounds promising.  Designed to be a daemon process on your machine that is designed to be connected to over unix sockets, shared memory, etc, and is meant to reduce the need for every language to have an implementation.  Should be well funded and documented, though it would potentially fall on us to write the python library that connects to it if we want to start using it in a "soon" timeline.

I'm still highly preferential to a native libp2p implementation.  There is a team from some University that has actually started working on exactly that as their final project and is likely to receive a grant for it so I think it's likely that we gain access to both a native implementation and the daemon approach in the near term future.

libp2p_bridge/grpc_server.py
```diff
@@ -0,0 +1,84 @@
...
+class GRPCServer(event_pb2_grpc.EventServicer):
+
+    def Receive(self, request, context):
+        response = make_response(True)  # Request succeeded
+        ret_bytes = dispatch(request.msgType, request.data)
+        receive_response = event_pb2.ReceiveResponse(
+            response=response,
+            data=ret_bytes,
+        )
+        print("Receive: request={}, response={}".format(request, receive_response))
```
carver: A logger is better than a `print()` statement here, for easy configurability of displayed info. Something like:

```py
class GRPCServer(event_pb2_grpc.EventServicer):
    logger = logging.getLogger('libp2p_bridge.grpc_server')

        ...
        self.logger.info("Receive: request=%s, response=%s", request, receive_response)
```

(switched to let the logger do the formatting, for eventual use by logging aggregators, etc)

libp2p_bridge/grpc_server.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    listen_addr = '{}:{}'.format(RPC_SERVER_LISTEN_IP, RPC_SERVER_PORT)
+    server.add_insecure_port(listen_addr)
+    server.start()
+    print("Server started")
+    try:
+        while True:
+            time.sleep(86400)
+    except KeyboardInterrupt:
+        server.stop(0)
```
carver: Maybe add a global `except Exception` that logs the uncaught exception with `logger.exception(...)`, and a finally block that runs `server.stop(0)`.

libp2p_bridge/grpc_server.py
```diff
@@ -0,0 +1,84 @@
...
+        GRPCServer(),
+        server,
+    )
+    listen_addr = '{}:{}'.format(RPC_SERVER_LISTEN_IP, RPC_SERVER_PORT)
+    server.add_insecure_port(listen_addr)
+    server.start()
+    print("Server started")
+    try:
+        while True:
+            time.sleep(86400)
```
carver: If this method sticks around for a while, it should probably turn into a `BaseService`. This we wouldn't need a `while True` that can only be ended by exception. Instead, we could just cancel the service.

libp2p_bridge/message.py
```diff
@@ -0,0 +1,89 @@
...
+            raise ValueError("bytes too short for a collation: len={}".format(len(b)))
+        return CollationRequest(
+            bytes_to_int(b[:INT_BYTES]),
+            bytes_to_int(b[INT_BYTES:INT_BYTES * 2]),
+            b[INT_BYTES * 2:].decode(),
+        )
+
+
+# FIXME: use plain class instead of Enum now
+class MsgType:
```
carver: I don't follow the comment. Why not an Enum?

libp2p_bridge/grpc_server.py
```diff
@@ -0,0 +1,84 @@
...
+            response=response,
+            data=ret_bytes,
+        )
+        print("Receive: request={}, response={}".format(request, receive_response))
+        return receive_response
+
+
+def run_grpc_server():
+    # TODO: should confirm how many workers to use
+    server = grpc.server(futures.ThreadPoolExecutor(max_workers=10))
```
carver: Am I reading this right that every "in-flight" request (a request sent, with no response received) will hang one thread? So this would limit to 10 in-flight requests.  I'm not really sure what's reasonable, but that sounds a little low. It's worth thinking about what happens if a few of your clients go offline or just time out (hanging a thread for a while).

libp2p_bridge/message.py
mhchia: Sorry, it should be changed to `Enum` indeed. Will fix it

libp2p_bridge/grpc_server.py
mhchia: Thanks a lot for pointing it out. Will change it.

libp2p_bridge/grpc_server.py
mhchia: > Am I reading this right that every "in-flight" request (a request sent, with no response received) will hang one thread? So this would limit to 10 in-flight requests. I'm not really sure what's reasonable, but that sounds a little low.

Yes, it is. I set this number just as a placeholder since I wasn't sure of how many threads to use. Should I just set it as None(or don't set this parameter), let it be `5 * num_of_processor`? (ref: [doc](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor)). This number seems more reasonable to me. However, might still be a relatively small number.

libp2p_bridge/grpc_server.py
mhchia: Good point! Will change it. Thank you.

libp2p_bridge/grpc_server.py
mhchia: Thanks for the pointer. I will change it according to the [guideline](https://github.com/ethereum/py-evm/blob/fcf73d071278fab5f85c221f89cb95fcbeb942a3/p2p/DEVELOPMENT.md)!

mhchia:
@carver Really appreciate your review. Does that make sense I just move `libp2p_bridge` into `p2p`? Will add a README afterward.

@pipermerriam Yes, I think the libp2p daemon is promising as well. It is under active and great development, and they are willing to support our requirements. My concern on this was, it is still currently under development. IMO we can simultaneously build the Python bindings(the logics in Python to connect the daemon) and help to use and develop the components we need, and keep working with the existing gRPC solution, until the native Python solution is usable.

mhchia:
#### TODO
Reflect on review
- [ ] Use `BaseService` in `GRPCServer`
- [ ] Add README for tutorial

p2p/libp2p_bridge/test_grpc.py
```diff
@@ -23,32 +23,16 @@
...
-    """
-    stub = make_event_stub()
-    cr = CollationRequest(1, 2, "")
-    req = event_pb2.ReceiveRequest(
-        peerID="",
-        topic="",
-        msgType=MsgType.CollationRequest,
-        data=cr.to_bytes(),
-    )
-    stub.Receive(req)
```
carver: Why did this test get removed?

mhchia: Ah actually this function is not called previously, so I removed it. But you're right this test currently lacks of the tests for `GRPCServer` using `make_event_stub`. Will Add more tests for this.

mhchia:
@carver Sorry for bothering again. I changed `GRPCServer` to inherit `BaseService`(by overriding the _run method: https://github.com/ethereum/py-evm/blob/5b5d580b75a680e14fc3002adb0ac80d4d6dcfd8/p2p/libp2p_bridge/grpc_server.py#L78-L97), and added a [`test_grpc_server.py`](https://github.com/ethereum/py-evm/blob/5b5d580b75a680e14fc3002adb0ac80d4d6dcfd8/p2p/libp2p_bridge/test_grpc_server.py#L35-L72) to specifically test the `GRPCServer` in asyncio. However, I kept getting errors like the following:
```
===================================================================================== FAILURES ======================================================================================
____________________________________________________________________ test_grpc_server_receive_collation_request _____________________________________________________________________

event_loop = <_UnixSelectorEventLoop running=False closed=False debug=False>

    @pytest.mark.asyncio
    async def test_grpc_server_receive_collation_request(event_loop):
        token = CancelToken("grpc_server", loop=event_loop)
        grpc_server = GRPCServer(token=token, loop=event_loop)
        asyncio.ensure_future(grpc_server.run())
        await asyncio.sleep(0.1)
        cr = CollationRequest(1, 2, "")
        req = event_pb2.ReceiveRequest(
            peerID="",
            msgType=MsgType.CollationRequest.value,
            data=cr.to_bytes(),
        )
        stub = make_stub()
>       resp = stub.Receive(req)

cr         = <p2p.libp2p_bridge.message.CollationRequest object at 0x10bd59fd0>
event_loop = <_UnixSelectorEventLoop running=False closed=False debug=False>
grpc_server = <p2p.libp2p_bridge.grpc_server.GRPCServer object at 0x10bcce4e0>
req        = msgType: 3
data: "\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000...0\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\000\002"

stub       = <p2p.libp2p_bridge.github.com.ethresearch.sharding_p2p_poc.pb.event.event_pb2_grpc.EventStub object at 0x10bcf4c50>
token      = <CancelToken: grpc_server>

test_grpc_server.py:67: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
/Users/mhchia/.pyenv/versions/py3.6.0eth/lib/python3.6/site-packages/grpc/_channel.py:514: in __call__
    return _end_unary_response_blocking(state, call, False, None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

state = <grpc._channel._RPCState object at 0x10bd08550>, call = <grpc._cython.cygrpc.SegregatedCall object at 0x10bd6d688>, with_call = False, deadline = None

    def _end_unary_response_blocking(state, call, with_call, deadline):
        if state.code is grpc.StatusCode.OK:
            if with_call:
                rendezvous = _Rendezvous(state, call, None, deadline)
                return state.response, rendezvous
            else:
                return state.response
        else:
>           raise _Rendezvous(state, None, None, deadline)
E           grpc._channel._Rendezvous: <_Rendezvous of RPC that terminated with:
E           	status = StatusCode.UNAVAILABLE
E           	details = "Socket closed"
E           	debug_error_string = "{"created":"@1538493962.016122000","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1095,"grpc_message":"Socket closed","grpc_status":14}"
E           >

call       = <grpc._cython.cygrpc.SegregatedCall object at 0x10bd6d688>
deadline   = None
state      = <grpc._channel._RPCState object at 0x10bd08550>
with_call  = False

/Users/mhchia/.pyenv/versions/py3.6.0eth/lib/python3.6/site-packages/grpc/_channel.py:448: _Rendezvous
-------------------------------------------------------------------------------- Captured log setup ---------------------------------------------------------------------------------
selector_events.py          65 DEBUG    Using selector: KqueueSelector
--------------------------------------------------------------------------------- Captured log call ---------------------------------------------------------------------------------
grpc_server.py              94 INFO     grpc_server started
======================================================================== 1 failed, 1 passed in 1.05 seconds =========================================================================
```
I don't quite understand the reason of this error, and am guessing it might be relevant to the fact that the server and the client are run in the same loop. Besides, I'm pretty unsure whether it is correct to use [`futures.ThreadPoolExecutor()`](https://github.com/ethereum/py-evm/blob/5b5d580b75a680e14fc3002adb0ac80d4d6dcfd8/p2p/libp2p_bridge/grpc_server.py#L86) in the coroutine.
Do you have any idea on this issue? I will keep investigating the cause. I will also really appreciate if there is any pointer.

carver:
> I don't quite understand the reason of this error, and am guessing it might be relevant to the fact that the server and the client are run in the same loop. Besides, I'm pretty unsure whether it is correct to use futures.ThreadPoolExecutor() in the coroutine. Do you have any idea on this issue?

Yeah, things get messy when trying to mix sync and async things. It's not obvious to me either, at a quick look.

If converting to a `BaseService` was going to be a straightforward change, then it would be worth it, but not if it blocks progress. So I'm happy postponing that change. Especially since the underlying protobuf and server aren't `async` methods, there's not as much a benefit.

mhchia:
> If converting to a BaseService was going to be a straightforward change, then it would be worth it, but not if it blocks progress. So I'm happy postponing that change. Especially since the underlying protobuf and server aren't async methods, there's not as much a benefit.

Really thanks a lot for the direction. In this case, I will change it back to the sync version for simplicity.

carver:
Cool, I think this is okay to go in if you're ready for a merge @mhchia

mhchia:
Hi @carver , I added a first version of [README](https://github.com/ethereum/py-evm/blob/c502e5dd2926471ca8cd73be7b7c18950990dd56/p2p/libp2p_bridge/docs/README.md), but the content inside is subject to changes, and also needs https://github.com/ethresearch/sharding-p2p-poc/pull/65 to be merged to be useful. May I merge this PR first, and possibly make changes later to this README? Thanks a lot!

carver:
Yeah, totally understandable that it's still in flux. We can merge now and iterate.

PR #1312, opened 2018-09-20
Turn on warnings by default, if in prerelease

carver:
### What was wrong?

`trinity` is spouting a bunch of warnings, and no one noticed. Presumably because none of us explicitly turned warnings on.

### How was it fixed?

Check if we are in a pre-release version (alpha or beta). If so, turn on warnings. This will help us catch warning things sooner.

I'm not totally convinced it's a good idea, but it did help me do some cleanups. Also, it seems to highlight a bunch of unclean connection handling.

#### Cute Animal Picture

### DIFF
--- a/trinity/utils/version.py
+++ b/trinity/utils/version.py
@@ -1,5 +1,7 @@
 import sys
 
+import pkg_resources
+
 from trinity import __version__
 
 
@@ -16,3 +18,12 @@ def construct_trinity_client_identifier() -> str:
         # mypy Doesn't recognize the `sys` module as having an `implementation` attribute.
         imp=sys.implementation,  # type: ignore
     )
+
+
+def is_prerelease() -> bool:
+    try:
+        distro = pkg_resources.get_distribution("trinity")
+        # mypy thinks that parsed_version is a tuple. Ignored...
+        return distro.parsed_version.is_prerelease  # type: ignore
+    except pkg_resources.DistributionNotFound:
+        return True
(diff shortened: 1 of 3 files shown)


trinity/utils/version.py
```diff
@@ -16,3 +18,12 @@ def construct_trinity_client_identifier() -> str:
         # mypy Doesn't recognize the `sys` module as having an `implementation` attribute.
         imp=sys.implementation,  # type: ignore
     )
+
+
+def is_prerelease() -> bool:
+    try:
+        distro = pkg_resources.get_distribution("trinity")
+        # mypy thinks that parsed_version is a tuple. Ignored...
+        return distro.parsed_version.is_prerelease  # type: ignore
```
cburgdorf: Am I right that all Trinity versions that we currently put out are pre-releases (as in, have `alpha` suffix)? Doesn't that mean we slow down sync for everyone trying out Trinity at this stage?

cburgdorf: Ah wait, maybe I misread. This does not turn on `PYTHONASYNCIODEBUG` by default?

carver: > This does not turn on PYTHONASYNCIODEBUG by default?

Right, my first draft turned it on at first, but it was way too slow, maybe >10x slower syncing. It was annoying even in development, and made it hard to tell which things might improve performance.

carver:
> Seems like something that should be part of the CLI unless I'm misunderstanding the context of this issue.

The `asyncio` logger can already be modified with the current logging arguments I believe. The `PYTHONWARNINGS` change makes sense to just leave as controlled by an environment variable IMO.

I don't think this should be merged until after the pending release, though, to give us time to clean up new warnings that we see.

PR #1323, opened 2018-09-26
Added beacon chain utils

### LINKED ISSUE #1322: beacon chain utils
### What is wrong?
For #930, the first thing we can do is moving some basic utils to Py-EVM.

### How can it be fixed
In beacon_chain repo [beacon_chain/utils](https://github.com/ethereum/beacon_chain/tree/master/beacon_chain/utils), all pure functions:

1. `bitfields.py` - the APIs for operating bitfield
2. `bls.py` - the APIs of BLS signature (BN128 curve currently, will change to BLS12-381 curve in other PRs).
3. `blake.py` - a wrapper of blake2b. The reason why we use `blake2b(x).digest()[:32]` instead of `blake2b(x, digest_size=32)` is because using default algorithm is easier to tinker and some libraries may not support digest sizes for generating 32 bytes output.

Action: move these utils APIs to `eth/utils/`, and the tests.

hwwhww:
### What was wrong?
Fixes #1322

### How was it fixed?
I was thinking of putting these APIs to `eth.beacon_chain.utils` at the beginining, but those functions will also be used for shard chains, so I put to `eth.utils` in the end.

#### Cute Animal Picture

dall_sheep

(yay I miss cute animal pictures!)

### DIFF
--- a/eth/utils/bitfield.py
+++ b/eth/utils/bitfield.py
@@ -37,17 +37,13 @@ def get_vote_count(bitfield: bytes) -> int:
     )
 
 
-def or_bitfields(bitfields: List[bytes], allow_different_size=False) -> bytes:
-    if allow_different_size:
-        bytes_length = max([len(b) for b in bitfields])
-    else:
-        bytes_length = len(bitfields[0])
-
+def or_bitfields(bitfields: List[bytes]) -> bytes:
+    bytes_length = len(bitfields[0])
     new = b''
     for i in range(bytes_length):
         byte = 0
         for bitfield in bitfields:
-            if not allow_different_size and len(bitfield) != bytes_length:
+            if len(bitfield) != bytes_length:
                 raise ValueError("The bitfield sizes are different")
             if i < len(bitfield):
                 byte = bitfield[i] | byte
--- a/eth/utils/bitfield.py
+++ b/eth/utils/bitfield.py
@@ -1,0 +1,41 @@
+from typing import (
+    List,
+)
+
+
+def has_voted(bitfield: bytes, index: int) -> bool:
+    return bool(bitfield[index // 8] & (128 >> (index % 8)))
+
+
+def set_voted(bitfield: bytes, index: int) -> bytes:
+    byte_index = index // 8
+    bit_index = index % 8
+    new_byte_value = bitfield[byte_index] | (128 >> bit_index)
+    return bitfield[:byte_index] + bytes([new_byte_value]) + bitfield[byte_index + 1:]
+
+
+def get_bitfield_length(bit_count: int) -> int:
+    """Return the length of the bitfield for a given number of attesters in bytes."""
+    return (bit_count + 7) // 8
+
+
+def get_empty_bitfield(bit_count: int) -> bytes:
+    return b"\x00" * get_bitfield_length(bit_count)
+
+
+def get_vote_count(bitfield: bytes) -> int:
+    votes = 0
+    for index in range(len(bitfield) * 8):
+        if has_voted(bitfield, index):
+            votes += 1
+    return votes
+
+
+def or_bitfields(bitfields: List[bytes]) -> bytes:
+    new = b''
+    for i in range(len(bitfields[0])):
+        byte = 0
+        for bitfield in bitfields:
+            byte = bitfield[i] | byte
+        new += bytes([byte])
+    return new
--- a/eth/utils/bitfield.py
+++ b/eth/utils/bitfield.py
@@ -1,8 +1,12 @@
 from typing import (
     List,
 )
+from cytoolz import (
+    curry,
+)
 
 
+@curry
 def has_voted(bitfield: bytes, index: int) -> bool:
     return bool(bitfield[index // 8] & (128 >> (index % 8)))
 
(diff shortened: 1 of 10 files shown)


eth/utils/blake.py
```diff
@@ -0,0 +1,10 @@
+try:
+    # Python 3.6
+    from hashlib import blake2b
+except ImportError:
+    # Python 3.5
```
pipermerriam: I don't think we need to support 3.5, so maybe we can remove this and only focus on 3.6+ for beacon things.

eth/utils/bitfield.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_empty_bitfield(bit_count: int) -> bytes:
+    return b"\x00" * get_bitfield_length(bit_count)
+
+
+def get_vote_count(bitfield: bytes) -> int:
+    votes = 0
+    for index in range(len(bitfield) * 8):
+        if has_voted(bitfield, index):
+            votes += 1
```
pipermerriam: Non mutative approach:

```python
# make has_voted curriable
@curry
def has_voted(bitfield, index):
    ...

return len(tuple(filter(has_voted(bitfield), (idx for idx in range(len(bitfield) * 8)))))
```

psuedo-code but it should convey the idea.  The `filter` could be replaced with a comprehension using `[v for v in thing if has_voted(...)]` which might increase readability.  Also line-breaks and indentation can improve readability a lot here.

eth/utils/bitfield.py
```diff
@@ -0,0 +1,41 @@
...
+    return votes
+
+
+def or_bitfields(bitfields: List[bytes]) -> bytes:
+    new = b''
+    for i in range(len(bitfields[0])):
+        byte = 0
+        for bitfield in bitfields:
+            byte = bitfield[i] | byte
```
pipermerriam: This will raise an `IndexError` if any of the bitfields are shorted in length than the bitfield at index `0`.

eth/utils/bitfield.py
```diff
@@ -0,0 +1,41 @@
...
+def get_vote_count(bitfield: bytes) -> int:
+    votes = 0
+    for index in range(len(bitfield) * 8):
+        if has_voted(bitfield, index):
+            votes += 1
+    return votes
+
+
+def or_bitfields(bitfields: List[bytes]) -> bytes:
+    new = b''
```
pipermerriam: A non-mutative approach which depending on desired behavior will require some pre-validation is as follows.  Some behavior which needs to be specified are things like what happens when the various bitfields have different lengths...

Also, the following code will need to be special cased for situations where there are `0` or `1` bitfields.

The following code does the following algorithm.

1. group all of the bytes from each bitfield in a sequence of sequences where the sequence at index `0` is all of the bytes at index `0` from each of the bitfields.
2. do a map reduce using the binary *or* operator on each of the sequences.
3. join the resulting bytes back into a single bitfield

```python
from cytoolz.curried import reduce

byte_slices = zip(*bitfields)
return bytes(map(reduce(operator.or_), byte_slices))
```

Above is untested psuedo code but it should convey the idea.

eth/utils/blake.py
hwwhww: It's currently covered by `py35-core` tests. Do you think I should separate `beacon-utils` tests from `core`? Or we should work on another new branch instead of the `master` branch?

hwwhww:
> Thoughts  on the `bls` utils just being added to `py_ecc`?

hwwhww:
> Thoughts  on the `bls` utils just being added to `py_ecc`?

Checked with Vitalik, the current `hash_to_G2` function has not been standardized. So it's not 100% ready to be in `py_ecc` library. I suggest we keep it in Py-EVM in the short term.

eth/utils/blake.py
pipermerriam: How about we spin up a new CI run and relocate these tests to live under `tests/beacon` or maybe `tests/core-36` and we can merge once we drop 3.5 in core.

pipermerriam: Actually, ignore that, lets just mark all of those `3.6` only tests with a pytest skip

https://docs.pytest.org/en/latest/skipping.html#id1

hwwhww:
@pipermerriam It turns out the CI failing is somehow relevant to this PR...  because I added `tests/core/b*` tests, and the `tests/core/builder-tools/*` tests are running *after* those new tests. So here timeout:
https://github.com/ethereum/py-evm/blob/e25731d94f57e66bcd2c2e64c68174542207109d/tests/core/builder-tools/test_chain_initializer.py#L42
I guess the expectation of `header.timestamp` is the time we call `chain = pipe(chain_class, genesis(), )`? However, now the result is actually **the timestamp of loading `eth.tools.builder.chain.builder.GENESIS_DEFAULTS`**, which is before all the tests:
https://github.com/ethereum/py-evm/blob/28f5129da9b953c376a2502d6a84e2357972d698/eth/tools/builder/chain/builders.py#L191-L206

I suggest to remove this assertion in this PR and open another issue for fixing it if need.

pipermerriam:
> I suggest to remove this assertion in this PR and open another issue for fixing it if need.

Can you just remove the `timestamp` from the hard coded defaults, reduce the assertion delta to be 2 seconds, and update `_get_default_genesis_params` to dynamically populate the `timestamp` value at runtime?

pipermerriam:
Also, good find.  We were confused as to why that was occasionally happening but never spent any time digging in.

PR #1326, opened 2018-09-26
Fix unhandled ValueError raised during handshake

carver:
### What was wrong?

An invalid public key was causing an uncaught `ValueError` during handshake

### How was it fixed?

Catch the `ValueError` raised by the encryption library on a bad public key, and re-raise a `DecryptionError`.

#### Cute Animal Picture

### DIFF
--- a/p2p/ecies.py
+++ b/p2p/ecies.py
@@ -34,6 +34,14 @@ CURVE = ec.SECP256K1()
 KEY_LEN = 32
 
 
+class _InvalidPublicKey(Exception):
+    """
+    A custom exception raised when trying to convert bytes
+    into an elliptic curve public key.
+    """
+    pass
+
+
 def generate_privkey() -> datatypes.PrivateKey:
     """Generate a new SECP256K1 private key and return it"""
     privkey = ec.generate_private_key(CURVE, default_backend())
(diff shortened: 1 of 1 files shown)


pipermerriam:
Meta comment, not requesting anything in this PR.  I would like for us to find a way to test these things.  We have probably 20+ of these patches/fixes that have gone in, catching various edge case exceptions that show up when peers send invalid information.  Our strategy for fixing them is fine but I suspect we are very quickly approaching a point where some of these will do one of the following:

- become a dead code path due to refactor resulting in the exception scenario not being possible anymore.
- masque an error in our implementation

The later of these actually doesn't worry me much because I suspect most of those cases will manifest as problems elsewhere with other clients disconnecting us.

However, if they become dead code paths, we'll just be living with that dead code forever which is lame.  Solution seems to be trying to figure out an avenue to test these, but I suspect those will end up being very hard to read tests that are hard to maintain and potentially fragile.

p2p/ecies.py
```diff
@@ -100,7 +100,12 @@ def decrypt(data: bytes, privkey: datatypes.PrivateKey, shared_mac_data: bytes =
 
     #  1) generate shared-secret = kdf( ecdhAgree(myPrivKey, msg[1:65]) )
     shared = data[1:1 + PUBKEY_LEN]
-    key_material = ecdh_agree(privkey, keys.PublicKey(shared))
+    try:
+        key_material = ecdh_agree(privkey, keys.PublicKey(shared))
+    except ValueError as exc:
+        raise DecryptionError(
```
pipermerriam: Requesting comment explaining why this happens.

Issue #1103, opened 2018-07-27
Create guide on how to write plugins


cburgdorf:
### What is wrong?

When the plugin API has settled a bit (so, not ready yet!), we should add a guide that explains how to create plugins.

### How can it be fixed

:writing_hand: :wine_glass: :writing_hand: :sun_with_face: :wine_glass: :writing_hand:

pipermerriam:
I think it would be good to write down some of the high level things like intended use cases and basics of the API early, and we can put some big loud `.. warning::` admonitions at the top of that page.

pipermerriam:
Also maybe link into the code to the current internals which are implemented as plugins.

cburgdorf:
I think I'm gonna sit down and come up with something after #1212 lands. The API is still very early and in flux but it's approaching a state where one can do more and more useful things. The usefulness will climb up when we begin to broadcast more and more events via the eventbus.

cburgdorf:
The guide landed a while ago (Albeit, there's still much room for improvement)

PR #1331, opened 2018-09-27
Refactor plugin support to exclusively use EventBus events

cburgdorf:
### What was wrong?

Currently plugins have two different ways of consuming events:

1. Comsuming `lahja` events through the `EventBus`

2. Consuming *plugin events*  through `handle_event`

With the upcoming `lahja` release, this can be unified.

### How was it fixed?

1. Plugins consume `EventBus` events only
2. `should_start` is gone. The model has turned. While previously the `PluginManager` called `_start()` on the plugins, plugins do now call `handle_start()` themselves.
3. Because of the former, it is now the responsible of the plugin base classes to flip a `running` property on which the `PluginManager` figures out which plugins need to be stopped
3. The `Node` does not depend on the `PluginManager` any more

### Known issues:

2. I think the `start` and `handle_start`  methods could use better names. The `handle_start` method is what plugins need to *call* to start the plugin whereas the `start` method is what they need to *implement*.

#### Cute Animal Picture

### DIFF
--- a/trinity/extensibility/plugin.py
+++ b/trinity/extensibility/plugin.py
@@ -115,18 +115,18 @@ class BasePlugin(ABC):
         """
         pass
 
-    def boot(self) -> None:
+    def start(self) -> None:
         """
         Prepare the plugin to get started and eventually cause ``start`` to get called.
         """
         self.running = True
-        self.start()
+        self._start()
         self.event_bus.broadcast(
             PluginStartedEvent(type(self))
         )
         self.logger.info("Plugin started: %s", self.name)
 
-    def start(self) -> None:
+    def _start(self) -> None:
         """
         The ``start`` method is called only once when the plugin is started. In the case
         of an `BaseIsolatedPlugin` this method will be launched in a separate process.
@@ -138,8 +138,19 @@ class BaseSyncStopPlugin(BasePlugin):
     """
     A ``BaseSyncStopPlugin`` unwinds synchronoulsy, hence blocks until shut down is done.
     """
-    def stop(self) -> None:
+    def _stop(self) -> None:
+        """
+        Stop the plugin. Should be overwritten by subclasses.
+        """
         pass
+
+    def stop(self) -> None:
+        """
+        Stop the plugin by delegating to
+        :meth:`~trinity.extensibility.plugin.BaseSyncStopPlugin._stop`
+        """
+        self._stop()
+        self.running = False
 
 
 class BaseAsyncStopPlugin(BasePlugin):
@@ -147,8 +158,19 @@ class BaseAsyncStopPlugin(BasePlugin):
     A ``BaseAsyncStopPlugin`` unwinds asynchronoulsy, hence needs to be awaited.
     """
 
-    async def stop(self) -> None:
+    async def _stop(self) -> None:
+        """
+        Asynchronously stop the plugin. Should be overwritten by subclasses.
+        """
         pass
+
+    async def stop(self) -> None:
+        """
+        Asynchronously stop the plugin by delegating to
+        :meth:`~trinity.extensibility.plugin.BaseAsyncStopPlugin._stop`
+        """
+        await self._stop()
+        self.running = False
 
 
 class BaseMainProcessPlugin(BasePlugin):
@@ -170,7 +192,7 @@ class BaseIsolatedPlugin(BaseSyncStopPlugin):
 
     _process: Process = None
 
-    def boot(self) -> None:
+    def start(self) -> None:
         """
         Prepare the plugin to get started and eventually cause ``start`` to get called.
         """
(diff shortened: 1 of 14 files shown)


cburgdorf:
@pipermerriam this is now based on the freshly released `lahja` `0.9.0` release and can be given a full review. The thing that I personal dislike the most about it is the naming of `start` and `handle_start` where `start` is a method to be **implemented** by plugins to wrap the boot process of the plugin and `handle_start` is what plugins need to call to actually start itself when they want to.

Maybe we should rename `handle_start` to `run` and `start` to `bootstrap`?

trinity/extensibility/plugin.py
```diff
@@ -85,35 +90,41 @@ def name(self) -> str:
...
         """
         Set the :class:`~trinity.extensibility.plugin.PluginContext` for this plugin.
         """
         self.context = context
 
-    def configure_parser(self, arg_parser: ArgumentParser, subparser: _SubParsersAction) -> None:
+    def ready(self) -> None:
         """
-        Called at startup, giving the plugin a chance to amend the Trinity CLI argument parser
+        Called after the plugin received its context and is ready to bootstrap itself.
```
pipermerriam: It'd be cool to have a large docstring for this class which diagram'd the lifecycle of a plugin (and probably a page in the docs once the API stablizes.)

trinity/extensibility/plugin.py
```diff
@@ -85,35 +90,41 @@ def name(self) -> str:
...
         """
-        Notify the plugin about an event, giving it the chance to do internal accounting right
-        before :meth:`~trinity.extensibility.plugin.BasePlugin.should_start` is called
+        Called at startup, giving the plugin a chance to amend the Trinity CLI argument parser
         """
-
         pass
 
-    def should_start(self) -> bool:
+    def handle_start(self) -> None:
```
pipermerriam: I like `bootstrap` or `boot`.

trinity/extensibility/plugin.py
```diff
@@ -85,35 +90,41 @@ def name(self) -> str:
...
-
-    def _start(self) -> None:
+        self.running = True
         self.start()
+        self.event_bus.broadcast(
+            PluginStartedEvent(type(self))
+        )
+        self.logger.info("Plugin started: %s", self.name)
 
     def start(self) -> None:
```
pipermerriam: I'm ok with `run` but I think I like `start` more since `stop` is a natural opposite.

trinity/extensibility/plugin_manager.py
```diff
@@ -219,14 +187,15 @@ def shutdown_blocking(self) -> None:
...
+        for plugin in self._plugin_store:
 
-            if not isinstance(plugin, BaseSyncStopPlugin):
+            if not isinstance(plugin, BaseSyncStopPlugin) or not plugin.running:
                 continue
 
             try:
                 self._logger.info("Stopping plugin: %s", plugin.name)
                 plugin.stop()
+                plugin.running = False
```
pipermerriam: This seems wrong for something external to the plugin to manage its state.  Shouldn't this be handled by the `plugin.stop` method?

trinity/extensibility/plugin.py
cburgdorf: Yes, I plan to improve docs for plugins soon! (https://github.com/ethereum/py-evm/issues/1103)

trinity/extensibility/plugin_manager.py
cburgdorf: Mmh...well, since `stop` is implemented by each individual plugin that would mean that plugin authors either need to remember to call `super().stop()` (bad because if they forget it, they are basically leaving the `PluginManager` thinking that the plugin would still be running when in reality it isn't.

Alternatively we do the same as we do with the `start` / `boot` pair which means there will be another method (`shutdown` ?) which sets `running=False` and then delegates to `stop()`. At least, that would be symmetric to the `start / boot` pair. 

However setting `plugin.running` from the manager lets us avoid this extra indirection and I'm personally ok with having the `PluginManager` make the flip here. Your call.

trinity/extensibility/plugin.py
cburgdorf: Right, then let's leave this be called `start` and rename `handle_start` to `boot`

trinity/extensibility/plugin_manager.py
pipermerriam: I'm generally a fan of the model you proposed with `stop/shutdown`.  One is internal, only accessed by the manager and handles the orchestration like setting flags, triggering `asyncio.Event`s, the other is the hook for the user to implement the logic.  We've been using this with `BaseService` for a while and I think it's working out well.

cburgdorf: Alright, will refactor that previous to merging this.

cburgdorf: @pipermerriam this https://github.com/ethereum/py-evm/pull/1331/commits/5ee9b53b73152eb0194bb9bf6660b39d73031028

PR #1338, opened 2018-09-30
Beacon types

### LINKED ISSUE #1333: Add beacon chain data classes
### What is wrong?
#930 - part 2 - moving the data classes from beacon_chain repo to Py-EVM.

### How can it be fixed
In [beacon_chain repo](https://github.com/ethereum/beacon_chain/tree/master/beacon_chain/state), there some classes:
- active_state.py
- attestation_record.py
- block.py
- crosslink_record.py
- crystallized_state.py
- shard_and_committee.py
- validator_record.py

In the Eth2.0 call yesterday, [Alexey pointed out some problems with SimpleSerialize](https://github.com/ethereum/eth2.0-pm/issues/8#issuecomment-424849447). At this point, I suggest we use RLP for now before it's standardized. @djrtwo any thoughts?

Action:
- Move these classes to `eth/beacon/types/`
- Move tests to `tests/core/beacon/types/`

Note: there are some handy cache data in `ActiveState`, I think we can remove it and add some other Py-EVM style cache mechanism later.

hwwhww:
### What was wrong?
Closes #1333 

### How was it fixed?
1. Added `eth.beacon.types` - currently, using RLP for serialization until we have standardized it. (maybe [SSZ](https://github.com/ethereum/beacon_chain/tree/master/ssz), or some hybrid solutions.)
2. Added `py36-beacon`, `py37-beacon` in CI.
3. For implementing `CrystallizedState.active_validator_indices` property, added `eth.beacon.helpers.get_active_validator_indices`. Will add other helper functions in other PRs.

#### Cute Animal Picture

sheep-bleat-communication-communicate-437618


djrtwo:
@pipermerriam Would you put these types in `beacon/state_machine/forks/fork_name` to be proactive in being able to handle forking when it inevitably comes?

eth/beacon/types/validator_record.py
```diff
@@ -0,0 +1,54 @@
...
+        # Current balance
+        ('balance', int128),
+        # Dynasty where the validator is inducted
+        ('start_dynasty', int64),
+        # Dynasty where the validator leaves
+        ('end_dynasty', int64),
+    ]
+
+    def __init__(self,
+                 pubkey=b'',
```
pipermerriam: Is the `0` public key a good reasonable default?

tests/beacon/conftest.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@pytest.fixture
+def slot_duration():
+    return SLOT_DURATION
+
+
+@pytest.fixture
+def sqrt_e_drop_time():
+    return SQRT_E_DROP_TIME
```
pipermerriam: This seems like a lot of code overhead for a bunch of things that are simple defaults.  We could drop all of these fixtures and just bake these defaults into the `config` fixture.

tests/beacon/conftest.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@pytest.fixture
+def config(base_reward_quotient,
```
pipermerriam: maybe namespace this as `beacon_config` since `config` is a very generic name.

tests/beacon/helpers.py
```diff
@@ -0,0 +1,18 @@
+from eth.beacon.config import (
+    DEFAULT_CONFIG,
+)
+from eth.beacon.types.validator_record import (
+    ValidatorRecord,
+)
+
+
+def mock_validator_record(pubkey, start_dynasty=0, config=DEFAULT_CONFIG):
```
pipermerriam: It seems problematic that this uses `DEFAULT_CONFIG` by default, yet there is another `config` object produced by the testing fixtures...  This seems like a good way to accidentally use two different configurations within the same test.

tests/beacon/types/test_block.py
```diff
@@ -0,0 +1,56 @@
...
+)
+def test_defaults(param, default_value, sample_block_params):
+    del sample_block_params[param]
+    block = Block(**sample_block_params)
+    assert getattr(block, param) == default_value
+
+
+def test_update_slot():
+    block = Block()
+    block = block.copy(slot_number=10)
```
pipermerriam: This test should probably be removed as it's testing 3rd party library functionality (which is tested as part of that library).

tests/beacon/types/test_block.py
```diff
@@ -0,0 +1,56 @@
...
+    attestations.append(AttestationRecord())
+    block2 = block.copy(
+        attestations=attestations
+    )
+    assert block2.num_attestations == 1
+
+
+def test_hash():
+    block = Block()
+    assert block.hash == blake(rlp.encode(block))
```
pipermerriam: Rather than testing the specific implementation, these tests might be better to test against an explicitly hash computed from known values.  As it stands, this is really testing the implementation.  Not zero value since it acts as a reasonable sanity check.

eth/beacon/types/block.py
```diff
@@ -0,0 +1,76 @@
...
+        ('pow_chain_ref', hash32),
+        # Hash of the active state
+        ('active_state_root', hash32),
+        # Hash of the crystallized state
+        ('crystallized_state_root', hash32),
+    ]
+
+    def __init__(self,  # noqa: F811
+                 parent_hash=ZERO_HASH32,
+                 slot_number=0,
```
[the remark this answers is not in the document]
djrtwo: agreed. Some of these defaults were put in place during early testing. 

You do need to be able to create a block without specifying `active_state_root` and `crystallized_state_root`.

The workflow is. Create a block, perform the state transitions to get the new state roots, insert the expected end state roots.

eth/beacon/types/crosslink_record.py
```diff
@@ -0,0 +1,34 @@
...
+        ('dynasty', int64),
+        # slot during which crosslink was added
+        ('slot', int64),
+        # The block hash
+        ('hash', hash32),
+    ]
+
+    def __init__(self,
+                 dynasty=0,
+                 slot=0,
```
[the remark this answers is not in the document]
djrtwo: There is a set of "genesis" crosslinks -- 1 for each shard -- that are inserted into the genesis state. These params match those genesis params.

That said, that should probably happen in an explicit separate process rather than be applied as defaults to each.

tests/beacon/conftest.py
djrtwo: I personally like this construction because I can do things like the following:

```python
@pytest.mark.parametrize(
    (
        'cycle_length,min_committee_size,shard_count'
    ),
    [
        (50, 10, 10),
        (20, 10, 100),
    ],
)
def test(genesis_crystallized_state,
         genesis_active_state,
         mock_make_block):
    do_test()
```
This allows for setting these configuration variables via `parametrize` and the config vars that are set this way percolate through the fixtures, whereas the ones untouched stay as defaults.

I agree it's a strange construction. Do you have a better alternative?

eth/beacon/types/attestation_record.py
```diff
@@ -0,0 +1,64 @@
...
+        # Last justified block
+        ('justified_slot', int64),
+        ('justified_block_hash', hash32),
+        # The actual signature
+        ('aggregate_sig', CountableList(int256)),
+    ]
+
+    def __init__(self,
+                 slot=0,
+                 shard_id=0,
```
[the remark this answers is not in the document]
hwwhww: @pipermerriam I think we mixed up "demonstration values" and "default values" together in beacon chain repo. Thanks for pointing out!

eth/beacon/types/validator_record.py
hwwhww: will remove all default values from ValidatorRecord.

hwwhww:
Thanks for the feedback! Applied all PR feedback expect for the fixture one. The tests configuration probably will be changed a lot in #1339 - `4. Tear down helpers` anyway.

tests/beacon/conftest.py
pipermerriam: Maybe something with environment variables and potentially the `monkeypatch` fixture, though I'm moving more in favor of removing the `config` *global* concept towards a `constants` module and configurable class based approach.

I.E. The `cycle_length` would be part of some class and retrieved with `BeaconChain.get_cycle_length()` with the default implementation returning `constants.CYCLE_LENGTH`.

- Easy to modify at runtime
- easy to integrate into the chain builder API
- avoids the use of globals
- doesn't require passing around a config object (which will result in annoying dependencies based on my experience

pipermerriam: Oh, and if other agree to go with this approach I would suggest not doing it in this PR, but rather getting this cleaned up, merging, and doing it in a subsequent one.

djrtwo: sounds good 👍

pipermerriam:
@hwwhww can you open an issue to track conversion of `config` as discussed here: https://github.com/ethereum/py-evm/pull/1338#discussion_r221757203

Issue #1339, opened 2018-10-01
Add beacon chain helper functions


hwwhww:
### What is wrong?
Part of #930 

- The [helper functions in beacon_chain](https://github.com/ethereum/beacon_chain/blob/master/beacon_chain/state/helpers.py) are all pure functions with a `config: Dict[string, Any]` parameter. 
- Some helpers functions are like ~90% similar to [the spec](https://notes.ethereum.org/s/Syj3QZSxm#Helper-functions) as a reference implementation. But to fit Py-EVM paradigm, we may have to refactor a lot.

### How can it be fixed
It may take multiple steps:
1. Move all pure functions to `eth/beacon/helpers.py` - I have a commit here: https://github.com/hwwhww/py-evm/commit/5724a7a30ce56979382c7ea07e29e08fcaf21254 
2. Introduce `eth.beacon.db.chain.BeaconChainDB`
3. Introduce `eth.beacon.state_machines.base.BaseStateMachine(Configurable)`
4. Introduce `eth.beacon.chains.chain.BeaconChain`
4. Tear down helpers: move some of them to `eth.beacon.state_machine.base.BaseStateMachine` or being static functions / properties in `eth.beacon.types.*` classes.

cc @djrtwo @pipermerriam

pipermerriam:
- The `config` object can probably be updated to be a `typing.NamedTuple` for stronger type safety.
- Modifying the helper functions to be non mutative looks like something that will go relatively quickly.
- Writing tests for all of these is probably the largest task (take care to not waste time writing exhaustive tests on things that are likely to change).

pipermerriam:
Also, ideally, rather than passing a `config` object around through these various helper functions, it would probably be better to just explicitly pass the specific configuration values as individual parameters.

pipermerriam:
On *second* thought, I'm curious if we can get away with the same pattern that the existing code uses, placing these in a `constants` file and importing them as needed.  For parameters that end up changing for various forks, we can place them in `eth.beacon.forks.fork_name.constants` (as danny suggested)

hwwhww:
Yes, that makes sense. We will have `Configurable` `StateMachine` to handle these chain configurations.

djrtwo:
Yeah, I agree the import is the way to go on these

hwwhww:
closed via #1341 and #1381

Issue #1340, opened 2018-10-02
Refactor BeaconConfig related fixtures


hwwhww:
### What is wrong?
In current [tests/beacon/conftest.py](https://github.com/ethereum/py-evm/blob/c78020fe0cf6b4d98b93264872dfd10c59757e06/tests/beacon/conftest.py#L135-L196), these fixtures are for setting configuration variables via `parametrize`, but still, strange.

### How can it be fixed
https://github.com/ethereum/py-evm/pull/1338#discussion_r221757203 by Piper

Removing the `config` *global* concept towards a `constants` module and configurable class based approach.

I.E. The `cycle_length` would be part of some class and retrieved with `BeaconChain.get_cycle_length()` with the default implementation returning `constants.CYCLE_LENGTH`.

- Easy to modify at runtime
- easy to integrate into the chain builder API
- avoids the use of globals
- doesn't require passing around a config object (which will result in annoying dependencies based on my experience

hwwhww:
closed via #1373

PR #1341, opened 2018-10-02
Port beacon chain helper functions

hwwhww:
### What was wrong?
#1339 - 1. Move all pure functions

### How was it fixed?
1. Move all `helpers.py` functions to `eth/beacon/helpers.py`
2. Move all `genesis_helpers.py` functions to `eth/beacon/genesis_helpers.py`

#### Cute Animal Picture

pexels-photo-208847

### DIFF
--- a/eth/beacon/config.py
+++ b/eth/beacon/config.py
@@ -1,5 +1,5 @@
-from collections import (
-    namedtuple,
+from typing import (
+    NamedTuple,
 )
 
 from eth_utils import denoms
@@ -19,20 +19,20 @@ SQRT_E_DROP_TIME = 2**20  # seconds
 assert SQRT_E_DROP_TIME % SLOT_DURATION == 0
 
 
-BeaconConfig = namedtuple(
+BeaconConfig = NamedTuple(
     'BeaconConfig',
-    [
-        'base_reward_quotient',
-        'default_end_dynasty',
-        'deposit_size',
-        'cycle_length',
-        'max_validator_count',
-        'min_committee_size',
-        'min_dynasty_length',
-        'shard_count',
-        'slot_duration',
-        'sqrt_e_drop_time',
-    ]
+    (
+        ('base_reward_quotient', int),
+        ('default_end_dynasty', int),
+        ('deposit_size', int),
+        ('cycle_length', int),
+        ('max_validator_count', int),
+        ('min_committee_size', int),
+        ('min_dynasty_length', int),
+        ('shard_count', int),
+        ('slot_duration', int),
+        ('sqrt_e_drop_time', int),
+    )
 )
 
 
@@ -47,7 +47,6 @@ def generate_config(*,
                     shard_count: int=SHARD_COUNT,
                     slot_duration: int=SLOT_DURATION,
                     sqrt_e_drop_time: int=SQRT_E_DROP_TIME) -> BeaconConfig:
-
     return BeaconConfig(
         base_reward_quotient=base_reward_quotient,
         default_end_dynasty=default_end_dynasty,
--- a/eth/beacon/config.py
+++ b/eth/beacon/config.py
@@ -49,16 +49,16 @@ def generate_config(*,
                     sqrt_e_drop_time: int=SQRT_E_DROP_TIME) -> BeaconConfig:
 
     return BeaconConfig(
-        base_reward_quotient=BASE_REWARD_QUOTIENT,
-        default_end_dynasty=DEFAULT_END_DYNASTY,
-        deposit_size=DEPOSIT_SIZE,
-        cycle_length=CYCLE_LENGTH,
-        max_validator_count=MAX_VALIDATOR_COUNT,
-        min_committee_size=MIN_COMMITTEE_SIZE,
-        min_dynasty_length=MIN_DYNASTY_LENGTH,
-        shard_count=SHARD_COUNT,
-        slot_duration=SLOT_DURATION,
-        sqrt_e_drop_time=SQRT_E_DROP_TIME
+        base_reward_quotient=base_reward_quotient,
+        default_end_dynasty=default_end_dynasty,
+        deposit_size=deposit_size,
+        cycle_length=cycle_length,
+        max_validator_count=max_validator_count,
+        min_committee_size=min_committee_size,
+        min_dynasty_length=min_dynasty_length,
+        shard_count=shard_count,
+        slot_duration=slot_duration,
+        sqrt_e_drop_time=sqrt_e_drop_time
     )
 
 
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -205,7 +205,8 @@ def get_attestation_indices(crystallized_state: 'CrystallizedState',
 def get_active_validator_indices(dynasty: int,
                                  validators: Iterable['ValidatorRecord']) -> Iterable[int]:
     """
-    TODO: Logic changed
+    TODO: Logic changed in the latest spec, will have to update when we add validator
+    rotation logic.
     https://github.com/ethereum/eth2.0-specs/commit/52cf7f943dc99cfd27db9fb2c03c692858e2a789#diff-a08ecec277db4a6ed0b3635cfadc9af1  # noqa: E501
     """
     o = []
@@ -219,24 +220,44 @@ def get_active_validator_indices(dynasty: int,
 # Shuffling
 #
 def _get_shuffling_committee_slot_portions(
-        active_validators_length: int,
+        active_validators_size: int,
         cycle_length: int,
         min_committee_size: int,
         shard_count: int) -> Tuple[int, int]:
-    if active_validators_length >= cycle_length * min_committee_size:
+    """
+    Returns committees number per slot and slots number per committee
+    """
+    # If there are enough active validators to form committees for every slot
+    if active_validators_size >= cycle_length * min_committee_size:
+        # One slot can be attested by many committees, but not more than shard_count // cycle_length
         committees_per_slot = min(
-            active_validators_length // cycle_length // (min_committee_size * 2) + 1,
+            active_validators_size // cycle_length // (min_committee_size * 2) + 1,
             shard_count // cycle_length
         )
+        # One committee only has to attest one slot
         slots_per_committee = 1
     else:
+        # One slot can only be asttested by one committee
         committees_per_slot = 1
+        # One committee has to asttest more than one slot
         slots_per_committee = 1
-        bound = cycle_length * min(min_committee_size, active_validators_length)
-        while(active_validators_length * slots_per_committee < bound):
+        bound = cycle_length * min(min_committee_size, active_validators_size)
+        while(active_validators_size * slots_per_committee < bound):
             slots_per_committee *= 2
 
     return committees_per_slot, slots_per_committee
+
+
+@to_tuple
+def _get_shards_and_committees_for_shard_indices(
+        shard_indices: Sequence[Sequence[int]],
+        shard_id_start: int,
+        shard_count: int) -> Iterable[ShardAndCommittee]:
+    for j, indices in enumerate(shard_indices):
+        yield ShardAndCommittee(
+            shard_id=(shard_id_start + j) % shard_count,
+            committee=indices
+        )
 
 
 @to_tuple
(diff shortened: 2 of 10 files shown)


eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+        current_block: 'Block',
+        slot: int,
+        beacon_config: 'BeaconConfig') -> Hash32:
+    """
+    Get the blockhash from active_state.recent_block_hashes by current block slot number.
+    """
+    cycle_length = beacon_config.cycle_length
+
+    sback = current_block.slot_number - cycle_length * 2
+    assert sback <= slot < sback + cycle_length * 2
```
pipermerriam: This should be converted to an `if/raise` (we avoid assertions in runtime code because they are easy to disable)

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+        active_state: 'ActiveState',
+        current_block: 'Block',
+        slot: int,
+        beacon_config: 'BeaconConfig') -> Hash32:
+    """
+    Get the blockhash from active_state.recent_block_hashes by current block slot number.
+    """
+    cycle_length = beacon_config.cycle_length
+
+    sback = current_block.slot_number - cycle_length * 2
```
pipermerriam: What does `sback` stand for?

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
 
 
+#
+# Get block hashes
+#
+def get_block_hash(
+        active_state: 'ActiveState',
+        current_block: 'Block',
+        slot: int,
+        beacon_config: 'BeaconConfig') -> Hash32:
```
pipermerriam: If #1340 is going to be prioritized soon, I have no issue leaving these in, but it may be prudent to go ahead and convert these functions to accept only the parameters from the `beacon_config` that they actually use.  Also, fine to clean it all up in one go in a subsequent PR

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+def get_hashes_to_sign(active_state: 'ActiveState',
+                       block: 'Block',
+                       beacon_config: 'BeaconConfig') -> List[Hash32]:
+    """
+    Given the head block to attest to, collect the list of hashes to be
+    signed in the attestation.
+    """
+    cycle_length = beacon_config.cycle_length
+
+    hashes = get_hashes_from_active_state(
```
pipermerriam: We can reduce some of the overhead here by changing this function to return `Iterable[Hash32]` and adjusting these lines to be:

```python
yield from get_hashes_from_active_state(...)
yield blake(block.hash)
```

In the current state we have to build the enumerable of hashes twice.

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+def get_signed_parent_hashes(active_state: 'ActiveState',
+                             block: 'Block',
+                             attestation: 'AttestationRecord',
+                             beacon_config: 'BeaconConfig') -> List[Hash32]:
+    """
+    Given an attestation and the block they were included in,
+    the list of hashes that were included in the signature
+    """
+    cycle_length = beacon_config.cycle_length
+    parent_hashes = get_hashes_from_active_state(
```
pipermerriam: Same here for using `yield from` to get rid of the `List`

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+    return active_state.recent_block_hashes[slot - sback]
+
+
+@to_tuple
+def get_hashes_from_active_state(active_state: 'ActiveState',
+                                 block: 'Block',
+                                 from_slot: int,
+                                 to_slot: int,
+                                 beacon_config: 'BeaconConfig') -> List[Hash32]:
+    hashes = [
```
pipermerriam: Update to use a `yield` pattern to get rid of the `List`

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+@to_tuple
+def get_shards_and_committees_for_slot(
+        crystallized_state: 'CrystallizedState',
+        slot: int,
+        beacon_config: 'BeaconConfig') -> List[ShardAndCommittee]:
+    cycle_length = beacon_config.cycle_length
+
+    start = crystallized_state.last_state_recalc - cycle_length
+    assert start <= slot < start + cycle_length * 2
+    return crystallized_state.shard_and_committee_for_slots[slot - start]
```
pipermerriam: Update to use `yield from` to get rid of the `List`

eth/beacon/helpers.py
```diff
@@ -1,17 +1,167 @@
...
+                beacon_config,
+            )
+        )
+    )
+
+    attestation_indices = []  # type: List[int]
+    if filtered_shards_and_committees_for_slot:
+        attestation_indices = filtered_shards_and_committees_for_slot[0].committee
+
+    return attestation_indices
```
pipermerriam: convert to `yield from` to get rid of the `List`

eth/beacon/helpers.py
```diff
@@ -20,3 +170,100 @@ def get_active_validator_indices(dynasty: int,
         if (validator.start_dynasty <= dynasty and dynasty < validator.end_dynasty):
             o.append(index)
     return o
+
+
+#
+# Shuffling
+#
+def shuffle(lst: List[Any],
```
pipermerriam: `values` or `items` is often a good variable name for this type of thing.

eth/beacon/helpers.py
```diff
@@ -20,3 +170,100 @@ def get_active_validator_indices(dynasty: int,
...
+            rand_max = 16777216 - 16777216 % remaining
+            if m < rand_max:
+                replacement_pos = (m % remaining) + i
+                o[i], o[replacement_pos] = o[replacement_pos], o[i]
+                i += 1
+    return o
+
+
+@to_tuple
+def split(lst: List[Any], number: int) -> List[Any]:
```
pipermerriam: I believe this function is the same as `cytoolz.partition_all`

eth/beacon/helpers.py
```diff
@@ -20,3 +170,100 @@ def get_active_validator_indices(dynasty: int,
...
+            shard_count // cycle_length
+        )
+        slots_per_committee = 1
+    else:
+        committees_per_slot = 1
+        slots_per_committee = 1
+        while (len(avs) * slots_per_committee < cycle_length * min_committee_size and
+               slots_per_committee < cycle_length):
+            slots_per_committee *= 2
+    o = []
```
pipermerriam: Should be trivial to convert this function to use a `yield` pattern.

eth/beacon/helpers.py
```diff
@@ -20,3 +170,100 @@ def get_active_validator_indices(dynasty: int,
...
+            len(avs) // cycle_length // (min_committee_size * 2) + 1,
+            shard_count // cycle_length
+        )
+        slots_per_committee = 1
+    else:
+        committees_per_slot = 1
+        slots_per_committee = 1
+        while (len(avs) * slots_per_committee < cycle_length * min_committee_size and
+               slots_per_committee < cycle_length):
+            slots_per_committee *= 2
```
pipermerriam: It seems like we should be able to compute the final value without the `while` loop with some simple algebra.

eth/beacon/helpers.py
```diff
@@ -20,3 +170,100 @@ def get_active_validator_indices(dynasty: int,
...
+        for i in range(number)
+    ]
+
+
+@to_tuple
+def get_new_shuffling(seed: Hash32,
+                      validators: List['ValidatorRecord'],
+                      dynasty: int,
+                      crosslinking_start_shard: int,
+                      beacon_config: 'BeaconConfig') -> List[List[ShardAndCommittee]]:
```
pipermerriam: This function could use a docstring as well as some inline comments to explain what is happening and why.

eth/beacon/helpers.py
```diff
@@ -20,3 +170,100 @@ def get_active_validator_indices(dynasty: int,
...
+#
+def get_proposer_position(parent_block: 'Block',
+                          crystallized_state: 'CrystallizedState',
+                          beacon_config: 'BeaconConfig') -> Tuple[int, int]:
+    shards_and_committees = get_shards_and_committees_for_slot(
+        crystallized_state,
+        parent_block.slot_number,
+        beacon_config,
+    )
+    assert shards_and_committees
```
pipermerriam: convert to `if/raise`

eth/beacon/helpers.py
hwwhww: For some functions, e.g., `get_block_hash` actually can just take `recent_block_hashes, current_block_slot_number, target_slot, cycle_length` as the parameters, `get_signed_parent_hashes` can just take `recent_block_hashes, block_slot_number, attestation, cycle_length` as parameters. I'd like to clean/slim them in this PR so that they will be easier to create tests.

eth/beacon/helpers.py
pipermerriam: Ah, I still think you can do this in a cleaner way with `partition_all` (though I did initially misunderstand what this function was doing).

```python
def split(seq, chunks):
    chunk_size = len(seq) // chunks
    if chunk_size < 1:
        raise ValueError("Cannot split into chunks, not enough values")
    return partition_all(chunk_size, seq)
```

pipermerriam: I think you can make the first parameter a `Sequence` instead of a `List` to make it more generic as well as using `yield_from` to get rid of the `List` return type.

pipermerriam: Oh, and the `Any` should be updated to use a generic type.

```python
from typing import TypeVar
TItem = TypeVar('TItem')
def split(seq: Sequence[TItem], number: int) -> Iterable[TItem]:
    ...
```

This allows us to maintain type safety for values that pass through this function.  The `shuffle` function above should be updated in the same manner (and can use the same generic type).

hwwhww: I forgot to mention that `split` function is [in-protocol](https://github.com/ethereum/eth2.0-specs/blob/master/specs/casper_sharding_v2.1.md), so we still have to produce the *same* result as the old split function after refactorings. Note that with `split()`, the end chunks have more elements than the front chunks. I think stick with the spec is safer.

eth/beacon/helpers.py
hwwhww: I reduced the problem to:
```python
while a * b < c:
    b ** 2
```

and then not sure where else can be reduced. 

There might be other way beyond trying to calcuate `_get_shuffling_committee_slot_portions()`, but I haven't figured it out...  😨

pipermerriam: Not 100% sure this is right, but I think it's equivalent.  You'd do `b = b ** 2 ** int(math.ceil(n))` to get the final `B` value.

photo_2018-10-02_13-38-37

hwwhww: Bad news:
1. slap myself(!!!). a typo and misled you, it should be:
    ```python
    b = 1
    while a * b < c:
        b *= 2
    ```
    So since `b=1` initially, it's `2^N <= c/a`. -> `N <= ln(c/a)`

2. Tried `B = 2 ** int(math.log2(c / a)) when c // a > 1` and benchmarked. Unfortunately it doesn't improve much in our case. Basically because `B` is `slots_per_committee`, the edge case numbers in the spec are:
    ```
    a is active_validators_length,  domain is between [MIN_COMMITTEE_SIZE, MAX_VALIDATOR_COUNT] = [128, 4194304]    # the lower bound may be higher
    c would be CYCLE_LENGTH * MIN_COMMITTEE_SIZE, smallest: 64 * 128
    ```
    So the possible value of B is between [1, 64]. But `math` API strategy has advantage when number is large.

Many many thanks! ✨

hwwhww:
@pipermerriam ready to get review again. :)

tests/beacon/utils/test_random.py
```diff
@@ -0,0 +1,33 @@
...
+    [
+        (
+            [i for i in range(100)],
+            b"\x8d\x8dd\xb9\x84\x05*Bw\xf5\x87\xdd\x04nj\xee\x82\xfc9\xb9\x19\x82'x!#p];\xa5\x8e\xc9",  # noqa: E501
+            [87, 32, 76, 41, 61, 9, 44, 78, 70, 42, 8, 6, 79, 37, 4, 67, 3, 66, 55, 0, 69, 98, 38, 47, 20, 31, 48, 97, 34, 33, 96, 63, 91, 15, 1, 77, 39, 64, 58, 53, 82, 49, 26, 29, 5, 65, 19, 86, 74, 23, 46, 14, 10, 51, 85, 89, 57, 73, 71, 52, 88, 30, 16, 81, 24, 22, 36, 45, 12, 2, 62, 93, 18, 72, 83, 95, 13, 21, 27, 28, 80, 43, 68, 59, 94, 60, 99, 56, 92, 54, 25, 75, 7, 90, 50, 35, 11, 84, 17, 40],  # noqa: E501
+        ),
+        (
+            [i for i in range(12)],
+            b"&P\x18'\x91\xf6\x8f\xc8;<\xb8\x8f\x1d\x92?\xad4\xb5\xb8\x0b\xca\x8a'\x8e\xae'\xcf7\xa3(\xb2\x8d",  # noqa: E501
+            [7, 3, 2, 5, 11, 9, 1, 0, 4, 6, 10, 8],
```
hwwhww: Temporary cross-client test cases: https://github.com/ethereum/beacon_chain/issues/131

eth/beacon/helpers.py
pipermerriam: If you know all possible b values, what about doing a binary search across `[1, 64]`?  That should give you a loop which is guaranteed to terminate and fewer computation steps.

pipermerriam: Also, I don't care which method you go with at this point.  Optimization rabbit hole is deep and it's fine to just pick an approach and move on.

tests/beacon/test_helpers.py
```diff
@@ -0,0 +1,384 @@
...
+    blocks, recent_block_hashes = generate_mock_recent_block_hashes(
+        genesis_block,
+        current_block_slot_number,
+        cycle_length,
+    )
+
+    block = blocks[current_block_slot_number]
+    attestation = AttestationRecord(**sample_attestation_record_params)
+    oblique_parent_hashes = [b'\x77' * 32]
+    attestation = attestation.copy(
```
pipermerriam: *nitpick*

I tend to try and avoid overwriting variables.  If a variable *changes* during the scope of a function that introduces cognitive load to *track* what state it should be in.  Nothing structurally wrong with doing it but if we avoid it we end up with an overall easier to read codebase.

A pattern that you can use for this case would be to combine this with the line above to bet:

```python
oblique_parent_hashes = [b'\x77' * 32]
attestation = AttestationRecord(**sample_attestation_record_params).copy(
    slot=10,
    oblique_parent_hashes=oblique_parent_hashes,
)
```

eth/beacon/config.py
```diff
@@ -49,16 +49,16 @@ def generate_config(*,
                     sqrt_e_drop_time: int=SQRT_E_DROP_TIME) -> BeaconConfig:
 
     return BeaconConfig(
```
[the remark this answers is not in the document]
hwwhww: Oops, sorry I misunderstood your previous comment. Fixed!

hwwhww:
Once again, learned a lot and big thanks for the guide and feedback ! 😊

eth/beacon/helpers.py
hwwhww: > what about doing a binary search across [1, 64]? 

Good point. Will construct and benchmark it later off this PR.
I personally think it's interesting to explore the solutions. 🚂But also keep the schedule in mind.

PR #1342, opened 2018-10-02
Rename ChainConfig to TrinityConfig

pipermerriam:
### What was wrong?

As part of https://github.com/ethereum/py-evm/pull/1299 I found that the logic within the `ChainConfig` was getting too complex.  

### How was it fixed?

To combat this, I decided to split the config into two parts.

- The `ChainConfig` which **only** handles configuration of the `Chain` object.
- The `TrinityConfig` which handles everything else.

This pull request does step 1 in this, renaming the existing `ChainConfig` to `TrinityConfig`.  A subsequent PR will introduce the new `ChainConfig`.

#### Cute Animal Picture

squirrel-and-pumpkin-002

### DIFF
--- a/scripts/db-shell.py
+++ b/scripts/db-shell.py
@@ -14,7 +14,7 @@ from eth.chains.ropsten import ROPSTEN_NETWORK_ID
 from eth.db.chain import ChainDB
 from eth.db.backends.level import LevelDB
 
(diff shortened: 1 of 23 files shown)


carver:
That p2p test has been flakier recently.

PR #1343, opened 2018-10-02
Fix mixup with chain_id and network_id

pipermerriam:
### What was wrong?

The `eth.chains.base.Chain` class had a `network_id` property.  This is *wrong* since `network_id` is *above* the EVM level.  This should have been `chain_id`

### How was it fixed?

- migrated `MAINNET_NEWORK_ID` and `ROPSTEN_...` to live under `trinity.constants`
- migrated `DEFAULT_PREFERRED_NODES` to also live under `trinity` to prevent some weird imports.
- added `MAINNET_CHAIN_ID` and `ROPSTEN_CHAIN_ID` to `eth` module.
- updated the `net` JSON-RPC module to return the `chain_id` (currently served over the event bus by the `Node` class)
  - In a subsequent PR I'd like to move this out of the `Node` class into something stand-alone which serves up the various config values like this.  still mentally fleshing this out.

#### Cute Animal Picture

 
squirrel-in-a-pumpkin

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -764,8 +764,6 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
         self.connected_nodes: Dict[Node, BasePeer] = {}
         self._subscribers: List[PeerSubscriber] = []
         self.event_bus = event_bus
-        if self.event_bus is not None:
-            self.run_task(self.handle_peer_count_requests())
 
     async def handle_peer_count_requests(self) -> None:
         async def f() -> None:
@@ -856,7 +854,9 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
     async def _run(self) -> None:
         # FIXME: PeerPool should probably no longer be a BaseService, but for now we're keeping it
         # so in order to ensure we cancel all peers when we terminate.
-        self.run_task(self._periodically_report_stats())
+        if self.event_bus is not None:
+            self.run_daemon_task(self.handle_peer_count_requests())
+        self.run_daemon_task(self._periodically_report_stats())
         await self.cancel_token.wait()
 
     async def stop_all_peers(self) -> None:
--- a/trinity/nodes/base.py
+++ b/trinity/nodes/base.py
@@ -30,6 +30,11 @@ from trinity.utils.db_proxy import (
     create_db_manager
 )
 
+from .events import (
+    NetworkIdRequest,
+    NetworkIdResponse,
+)
+
 
 class Node(BaseService):
     """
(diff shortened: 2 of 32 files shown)


p2p/peer.py
```diff
@@ -856,7 +854,9 @@ def _add_peer(self,
     async def _run(self) -> None:
         # FIXME: PeerPool should probably no longer be a BaseService, but for now we're keeping it
         # so in order to ensure we cancel all peers when we terminate.
-        self.run_task(self._periodically_report_stats())
+        if self.event_bus is not None:
+            self.run_daemon_task(self.handle_peer_count_requests())
+        self.run_daemon_task(self._periodically_report_stats())
```
pipermerriam: This is potentially not a good idea.  It'll cause the node to crash if peer stats reporting crashes.

p2p/peer.py
```diff
@@ -856,7 +854,9 @@ def _add_peer(self,
     async def _run(self) -> None:
         # FIXME: PeerPool should probably no longer be a BaseService, but for now we're keeping it
         # so in order to ensure we cancel all peers when we terminate.
-        self.run_task(self._periodically_report_stats())
+        if self.event_bus is not None:
+            self.run_daemon_task(self.handle_peer_count_requests())
```
pipermerriam: @cburgdorf can you :+1: / :-1: this change.  It moves the task into the `_run` method so that it only starts when the peer pool has started as well as making it a deamon task which will crash the whole trinity process if it fails.

cburgdorf: Yes, that makes sense. :+1:  That said, I think this will move soon into its own child service or something similar.

trinity/nodes/base.py
```diff
@@ -46,6 +51,25 @@ def __init__(self, plugin_manager: PluginManager, trinity_config: TrinityConfig)
         self._headerdb = self._db_manager.get_headerdb()  # type: ignore
 
         self._jsonrpc_ipc_path: Path = trinity_config.jsonrpc_ipc_path
+        self._network_id = trinity_config.network_id
+
+        self.event_bus = plugin_manager.event_bus_endpoint
+
+    async def handle_network_id_requests(self) -> None:
+        async def f() -> None:
+            # FIXME: There must be a way to cancel event_bus.stream() when our token is triggered,
```
cburgdorf: nitpick: The comment could be redirected to mention that this hack will go away as soon as https://github.com/ethereum/lahja/pull/14 is included in the next release

pipermerriam: Will cleanup independently

PR #1347, opened 2018-10-02
Piper/re organize trinity chain module

pipermerriam:
> built on top of https://github.com/ethereum/py-evm/pull/1343 so ignore the first commit or go review that and I'll get this rebased

### What was wrong?

While working on #1299 it became apparent that the way some of the trinity modules were organized was not correct.  This was based on finding places where we ended up with circular imports, or having to do odd import hoop jumping to make things work.

### How was it fixed?

Re-organized things.  New modules are:

- `trinity.initialization`
- `trinity.db.manager`
- `trinity.utils.mp` now has the multiprocessing exception data structures.
- moved some constants around.

#### Cute Animal Picture

squirrel-steals-carved-pumpkin-max-ellis-2


pipermerriam:
Ignore this for now.  Still sorting out test failures.

pipermerriam:
Ok, looks like it's going to go green, :+1: ready for review.

tests/trinity/core/chain-management/test_initialize_database.py
```diff
@@ -1,32 +0,0 @@
-import pytest
```
carver: Were these test files supposed to be moved instead of deleted?

trinity/constants.py
```diff
@@ -11,18 +11,20 @@
...
 SYNC_FULL = 'full'
 SYNC_LIGHT = 'light'
 
+# lahja endpoint names
 MAIN_EVENTBUS_ENDPOINT = 'main'
 NETWORKING_EVENTBUS_ENDPOINT = 'networking'
 
-MAXIMUM_OBJECT_MEMORY_BYTES = 10000000
-
+# Network IDs: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-155.md
```
carver: This link is to the list of chain ids. (notably, Ethereum Classic is on network ID 1 and chain ID 61)

I don't know of a better list of network IDs than this: https://ethereum.stackexchange.com/a/17101/1461

pipermerriam:
https://github.com/ethereum/py-evm/issues/1352

pipermerriam:
Another missing test that I previously assumed `ethereum/tests` covered.

Issue #1298, opened 2018-09-15
DecryptionError escaping


cburgdorf:
### What is wrong?

```
 <Node(0xaa36fdf33dd030378a0168efe6ed7d5cc587fafa3cdd375854fe735a2e11ea3650ba29644e2db48368c46e1f60e716300ba49396cd63778bf8a818c09bded46f@13.93.211.84:30303)>, <Node(0x78de8a0916848093c73790ead81d1928bec737d565119932b98c6b100d944b7a95e94f847f689fc723399d2e31129d182f7ef3863f2b4c820abbf3ab2722344d@191.235.84.50:30303)>, <Node(0x158f8aab45f6d19c6cbf4a089c2670541a8da11978a2f90dbf6a502a4a3bab80d288afdbeb7ec0ef6d92de563767f3b1ea9e8e334ca711e9f8e2df5a0385e8e6@13.75.154.138:30303)>, <Node(0x1118980bf48b0a3640bdba04e0fe78b1add18e1cd99bf22d53daac1fd9972ad650df52176e7c7d89d1114cfef2bc23a2959aa54998a46afcf7d91809f0855082@52.74.57.123:30303)>)\n'
b'    INFO  09-13 21:24:44     service  Starting fast-sync; current head: #0\n'
b'   ERROR  09-13 21:24:51        peer  Unexpected error during auth/p2p handshake with <Node(0x4aaef89859ce7474eb3f3601b2118d7f376029e0261135c9bcb2ab6371e3c47e53001976cffa43587ef5d6021667443f047be0736989042523ec13c5f1eabe42@47.100.184.138:30303)>\n'
b'Traceback (most recent call last):\n'
b'  File "/home/circleci/repo/p2p/peer.py", line 919, in connect\n'
b'    self.cancel_token))\n'
b'  File "/home/circleci/repo/p2p/cancellable.py", line 19, in wait\n'
b'    return await self.wait_first(awaitable, token=token, timeout=timeout)\n'
b'  File "/home/circleci/repo/p2p/cancellable.py", line 41, in wait_first\n'
b'    return await token_chain.cancellable_wait(*awaitables, timeout=timeout)\n'
b'  File "/home/circleci/repo/.tox/py36-trinity-integration/lib/python3.6/site-packages/cancel_token/token.py", line 152, in cancellable_wait\n'
b'    return done.pop().result()\n'
b'  File "/home/circleci/repo/p2p/peer.py", line 160, in handshake\n'
[...]

PR #1350, opened 2018-10-03
Solve DecryptionError Escaping

Bhargavasomu:
### What was wrong?
The DecryptionError needs to be escaped for Malformed Messages.

### How was it fixed?
Calling both the decrypt_body and decrypt_header in try/except blocks.

#### Cute Animal Picture


pipermerriam:
link: https://github.com/ethereum/py-evm/issues/1298

PR #1351, opened 2018-10-03
Solves DecryptionError Escaping

Bhargavasomu:
### What was wrong?
The DecryptionError needs to be escaped for Malformed Messages.

### How was it fixed?
Calling both the decrypt_body and decrypt_header in try/except blocks.

#### Cute Animal Picture

### DIFF
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -385,13 +385,21 @@ class BasePeer(BaseService):
 
     async def read_msg(self) -> Tuple[protocol.Command, protocol.PayloadType]:
         header_data = await self.read(HEADER_LEN + MAC_LEN)
-        header = self.decrypt_header(header_data)
+        try:
+            header = self.decrypt_header(header_data)
+        except DecryptionError as err:
+            self.logger.debug("Error in Decrypting message from peer")
+            raise MalformedMessage from err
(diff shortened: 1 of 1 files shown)


p2p/peer.py
```diff
@@ -385,13 +385,21 @@ def is_closing(self) -> bool:
...
         frame_size = self.get_frame_size(header)
         # The frame_size specified in the header does not include the padding to 16-byte boundary,
         # so need to do this here to ensure we read all the frame's data.
         read_size = roundup_16(frame_size)
         frame_data = await self.read(read_size + MAC_LEN)
-        msg = self.decrypt_body(frame_data, frame_size)
+        try:
+            msg = self.decrypt_body(frame_data, frame_size)
+        except DecryptionError as err:
+            self.logger.debug("Error in Decrypting message from peer")
```
pipermerriam: Can you update both of these debug statements to include contextual information about the peer and exception message like is done here: https://github.com/ethereum/py-evm/pull/1351/files#diff-0c5a90ea9349bdb0b8d152d6c4b24583R412

Bhargavasomu: Yaa made the Debugging statement so that it includes the peer, error and the header or the body data respectively. Please review the latest PR.

PR #1353, opened 2018-10-04
Separate Peer Message Handling from Syncing

carver:
### What was wrong?

This is a precursor to skeleton header syncing. I found it confusing to have the syncer be in charge of handling incoming requests from peers. I'm not sure anything is strictly wrong with it, although there is theoretically a very short period when switching from fast sync -> state sync or state -> regular, where we aren't responding to inbound peer requests. Most important to me, I just found it harder to read/parse what was going on.

### How was it fixed?

Key changes:
- New service (`EthRequestServer` and `LightRequestServer`) which handle incoming requests for data from other peers , like `GetReceipts`, `GetNodeData`, etc
- Strip out all command subscriptions from the chain syncer classes, using the new tip monitor to find out when a new block/peer arrives
- The core trinity server now perpetually runs a `RequestServer` from launch.

I'm not in :heart: with the name `RequestServer` for the service that listens to all incoming requests and responds appropriately. Server is used a lot, and request can mean outgoing requests too. Any ideas for better names? Maybe:
- `DataServer` (with corresponding `ETHDataServer` and `LESDataServer`)
- `InboundRequestServer`
- `PeerRequestServer`
- `DataRequestServer`

Names I don't like:
- `RequestHandler` -- there's already a handler, which does the actual work of collecting the data and responding. I like that "server" implies something actively listening for requests

#### Cute Animal Picture


trinity/p2p/tip_monitor.py
```diff
@@ -0,0 +1,123 @@
...
+            new_tip_event.set()
+
+    async def _run(self) -> None:
+        self.run_task(self._handle_msg_loop())
+        with self.subscribe(self._peer_pool):
+            await self.events.cancelled.wait()
+
+    @contextmanager
+    def _subscriber(self) -> Iterator[asyncio.Event]:
+        new_tip_event = asyncio.Event()
```
pipermerriam: You should go look into `asyncio.Condition` as I think it *might* fit this model better.  I had a hard time grocking it's intended use case but after some experimentation it appears to handle something like this well where there is a shared resource (the tip peer), something in charge of managing it (the `ChainTipMonitor`), and many subscribers that want to be notified when it's available.

Nothing wrong with this API though, just something you might investigate.

trinity/sync/common/chain.py
```diff
@@ -184,7 +148,7 @@ class PeerHeaderSyncer(BaseService):
     def __init__(self,
                  chain: AsyncChain,
                  db: AsyncHeaderDB,
-                 peer: HeaderRequestingPeer,
+                 peer: BaseChainPeer,
```
pipermerriam: You might pull these type changes out to a small PR to drop a few things from the diff.

trinity/p2p/tip_monitor.py
carver: I can't think of why the waiters would need exclusive access to the tip peer.  If I'm understanding `asyncio.Condition` right (a big *if*), it would add some overhead for acquiring and releasing locks at every new tip, which is unnecessary for this use case.

Returning the highest TD peer is mostly just a convenience -- every waiter currently asks for the highest td peer as soon as a new tip is found, so this saves them that lookup.

Technically, though, this class would be just as effective to simply `yield None` and then the waiter can decide what to do. Do you think that would be clearer?

pipermerriam: So I believe the difference is that you can re-use an `asyncio.Condition` which would allow the subscriber to grab the new head info and then re-wait the same `condition` until the next new tip data is received.  But yes, I'm pretty sure you're right about added overhead with the various locks, as well as whether we are even understating how `asyncio.Condition` is supposed to be used.

I'm fine with whatever, I think the current implementation is plenty clear (enough).  Mostly, I want to figure out how you're supposed to use that damn `asyncio.Condition` API because I continue to be perplexed by it.

trinity/server.py
```diff
@@ -344,6 +360,13 @@ def _make_syncer(self) -> LightChainSyncer:
             self.cancel_token,
         )
 
+    def _make_request_server(self) -> LightRequestServer:
+        return LightRequestServer(
+            self.headerdb,
+            cast(LESPeerPool, self.peer_pool),
```
pipermerriam: I'm surprised this `cast` is needed though I do remember there being some oddities with getting the types to be right with this.

trinity/protocol/common/servers.py
```diff
@@ -0,0 +1,150 @@
...
+        Identify the command, and react appropriately.
+        """
+        pass
+
+    @property
+    def msg_queue_maxsize(self) -> int:
+        # This is a rather arbitrary value, but when the sync is operating normally we never see
+        # the msg queue grow past a few hundred items, so this should be a reasonable limit for
+        # now.
+        return 2000
```
pipermerriam: This can be converted to a normal property.

trinity/protocol/eth/servers.py
```diff
@@ -0,0 +1,167 @@
...
+    MAX_BODIES_FETCH,
+    MAX_RECEIPTS_FETCH,
+    MAX_STATE_FETCH,
+)
+from trinity.protocol.eth.requests import HeaderRequest as ETHHeaderRequest
+from trinity.rlp.block_body import BlockBody
+
+
+class ETHPeerRequestHandler(BasePeerRequestHandler):
+    def __init__(self, db: AsyncChainDB, logger: TraceLogger, token: CancelToken) -> None:
```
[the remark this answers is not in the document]
carver: This is more inherited code, but I suspect the idea was that this handler isn't "big enough" to justify its own logger (like no one would want to tweak the log level of only this handler). Also, since it's not a service, it doesn't get one for free. Just speculating though. I don't have any strong feelings on how it should be.

trinity/protocol/common/servers.py
```diff
@@ -0,0 +1,150 @@
...
+                                             request: BaseHeaderRequest,
+                                             ) -> Tuple[BlockNumber, ...]:
+        """
+        Generate the block numbers for a given `HeaderRequest`.
+        """
+        if isinstance(request.block_number_or_hash, bytes):
+            header = await self.wait(
+                self.db.coro_get_block_header_by_hash(cast(Hash32, request.block_number_or_hash)))
+            return request.generate_block_numbers(header.block_number)
+        elif isinstance(request.block_number_or_hash, int):
```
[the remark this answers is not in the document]
carver: This is what happens when I try:

```
In [1]: from eth_typing import BlockNumber

In [2]: isinstance(3, BlockNumber)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-2-6a4cdd0b995b> in <module>()
----> 1 isinstance(3, BlockNumber)

TypeError: isinstance() arg 2 must be a type or tuple of types

In [3]: type(BlockNumber)
Out[3]: function
```

Did you mean something else?

trinity/server.py
carver: Found a way around it by making the `Server` into `Generic`

trinity/protocol/eth/servers.py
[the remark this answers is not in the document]
pipermerriam: I'm generally :-1: on this approach.  I have in a semi-anti-pattern bucket simply because it makes testing harder.

trinity/protocol/common/servers.py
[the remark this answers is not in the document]
pipermerriam: I think I must just be making things up.

PR #1348, opened 2018-10-02
added default server URL for mainnet and ropsten, improved help message

evgeniuz:
### What was wrong?

Users were using ethstats server URL without path. Fixes #1337.

### How was it fixed?

Added example with path in `--help` output, added default server URLs for mainnet and ropsten.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #1354, opened 2018-10-04
Trinity ethstats plugin: fix CI breakage due to constant having moved

veox:
### What was wrong?

PR #1343 was merged before (20 hours prior to) PR #1348, but the latter didn't pick up the changes for some reason, and hasn't been run against actual `master` at the moment of merge.

In other words, PR #1348 had all CircleCI integration check-boxes green prior to @cburgdorf squash-merging it, but that was against an old commit.

### How was it fixed?

So far, identified one minor change that caused breaks - network ID constants being moved to a different module.

### Cute Animal Picture

It's a crash!..

Source: [Soggydan Benenovitch on flickr](https://www.flickr.com/photos/soggydan/6092175826/in/pool-62721506@N00)

### DIFF
--- a/trinity/plugins/builtin/ethstats/plugin.py
+++ b/trinity/plugins/builtin/ethstats/plugin.py
@@ -7,8 +7,10 @@ from argparse import (
     _SubParsersAction,
 )
 
-from eth.chains.mainnet import MAINNET_NETWORK_ID
-from eth.chains.ropsten import ROPSTEN_NETWORK_ID
+from trinity.constants import (
+    MAINNET_NETWORK_ID,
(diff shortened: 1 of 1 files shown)


carver:
It looks like the last poke re-ran all tests. One way to poke just the one job is "Rerun job with SSH", even with the 10 minute SSH timeout, it's usually faster than re-running the whole workflow.

veox:
It seems that this was not the only breakage; the `ethstats` plugin is no longer able to communicate with the remote server.

I've done a `git bisect`, and that points the regression being PR #1343.

Will open an issue with details if I can't figure this out.

PR #1357, opened 2018-10-04
Remove Unnecessary Lazy Extension Code

Bhargavasomu:
### What was wrong?
The If Clause code will never be executed and `validate_lte` would replace its functionality always. Hence need to delete this lazy extension.

### How was it fixed?
The If Clause is to be deleted instead of the `validate_lte`, because if that was kept then we would be increasing memory without charging gas for the increased memory.

#### Cute Animal Picture

### DIFF
--- a/eth/vm/memory.py
+++ b/eth/vm/memory.py
@@ -48,12 +48,6 @@ class Memory(object):
             validate_length(value, length=size)
             validate_lte(start_position + size, maximum=len(self))
 
-            if len(self._bytes) < start_position + size:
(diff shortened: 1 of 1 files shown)


Bhargavasomu:
@pipermerriam please refer this issue. Yaa will start using Markdown wherever possible from now. Glad to be part of the community

Bhargavasomu:
@pipermerriam I have removed the lazy extension and everything seems fine to me. Could you review this and merge this?

PR #1358, opened 2018-10-04
Add `BeaconChainDB`

hwwhww:
### What was wrong?
#1339 - part 2.1: Introduce `eth.beacon.db.chain.BeaconChainDB` 

### How was it fixed?
It’s WIP! But I hope @pipermerriam & co can take a look at the architecture of `BeaconChainDB` earlier. 

Now I basically just combined / copy-pasted `eth.db.header.HeaderDB` + `eth.db.chain.ChainDB` because:
1. There’s no `Header` in beacon block.
2. BeaconBlock fields names and serialization are and will be so different from PoW `BaseBlock`, so I don’t think it makes sense to make `BaseBeaconBlock` inherit `BaseBlock`. Even `slot_number` is defined as `int64`, not as same as the serial number in the PoW chain.

I see I repeated a lot... but, do you think it’s acceptable? Or some significant fundamental changes needed?

Thanks!

#### Cute Animal Picture

alpacas


eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+class BaseBeaconChainDB(ABC):
+    db = None  # type: BaseAtomicDB
+
+    # FIXME: Define more abstractmethod
+
+    @abstractmethod
+    def __init__(self, db: BaseAtomicDB) -> None:
+        raise NotImplementedError("BeaconChainDB classes must implement this method")
```
pipermerriam: We've shifted to just putting `pass` statements here since it's actually impossible to instantiate an `ABC` class without overriding this method, thus it's impossible to actually hit this `raise` statement.

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+
+    def persist_block(self,
+                      block: BaseBeaconBlock) -> Tuple[Tuple[bytes, ...], Tuple[bytes, ...]]:
+        """
+        Persist the given block.
+        """
+        with self.db.atomic_batch() as db:
+            return self._persist_block(db, block)
+
+    @classmethod
```
pipermerriam: What's the motivation for this being a `classmethod`?

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+            canonical_head_hash = db[SchemaV1.make_canonical_head_hash_lookup_key()]
+        except KeyError:
+            raise CanonicalHeadNotFound("No canonical head set for this chain")
+        return cls._get_block_by_hash(db, canonical_head_hash)
+
+    def get_block_by_hash(self, block_hash: Hash32) -> BaseBeaconBlock:
+        return self._get_block_by_hash(self.db, block_hash)
+
+    @staticmethod
+    def _get_block_by_hash(db: BaseDB, block_hash: Hash32) -> BaseBeaconBlock:
```
pipermerriam: Curious what the justification/reason is for this pattern of implementing all of these as separate methods?  My initial reaction is :-1: but if there's a reason I'd like to hear it.

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+        """
+        NOTE: copied from ``eth.db.header.HeaderDB._header_exists``,
+        and changed "header" to "block".
+        """
+        validate_word(block_hash, title="Block Hash")
+        return block_hash in db
+
+    def persist_block_chain(
+        self,
+        headers: Iterable[BaseBeaconBlock]
```
pipermerriam: looks like this name didn't get updated to `blocks`.

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+    def _persist_block_chain(
+            cls,
+            db: BaseDB,
+            blocks: Iterable[BaseBeaconBlock]
+    ) -> Tuple[Tuple[BaseBeaconBlock, ...], Tuple[BaseBeaconBlock, ...]]:
+        """
+
+        NOTE: copied from ``eth.db.header.HeaderDB._persist_header_chain``,
+        and changed "header" to "block".
+        NOTE: moodified a lot
```
pipermerriam: Probably worth dropping these comments in cases where the modification sufficiently departs from the original one...

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+                        )
+                    )
+
+            is_genesis = first_block.parent_hash == GENESIS_PARENT_HASH
+            if not is_genesis and not cls._block_exists(db, first_block.parent_hash):
+                raise ParentNotFound(
+                    "Cannot persist block ({}) with unknown parent ({})".format(
+                        encode_hex(first_block.hash), encode_hex(first_block.parent_hash)))
+
+            score = 0 if is_genesis else cls._get_score(db, first_block.parent_hash)
```
pipermerriam: Preference would be to expand this to multiline `if/else` as I think it is easier to read.

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+
+        for block in blocks:
+            db.set(
+                block.hash,
+                rlp.encode(block),
+            )
+
+            # FIXME: It's a stub before we implement fork choice rule
```
pipermerriam: I think we have two conventions floating around, `TODO` and `FIXME`.  Anyone want to make a case for one-or-the-other?  It'd be nice to be able to do a single `grep TODO` or `grep FIXME` to find all of these.

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+
+        try:
+            previous_canonical_head = cls._get_canonical_head(db).hash
+            head_score = cls._get_score(db, previous_canonical_head)
+        except CanonicalHeadNotFound:
+            (
+                new_canonical_blocks,
+                old_canonical_blocks
+            ) = cls._set_as_canonical_chain_head(db, block.hash)
+        else:
```
pipermerriam: I think this block (the `try/except` above and the nested `if/else` below would be easier to read written as:

```python
try:
    ...
except CanonicalHeadNotFound:
    return cls._set_as_canonical_chain_head(...)

if score > head_score:
    return cls._set_as_canonical_chain_head(...)
else:
    return tuple(), tuple()
```

Removes a level of nesting and *mostly* decouples the two blocks of code.

eth/beacon/db/chain.py
```diff
@@ -0,0 +1,450 @@
...
+
+        for block in new_canonical_blocks:
+            cls._add_block_slot_to_hash_lookup(db, block)
+
+        db.set(SchemaV1.make_canonical_head_hash_lookup_key(), block.hash)
+
+        return new_canonical_blocks, tuple(old_canonical_blocks)
+
+    @classmethod
+    @to_tuple
```
pipermerriam: I think the two approaches are equivalent but I think generally `classmethod` belongs *closer* to the function definition and `to_tuple` should be above it.

eth/beacon/db/chain.py
hwwhww: My fault, according to `BaseHeaderDB(ABC)`, it might be:
```python
    @abstractmethod
    def __init__(self, db: BaseAtomicDB) -> None:
        self.db = db
```

eth/beacon/db/chain.py
hwwhww: Copied our conversation:

> Piper: I think we did it so that we could implement some of them as staticmethod or classmethod
which the type system won't allow you to override those things cleanly

eth/beacon/db/chain.py
hwwhww: 👍 these comments were just for me to track the source in case you want me to deduplicate the code.

eth/beacon/db/chain.py
hwwhww: Found this [PEP 350 -- Codetags](https://www.python.org/dev/peps/pep-0350/#mnemonics):
* To do: Informal tasks/features that are pending completion.
* Fix me: Areas of problematic or ugly code needing refactoring or cleanup.

In the case of this PR, these FIXMEs should be changed to TODOs. I'll fix it. 👍

eth/beacon/db/chain.py
hwwhww: I tried:
```python
@to_tuple
@classmethod
def _find_new_ancestors(...)
```
then got error: `TypeError: 'classmethod' object is not callable`. It looks like we have to go through `to_tuple` and then `@classmethod` when calling it.

eth/beacon/db/chain.py
hwwhww: oops sorry Jason! Wrongly tagged! Asking @pipermerriam

hwwhww: Updates:
Sorry, should be `headers[-1]` and it's not a bug because `header` will be set to `headers[-1]` in the previous for loop automatically.

eth/beacon/validation.py
```diff
@@ -0,0 +1,11 @@
+from eth.validation import (
+    validate_gte,
+    validate_lte,
+    validate_is_integer,
+)
+
+
+def validate_slot(slot):
```
pipermerriam: missing type (which should be `Any`) for validation functions.

eth/validation.py
```diff
@@ -198,7 +198,7 @@ def validate_unique(values, title="Value"):
         )
 
 
-def validate_block_number(block_number, title="Block Number"):
+def validate_block_number(block_number):
```
pipermerriam: Instead of removing the `title` from here, can you change the inner calls to pass the `title` down?

eth/beacon/db/chain.py
carver: The primary motivation for me is: we want to prevent accidental reads/writes of `self.db` in `_persist_block()`, which would break atomicity of `persist_block()`

eth/db/header.py
```diff
@@ -251,21 +254,12 @@ def _persist_header_chain(
...
-            else:
-                new_canonical_headers = tuple()
-                old_canonical_headers = tuple()
+            return cls._set_as_canonical_chain_head(db, header.hash)
 
-        return new_canonical_headers, old_canonical_headers
+        if score > head_score:
+            return cls._set_as_canonical_chain_head(db, header.hash)
+        else:
+            return tuple(), tuple()
```
carver: Nice, a much easier read.

PR #1361, opened 2018-10-04
Use BaseChainPeer instead of HeaderRequestingPeer

carver:
### What was wrong?

_I hadn't had a good tussle with mypy for a while, and I like punishment._

`HeaderRequestingPeer` was an old "poor coder's" version of `BaseChainPeer`

### How was it fixed?

_Rope-a-dope with mypy, plus a crucial assist from cburgdorf._

Replaced `HeaderRequestingPeer` with `BaseChainPeer`, and resolve the mypy fallout.

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/handlers.py
+++ b/trinity/protocol/common/handlers.py
@@ -1,11 +1,16 @@
 from abc import ABC, abstractmethod
 from typing import (
     Any,
+    Awaitable,
+    Callable,
     Dict,
     Iterator,
+    Tuple,
     Type,
+    TYPE_CHECKING,
 )
 
+from eth.rlp.headers import BlockHeader
 from p2p.peer import BasePeer
 
 from trinity.protocol.common.exchanges import (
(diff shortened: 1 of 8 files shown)


trinity/protocol/common/handlers.py
```diff
@@ -46,3 +51,16 @@ def get_stats(self) -> Dict[str, str]:
             for exchange
             in self
         }
+
+
+if TYPE_CHECKING:
+    from mypy_extensions import DefaultArg
+    BlockHeadersCallable = Callable[
+        [BaseExchangeHandler, int, int, DefaultArg(int, 'skip'), DefaultArg(int, 'reverse')],
```
pipermerriam: SUPER FANCY!

trinity/protocol/common/handlers.py
```diff
@@ -46,3 +51,16 @@ def get_stats(self) -> Dict[str, str]:
             for exchange
             in self
         }
+
+
+if TYPE_CHECKING:
+    from mypy_extensions import DefaultArg
```
cburgdorf: @carver why didn't we need to install `mypy_extensions` in order for this to work?

carver: IDK for sure, but it seems to come for free with `mypy`. I'm trusting that CI would have caught any problems. Since it doesn't matter in production, it basically only matters in CI and dev machines, I didn't put more thought into it.

PR #1363, opened 2018-10-05
Fix Minor Bugs in Tools

Bhargavasomu:
### What was wrong?
There are 2 issues with the function `normalize_bytes` in eth/tools/fixtures/normalization.py#L68
1) The `if condition` should be checking whether the `len(value)` is 0 first, and then move on to verifying `is_hex(value)`. The problem arises when `value = b''`
2) As per the `eth_util library`, `is_hex function` takes only strings as argument. So the bytes should be converted to string using value = value.decode('utf-8') in the beginning of the function.

### How was it fixed?
1) The `if condition` was reversed
2) `bytes` is converted to `string` using `decode('utf-8')` function

#### Cute Animal Picture

### DIFF
--- a/eth/tools/fixtures/normalization.py
+++ b/eth/tools/fixtures/normalization.py
@@ -24,6 +24,7 @@ from eth_utils import (
     is_hex,
     is_integer,
     is_string,
+    is_text,
     to_bytes,
     to_canonical_address,
     to_dict,
--- a/eth/tools/fixtures/normalization.py
+++ b/eth/tools/fixtures/normalization.py
@@ -66,13 +66,10 @@ def normalize_int(value):
 
 
 def normalize_bytes(value):
-    # type(value) maybe also bytes. But for `is_hex(value)` function
-    # value has to be only of the type string
-    value = value.decode('utf-8')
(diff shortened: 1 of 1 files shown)


eth/tools/fixtures/normalization.py
```diff
@@ -66,7 +66,10 @@ def normalize_int(value):
 
 
 def normalize_bytes(value):
-    if is_hex(value) or len(value) == 0:
+    # type(value) maybe also bytes. But for `is_hex(value)` function
+    # value has to be only of the type string
+    value = value.decode('utf-8')
+    if len(value) or is_hex(value) == 0:
```
pipermerriam: This isn't correct, the `== 0` didn't stay with the `len` check.

Probably best to update this to:

```python
if is_bytes(value):
    return b''
elif is_text(value) and is_hex(value):
    return decode_hex(value)
else:
    raise TypeError(...)
```

Bhargavasomu: Oops sry about that. Will make the changes ASAP

Bhargavasomu:
@pipermerriam I have made the necessary changes, please merge issue #1357

PR #1364, opened 2018-10-05
Use chain builder tools to cleanup benchmarks

Bhargavasomu:
### What was wrong?
The current benchmarks use an impromptu API to create the chains that are used for benchmarking.

### How was it fixed?
We now have some more solid `builder tools` to create chains. These tools allow the creation of chains using a declarative, fluent API which we do use already in several tests. These `builder tools` are now integrated in `Benchmarks` for building the chains.

#### Cute Animal Picture

### DIFF
--- a/scripts/benchmark/utils/chain_plumbing.py
+++ b/scripts/benchmark/utils/chain_plumbing.py
@@ -1,6 +1,9 @@
 from typing import (
+    Any,
+    Dict,
     Iterable,
-    Type
+    Tuple,
+    Type,
 )
 
 from eth_keys import (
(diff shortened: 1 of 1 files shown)


Bhargavasomu:
[
screenshot from 2018-10-06 00-39-04
](url)

Got this error in pylint, but everything seems to work. How can I get rid of this py36-lint error?

pipermerriam:
I suspect this is a deficiency with `mypy` + curried functions.  It probably makes sense to move away from the curry model and just do explicit functions as I think most everything in the builder can be fit into that model.

pipermerriam:
https://github.com/ethereum/py-evm/issues/1366 should fix this.

cburgdorf:
>I suspect this is a deficiency with mypy + curried functions

@pipermerriam It's not the currying per se but rather that the `curry` decorator from `cytoolz` does not come with type annotations and no type annotations seem to exist in typeshed for it. It should be easy to create our own `curry` decorator in `eth_utils`, enable distribution of type annotations for `eth_utils` and then use that.

Btw, the same applies to other decorators such as `to_tuple`. We currently errase type hints for functions that we apply something like `to_tuple` on. That will change as soon as we enable type hint distribution for `eth_utils` as per PEP561 ([ridiculously simple](https://github.com/ethereum/eth-typing/commit/b2fbd48d469fa6bf230ab8c2cd506e232336ea7d))

cburgdorf:
@Bhargavasomu I took that one over to resolve the mypy issue and apply a minor refactoring to allow benchmarks to pass in the genesis state rather than hard coding it within `get_chain`.

Thanks for doing the bulk of the work! :tada: 

I think the two remaining functions are so trivial that I'm tempted to say we should enhance the *builder* API to get rid of `chain_plumping.py` altogether. That said, I'm not entirely sure the creation of multiple chains that basically follow the same recipe but start at different VMs, fits into the *builder* pattern of that API.

Bhargavasomu:
@cburgdorf I am just curious of how the error of cannot call function of unknown type did not occur, even though the builder tools were not type hinted. Could you please explain how it got fixed?

cburgdorf:
>Could you please elaborate on this for my understanding?

Let me try! It has to do with the way these methods are defined on the `eth.builders.chain.api`

It is basically a class `API` that defines class members that point to these methods.

https://github.com/ethereum/py-evm/blob/c413ec9e1d3df0e0b91b27f00cdc6d160719c664/eth/tools/builder/chain/__init__.py#L39-L62
...

Then the API is exposed as `api`

https://github.com/ethereum/py-evm/blob/c413ec9e1d3df0e0b91b27f00cdc6d160719c664/eth/tools/builder/chain/__init__.py#L95

Somewhere in this process mypy loses some information about what these class members actually refer to, namely the functions that are defined standalone as:

https://github.com/ethereum/py-evm/blob/c413ec9e1d3df0e0b91b27f00cdc6d160719c664/eth/tools/builder/chain/builders.py#L47-L84
...

I guess, that may work with importing `api` as well if the entire *builders* module was typed but since it isn't mypy seems to be unable to infer what the types of these class members actually are.

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -70,56 +67,33 @@
...
-            ),
-        ])
-    )
-
-
-def get_all_chains() -> Iterable[MiningChain]:
+def get_all_chains(genesis_state: GenesisState=DEFAULT_GENESIS_STATE) -> Iterable[MiningChain]:
     for vm in ALL_VM:
-        chain = get_chain(vm)
+        chain = get_chain(vm, DEFAULT_GENESIS_STATE)
```
carver: I suppose this was intended to be `get_chain(vm, genesis_state)`.

cburgdorf: When did I become so sloppy... :see_no_evil: https://github.com/ethereum/py-evm/pull/1376

PR #1372, opened 2018-10-07
[WIP] Extend round trip API to GetBlockBodies LES protocol command

hoi:
### What was wrong?

 * [ ]  `GetBlockBodies -> BlockBodies`

### How was it fixed?

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/les/exchanges.py
+++ b/trinity/protocol/les/exchanges.py
@@ -11,20 +11,27 @@ from eth.rlp.headers import BlockHeader
 from trinity.protocol.common.exchanges import (
     BaseExchange,
 )
+# Q: What is the order to imports?  and why sometimes new line between imports and sometimes not?
+from trinity.protocol.common.types import BlockBodyBundles
+from trinity.rlp.block_body import BlockBody
 from trinity.utils.les import (
     gen_request_id,
 )
 
 from .normalizers import (
     BlockHeadersNormalizer,
+    GetBlockBodiesNormalizer,
 )
 from .requests import (
     GetBlockHeadersRequest,
+    GetBlockBodiesRequest,
 )
 from .trackers import (
     GetBlockHeadersTracker,
+    GetBlockBodiesTracker,
 )
 from .validators import (
+    GetBlockBodiesValidator,
     GetBlockHeadersValidator,
     match_payload_request_id,
 )
@@ -53,6 +60,28 @@ class GetBlockHeadersExchange(LESExchange[Tuple[BlockHeader, ...]]):
 
         command_args = original_request_args + (gen_request_id(),)
         request = self.request_class(*command_args)  # type: ignore
+
+        return await self.get_result(
+            request,
+            self._normalizer,
+            validator,
+            match_payload_request_id,
+            timeout,
+        )
+
+# Q: Where can I find the correct signature for this class?
+class GetBlockBodiesExchange(LESExchange[Tuple[BlockBody, ...]]):
+    _normalizer = GetBlockBodiesNormalizer()
+    request_class = GetBlockBodiesRequest
+    tracker_class = GetBlockBodiesTracker
+
+    async def __call__(self,  # type: ignore
+                       headers: Tuple[BlockHeader, ...],
+                       timeout: float = None) -> BlockBodyBundles:
+        validator = GetBlockBodiesValidator(headers)
+
+        block_hashes = tuple(header.hash for header in headers)
+        request = self.request_class(block_hashes, gen_request_id())
 
         return await self.get_result(
             request,
--- a/trinity/protocol/les/exchanges.py
+++ b/trinity/protocol/les/exchanges.py
@@ -11,9 +11,7 @@ from eth.rlp.headers import BlockHeader
 from trinity.protocol.common.exchanges import (
     BaseExchange,
 )
-# Q: What is the order to imports?  and why sometimes new line between imports and sometimes not?
 from trinity.protocol.common.types import BlockBodyBundles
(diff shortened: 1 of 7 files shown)


trinity/protocol/les/exchanges.py
```diff
@@ -11,20 +11,27 @@
 from trinity.protocol.common.exchanges import (
     BaseExchange,
 )
+# Q: What is the order to imports?  and why sometimes new line between imports and sometimes not?
```
pipermerriam: ordering is:

1. standard library
2. 3rd party
3. local library
4. local modules (these are all `from . import` style imports

Within each section we alphabetize and *normally* do whitespace breaks between libraries.

Within each of those sections

trinity/protocol/les/normalizers.py
```diff
@@ -8,13 +8,22 @@
 from eth.rlp.headers import BlockHeader
 
 from trinity.protocol.common.normalizers import BaseNormalizer
+from trinity.rlp.block_body import BlockBody
 
 TResult = TypeVar('TResult')
 LESNormalizer = BaseNormalizer[Dict[str, Any], TResult]
 
 
+# Q: Shouldn't this be named GetBlockHeadersNormalizer?
```
pipermerriam: No, because it normalizes the response which is a `Blockheaders` message (responding to a `GetBlockHeaders` request).

trinity/protocol/les/trackers.py
```diff
@@ -37,3 +39,21 @@ def _get_result_size(self, result: Tuple[BlockHeader, ...]) -> int:
...
         return len(result)
+
+
+BaseGetBlockBodiesTracker = BasePerformanceTracker[
+    GetBlockBodiesRequest,
+    Tuple[BlockBody, ...],
+]
+
+
+# Q: Where can I find the signature for this class?
```
pipermerriam: I don't know what you mean by *signature*

trinity/protocol/les/normalizers.py
```diff
@@ -8,13 +8,22 @@
...
 class BlockHeadersNormalizer(LESNormalizer[Tuple[BlockHeader, ...]]):
     @staticmethod
     def normalize_result(message: Dict[str, Any]) -> Tuple[BlockHeader, ...]:
         result = message['headers']
         return result
+
+
+class GetBlockBodiesNormalizer(LESNormalizer[Tuple[BlockBody, ...]]):
```
carver: Yeah, so for the same reason, the `Get` should be dropped here. This normalizer only acts on the response value, which is sent with the `BlockBodies` command.

trinity/protocol/les/normalizers.py
```diff
@@ -8,13 +8,22 @@
...
+    @staticmethod
+    def normalize_result(message: Dict[str, Any]) -> Tuple[BlockBody, ...]:
+        result = message['bodies']
+        return result
```
carver: I'm fairly sure that we'll want to return block body bundles here, similar to how the eth version does:
https://github.com/ethereum/py-evm/blob/9ba2c1fae5de83ae3fe55688197630e65262f1b9/trinity/protocol/eth/normalizers.py#L48-L57

but everywhere `msg` is in this snippet, using `result` from the PR.

This is going to change the type signature of the the exchange, and so will affect the validator, etc.

trinity/protocol/les/trackers.py
carver: I'm not sure either, feel free to hop on chat, since it may require some back-and-forth to resolve.

carver:
Also, maybe best to reduce the scope on each PR to just one exchange, in this case: `GetBlockBodiesExchange`. (I only mention it because the PR summary includes the list of all the different exchanges)

carver:
... and excellent cute animal picture :100:

trinity/protocol/les/normalizers.py
hoi: ah! makes sense

trinity/protocol/les/trackers.py
hoi: ah, i just meant the input arguments and the expected return

hoi:
> Also, it's probably best to reduce the scope on each PR to just one exchange, in this case: `GetBlockBodiesExchange`. (I only mention it because the PR summary includes the list of all the different exchanges)

I agree.  That was my intention too.  Thanks!

trinity/protocol/les/exchanges.py
hoi: ah, great!  I wonder if my ide can set this order.  I've done it in the past for my IntelliJ.

trinity/protocol/les/normalizers.py
hoi: Ah got it

carver:
Here's a suggestion for how to run a local smoke test to make sure everything is working as expected. Use the new API you created to replace this code:

https://github.com/ethereum/py-evm/blob/7d0f1888e5b43189bc82852e16dc38f8e58ea2b0/trinity/sync/light/service.py#L170-L172

With (roughly):

```py
 block_body_bundles = peer.requests.get_block_bodies((block_hash, ))
```

Then you'll extract the block and return it.

---

After making that change, you should be able to run a light trinity node and use web3.py to call `w3.eth.getBlock('latest')` and get an expected block. (probably on ropsten because mainnet light servers are scarce) If that still works as expected, then the smoke test works well.

---

Note that the above change should actually be made permanently in trinity. It doesn't have to be made in *this* PR, but it would be okay to if you want.

carver:
Go ahead and rebase this, fix lint, and force-push :+1:

pipermerriam:

PR #1373, opened 2018-10-08
Add `BeaconStateMachine` outline

hwwhww:
### What was wrong?
#1339 - part 3: Introduce `eth.beacon.state_machines.base.BaseStateMachine(Configurable)`

### How was it fixed?

1. **Note: the spec changed recently, so it doesn’t match [the current spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/beacon-chain.md)! This PR is closer to [this version](https://github.com/ethereum/eth2.0-specs/blob/0001b7b9de2bf87ff267547acdb99788cf9b463c/specs/casper_sharding_v2.1.md).**
2. State storage
	1. The beacon chain states are quite different from EVM chain:
		* The crystallized state recalculation, which happens only if `block.slot_number >= last_state_recalc + CYCLE_LENGTH`, and affects the `CrystallizedState` and `ActiveState`
		* The per-block processing, which happens every block (if during a crystallized state recalculation block, it happens after the crystallized state recalculation), and affects the `ActiveState` only
		* Both states are being serialized with RLP (or SSZ later). No tree data structure currently, but **possible will be**. **Now the state root is actually just hashed state.**
	2. The `CrystallizedState` is recalculated every cycle (~17 minutes with the latest spec). Now my strategy is simply store it by:
		1. `crystallized_state_root: Hash32` -> `crystallized_state: CystallizedState`
		2. `last_state_recalc` slot number ->  `crystallized_state_root: Hash32` for canonical chain
		3. If there is forking (e.g., the same slot number conflicted), records the old state root to `deletable_state_roots` DB storage. Could be clean up periodically or when some blocks are finalized.
	3. The `ActiveState` is recalculated every block (~16 seconds). 
		1. The `ActiveState` will be backed up when `CrystallizedState` is persisted.
		2. Since the `CYCLE_LENGTH` is 64 slots, with the backup ActiveState + recent blocks, it can be reproduced.
	4. Not a perfect solution (high disk I/O!), but since the spec hasn't been finalized, the data structure might be changed soon. (In fact, `ActiveState` grows to 4 fields now!) I think we can leave the storage optimization later.
3. Config
	1. After another look, I think if we want to mock important parameters in tests,  it can be configured more like `opcodes` in PoW chain, not `constants`. But now it’s pretty messy.
4. StateMachine
	1. Copied the structure of VM.
	2. Assumed that the DB already has essential data.

### TODO
1. Now there are two `BeaconConfig`s, needs to be merged.
2. Check and refactor `from_old_active_and_blocks` thing.

#### Cute Animal Picture

ohfu


eth/beacon/db/schema.py
```diff
@@ -35,3 +59,18 @@ def make_block_slot_to_hash_lookup_key(slot: int) -> bytes:
     @staticmethod
     def make_block_hash_to_score_lookup_key(block_hash: Hash32) -> bytes:
         return b'beacon:block-hash-to-score:%s' % block_hash
+
+    #
+    # States
+    #
+    @staticmethod
+    def make_slot_to_crystallized_state_lookup_key(slot: int) -> bytes:
+        return b'beacon:slot-to-crystallized-state:%d' % slot
```
pipermerriam: still probably want the `v1` namespace to be present somewhere in this string.

eth/beacon/state_machines/base.py
```diff
@@ -0,0 +1,185 @@
...
+                self.chaindb,
+                depth,
+                start_slot=old_active_state_slot
+            )
+            blocks = blocks[::-1]
+
+            self._active_state = self.get_active_state_class().from_old_active_and_blocks(
+                old_active_state,
+                blocks,
+                self.config.CYCLE_LENGTH * 2,
```
pipermerriam: It appears that this is the only value that is accessed in `self.config`.  Does it make sense to just have this be a class level variable?

eth/beacon/helpers.py
```diff
@@ -268,11 +267,14 @@ def _get_shards_and_committees_for_shard_indices(
 
 
 @to_tuple
-def get_new_shuffling(seed: Hash32,
+def get_new_shuffling(*,
```
pipermerriam: :heart: 

I just deleted a comment suggesting that you adopt this exact pattern!

One thing to think about.  Invoking functions using keyword arguments has a runtime cost.  I suspect that for this function, that runtime cost is going to be trivial compared to the execution of the body of the functions, but for any function that needs to be highly performant, this should be taken into account.

eth/beacon/state_machines/base.py
```diff
@@ -74,7 +74,7 @@ def logger(self):
     @classmethod
     def get_block_class(cls) -> Type[BaseBeaconBlock]:
         """
-        Return the :class:`~eth.beacon.types.blocks.BeaconBlock` class that this
+        Returns the :class:`~eth.beacon.types.blocks.BeaconBlock` class that this
```
pipermerriam: I believe the original `Return` was the proper format we are using for docstrings (cc @cburgdorf)

eth/beacon/state_machines/base.py
```diff
@@ -83,25 +83,30 @@ def get_block_class(cls) -> Type[BaseBeaconBlock]:
...
     @to_tuple
-    def get_prev_blocks_to_slot(cls,
-                                last_block_hash: Hash32,
-                                chaindb: BaseBeaconChainDB,
-                                depth: int,
-                                start_slot: int) -> Iterable[BaseBeaconBlock]:
+    def get_prev_blocks(cls,
+                        last_block_hash: Hash32,
+                        chaindb: BaseBeaconChainDB,
+                        searching_depth: int,
```
pipermerriam: *nitpick*

I think `search_depth` is a more natural name for this.  And maybe even `max_search_depth` since this seems to be an upper bound.

eth/beacon/state_machines/base.py
```diff
@@ -83,25 +83,30 @@ def get_block_class(cls) -> Type[BaseBeaconBlock]:
...
+                        target_slot_number: int) -> Iterable[BaseBeaconBlock]:
+        """
+        Returns the previous blocks.
+
+        Since the slot numbers might be not continuous. The searching depth is ``searching_depth``
+        and the target slot depth is ``target_slot_number``.
```
pipermerriam: Suggested rephrasing.

```
Slot numbers are not guaranteed to be contiguous since it is possible for there
to be no block at a given slot.  The search is bounded by two parameters. 

- `max_search_depth` - The maximum number of slots below the slot of the block denoted by `last_block_hash` that we should search.
- `min_slot_depth` - The slot number for which we should not search any deeper if reached, though we do include the first block encountered below this slot number before terminating the search.
```

eth/beacon/state_machines/base.py
```diff
@@ -83,25 +83,30 @@ def get_block_class(cls) -> Type[BaseBeaconBlock]:
...
-    def get_prev_blocks_to_slot(cls,
-                                last_block_hash: Hash32,
-                                chaindb: BaseBeaconChainDB,
-                                depth: int,
-                                start_slot: int) -> Iterable[BaseBeaconBlock]:
+    def get_prev_blocks(cls,
+                        last_block_hash: Hash32,
+                        chaindb: BaseBeaconChainDB,
+                        searching_depth: int,
+                        target_slot_number: int) -> Iterable[BaseBeaconBlock]:
```
pipermerriam: Maybe `min_slot_number` since minimum is somewhat of an *optional* constraint and `target` is likely to be interpreted as a firm constraint.  With the term target, my intuition would say that this search would continue until it hit this slot number.

eth/beacon/state_machines/base.py
```diff
@@ -83,25 +83,30 @@ def get_block_class(cls) -> Type[BaseBeaconBlock]:
...
 
-        for _ in range(depth):
+        for _ in range(searching_depth):
             yield block
             try:
                 block = chaindb.get_block_by_hash(block.parent_hash)
             except (IndexError, BlockNotFound):
                 break
-            if block.slot_number < start_slot:
+            if block.slot_number < target_slot_number:
```
pipermerriam: I find it interesting and not intuitive that this search will include blocks which have a slot number lower than `target_slot_number`.  I assume that is part of the spec, but it might be something worth leaving an inline comment to be sure future readers of this code know that behavior is intentional.

eth/beacon/state_machines/base.py
```diff
@@ -144,39 +149,44 @@ def active_state(self) -> ActiveState:
...
+            else:
+                # Get recent blocks after last ActiveState backup.
+                searching_depth = self.config.CYCLE_LENGTH * 2
+                blocks = self.get_prev_blocks(
+                    last_block_hash=self.parent_block.hash,
+                    chaindb=self.chaindb,
+                    searching_depth=searching_depth,
+                    target_slot_number=backup_active_state_slot
+                )
+                blocks = blocks[::-1]
```
pipermerriam: I had to go check that this was reversing the `blocks`.  I think the idiomatic way to do this is to call `reversed()` on blocks, or better yet, to do `blocks = reversed(self.get_prev_blocks(...))` so that you don't have to define the `blocks` variable twice.

eth/beacon/state_machines/forks/serenity/active_states.py
```diff
@@ -1,14 +1,25 @@
...
+
 
 class SerenityActiveState(ActiveState):
     @classmethod
-    def from_old_active_and_blocks(cls, old_active_state, blocks, recent_block_hashes_length):
-        recent_block_hashes = old_active_state.recent_block_hashes
-        pending_attestations = old_active_state.pending_attestations
+    def from_backup_active_state_and_blocks(cls,
+                                            backup_active_state: ActiveState,
+                                            blocks: Sequence['BaseBeaconBlock']) -> ActiveState:
```
pipermerriam: :+1: to using `Sequence`.  I think we (and largely me) have been defining these as `Tuple` or `List` too often which isn't strictly necessary and doesn't provide much extra safety.

eth/beacon/state_machines/forks/serenity/configs.py
```diff
@@ -3,12 +3,11 @@
 from eth.beacon.state_machines.configs import BeaconConfig
 
 
-SENERITY_CONFIG = BeaconConfig(
```
pipermerriam: :rofl:

eth/beacon/helpers.py
hwwhww: *If* we move it into StateMachine, we can reduce some parameters by using  `self.config.CYCLE_LENGTH`, `self.config.min_committee_size`, `self.config.shard_count`.

eth/beacon/state_machines/base.py
hwwhww: I don't have any preferences! 😆 Searched for `eth` with this PR, there are 81 `Returns` and 64 `Return` now. I can change it back.

eth/beacon/state_machines/base.py
hwwhww: > I find it interesting and not intuitive that this search will include blocks which have a slot number lower than target_slot_number

Ohh, my goal is making this search return the blocks that **are greater than or equal to** `target_slot_number` (`min_slot_number`). It should return at least one block (the block that produced `backup_active_state`).

Long story version:

1. Startup: Setup empty states
2. Apply the genesis block (`B0`) and get the genesis states. Persist these states (`C0` and `A0`)
	* `recent_block_hashes: [ZERO_HASH32]  * (CYCLE_LENGTH * 2)`
4. Apply block `B1`, update active state (`A1`)
    ```python
    recent_block_hashes = get_new_recent_block_hashes(
            old_block_hashes=recent_block_hashes,
            parent_slot=B0.slot_number,
            current_slot=B1.slot_number,
            parent_hash=B0.hash
    )
    ```
	* `recent_block_hashes: [ZERO_HASH32] * (CYCLE_LENGTH * 2 - 1) + [B0.hash]`
5. Apply block `B2`, update active state (`A2`)
    ```python
    recent_block_hashes = get_new_recent_block_hashes(
            old_block_hashes=recent_block_hashes,
            parent_slot=B1.slot_number,
            current_slot=B2.slot_number,
            parent_hash=B1.hash
    )
    ```
	* `recent_block_hashes: [ZERO_HASH32] * (CYCLE_LENGTH * 2 - 2) + [B0.hash] + [B1.hash]`
6. Apply block `B3`, update active state (`A3`)
…
7. Apply block `B64`, update active state (`A64`) and crystallized state (`C64`), and also persist states.


Assume that we have the backup active state `A0` and want to create `B3’` (!= `B3`), so we have to reproduce `A2`. It requires the blocks starting from **the parent** of  “the first block after `backup_active_state_slot`”. This parent block should be the block that produced `A0`, so I set `min_slot_number=backup_active_state_slot`.

Sorry, it's long. Do you think it makes sense?

eth/beacon/state_machines/base.py
cburgdorf: Correct, we try to streamline docs to always be imperative present tense. I *believe* that is what most documentation uses but it is at least inline with what the official Python docs use.

Here's a rule of thumb. If this sounds correct, it is correct:

*If this method is called it will: `<doc-string>`*

Example:

If this method is called it will *return the `BeaconBlock` class*

eth/beacon/db/chain.py
```diff
@@ -92,25 +96,49 @@ def persist_block_chain(
         raise NotImplementedError("BeaconChainDB classes must implement this method")
 
     #
-    # Raw Database API
+    # Crystallized State
     #
     @abstractmethod
-    def exists(self, key: bytes) -> bool:
+    def get_crystallized_state_by_root(self, state_root: Hash32) -> CrystallizedState:
```
cburgdorf: ntpick: I think we recently settled on a preference to just use `pass` here instead of  raising a `NotImplementedError` because the `ABC` module already takes care of not allowing these to be called, rendering the `raise` obsolete.

eth/beacon/state_machines/forks/serenity/active_states.py
cburgdorf: When judging between `Iterable` and `Sequence` my preference is to always use the most flexible type that fulfills the goal. 

For example, if you do not need to access `len()` (or e.g index access), `Iterable` will do it and is more flexible. For instance, it allows you to also pass in an iterator (`yield "foo"`). 

But if you need to get the length or need to access elements at a specific index, then one has to tighten the type to something like `Sequence` because `Iterable` simply can't give you that (e.g. it is perfectly valid for an `Iterable` to be infinite, so not having an answer to `len()` at all).

eth/beacon/state_machines/forks/serenity/configs.py
```diff
@@ -0,0 +1,16 @@
...
+
+SERENITY_CONFIG = BeaconConfig(
+    BASE_REWARD_QUOTIENT=2**15,
+    DEFAULT_END_DYNASTY=9999999999999999999,
+    DEPOSIT_SIZE=32 * denoms.ether,  # WEI
+    CYCLE_LENGTH=64,  # slots
+    MIN_COMMITTEE_SIZE=128,  # validators
+    MIN_DYNASTY_LENGTH=256,  # slots
+    SHARD_COUNT=1024,  # shards
+    SLOT_DURATION=8,  # seconds
```
cburgdorf: Should that rather be named `SLOT_DURATION_SECONDS` and `SQRT_E_DROP_SECONDS`? Alternatively, instead of assigning `int` here, we could assign `Second` which would be a `NewType` and prevent arbitrary `int` values to be assigned. /cc @pipermerriam @carver

eth/beacon/state_machines/configs.py
```diff
@@ -0,0 +1,18 @@
...
+    'BeaconConfig',
+    (
+        ('BASE_REWARD_QUOTIENT', int),
+        ('DEFAULT_END_DYNASTY', int),
+        ('DEPOSIT_SIZE', int),
+        ('CYCLE_LENGTH', int),
+        ('MIN_COMMITTEE_SIZE', int),
+        ('MIN_DYNASTY_LENGTH', int),
+        ('SHARD_COUNT', int),
+        ('SLOT_DURATION', int),
```
cburgdorf: As noted in another comment, the following two are good candidates to use a `Second` `NewType` rather than using an `int`.

eth/beacon/state_machines/base.py
pipermerriam: Just noting that this is the same for git commit messages which at least makes it easier for me to remember.

eth/beacon/state_machines/base.py
pipermerriam: Just noting that I was wrong in my first read of the code and yes, this does indeed only return down to block **greater than or equal to** `target_slot_number`

eth/beacon/state_machines/forks/serenity/configs.py
hwwhww: Renaming: now these constants (configs) names follow the previous spec - and the spec is changing. Recently Justin proposed that using slot as the common time unit, except for `SLOT_DURATION`.  e.g., `SQRT_E_DROP_TIME` in second became `SQRT_E_DROP_TIME` in slot. I will clean up all these recent changes after porting old codebase.

For beacon chain, some other `NewType` candidates: `Slot`, `ShardID`. And [with the latest spec](https://github.com/ethereum/eth2.0-specs/blob/cbd254a3090771e79ccc8eb8e21c7b32dba93c45/specs/beacon-chain.md#constants): `ValidatorStatus`, `SpecialRecordType`, `ValidatorSetDeltaFlag`.

eth/beacon/state_machines/forks/serenity/active_states.py
hwwhww: @cburgdorf thank you for elaborating! I will keep the current `Iterable`s.

eth/beacon/state_machines/forks/serenity/configs.py
hwwhww: #1383

PR #1374, opened 2018-10-08
Chain tip monitor

carver:
### What was wrong?

Sometimes a class wants to know if there is a chance of a new tip (based on a new peer or a new block).  Previously the best way to do it was to become a `PeerSubscriber`. That adds a lot of extra code overhead.

### How was it fixed?

Added a new tip monitor class. When you want to act on potentially new tip info, simply do:
```py
tip_monitor = ETHChainTipMonitor(...)
asyncio.ensure_future(tip_monitor.run())

async for highest_td_peer in tip_monitor.wait_tip_info():
    do_something_with_new_tip(highest_td_peer)
```

Pulled out of #1353 -- with a few random logging improvements.

The `highest_td_peer` getting returned is a little weird here. It is the closest implementation to the old way of doing things. One alternative would be to return the peer that provided the new tip info. I kept the current implementation, because it shouldn't be hard to change later. Knowing what other context we want to use it in might help design a better API, at that point. I'm open to suggestions, of course.

Note that it doesn't significantly reduce the size of the syncing clients yet, but it will as part of #1353 -- when the `BaseHeaderChainSyncer` will not have to be a `PeerSubscriber` anymore, so can drop several methods.

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/monitors.py
+++ b/trinity/protocol/common/monitors.py
@@ -76,7 +76,7 @@ class BaseChainTipMonitor(BaseService, PeerSubscriber):
             new_tip_event.set()
 
     async def _run(self) -> None:
-        self.run_task(self._handle_msg_loop())
+        self.run_daemon_task(self._handle_msg_loop())
         with self.subscribe(self._peer_pool):
             await self.wait(self.events.cancelled.wait())
 
--- a/trinity/protocol/common/monitors.py
+++ b/trinity/protocol/common/monitors.py
@@ -1,0 +1,93 @@
+import asyncio
+from contextlib import contextmanager
+from typing import (
+    AsyncIterator,
+    Iterator,
+    Set,
+)
+
+from cancel_token import CancelToken
+from eth_utils import ValidationError
+
+from p2p.exceptions import NoConnectedPeers
+from p2p.peer import BasePeer, PeerSubscriber
+from p2p.service import BaseService
+
+from trinity.protocol.common.peer import BaseChainPeer, BaseChainPeerPool
+from trinity.protocol.eth import commands
+
+
+class BaseChainTipMonitor(BaseService, PeerSubscriber):
+    """
+    Monitor for potential changes to the tip of the chain: a new peer or a new block
+
+    Subclass must specify :attr:`subscription_msg_types`
+    """
+
+    def __init__(
+            self,
+            peer_pool: BaseChainPeerPool,
+            token: CancelToken = None) -> None:
+        super().__init__(token)
+        self._peer_pool = peer_pool
+        # There is one event for each subscriber, each one gets set any time new tip info arrives
+        self._subscriber_notices: Set[asyncio.Event] = set()
+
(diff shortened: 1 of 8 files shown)


carver:
The test is failing because trinity is not connecting to the geth client. That's interesting for a couple reasons:
1. The connection line is *hanging* not timing out. `cancellable_wait` isn't doing what I expect it to here:  https://github.com/ethereum/py-evm/blob/f7a33ea2037e45b535682ba12efa8b985dae2c97/p2p/auth.py#L120-L123
1. I have no idea why this PR would affect the core connection logic

trinity/protocol/common/monitors.py
```diff
@@ -0,0 +1,93 @@
...
+            raise ValidationError("%s is cancelled, new tip info is impossible", self)
+        elif not self.is_running:
+            await self.events.started.wait()
+
+        with self._subscriber() as new_tip_event:
+            while self.is_operational:
+                try:
+                    highest_td_peer = self._peer_pool.highest_td_peer
+                except NoConnectedPeers:
+                    # no peers are available right now, skip the new tip info yield
```
pipermerriam: I think this might be where your problem is.  This loop looks like it won't ever yield the process if this `except` clause is hit.  I think you need to inject an `await self.sleep(0)` here somewhere to ensure that you yield the event loop at some point.

trinity/protocol/common/monitors.py
```diff
@@ -0,0 +1,93 @@
...
+
+                await self.wait(new_tip_event.wait())
+                new_tip_event.clear()
+
+    @property
+    def msg_queue_maxsize(self) -> int:
+        # This is a rather arbitrary value, but when the sync is operating normally we never see
+        # the msg queue grow past a few hundred items, so this should be a reasonable limit for
+        # now.
+        return 2000
```
pipermerriam: Can't we just set this as a normal property?

trinity/protocol/common/monitors.py
```diff
@@ -0,0 +1,93 @@
...
+            peer, cmd, msg = await self.wait(self.msg_queue.get())
+            if isinstance(cmd, commands.NewBlock):
+                self._notify_tip()
+
+    def _notify_tip(self) -> None:
+        for new_tip_event in self._subscriber_notices:
+            new_tip_event.set()
+
+    async def _run(self) -> None:
+        self.run_task(self._handle_msg_loop())
```
pipermerriam: Should be a daemon right?

trinity/protocol/common/monitors.py
carver: Yup, looks right to me, thanks!

carver:
I think there's probably value in extending this to broadcast an appropriate event, but I don't think it needs to be tied to this PR. As I started to implement it, the direction that made most sense to me was all additive (rather than replacing the current implementation). So I think I'll go ahead and merge, and post a WIP of extending this with events for plugins.

Issue #1198, opened 2018-08-20
Move peer pool interactions to be event bus based.


pipermerriam:
### What is wrong?

The `PeerPool` is used widely across most components.  Our current model for sharing APIs across process boundaries is using the `Manager` and `Proxy` object pattern from the `multiprocessing` module.

This is however proving an untenable solution.

- Significant overhead and *setup* needed.
- Hides the *overhead* of cross process communication
- Doesn't cooperate well with type checker.

### How can it be fixed

As proposed here: https://github.com/ethereum/py-evm/issues/1075#issuecomment-413489849

We'll establish a central event bus for message passing between components.  The `PeerPool` will then expose an external *public* API through the event bus from which consumers of the peer pool APIs can interact.

> Note: This likely extends to the `Peer` classes themselves as well since components will need to do more than just talk with the peer pool, but also, interact with the peer instances themselves.

cburgdorf:
Here's a PoC full of bugs https://github.com/ethereum/py-evm/pull/1202

cburgdorf:
@pipermerriam  Here's just a quick recap of our recent call. Feel free to edit to make changes.

**PeerPoolAPI**

```
class PeerPoolAPI:

    def __init__(self, event_bus: Endpoint):
        self.event_bus = event_bus


    async def get_peer_count() -> int:
        response = await self.event_bus.request(PeerCountRequest())
        return response.peer_count

    async def other_peer_pool_thingy() -> Whatever:
        ...
```

This `PeerPoolAPI` can be instantiated and used in any process (and hence also any isolated plugin) that holds a connection to the event bus (which by default all plugins do)

In general, the return values of this API will be shallow DTOs and hence there will exist similar APIs e.g. for specific peer protocol operations that consume such DTOs as call parameters and will match them in their local process against a specific peer (fat object, non-dto sibling) and invoke specific actions.

**EventBusHandler**

Inside the current `PeerPool` there should run an `EventBusHandler` that listens for all incoming events and responds to them. Same thing for other APIs (e.g peer APIs).

This should run as child services.

PR #1375, opened 2018-10-09
[WIP] broadcast new peer and chain tip events

carver:
### What was wrong?

Plugins will probably be interested in new tip events as well. The newest tip monitor class isn't suited for plugin usage.

### How was it fixed?

Broadcast events when a new peer joins the pool, and when a new eth block is received (via regular or light protocol)

Builds on #1374 

#### Cute Animal Picture

### DIFF
--- a/p2p/events.py
+++ b/p2p/events.py
@@ -1,11 +1,23 @@
 from typing import (
     Type,
+    TYPE_CHECKING,
 )
 
 from lahja import (
     BaseEvent,
     BaseRequestResponseEvent,
 )
+
+if TYPE_CHECKING:
+    from p2p.peer import BasePeer
+
+
+class PeerConnectedEvent(BaseEvent):
+    """
+    Broadcasted when a new peer of any kind connects to the peer pool.
+    """
+    def __init__(self, peer: 'BasePeer'):
+        self.peer = peer
 
 
 class PeerCountResponse(BaseEvent):
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -86,6 +86,7 @@ from .constants import (
 )
 
 from .events import (
+    PeerConnectedEvent,
     PeerCountRequest,
     PeerCountResponse,
 )
@@ -864,6 +865,8 @@ class BasePeerPool(BaseService, AsyncIterable[BasePeer]):
             peer.add_subscriber(subscriber)
             for msg in msgs:
                 subscriber.add_msg(msg)
+        if self.event_bus is not None:
+            self.event_bus.broadcast(PeerConnectedEvent(peer))
 
     async def _run(self) -> None:
         # FIXME: PeerPool should probably no longer be a BaseService, but for now we're keeping it
--- a/trinity/protocol/common/events.py
+++ b/trinity/protocol/common/events.py
@@ -1,0 +1,11 @@
+from p2p.peer import PeerMessage
+
+from trinity.extensibility.events import BaseEvent
+
+
+class NewChainTipEvent(BaseEvent):
+    """
+    Broadcasted when a new tip (in regular or light protocol) is received from a peer.
+    """
+    def __init__(self, peer_message: PeerMessage) -> None:
+        self.peer_message = peer_message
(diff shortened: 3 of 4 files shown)


p2p/events.py
```diff
@@ -1,12 +1,24 @@
...
 
+if TYPE_CHECKING:
+    from p2p.peer import BasePeer
+
+
+class PeerConnectedEvent(BaseEvent):
+    """
+    Broadcasted when a new peer of any kind connects to the peer pool.
+    """
+    def __init__(self, peer: 'BasePeer'):
```
pipermerriam: I would expect that we can't transmit peer instances over multiprocessing pipes...  am I missing something?

trinity/protocol/common/events.py
```diff
@@ -0,0 +1,11 @@
...
+
+from trinity.extensibility.events import BaseEvent
+
+
+class NewChainTipEvent(BaseEvent):
+    """
+    Broadcasted when a new tip (in regular or light protocol) is received from a peer.
+    """
+    def __init__(self, peer_message: PeerMessage) -> None:
+        self.peer_message = peer_message
```
pipermerriam: This seems like it would be good to transmit this data in a normalized fashion rather than the raw message (which is what I assume this is doing).

p2p/events.py
cburgdorf: Every event that travels across processes (that is every event except those that are `BroadcastConfig(internal=True)`) needs to be pickable. I assume that a `BasePeer` would not be pickable.

My vision for how these things should work is basically that we would create shallow DTOs for this purpose which can be matched back to the original peer based on some unique identifier ([see also](https://github.com/ethereum/py-evm/issues/1198#issuecomment-417740431)).

p2p/peer.py
```diff
@@ -864,6 +865,8 @@ def _add_peer(self,
             peer.add_subscriber(subscriber)
             for msg in msgs:
                 subscriber.add_msg(msg)
+        if self.event_bus is not None:
+            self.event_bus.broadcast(PeerConnectedEvent(peer))
```
cburgdorf: I'm not sure how often `PeerConnnectedEvent` occurs in normal networking traffic but let me just raise awareness that broadcasting very frequent events via the event bus may cause performance issues. There's a lot that we can optimize about it (see, https://github.com/ethereum/lahja/issues/2 and https://github.com/ethereum/lahja/issues/1) but as of today, every event broadcasted travels at least one hop from the process that it originates from (`networking` in this case) to the `main` process where the central event bus coordinator lives and then from there to each and every other connected process (unless `BroadcastConfig(filter_endpoint='some_specific_endpoint')` was used).

trinity/protocol/common/events.py
carver: It's a named tuple with some important bits split out https://github.com/ethereum/py-evm/blob/fe2019b8af7253810328181be50d9638b98f0edf/p2p/peer.py#L616-L619

So it's going to have the same pickleable problem as the others.

p2p/peer.py
carver: It shouldn't be frequent at all. Typically once every several seconds. Although it's possible that some certain corner case might cause peers to drop and reconnect a lot in a short period. I wouldn't expect it to be more than a burst.

trinity/protocol/common/events.py
pipermerriam: Yeah, I was actually thinking that it would be a custom payload which contained maybe a peer identifier and then the normalized tip info (`head_td`, `head_number`, `head_hash`?)

pipermerriam:

carver:
This got stale, and it seems like @cburgdorf is working on related/overlapping things. Closing for now.

FWIW, I'm not excited about having maintain another class for every class we want to send across the event bus (eg~ the "real" BasePeer, and the one that can be pickled). If there's any way to make the original objects pickle-able, that seems like the best option. Second best option seems like: some kind of proxy object that sends a request to the live object for data. I think those things would be easier to maintain, requiring fewer changes every time a class is modified.

cburgdorf:
>some kind of proxy object that sends a request to the live object for data.

I came up with some kind of proxy pattern back when we bridged the `LightDispatchChain` to be available via the eventbus. It was basically maintenance free but relied on a certain type of magic (call stack introspection)

https://github.com/ethereum/py-evm/pull/1262#issuecomment-420232866

We then settled on a very explicit pattern instead which, so far, does still feel manageable to me. The stronger type safety guarantees that we have today do a fairly good job to yield errors when API changes need to spread across multiple places.

But I'm happy to re-evaluate options when we eventbusify the `PeerPool`.

PR #1377, opened 2018-10-09
Migrate util usage to eth_utils

cburgdorf:
### What was wrong?

Came across this while working on something else. Seems we had some helpers in `eth.utils.numeric` that we also have in `eth_utils`. Guess those were in place from a time where `eth_utils did not exist or at least did not have these methods.

### How was it fixed?

Migrated to `eth_utils`.

#### Cute Animal Picture

### DIFF
--- a/eth/consensus/pow.py
+++ b/eth/consensus/pow.py
@@ -14,6 +14,7 @@ from eth_typing import (
     Hash32
 )
 from eth_utils import (
+    big_endian_to_int,
     ValidationError,
 )
 
@@ -21,9 +22,6 @@ from eth_hash.auto import keccak
 
 from eth.utils.hexadecimal import (
     encode_hex,
-)
-from eth.utils.numeric import (
-    big_endian_to_int,
 )
 from eth.validation import (
     validate_length,
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -19,7 +19,10 @@ from trie import (
 )
 
 from eth_hash.auto import keccak
-from eth_utils import encode_hex
+from eth_utils import (
+    encode_hex,
+    int_to_big_endian,
+)
 
 from eth.constants import (
     BLANK_ROOT_HASH,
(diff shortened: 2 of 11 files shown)


pipermerriam:
These are the only lines where that function is used where the difference in behavior would manifest:

- https://github.com/ethereum/py-evm/blob/master/eth/utils/transactions.py#L51
- https://github.com/ethereum/py-evm/blob/master/eth/vm/forks/spurious_dragon/transactions.py#L22
- (maybe) https://github.com/ethereum/py-evm/blob/master/eth/vm/stack.py#L75

pipermerriam:
The two functions behave the same in all cases **except** for `0`.

- `eth_utils.int_to_big_endian(0) -> b'\x00'`
- `eth.utils.numeric.int_to_big_endian(0) -> b''`

pipermerriam:
> IMO `None` should be handled as a separate case rather than expecting `int_to_big_endian` to convert it to `b''`.

Yeah, I would expect/want `int_to_big_endian(None)` to throw an exception.

cburgdorf:
>Interesting, it looks like the default value of None would crash here

https://github.com/ethereum/py-evm/blob/19aed7a379ca643ac051ffd94ea2d8dc8c01579f/eth/utils/transactions.py#L46-L51

But the `if chain_id` prevents it from hitting that branch anyway. Btw, if you hit `y` on your keyboard, github generates a link to the commit (rather than the branch) and does then inline the code snippet instead of just rendering the link.

However, it's not that. The reason this failed is because our `modexp` precompile had a bug that previously wasn't exposed because `int_to_big_endian` had a bug (kinda).

I changed the `modexp` precompile to be inline with what geth does.

carver:
> Btw, if you hit y on your keyboard, github generates a link to the commit (rather than the branch)

neat! :tada: 

> But the if chain_id prevents it from hitting that branch anyway.

oops :man_facepalming: 

> I changed the `modexp` precompile to be inline with what geth does.

:+1:

PR #1378, opened 2018-10-09
Fix bug in CREATE2 address generation + add test

cburgdorf:
### What was wrong?

The generated address for the new CREATE2 opcode turned out to be wrong. By the time that I implemented it, there were no test cases so the implementation was a shot in the dark (and still kinda is, as this only partly covers what CREATE2 does).

Now we do have test cases and they discovered a bug in our implementation.

### How was it fixed?

Correctly interpret salt + add tests.

#### Cute Animal Picture

### DIFF
--- a/tests/core/address-utils/test_address_generation.py
+++ b/tests/core/address-utils/test_address_generation.py
@@ -1,0 +1,68 @@
+import pytest
+
+from eth_utils import (
+    decode_hex,
+    big_endian_to_int,
+    is_same_address,
+)
+from eth.utils.address import (
+    generate_safe_contract_address,
+)
+
+
+@pytest.mark.parametrize(
+    # Test cases from: https://eips.ethereum.org/EIPS/eip-1014
+    'origin, salt, code, expected',
+    (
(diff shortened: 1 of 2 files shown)


tests/core/address-utils/test_address_generation.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+)
+def test_generate_safe_contract_address(origin, salt, code, expected):
+    address = generate_safe_contract_address(
+        decode_hex(origin),
+        big_endian_to_int(decode_hex(salt)),
+        decode_hex(code)
+    )
+
+    assert to_checksum_address(encode_hex(address)) == expected
```
pipermerriam: This is generally fine, but our idiomatic way to compare addresses is to use `eth_utils.is_same_address` which handles comparison of differently formatted addresses.

cburgdorf: Ah, now that you mentions it, I remember you mentioned it before :sweat_smile: Fixed it.

Issue #1275, opened 2018-09-10
Snappy compression


pipermerriam:
### What is wrong?

Based on a note within the AllCoreDevs channel snappy compression may be manditory very soon.  

### How can it be fixed

We'll need to implement compression/decompression in our networking layer.

Bhargavasomu:
Is using the `python-snappy` library ok? Or do we need to implement it from scratch in `cython` preferably for speed?

Bhargavasomu:
## Communication Between Peers
* We could have a `compress command` for the message before sending it in `peer.py` before the line 555 https://github.com/ethereum/py-evm/blob/6aa10e54a274c7068a535f279fe0396c5e275fa7/p2p/peer.py#L555
* In the receiving side, we could include a `decompress command` in `peer.py` after receiving the compressed message, and then returning the decompressed msg i.e., after the line 334 (without the return though)
https://github.com/ethereum/py-evm/blob/6aa10e54a274c7068a535f279fe0396c5e275fa7/p2p/peer.py#L334

## Discovery Protocol
We could include Snappy Compression here too.
* We could add a compress command in `discovery.py` before line 424
https://github.com/ethereum/py-evm/blob/6aa10e54a274c7068a535f279fe0396c5e275fa7/p2p/discovery.py#L424
* We then add a decompress command in `discovery.py` after line 414 and then continue the rest of the function
https://github.com/ethereum/py-evm/blob/6aa10e54a274c7068a535f279fe0396c5e275fa7/p2p/discovery.py#L414

Is this ok or am I wrong?

pipermerriam:
cc @carver and @cburgdorf I'm getting started with https://github.com/ethereum/py-snappy for a pure python implementation, but in the meantime @Bhargavasomu you can use the `python-snappy` bindings.

Bhargavasomu:
The function `read_msg` in `p2p/peer.py` is responsible for reading the messages in the channel. It always reads the `header` and the `body` given that they are of fixed length to all the peers which is `HEADER_LEN + MAC_LEN` and `read_size + MAC_LEN`
https://github.com/ethereum/py-evm/blob/740dc53a1508b648cc035e0fac1dbf34dd419e59/p2p/peer.py#L386
[...]

PR #1380, opened 2018-10-09
Add Snappy Compression

Bhargavasomu:
### What was wrong?
Need to introduce `Snappy Compression` in Networking Layer
Issue : #1275

### How was it fixed?
Modified the `send` and `receive` functions so as to compress and decompress using `python-snappy`.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses:max_bytes(150000):strip_icc()/iStock-619961796-edit-59cabaf6845b3400111119b7.jpg)

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -412,6 +412,7 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
         address = kademlia.Address(ip_address, udp_port)
         # The prefix below is what geth uses to identify discv5 msgs.
         # https://github.com/ethereum/go-ethereum/blob/c4712bf96bc1bae4a5ad4600e9719e4a74bde7d5/p2p/discv5/udp.go#L149  # noqa: E501
+        data = snappy.decompress(data)
         if text_if_str(to_bytes, data).startswith(V5_ID_STRING):
             self.receive_v5(address, cast(bytes, data))
         else:
@@ -421,7 +422,8 @@ class DiscoveryProtocol(asyncio.DatagramProtocol):
         self.logger.error('error received: %s', exc)
 
     def send(self, node: kademlia.Node, message: bytes) -> None:
-        self.transport.sendto(message, (node.address.ip, node.address.udp_port))
+        compressed_msg = snappy.compress(message)
+        self.transport.sendto(compressed_msg, (node.address.ip, node.address.udp_port))
 
     async def stop(self) -> None:
         self.logger.info('stopping discovery')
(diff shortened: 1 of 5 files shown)


p2p/discovery.py
```diff
@@ -412,6 +412,7 @@ def datagram_received(self, data: Union[bytes, Text], addr: Tuple[str, int]) ->
         address = kademlia.Address(ip_address, udp_port)
         # The prefix below is what geth uses to identify discv5 msgs.
         # https://github.com/ethereum/go-ethereum/blob/c4712bf96bc1bae4a5ad4600e9719e4a74bde7d5/p2p/discv5/udp.go#L149  # noqa: E501
+        data = snappy.decompress(data)
```
pipermerriam: We prefer not to overwrite variables.  I'd adjust the "compressed" value to be `raw_data` and the decompressed value to be assigned to `data`.

Bhargavasomu:
screenshot from 2018-10-10 00-07-20
This above error is occurring because the user needs to pre-install the `libsnappy-dev` library preferabbly through `sudo apt-get install libsnappy-dev`. How do I setup the installation of the library?

p2p/discovery.py
Bhargavasomu: @pipermerriam will do the changes as requested. Also I think tox.ini has to be modified so that it sets up the environment accordingly. Any insights into how I can do the tox.ini modification?

pipermerriam: The modification likely needs to take place in the `./.circleci/config` file to do the `apt` install.

You can see how that is done in the `py-snappy` repo here: https://github.com/ethereum/py-snappy/blob/master/.circleci/config.yml#L23-L25

Bhargavasomu: @pipermerriam I believe that some modifications need to be done to the tox.ini file too, because a lot of them are failing because of tox

Bhargavasomu:
I guess adding the Try Catch exceptions should take care of the failed test cases.

pipermerriam:
@Bhargavasomu I don't think that's the right approach.  The `try/except` would be intended to catch the error and likely re-raise under a different exception.

What I suspect is happening here is that the compressed data has a different size and so the calls to `readexactly` are returning either too little or too much data and thus it can't be decompressed.

carver:
> thoughts on putting this behind a feature flag?

Yeah, seems like it should be opt-in until we can set it up to install automatically as part of the `pip install`. Maybe an environment variable, since I don't think we will always need to give this option. It's more awkward to remove a CLI flag, later.

Bhargavasomu:
Sending the compressed message length first and then sending the compressed message so that it could be decompressed. Still trinity fails. Is it because the testing in trinity doesn't first compress the message while testing?

pipermerriam:
@Bhargavasomu I don't think this is the right issue for you to work on.  Looks like there is some low level p2p networking know-how needed and without digging in myself I don't know what the right approach should be here.

PR #1384, opened 2018-10-10
Colorize shell output based on log level

cburgdorf:
### What was wrong?

Do you spot the `ERROR` in here?

image

Me, neither!

### How was it fixed?

Colors! :rainbow: 

image

With this change, everything worse than `ERROR` is printed bold and red and everything worse than `WARNING` (but less bad than `ERROR`) is printed bold and yellow.

Notice that this copies the `shellart` util that we already use in the benchmark to Trinity because there doesn't seem to be a common space to share between both projects and the file is simple and small enough that I think the copy is ok.

#### Cute Animal Picture

### DIFF
--- a/trinity/utils/logging.py
+++ b/trinity/utils/logging.py
@@ -28,6 +28,11 @@ from eth.tools.logging import (
     TraceLogger,
 )
 
+from trinity.utils.shellart import (
+    bold_red,
+    bold_yellow,
+)
+
 if TYPE_CHECKING:
     from multiprocessing import Queue  # noqa: F401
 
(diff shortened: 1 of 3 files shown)


trinity/utils/logging.py
```diff
@@ -42,6 +47,13 @@ def __init__(self, fmt: str, datefmt: str) -> None:
 
     def format(self, record: logging.LogRecord) -> str:
         record.shortname = record.name.split('.')[-1]  # type: ignore
+
+        if record.levelno >= logging.ERROR:
+            return bold_red(super().format(record))
+
+        if record.levelno >= logging.WARNING:
```
pipermerriam: *nitpick* but the return statements make these *implicit* `elif` statements.  Maybe go ahead and just make this one single `if/elif/else` block?

PR #1386, opened 2018-10-11
Fix and improve logging output

cburgdorf:
### What was wrong?

1. Trinity always prints `Keyboard Interrupt: Stopping` on shutdown even if the shutdown was not initiated by a `KeyboardInterrupt`
2. The `message` parameter to `kill_trinity_gracefully` was never used and hence removed

### How was it fixed?

1. Trinity now prints the reason for the shutdown next to the `Shutting down Trinity` as well as next to the `Trinity shutdown complete` message, making it more obvious to the user why Trinity was shutdown
2. `message` parameter was removed

#### Cute Animal Picture

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -262,7 +262,8 @@ def trinity_boot(args: Namespace,
             networking_process,
             plugin_manager,
             main_endpoint,
-            event_bus
+            event_bus,
+            ev.reason
         )
     )
 
@@ -279,7 +280,8 @@ def trinity_boot(args: Namespace,
             networking_process,
             plugin_manager,
             main_endpoint,
-            event_bus
+            event_bus,
+            reason="CTRL+C / Keyboard Interrupt"
         )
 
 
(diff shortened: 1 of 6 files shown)


trinity/main.py
```diff
@@ -262,7 +262,8 @@ def trinity_boot(args: Namespace,
             networking_process,
             plugin_manager,
             main_endpoint,
-            event_bus
+            event_bus,
+            ev.reason
```
cburgdorf: Just pointing out that `lahja` provides type safety for these events which means using something like `ev.foobar` results in a `mypy` error. However, for some reason, our type safety from mypy running inside `tox` is currently degraded (see https://github.com/python/mypy/issues/5771)

pipermerriam: Thoughts on add a temporary new CI run to run `mypy` outside of tox in circle?

cburgdorf: Fixed with #1389

PR #1388, opened 2018-10-11
unknown unknowns

cburgdorf:
Just testing unknown unknowns


cburgdorf:
Just testing if this behaves as expected. The mypy errors here are exactly what we want. It's all the stuff that previously wasn't type safe because `eth-typing` didn't distribute type hints as per PEP561.

PR #1385, opened 2018-10-11
Fix wrong types

cburgdorf:
### What was wrong?

Type hints were wrong.

What is super weird about it: Somehow, `mypy` inside `tox` did not catch this. Running the same `mypy` command outside of `tox` (with the same `mypy` version) caught this though.

```
$ tox -e py36-lint
py36-lint develop-inst-nodeps: /home/cburgdorf/Documents/hacking/ef/py-evm
py36-lint installed: asn1crypto==0.24.0,cffi==1.11.5,cryptography==2.3.1,cytoolz==0.9.0.1,eth-bloom==1.0.1,eth-hash==0.2.0,eth-keys==0.2.0b3,eth-typing==1.3.0,eth-utils==1.2.2,flake8==3.5.0,idna==2.7,lru-dict==1.1.6,mccabe==0.6.1,mypy==0.630,mypy-extensions==0.4.1,py-ecc==1.4.3,-e git+https://github.com/ethereum/py-evm.git@cafd4a6b9bf11bc69178331d73c20bdc24fec3b9#egg=py_evm,pycodestyle==2.3.1,pycparser==2.19,pyethash==0.1.27,pyflakes==1.6.0,rlp==1.0.3,six==1.11.0,toolz==0.9.0,trie==1.3.8,typed-ast==1.1.0
py36-lint runtests: PYTHONHASHSEED='1314036567'
py36-lint runtests: commands[0] | flake8 /home/cburgdorf/Documents/hacking/ef/py-evm/eth
py36-lint runtests: commands[1] | flake8 /home/cburgdorf/Documents/hacking/ef/py-evm/tests
py36-lint runtests: commands[2] | flake8 /home/cburgdorf/Documents/hacking/ef/py-evm/p2p
py36-lint runtests: commands[3] | flake8 /home/cburgdorf/Documents/hacking/ef/py-evm/trinity
py36-lint runtests: commands[4] | flake8 /home/cburgdorf/Documents/hacking/ef/py-evm/scripts
py36-lint runtests: commands[5] | mypy --follow-imports=silent --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics scripts/benchmark
py36-lint runtests: commands[6] | mypy --follow-imports=silent --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p p2p -p trinity
[...]

PR #1389, opened 2018-10-11
Ensure mypy checks correctly in CI

cburgdorf:
### What was wrong?

As seen with #1385 and https://github.com/python/mypy/issues/5771 `mypy` currently doesn't provide us the type safety that we expect it should.

### How was it fixed?

Turns out, we didn't install all expected dependencies for the `mypy` run. That means, `mypy` was checking the source code but didn't have any of the other dependencies installed. Having the dependencies installed is important for mypy to infer the type info it needs to do its checks.

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -92,7 +92,7 @@ commands=
 
 
 [common-lint]
-deps = .[lint]
+deps = .[eth,p2p,trinity,lint]
 setenv=MYPYPATH={toxinidir}:{toxinidir}/stubs
 commands=
     flake8 {toxinidir}/eth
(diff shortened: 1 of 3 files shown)


cburgdorf:
@pipermerriam this now catches the bug that is fixed with #1385

pipermerriam:
Can we pull in #1385 on top of this and merge them together?

cburgdorf:
>Can we pull in #1385 on top of this and merge them together?

I take this as an **approval** as the complexity of the change is rather low.

Issue #1381, opened 2018-10-09
Port beacon chain


hwwhww:
> I found that #1339 became a huge ticket beyond just "helper functions". Instead, open this one for an overview.

![](https://i.imgur.com/NZBO33B.png)

### Role
- Provides database interface.

### Status:
- [x] DB APIs are done in [py-evm#1373](https://github.com/ethereum/py-evm/pull/1373) 
- [ ] **[TODO]** [Fork choice rule](https://github.com/ethereum/eth2.0-specs/blob/master/specs/beacon-chain.md#beacon-chain-fork-choice-rule) -> waiting for the spec to  be finalized.

### Role
- Provides state transition.

### Status:
- - [x] Basic forkable architecture: [py-evm#1373](https://github.com/ethereum/py-evm/pull/1373)
- - [ ] Port functions from `beacon_chain`: Under construction :construction: 

#### Process
- [`import_block`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L709-L743): needs to extended `compute_state_transition` logic.
    - TODO: set state roots of the block
- [`compute_cycle_transitions`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L686-L706)
- [`compute_dynasty_transition`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L662-L683)
- [`process_block`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L258-L295)
- [`process_updated_crosslinks`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L298-L337)
- [`initialize_new_cycle`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L340-L421)
- [`fill_recent_block_hashes`](https://github.com/ethereum/beacon_chain/blob/ccf75cc679601b53e03fb5a410fc3262c503ace1/beacon_chain/state/state_transition.py#L424-L437)
[...]

PR #1390, opened 2018-10-12
Per block transition and attestation

hwwhww:
Apologize for the fat PR. 🙈 

### What was wrong?
#1381 - StateMachine - per block transition and attestation

#### Follow [the beacon chain spec - Per-block processing section](https://github.com/ethereum/eth2.0-specs/blob/fe74c7e299ff90d56f4099733754c0e94e645ccc/specs/beacon-chain.md#per-block-processing) **except**:
1. No `update_ancestor_hashes`  design in this PR
2. No RANDAO reveal design in this PR
3. No versioning design in this PR

#### Modifications:
1. `helpers.py`
	1. Bugfix `get_hashes_to_sign` - shouldn’t re-hash the latest hash
	2. Remove multiple slots per committee option: https://github.com/ethereum/eth2.0-specs/pull/62
	3. Rename `get_proposer_position` to `get_block_committees_info`, it will return `BlockCommitteesInfo` as a simple structure of the committee information context.
2. `eth.beacon.state_machines.base.py`
	1. **Import block APIs**
		1. Import the given block into state machine
		2. Call process APIs; if valid, update the states in memory
	2. **Process block APIs** (all are classmethods)
		1. Pure functions for processing a block
		2. `compute_per_block_transtion`: compute for every block, will return the updated `active_state`. 
			1. Validate the `validate_parent_block_proposer` and `attestations`
			2. Return the new `active_state`
		3. (TODO in other PRs) `compute_cycle_transitions`: compute the cycle transitions. (`CYCLE_LENGTH = 64 slots`), will return the updated `crystallized_state` and `active_state`.
	3. **Proposer APIs** 
		1.  Proposer story, please see `test_propose_block_and_validate_attestation` tests:
			 1. The block proposer of slot 1:
				 1. Creates a block and process the block `B1`.
					 * Note: no shard block validation currently, so I just set `shard_block_hash=ZERO_HASH32`.
				 2. Attests the block and create an `AttestationRecord` of `B1`
				 3. Broadcasts `B1` **and the attestation record of  `B1`.**
			 2. The other attesters validate `B1`.
				 * If it's valid, sign the same message.
			 3. The next block proposer or other nodes collect  the votes
				* Aggregate the signatures and build attestation record.
				* Broadcast the attestation record.
			 4. The block proposer of slot 2:
				 1. Create a block `B2` with the aggregated attestation records
				 2. Attests the block and create an `AttestationRecord`
				 3. Broadcasts `B2` **and the attestation record of  `B2`.**
                          (Continue)
	4. **Validation APIs** (all are classmethods)
		1. proposer validation 
		2. attestation validation

#### TODO
- Add tests for validation APIs
- Add tests for aggregate signature
- Add test for multiple attestations and multiple blocks.
- Add assertions in the current tests.
- Too many arguments passing! 
	* Yep I see > 6 is a bad signal. I’d prefer to keep using pure function in processing the block, so need to refactor it.
	* FYI: The only reason why Proposer APIs are not classmethods is because it calls `self.get_attestation_record_class()`, any other way to avoid it?

@djrtwo: the proposer logic is partially from old `mock_make_attestations` and partially something new, could you verify the description above? Thanks!

#### Cute Animal Picture
Cat-sper!
caspercat_by_steve_begin

### DIFF
--- a/eth/beacon/aggregation.py
+++ b/eth/beacon/aggregation.py
@@ -1,0 +1,79 @@
+from typing import (
+    Iterable,
+    Tuple,
+)
+from cytoolz import (
+    pipe
+)
+
+
+from eth_typing import (
+    Hash32,
+)
+
+from eth.utils import bls
+from eth.utils.bitfield import (
+    set_voted,
+)
+from eth.utils.blake import blake
+
+
+def create_signing_message(slot: int,
+                           parent_hashes: Iterable[Hash32],
+                           shard_id: int,
+                           shard_block_hash: Hash32,
+                           justified_slot: int) -> bytes:
+    """
+    Return the signining message for attesting.
+    """
+    # TODO: Will be updated with SSZ encoded attestation.
+    return blake(
+        slot.to_bytes(8, byteorder='big') +
+        b''.join(parent_hashes) +
+        shard_id.to_bytes(2, byteorder='big') +
+        shard_block_hash +
+        justified_slot.to_bytes(8, 'big')
+    )
+
+
+def verify_votes(
+        message: bytes,
+        votes: Iterable[Tuple[int, bytes, int]]) -> Tuple[Tuple[bytes, ...], Tuple[int, ...]]:
+    """
+    Verify the given votes.
+
+    vote: (committee_index, sig, public_key)
+    """
+    sigs_with_committe_info = tuple(
+        (sig, committee_index)
+        for (committee_index, sig, public_key)
+        in votes
+        if bls.verify(message, public_key, sig)
+    )
+    try:
+        sigs, committee_indices = zip(*sigs_with_committe_info)
+    except ValueError:
+        sigs = tuple()
+        committee_indices = tuple()
+
+    return sigs, committee_indices
+
+
+def aggregate_votes(bitfield: bytes,
+                    sigs: Iterable[bytes],
+                    voting_sigs: Iterable[bytes],
+                    voting_committee_indices: Iterable[int]) -> Tuple[bytes, Tuple[int]]:
+    """
+    Aggregate the votes.
+    """
+    # Update the bitfield and append the signatures
+    sigs = tuple(sigs) + tuple(voting_sigs)
+    bitfield = pipe(
+        bitfield,
+        *(
+            set_voted(index=committee_index)
+            for committee_index in voting_committee_indices
+        )
+    )
+
+    return bitfield, bls.aggregate_sigs(sigs)
--- a/eth/beacon/block_committees_info.py
+++ b/eth/beacon/block_committees_info.py
@@ -1,6 +1,6 @@
 from typing import (
-    Iterable,
     NamedTuple,
+    Tuple,
     TYPE_CHECKING,
 )
 
@@ -15,6 +15,6 @@ BlockCommitteesInfo = NamedTuple(
         ('proposer_index_in_committee', int),
         ('proposer_shard_id', int),
         ('proposer_committee_size', int),
-        ('shards_and_committees', Iterable['ShardAndCommittee'])
+        ('shards_and_committees', Tuple['ShardAndCommittee'])
     )
 )
--- a/eth/beacon/block_committees_info.py
+++ b/eth/beacon/block_committees_info.py
@@ -1,0 +1,47 @@
+from typing import (
+    Iterable,
+    TYPE_CHECKING,
+)
+
+if TYPE_CHECKING:
+    from eth.beacon.types.shard_and_committees import ShardAndCommittee  # noqa: F401
+
+
+class BlockCommitteesInfo:
+    _proposer_index = None  # validator index
+    _proposer_index_in_committee = None
+    _proposer_shard_id = None
+    _proposer_committee_size = None
+    _shards_and_committees = None
+
+    def __init__(self,
+                 proposer_index: int,
+                 proposer_index_in_committee: int,
+                 proposer_shard_id: int,
+                 proposer_committee_size: int,
+                 shards_and_committees: Iterable['ShardAndCommittee']) -> None:
+        self._proposer_index = proposer_index
+        self._proposer_index_in_committee = proposer_index_in_committee
+        self._proposer_shard_id = proposer_shard_id
+        self._proposer_committee_size = proposer_committee_size
+        self._shards_and_committees = shards_and_committees
+
+    @property
+    def proposer_index(self) -> int:
+        return self._proposer_index
+
+    @property
+    def proposer_index_in_committee(self) -> int:
+        return self._proposer_index_in_committee
+
+    @property
+    def proposer_shard_id(self) -> int:
+        return self._proposer_shard_id
+
+    @property
+    def proposer_committee_size(self) -> int:
+        return self._proposer_committee_size
+
+    @property
+    def shards_and_committees(self) -> Iterable['ShardAndCommittee']:
+        return self._shards_and_committees
--- a/eth/beacon/block_committees_info.py
+++ b/eth/beacon/block_committees_info.py
@@ -1,5 +1,6 @@
 from typing import (
     Iterable,
+    NamedTuple,
     TYPE_CHECKING,
 )
 
@@ -7,41 +8,13 @@ if TYPE_CHECKING:
     from eth.beacon.types.shard_and_committees import ShardAndCommittee  # noqa: F401
 
 
-class BlockCommitteesInfo:
-    _proposer_index = None  # validator index
-    _proposer_index_in_committee = None
-    _proposer_shard_id = None
-    _proposer_committee_size = None
-    _shards_and_committees = None
-
-    def __init__(self,
-                 proposer_index: int,
-                 proposer_index_in_committee: int,
-                 proposer_shard_id: int,
-                 proposer_committee_size: int,
-                 shards_and_committees: Iterable['ShardAndCommittee']) -> None:
-        self._proposer_index = proposer_index
-        self._proposer_index_in_committee = proposer_index_in_committee
-        self._proposer_shard_id = proposer_shard_id
-        self._proposer_committee_size = proposer_committee_size
-        self._shards_and_committees = shards_and_committees
-
-    @property
-    def proposer_index(self) -> int:
-        return self._proposer_index
-
-    @property
-    def proposer_index_in_committee(self) -> int:
-        return self._proposer_index_in_committee
-
-    @property
-    def proposer_shard_id(self) -> int:
-        return self._proposer_shard_id
-
-    @property
-    def proposer_committee_size(self) -> int:
-        return self._proposer_committee_size
-
-    @property
-    def shards_and_committees(self) -> Iterable['ShardAndCommittee']:
-        return self._shards_and_committees
+BlockCommitteesInfo = NamedTuple(
+    'BlockCommitteesInfo',
+    (
+        ('proposer_index', int),
+        ('proposer_index_in_committee', int),
+        ('proposer_shard_id', int),
+        ('proposer_committee_size', int),
+        ('shards_and_committees', Iterable['ShardAndCommittee'])
+    )
+)
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -435,7 +435,7 @@ def aggregate_attestation_record(last_justified_slot: int,
         block.slot_number,
         parent_hashes,
         proposer_attestation.shard_id,
-        block.shard_block_hash,
+        proposer_attestation.shard_block_hash,
         last_justified_slot,
     )
 
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -439,7 +439,7 @@ def aggregate_attestation_record(last_justified_slot: int,
         block.shard_block_hash,
         last_justified_slot,
     )
-    
+
     bitfield, sigs = aggregate_votes(
         message,
         votes,
@@ -475,7 +475,7 @@ def aggregate_votes(message: bytes,
     bitfield = pipe(
         bitfield,
         *(
-            functools.partial(set_voted, index=committee_index)
+            set_voted(index=committee_index)
             for committee_index in committee_indices
         )
     )
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -2,9 +2,6 @@ from itertools import (
     repeat,
 )
 
-from cytoolz import (
-    pipe
-)
 from typing import (
     Any,
     Iterable,
@@ -21,11 +18,6 @@ from eth_typing import (
     Hash32,
 )
 
-from eth.utils import bls
-from eth.utils.bitfield import (
-    set_voted,
-)
-from eth.utils.blake import blake
 from eth.utils.numeric import (
     clamp,
 )
@@ -394,105 +386,3 @@ def get_block_committees_info(parent_block: 'BaseBeaconBlock',
         proposer_committee_size=proposer_committee_size,
         shards_and_committees=shards_and_committees,
     )
-
-
-#
-# Signatures and Aggregation
-#
-def create_signing_message(slot: int,
-                           parent_hashes: Iterable[Hash32],
-                           shard_id: int,
-                           shard_block_hash: Hash32,
-                           justified_slot: int) -> bytes:
-    # TODO: will be updated to hashed encoded attestation
-    return blake(
-        slot.to_bytes(8, byteorder='big') +
-        b''.join(parent_hashes) +
-        shard_id.to_bytes(2, byteorder='big') +
-        shard_block_hash +
-        justified_slot.to_bytes(8, 'big')
-    )
-
-
-def aggregate_attestation_record(last_justified_slot: int,
-                                 recent_block_hashes: Iterable[Hash32],
-                                 block: 'BaseBeaconBlock',
-                                 votes: Iterable[Tuple[int, bytes, int]],
-                                 proposer_attestation: 'AttestationRecord',
-                                 cycle_length: int) -> 'AttestationRecord':
-    """
-    Aggregate the votes.
-
-    TODO: Write tests
-    """
-    # Get signing message
-    parent_hashes = get_hashes_to_sign(
-        recent_block_hashes,
-        block,
-        cycle_length,
-    )
-    message = create_signing_message(
-        block.slot_number,
-        parent_hashes,
-        proposer_attestation.shard_id,
-        proposer_attestation.shard_block_hash,
-        last_justified_slot,
-    )
-
-    # Verify
-    # TODO: Verification is very slow, needs to compute in parallel
-    voting_sigs, committee_indices = verify_votes(message, votes)
-
-    # Aggregate the votes
-    bitfield, sigs = aggregate_votes(
-        bitfield=proposer_attestation.bitfield,
-        sigs=proposer_attestation.aggregate_sig,
-        voting_sigs=voting_sigs,
-        voting_committee_indices=committee_indices
-    )
-
-    return proposer_attestation.copy(
-        bitfield=bitfield,
-        sigs=bls.aggregate_sigs(sigs),
-    )
-
-
-def verify_votes(
-        message: bytes,
-        votes: Iterable[Tuple[int, bytes, int]]) -> Tuple[Tuple[bytes, ...], Tuple[int, ...]]:
-    """
-    Verify the given votes
-    """
-    sigs_with_committe_info = tuple(
-        (sig, committee_index)
-        for (committee_index, sig, public_key)
-        in votes
-        if bls.verify(message, public_key, sig)
-    )
-    try:
-        sigs, committee_indices = zip(*sigs_with_committe_info)
-    except ValueError:
-        sigs = tuple()
-        committee_indices = tuple()
-
-    return sigs, committee_indices
-
-
-def aggregate_votes(bitfield: bytes,
-                    sigs: Iterable[bytes],
-                    voting_sigs: Iterable[bytes],
-                    voting_committee_indices: Iterable[int]) -> Tuple[bytes, Tuple[int]]:
-    """
-    Aggregate the votes
-    """
-    # Update the bitfield and append the signatures
-    sigs = tuple(sigs) + tuple(voting_sigs)
-    bitfield = pipe(
-        bitfield,
-        *(
-            set_voted(index=committee_index)
-            for committee_index in voting_committee_indices
-        )
-    )
-
-    return bitfield, bls.aggregate_sigs(sigs)
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -1,4 +1,3 @@
-import functools
 from itertools import (
     repeat,
 )
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -45,7 +45,7 @@ def _get_element_from_recent_list(
         target_slot: int,
         slot_relative_position: int) -> Any:
     """
-    Returns the element from ``target_list`` by the ``target_slot`` number,
+    Return the element from ``target_list`` by the ``target_slot`` number,
     where the the element should be at ``target_slot - slot_relative_position``th
     element of the given ``target_list``.
     """
@@ -75,7 +75,7 @@ def get_block_hash(
         slot: int,
         cycle_length: int) -> Hash32:
     """
-    Returns the blockhash from ``ActiveState.recent_block_hashes`` by
+    Return the blockhash from ``ActiveState.recent_block_hashes`` by
     ``current_block_slot_number``.
     """
     if len(recent_block_hashes) != cycle_length * 2:
@@ -192,7 +192,7 @@ def get_attestation_indices(crystallized_state: 'CrystallizedState',
                             attestation: 'AttestationRecord',
                             cycle_length: int) -> Iterable[int]:
     """
-    Returns committee of the given attestation.
+    Return committee of the given attestation.
     """
     shard_id = attestation.shard_id
 
@@ -231,7 +231,7 @@ def _get_shuffling_committee_slot_portions(
         min_committee_size: int,
         shard_count: int) -> Tuple[int, int]:
     """
-    Returns committees number per slot and slots number per committee.
+    Return committees number per slot and slots number per committee.
     """
     # If there are enough active validators to form committees for every slot
     if active_validators_size >= cycle_length * min_committee_size:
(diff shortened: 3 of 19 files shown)


djrtwo:
@pipermerriam I'm going to review this PR from a spec validity perspective

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+
+        The proposer broadcasts zir attestation with the proposed block.
+        """
+        block_committees_info = get_block_committees_info(
+            block,
+            post_crystallized_state,
+            cycle_length,
+        )
+        # Vote
+        attester_bitfield = set_voted(
```
pipermerriam: Not for this PR, but it might be nice to create a class to represent the voting bitfield

```python
class VoteBitfield(bytes):
    @classmethod
    def create_empty(cls, size):
        ...

    def set_voted(self, idx):
        ...

    def get_vote(self, idx):
        ...
```

I don't necesarrily think those are the best APIs, just meant to be illustrative.

API for interacting with it could be similar to the `Receipt.bloom_filter`.  Also, might be nice for the class to be immutable so that each method call returns the updated bitfield, allowing you to chain multiple calls.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+            get_empty_bitfield(block_committees_info.proposer_committee_size),
+            block_committees_info.proposer_index_in_committee,
+        )
+
+        # Get justified_slot and justified_block_hash
+        justified_slot = post_crystallized_state.last_justified_slot
+        justified_block_hash = chaindb.get_canonical_block_by_slot(justified_slot).hash
```
pipermerriam: It may be nice to have both `get_canonical_block_hash_by_slot` and `get_canonical_block_by_slot` so that you can save the extra overhead of instantiating the block and then hashing it.  It's likely that the implementation already has to lookup the hash in order to lookup the block body.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+            block.shard_block_hash,
+            crystallized_state.last_justified_slot,
+        )
+        # Update the bitfield and append the signatures
+        bitfield = proposer_attestation.bitfield
+        sigs = []
+        for (committee_index, sig, public_key) in votes:
+            if bls.verify(message, public_key, sig):
+                bitfield = set_voted(bitfield, committee_index)
+                sigs.append(sig)
```
pipermerriam: Just looking at what this looks like without the mutation

```python
sigs_with_committe_info = tuple(
    sig, committee_index
    for (committee_index, sig, public_key)
    in votes
    if bls.verify(message, public_key, sig)
)
sigs, commitee_indices = zip(*sigs_with_committe_info)   # may need to handle edge case of empty list
bitfield = pipe(
    proposer_attestation.bitfield
    *(functools.partial(set_voted, index=committee_index) for committee_index in committee_indices)
)
```

That's all psuedo code, probably some minor errors, but it removes the mutation and seems readable enough).  The `functools.partial` can be removed if `set_voted` was made curriable.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+        if block.slot_number == 0:
+            return
+
+        block_committees_info = get_block_committees_info(
+            parent_block,
+            crystallized_state,
+            cycle_length,
+        )
+
+        if len(block.attestations) == 0:
```
pipermerriam: Rather than pre-check, I think the more idiomatic way to do this is:

```python
try:
    attestation = block.attestations[0]
except IndexError:
    raise ValidationError(...)
```

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+        )
+        cls.validate_aggregate_sig(
+            crystallized_state,
+            attestation,
+            attestation_indices,
+            parent_hashes,
+        )
+
+    @classmethod
+    def validate_slot(cls,
```
pipermerriam: This looks like it can be a staticmethod which also suggests it could be extracted to a standalone to simplify the class (but still tacked onto the class as a `staticmethod` for convenience if you want).

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+                "Attestation slot number too low:\n"
+                "\tFound: %s, Needed greater than or equalt to: %s" %
+                (
+                    attestation.slot,
+                    max(parent_block.slot_number - cycle_length + 1, 0)
+                )
+            )
+
+    @classmethod
+    def validate_justified(cls,
```
pipermerriam: Same about `staticmethod`, extraction to standalone.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+                "justified_block_hash %s is not in the canonical chain" %
+                attestation.justified_block_hash
+            )
+        if justified_block.slot_number != attestation.justified_slot:
+            raise ValidationError(
+                "justified_slot %s doesn't match justified_block_hash" % attestation.justified_slot
+            )
+
+    @classmethod
+    def validate_bitfield(cls,
```
pipermerriam: Same about `staticmethod`, extraction to standalone.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+        Return the :class:`~eth.beacon.types.attestation_records.AttestationRecord` class that this
+        StateMachine uses for the current fork version.
+        """
+        if cls.attestation_record_class is None:
+            raise AttributeError("No `attestation_record_class` has been set for this StateMachine")
+        else:
+            return cls.attestation_record_class
+
+    @classmethod
+    def create_signing_message(cls,
```
pipermerriam: This looks like it could be extracted to a standalone function (which will simplify this class which is quite large).  Or, at minimum, it can be made into a `staticmethod`

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+
+        # check if end bits are zero
+        last_bit = len(attestation_indices)
+        if last_bit % 8 != 0:
+            for i in range(8 - last_bit % 8):
+                if has_voted(attestation.attester_bitfield, last_bit + i):
+                    raise ValidationError("Attestation has non-zero trailing bits")
+
+    @classmethod
+    def validate_aggregate_sig(cls,
```
pipermerriam: same about `staticmethod` -> extraction to standalone (since `create_signing_message` is also able to be made standalone.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+        )
+        if not bls.verify(message, bls.aggregate_pubs(pub_keys), attestation.aggregate_sig):
+            raise ValidationError("Attestation aggregate signature fails")
+
+    @classmethod
+    def validate_version(cls,
+                         crystallized_state: CrystallizedState,
+                         attestation: 'AttestationRecord') -> None:
+        # TODO: it's a stub
+        return
```
pipermerriam: *nitpick*, normally `pass` instead of `return`

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+
+    #
+    # State roots validation
+    #
+    @classmethod
+    def validate_state_roots(cls,
```
pipermerriam: Same about extraction to standalone.

tests/beacon/state_machines/test_proposer.py
```diff
@@ -0,0 +1,149 @@
...
+
+from eth.beacon.helpers import (
+    get_block_committees_info,
+    get_new_recent_block_hashes,
+)
+
+
+@pytest.fixture
+def fixture_sm_class(config):
+    return SerenityBeaconStateMachine.configure(
```
pipermerriam: dropping the `Beacon` from the class name would make this a bit less of a mouthful to read: `SerenityStateMachine`

fits with the EVM naming convention since we don't name things `FrontierEVMVM` or `FrontierChainVM`

tests/beacon/state_machines/test_proposer.py
```diff
@@ -0,0 +1,149 @@
...
+        sm.crystallized_state,
+        sm.active_state,
+        block_1_shell,
+        sm.chaindb,
+        sm.config,
+    )
+    block_1 = block_1_shell.copy(
+        active_state_root=active_state_1.hash,
+    )
+    sm.import_block(block_1)
```
pipermerriam: I suspect this still has value as a smoke test but surprised to see this test has no assertions.

tests/beacon/state_machines/test_proposer.py
```diff
@@ -0,0 +1,149 @@
...
+@pytest.mark.parametrize(
+    (
+        'num_validators,'
+        'cycle_length,min_committee_size,shard_count'
+    ),
+    [
+        (1000, 20, 10, 100),
+    ]
+)
+def test_propose_block_and_validate_attestation(fixture_sm_class,
```
pipermerriam: Similar lack of assertions (though I recognize that the `validate_` calls are effectively assertions.

eth/beacon/state_machines/base.py
```diff
@@ -55,9 +76,17 @@ class BeaconStateMachine(BaseBeaconStateMachine):
     _crytallized_state = None  # type: CrystallizedState
     _active_state = None  # type: ActiveState
 
-    def __init__(self, chaindb: BaseBeaconChainDB, block: BaseBeaconBlock) -> None:
+    def __init__(self, chaindb: BaseBeaconChainDB, block: BaseBeaconBlock=None) -> None:
         self.chaindb = chaindb
-        self.block = self.get_block_class().from_block(block)
+        if self.block is None:
```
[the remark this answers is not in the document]
hwwhww: Sorry, it's typo. Should be: `if block is None:`

eth/beacon/state_machines/base.py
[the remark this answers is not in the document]
hwwhww: It came from the spec. I think it's kind of a convention in cryptography when we call the thing we sign for? Because in here, the thing we sign with BLS *is* the whole`blake(...)` part.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+                                    chaindb: BaseBeaconChainDB,
+                                    cycle_length: int) -> ActiveState:
+        """
+        Process ``block`` and return the new ActiveState.
+
+
+        TODO: It doesn't match the latest spec.
+        There will be more fields need to be updated in ActiveState.
+        """
+        parent_block = chaindb.get_block_by_hash(block.parent_hash)
```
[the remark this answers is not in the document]
hwwhww: @pipermerriam I left some TODOs hints in [`BeaconChainDB`](https://github.com/ethereum/py-evm/blob/96fc4ede33378823b470f6540245b1fc94752d1d/eth/beacon/db/chain.py#L635-L650). Am I correct that since All PoW chain forks use the same block fields format, so currently EVM doesn't handle it as well?

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+    def propose_block(
+        self,
+        crystallized_state: CrystallizedState,
+        active_state: ActiveState,
+        block: BaseBeaconBlock,
+        shard_id: int,
+        shard_block_hash: Hash32,
+        chaindb: BaseBeaconBlock,
+        config: BeaconConfig,
+        private_key: int
```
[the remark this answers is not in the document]
hwwhww: Yes, the key pairing is in ECC family. But the curve will need to be upgraded; right now we are using `BN128` in the codebase, but will have to change to [`BLS12-381`](https://z.cash/blog/new-snark-curve/).

Boneh–Lynn–Shacham (BLS) is for aggregating the signatures into one aggregated signature, and we only have to verify the aggregated signature with public keys instead of putting every single sigs on chain.

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +222,562 @@ def active_state(self) -> ActiveState:
...
+                              block: BaseBeaconBlock,
+                              shard_id: int,
+                              shard_block_hash: Hash32,
+                              chaindb: BaseBeaconBlock,
+                              cycle_length: int,
+                              private_key: int) -> 'AttestationRecord':
+        """
+        Return the initial attestation by the block proposer.
+
+        The proposer broadcasts zir attestation with the proposed block.
```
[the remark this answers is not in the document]
hwwhww: Ah.. was the gender-neutral possessive... but I can change to `their`.

hwwhww:
### Changelog
1. Fix a lot PR feedback! Thanks @pipermerriam ✨ 
2. Reorg some state machine fixtures.
3. Move most of validation functions into `eth.beacon.state_machines.validation`. Only keep staticmethod `validate_attestation` for overviewing in StateMachine. (Or I should move all?) 
4. Move `create_signing_message` and `aggregate_attestation_record` to `helper.py`.
5. `aggregate_attestation_record` -> two steps (i)  `aggregate_attestation_record` and  (ii) `aggregate_votes`.
6. Add test for `aggregate_votes`. Also found the paring is too slow to add more `hypothesis` case… :|
7. Add assertions to the tests.

eth/beacon/block_committees_info.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+BlockCommitteesInfo = NamedTuple(
+    'BlockCommitteesInfo',
+    (
+        ('proposer_index', int),
+        ('proposer_index_in_committee', int),
+        ('proposer_shard_id', int),
+        ('proposer_committee_size', int),
+        ('shards_and_committees', Iterable['ShardAndCommittee'])
```
pipermerriam: Iterable is probably not what you want since that would allow a generator.  I'd actually recommend being very specific here and choosing `Tuple`.

eth/beacon/helpers.py
```diff
@@ -360,20 +366,117 @@ def get_proposer_position(parent_block: 'BaseBeaconBlock',
...
+    try:
+        sigs, committee_indices = zip(*sigs_with_committe_info)
+    except ValueError:
+        sigs = ()
+        committee_indices = ()
+
+    sigs = sigs + tuple(pre_sigs)
+    bitfield = pre_bitfield
+    bitfield = pipe(
+        bitfield,
```
pipermerriam: You *should* be able to just use `pre_bitfield` here.

tests/beacon/test_helpers.py
```diff
@@ -376,16 +381,60 @@ def mock_get_shards_and_committees_for_slot(parent_block,
...
+        (5),
+        (9),
+    ],
+)
+def test_aggregate_votes(votes_count, privkeys, pubkeys):
+    bit_count = 10
+    pre_bitfield = get_empty_bitfield(bit_count)
+    pre_sigs = ()
+
+    random_votes = random.sample([i for i in range(bit_count)], votes_count)
```
pipermerriam: We've *typically* tried to use `hypothesis` when we need randomness in tests so that the randomness can be reproduced upon failure.  It's not a hard requirement, but in cases where you are using random inputs to test something and the randomness can be a source of failure, it's probably what you should do.

You can do as far as having hypothesis generate values for you, or if you still just want to use the built-in python `random` standard library module, you can use this https://hypothesis.readthedocs.io/en/latest/data.html#hypothesis.strategies.random_module

eth/beacon/block_committees_info.py
cburgdorf: :+1: to what @pipermerriam said.

Taking the chance to elaborate on my comment here: https://github.com/ethereum/py-evm/pull/1373#discussion_r224106008

I think as a rule of thumb: APIs *taking parmeters* (e.g. `def do_something(things: Iterable[Foo])` should accept the most flexible interface that the API can work with. I can hardly think of a reason why an API should constrain a parameter to a concrete type (e.g. `Tuple[Foo]`) if there is nothing intrinsically preventing that API from working with an `Iterable[Foo]`.

However, **Return types** should *usually* be concrete. Because, most of the time an API returning *something*, does have intrinsic knowledge about the type anyway and there is nothing to gain from making the return type more abstract if that return type fulfills that abstract interface anyway.

E.g. There is nothing to gain from doing this:

```python
def get_foo() -> Iterable[Foo]
    return [foo, too]
```

Even though that works, we are not improving flexibility here. If we make the return type `List[Foo]`, that `List[Foo]` still fulfills being `Iterable[Foo]` as well as `Sequence[Foo]` implicitly, so by making the return type somthing like `Iterable[Foo]` when in reality we have a `List[Foo]` is in fact only limiting the flexibility here.

Coming back to the code in question `shards_and_committees` is, after all, just a property on `BlocksCommitteesInfo`, hence making it fall in the return type bucket.

Related: [Covariance and Contravariance](https://en.wikipedia.org/wiki/Covariance_and_contravariance_(computer_science))

carver: Also, here is an alternative syntax, if you like it better:

```py
class BlockCommitteesInfo(NamedTuple):
  proposer_index: int
  proposer_index_in_committee: int
  # etc...
```

cburgdorf: @carver I *think* that syntax does not work with Python 3.5 though

cburgdorf: I'd be in support to drop 3.5 at any time 🤭😅

pipermerriam: Two things,

- Can't drop 3.5 until we do it in web3.py  (is v5 py36+?)
- @cburgdorf would you mind migrating that writup to https://github.com/pipermerriam/ethereum-dev-tactical-manual, maybe under a new section for static type hinting.

carver: #1391 for discussion

hwwhww:
### Changelog
1. Fix `test_aggregate_votes` test and add `verify_votes` function
2. Add `BlockProposal` NamedTuple for having less arguments
3. Move `validate_attestation` to `eth/beacon/state_machines/validation.py`
4. Fix `validate_attestation` (should use updated `recent_block_hashes`) and add tests for every edge cases
5. Propose one more block (`block_2`) in `test_propose_block` - many lines are repeated, will consider have `evm.tools.builder.chain` tool in other PRs! (maybe after we have cycle transitions?)

eth/beacon/state_machines/base.py
```diff
@@ -55,9 +76,17 @@ class BeaconStateMachine(BaseBeaconStateMachine):
...
     _active_state = None  # type: ActiveState
 
-    def __init__(self, chaindb: BaseBeaconChainDB, block: BaseBeaconBlock) -> None:
+    def __init__(self, chaindb: BaseBeaconChainDB, block: BaseBeaconBlock=None) -> None:
         self.chaindb = chaindb
-        self.block = self.get_block_class().from_block(block)
+        if block is None:
+            # Build a child block of current head
+            head_block = self.chaindb.get_canonical_head()
+            self.block = self.get_block_class()(*head_block).copy(
```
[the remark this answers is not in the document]
hwwhww: Right now maybe just in testing.

eth/beacon/helpers.py
```diff
@@ -360,20 +366,133 @@ def get_proposer_position(parent_block: 'BaseBeaconBlock',
...
+        block.slot_number,
+        parent_hashes,
+        proposer_attestation.shard_id,
+        proposer_attestation.shard_block_hash,
+        last_justified_slot,
+    )
+
+    # Verify
+    # TODO: Verification is very slow, needs to compute in parallel
+    voting_sigs, committee_indices = verify_votes(message, votes)
```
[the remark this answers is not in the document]
hwwhww: > What is going to be calling this function? This is for a block producer bundling attestations of the previous block? Note that they should be including attestations not just from the previous block but for any within the valid range (defined by parent block and CYCLE_LENGTH) that have not been included

> Note that they should be including attestations not just from the previous block but for any within the valid range (defined by parent block and CYCLE_LENGTH)

You're right. Verifying the attestations sigs *right after receiving it* makes more sense. (like an AttestationPool)
So upon receiving the Attestation:
1. Call `validate_attestation`
2. Store the attestation in the AttestationPool

When the client wants to aggregate and  the attestation pack: 
1. Check if we can reuse attestations from other block and set `oblique_parent_hashes`.
2. Aggregate the attestations that are siging for the same message.

Anything else I missed?

> We can aggregate and include attestations that are not the same as the proposer_attestation. For example, we can include votes that were for a short range fork rather than for the what the proposer submit as the head. Or some of the votes could vote for the proposer's block but for a different shard_block_hash

Right. I will rename `proposer_attestation` to something more general.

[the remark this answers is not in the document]
hwwhww: > Right. I will rename proposer_attestation to something more general.

@djrtwo 

Second thought - I will just remove `aggregate_attestation_record` from here. It's not part of StateMachine. 😆 

But please check the correctness of the story above. Thanks!

eth/beacon/block_committees_info.py
cburgdorf: @pipermerriam proposed section on typing (choosing types for function parameters and return values)! https://github.com/pipermerriam/ethereum-dev-tactical-manual/pull/24

eth/beacon/helpers.py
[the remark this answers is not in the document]
djrtwo: Looks good. The AttestationPool and who should be doing the aggregation might change depending on the network-level protocol we define for aggregation.

I would note that the AttestationPool should likely only include attestations of single signers (non-aggregated). Also that when we receive a new block (or create one) we should clean up our attestation pool, removing any of the attestations that were included in the block.

> Check if we can reuse attestations from other block
I would be clear that this is "reuse attestations from a block on a short range fork"

> Aggregate the attestations that are signing for the same message.
And can be validly included in the block (not too old, etc)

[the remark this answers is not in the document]
djrtwo: Right, the aggregation is more of the Node or Validator's responsibility.

eth/beacon/aggregation.py
```diff
@@ -0,0 +1,79 @@
...
+    try:
+        sigs, committee_indices = zip(*sigs_with_committe_info)
+    except ValueError:
+        sigs = tuple()
+        committee_indices = tuple()
+
+    return sigs, committee_indices
+
+
+def aggregate_votes(bitfield: bytes,
```
djrtwo: I think it might be more natural to pass in an Iterable of `AttestationRecords` and return an `AttestationRecord`.

Attestations will generally be passed around and deserialized as attestation records rather than deconstructed tuples of sigs and indices

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +224,272 @@ def active_state(self) -> ActiveState:
...
+            crystallized_state: CrystallizedState,
+            active_state: ActiveState,
+            block: BaseBeaconBlock,
+            chaindb: BaseBeaconChainDB,
+            config: BeaconConfig) -> Tuple[BaseBeaconBlock, CrystallizedState, ActiveState]:
+        """
+        Process ``block`` and return the new crystallized state and active state.
+        """
+        # Process per block state changes (ActiveState)
+        processing_active_state = cls.compute_per_block_transtion(
```
djrtwo: Typo
`compute_per_block_transtion` -> `compute_per_block_transition`

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +224,272 @@ def active_state(self) -> ActiveState:
...
+            validate_attestation(
+                block,
+                parent_block,
+                crystallized_state,
+                recent_block_hashes,
+                attestation,
+                chaindb,
+                cycle_length,
+            )
+
```
djrtwo: TODO: validate `block.specials` and append `active_state.pending_specials`

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +224,272 @@ def active_state(self) -> ActiveState:
...
+        active_state: ActiveState,
+        block_proposal: 'BlockProposal',
+        chaindb: BaseBeaconChainDB,
+        config: BeaconConfig,
+        private_key: int
+    ) -> Tuple[BaseBeaconBlock, CrystallizedState, ActiveState, 'AttestationRecord']:
+        """
+        Propose the given block.
+        """
+        block, post_crystallized_state, post_active_state = self.process_block(
```
djrtwo: If `process_block` is to be used for calculating the state roots for a block proposal, then it needs to have the option of being run without processing the signatures from the attestations. maybe another arg -- `validate_signatures`

We can assume the attestations being added were already checked as valid when initially received/created. No need to recheck here!

eth/beacon/state_machines/base.py
```diff
@@ -193,9 +224,272 @@ def active_state(self) -> ActiveState:
...
+            block_proposal.shard_block_hash,
+            justified_slot,
+        )
+        sigs = [
+            bls.sign(
+                message,
+                private_key,
+            )
+        ]
+        aggregate_sig = bls.aggregate_sigs(sigs)
```
djrtwo: No need to aggregate the single sig. A sig and an aggregate sig are indistinguishable in BLS

eth/beacon/state_machines/validation.py
```diff
@@ -0,0 +1,268 @@
...
+# Parent block proposer validation
+#
+def validate_parent_block_proposer(crystallized_state: 'CrystallizedState',
+                                   block: 'BaseBeaconBlock',
+                                   parent_block: 'BaseBeaconBlock',
+                                   cycle_length: int) -> None:
+    if block.slot_number == 0:
+        return
+
+    block_committees_info = get_block_committees_info(
```
djrtwo: I really like the BlockCommittee info :)

tests/beacon/test_aggregation.py
```diff
@@ -0,0 +1,68 @@
...
+        (0),
+        (9),
+    ],
+)
+def test_aggregate_votes(votes_count, random, privkeys, pubkeys):
+    bit_count = 10
+    pre_bitfield = get_empty_bitfield(bit_count)
+    pre_sigs = ()
+
+    random_votes = random.sample([i for i in range(bit_count)], votes_count)
```
ChihChengLiang: nitpick: could `[i for i in range(bit_count)]` be replaced with ` range(bit_count)`?
```python
>>> import random
>>> random.seed(5566)
>>> random.sample([i for i in range(10)], 5)
[0, 9, 2, 8, 4]
>>> random.seed(5566)
>>> random.sample(range(10), 5)
[0, 9, 2, 8, 4]
```

hwwhww:
@djrtwo @ChihChengLiang thanks for the feedback! ✨ 
I'm going to merge this PR and make other issues for following up attestation aggregation.

Issue #1359, opened 2018-10-04
Have at least some benchmarks run against the actual DB rather than the MemoryDB


cburgdorf:
### What is wrong?

Currently, all our benchmarks run against the `MemoryDB`. Unfortunately, that means we are missing out to measure pretty much all optimizations that are about reducing the number of hits to read/write to the database.

In other words, if some optimizations save e.g. two hits to the DB we will barely see any difference in the benchmarks as data updates in-memory are just super fast compared to the actual cost when we hit the disk (which will be the case with something like `LevelDB`.

### How can it be fixed

Run at least *some* benchmarks against `LevelDB`. I say *some* because the one argument against this, is that it will increase our CI times.

pipermerriam:
two things:

1. we can help the current state of affairs a little by promoting our benchmarking CI runs to run earlier in the process.  That's more of a scheduling optimization but it will help.
2. in CI, we can reduce the granularity that our benchmarks run, only running the most *important* ones and then maybe a *random* subset of the other ones (ideally deterministically random so that any failures are reproducable within a branch).

veox:
A few issues with running benchmarks in remote-hosted CI:

* you never know how "busy" the VM host is serving other VM guests - there's need to measure the "raw" device performance, too, so the benchmark is normalised to something comparable between runs;
* you never know if the remote CI is completely memory-hosted, where the "disk drive" is a "RAM disk";
* (there was something else, but I forgot :disappointed:).

pipermerriam:
It's less about the actual benchmark performance and more about making sure the benchmarks work.

Bhargavasomu:
What should I use as the `path` for the `LevelDB initialization`? Is it `/tmp/mainnet.db`?

pipermerriam:
I'd recommend using the [tmpdir](https://docs.pytest.org/en/latest/tmpdir.html) fixture that comes with pytest or the [`tempfile`](https://docs.python.org/3.5/library/tempfile.html) standard library module to generate a new temporary directory for each run.  If you use the standard library `tempfile` module, make sure that the directory is removed after the test.

Bhargavasomu:
![screenshot from 2018-10-14 10-28-43](https://user-images.githubusercontent.com/15922456/46912760-1c36c780-cf9c-11e8-8758-2e6a6440c6f2.png)
How to fix this issue in Circle CI?

pipermerriam:
I've restarted those tests (looks like you have a linting error though)

[...]

PR #1393, opened 2018-10-14
Memory db to level db

Bhargavasomu:
### What was wrong?
Currently, all our benchmarks run against the `MemoryDB`. Unfortunately, that means we are missing out to measure pretty much all optimizations that are about reducing the number of hits to read/write to the database.

In other words, if some optimizations save e.g. two hits to the DB we will barely see any difference in the benchmarks as data updates in-memory are just super fast compared to the actual cost when we hit the disk (which will be the case with something like `LevelDB`.

Issue : #1359 

### How was it fixed?
While creating the chains, for each benchmark the chains were created one with `MemoryDB` and one with `LevelDB` in `chain_plumbing.py` which is the base for all the benchmarks.

#### Cute Animal Picture

### DIFF
--- a/scripts/benchmark/utils/chain_plumbing.py
+++ b/scripts/benchmark/utils/chain_plumbing.py
@@ -2,11 +2,9 @@ from pathlib import (
     Path,
 )
 import tempfile
-
 from typing import (
     Any,
     Dict,
-    Generator,
     Iterable,
     Tuple,
     Type,
@@ -90,9 +88,8 @@ DEFAULT_GENESIS_STATE = [
 GenesisState = Iterable[Tuple[Address, Dict[str, Any]]]
 
 
-def get_chain(vm: Type[BaseVM], genesis_state: GenesisState) -> Generator[MiningChain, None, None]:
+def get_chain(vm: Type[BaseVM], genesis_state: GenesisState) -> Iterable[MiningChain]:
 
-    # Chain uses LevelDB
     with tempfile.TemporaryDirectory() as temp_dir:
         level_db_obj = LevelDB(Path(temp_dir))
         level_db_chain = build(
@@ -106,5 +103,4 @@ def get_chain(vm: Type[BaseVM], genesis_state: GenesisState) -> Generator[Mining
 
 def get_all_chains(genesis_state: GenesisState=DEFAULT_GENESIS_STATE) -> Iterable[MiningChain]:
     for vm in ALL_VM:
-        chain = get_chain(vm, genesis_state)
-        yield next(chain)
+        yield from get_chain(vm, genesis_state)
--- a/scripts/benchmark/utils/chain_plumbing.py
+++ b/scripts/benchmark/utils/chain_plumbing.py
@@ -1,6 +1,12 @@
+from pathlib import (
+    Path,
+)
+import tempfile
+
 from typing import (
     Any,
     Dict,
+    Generator,
     Iterable,
     Tuple,
     Type,
@@ -25,8 +31,11 @@ from eth import (
 from eth.chains.base import (
     MiningChain,
 )
+from eth.db.backends.level import (
+    LevelDB,
+)
 from eth.vm.base import (
-    BaseVM
+    BaseVM,
 )
 from eth.chains.mainnet import (
     BaseMainnetChain,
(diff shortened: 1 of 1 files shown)


scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -81,19 +92,28 @@
...
+        memory_db_obj = AtomicDB()
+        memory_db_chain = get_chain(vm, memory_db_obj, genesis_state)
+        yield memory_db_chain
+
+        # Chain using LevelDB
+        with tempfile.TemporaryDirectory() as temp_dir:
+            level_db_obj = LevelDB(temp_dir)
+            level_db_chain = get_chain(vm, level_db_obj, genesis_state)
+            yield level_db_chain
+        del temp_dir
```
pipermerriam: I think this is not necessary since you are using the context manager API to get the temporary directory.

Bhargavasomu: You mean the `del temp_dir` part right?

Bhargavasomu:
@pipermerriam could you please re-run the tests which is showing `GitHub rate limited`

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -81,19 +93,27 @@
...
     for vm in ALL_VM:
-        chain = get_chain(vm, genesis_state)
-        yield chain
+        # Chain using MemoryDB
+        memory_db_obj = AtomicDB()
+        memory_db_chain = get_chain(vm, memory_db_obj, genesis_state)
+        yield memory_db_chain
+
+        # Chain using LevelDB
+        with tempfile.TemporaryDirectory() as temp_dir:
```
cburgdorf: Looking at the CI run, this doesn't increase the time these benchmarks take as much as I had feared. With that in mind, I'd say we should default to run all tests using `LevelDB` and simplify the logic here. Basically, I think `get_all_chains` can roll back to what it looked like before and the `LevelDB` setup code can move into `get_chain`.

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -1,3 +1,6 @@
+import tempfile
```
cburgdorf: Ordering within each block (stdlib, 3rd party, own) should be alphabetically (See https://github.com/pipermerriam/ethereum-dev-tactical-manual/blob/master/style-guide.md#import-statements) so `pathlib` should move up. Also, for consistency let's make it

```
from pathlib import (
    Path,
)
```

Bhargavasomu:
@cburgdorf I have made the recommended changes, and winded up `LevelDB` chain in `get_chain()`. The output of the runner is as following.
screenshot from 2018-10-16 07-56-05
The second duplicate name for each Chain is the `LevelDB benchmark`. We should probably structure the output format so that there is indication of which DB the benchmark is using.

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -81,19 +93,32 @@
 GenesisState = Iterable[Tuple[Address, Dict[str, Any]]]
 
 
-def get_chain(vm: Type[BaseVM], genesis_state: GenesisState) -> MiningChain:
+def get_chain(vm: Type[BaseVM], genesis_state: GenesisState) -> Iterable[MiningChain]:
 
-    chain = build(
+    # Chain uses MemoryDB
```
cburgdorf: Sorry, I probably wasn't clear with my previous comment. I meant we can drop usage of `MemoryDB` entirely and just go with `LevelDB` exclusively. The only reason `MemoryDB` was used at all is because:

a.) it was simple to start with
b.) it was thought that running benchmarks against `LevelDB` may be increase benchmark time too much

Turned out wrong, so it doesn't seem there's a good reason to keep `MemoryDB` at all here.

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -81,19 +93,32 @@
...
+            genesis(db=level_db_obj, params=GENESIS_PARAMS, state=genesis_state)
+        )
+        yield level_db_chain
 
 
 def get_all_chains(genesis_state: GenesisState=DEFAULT_GENESIS_STATE) -> Iterable[MiningChain]:
     for vm in ALL_VM:
-        chain = get_chain(vm, genesis_state)
-        yield chain
+        chains = get_chain(vm, genesis_state)
```
cburgdorf: This can roll back to just yield one single chain

scripts/benchmark/utils/chain_plumbing.py
```diff
@@ -1,3 +1,9 @@
+from pathlib import (
+    Path,
+)
+
```
cburgdorf: *nitpick:* line breaks should only appear between the three groups: stdlib, 3rd party, own but not within imports of each group.

Bhargavasomu:
@cburgdorf looks like we have to `yield` the chain instead of `return`, because we are using the `tempfile` with `context manager`. When we do a `return`, the file gets deleted (`DB doesn't exist anymore`) and the subsequent `benchmark life cycle` may not run properly.

cburgdorf:
>looks like we have to `yield` the chain instead of `return`

@Bhargavasomu yep, that's right!

Bhargavasomu:
@cburgdorf there is an issue with `Circle CI` limiting the checkmark of the `py36-native-blockchain` testcase. The details of the execution shows success but it is not being updated here. Could you please rerun that testcase as I don't have access to it.

cburgdorf:
@Bhargavasomu yep, that Circle CI issue is annoying. We already contacted support.

I added a minor cleanup in a separate commit for you to see the changes. I also squashed your commits into one to keep the history tidy. Will merge as soon as it goes green. Thank you for your contribution!

Issue #1024, opened 2018-07-17
Trinity client identifier should include git commit ID when in dev mode


gsalgado:
### What is wrong?

When developing it's important to know the commit ID that was in use on a given trinity run

### How can it be fixed

Maybe change `construct_trinity_client_identifier()` to include that when available?

PR #1394, opened 2018-10-14
Include Git Commit ID in Trinity Version String

kev946:
Addresses #1024

### DIFF
--- a/trinity/__init__.py
+++ b/trinity/__init__.py
@@ -1,14 +1,37 @@
-import pkg_resources
+import re
+import subprocess
 import sys
+from pathlib import Path
+
+import pkg_resources
+
+
+def find_version(setup_trinity_path: Path) -> str:
+    version = ''
+    try:
+        text = setup_trinity_path.read_text()
+        version_match = re.search(r"version=['\"]([^'\"]*)['\"]", text)
+        if version_match:
+            version = version_match.group(1)
+    except IOError:
+        pass
+    return version
+
 
 # TODO: update this to use the `trinity` version once extracted from py-evm
 __version__: str
+SETUP_FILE_PATH = Path(__file__).parent.parent
 try:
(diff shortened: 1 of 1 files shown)


trinity/__init__.py
```diff
@@ -1,14 +1,46 @@
-import pkg_resources
+import os
+import re
+import subprocess
 import sys
 
+import pkg_resources
+
+
+def read(*parts: str) -> str:
```
pipermerriam: I think you can drop this function and instead do:

```python
from pathlib import Path
SETUP_FILE_PATH = Path(__file__).parent.parent / "setup_trinity"
```

Then in `find_version`:

```python
def find_version(trinity_setup_path: Path) -> str:
    version_file = trinity_setup_path.read_text()
```

We've been using `pathlib.Path` objects more and more and they are very nice to work with.

trinity/__init__.py
```diff
@@ -1,14 +1,46 @@
...
+        return version_match.group(1)
+    return ''
+
+
+def get_git_commit_id(path: str) -> str:
+    try:
+        return subprocess.check_output(['git', '-C', path, 'rev-parse', 'HEAD'],
+                                       stderr=subprocess.DEVNULL, encoding='UTF-8')
+    except subprocess.CalledProcessError:
+        return ''
```
pipermerriam: I think I'd rather the exception be handled at the call site than within this function.  Otherwise there is *some* ambiguity as to whether the subprocess check actually *somehow* returned an empty string or whether the call threw an exception.  Minor semantics, but I think it's a slightly better approach.

pipermerriam:
TODO: weird test failures that seem unrelated.  Also, I'd like some tests added to verify that this code behaves well in both development environments and in cases where the package is installed (and thus doesn't have access to the `trinity_setup.py` file.

pipermerriam:

fselmo:
stale

PR #1400, opened 2018-10-16
Add `compute_cycle_transitions`

hwwhww:
### What was wrong?
#1381

Before adding the [whole cycle transition process](https://github.com/ethereum/eth2.0-specs/blob/master/specs/beacon-chain.md#state-recalculations-every-cycle_length-slots), first, adding `compute_cycle_transitions`.

### How was it fixed?
1. `compute_cycle_transitions` does two things:
    1. Check if `block.slot_number >= crystallized_state.last_state_recalc + config.CYCLE_LENGTH`; if yes: 
        - Call `compute_per_cycle_transition` (a.k.a. `initialize_new_cycle` in `beacon_chain` repo) to compute new cycle and get updated states.
            - TODO in other PRs: `compute_crosslinks` will call `compute_crosslinks` and `apply_rewards_and_penalties`.
        - Check if it's ready for dynasty transition.
            - If yes, call `compute_dynasty_transition`
    2. Return the updated states.
2. Renamed `total_deposits` to `total_balance` as the latest spec.

#### Cute Animal Picture

512px-cheney s_dogs_dressed_for_halloween_by david bohrer

### DIFF
--- a/eth/beacon/state_machines/base.py
+++ b/eth/beacon/state_machines/base.py
@@ -385,6 +385,11 @@ class BeaconStateMachine(BaseBeaconStateMachine):
                 block,
             )
 
+            # TODO: it's a STUB before we implement compute_per_cycle_transition
+            crystallized_state = crystallized_state.copy(
+                last_state_recalc=crystallized_state.last_state_recalc + cls.config.CYCLE_LENGTH
+            )
+
             if cls.ready_for_dynasty_transition(crystallized_state, block):
                 crystallized_state = cls.compute_dynasty_transition(
                     crystallized_state,
@@ -409,10 +414,10 @@ class BeaconStateMachine(BaseBeaconStateMachine):
     # Crosslinks
     #
     @classmethod
-    def compute_crosslinks(cls,
-                           crystallized_state: CrystallizedState,
-                           active_state: ActiveState,
-                           block: BaseBeaconBlock) -> Tuple['CrosslinkRecord', ...]:
+    def update_crosslinks(cls,
+                          crystallized_state: CrystallizedState,
+                          active_state: ActiveState,
+                          block: BaseBeaconBlock) -> Tuple['CrosslinkRecord', ...]:
         # TODO
         return ()
 
--- a/eth/beacon/state_machines/base.py
+++ b/eth/beacon/state_machines/base.py
@@ -260,7 +260,6 @@ class BeaconStateMachine(BaseBeaconStateMachine):
             self.active_state,
             block,
             self.chaindb,
-            self.config,
             is_validating_signatures=True,
         )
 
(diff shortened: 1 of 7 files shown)


eth/beacon/state_machines/base.py
```diff
@@ -303,17 +304,18 @@ def process_block(
         )
 
         # Process per cycle state changes (CrystallizedState and ActiveState)
-        processed_crystallized_state, processed_active_state = cls.compute_cycle_transitions(
+        processed_crystallized_state, processed_active_state = self.compute_cycle_transitions(
             crystallized_state,
             processing_active_state,
+            block,
+            config,
```
pipermerriam: Making note that passing the `config` object around like this seems like an API that could be improved.  IIUC the `BeaconConfig` should be both immutable and static within the context of a single fork.  For this reason, does it make sense to make this a class property?

I'm starting to like this pattern and am considering what it would look like to port it over to the main `Chain` and `VM` APIs.  Many of the `constants` would become config properties which would ideally reduce some of the reliance on inheritance by giving each `VM` it's own `config` object (think replacing `VM.get_block_reward()` with `VM.config.block_reward`.

eth/beacon/state_machines/base.py
```diff
@@ -368,33 +373,108 @@ def compute_per_block_transition(cls,
...
             cls,
             crystallized_state: CrystallizedState,
-            active_state: ActiveState) -> Tuple[CrystallizedState, ActiveState]:
-        # TODO: it's a stub
+            active_state: ActiveState,
+            block: BaseBeaconBlock) -> Tuple[CrystallizedState, ActiveState]:
+        """
+        Compute the cycle transitions and return processed CrystallizedState and ActiveState.
+        """
+        while block.slot_number >= crystallized_state.last_state_recalc + cls.config.CYCLE_LENGTH:
```
djrtwo: This `while` loop will never end with the current stub of `compute_per_cycle_transition`.

Consider updating the stub to at least update `new_crystallized_state.last_state_recalc += crystallized_state.last_state_recalc + CYCLE_LENGTH`

eth/beacon/state_machines/base.py
```diff
@@ -368,33 +373,108 @@ def compute_per_block_transition(cls,
...
+        Initialize a new cycle.
+        """
+        # TODO
         return crystallized_state, active_state
 
+    #
+    # Crosslinks
+    #
+    @classmethod
+    def compute_crosslinks(cls,
```
djrtwo: Name might be better `update_crosslinks`

eth/beacon/state_machines/base.py
djrtwo: and add a sanity check test that we can go through a cycle transition without it hanging in the while loop

pipermerriam: **if** it is possible to determine up-front how many iterations this will take (which I recognize may not be possible) I think it might be prudent to do that so that we can write this loop in a manner that is guaranteed to terminate.

Alternatively, can we compute a reasonable upper bound and set the loop to only run that long.  The inner loop can use a `break` and then we can add an exception to the loop's `else` clause to prevent a possible infinite loop.

All of this assumes that the algorithms involved here are sufficiently complex that we won't be **really** sure this loop's exit condition will be met which currently seems to be the case.

hwwhww: > if it is possible to determine up-front how many iterations this will take

I think yes, if no accident, it should be `(block.slot_number - crystallized_state.last_state_recalc) // cls.config.CYCLE_LENGTH` times since according to the spec, the `crystallized_state.last_state_recalc` will be updated to `crystallized_state.last_state_recalc + cls.config.CYCLE_LENGTH` at the end of `compute_per_cycle_transition` too.

hwwhww:
@pipermerriam @djrtwo ready for review again, thanks!

eth/beacon/state_machines/base.py
pipermerriam: I see you pinged me for review but I don't see any of the suggested changes to this loop's structure.  Is that being done separately?

djrtwo:
Approved

Issue #1398, opened 2018-10-16
Add type hints to eth module


cburgdorf:
### Background

[Type hints](https://www.python.org/dev/peps/pep-0484/) allow us to perform static type checking, among other things. They raise the security bar by catching bugs at development time that, without type support, may turn into runtime bugs.

[This stackoverflow answer](https://stackoverflow.com/questions/32557920/what-are-type-hints-in-python-3-5/32558710#32558710) does a great job at describing their main benefits.

### What is wrong?

We currently enforce type hints for the entire `p2p` and `trinity` package. While the `eth` package does have some type hints already, the coverage isn't at 100 % and while it is agreed up on that new code needs to land with type hints, there is not technical enforcement of this rule in place yet.

This needs to be fixed by:

1. Adding all missing type hints.
2. Enforcing (stricter) type checking in CI

### How

There does exist tooling ([monkeytype](https://engineering.instagram.com/let-your-code-type-hint-itself-introducing-open-source-monkeytype-a855c7284881)) to the generation of type hints for existing code bases. From my personal experience `monkeytype` *can* be helpful but does still require manual fine tuning. Also, manually adding these type hints does serve as a great boost to the general understanding of the code base as it forces one to think about the code.

1. Run `mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth`

2. Eliminate every reported error by adding the right type hint

3. This should probably be done incrementally in roughly the following steps.

- `eth.tools` (eligible for payout of 150 DAI)
- `eth.vm` (eligible for payout of 350 DAI)
- everything that is not `eth.vm` (eligible for payout of 300 DAI)

The reason for this order is, that it makes it easy to incrementally update the `tox.ini` to enforce stricter rules for such each package without cluttering `tox.ini` too much. 

4. Send frequent pull requests for chunks of work

### Definition of done

This issue is done when the following criteria are met:

1. The following configuration in `tox.ini` is changed.

https://github.com/ethereum/py-evm/blob/8f74e6b920e1fd4d810724ea842e8a8344ea7094/tox.ini#L107

It needs to be:

```
mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth
```

[...]

PR #1404, opened 2018-10-17
Add type hints For eth/utils module

Bhargavasomu:
### What was wrong?
The `eth` module was not `type hinted` till now. This needs to be fixed and also the `tox.ini` file needs to be updated.
Issue : #1398

### How was it fixed?
Adding `type hints` to the arguments and the function. The phase of adding the type hints is as follows
* eth.utils (Started previously and hence wanted to finish it)
* eth.tools
* eth.vm
* Everything else which is leftover

#### Cute Animal Picture


Bhargavasomu:
@cburgdorf I have done the `type hinting` for `eth/utils` module. Please verify it.

eth/utils/numeric.py
```diff
@@ -69,12 +72,13 @@ def is_odd(value: int) -> bool:
     return value % 2 == 1
 
 
-def get_highest_bit_index(value):
+def get_highest_bit_index(value: int) -> int:
```
cburgdorf: I think the preferred way is:

```python
def get_highest_bit_index(value: int) -> int:
    value >>= 1
    for bit_length in itertools.count():
        if not value:
            return bit_length
        value >>= 1

    raise Exception("Invariant: unreachable code path")
```

eth/utils/rlp.py
```diff
@@ -11,9 +16,15 @@
     ValidationError,
 )
 
+from eth.rlp.blocks import (
+    BaseBlock,
+)
+
+DiffObjectType = Union[None, Iterable[Tuple[str, str, str]]]
```
cburgdorf: `Union[None, Something]` should be equivalent to `Optional[Something]` but the latter reads more naturally

eth/utils/state.py
```diff
@@ -1,14 +1,45 @@
...
+DiffType = Iterable[Tuple[Address, str, Union[int, bytes], Union[int, bytes]]]
+
 
 @to_tuple
-def diff_account_db(expected_state, account_db):
+def diff_account_db(expected_state: AccountState,
+                    account_db: BaseAccountDB) -> DiffType:
+
     for account, account_data in sorted(expected_state.items()):
+        assert isinstance(account_data['balance'], int)
```
cburgdorf: Same here about the `assert` statements. `TypedDict` should help.

tox.ini
```diff
@@ -104,7 +104,8 @@ commands=
     {[common-lint]commands}
     flake8 {toxinidir}/tests --exclude="trinity,p2p"
     # TODO: Drop --ignore-missing-imports once we have type annotations for eth_utils, coincurve and cytoolz
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
+	mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics eth/utils
```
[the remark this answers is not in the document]
pipermerriam: @cburgdorf how about just two mypy runs?  one that has the current config and one that we slowly expand to cover only the modules which have been updated.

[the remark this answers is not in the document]
cburgdorf: @pipermerriam yes, that's my plan. I just think we do not need to do it for each and every submodule (hence,`eth.tools`, `eth.vm`, then the rest) especially if we consider this whole thing to be finished over the course of the next, say, two weeks.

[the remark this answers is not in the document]
Bhargavasomu: @cburgdorf got it. But since I have started on `eth.utils` first unfortunately, how about we have one mypy for the normal configuration that we see today, one for the `eth.utils` submodule and one for the presently working submodule (`eth.tools` and `eth.vm`)?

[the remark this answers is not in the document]
cburgdorf: Cool with me!

Bhargavasomu:
I added the mypy-extensions module to `setup.py`, but still travis is showing "No module named `mypy_extensions`"

cburgdorf:
@Bhargavasomu it comes with `mypy` by default and is already being used here:

https://github.com/ethereum/py-evm/blob/e68e7c36a75ba7287858f479fb82141d656c82e6/trinity/protocol/common/handlers.py#L59-L69

eth/utils/state.py
```diff
@@ -1,14 +1,50 @@
...
+from eth.db.account import (
+    BaseAccountDB,
+)
+
+
+# Mapping from address to account state.
+# 'balance', 'nonce' -> int
+# 'code' -> bytes
+# 'storage' -> Dict[int, int]
+AccountDetails = TypedDict('AccountDetails',
```
cburgdorf: :heart:

tox.ini
```diff
@@ -104,7 +104,9 @@ commands=
     {[common-lint]commands}
     flake8 {toxinidir}/tests --exclude="trinity,p2p"
     # TODO: Drop --ignore-missing-imports once we have type annotations for eth_utils, coincurve and cytoolz
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
+	mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
```
cburgdorf: Looks like you accidentally used tabs instead of four spaces here.

Bhargavasomu:
@cburgdorf, since the whole of `eth` is not type hinted the following command would fail right?
`mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth`
This is present in the present `tox.ini` file, but it would fail eventually because the whole `eth` module is not yet type hinted fully. So shouldn't we remove that command till we finish the whole type hinting? Is there something I am missing?

Bhargavasomu:
1) @cburgdorf then how would we know exactly why the testcases are failing for, because of the requirement of the type hinting of the whole `eth` module or the faults in the linting or some scenario in the present module for which we have made changes for.
2) Also the above test cases are failing because we haven't put the typing for all of the `eth` module. Is that ok for them to fail, till we have completed the type hinting for the whole module.
3) If yes, then can I move onto Type Hinting the other modules (`eth.tools`)?

cburgdorf:
1. Stricter type hints do automatically qualify less strict checks. That said, if you add stricter type hints in e.g. `eth.utils` that may unveil issues in other parts of the code base, hence you see that now some other things outside of `eth.utils` fail. Fixing these can either mean:

a.) something in `eth.utils` is typed incorrectly
b.) something outside of `eth.utils` is typed incorrectly

This is why, even if you aim to just add type hints for module `x` it may involve changing things in module `y` in the same PR. 

>Is that ok for them to fail, till we have completed the type hinting for the whole module.

The rules are:

1. We do not lower type checks at any point
2. We do never merge PRs that fail in CI

To elaborate on the second point. If we decided to merge a PR that fails CI, even if it is only for type hints, it becomes hard / impossible to spot any other CI failures as `master` would be broken for everyone else.

Bhargavasomu:
@cburgdorf I have made the necessary changes so that the present `eth` type hinting is not breaked. I have also added the necessary type hinting for `eth.utils` and `eth.tools.builder`. But the test cases are failing because of `ImportError: No module named 'mypy_extensions'`. Then I took a look at this
screenshot from 2018-10-19 17-41-45
There is no `mypy` installed as part of the `setup` and hence `mypy-extensions` module won't exist also. So I guess we would have to include mypy-extensions as part of the setup ourselves.

cburgdorf:
>Eventhough `mypy` is part of the setup, it is not being installed here

Right, I guess `mypy` is currently only installed for the linting jobs itself. I guess there are two ways to go from here:

1. Add `mypy` as a dependency to `eth`

https://github.com/ethereum/py-evm/blob/e68e7c36a75ba7287858f479fb82141d656c82e6/setup.py#L7-L19

2. Use `if TYPE_CHECKING` to only import from `mypy_extensions` during `mypy` runs

https://github.com/ethereum/py-evm/blob/e68e7c36a75ba7287858f479fb82141d656c82e6/trinity/protocol/common/handlers.py#L58-L59

Honestly, I have no preference. Since we already use `if TYPE_CHECKING` to guard around the `mypy_extensions` you may just continue doing that. On the other hand I don't see why we should treat this library any different than other libraries we depend on.

Feel free to try and pick any of both options. Chances are @carver or @pipermerriam have a stronger preference for one or the other way but it will be trivial to move back and forth so whatever moves this forward is fine by me for now.

eth/utils/rlp.py
```diff
@@ -11,9 +16,15 @@
...
+from eth.rlp.blocks import (
+    BaseBlock,
+)
+
+DiffObjectType = Optional[Iterable[Tuple[str, str, str]]]
+
 
 @to_tuple
-def diff_rlp_object(left, right):
+def diff_rlp_object(left: BaseBlock, right: BaseBlock) -> DiffObjectType:
```
cburgdorf: Since this is only being used once, I'd prefer to use the type in it longer form and just reformat as:

```
def diff_rlp_object(left: BaseBlock,
                             right: BaseBlock) -> Optional[Iterable[Tuple[str, str, str]]]
...
```

eth/utils/state.py
```diff
@@ -1,14 +1,50 @@
+from typing import (
+    Dict,
+    Iterable,
+    Tuple,
+    Union,
+)
+
+from mypy_extensions import (
```
cburgdorf: Based on what you choose from the [stated options](https://github.com/ethereum/py-evm/pull/1404#issuecomment-431346890) you may need to add an `if TYPECHECKING` here to only do that import under `mypy` runs.

eth/utils/state.py
```diff
@@ -1,14 +1,50 @@
...
+    TypedDict,
+)
+
 from eth_utils import (
     to_tuple,
 )
 
+from eth_typing import (
```
cburgdorf: nitpick: `eth_typing` should be above `eth_utils` (alphabetical order within 3rd party libraries)

eth/utils/state.py
```diff
@@ -1,14 +1,50 @@
...
+    Address,
+)
+
+from eth.db.account import (
+    BaseAccountDB,
+)
+
+
+# Mapping from address to account state.
```
cburgdorf: I think we can get rid of this comment now that we are using a `TypedDict`

eth/utils/state.py
```diff
@@ -1,14 +1,50 @@
...
+# 'storage' -> Dict[int, int]
+AccountDetails = TypedDict('AccountDetails',
+                           {'balance': int,
+                            'nonce': int,
+                            'code': bytes,
+                            'storage': Dict[int, int]
+                            })
+AccountState = Dict[Address, AccountDetails]
+
+DiffType = Iterable[Tuple[Address, str, Union[int, bytes], Union[int, bytes]]]
```
cburgdorf: Let's rename this to `AccountDiff`.

tox.ini
```diff
@@ -105,6 +105,9 @@ commands=
     flake8 {toxinidir}/tests --exclude="trinity,p2p"
     # TODO: Drop --ignore-missing-imports once we have type annotations for eth_utils, coincurve and cytoolz
     mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
+    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics eth/utils
```
cburgdorf: Rather than using `eth/utils` and `eth/tools/builder` can you use `-p eth.utils` and `-p eth.tools.builder`

Bhargavasomu:
@cburgdorf sorry for pinging you a lot, but these errors don't seem to be related to the `type hinting`. Some assert statements are failing because of the `fixtures` which I am not able to understand properly. Could you please tell me what to do?

cburgdorf:
>but these errors don't seem to be related to the type hinting. Some assert statements are failing because of the `fixtures`

The reason you are see these is because you (accidentally) changed the fixed commit of the `fixtures` sub repository in your PR to point to newer test cases.

submodule

These newer test cases do fail unless you rebase your branch against latest `master` where changes have been made to make these test cases work.

cburgdorf:
@Bhargavasomu have a look at https://github.com/ethereum/py-evm/pull/1412

1. Resolved `fixtures` issue + rebased against latest master
1. Add `"mypy_extensions>=0.4.1,<1.0.0",` to the `eth` section (instead of `mypy`
2. Change `tox.ini` do install the `eth` dependencies

cburgdorf:
@Bhargavasomu I'm going to drive this home in https://github.com/ethereum/py-evm/pull/1412 and apply some fine tuning on top of it.

You can consider this done and move on adding type hints to other parts. Thank you for pushing this forward. Great work! :+1:  I guess the learnings from this PR will make the upcoming PR chunks easier.

One thing regarding commit history. I'm quite keen on getting clean semantic commits, which means if there are many wip commits, I'll squash them all together into one semantic commit. You can avoid this by cutting semantic commits yourself and force pushing into the PR.

Bhargavasomu:
@cburgdorf sorry for bugging you a lot through this PR. I definitely learnt a lot which would help me type hinting the rest of the parts. Thankyou

cburgdorf:
>sorry for bugging you a lot through this PR. I definitely learnt a lot which would help me type hinting the rest of the parts. Thankyou

No problem! I know that adding types can be quite challenging in the beginning. Especially in `py-evm`. I'm super happy to have you do this and I'm excited for `py-evm` soon be fully typed!

Issue #1105, opened 2018-07-27
EIP1283 - Net gas metering


pipermerriam:
### What is wrong?

~~ttps://eips.ethereum.org/EIPS/eip-1087~~
https://eips.ethereum.org/EIPS/eip-1283 is coming in Constantinople

### How can it be fixed

Implement the changes to how `SSTORE` costs are computed.

veox:
There are now two incompatible proposals (the latter a counter-proposal by Parity Tech due to their implementation specifics):

* [EIP-1087: Net gas metering for SSTORE operations](https://eips.ethereum.org/EIPS/eip-1087)
* [EIP-1283: Net gas metering for SSTORE without dirty maps](https://eips.ethereum.org/EIPS/eip-1283)

-----

A copy-paste from [gitter/ethereum/AllCoreDevs](https://gitter.im/ethereum/AllCoreDevs?at=5b6f2bf4a6af14730b2088a4):

> @veox
> 
> (...)
> 
> **EIP-1087 introduces a map (of `bool`s) for marking "dirty" storage,** but doesn't make comparisons to the original values until the post-processing step. Post-processing can be delayed until a transaction's execution is complete, so that the refund logic can be applied outside the EVM.
> 
> EIP-1283 needs a "cache" (a map, most likely - of `uint256`s) of original storage values, so that it can determine gas_used when doing a write (20000 or 5000). This cache must now necessarily reside within the EVM. In other words:
> 
> **EIP-1283 introduces a map (of `uint256`s) for "original" value look-ups.**
> 
> (The above might not make sense depending on how your favourite implementation delineates "inside the EVM". Or maybe I'm battling a chimera here, to understand the trade-offs. Sorry if noise.)

To which the author of EIP-1283 responds:

> @sorpaas
> 
> Not quite the case, actually. EIP-1283's "map" is needed to be kept for every client right now -- all clients already have this information because they need to handle it in case a transaction reverts. Just like a storage slot's _current value_ and _new value_, most clients won't need any new extra structs to fetch that info.
> 
> The issue with EIP-1087 is that the "dirty map" concept is in conflict with one of the optimization we do after EIP-658, so for parity-ethereum, we can't have that without getting an extra struct.
> 
> So you may say:
> 
> * EIP-1087 needs a map (of `uint256`s) for "original" value look-ups, and introduces a map (of `bool`s) for marking "dirty" storage.
> * EIP-1283 needs a map (of `uint256`s) for "original" value look-ups.
> 
[...]

PR #1410, opened 2018-10-18
EIP-1283

cburgdorf:
### What was wrong?

Need to implement  EIP-1283 #1105 

### How was it fixed?

- Overwritten `SSTORE` opcode in constantinople
- Implemented it by following the spec as closely as possible.
- changed `GasMeter` to allow passing a `RefundStrategy` to allow negative refunds (will balance out within transaction. That said, I wonder if we should introduce a new additional check that raises **if it does not balance out** on the tx level
- added tests

#### Cute Animal Picture


cburgdorf:
Note to self (from spec):

>If an implementation uses “execution-frame level” refund counter (a new refund counter is created at each call frame, and then merged back to parent when the call frame finishes), then the refund counter needs to be changed to signed – at internal calls, a child refund can go below zero.

(This, I believe, why some test cases are not working yet)

eth/vm/logic/storage.py
```diff
@@ -49,6 +49,71 @@ def sstore(computation):
...
+        slot=slot,
+        original=True
+    )
+
+    is_currently_empty = not bool(current_value)
+    is_going_to_be_empty = not bool(value)
+
+    gas_refund = 0
+
+    if current_value == value:
```
pipermerriam: This whole block looks like an excellent candidate for a stand-alone utility function which can be tested in isolation.

Also, if we convert all of the inputs to booleans:

```python
def get_gas_cost_and_refund(original_is_zero, current_is_zero, value_is_zero, original_equals_current, original_equals_value):
    ...
```

We end up with a function that has `2**5 -> 32` possible input combinations for which we should be able to explicitely specify what the expected costs are.

cburgdorf: Unfortunately though, following the spec causes negative refunds for some configurations for us. I believe the reason for this is:

>If an implementation uses “execution-frame level” refund counter (a new refund counter is created at each call frame, and then merged back to parent when the call frame finishes), then the refund counter needs to be changed to signed – at internal calls, a child refund can go below zero.

This seems to apply to us. So, if you refund counter would be a single instance *per transaction*, then we would not get negative refunds, but since our refund counter is per Computation/Message, we get negatives.

That doesn't contradict what you wrote, we could just assert getting negative results for some of the combinations. I still have to taker a closer look to see what would be the best way forward.

pipermerriam: I think that implies that our `GasMeter` may need to be changed since it doesn't allow for negative refunds.  And likely we just need to create a new `ConstantinopleGasMeter` since the logic isn't needed for pre-constantinople forks.

cburgdorf: I'm trying out something more compositional right now but I'm not sure if I like it because I think it doesn't play so well with the rest of our architecture (e.g. we do not use a DI framework which takes the boilerplate out of composition). So I may fall back to what you suggest but I want to give it a chance at least.

eth/vm/forks/constantinople/storage.py
```diff
@@ -0,0 +1,73 @@
...
+from eth.vm.forks.constantinople import (
+    constants
+)
+
+from eth.utils.hexadecimal import (
+    encode_hex,
+)
+
+
+def sstore_eip1283(computation):
```
cburgdorf: I moved this out into `constantinople/storage.py` because I strongly feel cluttering our main code base with fork specific logic is wrong. We currently have a bit of a mix (with the majority of fork specific logic being nicely isolated in `vm/forks/*`) Kinda related to our discussion we had in https://github.com/ethereum/py-evm/issues/1146

eth/vm/gas_meter.py
```diff
@@ -16,17 +17,49 @@
...
 
     gas_refunded = None  # type: int
     gas_remaining = None  # type: int
 
     logger = cast(TraceLogger, logging.getLogger('eth.gas.GasMeter'))
 
-    def __init__(self, start_gas: int) -> None:
+    def __init__(self,
+                 start_gas: int,
+                 refund_strategy: RefundStrategy = default_refund_strategy) -> None:
```
cburgdorf: @pipermerriam @carver I wanted to avoid `ConstantinopleGasMeter` and reached for a more compositional approach that passes in a `RefundStrategy`. I don't have strong feelings about it. If you rather like this to follow the path that the rest of the code base uses (introducing `ConstantinopleGasMeter` and overwriting `refund_gas`) that is cool with me as well. In *general*, I'm more a friend of the compositional approach, it's just that this change is so tiny that I think there isn't much harm from an inheritance approach either.

eth/vm/gas_meter.py
```diff
@@ -16,17 +17,49 @@
...
+
+def allow_negative_refund_strategy(meter: "GasMeter", amount: int) -> None:
+    meter.gas_refunded += amount
+
+    meter.logger.trace(
+        'GAS REFUND: %s + %s -> %s',
+        meter.gas_refunded - amount,
+        amount,
+        meter.gas_refunded,
+    )
```
carver: Looks like these logs can stay in the `GasMeter`

eth/vm/gas_meter.py
carver: Yeah, seems promising, and a reasonable place to experiment with the approach because it's so simple. :+1:

eth/vm/gas_meter.py
```diff
@@ -16,17 +17,49 @@
...
+
+
+def allow_negative_refund_strategy(meter: "GasMeter", amount: int) -> None:
+    meter.gas_refunded += amount
```
carver: So it raises a flag for me when there is a mutual dependency (`GasMeter` on the refund strategy and refund strategy on GasMeter).

One alternative:
```py
def allow_negative_refund_strategy(previous_refund_total: int, amount: int) -> None:
    return previous_refund_total + amount
```

Which makes the strategy both stateless and side-effect-free.

eth/vm/gas_meter.py
```diff
@@ -70,14 +103,4 @@ def return_gas(self, amount: int) -> None:
...
-
-        self.gas_refunded += amount
-
-        self.logger.trace(
-            'GAS REFUND: %s + %s -> %s',
-            self.gas_refunded - amount,
-            amount,
-            self.gas_refunded,
-        )
+        return self.refund_strategy(self, amount)
```
carver: Then this would be something like:
```suggestion
        self.gas_refunded = refund_strategy(self.gas_refunded, amount)
        
        self.logger.trace(	
            'GAS REFUND: %s + %s -> %s',	
            self.gas_refunded - amount,	
            amount,	
            self.gas_refunded,	
        )

        return self.gas_refunded
```

eth/vm/gas_meter.py
cburgdorf: I was wondering about that and then said to myself: If it's really just this single line, why not duplicate it to allow *entirely* swapping out the implementation including the generated log.

eth/vm/gas_meter.py
cburgdorf: I also thought about that but it is limiting the usefulness heavily. E.g. what if you want to implement a strategy that throws if e.g. `gas_refunded` or `gas_remaining` is above some treshold (just making things up). Sure, we could pass these as well but at this point we may just pass the entire meter and let the function inspect whatever it wants?

eth/vm/forks/constantinople/storage.py
pipermerriam: I'm completely onboard with this.  I've got a code reorganization building up in my head to move all of the EVM stuff under `eth.evm` the same way the beacon chain stuff is under `eth.beacon`, which would also meaning moving a lot of stuff thats in `eth.validation`, `eth.constants`, `eth.rlp` etc around a bit to match.

eth/vm/gas_meter.py
carver: Ok, I'm not married to that particular alternative, but the mutual dependency seems fairly undesirable. If there are no other options come to mind, then we can just drop the compositional approach.

cburgdorf: Well ok, maybe designing this in the most flexible way shouldn't be our main priority for now. We can design it in the stateless way you suggest as it is flexible enough to fulfill our current needs and if we ever come across the need for an alternative refund strategy that this system can't handle we can reconsider.

eth/vm/forks/constantinople/constants.py
```diff
@@ -2,5 +2,11 @@
 
 
 GAS_EXTCODEHASH_EIP1052 = 400
+GAS_SSTORE_NOOP_EIP1283 = 200
```
pipermerriam: It'd be ideal if the `EIP1283` namespace was located before the parts that are different.

```python
GAS_SSTORE_EIP1283_INIT = ...
GAS_SSTORE_EIP1283_CLEAN = ...
```

Makes the namespace more effective imho.

eth/vm/forks/constantinople/storage.py
```diff
@@ -0,0 +1,73 @@
...
+    gas_refund = 0
+
+    if current_value == value:
+        gas_cost = constants.GAS_SSTORE_NOOP_EIP1283
+    else:
+        if original_value == current_value:
+            if original_value == 0:
+                gas_cost = constants.GAS_SSTORE_INIT_EIP1283
+            else:
+                gas_cost = constants.GAS_SSTORE_CLEAN_EIP1283
```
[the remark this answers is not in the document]
cburgdorf: I looked into `geth` and `ethereum-js` and I think one more and they all used these names, so I just copied them from there. My preference would be to stick to that name just for the sake of sharing the vocabulary with them.

eth/db/account.py
```diff
@@ -375,8 +375,8 @@ def account_is_empty(self, address):
     #
     # Internal
     #
-    def _get_account(self, address):
-        rlp_account = self._journaltrie.get(address, b'')
+    def _get_account(self, address, from_journal=True):
+        rlp_account = (self._journaltrie if from_journal else self._trie).get(address, b'')
```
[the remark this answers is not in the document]
cburgdorf: Good catch!

eth/vm/forks/constantinople/storage.py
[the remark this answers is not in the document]
carver: Fair enough convention vs clarity is a tightrope walk that I feel ambivalent about. If it were my PR, I'd probably pick the name that I thought was clearest and add a comment in the constant file to reference the conventional name. But I don't feel strongly enough to try to convince you to change it. :D

PR #1412, opened 2018-10-19
Christoph/fix/typehints

cburgdorf:
This is mostly #1404 by @Bhargavasomu with some minimal fine tuning on top.

### What was wrong?
The `eth` module was not `type hinted` till now. This needs to be fixed and also the `tox.ini` file needs to be updated.
Issue : #1398

### How was it fixed?
Adding `type hints` to the arguments and the function. The phase of adding the type hints is as follows
* eth.utils (Started previously and hence wanted to finish it)
* eth.tools
* eth.vm
* Everything else which is leftover

#### Cute Animal Picture

### DIFF
--- a/eth/tools/builder/chain/builders.py
+++ b/eth/tools/builder/chain/builders.py
@@ -1,11 +1,29 @@
 import functools
 import time
-from typing import TYPE_CHECKING
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    List,
+    Tuple,
+    Type,
+    Union,
+)
 
 from cytoolz import (
     curry,
     merge,
     pipe,
+)
+
+from mypy_extensions import (
+    TypedDict,
+)
+
+from eth_typing import (
+    Address,
+    BlockNumber,
 )
 
 from eth_utils import (
@@ -19,10 +37,19 @@ from eth.chains.base import (
     BaseChain,
     MiningChain,
 )
-from eth.db.backends.memory import MemoryDB
 from eth.db.atomic import AtomicDB
-from eth.validation import (
-    validate_vm_configuration,
+from eth.db.backends.base import (
+    BaseAtomicDB,
+)
+from eth.db.backends.memory import (
+    MemoryDB,
+)
+from eth.rlp.blocks import (
+    BaseBlock,
+)
+from eth.rlp.headers import (
+    BlockHeader,
+    HeaderParams,
 )
 from eth.tools.mining import POWMiningMixin
 from eth.tools._utils.mappings import (
@@ -30,6 +57,12 @@ from eth.tools._utils.mappings import (
 )
 from eth.tools._utils.normalization import (
     normalize_state,
+)
+from eth.validation import (
+    validate_vm_configuration,
+)
+from eth.vm.base import (
+    BaseVM,
 )
 from eth.vm.forks import (
     FrontierVM,
@@ -40,11 +73,24 @@ from eth.vm.forks import (
     ConstantinopleVM,
 )
 
-if TYPE_CHECKING:
-    from typing import Dict, Union  # noqa: F401
+
+VMConfigurationType = Iterable[Tuple[int, Type[BaseVM]]]
 
 
-def build(obj, *applicators):
+# 'balance', 'nonce' -> int
+# 'code' -> bytes
+# 'storage' -> Dict[int, int]
+AccountDetails = TypedDict('AccountDetails',
+                           {'balance': int,
+                            'nonce': int,
+                            'code': bytes,
+                            'storage': Dict[int, int]
+                            })
+AccountState = Dict[Address, AccountDetails]
+GeneralStateType = Union[AccountState, List[Tuple[Address, Dict[str, Union[int, bytes, Dict[int, int]]]]]]  # noqa: E501
+
+
+def build(obj: Any, *applicators: Callable[..., Any]) -> Any:
     """
     Run the provided object through the series of applicator functions.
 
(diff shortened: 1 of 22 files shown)


eth/tools/builder/chain/builders.py
```diff
@@ -40,11 +77,12 @@
     ConstantinopleVM,
 )
 
-if TYPE_CHECKING:
-    from typing import Dict, Union  # noqa: F401
 
+VMConfigurationType = Iterable[Tuple[int, Type[BaseVM]]]
+GeneralStateType = Union[AccountState, List[Tuple[Address, Dict[str, Union[int, bytes, Dict[int, int]]]]]]  # noqa: E501
```
pipermerriam: Looks like the inner `Dict` should be a `TypedDict`

pipermerriam: I think the inner `Dict` should be updated to be an `AccountDetals`.

eth/typing.py
```diff
@@ -0,0 +1,26 @@
...
+    Address,
+)
+from mypy_extensions import (
+    TypedDict,
+)
+
+
+# TODO: Move into eth_typing
+
+AccountDetails = TypedDict('AccountDetails',
```
pipermerriam: Ooooh!  `TypedDict` is cool.

eth/utils/module_loading.py
```diff
@@ -1,8 +1,13 @@
...
 from importlib import import_module
 
+from typing import (
+    Any,
+    Tuple,
+)
 
-def import_string(dotted_path):
+
+def import_string(dotted_path: str) -> Any:
```
pipermerriam: can you use `types.ModuleType` here instead of `Any`?

eth/utils/numeric.py
```diff
@@ -12,7 +15,7 @@
 )
 
 
-def int_to_bytes32(value):
+def int_to_bytes32(value: Union[int, bool]) -> bytes:
```
pipermerriam: I think `Hash32` is more appropriate than `bytes`

eth/utils/transactions.py
```diff
@@ -43,7 +47,10 @@ def extract_signature_v(v: int) -> int:
         return V_OFFSET
 
 
-def create_transaction_signature(unsigned_txn, private_key, chain_id=None):
+def create_transaction_signature(unsigned_txn: BaseTransaction,
+                                 private_key: datatypes.PrivateKey,
+                                 chain_id: int=None) -> Tuple[int, int, int]:
```
pipermerriam: candidate for eth-typing `VRS = NewType("VRS", Tuple[int, int, int])`

also, `UINT256 = NewType("UINT256", int)` for these values that are specifically bounded within the 256 bit integer space.

cburgdorf: So `VRS` would then be a `VRS = NewType("VRS", Tuple[UINT256, UINT256, UINT256])`?

pipermerriam: Yes, though `v` *might* have a lower ceiling than the others (not sure what it is, might be a single byte).

cburgdorf: Don't talk low level math with me! What does `v` even stand for? Vitalik? ;)

eth/typing.py
cburgdorf: Yep, they let us avoid sprinkling the code with `assert` statements https://github.com/ethereum/py-evm/pull/1404#discussion_r226073247

eth/utils/transactions.py
pipermerriam: https://bitcoin.stackexchange.com/questions/38351/ecdsa-v-r-s-what-is-v/38909

`v` is the recovery id and looks to be reasonably bounded within a single byte (2**8 or `UINT8`)

cburgdorf:
@Bhargavasomu thanks again for doing the bulk of the work! You may want to check out the few extra commits on top just to get a feel what kind of changes were made.

tox.ini
```diff
@@ -45,7 +45,7 @@ commands=
     native-state-metropolis: pytest {posargs:tests/json-fixtures/test_state.py -k Metropolis}
     lightchain_integration: pytest --integration {posargs:tests/trinity/integration/test_lightchain_integration.py}
 
-deps = .[p2p,trinity,eth-extra,test]
+deps = .[eth,p2p,trinity,eth-extra,test]
```
carver: This is confusing to me: `.` should already be installing those packages:
https://github.com/cburgdorf/py-evm/blob/db8bf2e1003bb18542a3e36ff46fbffabfddc808/setup.py#L102

setup.py
```diff
@@ -12,6 +12,7 @@
         "eth-typing>=2.0.0,<3.0.0",
         "eth-utils>=1.3.0b0,<2.0.0",
         "lru-dict>=1.1.6",
+        "mypy_extensions>=0.4.1,<1.0.0",
```
carver: Feels a little weird, like including `flake8` in the distribution. Don't feel strongly about it though.

cburgdorf: At the end of the day, `mypy_extensions` is a library as any other and we are using it now for things like `TypedDict`. Notice that this is not just a type hint per se. We have to have it in order for code such as the following to work.

https://github.com/ethereum/py-evm/blob/7ad4259f574db2214b03da3e8781dca5bbac04e5/eth/typing.py#L20-L25

tox.ini
cburgdorf: Thanks for pointing that out, now I'm confused as well. Let's see https://github.com/ethereum/py-evm/pull/1417

Bhargavasomu:
@cburgdorf I noticed that you have moved some type definitions to `eth.typing` and importing from there, which is fancy BTW. I haven't known we could do this, definitely will use this method for frequently invoking type definitions

cburgdorf:
Yep, and in the long run we might move some if not all to https://github.com/ethereum/eth-typing

PR #1413, opened 2018-10-19
BeaconChain

hwwhww:
### What was wrong?
#1381 - `BeaconChain`

### How was it fixed?
- Basically copied from PoW `Chain`, but changed `header` to `block`.
- Made `BeaconChainDB.persist_block` return Blocks instead of Hashes
- Added `create_serenity_block_from_parent` as PoW chain does, but at this moment I'm not sure if we will need it.
- Made `BaseBeaconBlock` have some `@overload` initialization functions like: https://github.com/ethereum/py-evm/blob/master/eth/rlp/headers.py.
    - I'm not sure if it's okay to not have type hint in the third `__init__`. But also don't know how to add type hint in this case.
    - If I added the same type hint as the second `__init__`, mypy shows: `Overloaded function implementation does not accept all possible arguments of signature 1`. Should I use [this](https://github.com/python/mypy/issues/4619) workaround? @cburgdorf any chance you can take a look? 🙏 

#### TODO
More tests.

#### Cute Animal Picture

by_jess2284


eth/beacon/chains/base.py
```diff
@@ -0,0 +1,360 @@
...
+    #
+    # Helpers
+    #
+    @classmethod
+    @abstractmethod
+    def get_chaindb_class(cls) -> Type[BaseBeaconChainDB]:
+        pass
+
+    @classmethod
+    def get_sm_configuration(cls) -> Tuple[Tuple[int, Type['BaseBeaconStateMachine']], ...]:
```
pipermerriam: I think you can drop this method, likely getting dropped from the main `Chain` implementation (it was only added to be able to fetch this stuff over a multiprocessing pipe but that is no longer a viable API, see #1299 )

eth/beacon/chains/base.py
```diff
@@ -0,0 +1,360 @@
...
+    @classmethod
+    def get_sm_class(cls, block: BaseBeaconBlock) -> Type['BaseBeaconStateMachine']:
+        """
+        Returns the StateMachine instance for the given block slot number.
+        """
+        return cls.get_sm_class_for_block_slot(block.slot_number)
+
+    @abstractmethod
+    def get_sm(self, block: BaseBeaconBlock) -> 'BaseBeaconStateMachine':
+        raise NotImplementedError("Chain classes must implement this method")
```
pipermerriam: should be `pass`

eth/beacon/chains/base.py
```diff
@@ -0,0 +1,360 @@
...
+            block: BaseBeaconBlock,
+            perform_validation: bool=True
+    ) -> Tuple[BaseBeaconBlock, Tuple[BaseBeaconBlock, ...], Tuple[BaseBeaconBlock, ...]]:
+        pass
+
+
+class BeaconChain(BaseBeaconChain):
+    """
+    A Chain is a combination of one or more StateMachine classes.  Each StateMachine is associated
+    with a range of blocks.  The Chain class acts as a wrapper around these other
```
pipermerriam: `s/blocks/slots/`

eth/beacon/chains/base.py
```diff
@@ -0,0 +1,360 @@
...
+    StateMachine classes, delegating operations to the appropriate StateMachine depending on the
+    current block slot number.
+    """
+    logger = logging.getLogger("eth.beacon.chains.chain.BeaconChain")
```
pipermerriam: extra `chain` in the path, suggest just `eth.beacon.chains.BeaconChain`

hwwhww:
Replace with https://github.com/ethereum/trinity/pull/156

PR #1416, opened 2018-10-19
Base implementation of RocksDB support

pipermerriam:
### What was wrong?

LevelDB doesn't support having multiple processes reading from the same database.  We worked around this by having a dedicated database process which has a direct handle on the database and all other processes interact with it over multiprocessing pipes.

This is not ideal.  It incurs extra overhead both with respect to performance and code complexity.

### How was it fixed?

Implemented a RocksDB backend which uses an alternate underlying database called rocksdb -> https://rocksdb.org/

This PR establishes the database backend class and adds a CLI argument to allow opt-in use of the new database backend.  A subsequent PR will tackle removal of the database process entirely.

#### Cute Animal Picture


eth/db/backends/level.py
```diff
@@ -50,6 +49,9 @@ def _exists(self, key: bytes) -> bool:
         return self.db.get(key) is not None
 
     def __delitem__(self, key: bytes) -> None:
+        v = self.db.get(key)
+        if v is None:
+            raise KeyError(key)
```
pipermerriam: Without this the leveldb backend isn't fully compliant with the `BaseDB` api.

tests/database/test_leveldb_db_backend.py
```diff
@@ -1,66 +0,0 @@
...
-
-def test_raises_if_db_path_is_not_specified(config_env):
-    with pytest.raises(TypeError):
-        get_db_backend()
-
-
-def test_set_and_get(memory_db, level_db):
-    level_db.set(b'1', b'1')
-    memory_db.set(b'1', b'1')
-    assert level_db.get(b'1') == memory_db.get(b'1')
```
pipermerriam: These were removed as it appeared all of these tests are covered by `test_base_db_api.py`

trinity/main.py
```diff
@@ -321,9 +314,9 @@ def kill_trinity_gracefully(logger: logging.Logger,
 
 @setup_cprofiler('run_database_process')
 @with_queued_logging
-def run_database_process(trinity_config: TrinityConfig, db_class: Type[BaseDB]) -> None:
+def run_database_process(trinity_config: TrinityConfig) -> None:
     with trinity_config.process_id_file('database'):
-        base_db = db_class(db_path=trinity_config.database_dir)
+        base_db = trinity_config.db_class(db_path=trinity_config.database_dir)
```
carver: I think rocks and level use names that prevent filepath collisions, but worth double checking.

pipermerriam: I hadn't gotten this far but yes, we should be very sure that we cannot re-use a database directory that was previously used with another backend.  

I thought about fixing this by namespacing the directory based on the backend but... then we could get into a situation where the user switches the backend midstream and effectively has two versions of the same chain.

So maybe just a small file we place somewhere to be able to check what database is being used.

.circleci/config.yml
```diff
@@ -30,7 +30,7 @@ common: &common
           sudo apt-get install -y liblz4-dev libsnappy-dev libgflags-dev zlib1g-dev libbz2-dev libzstd-dev
           if [ ! -f "/root/project/rocksdb/librocksdb.a" ]; then
             git clone https://github.com/facebook/rocksdb
-            cd rocksdb/ && git checkout v5.8.8 && make install-shared INSTALL_PATH=/usr
```
carver: Haha, where are those bash aliases when you need them?

tests/database/test_base_db_api.py
```diff
@@ -80,7 +80,7 @@ def test_database_api_missing_key_retrieval(db):
 
 
 def test_database_api_missing_key_for_deletion(db):
-    db.delete(b'does-not-exist')
```
carver: I think the idea here was that `delete()` kind of mirrors `dict.get` which returns `None` instead of raising an error. ie:
https://github.com/ethereum/py-evm/blob/bad0c8e694eb34e14590461d3a9b06a25624d745/eth/db/backends/base.py#L44-L48

So this line was testing that `.delete()`ing a missing key doesn't crash. Admittedly it wasn't documented anywhere and it should maybe be in its own test. :see_no_evil:

carver:
(After CircleCI kicks off and goes green, of course, plus some commit squashing)

trinity/main.py
pipermerriam: TODO: make sure leveldb and rocksdb databases cannot use the same root directory

pipermerriam:
Alright, last thing is to ensure that the rocksdb build is properly cached (since it takes about 8-9 minutes to build and we don't want to do that in every ci run).

pipermerriam:
I CURRENTLY HATE EVERYTHING.  The `lightchain_integration` tests are failing in ways that they shouldn't..... grumble.

trinity/config.py
```diff
@@ -169,19 +183,49 @@ def data_dir(self) -> Path:
     def data_dir(self, value: str) -> None:
         self._data_dir = Path(value).resolve()
 
+    @property
+    def is_db_backend_leveldb(self) -> bool:
+        return self.db_backend == DB_LEVEL
+
+    @property
+    def is_db_backend_rocksdb(self) -> bool:
+        return self.db_backend == DB_ROCKS
```
carver: I want to say YAGNI about these `is_db_b`... properties, and go back to just running these lines in the if test internally. Do we lose much by dropping the methods?

trinity/main.py
```diff
@@ -168,6 +164,21 @@ def main() -> None:
                 "inside the XDG_TRINITY_ROOT path"
             )
 
+    # Verify that the database engine that trinity is configured to use matches
+    # the existing on-disk engine.
+    if trinity_config.db_backend != trinity_config.on_disk_database_engine:
```
carver: Since both `db_backend` and `database_engine` are storing those constants in `{'rocks', 'level'}`, can we unify the names a bit? I think I like engine better, because `backend` implies to me that I'm going to get a `LevelDB` or `LevelDB()` back instead of `'level'`.

So maybe `config.db_engine` and `config.on_disk_db_engine`?

trinity/utils/chains.py
```diff
@@ -94,6 +95,32 @@ def get_database_socket_path(data_dir: Path) -> Path:
...
+    elif sync_mode == SYNC_LIGHT:
+        return database_base_dir / 'light'
+    else:
+        raise ValueError(
+            f"Unknown sync mode: `{sync_mode}`, must be one of "
+            f"{SYNC_FULL}/{SYNC_LIGHT}"
+        )
+
+
+DATABASE_ENGINE_LOCK_NAME = '.trinity.db_engine.lock'
```
carver: I see what you're saying with `lock`, but also I'm used to lock files being things that should go away after shutdown.

Maybe?
- `.trinity.db_engine.freeze`
- `.trinity.db_engine.frozen`
- `.trinity.db_engine.initialized`
- `.trinity.db_engine.selected`

(and all the other places that use `lock` for this choice)

.circleci/config.yml
```diff
@@ -1,4 +1,4 @@
-version: 2.0
+version: 2.1
```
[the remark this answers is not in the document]
pipermerriam: It's superfluous.  I should remove it.

.circleci/config.yml
```diff
@@ -70,7 +90,7 @@ geth_steps: &geth_steps
           geth version
     - run:
         name: run tox
-        command: ~/.local/bin/tox -r
```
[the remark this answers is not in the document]
pipermerriam: Another thing that didn't get reverted.  will remove.

.circleci/install_rocksdb.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+set -u
```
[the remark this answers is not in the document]
pipermerriam: TIL there is a longform.

docs/guides/trinity/quickstart.rst
```diff
@@ -26,6 +26,12 @@ we need to install the ``python3-pip`` package through the following command.
 
   apt-get install python3-pip
 
+Trinity also requires RocksDB which can be installed with the following command:
+
+.. code:: sh
+
+  apt-get install liblz4-dev lib-rocksdb5.8
```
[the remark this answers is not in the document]
pipermerriam: Not stale.  Requirements for rocksdb installation, will double check but I believe these are directly from their readme.

On a related note, I think our installation story is about to get ungood and we should look into debian packages and/or brew.  We're gonna have to do it eventually...

eth/db/backends/rocks.py
```diff
@@ -0,0 +1,139 @@
...
+        return self.db.get(key) is not None
+
+    def __delitem__(self, key: bytes) -> None:
+        exists, _ = self.db.key_may_exist(key)
+        if not exists:
+            raise KeyError(key)
+        self.db.delete(key)
+
+    @contextmanager
+    def atomic_batch(self) -> Generator['RocksDBWriteBatch', None, None]:
```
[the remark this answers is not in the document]
pipermerriam: likely copy/paste from `LevelDB`.  Will change.

trinity/config.py
pipermerriam: I wrote these because I don't want to have to import the `DB_ROCKS` and `DB_LEVEL` constants in the places that these checks happen.  I ended up doing the same with `is_full_sync` and `is_fast_sync`.  Unless you feel strongly I'd like to leave them.

trinity/utils/chains.py
pipermerriam: went with `marker`

Issue #1402, opened 2018-10-16
ImportWarning: can't resolve package from __spec__ or __package__


carver:
### What is wrong?

When running `trinity`, there are several warnings like this:
```
ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
```

This is one of many different kinds of warnings mentioned in #1370 

The goal is not simply to hide these warnings but to resolve them.

### How can it be fixed

It will take some investigation. You should be able to see stack trace info for the warning by starting trinity with:
```sh
PYTHONWARNINGS=error::ImportWarning trinity
```

Then dig into each offending module to investigate.

[This bug report](https://bugs.launchpad.net/lxml/+bug/1732505) has some ideas about resolving the warnings.

pipermerriam:
Should have been fixed with #1419

Issue #1370, opened 2018-10-07
Syncing throws lots of errors


cburgdorf:
* py-evm Version: x.x.x

Latest master on `6aa10e54`

* OS: osx/linux/win

Linux

* Python Version (python --version):

$ python --version

Python 3.6.5

* Environment (output of `pip freeze`):

```
$ pip freeze
alabaster==0.7.12
apipkg==1.5
argh==0.26.2
asn1crypto==0.24.0
async-generator==1.10
async-lru==0.1.0
asyncio-cancel-token==0.1.0a2
atomicwrites==1.2.1
attrdict==2.0.0
attrs==18.2.0
Babel==2.6.0
backcall==0.1.0
bleach==3.0.0
bloom-filter==1.3
bumpversion==0.5.3
cachetools==2.1.0
certifi==2018.8.24
cffi==1.11.5
chardet==3.0.4
cmarkgfm==0.4.2
coincurve==8.0.2
colorama==0.3.9
coverage==4.5.1
cryptography==2.3.1
cytoolz==0.9.0.1
decorator==4.3.0
docopt==0.6.2
docutils==0.14
eth-abi==1.2.0
eth-account==0.2.3
eth-bloom==1.0.1
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.0b3
eth-rlp==0.1.2
eth-typing==1.3.0
eth-utils==1.2.2
execnet==1.5.0
flake8==3.5.0
future==0.16.0
hexbytes==0.1.0
hypothesis==3.69.5
idna==2.7
ifaddr==0.1.4
imagesize==1.1.0
ipaddress==1.0.22
ipython==6.5.0
ipython-genutils==0.2.0
jedi==0.13.1
Jinja2==2.10
lahja==0.8.0
lru-dict==1.1.6
lxml==4.2.5
MarkupSafe==1.0
mccabe==0.6.1
more-itertools==4.3.0
mypy==0.620
netdisco==2.1.0
netifaces==0.10.7
packaging==18.0
parsimonious==0.8.0
parso==0.3.1
pathtools==0.1.2
pexpect==4.6.0
pickleshare==0.7.5
pkginfo==1.4.2
pluggy==0.7.1
plyvel==1.0.5
prompt-toolkit==1.0.15
ptyprocess==0.6.0
py==1.6.0
py-ecc==1.4.3
[...]

PR #1419, opened 2018-10-19
Suppress ImportWarning messages

rayrapetyan:
This commit fixes ImportWarning-s issue (#1402, #1370). As mentioned in referred issue, this is a Cython bug fixed in master and not yet available. We can't fix it in py-evm modules.
A suggested [workaround](https://bugs.launchpad.net/lxml/+bug/1732505/comments/4) was used.

### DIFF
--- a/eth/_warnings.py
+++ b/eth/_warnings.py
@@ -1,0 +1,10 @@
+from contextlib import contextmanager
+import warnings
+
+
+# TODO: drop once https://github.com/cython/cython/issues/1720 is resolved
+@contextmanager
+def catch_and_ignore_import_warning():
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', category=ImportWarning)
+        yield
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -29,18 +29,6 @@ from eth_typing import (
     Hash32,
 )
 
-from eth_utils import (
-    to_set,
-    ValidationError,
-)
-from eth_utils.toolz import (
-    assoc,
-    compose,
-    groupby,
-    iterate,
-    take,
-)
-
 from eth.db.backends.base import BaseAtomicDB
 from eth.db.chain import (
     BaseChainDB,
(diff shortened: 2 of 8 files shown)


rayrapetyan:
I did everything I could to bypass CircelCI checks (even re-created pull request from the scratch), but it keeps an old copy of chains/base.py file and fails on all imports there... "Erroneous" line numbers correspond to empty lines in base.py.

carver:
Is it working for you when you run tox locally?

rayrapetyan:
Yeah, kind of... Nevermind, I’ve found a workaround. Please review. Thanks!

pipermerriam:
FYI, I think this is the core issue: https://github.com/cython/cython/issues/1720

pipermerriam:
@rayrapetyan I'm inclined to wrap this so that it's easier for us to clean these up later.

```python
import contextlib
import warnings

def catch_and_ignore_import_warning():
    with warnings.catch_warnings():
        warnings.simplefilter('ignore', category=ImportWarning)
        yield
```

This should then be usable like this:

```python
with catch_and_ignore_import_warning():
    from eth_utils import (
        to_set,
        ValidationError,
    )
```

This will have to go somewhere that we can get it without triggering any of these warnings.  I'd be fine putting it somewhere like `eth._warnings` (`eth/_warnings.py`).

Thoughts?

rayrapetyan:
Well it looks reasonable for a permanent fix, but for a temp fix and just 4 places introducing a separate file on a root level... looks an overkill, but up to you, I can do that.

rayrapetyan:
After syncing master there are 7 places to fix, so looks reasonable to save one line in each place, committed with _warnings.py file.

pipermerriam:
> Well it looks reasonable for a permanent fix, but for a temp fix and just 4 places introducing a separate file on a root level... looks an overkill, but up to you, I can do that.

- DRY: if we found that there was an extra piece of logic needed for this to work right a week from now we might not remember to update all of the call sites.
- In a similar vein it will be easier to find all of the locations and clean them up when the core issue is fixed.

:+1: Will give this some final review on Monday.

PR #1420, opened 2018-10-21
Complete Type Hinting for `eth.tools`

Bhargavasomu:
### What was wrong?
Issue : #1398

### How was it fixed?
Adding the type hints for `eth.tools` module completely.

#### Cute Animal Picture

### DIFF
--- a/eth/tools/_utils/hashing.py
+++ b/eth/tools/_utils/hashing.py
@@ -2,10 +2,15 @@ from eth_hash.auto import keccak
 
 import rlp
 
+from typing import (
+    Iterable,
+    Tuple,
+)
+
 from eth.rlp.logs import Log
 
 
-def hash_log_entries(log_entries):
+def hash_log_entries(log_entries: Iterable[Tuple[bytes, bytes, bytes]]) -> bytes:
     """
     Helper function for computing the RLP hash of the logs from transaction
     execution.
--- a/eth/tools/_utils/hashing.py
+++ b/eth/tools/_utils/hashing.py
@@ -7,10 +7,14 @@ from typing import (
     Tuple,
 )
 
+from eth_typing import (
+    Hash32,
+)
+
 from eth.rlp.logs import Log
 
 
-def hash_log_entries(log_entries: Iterable[Tuple[bytes, bytes, bytes]]) -> bytes:
+def hash_log_entries(log_entries: Iterable[Tuple[bytes, bytes, bytes]]) -> Hash32:
     """
     Helper function for computing the RLP hash of the logs from transaction
     execution.
--- a/eth/tools/_utils/normalization.py
+++ b/eth/tools/_utils/normalization.py
@@ -5,6 +5,7 @@ from typing import (
     Any,
     AnyStr,
     Callable,
+    cast,
     Dict,
     Iterable,
     List,
@@ -56,8 +57,9 @@ from eth.tools._utils.mappings import (
 from eth.typing import (
     AccountState,
     GeneralState,
-    NormalizerType,
+    Normalizer,
     TransactionDict,
+    TransactionNormalizer,
 )
 
 
@@ -127,7 +129,7 @@ robust_decode_hex = eth_utils.curried.hexstr_if_str(to_bytes)
 #
 def dict_normalizer(formatters: Dict[Any, Callable[..., Any]],
                     required: Iterable[Any]=None,
-                    optional: Iterable[Any]=None) -> NormalizerType:
+                    optional: Iterable[Any]=None) -> Normalizer:
 
     all_keys = set(formatters.keys())
 
@@ -154,7 +156,7 @@ def dict_normalizer(formatters: Dict[Any, Callable[..., Any]],
     return normalizer
 
 
-def dict_options_normalizer(normalizers: Iterable[NormalizerType]) -> NormalizerType:
+def dict_options_normalizer(normalizers: Iterable[Normalizer]) -> Normalizer:
 
     def normalize(d: Dict[Any, Any]) -> Dict[str, Any]:
         first_exception = None
@@ -247,7 +249,7 @@ normalize_state = compose(
 )
 
 
-normalize_main_transaction = dict_normalizer({    # type: ignore # Overwrite type hint not yet supported # noqa: 501
+normalize_main_transaction = cast(Normalizer, dict_normalizer({
     "data": normalize_bytes,
     "gasLimit": normalize_int,
     "gasPrice": normalize_int,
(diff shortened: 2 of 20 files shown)


Bhargavasomu:
@cburgdorf I had to change the `configurable` function in `eth.utils.database` to `Type[T]`. 
* For the `eth.tools` module, most of them were meant to be generic I guess, and hence I had to use `Any` a lot of times. I tried tightening the types instead of using `Any`, but it resulted in type error somewhere else. Please suggest me if we could tighten the types anywhere. Will make the changes.
* I included some own defined types in `eth.typing`

cburgdorf:
@Bhargavasomu I'll give this a proper review tomorrow!

eth/tools/_utils/hashing.py
```diff
@@ -2,10 +2,15 @@
...
+from typing import (
+    Iterable,
+    Tuple,
+)
+
 from eth.rlp.logs import Log
 
 
-def hash_log_entries(log_entries):
+def hash_log_entries(log_entries: Iterable[Tuple[bytes, bytes, bytes]]) -> bytes:
```
cburgdorf: As a rule of thumb, `keccak` will always return a hash of 32 bytes, hence everything that returns `keccak(x)` should always have the return type `Hash32` rather than just arbitrary bytes.

eth/tools/_utils/normalization.py
```diff
@@ -41,12 +53,19 @@
...
+    TransactionType,
+)
+
 
 #
 # Primitives
 #
 @functools.lru_cache(maxsize=1024)
-def normalize_int(value):
+def normalize_int(value: Any) -> int:
```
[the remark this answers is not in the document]
cburgdorf: Might even qualify for a type alias (not a `NewType`) to be put in `eth.typing`:

```
IntLike = Union[int, bytes, eth_typing.HexStr, str]
```

Alternative names: `IntConvertible`, `IntRetrievable`

eth/typing.py
```diff
@@ -32,4 +34,8 @@
     List[Tuple[Address, Dict[str, Union[int, bytes, Dict[int, int]]]]]
 ]
 
+TransactionType = Dict[str, Union[bytes, int, str]]
```
cburgdorf: Did you try locking this down to a `TypedDict`? Also, in general I prefer not to use the suffix `Type` for any new types / aliases that we create. It feels redundant to me (we also say `Chain`, `BaseComputation`, `BaseVM` and not `ChainType`, `BaseComputationType` or `BaseVMType`.

cburgdorf: As a name suggestion, maybe just `TransactionDict`?

Bhargavasomu:
```
def normalize_int(value: Union[int, bytes, str, ...]) -> int
    """
    Robust to integer conversion, handling hex values, string representations,
    and special cases like `0x`.
    """
    if is_integer(value):
        return value
   ...
   ...
```

@cburgdorf @pipermerriam if the type hinting was changed as above, then the `return value` would return type ` Union[int, bytes, str, ...]` instead of `int` which would throw an error. I have tested those and tried tightening the types prior to this PR.

carver: Inside the `if is_integer` block, it seems reasonable to `cast(int, value)`.

Bhargavasomu:
In general how do I overwrite the Type hinting of a variable? This arises in the following scenario.
```
def dict_normalizer(formatters: Dict[Any, Callable[..., Any]],
                    required: Iterable[Any]=None,
                    optional: Iterable[Any]=None) -> Callable[[Dict[Any, Any]], Dict[str, Any]]:
    ...
    ...
```
This above function is generic and used by a lot of functions. But I need to tighten the functions which are using this function. Explained in below example.
```
normalize_main_transaction = dict_normalizer({
    "data": normalize_bytes,
    "gasLimit": normalize_int,
    "gasPrice": normalize_int,
    "nonce": normalize_int,
    "secretKey": normalize_bytes,
    "to": normalize_to_address,
    "value": normalize_int,
})
```
The normalize_main_transaction function returns object of the type 
`Callable[[Dict[Any, Any]], Dict[str, Any]]`. But I need the type of `normalize_main_transaction` function to be of the type `Callable[TransactionDict], TransactionDict]`.
 
How can the above be achieved?

eth/typing.py
Bhargavasomu: If we need to cast it, then we would need to cast it to a new variable. How do we `overwrite` the type of a variable, which is already type hinted? This is because, mypy is throwing error when trying to overwrite the type hinting of the variables which are already type hinted.
```
value = cast(int, value)
return value
```
The following error occurs `Incompatible return value type (got "Union[int, bytes, ...]", expected "int")`

carver: I think it's fine to: `return cast(int, value)`

I'm not sure why mypy doesn't like the cast. @cburgdorf is our resident mypy wrangler.

Bhargavasomu:
@cburgdorf I have done some searching regarding `mypy` and found that there are some issues which are still in deveopment by the `mypy community`. They are as follows.
* https://github.com/python/mypy/issues/4029
* Overwriting the type hints of the variables is still not possible.

Because of this, in the code, I have used `variable inline annotations` wherever possible along with `type: ignore`. But `circle.ci` is showing the following error because of which the test cases are failing.
screenshot from 2018-10-24 15-06-49

And by the `variable annotation`, it means the following.
screenshot from 2018-10-24 15-08-06

How should I proceed with this?

cburgdorf:
>But circle.ci is showing the following error because of which the test cases are failing.

Since `eth` is still compatible with Python 3.5 you need to use the ugly comment notation for variable type annotations. Here's an example on how to do it.

https://github.com/ethereum/py-evm/blob/e48b279bb53928ea1a6941109398b2ad6afb67bb/eth/chains/base.py#L116-L119

Once we drop Python 3.5 support for the `eth` module (I expect that to happen within the next months) we can migrate those to the proper type annotation syntax.

Bhargavasomu:
@cburgdorf I don't remember changing anything related to `trinity-lightchain`(went through the commit history too). The test cases are failing for that particularly. Is this because the module is to be `rebased`?

pipermerriam:
@Bhargavasomu what you're seeing looks to be the same issue that I fixed in #1416 

https://github.com/ethereum/py-evm/pull/1416/files#diff-1d37e48f9ceff6d8030570cd36286a61L269

Bhargavasomu:
@pipermerriam , so maybe I should wait till #1416 gets merged?

Bhargavasomu:
@pipermerriam this is causing issues in the `setup` I guess. Something in installing the `rocksdb` is causing the error.

Bhargavasomu:
@cburgdorf I have made the changes as required, the tests are failing because of some `RocksDB` installation, which I think is being taken care of. Could you please review it again and tell me whether anything else is to be added regarding the `type hinting` of this module?

cburgdorf:
@Bhargavasomu I'll take a look today

Bhargavasomu:
@cburgdorf forgot about the NormalizerType, sorry about that. Will make the PR for the `eth.vm` as soon as possible. It's great working with you Christoph. Thankyou.

cburgdorf:
@Bhargavasomu Likewise! You are doing great work and I'm happy to have you contributing :+1: 

I also just [triggered](https://github.com/ethereum/py-evm/issues/1398#issuecomment-433349890) the payout of the first 150 DAI for this milestone.

PR #1421, opened 2018-10-21
Implement eth_syncing endpoint

rayrapetyan:
Implemented eth_syncing endpoint as per https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_syncing specs.

### DIFF
--- a/tests/trinity/core/json-rpc/test_ipc.py
+++ b/tests/trinity/core/json-rpc/test_ipc.py
@@ -454,7 +454,6 @@ def mock_peer_count(count):
     ],
 )
 async def test_peer_pool_over_ipc(
-        monkeypatch,
         jsonrpc_ipc_pipe_path,
         request_msg,
         event_bus_setup_fn,
@@ -473,10 +472,10 @@ async def test_peer_pool_over_ipc(
     assert result == expected
 
 
-def mock_syncing(syncing):
+def mock_syncing(is_syncing, progress=None):
     async def mock_event_bus_interaction(bus):
         async for req in bus.stream(SyncingRequest):
-            bus.broadcast(SyncingResponse(syncing), req.broadcast_config())
+            bus.broadcast(SyncingResponse(is_syncing, progress), req.broadcast_config())
             break
 
     return mock_event_bus_interaction
@@ -493,7 +492,7 @@ def mock_syncing(syncing):
         ),
         (
             build_request('eth_syncing'),
-            mock_syncing(SyncProgress(0, 1, 2)),
+            mock_syncing(True, SyncProgress(0, 1, 2)),
             {'result': {'startingBlock': 0, 'currentBlock': 1, 'highestBlock': 2}, 'id': 3,
              'jsonrpc': '2.0'},
         ),
@@ -503,7 +502,6 @@ def mock_syncing(syncing):
     ],
 )
 async def test_eth_over_ipc(
-        monkeypatch,
         jsonrpc_ipc_pipe_path,
         request_msg,
         event_bus_setup_fn,
--- a/tests/trinity/core/json-rpc/test_ipc.py
+++ b/tests/trinity/core/json-rpc/test_ipc.py
@@ -22,6 +22,13 @@ from trinity.nodes.events import (
     NetworkIdRequest,
     NetworkIdResponse,
 )
+from trinity.sync.common.events import (
+    SyncingRequest,
+    SyncingResponse,
+)
+from trinity.sync.common.types import (
+    SyncProgress
+)
 
 from trinity.utils.version import construct_trinity_client_identifier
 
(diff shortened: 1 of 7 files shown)


trinity/sync/common/events.py
```diff
@@ -0,0 +1,26 @@
...
+
+from trinity.sync.common.types import (
+    SyncProgress
+)
+
+
+class SyncingResponse(BaseEvent):
+
+    def __init__(self, syncing: Union[bool, SyncProgress]) -> None:
+        self.syncing = syncing
```
pipermerriam: What if this was split into two properties.

```python
class SyncingResponse(BaseEvent):
    is_syncing: bool
    progress: Optional[SyncProgress]
```

This should give slightly better type safety in the places where it is used so that you don't have to do a type check on `self.syncing`.

trinity/rpc/modules/eth.py
```diff
@@ -264,5 +267,12 @@ class Eth(RPCModule):
     async def protocolVersion(self) -> str:
         return "63"
 
-    async def syncing(self) -> bool:
-        raise NotImplementedError()
+    async def syncing(self) -> Union[bool, Dict[str, BlockNumber]]:
```
pipermerriam: For stronger type safety we've started using `mypy_extensions.TypedDict` for these types of dictionaries with known keys and value types.

https://github.com/python/mypy/blob/a7f0263f29317985e0a2ca6f85ba5b7666401347/extensions/mypy_extensions.py#L74-L99

trinity/sync/common/chain.py
```diff
@@ -134,6 +166,16 @@ def _get_peer_header_syncer(self, peer: BaseChainPeer) -> Iterator['PeerHeaderSy
                 new_headers = tuple(h for h in header_batch if h not in self.header_queue)
                 await self.wait(self.header_queue.add(new_headers))
 
+    def syncing(self) -> Union[bool, SyncProgress]:
+        if not self._syncing:
+            return False
+        return self._peer_header_syncer.sync_progress
+
+    async def handle_syncing_requests(self) -> None:
```
pipermerriam: Thoughts on renaming this to `handle_sync_status_requests` or `handle_sync_info_requests`.  The current name could easily be confused as something which is related to the actually code which performs chain sync.

trinity/sync/common/chain.py
```diff
@@ -134,6 +166,16 @@ def _get_peer_header_syncer(self, peer: BaseChainPeer) -> Iterator['PeerHeaderSy
                 new_headers = tuple(h for h in header_batch if h not in self.header_queue)
                 await self.wait(self.header_queue.add(new_headers))
 
+    def syncing(self) -> Union[bool, SyncProgress]:
```
pipermerriam: Can we rename this to something like `get_sync_info` or `get_sync_status`?

trinity/sync/common/chain.py
```diff
@@ -306,6 +349,7 @@ def get_target_header_hash(self) -> Hash32:
 
             yield headers
             last_received_header = headers[-1]
+            self.sync_progress.current_block = last_received_header.block_number
```
pipermerriam: See my comment below about named tuples.  Instead of updating the existing object (and thus mutating it), I'd prefer we created an all new `SyncProgress` item.

trinity/sync/common/types.py
```diff
@@ -0,0 +1,13 @@
+from eth_typing import (
+    BlockNumber,
+)
+
+
+class SyncProgress:
```
pipermerriam: Can you update this to use a `typing.NamedTuple`

https://docs.python.org/3/library/typing.html#typing.NamedTuple

To make updating the `current_block` easy you could do something like:

```python
class SyncProgress(NamedTuple):
    ...

    def update_current_block(self, new_current_block):
        return type(self)(self.starting_block, new_current_block, self.highest_block)
```

tests/trinity/core/json-rpc/test_ipc.py
```diff
@@ -464,3 +471,52 @@ def mock_peer_count(count):
...
+            {'result': {'startingBlock': 0, 'currentBlock': 1, 'highestBlock': 2}, 'id': 3,
+             'jsonrpc': '2.0'},
+        ),
+    ),
+    ids=[
+        'eth_syncing_F', 'eth_syncing_T',
+    ],
+)
+async def test_eth_over_ipc(
+        monkeypatch,
```
pipermerriam: Doesn't look like `monkeypatch` is used anywhere in this function.

trinity/sync/common/chain.py
rayrapetyan: Hm, any reason we are afraid of mutating objects in this project? Introducing setter is ok, but recreating whole object.. why?

pipermerriam: We have a reasonably strong convention across all of our codebases to avoid mutability unless faced with a specific reason why it is necessary.  This gets us two objective benefits.

- Reduced cognitive overhead.  When reading non mutative code, you don't have to track state which makes reading the code significantly easier.
- Avoiding multiple types of bugs entirely.

So, while in many cases, a mutation based approach may be fine, we still opt to find a non-mutative way.

rayrapetyan: Well, in both cases we change memory behind self.sync_progress variable, complexity of tracking this change in code is same for both approaches. The only difference is changing only one values vs all three values at once, two of which are always redundant. I hardly can see any benefits but wasting of CPU cycles here. However, if it's a "holy rule" of a project - I understand and can follow it.

pipermerriam: I understand that it may seem un-necessary, and it might be, but I'd like it updated to remove the mutation.

cburgdorf: >Well, in both cases we change memory behind self.sync_progress variable, complexity of tracking this change in code is same for both approaches

Not entirely. Some code path could hold on to an instance of `SyncProgress` and do something with it based on the state of its properties. And with the solution that uses mutation, these properties could just change which basically means, you have to keep in mind that they could technically just change from one moment to the next.

In the non-mutation version, a code path that holds on to a`SyncProgress` object never has to worry about its state changing. It will always stay the way it is.

pipermerriam:
:+1: to merge when tests go green.

PR #1424, opened 2018-10-22
API docs + some fine tuning for Plugin API

cburgdorf:
### What was wrong?

Plugins didn't have proper API docs yet. There were also some easy to fix glitches that this PR addresses. Notice that this is just a first step towards a full detailed plugin guide which is coming in a separate PR (#1429 )

### How was it fixed?

- created API docs
- created `TrinityBootInfo` type to reduce complexity of several method signatures
- Use `pass` instead of `raise` for abstract methods
- Fix that ugly thing were we return `None` as a context

#### Cute Animal Picture

### DIFF
--- a/docs/api/trinity/api.cli.rst
+++ b/docs/api/trinity/api.cli.rst
@@ -1,0 +1,46 @@
+Command Line Interface (CLI)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: shell
+
+    usage: trinity [-h] [--version] [--trinity-root-dir TRINITY_ROOT_DIR]
+                [-l {debug,info}] [--network-id NETWORK_ID | --ropsten]
+                [--sync-mode {full,light} | --light] [--data-dir DATA_DIR]
+                [--nodekey NODEKEY] [--nodekey-path NODEKEY_PATH]
+                {console,attach} ...
+
+    positional arguments:
+    {console,attach}
+        console             run the chain and start the trinity REPL
+        attach              open an REPL attached to a currently running chain
+
+    optional arguments:
+    -h, --help            show this help message and exit
+
+    sync mode:
+    --version             show program's version number and exit
+    --trinity-root-dir TRINITY_ROOT_DIR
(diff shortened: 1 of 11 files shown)


docs/api/trinity/api.cli.rst
```diff
@@ -0,0 +1,46 @@
...
+                            network. Shortcut for `--networkid=3`
+
+    sync mode:
+    --sync-mode {full,light}
+    --light               Shortcut for `--sync-mode=light`
+
+    chain:
+    --data-dir DATA_DIR   The directory where chain data is stored
+    --nodekey NODEKEY     Hexadecimal encoded private key to use for the nodekey
+    --nodekey-path NODEKEY_PATH
```
pipermerriam: side note: we should be able to merge the functionality of this argument in with the `--nodekey` argument and just do detection on whether the arg is referencing a file or that it's just a string nodekey.

docs/api/trinity/extensibility/index.rst
```diff
@@ -0,0 +1,16 @@
+Extensibility
+=============
+
+.. note::
```
pipermerriam: Maybe `.. warning::`?

trinity/extensibility/plugin.py
```diff
@@ -80,18 +118,25 @@ class BasePlugin(ABC):
     @abstractmethod
     def name(self) -> str:
         """
-        Describe the name of the plugin
+        Describe the name of the plugin.
         """
         raise NotImplementedError(
```
pipermerriam: you could replace this `NotImplementedError` with a `pass` (our current convention).

docs/api/trinity/api.cli.rst
cburgdorf: Created an issue https://github.com/ethereum/py-evm/issues/1428

cburgdorf:
@pipermerriam I decided to spin out the guide into a separate PR (#1429) as writing the guide will still take a bit more time and I think we can get this in independently from the guide.

Issue #1426, opened 2018-10-23
Convert CircleCI to use docker images for testing.


pipermerriam:
### What is wrong?

Trying to get our ci environment to be *just right* is hard.   Case-in-point, lots of thrash trying to get rocksdb installed and then to cache that installation.

### How can it be fixed

Switch over to using custom docker images.  Lots to figure out, but basically we create images for each of our build environments and test using those docker images.

This shifts the complexity into building docker images, but I think that will be much more maintainable.

PR #1430, opened 2018-10-25
Christoph/review/1420

cburgdorf:
Temp PR to diagnose errors

### DIFF
--- a/eth/tools/_utils/git.py
+++ b/eth/tools/_utils/git.py
@@ -3,6 +3,6 @@ import subprocess
 from eth_utils import to_text
 
 
-def get_version_from_git():
+def get_version_from_git() -> str:
     version = subprocess.check_output(["git", "describe"]).strip()
     return to_text(version)
--- a/eth/tools/_utils/hashing.py
+++ b/eth/tools/_utils/hashing.py
@@ -2,10 +2,15 @@ from eth_hash.auto import keccak
 
 import rlp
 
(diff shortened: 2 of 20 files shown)


Bhargavasomu:
Some installation problem with the `RocksDB`.

cburgdorf:
@Bhargavasomu yep, trying to figure out why...because I haven't had any problems after rebasing https://github.com/ethereum/py-evm/pull/1424 on top of latest master

Bhargavasomu:
@cburgdorf , #1181 failed for the same reasons, by looking at the details given by `circleci`

Bhargavasomu:
@cburgdorf #1431 seems to make the necessary header files accessible. Could you please take a look at it? I have made a link to the header files, which I don't think was previously present. But it is taking a lot of time to run the test cases. Is this because I have cleared the cache somewhere?

cburgdorf:
@Bhargavasomu nice find. I rebased this on top of #1431 just to double check :)

pipermerriam:
I'm fine with us rolling back #1416 for now until we get #1426 in place or something similar which makes this more reliable.

pipermerriam:
you can bust the cache by incrementing the `v1/v2` numbers in the cache strings in the `./circleci/config` file.  But given that this has happened more than once (already happened to me once while trying to get the PR green), I suspect it'll happen again.

PR #1432, opened 2018-10-25
Revert "Base implementation of RocksDB support (#1416)"

cburgdorf:
### What was wrong?

Rolling back the RocksDB switch as the installation doesn't yet work as smoothly as we would like it to be.

### How was it fixed?

This reverts commit 61a9a37882b4dce596a1d9d364183f4a5bcd1c3f.

#### Cute Animal Picture


cburgdorf:
@pipermerriam I'm going to merge this based on the agreement in the gitter chat. Reasoning is that otherwise new incoming PRs fail on CI. If it turns out that I should not have merged it you'll just kick my ass at devcon :sweat_smile:

PR #1433, opened 2018-10-25
Move mainnet DAO fork config into mainnet configuration

pipermerriam:
### What was wrong?

The default `HomesteadVM` came with the dao fork block number already set.  This didn't seem semantically correct since the block number would differ based on the chain configuration and what chain it was used in.

### How was it fixed?

Changed to have the block number configured within the context of the mainnet chain.

#### Cute Animal Picture

21jan16_161954

### DIFF
--- a/eth/chains/mainnet/__init__.py
+++ b/eth/chains/mainnet/__init__.py
@@ -13,6 +13,7 @@ from .constants import (
     HOMESTEAD_MAINNET_BLOCK,
     SPURIOUS_DRAGON_MAINNET_BLOCK,
     DAO_FORK_MAINNET_EXTRA_DATA,
+    DAO_FORK_MAINNET_BLOCK,
 )
 from eth import constants
 
@@ -60,7 +61,7 @@ class MainnetDAOValidatorVM:
 
 
 class MainnetHomesteadVM(MainnetDAOValidatorVM, HomesteadVM):
-    pass
+    dao_fork_block_number = DAO_FORK_MAINNET_BLOCK
 
 
 MAINNET_FORK_BLOCKS = (
--- a/eth/chains/mainnet/__init__.py
+++ b/eth/chains/mainnet/__init__.py
@@ -40,10 +40,8 @@ class MainnetDAOValidatorVM:
         super().validate_header(header, previous_header, check_seal)  # type: ignore
 
         # The special extra_data is set on the ten headers starting at the fork
-        extra_data_block_nums = range(
-            cls.get_dao_fork_block_number(),
-            cls.get_dao_fork_block_number() + 10,
-        )
+        dao_fork_at = cls.get_dao_fork_block_number()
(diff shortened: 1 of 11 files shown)


eth/vm/forks/homestead/__init__.py
```diff
@@ -18,7 +15,7 @@
 
 class MetaHomesteadVM(FrontierVM):
     support_dao_fork = True
-    dao_fork_block_number = DAO_FORK_MAINNET_BLOCK
+    dao_fork_block_number = 0
```
pipermerriam: cc @carver thoughts on how we should proceed here? I'm thinking of changing both of this to default to `None` and to add a runtime check that blows up if the value hasn't been explicitely set.

carver: How about:

```py
@abstractmethod
@property
def doa_fork_block_number(self):
  pass
```

pipermerriam: This didn't work well due to use using inheritance for chains.  I went with a `classmethod` implementation.  take a look and tell me what you think.

eth/chains/mainnet/constants.py
```diff
@@ -13,7 +18,7 @@
 #
 # DAO Block
 #
-DAO_FORK_MAINNET_BLOCK = 1920000
+DAO_FORK_MAINNET_BLOCK = cast(BlockNumber, 1920000)
```
carver: Is `BlockNumber(1920000)` a thing in py3.5?

eth/vm/forks/homestead/__init__.py
```diff
@@ -18,7 +17,13 @@
 
 class MetaHomesteadVM(FrontierVM):
     support_dao_fork = True
-    dao_fork_block_number = DAO_FORK_MAINNET_BLOCK
+    dao_fork_block_number = None  # type: Optional[BlockNumber]
```
carver: Maybe move to private `_dao_fork`...?

- Helps make sure that there aren't any stray references to `VM.dao_fork_block_number` out there
- Encourages new usages to call the `get_dao_fork`... because the other one looks private.

eth/chains/mainnet/__init__.py
```diff
@@ -39,7 +40,10 @@ def validate_header(cls, header, previous_header, check_seal=True):
         super().validate_header(header, previous_header, check_seal)  # type: ignore
 
         # The special extra_data is set on the ten headers starting at the fork
-        extra_data_block_nums = range(cls.dao_fork_block_number, cls.dao_fork_block_number + 10)
+        extra_data_block_nums = range(
+            cls.get_dao_fork_block_number(),
+            cls.get_dao_fork_block_number() + 10,
```
carver: Now that this had to split lines, might be worth getting the fork number first, so the range can be on one line. Like:

```py
fork_num = cls.get_dao_fork_block_number()
extra_data_block_nums = range(fork_num, fork_num + 10)
```

PR #1434, opened 2018-10-25
Fix mainnet genesis configuration

pipermerriam:
### What was wrong?

The parameters for our mainnet genesis were  incorrect.

### How was it fixed?

Updated to use correct genesis parameters.

#### Cute Animal Picture

squirrel-pumpkin-funny-photography-max-ellis-5


tests/core/chain-object/test_gas_estimation.py
```diff
@@ -30,11 +30,11 @@ def chain(chain_without_block_validation):
         (b'\xff' * 10, None, ADDR_1010, True, 21680),
         (b'\xff' * 10, None, ADDR_1010, False, 21680),
         # sha3 precompile
-        (b'\xff' * 32, None, ADDRESS_2, True, 35381),
-        (b'\xff' * 32, None, ADDRESS_2, False, 35369),
-        (b'\xff' * 320, None, ADDRESS_2, True, 54888),
+        (b'\xff' * 32, None, ADDRESS_2, True, 35333),
+        (b'\xff' * 32, None, ADDRESS_2, False, 35345),
+        (b'\xff' * 320, None, ADDRESS_2, True, 54840),
```
pipermerriam: These gas numbers changed because the genesis difficulty they are operating against changed which changes the gas limit for the `pending_header` that the estimation is done against.

PR #1438, opened 2018-10-25
Fix warnings when printing Trinity cli header

pipermerriam:
### What was wrong?

The ASCII "trinity" header that is printed out when running the trinity command uses forward slashes.  These cause warnings because python thinks they are supposed to be escape sequences.

### How was it fixed?

Update header to use raw strings.

#### Cute Animal Picture

awesomelycute-com-2607

### DIFF
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -81,15 +81,15 @@ from trinity.utils.version import (
 PRECONFIGURED_NETWORKS = {MAINNET_NETWORK_ID, ROPSTEN_NETWORK_ID}
 
 
-TRINITY_HEADER = (
+TRINITY_HEADER = "\n".join((
     "\n"
-    "      ______     _       _ __       \n"
-    "     /_  __/____(_)___  (_) /___  __\n"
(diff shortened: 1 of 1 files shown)


trinity/main.py
```diff
@@ -81,15 +81,15 @@
...
+TRINITY_HEADER = "\n".join((
     "\n"
-    "      ______     _       _ __       \n"
-    "     /_  __/____(_)___  (_) /___  __\n"
-    "      / / / ___/ / __ \/ / __/ / / /\n"
-    "     / / / /  / / / / / / /_/ /_/ / \n"
-    "    /_/ /_/  /_/_/ /_/_/\__/\__, /  \n"
-    "                           /____/   "
-)
+    r"      ______     _       _ __       ",
```
carver: The PR appears to add an extra `\n` above the header, compared to the pre-diff. I don't really care, just making sure it was intentional.

pipermerriam: It was intentional

PR #1442, opened 2018-10-28
Refactor EVM stack to store only int

Bhargavasomu:
### What was wrong?
EVM stack (`eth.vm.stack`) previously stored items of both `int` and `bytes`. This causes typing issues and morevover `bytes` data can be converted to `int`, and that would be a better design. This is because while popping the data, we can be sure of only 1 type(`int`), instead of `union of int and bytes`.

### How was it fixed?
Stack was made sure that it stored only items of the type `int`. Data of type `bytes` were converted to `int` and then stored in the stack. Also instead of having a single `pop` function, 2 functions `pop_ints` and `pop_bytes` were introduced, and they are used instead of the traditional `pop` function.

#### Cute Animal Picture

### DIFF
--- a/eth/validation.py
+++ b/eth/validation.py
@@ -167,9 +167,7 @@ def validate_uint256(value, title="Value"):
 
 
 def validate_stack_item(value):
-    if isinstance(value, bytes) and len(value) <= 32:
-        return
-    elif isinstance(value, int) and 0 <= value <= UINT_256_MAX:
+    if isinstance(value, int) and 0 <= value <= UINT_256_MAX:
         return
     raise ValidationError(
         "Invalid Stack Item: Must be either a length 32 byte "
--- a/eth/vm/computation.py
+++ b/eth/vm/computation.py
@@ -291,17 +291,25 @@ class BaseComputation(Configurable, ABC):
         """
         return self._gas_meter.refund_gas(amount)
 
-    def stack_pop(self, num_items=1, type_hint=None):
+    def stack_pop_ints(self, num_items=1):
         """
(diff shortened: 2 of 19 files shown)


Bhargavasomu:
I don't remember changing any of the core logic, but still the `assert` statements are failing. Can someone help me with this?

cburgdorf:
@Bhargavasomu From our discussion in https://github.com/ethereum/py-evm/issues/1398 it seems that we have two different problems to address.

1. The Stack API sometimes returns single items and sometimes returns multiple items. I think this is the main reason that makes adding type hints hard and led to a very explicit API proposal from @carver with methods such as `pop1`, `pop2` etc.

2. The other issue is that the stack currently allows mixing of `int` and `bytes`.

It seems that this PR only addresses the second issue but the first issue is actually same (if not more) important.

I'd also like to see an actual performance comparison of the different approaches that were proposed. 

If this turns out to be too big of an distraction in the process of adding type hints I'd propose to keep the Stack API as is for now and use some `# type: ignore`'s  where needed.

I can assist with changing the Stack API and make performance comparisons. But please don't read this as discouraging you from changing the Stack API. If you really want to go down this rabbit whole, feel free to do so but I believe it will be some work and we want to see some performance numbers before we commit to some change here. Hence, if you just want to move the type hints forward, I'd happily accept some `# type: ignore`'s for now until we settle on a change for the Stack API.

Bhargavasomu:
@cburgdorf I would be delighted to undertake the change of the `Stack API` with your guidance, as the task is exciting and would push me to explore more about the codebase. I will use `type: ignore` for now and after the type hinting is over with, will tackle this issue.

PR #1443, opened 2018-10-28
merge --nodekey and --nodekey-path arguments

### LINKED ISSUE #1428: Merge --nodekey and nodekey-path command line arguments
### What is wrong?

UX could be improved by merging functionality of `--nodekey-path` with the `--nodekey`  argument. (https://github.com/ethereum/py-evm/pull/1424#discussion_r227424899)

### How can it be fixed

We should be able to merge the functionality of `--nodekey-path` with the `--nodekey` argument and just do detection on whether the arg is referencing a file or that it's just a string nodekey.

jvmaia:
### What was wrong?
closes #1428 

### How was it fixed?
Removing the argument in file `cli_parser.py` and merging the options in `chains.py`.

#### Cute Animal Picture


rayrapetyan:
@jvmaia,I don't think finding "/" in argument is a good way to determine a path. It's a). Not a cross-platform solution (windows uses "\") b). Will not work for cases when path doesn't contain slashes at all (e.g. specified file in current dir). I would suggest to use e.g. "os.path.isfile" instead.

jvmaia:
I was in doubt about that, thanks for the suggestion

trinity/utils/chains.py
```diff
@@ -135,12 +135,11 @@ def construct_trinity_config_params(
     if args.data_dir is not None:
         yield 'data_dir', args.data_dir
 
-    if args.nodekey_path and args.nodekey:
-        raise ValueError("Cannot provide both nodekey_path and nodekey")
-    elif args.nodekey_path is not None:
-        yield 'nodekey_path', args.nodekey_path
-    elif args.nodekey is not None:
-        yield 'nodekey', decode_hex(args.nodekey)
+    if args.nodekey:
```
cburgdorf: Can you change this to:

```suggestion
    if args.nodekey is not None:
```

I *think* that technically, there aren't any bad things to expect with the plain falsey check so this is mostly for consistency with the rest of the code.

cburgdorf: Can you change this to:

```suggestion
    if args.nodekey is not None:
```

I *think* that technically, there aren't any bad things to expect with the plain falsey check so this is mostly for consistency with the rest of the code.

jvmaia: Fixed it, commits squashed, thanks for the feedback

PR #1444, opened 2018-10-30
More work on the plugin guide

cburgdorf:
### What was wrong?

Plugin guide is still rough. Needs to be smoooooth.

### How was it fixed?

- made `status` a readonly property to allow better documentation (and frankly improve encapsulation)
- More guidance on how to use plugins including a fully working `PeerCountReporter` example
- rebased on #1450 to also cover plugin distribution

#### Cute Animal Picture


cburgdorf:
This is now based on #1450 to also explain how to distribute plugins via pip!

docs/guides/trinity/writing_plugins.rst
```diff
@@ -245,19 +267,69 @@ We may want a plugin to only start running if:
...
+``pip``. The good news is, it's not hard at all!
+
+In this guide, we won't go into details about how to create Python packages as this is already
+`covered in the official Python docs <https://packaging.python.org/tutorials/packaging-projects/>`_
+.
+
+Once we have a ``setup.py`` file, all we have to do is to expose our plugin under
+``trinity.plugins`` via the ``entry_points`` section.
+
+.. literalinclude:: ../../../tests/trinity/integration/trinity_test_plugin/setup.py
```
cburgdorf: @pipermerriam this sucks because we use a different example here than in the rest of the guide which will confuse the reader. Having thought about this, I think it would be better if the example plugin that is used in this guide (`PeerCountReportPlugin`) had a `setup.py` file and then the integration test would simply use that plugin instead of defining its own.

Just, the one thing that bugs me about this plan is that I'm not sure if we want to have things like `setup.py` and the `tar.gz` package within `trinity/plugins/examples`. As in, I'm not sure if that is idiomatic code organization or if that may even come with weird side effects.

I think it would make more sense to have example plugins (or maybe even all plugins?) as a top level directory?

pipermerriam: > I think it would make more sense to have example plugins (or maybe even all plugins?) as a top level directory?

This was the first thing that came to mind.

trinity-plugins/examples/peer_count_reporter/plugin.py
```diff
@@ -51,7 +51,7 @@ def configure_parser(self,
 
     def on_ready(self) -> None:
         if self.context.args.report_peer_count:
-            self.start()
+            print("FOOBAR")
```
cburgdorf: @pipermerriam I've spend hours on figuring out why the entrypoints mechanism doesn't the way I want and now I need to look at something else and hope for bright ideas from someone else.

Here's what happened:

1. I moved the `PeerCountReporterPlugin` out of `trinity.plugins` so that it is an entirely different module
2. The entrypoint mechanism works if the plugin is simple (the `printt("FOOBAR")` works)
3, As soon as we call `self.start` (which then launchs a new process etc), the following happens:

```
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.6/multiprocessing/spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "/usr/lib/python3.6/multiprocessing/spawn.py", line 115, in _main
    self = reduction.pickle.load(from_parent)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity-plugins/examples/peer_count_reporter/plugin.py", line 10, in <module>
    from trinity.extensibility import BaseIsolatedPlugin
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/__init__.py", line 29, in <module>
    from .main import (  # noqa: F401
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/main.py", line 52, in <module>
    from trinity.plugins.registry import (
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/plugins/registry.py", line 48, in <module>
    entry_point.load()() for entry_point in pkg_resources.iter_entry_points('trinity.plugins')
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/trinity/plugins/registry.py", line 48, in <listcomp>
    entry_point.load()() for entry_point in pkg_resources.iter_entry_points('trinity.plugins')
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2346, in load
    return self.resolve()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm/venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2356, in resolve
    raise ImportError(str(exc))
ImportError: module 'plugin' has no attribute 'PeerCountReporterPlugin'
```

Steps to reproduce:

1. Checkout this branch
2. run `pip install -e trinity-plugins/examples/peer_count_reporter`
3. `trinity --report-peer-count`

Notice the mentioned error. So, the plugin works in general (which is the `--report-peer-count` is recognized just fine) but as then there's something that just falls apart. And the exact same plugin works just fine when its under `trinity.examples`  instead of sitting somewhere external and being plugged in through the `entrypoints` mechanism.

This did not pop up earlier because the existing test for the plugin discovery mechanism used the most simple plugin that just defined a name but did nothing else. In other words, we haven't figured out enough about the entrypoints mechanism to really enable people to write external plugins :sob:

carver: I don't think any of us have experience with the entry points mechanism, so it's probably just going to take a lot of web surfing to figure it out.

cburgdorf: >so it's probably just going to take a lot of web surfing to figure it out

Yep, that's what has ruined my day. There's just something inside me that thinks "This dude Jason is an amazing web surfer. He really knows how to Google stuff" haha.

Joking. I'll continue digging. Just wanted to share the pain and frustration ;)

cburgdorf: At least, I nailed it down to being an issue with `multiprocessing` and `entry points`. It's only the other plugin types that do not spawn new processes do work just fine. Digging deeper...

pipermerriam: I'd add a logging statement to print out what `plugin` is in that context as I suspect it's not the module you think it is.  I'd also log the value of `sys.path`.

cburgdorf: @pipermerriam thanks I tried that but couldn't find the info that I had hoped for. However, it seems I have tracked it down to be related to the way we launch processes. E.g. this fails:

```
        from trinity.utils.mp import ctx
        process = ctx.Process(
            target=self.foo,
        )

        process.start()
```

This works:

```
       from multiprocessing import Process
        process = Process(
            target=self.foo,
        )

        process.start()
```

Anyway, I'll continue looking into this tomorrow...family time ;)

cburgdorf: Further tracked it down, it's precisely about `fork` vs `spawn`. The bug does not happen if this is set to `fork`

https://github.com/ethereum/py-evm/blob/9870841a126fb53a6cbab2ef7bf3d46262fa23a8/trinity/utils/mp.py#L20

I have multiple thoughts to that:

1. `fork` is more efficient as it starts with shared memory and uses copy-on-write to guarantee isolation. It is also the default on Unix. See https://unix.stackexchange.com/questions/58145/how-does-copy-on-write-in-fork-handle-multiple-fork and https://stackoverflow.com/questions/2276117/python-multiprocessing-process-vs-standalone-python-vm

2. In general `spawn` provides slightly more isolation for edge cases (e.g. forking multi threaded processes may be problematic).

3. Changing the process launching strategy *because of this* feels like a sledge hammer that probably causes other (subtle) bugs so I want to continue finding an alternative fix.

cburgdorf: YES! :sparkles:  I figured out how to make it work with `spawn`. And it's pretty straight forward. Will have a fix soon after I come back from my physio therapy.

pipermerriam: FYI, if memory serves, the choice to use `spawn` may have been somewhat arbitrary and not driven by any specific requirement so it isn't out of the question for us to change it in the face of an objective reason (though I think we also need to take into account what this will do to our future ability to work in windows.

cburgdorf: @pipermerriam it's now working with `spawn` and the fix was really easy. See the commit message in https://github.com/ethereum/py-evm/pull/1444/commits/2a0c3ba33446cee4ab01108d384104a519e64821

Do you want to review this again? External plugins are now in `trinity-external-plugins` while the built-in ones remain directly in trinity.

pipermerriam: > Do you want to review this again? 

Nope

carver: >  it's now working with spawn and the fix was really easy. See the commit message in 2a0c3ba

Great find @cburgdorf !

PR #1450, opened 2018-11-06
Enable auto discovery of plugins

cburgdorf:
### What was wrong?

There's currently no way to simply distribute plugins via pip.

### How was it fixed?

Pick up plugins based on `entry_points` mechanism.

#### Cute Animal Picture


cburgdorf:
This landed as part of https://github.com/ethereum/py-evm/pull/1444

PR #1451, opened 2018-11-07
Enables eth.vm Type Hinting

Bhargavasomu:
### What was wrong?
Needed type hinting for `eth.vm` as part of #1398 

### How was it fixed?
Adding Type Hints to `eth.vm`

#### Cute Animal Picture


Bhargavasomu:
I have used `type: ignore` in the following scenarios.
* Overwriting of Inherited Variables or Functions (type may change)
*  In places where `staticmethod` function was used
* Used a bunch of type ignores in `eth.vm.forks` because of `Liskov Substitution Principle`, which gives the following error `Signature of "<variable or function>" incompatible with supertype "<ClassName>"`

For now the type hint of the function `stack_pop` in `eth.vm.computation` has been set to `Any`. This needs to be changed when the `Stack API` has been modified as discussed in #1398 (which will be taken up after this issue is closed).

That said, @cburgdorf could you please review the changes I have made?

eth/vm/base.py
```diff
@@ -237,17 +263,17 @@ def get_uncle_reward(block_number: int, uncle: BaseBlock) -> int:
...
     @classmethod
     @abstractmethod
-    def create_unsigned_transaction(cls, *args, **kwargs):
+    def create_unsigned_transaction(cls, *args: Any, **kwargs: Any) -> Any:
         raise NotImplementedError("VM classes must implement this method")
 
     @classmethod
     @abstractmethod
-    def get_transaction_class(cls):
+    def get_transaction_class(cls) -> Any:
```
cburgdorf: Guess this should be `Type[BaseTransaction]`

eth/vm/base.py
```diff
@@ -237,17 +263,17 @@ def get_uncle_reward(block_number: int, uncle: BaseBlock) -> int:
     # Transactions
     #
     @abstractmethod
-    def create_transaction(self, *args, **kwargs):
+    def create_transaction(self, *args: Any, **kwargs: Any) -> Any:
```
cburgdorf: Guess this should return `BaseTransaction`

eth/vm/base.py
```diff
@@ -296,12 +324,12 @@ def validate_uncle(
     #
     @classmethod
     @abstractmethod
-    def get_state_class(cls):
+    def get_state_class(cls) -> Any:
```
cburgdorf: Guess this should return `Type[BaseState]`

eth/vm/base.py
```diff
@@ -440,12 +473,15 @@ def apply_all_transactions(self, transactions, base_header):
             receipts.append(receipt)
             computations.append(computation)
 
-        return result_header, receipts, computations
+        receipts_tuple = tuple(receipts)
```
cburgdorf: You can change `receipts` and `computations` to be tuples and then concat and reassign them in the loop. Something like:

```python
receipts = receipts + (receipt,)
```

Then you do not have to start with lists and convert them to tuples but simply build up the tuple you want directly.

eth/vm/forks/byzantium/__init__.py
```diff
@@ -70,10 +76,10 @@ class ByzantiumVM(SpuriousDragonVM):
     _state_class = ByzantiumState  # type: Type[BaseState]
 
     # Methods
-    create_header_from_parent = staticmethod(create_byzantium_header_from_parent)
-    compute_difficulty = staticmethod(compute_byzantium_difficulty)
+    create_header_from_parent = staticmethod(create_byzantium_header_from_parent)   # type: ignore
+    compute_difficulty = staticmethod(compute_byzantium_difficulty)     # type: ignore
```
cburgdorf: Can we use comment type hints to give these methods the proper `Callable` type instead of ignoring them or is there a reason why that wouldn't work?

eth/vm/forks/byzantium/transactions.py
```diff
@@ -1,21 +1,38 @@
...
+from eth.vm.forks.spurious_dragon.transactions import (
+    SpuriousDragonTransaction,
+    SpuriousDragonUnsignedTransaction,
+)
+
 
 class ByzantiumTransaction(SpuriousDragonTransaction):
     @classmethod
-    def create_unsigned_transaction(cls, *, nonce, gas_price, gas, to, value, data):
+    def create_unsigned_transaction(cls,                 # type: ignore
```
cburgdorf: I'd be interested to know why we need the ignore here

eth/vm/forks/byzantium/transactions.py
```diff
@@ -1,21 +1,38 @@
...
+                                    gas: int,
+                                    to: Address,
+                                    value: int,
+                                    data: bytes) -> BaseTransaction:
         return ByzantiumUnsignedTransaction(nonce, gas_price, gas, to, value, data)
 
 
 class ByzantiumUnsignedTransaction(SpuriousDragonUnsignedTransaction):
-    def as_signed_transaction(self, private_key, chain_id=None):
+    def as_signed_transaction(self,                          # type: ignore
```
cburgdorf: Same here, why do we need the ignore here?

eth/vm/forks/constantinople/__init__.py
```diff
@@ -28,11 +28,11 @@ class ConstantinopleVM(ByzantiumVM):
     _state_class = ConstantinopleState  # type: Type[BaseState]
 
     # Methods
-    create_header_from_parent = staticmethod(create_constantinople_header_from_parent)
-    compute_difficulty = staticmethod(compute_constantinople_difficulty)
+    create_header_from_parent = staticmethod(create_constantinople_header_from_parent)  # type: ignore  # noqa: E501
+    compute_difficulty = staticmethod(compute_constantinople_difficulty)    # type: ignore
```
cburgdorf: Looks again as if we could use type hint comments

eth/vm/forks/constantinople/transactions.py
```diff
@@ -10,12 +18,21 @@
 
 class ConstantinopleTransaction(ByzantiumTransaction):
     @classmethod
-    def create_unsigned_transaction(cls, *, nonce, gas_price, gas, to, value, data):
+    def create_unsigned_transaction(cls,             # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/constantinople/transactions.py
```diff
@@ -10,12 +18,21 @@
...
+                                    gas: int,
+                                    to: Address,
+                                    value: int,
+                                    data: bytes) -> BaseTransaction:
         return ConstantinopleUnsignedTransaction(nonce, gas_price, gas, to, value, data)
 
 
 class ConstantinopleUnsignedTransaction(ByzantiumUnsignedTransaction):
-    def as_signed_transaction(self, private_key, chain_id=None):
+    def as_signed_transaction(self,                         # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/frontier/__init__.py
```diff
@@ -62,22 +65,22 @@ class FrontierVM(VM):
     _state_class = FrontierState  # type: Type[BaseState]
 
     # methods
-    create_header_from_parent = staticmethod(create_frontier_header_from_parent)
-    compute_difficulty = staticmethod(compute_frontier_difficulty)
+    create_header_from_parent = staticmethod(create_frontier_header_from_parent)    # type: ignore
+    compute_difficulty = staticmethod(compute_frontier_difficulty)      # type: ignore
```
cburgdorf: Should be a case for type hint comments

eth/vm/forks/frontier/headers.py
```diff
@@ -85,7 +94,7 @@ def create_frontier_header_from_parent(parent_header, **header_params):
     return header
 
 
-def configure_frontier_header(vm, **header_params):
+def configure_frontier_header(vm: "FrontierVM", **header_params: Any) -> BlockHeader:
```
cburgdorf: We may be able to tighten the `header_params` with the `mypy_extensions` but that is nothing that needs to be addressed in this PR. Just pointing it out.

eth/vm/forks/frontier/transactions.py
```diff
@@ -76,13 +83,20 @@ def get_message_for_signing(self):
         ))
 
     @classmethod
-    def create_unsigned_transaction(cls, *, nonce, gas_price, gas, to, value, data):
+    def create_unsigned_transaction(cls,                     # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/frontier/transactions.py
```diff
@@ -92,7 +106,7 @@ def validate(self):
         validate_is_bytes(self.data, title="Transaction.data")
         super().validate()
 
-    def as_signed_transaction(self, private_key):
+    def as_signed_transaction(self, private_key: PrivateKey) -> BaseTransaction:    # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/homestead/transactions.py
```diff
@@ -40,12 +47,19 @@ def get_message_for_signing(self):
         ))
 
     @classmethod
-    def create_unsigned_transaction(cls, *, nonce, gas_price, gas, to, value, data):
+    def create_unsigned_transaction(cls,        # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/homestead/transactions.py
```diff
@@ -40,12 +47,19 @@ def get_message_for_signing(self):
...
+                                    gas: int,
+                                    to: Address,
+                                    value: int,
+                                    data: bytes) -> BaseTransaction:
         return HomesteadUnsignedTransaction(nonce, gas_price, gas, to, value, data)
 
 
 class HomesteadUnsignedTransaction(FrontierUnsignedTransaction):
-    def as_signed_transaction(self, private_key):
+    def as_signed_transaction(self, private_key: PrivateKey) -> BaseTransaction:    # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/spurious_dragon/transactions.py
```diff
@@ -33,33 +46,42 @@ def get_message_for_signing(self):
             ))
 
     @classmethod
-    def create_unsigned_transaction(cls, *, nonce, gas_price, gas, to, value, data):
+    def create_unsigned_transaction(cls,         # type: ignore
```
cburgdorf: Can we get rid of this ignore?

eth/vm/forks/spurious_dragon/transactions.py
```diff
@@ -33,33 +46,42 @@ def get_message_for_signing(self):
...
-    def v_min(self):
+    def v_min(self) -> int:     # type: ignore
         if is_eip_155_signed_transaction(self):
             return 35 + (2 * self.chain_id)
         else:
             return 27
 
     @property
-    def v_max(self):
+    def v_max(self) -> int:     # type: ignore
```
cburgdorf: interested to learn about the reason for the ignore here and above

eth/tools/fixtures/fillers/_utils.py
```diff
@@ -72,7 +73,7 @@ def calc_state_root(state: AccountState, account_db_class: Type[BaseAccountDB])
 
 def generate_random_keypair() -> Tuple[bytes, Address]:
     key_object = keys.PrivateKey(pad32(int_to_big_endian(random.getrandbits(8 * 32))))
-    return key_object.to_bytes(), key_object.public_key.to_canonical_address()
+    return key_object.to_bytes(), cast(Address, key_object.public_key.to_canonical_address())
```
cburgdorf: I believe this could get rid of the `cast` and simply be `Address(key_object.public_key.to_canonical_address())`

eth/vm/state.py
```diff
@@ -175,14 +200,16 @@ def get_ancestor_hash(self, block_number):
             ancestor_depth >= len(self.execution_context.prev_hashes)
         )
         if is_ancestor_depth_out_of_range:
-            return b''
+            return cast(Hash32, b'')
```
cburgdorf: Don't cast here, do a `Hash32(b'')` instead. In general: a cast always means that you are smarter than the type checker which means you can even cast something silly as `cast(Hash32, "foobar")`. So, if we are creating the type here, we don't need the cast and can simply use the constructor.

tox.ini
```diff
@@ -107,6 +107,7 @@ commands=
     mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
     mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth.utils
     mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth.tools
+    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth.vm
```
cburgdorf: Rad! This is lifting up the type guarantees for a huge part of the code base :heart:

eth/vm/base.py
Bhargavasomu: I tried that approach as follows in the beginning for which I got an error,
`receipts += (receipt)`.
But this method worked and optimized too. Glad to make both `type hinting` and `optimization` happen at the same time.

eth/tools/fixtures/fillers/_utils.py
Bhargavasomu: @cburgdorf I didn't know we could do that! Thankyou for pointing it out, will change it immediately.

eth/vm/forks/byzantium/transactions.py
Bhargavasomu: @cburgdorf this is the error we would be getting if we didn't ignore this situation.
`error: Signature of "create_unsigned_transaction" incompatible with supertype "BaseTransaction"`
This is occurring because the `create_unsigned_transaction` function in BaseTransaction is defined as follows
```
def create_unsigned_transaction(self, *args: Any, **kwargs: Any) -> 'BaseTransaction':
```
Whereas the same function is inherited and used in `class ByzantiumTransaction` as follows
```
def create_unsigned_transaction(cls,
                                    *,
                                    nonce: int,
                                    gas_price: int,
                                    gas: int,
                                    to: Address,
                                    value: int,
                                    data: bytes) -> BaseTransaction:
```

This phenomena was named as `Liskov Substitution Principle` when I googled it and mypy doesn't support such things as of now (Please correct me if I am wrong).

cburgdorf: LSP was violated in a bunch of places here (not by your changes, it's been violated before). I created another PR where I will aggregate some fixes. I address this particular problem in this commit.

https://github.com/ethereum/py-evm/pull/1452/commits/a992d2139b56c307340557360d4035772d06a070

eth/vm/forks/byzantium/__init__.py
cburgdorf: Turns out these are a bit harder to solve and I'm ok to address these in another PR

Bhargavasomu:
So the 2 takeaways I guess are 
* To tighten the signatures and types of the abstract functions (instead of `Any`)
* Typecasting directly without the usage of `cast` (`Address(...)`)

Will keep in mind the above things while type hinting the rest of the modules. Learned a lot of new things from here. Thankyou for helping me out here @cburgdorf.

eth/chains/base.py
```diff
@@ -592,7 +610,8 @@ def get_transaction_result(
         This is referred to as a `call()` in web3.
         """
         with self.get_vm(at_header).state_in_temp_block() as state:
-            computation = state.costless_execute_transaction(transaction)
+            # Ignore is to not bleed the SpoofTransaction deeper into the code base
+            computation = state.costless_execute_transaction(transaction)  # type: ignore
```
carver: Probably a better solution here is an abstract interface for the transaction that `SpoofTransaction` and the main RLP Transaction both implement.

cburgdorf: That ignore is gone in the final version that was merged.

PR #1456, opened 2018-11-09
Eth leftover type hinting

### LINKED ISSUE #1398: Add type hints to eth module

Bhargavasomu:
### What was wrong?
Completes the type hinting in the `eth` module.
Closes #1398 

### How was it fixed?
Adding the type hints manually and adjusting in other modules like `trinity` and `p2p`.

#### Cute Animal Picture

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -12,6 +12,7 @@ from typing import (  # noqa: F401
     cast,
     Dict,
     Generator,
+    Iterable,
     Iterator,
     List,
     Optional,
@@ -32,6 +33,12 @@ from eth_utils import (
     encode_hex,
 )
 
+from eth.constants import (
+    BLANK_ROOT_HASH,
+    EMPTY_UNCLE_HASH,
+    MAX_UNCLE_DEPTH,
+)
+
 from eth.db.backends.base import BaseAtomicDB
 from eth.db.chain import (
     BaseChainDB,
@@ -40,11 +47,7 @@ from eth.db.chain import (
 from eth.db.header import (
     HeaderDB,
 )
-from eth.constants import (
-    BLANK_ROOT_HASH,
-    EMPTY_UNCLE_HASH,
-    MAX_UNCLE_DEPTH,
-)
+
 from eth.estimators import (
     get_gas_estimator,
 )
@@ -53,15 +56,7 @@ from eth.exceptions import (
     TransactionNotFound,
     VMNotFound,
 )
-from eth.utils.spoof import (
-    SpoofTransaction,
-)
-from eth.validation import (
-    validate_block_number,
-    validate_uint256,
-    validate_word,
-    validate_vm_configuration,
-)
+
 from eth.rlp.blocks import (
     BaseBlock,
 )
@@ -76,6 +71,14 @@ from eth.rlp.transactions import (
     BaseTransaction,
     BaseUnsignedTransaction,
 )
+
+from eth.typing import (
+    AccountState,
+)
+
+from eth.utils.spoof import (
+    SpoofTransaction,
+)
 from eth.utils.db import (
     apply_state_dict,
 )
@@ -88,9 +91,15 @@ from eth.utils.headers import (
 from eth.utils.rlp import (
     validate_imported_block_unchanged,
 )
-from eth.typing import (
-    AccountState,
+
+from eth.validation import (
+    validate_block_number,
+    validate_uint256,
+    validate_word,
+    validate_vm_configuration,
 )
+from eth.vm.computation import BaseComputation
+from eth.vm.state import BaseState      # noqa: F401
 
 from eth._warnings import catch_and_ignore_import_warning
 with catch_and_ignore_import_warning():
@@ -107,7 +116,10 @@ with catch_and_ignore_import_warning():
     )
 
 if TYPE_CHECKING:
-    from eth.vm.base import BaseVM  # noqa: F401
+    from eth.vm.base import (     # noqa: F401
+        BaseVM,
+        VM,
+    )
 
 
 class BaseChain(Configurable, ABC):
@@ -164,7 +176,7 @@ class BaseChain(Configurable, ABC):
         return cls.get_vm_class_for_block_number(header.block_number)
 
     @abstractmethod
-    def get_vm(self, header: BlockHeader=None) -> 'BaseVM':
+    def get_vm(self, header: BlockHeader=None) -> 'VM':
         raise NotImplementedError("Chain classes must implement this method")
 
     @classmethod
@@ -196,11 +208,11 @@ class BaseChain(Configurable, ABC):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def get_canonical_head(self):
+    def get_canonical_head(self) -> BlockHeader:
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def get_score(self, block_hash):
+    def get_score(self, block_hash: Hash32) -> int:
         raise NotImplementedError("Chain classes must implement this method")
 
     #
@@ -227,11 +239,14 @@ class BaseChain(Configurable, ABC):
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def get_canonical_block_hash(self, block_number):
+    def get_canonical_block_hash(self, block_number: BlockNumber) -> Hash32:
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def build_block_with_transactions(self, transactions, parent_header):
+    def build_block_with_transactions(self,
+                                      transactions: Tuple[BaseTransaction, ...],
+                                      parent_header: BlockHeader=None
+                                      ) -> Tuple[BaseBlock, Tuple[Receipt, ...], Tuple[BaseComputation, ...]]:      # noqa: E501
         raise NotImplementedError("Chain classes must implement this method")
 
     #
@@ -320,7 +335,7 @@ class Chain(BaseChain):
     current block number.
     """
     logger = logging.getLogger("eth.chain.chain.Chain")
-    gas_estimator = None  # type: Callable
+    gas_estimator = None     # type: Callable[[BaseState, BaseTransaction], int]
 
     chaindb_class = ChainDB  # type: Type[BaseChainDB]
 
@@ -334,8 +349,8 @@ class Chain(BaseChain):
 
         self.chaindb = self.get_chaindb_class()(base_db)
         self.headerdb = HeaderDB(base_db)
-        if self.gas_estimator is None:
-            self.gas_estimator = get_gas_estimator()  # type: ignore
+        if self.gas_estimator is None:                  # type: ignore
+            self.gas_estimator = get_gas_estimator()    # type: ignore
 
     #
     # Helpers
@@ -403,7 +418,7 @@ class Chain(BaseChain):
     #
     # VM API
     #
-    def get_vm(self, at_header: BlockHeader=None) -> 'BaseVM':
+    def get_vm(self, at_header: BlockHeader=None) -> 'VM':
         """
         Returns the VM instance for the given block number.
         """
@@ -414,7 +429,9 @@ class Chain(BaseChain):
     #
     # Header API
     #
-    def create_header_from_parent(self, parent_header, **header_params):
+    def create_header_from_parent(self,
+                                  parent_header: BlockHeader,
+                                  **header_params: HeaderParams) -> BlockHeader:
         """
         Passthrough helper to the VM class of the block descending from the
         given header.
@@ -432,7 +449,7 @@ class Chain(BaseChain):
         validate_word(block_hash, title="Block Hash")
         return self.chaindb.get_block_header_by_hash(block_hash)
 
-    def get_canonical_head(self):
+    def get_canonical_head(self) -> BlockHeader:
         """
         Returns the block header at the canonical chain head.
 
@@ -440,7 +457,7 @@ class Chain(BaseChain):
         """
         return self.chaindb.get_canonical_head()
 
-    def get_score(self, block_hash):
+    def get_score(self, block_hash: Hash32) -> int:
         """
         Returns the difficulty score of the block with the given hash.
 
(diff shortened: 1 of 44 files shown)


Bhargavasomu:
@cburgdorf I have tried my level best to tighten the types. I have used `type: ignore` in the following conditions.
* Using `type: ignore` in `eth.chains.base` because the following error is coming 
`Invalid self argument "Chain" to attribute function "gas_estimator" with type "Callable[[BaseState, BaseTransaction], int]"`
* Using `type:ignore` where `__contains__` is defined and that particular class inherits MutableMapping or Mapping
* Where the breakage of LSP is causing mypy problem (Mentioned in code comments)

Apart from that in some places I was forced to use `VM` instead of `BaseVM` because the `BaseVM` class didn't have the `apply_all_transactions` method.

Will make the improvements as per the review.

eth/chains/base.py
```diff
@@ -164,7 +176,7 @@ def get_vm_class(cls, header: BlockHeader) -> Type['BaseVM']:
         return cls.get_vm_class_for_block_number(header.block_number)
 
     @abstractmethod
-    def get_vm(self, header: BlockHeader=None) -> 'BaseVM':
+    def get_vm(self, header: BlockHeader=None) -> 'VM':
```
cburgdorf: If you create an abstract method `apply_all_transactions` in `BaseVM` you should be able to keep these here on `BaseVM` rather than `VM`

eth/chains/base.py
```diff
@@ -227,11 +239,14 @@ def get_canonical_block_by_number(self, block_number: BlockNumber) -> BaseBlock:
...
-    def get_canonical_block_hash(self, block_number):
+    def get_canonical_block_hash(self, block_number: BlockNumber) -> Hash32:
         raise NotImplementedError("Chain classes must implement this method")
 
     @abstractmethod
-    def build_block_with_transactions(self, transactions, parent_header):
+    def build_block_with_transactions(self,
+                                      transactions: Tuple[BaseTransaction, ...],
+                                      parent_header: BlockHeader=None
+                                      ) -> Tuple[BaseBlock, Tuple[Receipt, ...], Tuple[BaseComputation, ...]]:      # noqa: E501
```
cburgdorf: I think our preference in cases like this is to move the entire `self` one line down to have more horizontal space and get rid of the `# noqa: E501`

eth/chains/base.py
```diff
@@ -320,7 +335,7 @@ class Chain(BaseChain):
     current block number.
     """
     logger = logging.getLogger("eth.chain.chain.Chain")
-    gas_estimator = None  # type: Callable
+    gas_estimator = None     # type: Callable[[BaseState, BaseTransaction], int]
```
cburgdorf: I noticed a couple of type comments come with an unnecessary number of white space before. That should be changed.

eth/chains/base.py
```diff
@@ -524,7 +541,10 @@ def get_canonical_block_hash(self, block_number: BlockNumber) -> Hash32:
         """
         return self.chaindb.get_canonical_block_hash(block_number)
 
-    def build_block_with_transactions(self, transactions, parent_header=None):
+    def build_block_with_transactions(self,
```
cburgdorf: same here, preferred way is to move the self one line down to get more horizontal space

eth/chains/base.py
```diff
@@ -627,7 +647,7 @@ def estimate_gas(
         if at_header is None:
             at_header = self.get_canonical_head()
         with self.get_vm(at_header).state_in_temp_block() as state:
-            return self.gas_estimator(state, transaction)
+            return self.gas_estimator(state, transaction)               # type: ignore
```
cburgdorf: too much whitespace before comment. I also want to see if we can get rid of the `ignore` altogether but I haven't yet.

eth/chains/base.py
```diff
@@ -890,7 +912,7 @@ def mine_block(self, *args: Any, **kwargs: Any) -> BaseBlock:
         self.header = self.create_header_from_parent(mined_block.header)
         return mined_block
 
-    def get_vm(self, at_header: BlockHeader=None) -> 'BaseVM':
+    def get_vm(self, at_header: BlockHeader=None) -> 'VM':
```
cburgdorf: Related to my previous comment, this could go back to `BaseVM`

eth/chains/header.py
```diff
@@ -82,7 +93,7 @@ class HeaderChain(BaseHeaderChain):
 
     def __init__(self, base_db: BaseDB, header: BlockHeader=None) -> None:
         self.base_db = base_db
-        self.headerdb = self.get_headerdb_class()(base_db)
+        self.headerdb = self.get_headerdb_class()(cast(BaseAtomicDB, base_db))
```
cburgdorf: Note to self: I have to double check why we need the cast here

eth/chains/header.py
```diff
@@ -99,15 +110,15 @@ def from_genesis_header(cls,
         """
         Initializes the chain from the genesis header.
         """
-        headerdb = cls.get_headerdb_class()(base_db)
+        headerdb = cls.get_headerdb_class()(cast(BaseAtomicDB, base_db))
```
cburgdorf: same here

eth/db/cache.py
```diff
@@ -8,24 +10,24 @@ class CacheDB(BaseDB):
...
         self._db = db
         self._cache_size = cache_size
         self.reset_cache()
 
-    def reset_cache(self):
+    def reset_cache(self) -> None:
         self._cached_values = LRU(self._cache_size)
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: Any) -> Any:
```
cburgdorf: Could these `Any` all be `bytes`?

eth/db/journal.py
```diff
@@ -121,7 +121,7 @@ def commit_changeset(self, changeset_id: uuid.UUID) -> Dict[bytes, bytes]:
     #
     # Database API
     #
-    def __getitem__(self, key: bytes) -> Union[bytes, DeletedEntry]:
+    def __getitem__(self, key: bytes) -> Union[bytes, DeletedEntry]:    # type: ignore # Breaks LSP
```
cburgdorf: I'll double check this and other cases of `# type: ignore` in another round of review when I get my hands on the code. Just reviewing through github now.

tox.ini
```diff
@@ -104,10 +104,7 @@ commands=
     {[common-lint]commands}
     flake8 {toxinidir}/tests --exclude="trinity,p2p"
     # TODO: Drop --ignore-missing-imports once we have type annotations for eth_utils, coincurve and cytoolz
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs -p eth
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth.utils
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth.tools
-    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth.vm
+    mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth
```
cburgdorf: Exciting! :heart:

trinity/chains/light.py
```diff
@@ -59,7 +59,7 @@
 )
 
 if TYPE_CHECKING:
-    from eth.vm.base import BaseVM  # noqa: F401
+    from eth.vm.base import VM  # noqa: F401
```
cburgdorf: Another case where we could stay on `BaseVM` if we just ensure we create that abstract method

Bhargavasomu:
@cburgdorf made the changes requested. Please take a look.

eth/db/diff.py
```diff
@@ -53,72 +66,74 @@ class DBDiffTracker(MutableMapping):
...
 
-    def __getitem__(self, key):
+    def __getitem__(self, key: bytes) -> bytes:
         result = self._changes.get(key, NEVER_INSERTED)
         if result in (DELETED, NEVER_INSERTED):
-            raise DiffMissingError(key, result)
+            raise DiffMissingError(key, cast(MissingReason, result))
         else:
-            return result
+            return cast(bytes, result)
```
cburgdorf: @pipermerriam @carver I'm worried about the usage of `cast` in performance sensitive code paths.

Here's @gvanrossum from 2016

>I would really hope we won't need to use or encourage casts here, since it's an expensive runtime operation (at least it is until we teach CPython about it -- right now it is a user-defined function, which is incredibly slow compared to a plain assignment)

I'm not sure if anything has changed about this in the meantime but if not then I assume that even if `cast` is a `noop` at runtime it is still one extra function call and hence will slow things down. Therefore I'd advocate to use `type: ignore` in such performance sensitive code paths. Thoughts?

cburgdorf:
@Bhargavasomu well, it shouldn't matter for **most** cases but unless CPython optimized out the `cast` it is still a noop (a function that does nothing) and if a function is called in performance critical code which is called many times per in a loop then that may still be a problem.

See also my citation of Guido van Rossum (creator of Python)

>I would really hope we won't need to use or encourage casts here, since it's an expensive runtime operation (at least it is until we teach CPython about it -- right now it is a user-defined function, which is incredibly slow compared to a plain assignment)

I'm just trying to play it safe here. So my personal preference **for performance critical code** would be to use `type: ignore` over `cast` if needed (at all).

eth/db/diff.py
carver: If the best solution we have is a cast, then yes, a `type: ignore` seems like a fine alternative to me. (given that it's in a performance-sensitive line, and the line is doing exactly one thing -- we don't want to accidentally ignore other type checks on the same line).

Bhargavasomu:
@cburgdorf could you please tell me which files are performance critical, so that I could replace `cast` with `type: ignore`?

cburgdorf:
@Bhargavasomu I'll take care of that. Applying some final finish.

cburgdorf:
@Bhargavasomu solid work! :+1: 

I wrangled with some corner cases to get rid of some `cast` and `type: ignore` and also migrated perf critical `cast`s to `type: ignore` (as discussed).

I believe there are still plenty of cases where we could do better so if you like that sort of stuff feel free to continue wrangling with `type: ignore`, `cast` and `Any` and send follow up PRs :)

Bhargavasomu:
Thankyou for helping me out @cburgdorf

rayrapetyan:
hey guys, what about other bounty-marked tickets merged recently lol ))

cburgdorf:
@rayrapetyan are you saying we missed paying out a bounty? Can you point me to it?

rayrapetyan:
@cburgdorf, e.g. #1402, thanks!

cburgdorf:
@rayrapetyan I'm afraid that ticket did not have a Gitcoin bounty assigned. I assume that the `bounty` label was added with the intention to channel it through the bounty process but then it didn't happen for whatever reasons (probably just forgotten)

I'm afraid we can't retroactively give out money for work that has already happened and had no bounty associated at that time. But I recognize that you've been doing great work and there's plenty of opportunity for you to do paid work on this project.

How would you feel about writing a Proof of Concept GraphQL plugin as described in https://github.com/ethereum/py-evm/issues/1462. Is this something that you'd have fun with?

PR #1463, opened 2018-11-11
Ensure ethstats plugin handles incorrect server responses

cburgdorf:
### What was wrong?

When the ethstats plugin connects without secret we get a string response instead of the expected JSON object. This then further leads to a `TypeError` if we try to access an attribute on that string.

### How was it fixed?

Check if `raw_message` is a string and raise an exception. Also ensure ethstats errors are logged as `warning` not `info`.

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/builtin/ethstats/ethstats_client.py
+++ b/trinity/plugins/builtin/ethstats/ethstats_client.py
@@ -59,7 +59,7 @@ class EthstatsClient(BaseService):
             try:
                 message: EthstatsMessage = self.deserialize_message(json_string)
             except EthstatsException as e:
-                self.logger.info('Cannot parse message from server: %s' % e)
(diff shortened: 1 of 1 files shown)


trinity/plugins/builtin/ethstats/ethstats_client.py
```diff
@@ -84,6 +84,9 @@ def deserialize_message(self, json_string: str) -> EthstatsMessage:
         except json.decoder.JSONDecodeError as e:
             raise EthstatsException('Received incorrect JSON: %s' % e)
 
+        if isinstance(raw_message, str):
+            raise EthstatsException(f'Received incorrect payload: {raw_message}')
```
pipermerriam: "invalid payload" seems more appropriate.

PR #1464, opened 2018-11-12
Ensure ethstats plugin uses a timeout to wait on PeerCount

### LINKED ISSUE #1379: `ethstats` plugin stuck trying to get peer count
* `py-evm`: `trinity-v0.1.0-alpha.16-71-gca9f90a6` (current `master` at commit ca9f90a6c08699c2aa8989b233a23cd533f0aa02)
* OS: Arch Linux
* Python: CPython 3.7.0
* Environment (output of `pip freeze`): NAH

### What is wrong?

After PR #1343 (blamed by `git bisect`), the `ethstats` plug-in stopped working again (in both "light" mode and "full" sync mode). For ref, run as:

```
PYTHONWARNINGS='ignore' trinity --network-id=3 --sync-mode=light --log-level=debug \
    --ethstats --ethstats-server-url="wss://ropsten-stats.parity.io/api" \
    --ethstats-server-secret="SECRET" \
    --ethstats-node-id="veox-trinity-dev-lightclient-debug" \
    --ethstats-node-contact="veox" \
    --max-peers=0
```

A traceback (don't read in: it's misleading):

``` python
   ERROR  10-09 17:53:26        EthstatsClient  Unexpected error in <trinity.plugins.builtin.ethstats.ethstats_client.EthstatsClient object at 0x7fcdd4ee1438>, exiting
  File "/home/veox/src/py-evm/p2p/service.py", line 103, in run
    await self._run()
  File "/home/veox/src/py-evm/trinity/plugins/builtin/ethstats/ethstats_client.py", line 52, in _run
    self.recv_handler(),
  File "/home/veox/src/py-evm/p2p/cancellable.py", line 41, in wait_first
    return await token_chain.cancellable_wait(*awaitables, timeout=timeout)
  File "/home/veox/src/py-evm/.virtualenv/py-evm/lib/python3.7/site-packages/cancel_token/token.py", line 152, in cancellable_wait
    return done.pop().result()
  File "/home/veox/src/py-evm/trinity/plugins/builtin/ethstats/ethstats_client.py", line 59, in recv_handler
    json_string: str = await self.websocket.recv()
  File "/home/veox/src/py-evm/.virtualenv/py-evm/lib/python3.7/site-packages/websockets/protocol.py", line 352, in recv
    yield from self.ensure_open()
  File "/home/veox/src/py-evm/.virtualenv/py-evm/lib/python3.7/site-packages/websockets/protocol.py", line 514, in ensure_open
    self.close_code, self.close_reason) from self.transfer_data_exc
websockets.exceptions.ConnectionClosed: WebSocket connection is closed: code = 1000 (OK), no reason
```

The "blocker", at least within `ethstats`, are these lines:

https://github.com/ethereum/py-evm/blob/ca9f90a6c08699c2aa8989b233a23cd533f0aa02/trinity/plugins/builtin/ethstats/ethstats_service.py#L137-L139

A response for `PeerCountRequest` never arrives from the event bus. The plugin keeps waiting, never submitting any stats; eventually the remote server closes the `websocket`. That makes it look like the issue is network-related.

Commenting out this retrieval, and hard-coding the peer count number to be reported as `1337` is sufficient for the node to start behaving "normally": e.g. regularly submit latest chain head data to the remote, etc..

### How can it be fixed

The actual "culprit" is unknown to me. :(

Starting a `trinity --network-id=3 attach` session and requesting `web3.net.peerCount` works fine, so this might be related to plugin/service contexts. (The `attach` and `ethstats` plugins are quite different.)

Tried looking at whether `self.context.event_bus._running` for the plugin is `True`. In all cases, it already was by the time it reached that "hanging" point. So probably not a case of an event bus that's not yet running.

PR #1343 _did_ make some changes to how peer count requests are handled. I don't understand it exactly, so can't say for sure that it's relevant. Probably best to start there.

cburgdorf:
### What was wrong?

Previously, the Ethstats Plugin waited on the PeerCount
forever. But since the PeerPool is not ready by the time
that the Ethstats plugin sends the request, it just hangs
there waiting and at some point the websocket connection
times out.

### How was it fixed?

This change introduces a timeout of 500ms for the PeerPool
to answer. If the PeerPool does not answer within 500ms,
we assume 0 peers and move on.

Fixes #1379

I tested this against ethstats and it seems the plugin stays alive and reports the peer count correctly as seen in the following screenshot.

image

#### Cute Animal Picture

### DIFF
--- a/trinity/plugins/builtin/ethstats/ethstats_service.py
+++ b/trinity/plugins/builtin/ethstats/ethstats_service.py
@@ -7,7 +7,6 @@ from eth.chains.base import (
 )
 from p2p.events import (
     PeerCountRequest,
-    PeerCountResponse,
 )
 from p2p.service import (
     BaseService,
@@ -134,13 +133,19 @@ class EthstatsService(BaseService):
         }
 
     async def get_node_stats(self) -> EthstatsData:
-        response: PeerCountResponse = await self.context.event_bus.request(
-            PeerCountRequest()
-        )
+
+        try:
+            peer_count = (await self.wait(
+                self.context.event_bus.request(PeerCountRequest()),
+                timeout=0.5
+            )).peer_count
+        except TimeoutError:
+            self.logger.warning("Timeout: PeerPool did not answer PeerCountRequest")
+            peer_count = 0
 
         return {
             'active': True,
-            'peers': response.peer_count,
+            'peers': peer_count,
         }
 
     def get_chain(self) -> BaseChain:


trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -134,13 +134,19 @@ def get_node_block(self) -> EthstatsData:
         }
 
     async def get_node_stats(self) -> EthstatsData:
-        response: PeerCountResponse = await self.context.event_bus.request(
-            PeerCountRequest()
-        )
+
+        try:
+            peer_count = (await asyncio.wait_for(
```
pipermerriam: Since this is inside of a `BaseService` class , I think we need to do `self.wait(awaitable, timeout=0.5)` so that it properly respect any triggered cancellations.

trinity/plugins/builtin/ethstats/ethstats_service.py
```diff
@@ -134,13 +134,19 @@ def get_node_block(self) -> EthstatsData:
...
+                self.context.event_bus.request(PeerCountRequest()),
+                timeout=0.5
+            )).peer_count
+        except asyncio.TimeoutError:
+            self.logger.warning("Timeout: PeerPool did not answer PeerCountRequest")
+            peer_count = 0
```
pipermerriam: Maybe instead of defaulting to zero it should retry up to `N` times or for up to `T` seconds since this is more likely a race condition that if we wait a little we should be able to get a *real* answer.

I don't feel strongly about this.

cburgdorf: That's true. I deliberately went for a stop gap solution here because I believe we want to revisit this anyway when we start adding more infos to the `get_node_stats` response. So if you don't feel strongly about getting this fixed *now* I'd prefer to just keep it as is for now.

PR #1469, opened 2018-11-15
Log when fast-sync has finished

cburgdorf:
### What was wrong?

We currently do not log when the fast-sync has finished. Two reasons to have this:

1. Achieve symmetry (we do log when fast sync begins)
2. Have it easier to grep the logs to see if a fast-sync finished or not

### How was it fixed?

Add a `info` log when the fast sync has finished.

#### Cute Animal Picture

### DIFF
--- a/trinity/sync/full/service.py
+++ b/trinity/sync/full/service.py
@@ -42,7 +42,7 @@ class FullNodeSyncer(BaseService):
         # See https://github.com/ethereum/py-evm/issues/654 for more details
         if head.timestamp < time.time() - FAST_SYNC_CUTOFF:
             # Fast-sync chain data.
-            self.logger.info("Starting fast-sync; current head: #%d", head.block_number)
+            self.logger.info("Starting fast-sync; current head: %s", head)
             fast_syncer = FastChainSyncer(
                 self.chain,
                 self.chaindb,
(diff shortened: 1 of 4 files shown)


trinity/sync/full/service.py
```diff
@@ -50,14 +50,16 @@ def __init__(self,
                 self.cancel_token,
             )
             await fast_syncer.run()
+
+            head = await self.wait(self.chaindb.coro_get_canonical_head())
+            self.logger.info("Finished fast fast-sync; current head: #%d", head.block_number)
```
pipermerriam: Thoughts on logging the start and end block?  If we don't overwrite the `head` variable we can have both the start and end.  Also, it'd be good to output the first 4-6 hex chars of the block hash since small re-orgs complicate things and it's useful to have in the logs.

carver: Agreed about showing start & end.

`str(head)` shows the info we want, I believe (the prefix of the block hash & block number).

cburgdorf: Yep, looks like this:

```
Finished fast fast-sync; previous head: <BlockHeader #902469 0233bc67>, current head: <BlockHeader #902469 0233bc67>
```

So, it is formatted as `<BlockHeader #num hash>` but I actually think I like that.
I also changed the other places for consistency.

There are a few other places in the rest of the code base that currently print `%d head.blocknumber` that could also be changed to `%s head` for consistency.

cburgdorf: I changed all the other places that could also benefit from showing the hash https://github.com/ethereum/py-evm/pull/1469/commits/236e8295d36066245f28120d54a1f4297e74f5c9

PR #1470, opened 2018-11-15
Ensure state tests only run against the intended fork

cburgdorf:
### What was wrong?

The statetest fixtures have different definitions
for each fork and we must ensure to only run
test against against the intended fork (e.g run
Constantinople state tests against Constantinople VM).

We can not rely on `pytest -k` matching for that
as that matches against an identification string that
includes the path and name of the test file which in
some cases also contains fork names. A test file may
be named "ConstantinopleSomething.json" but still
contains individual definitions per fork.

### How was it fixed?

Add a mechanism so that the intended fork can be passed as `--fork` and then internally filter the generated fixtures against that.

Notice that this does only apply to state tests as these are the only test files expands into multiple test cases per fork.

The following table compares the number of executed tests per fork based on this new scheme. 

| State test  | Run before | Run after |
| ------------- | ------------- | -------------|
| Frontier |1033|942
| Homestead  | 2196 | 2031 
| EIP150 |1281|1112
| EIP158  | 1177  | 1168  |
| Byzantium |4714|4714

#### Cute Animal Picture

### DIFF
--- a/eth/tools/fixtures/loading.py
+++ b/eth/tools/fixtures/loading.py
@@ -77,7 +77,7 @@ def load_fixture(fixture_path: str,
 def filter_fixtures(all_fixtures: Iterable[Any],
                     fixtures_base_dir: str,
                     mark_fn: Callable[[str, str], bool]=None,
-                    ignore_fn: Callable[[str, str], bool]=None) -> Any:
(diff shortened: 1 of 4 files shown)


pipermerriam:
It's pretty nice that this resulted in a non-trivial drop in total number of tests we run.  :+1:

PR #1473, opened 2018-11-16
Cleanup classifiers and add python_requires to setup.py

cburgdorf:
### What was wrong?

Outdated info in `setup.py`. Also pip should prevent people from trying to install Trinity on Python versions below `3.6`.

Related: https://github.com/ethereum/snake-charmers-internal-handbook/issues/3

### How was it fixed?

Magic!

I believe that the `python_requires` change does only affect people who try to install Trinity from pypi as anyone else who would `pip install -e [trinity]` directly from the repository root, would hit the `setup.py` file instead of the `setup_trinity.py` file.

#### Cute Animal Picture

### DIFF
--- a/setup_trinity.py
+++ b/setup_trinity.py
@@ -23,12 +23,13 @@ setup(
     keywords='ethereum blockchain evm trinity',
     packages=[],
     classifiers=[
-        'Development Status :: 2 - Pre-Alpha',
+        'Development Status :: 3 - Alpha',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
         'Programming Language :: Python :: 3.6',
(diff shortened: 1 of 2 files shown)


setup_trinity.py
```diff
@@ -23,12 +23,13 @@
...
     packages=[],
     classifiers=[
-        'Development Status :: 2 - Pre-Alpha',
+        'Development Status :: 3 - Alpha',
         'Intended Audience :: Developers',
         'License :: OSI Approved :: MIT License',
         'Natural Language :: English',
         'Programming Language :: Python :: 3.6',
     ],
+    python_requires=">=3.6"
```
pipermerriam: Upper bound of `<4` seems to be what we're doing elsewhere.

PR #1476, opened 2018-11-18
Ensure network connection is closed when handshake with peer fails

### LINKED ISSUE #1403: ResourceWarning: unclosed resource <TCPTransport
### What is wrong?

When running `trinity`, there are several warnings like this:
```
ResourceWarning: unclosed resource <TCPTransport closed=False reading=False 0x2ab9dc8>
```

_You'll probably find more helpful results online with `ResourceWarning: unclosed transport <_SelectorSocketTransport ...`._

_The text in the original quoted warning is produced by `uvloop` which reports the error differently and has less hits on a web search._

This is one of many different kinds of warnings mentioned in #1370 

The goal is not simply to hide these warnings but to resolve them.

### How can it be fixed

It will take some investigation. You should be able to see stack trace info for the warning by starting trinity with:
```sh
PYTHONWARNINGS=error::ResourceWarning trinity
```

It looks like roughly the solution is to do an `await asyncio.sleep(0)` in between closing a transport and letting the variable go out of scope (or explicitly deleting the instance). Or maybe there is a transport that is never being explicitly closed.

pipermerriam:
fixes #1403 

### What was wrong?

As part of our peer connection logic we initiate a low level connection, then create a `Peer` instance, then perform a handshake, followed by starting the peer service.  If anything goes wrong before the service starts, the connection will be left open causing these warnings.

### How was it fixed?

wrapped the handshake step in a `try/except` which closes the connection in the event of an error.

#### Cute Animal Picture

bearlive05_04

### DIFF
--- a/p2p/auth.py
+++ b/p2p/auth.py
@@ -57,8 +57,16 @@ async def handshake(
     use_eip8 = False
     initiator = HandshakeInitiator(remote, privkey, use_eip8, token)
     reader, writer = await initiator.connect()
-    aes_secret, mac_secret, egress_mac, ingress_mac = await _handshake(
-        initiator, reader, writer, token)
+    try:
+        aes_secret, mac_secret, egress_mac, ingress_mac = await _handshake(
+            initiator, reader, writer, token)
+    except Exception:
+        if not reader.at_eof():
+            reader.feed_eof()
+        writer.close()
+        await asyncio.sleep(0)
+        raise
+
     return aes_secret, mac_secret, egress_mac, ingress_mac, reader, writer
 
 
--- a/p2p/auth.py
+++ b/p2p/auth.py
@@ -61,6 +61,10 @@ async def handshake(
         aes_secret, mac_secret, egress_mac, ingress_mac = await _handshake(
             initiator, reader, writer, token)
     except Exception:
+        # Note: This is one of two places where we manually handle closing the
+        # reader/writer connection pair in the event of an error during the
+        # peer connection and handshake process.
+        # See `p2p.peer.handshake` for the other.
         if not reader.at_eof():
             reader.feed_eof()
         writer.close()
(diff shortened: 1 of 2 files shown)


pipermerriam:
I'll need to dig in a bit to validate where I think the remainder of these are coming from.  The first place I want to investigate is:

https://github.com/ethereum/py-evm/blob/1f60fd8576d4959c7d4642f52ac56158658cc9ae/p2p/peer.py#L934-L936

First we initiate the connection and then call `PeerPool.start_peer(peer)`.  This ends up calling `self.run_child_service()` which *starts* the peer's service component, but only by adding it as a task that it tracks.  I'm looking for whether it's possible for the peer to disconnect before reaching:

https://github.com/ethereum/py-evm/blob/1f60fd8576d4959c7d4642f52ac56158658cc9ae/p2p/service.py#L101-L103

If so, then the `_cleanup` will never be called.

pipermerriam:
Some loose experimentation suggests this is a good enough starting point that it could be merged and we can chase down the remaining warnings in a later PR.

pipermerriam:
Nevermind, 3149871 seems to have cleaned up the remaining warnings.

p2p/auth.py
```diff
@@ -57,8 +57,16 @@
     use_eip8 = False
     initiator = HandshakeInitiator(remote, privkey, use_eip8, token)
     reader, writer = await initiator.connect()
-    aes_secret, mac_secret, egress_mac, ingress_mac = await _handshake(
-        initiator, reader, writer, token)
+    try:
+        aes_secret, mac_secret, egress_mac, ingress_mac = await _handshake(
```
cburgdorf: I wonder if this could be rewritten using an `asynccontextmanager` so that becomes:

```python

async with initiator.connect() as reader, writer:
    aes_secret, mac_secret, egress_mac, ingress_mac = await _handshake(
            initiator, reader, writer, token)
```

pipermerriam: In theory yes.  I've thought about this a bit but I'm unsure enough about it that I'd like to punt, with the compromise being adding a larger comment to the section explaining why the solution isn't ideal.

pipermerriam:
Don't restart the jobs that failed to report in due to rate limiting, I'm surfacing this with github support and they need to be able to view the failures.

Issue #1466, opened 2018-11-14
EIP170 Contract Size Limit


fubuloubu:
### What is wrong?
Double checking this is the intended functionality.

EIP170 states that the contract size limit was changed to `2**14 + 2**13` which is 24,576 bytes. The following line implementations the constant for EIP170, but it is off by one.

https://github.com/ethereum/py-evm/blob/0ff7bc0f820250d7f15e804bc09870776ea34eb5/eth/vm/forks/spurious_dragon/constants.py#L13

### How can it be fixed
Validate that the number is correctly applied. If it is, consider making the number match and modifying the condition that tests for it so that it works with the actual number described in EIP170.

carver:
> Validate that the number is correctly applied.

Yup a [quick search](https://github.com/ethereum/py-evm/search?q=EIP170_CODE_SIZE_LIMIT&unscoped_q=EIP170_CODE_SIZE_LIMIT) shows that it is functioning correctly.

https://github.com/ethereum/py-evm/blob/ae1e3e4fc133f36db2238f61dac05a4e4e344319/eth/vm/forks/spurious_dragon/computation.py#L41

---

> consider making the number match and modifying the condition that tests for it

The change seems fine to me, I agree that it's at least a slight improvement.

pipermerriam:
Might be good for us to write a non-JSON-fixture based test to test for this specific thing.  We rely on the JSON-fixture tests for compliance with these things, but it'd be good to have a legible test as well (similar to what @cburgdorf has done for many of our opcodes).

glaksmono:
Would take to take a stab on this

pipermerriam:
@glaksmono :+1:

dherykw:
@pipermerriam , is the issue solved with the #1799 pull request?

e3243eric:
I dug into this issue. I have some thoughts about this. Please let me know if there are any mistakes.

- opcodes for deploy contract are the "init code". It copies/stores the contract code in memory and should return the memory address at the end.

- The condition of this line should be '>' instead of '>='.
https://github.com/ethereum/py-evm/blob/12f3bee964dc07bf3425beb3776639ad06f0648c/eth/vm/forks/spurious_dragon/computation.py#L86

kclowes:
Completed via #2083

PR #1483, opened 2018-11-20
[WIP] Update the contract limit size to follow EIP170 and add tests against it

glaksmono:
### What was wrong?
EIP170 states that the contract size limit was changed to `2**14 + 2**13` which is `24,576 bytes`. The following line implementations the constant for EIP170, but it is off by one.

Issue Reference: https://github.com/ethereum/py-evm/issues/1466

### How was it fixed?

* Update the contract limit size to follow EIP170
* Add tests against the contract size limit

#### Cute Animal Picture


tests/core/vm/test_spurious_dragon_computation.py
```diff
@@ -0,0 +1,80 @@
...
+)
+
+
+NORMALIZED_ADDRESS_A = "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6"
+NORMALIZED_ADDRESS_B = "0xcd1722f3947def4cf144679da39c4c32bdc35681"
+CANONICAL_ADDRESS_A = to_canonical_address("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6")
+CANONICAL_ADDRESS_B = to_canonical_address("0xcd1722f3947def4cf144679da39c4c32bdc35681")
+CONTRACT_CODE_A = b""
+CONTRACT_CODE_B = b""
+CONTRACT_CODE_C = b""
```
glaksmono: @pipermerriam Would like to get suggestion on what is the best way of having different kinds of contract code in our tests. Do we usually just hardcode the `bytecode` here?

dherykw:
Is this issue still open? it seems that is pretty close to being resolved, May I give it a try?

eth/vm/forks/spurious_dragon/constants.py
```diff
@@ -10,4 +10,4 @@
 
 
 # https://github.com/ethereum/EIPs/issues/170
-EIP170_CODE_SIZE_LIMIT = 24577
+EIP170_CODE_SIZE_LIMIT = 24576
```
pipermerriam:
```suggestion
EIP170_CODE_SIZE_LIMIT = 24576  # 2**14 + 2**13
```

tests/core/vm/test_spurious_dragon_computation.py
```diff
@@ -0,0 +1,85 @@
...
+)
+
+from eth.vm.transaction_context import (
+    BaseTransactionContext,
+)
+
+
+NORMALIZED_ADDRESS_A = "0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6"
+NORMALIZED_ADDRESS_B = "0xcd1722f3947def4cf144679da39c4c32bdc35681"
+CANONICAL_ADDRESS_A = to_canonical_address("0x0f572e5295c57f15886f9b263e2f6d2d6c7b5ec6")
```
pipermerriam:
```suggestion
CANONICAL_ADDRESS_A = to_canonical_address(NORMALIZED_ADDRESS_A)
```

tests/core/vm/test_spurious_dragon_computation.py
```diff
@@ -0,0 +1,85 @@
...
+CANONICAL_ADDRESS_B = to_canonical_address("0xcd1722f3947def4cf144679da39c4c32bdc35681")
```
pipermerriam:
```suggestion
CANONICAL_ADDRESS_B = to_canonical_address(NORMALIZED_ADDRESS_B)
```

tests/core/vm/test_spurious_dragon_computation.py
pipermerriam: Typically yes.  I assume we don't need complex code and we could just do something like: `CODE = b'\x01' * 20000` or something to keep it concise.

tests/core/vm/test_spurious_dragon_computation.py
```diff
@@ -0,0 +1,85 @@
...
+        gas=100,
+    )
+    computation = SpuriousDragonComputation(
+        state=state,
+        message=message_contract_code_a,
+        transaction_context=transaction_context,
+    )
+
+    """
+    TODO: CONTRACT_CODE_B size is equal to EIP170_CODE_SIZE_LIMIT
```
pipermerriam: Test should probably be parametrized (or use a parametrized fixture) instead of testing all three in the same test.

PR #1490, opened 2018-11-20
Replace unicode char with three dots

cburgdorf:
### What was wrong?

Recent logging changes introduced a non-ascii symbol that may cause an exception if it isn't supported.

### How was it fixed?

Using `...` seams like the safer option.

#### Cute Animal Picture


pipermerriam:
Still confused as to what is causing the attempt at `ascii` encoding.  Works fine on my machine :smiling_imp:

cburgdorf:
¯\_(ツ)_/¯

PR #1492, opened 2018-11-20
Christoph/feat/custom genesis

cburgdorf:
### What was wrong?

This is an attempt to pickup and finish the work that was started in #1299. Goal is to support custom genesis files.

This is just an early PR after an initial rebase to see where things are.

#### Cute Animal Picture


trinity/utils/eip1085.py
```diff
@@ -0,0 +1,277 @@
...
+    TangerineWhistleVM,
+    FrontierVM,
+    HomesteadVM as BaseHomesteadVM,
+    SpuriousDragonVM,
+    ByzantiumVM,
+    ConstantinopleVM,
+)
+
+
+T_RAW_EIP1085 = Dict[str, Any]
```
cburgdorf: @pipermerriam just hinting that the `T_*` notation is usually associated with type params (used with generics) and not for type aliases which should use `PascalCase` as other regular classes

pipermerriam:
> I will continue this is in #1299 to keep the discussion at a central place.

Feel free to start a new issue at any point to clean up the discussion history and better capture the current approach and simply reference #1299 for historical purposes.

Issue #1494, opened 2018-11-21
Regular sync throws exception


cburgdorf:
### What is wrong?

I wanted to force a regular sync (so I slightly modified the code to do that until we have #1480). 

I ran into the following exception which I believe is a regression from when we introduced atomic writes.

```
 WARNING  11-21 17:36:26    RegularChainSyncer  Task <coroutine object RegularChainSyncer._import_ready_blocks at 0x7f6262706780> finished unexpectedly: 'DBProxy' object has no attribute 'atomic_batch'
```

### How can it be fixed

Probably need to expose `atomic_batch` on the `DBProxy`

cburgdorf:
This is looking to be problematic. Just exposing `atomic_batch` on the `DBProxy` doesn't seem to work.

```
 WARNING  11-21 18:02:45    RegularChainSyncer  Task <coroutine object RegularChainSyncer._import_ready_blocks at 0x7ff7d7282728> finished unexpectedly: 
---------------------------------------------------------------------------
Unserializable message: Traceback (most recent call last):
  File "/usr/lib/python3.6/multiprocessing/managers.py", line 283, in serve_client
    send(msg)
  File "/usr/lib/python3.6/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/lib/python3.6/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
TypeError: can't pickle generator objects
```

:sob:

pipermerriam:
@cburgdorf I think atomic batch has to be implemented on the client side of the two processes, wrapping the proxy DB such that it knows how to do the atomic batching locally.

PR #1496, opened 2018-11-21
Ensure atomic_batch works over DBProxy

cburgdorf:
### What was wrong?

Chains access the db via the `DBProxy`. When `atomic_batch` was added it wasn't added to the `DBProxy` which results in run time crashes for several chain APIs. See #1494 

### How was it fixed?

Added `atomic_batch` API to the `DBProxy`

#### Cute Animal Picture


cburgdorf:
@pipermerriam maybe we should change our tests for fast/regular sync to also use the `DBProxy` (similar to the test in this PR). Then we have a more realistic test scenario and catch these things during CI.

PR #1497, opened 2018-11-21
Move some sync work into an isolated plugin

cburgdorf:
### What was wrong?

This is a spin off from #1299 to reduce scope of #1299. Will soon come to life.

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/core/p2p-proto/test_sync.py
+++ b/tests/trinity/core/p2p-proto/test_sync.py
@@ -9,7 +9,10 @@ from eth import constants
 from eth.tools.mining import POWMiningMixin
 from eth.vm.forks.frontier import FrontierVM
 
-
+from trinity.plugins.builtin.block_importer import (
+    BlockImportHandler,
+    EventBusBlockImporter,
+)
 from trinity.protocol.eth.servers import ETHRequestServer
 from trinity.protocol.les.peer import LESPeer
 from trinity.protocol.les.servers import LightRequestServer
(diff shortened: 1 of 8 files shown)


tests/trinity/core/p2p-proto/test_sync.py
```diff
@@ -80,20 +83,33 @@ def finalizer():
...
+    client_block_import_handler = BlockImportHandler(client_chain, event_bus)
+    client_block_import_api = EventBusBlockImporter(event_bus)
+
     client = RegularChainSyncer(
-        FrontierTestChain(chaindb_fresh.db),
+        client_chain,
         chaindb_fresh,
-        MockPeerPoolWithConnectedPeers([client_peer]))
+        MockPeerPoolWithConnectedPeers([client_peer]),
+        block_import_fn=client_block_import_api.coro_import_block)
```
cburgdorf: @carver @pipermerriam This test is a simulation of the idea that I described here: https://github.com/ethereum/py-evm/pull/1299#issuecomment-440648907

This offloads the `coro_import_block` call so that it doesn't run in the `networking` process (as it does when we merge  #1299) but in the process that hosts the `BlockImportHandler`. The idea is that this would run as an isolated plugin. A separate process just dedicated for block processing / validation.

The test doesn't do *the real thing* yet as there aren't multiple processes involved. But other than that, the setup isn't fake and the test works.

I'd like to turn this into something that can be enabled via feature flag (e.g. `--dedicated-chain-process`) so that we could essentially A/B test trinity and look at actual numbers.

What's kinda cool about this approach of using the event bus to delegate to the actual processor is that it enables loose coupling. Anyone can write a plugin that listens for these events and take whatever approach they want to boost up the processing. E.g. you could write a plugin that uses https://dask.org to scale out to many processes (or whatever other magic it uses)

cburgdorf:
Sweet! I wired it up and here's the `BlockImportHandler` running in an isolated plugin, in it's own process, doing the costly block importing part to keep the `networking` process free of that work.

Again, this is essentially what happens in *current* master as well through the usage of the `ChainProxy` but what changes once we merge #1299. 

**BUT**, here's another thing: With the `ChainProxy` we *implicitly* route every chain access to another process (the db process under that model).

With this event bus based solution, we **explicitly** decide which APIs we want to offload to an external process and which are better consumed locally (after #1299, we use a local chain). This may actually turn out as a huge advantage because we might be also calling APIs that are better called locally as they are too inexpensive to pay the extra IO penalty of shifting them to another process.

```
INFO  11-22 17:52:10    RegularChainSyncer  Imported block 271 (0 txs) in 0.07 seconds
   DEBUG  11-22 17:52:10    BlockImportHandler  Importing block: Block #272
   DEBUG  11-22 17:52:10            FrontierVM  BLOCK REWARD: 5000000000000000000 -> b'\xf8\xff\xfd\xcbk\x8b\x887+j\xe4"W\xf6m\x93\xe5\x18\x9a\xc9'
   DEBUG  11-22 17:52:10                 Chain  IMPORTED_BLOCK: number 272 | hash 0x8e9a27332450fcdc845e6e20c4a6bd2d2ffc13cf096d4fdd4090ea16b47add19
    INFO  11-22 17:52:10    RegularChainSyncer  Imported block 272 (0 txs) in 0.07 seconds
   DEBUG  11-22 17:52:10    BlockImportHandler  Importing block: Block #273
   DEBUG  11-22 17:52:10            FrontierVM  BLOCK REWARD: 5000000000000000000 -> b'(\x92\x1eN,\x9d\x84\xf4\xc0\xf0\xc0\xce\xb9\x91\xf4WQ\xa0\xfe\x93'
   DEBUG  11-22 17:52:10                 Chain  IMPORTED_BLOCK: number 273 | hash 0x98dfac36f6e27ef7c538b32e97af049bff79179f20a077dd368be2e76766086d
```

pipermerriam:

pipermerriam:
Ping @cburgdorf to migrate this to trinity repo.

cburgdorf:
I'm going to close this for now. It was only meant as a PoC.

PR #1498, opened 2018-11-21
Prototype parallelization of normalization for peer responses.

pipermerriam:
POC for #1495 

### What was wrong?

Some basic measurement suggest that batches block bodies can take a long time to go through the normalization step (in the realm of 3-6 seconds).

### How was it fixed?

Very crude adjustment of how *slow* normalizers function to allow them to be parallelized.

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/common/managers.py
+++ b/trinity/protocol/common/managers.py
@@ -16,6 +16,7 @@ from cancel_token import CancelToken
 from eth_utils import (
     ValidationError,
 )
+from eth_utils.toolz import concat
 
 from p2p.exceptions import PeerConnectionLost
 from p2p.peer import BasePeer, PeerSubscriber
@@ -27,6 +28,7 @@ from p2p.protocol import (
 from p2p.service import BaseService
 
 from trinity.exceptions import AlreadyWaiting
+from trinity.utils.timer import Timer
 
 from .constants import ROUND_TRIP_TIMEOUT
 from .normalizers import BaseNormalizer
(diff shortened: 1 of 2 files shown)


pipermerriam:
This fix reduces the upper bound on the runtime for these to be closer to 1-2 seconds, though it doesn't appear to speed up the rate at which we can ingest blocks by any significant amount.

carver:
> it doesn't appear to speed up the rate at which we can ingest blocks by any significant amount.

This doesn't surprise me a ton. We are already processing in parallel by requesting blocks from multiple peers. Doubly parallel shouldn't have a meaningful impact on total throughput (maybe even making it worse?), unless we're connected to many fewer peers than there are cores.

If normalization has become the bottleneck, then we probably have to take other steps, like directly optimizing the normalization steps.

pipermerriam:
@carver it appears that I'm able to reduce the overall wait time for normalization which I believe just means that this does a better job of taking advantage of all of the worker processes, but since that didn't increase overall blocks-per-second throughput I believe that it means our primary bottleneck is elsewhere.

Some things to note are that my machine has 12 cores and thus more worker processes than a lower end machine.  Assuming we went this route I think we'd want something that monitored the *rate* that batches are being processed and adjusted the batch size dynamically as the optimal batch size will change based on both 1) number of peers we are requesting data from 2) number of cores on the machine.

I briefly tinkered with adjusting the size of the normalization batches dynamically.  After some futzing with it, I think I have some better intuition about how such a thing would work, but it got more complicated than I was willing to invest  the time to figure out.  Basically, if you think of a range of batch sizes `[1, n]` that we use to batch normalization across multiple cores, then the performance of a given batch size should roughly resemble a parabola where:

- Too high a batch size doesn't effectively use all of the cores and thus is slow.
- Too low a batch size **either** incurs too high an overhead cost *or* the cores are too busy and so execution has to wait for workers, and thus it is slow.

With this concept we should be able to dynamically adjust the batch size such that it finds the optimal place on that curve (which changes based on how many peers are connected and how heavy the load is on the machine).

Nothing actionable here, just wanted to dump some thoughts.  Given that this doesn't improve actual sync performance I'm going to close this without merging and we can come back to it later once we've figured out where the other bottlenecks are.

PR #1500, opened 2018-11-22
Speed up receipt validation.

pipermerriam:
### What was wrong?

A receipt with many logs from the same address or with the same topics will end up checking each duplicate against the bloom filter, resulting in many extra `keccak` calls.

### How was it fixed

Keep track of the values that have already been checked and skip any subsequent duplicates.


pipermerriam:
yellow tests are green but got rate limited.

Issue #1480, opened 2018-11-19
Add command line flag for doing a full "regular" sync


pipermerriam:
### What is wrong?

It should be possible to perform a "regular" sync. Meaning sequential download of blocks and sequential application of those blocks to build the chain state.

### How can it be fixed

First, we need to fix an internal naming issue where we refer to our *fast* sync using `SYNC_FULL`.  This should be renamed to `SYNC_FAST`, and use the *fast* terminology in place of *full* everywhere that is currently used.

Then, add a new option for `SYNC_FULL` which just does a regular block-by-block sync.  The cli should support the following:

- `--sync-mode=light/full/fast` with `fast` being the default.
- `--light` as a shortcut for `--sync-mode=light` (already supported)
- `--fast` as a shortcut for `--sync-mode=fast`
- `--full` as a shortcut for `--sync-mode=full`

The `trinity.sync` module will need to be re-organized as well, moving `trinity.sync.full` to `trinity.sync.fast` and adding the new syncer under `trinity.sync.full`.

carver:
Note that the "first" and "then" steps above should probably be separate PRs to make sure all tests pass after the first step.

Hm, geth used to have those shortcuts (eg~ `--fast`) and seems to have dropped them since. We should consider not adding them in the first place.

cburgdorf:
Am I right to assume that we still need to ensure we have the entire state for the current head before we start applying blocks with their transactions (how should it work without that :thinking:?)

pipermerriam:
> how should it work without that

Yes, regular sync should likely do a validate step upon startup (and honestly maybe before applying **any** block to check if the parent block's state root is present in the database so that we can throw a user friendly error in the event that it is not.

cburgdorf:
Cool, I feel like I'm starting to how this called Ethiriuum works.

carver:
It surely already raises an error if the state is missing (_he speculated with out finding it in the code_), but catching it and raising a friendlier message could be reasonable.

glaksmono:
Would like to take a stab on this

cburgdorf:
>but catching it and raising a friendlier message could be reasonable.

I guess we should even go one step further and hint the user that they could recreate the state using something like `--rebuild-state`.

When started with this command, trinity would take the genesis state and then apply all the blocks that it already has until it has build the state up to where its `HEAD` is. Then, from there, the user could run `trinity --full` and get into regular syncing. Does that make sense?

cburgdorf:
[...]

PR #1501, opened 2018-11-22
Fix sync mode naming: change all of the SYNC_FULL to SYNC_FAST

glaksmono:
### What was wrong?

We need to fix an internal naming issue where we refer to our fast sync using `SYNC_FULL`. This should be renamed to `SYNC_FAST`, and use the fast terminology in place of full everywhere that is currently used.

Issue Reference: https://github.com/ethereum/py-evm/issues/1480

### How was it fixed?

Change the name from `SYNC_FULL` to `SYNC_FAST`

#### Cute Animal Picture

### DIFF
--- a/scripts/db-shell.py
+++ b/scripts/db-shell.py
@@ -16,6 +16,7 @@ from trinity.config import TrinityConfig
 from trinity.constants import (
     MAINNET_NETWORK_ID,
     ROPSTEN_NETWORK_ID,
+    SYNC_FAST,
     SYNC_FULL,
     SYNC_LIGHT,
 )
@@ -30,9 +31,13 @@ if __name__ == '__main__':
     network_id = MAINNET_NETWORK_ID
     if args.ropsten:
         network_id = ROPSTEN_NETWORK_ID
-    sync_mode = SYNC_FULL
-    if args.light:
+
+    if args.full:
+        sync_mode = SYNC_FULL
+    elif args.light:
         sync_mode = SYNC_LIGHT
+    else:
+        sync_mode = SYNC_FAST
 
     cfg = TrinityConfig(network_id, sync_mode=sync_mode)
     chaindb = ChainDB(LevelDB(cfg.database_dir))
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -33,6 +33,7 @@ from trinity.constants import (
     DEFAULT_PREFERRED_NODES,
     MAINNET_NETWORK_ID,
     ROPSTEN_NETWORK_ID,
+    SYNC_FAST,
     SYNC_FULL,
     SYNC_LIGHT,
 )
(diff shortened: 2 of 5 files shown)


scripts/db-shell.py
```diff
@@ -30,7 +31,9 @@
     network_id = MAINNET_NETWORK_ID
     if args.ropsten:
         network_id = ROPSTEN_NETWORK_ID
-    sync_mode = SYNC_FULL
+    sync_mode = SYNC_FAST
+    if args.full:
+        sync_mode = SYNC_FULL
```
carver: These are three exclusive options, maybe `if/elif/else fast`.

trinity/config.py
```diff
@@ -176,7 +177,9 @@ def database_dir(self) -> Path:
 
         This is resolved relative to the ``data_dir``
         """
-        if self.sync_mode == SYNC_FULL:
+        if self.sync_mode == SYNC_FAST:
+            return self.data_dir / DATABASE_DIR_NAME / "fast"
```
carver: Hm, this is unfortunate. It will cause everyone not on a light client to re-sync from scratch :-/

I'm not sure that we even need a separate DB dir for fast vs full, but maybe that's just me not wanting to trigger a resync for everyone (without manual intervention of moving their data directory).

Counter-argument: it's alpha software, suck it up testers :P

I guess I land on: we keep this change, but add a big warning to users in the next release notes saying that they should move their data dir if they don't want to resync.

trinity/config.py
```diff
@@ -272,7 +275,7 @@ def node_class(self) -> Type['Node']:
                 raise NotImplementedError(
                     "Only the mainnet and ropsten chains are currently supported"
                 )
-        elif self.sync_mode == SYNC_FULL:
+        elif self.sync_mode == SYNC_FULL or self.sync_mode == SYNC_FAST:
```
carver: If you like: `elif self.sync_mode in {SYNC_FULL, SYNC_FAST}:`

trinity/config.py
```diff
@@ -283,7 +286,7 @@ def node_class(self) -> Type['Node']:
                 )
         else:
             raise NotImplementedError(
-                "Only full and light sync modes are supported"
+                self.sync_mode + " sync mode is not supported"
```
carver: We prefer "f-strings" in trinity for everything but `logging`. So:

```py
raise NotImplementedError(f"{self.sync_mode} sync mode is not supported")
```

glaksmono:
Not really sure why I'm seeing these on CircleCI:

```
Build-agent version 0.1.1073-1f69f340 (2018-11-20T18:07:03+0000)
Starting container circleci/python:3.6
  image is cached as circleci/python:3.6, but refreshing...

Error response from daemon: unauthorized: authentication required
```

@carver @pipermerriam

glaksmono:
Moving this PR to https://github.com/ethereum/py-evm/pull/1513

PR #1511, opened 2018-11-23
New SeededAtomicDB: in-memory atomic, backed by read-only DB

carver:
### What was wrong?

I wanted to load a LevelDB database from disk into an in-memory Chain DB, and guarantee that it could not be modified. It is relatively straightforward to implement, but I wanted to test that everything worked as expected, and make it available for re-use.

(This was inspired by working on an integration test for #1392 which includes ~1k blocks that should be loaded from disk, but not modified)

### How was it fixed?

Added a `SeededAtomicDB`, which is a tiny deviation from `AtomicDB`: all changes go into a `BatchDB` which is never committed. So it looks as if all changes are made, but the underlying is never changed.

I'll include an example case in a new PR that updates the `tests/trinity/core/p2p-proto/test_sync.py` to use a prebuilt chain with the `SeededAtomicDB` (lightchain integration could use the same tool). This will shave a few seconds off of the sync tests (and make it feasible to use ~1k real headers with PoW seals in a header skeleton test).

Thoughts on a better name? Some alternatives, plus things I don't like about them:
- `PreloadedAtomicDB` (close 2nd to Seeded, I could be convinced)
- `ReadThroughAtomicDB` (wordy)
- `EphemeralDB` (not clear from the name how it's different from MemoryDB)
- `DiskReadMemoryWriteDB` (the DB doesn't actually enforce the read db is on disk)
- `LockedAtomicDB` (locked implies you can't write, but you can)

#### Cute Animal Picture

### DIFF
--- a/eth/db/atomic.py
+++ b/eth/db/atomic.py
@@ -16,6 +16,7 @@ from eth.db.diff import (
 )
 from eth.db.backends.base import BaseDB, BaseAtomicDB
 from eth.db.backends.memory import MemoryDB
+from eth.db.batch import BatchDB
 
 
 class AtomicDB(BaseAtomicDB):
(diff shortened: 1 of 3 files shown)


pipermerriam:
See my comment in #1512 about questioning the necessity of this.

carver:
Closing in favor of #1523

PR #1512, opened 2018-11-23
Use seeded atomic db for sync tests

carver:
### What was wrong?

We were re-calculating the proof-of-work for 20 headers, three times over, when running `p2p-proto/test_sync.py` tests. This was a waste, and extending it to more headers would have been impractical.

### How was it fixed?

- Pre-generate the twenty headers used for sync test and save to disk.
- Upgraded from `Frontier` to `Byzantium` in sync test
- Load headers disk into an in-memory DB (by building on top of #1511) -- Any comments about `SeededAtomicDB` can go there.
- Added the script used to pregenerate the headers.

The only change unique to this PR is c754f22

This change paves the way for testing #1392 with ~1k headers.

#### Cute Animal Picture


tests/trinity/core/integration_test_helpers.py
```diff
@@ -96,3 +106,52 @@ class FakeAsyncChain(MiningChain):
...
+    GENESIS_STATE = {
+        FUNDED_ACCT.public_key.to_canonical_address(): {
+            "balance": 100000000000000000,
+            "code": b"",
+            "nonce": 0,
+            "storage": {}
+        }
+    }
+
+    return PoWMiningChain.from_genesis(db, GENESIS_PARAMS, GENESIS_STATE)
```
pipermerriam: This whole function seems like it should use the chain builder API.  Any reason you didn't?  I suspect it'd be significantly shorter and still easy to read and see what it is doing.

tests/trinity/core/p2p-proto/test_sync.py
```diff
@@ -157,72 +155,33 @@ def finalizer():
...
-            gas_price=1234,
-            gas=1234000,
-            to=RECEIVER.public_key.to_canonical_address(),
-            value=i,
-            data=b'',
-        )
-        chain.apply_transaction(tx.as_signed_transaction(SENDER))
-        chain.mine_block()
+def leveldb_20(tmpdir):
+    fresh_path = tmpdir.join('20-headers.ldb')
```
pipermerriam: Somewhat trivial but the stdlib versions have context manager APIs for temporary directories have context manager APIs which do cleanup for you.

tests/trinity/integration/fixtures/20pow_headers.ldb/CURRENT
```diff
@@ -0,0 +1 @@
+MANIFEST-000027
```
pipermerriam: What do you think about zipping this directory and to add unzipping it as part of the CI setup.  We should only have to do the unzip once as a `session` scoped fixture, and then we can do the directory copy from there. 

- It cleans up the git history a bit by not having each of the leveldb files laying around
- It prevents us *accidentally* connecting to and modifying the canonical database.

carver:
> Since we do a fully copy of the leveldb database into a temporary directory I don't see why we need an extra layer to keep us from modifying it.

Yeah, I think are right. I added the directory copy at the last minute, because I forgot that LevelDB makes file changes on disk even if you only ever read from it, and the dirty git tree was bothering me. I'll take a crack at replacing this with a directory-copy-only approach.

tests/trinity/core/integration_test_helpers.py
carver: It was a fairly dumb copy from the old implementation. I'll try out the chain builder API.

carver:
Closing in favor of #1523

PR #1513, opened 2018-11-24
Fix sync mode naming: change all of the SYNC_FULL to SYNC_FAST

glaksmono:
### What was wrong?

We need to fix an internal naming issue where we refer to our fast sync using `SYNC_FULL`. This should be renamed to `SYNC_FAST`, and use the fast terminology in place of full everywhere that is currently used.

Issue Reference: https://github.com/ethereum/py-evm/issues/1480

### How was it fixed?

Change the name from `SYNC_FULL` to `SYNC_FAST`

#### Cute Animal Picture

### DIFF
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -34,6 +34,7 @@ from trinity.constants import (
     MAINNET_NETWORK_ID,
     ROPSTEN_NETWORK_ID,
     SYNC_FULL,
+    SYNC_FAST,
     SYNC_LIGHT,
 )
 from trinity.utils.chains import (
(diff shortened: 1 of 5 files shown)


trinity/config.py
```diff
@@ -457,7 +464,7 @@ def node_class(self) -> Type['Node']:
         from trinity.nodes.full import FullNode
         from trinity.nodes.light import LightNode
 
-        if self.is_full_mode:
+        if self.is_full_mode or self.is_fast_mode:
```
glaksmono: @pipermerriam currently I'm making `SYNC_FULL` and `SYNC_FAST` to use the same implementation here. I believe the current `FULL` implementation (i.e. `from trinity.nodes.full import FullNode`) will all need to be renamed to `FAST` (i.e. `from trinity.nodes.fast import FastNode`)?

cburgdorf: >So in the case of `Fast` syncing strategy, which node types that it should be using?

The node type remains being `FullNode`. So in terms of node types there are only `FullNode` and `LightNode` so far.

Issue #62, opened 2017-08-02
Modify *trace* logging format to conform to spec


pipermerriam:
### What is wrong?

There is a spec of some form which the data generated by *tracing* a transaction execution generates.  Py-EVM needs to conform to that.

### How can it be fixed

Talk with @cdetrio and figure out where this format is documented if anywhere and then make the appropriate changes.

cdetrio:
Here's some notes on a standard trace format: https://github.com/ethereum/tests/issues/249

rayrapetyan:
@pipermerriam, @cdetrio could you guide here - is there something else beyond extending current logging message in "apply_computation" function?

pipermerriam:
@rayrapetyan the `logging.trace` statements and this are only loosely related.  Currently, all of the `TRACE` and `logging.trace` level logging is simply normal python logging.

This issue is about the ability to generate debug information about EVM execution.  See https://github.com/ethereum/go-ethereum/wiki/Management-APIs#debug_traceblock for the format that go-ethereum uses.

I had originally hoped that we could facilitate this via the python standard library `logging` APIs.  The idea was use something like [this API on the `logging.Filter` class](https://docs.python.org/3.5/howto/logging-cookbook.html#filters-contextual) to impart the contextual information about the EVM execution state (stack, memory, gas, etc).  Then we could use a special `logging.Handler` to extract the relevant logging statements and extract the data that way.

I'm not set on this implementation plan, just on having some way to get EVM trace data from transaction and block execution.

rayrapetyan:
So this issue about implementing non-standard debug_* APIs and expose them via JSON-RPC, correct? If yes - is rpc/modules/debug.py a good place to start coding from? Or a more generic rpc/modules/manage.py is better?

pipermerriam:
I think this needs to be implemented in the `eth` module and and then exposed via something like `rpc/modules/debug.py`.

rayrapetyan:
In go implementation, debug_traceBlock is based on debug_traceTransaction calls, which supports pretty advanced options, e.g. optional JavaScript-based transaction tracing. Do we need to support this as well?

pipermerriam:
@rayrapetyan no, for now I think it's adequate to start with just the full trace data as documented in the `debug_traceTransaction` endpoint.  Things like disabling stack/storage/etc are probably features we can push to a subsequent PR.

rayrapetyan:
[...]

PR #1515, opened 2018-11-24
debug_traceTransaction endpoint implementation

rayrapetyan:
### What was wrong?
There was no way to trace transactions in py-evm

### How was it fixed?
A basic tracing support was added.

### Description
New functionality is based on a go-ethereum's debug_traceTransaction logic (https://github.com/ethereum/go-ethereum/wiki/Tracing:-Introduction). Only basic parts were ported so far (e.g. no JS tracers support). Closes issue #62.

List of open questions to discuss:
- [x] Should we drop output of computation.logger.trace as it mostly duplicates a new tracer's output?
- [x] Find a better way to turn tracer on\off (right now it's a member of VM's state class and we just check if it's "not None" from a computation instance). Another way could be passing Tracer as param all the way down in chain of calls: VM.state.apply_transaction() -> execute_transaction() -> TransactionExecutor.build_computation() -> computation.apply_computation().
- [x] Cross-compare output with go-ethereum tests. Problem is I can't find any tests related to non-JS (basic) tracer there. I've performed cross-comparison of multiple transactions and output is identical except few bugs in geth: e.g. sometimes it initializes memory with zero after push into stack, or do not populate error field in case error fired etc).
- [x] "gas" field in response represents a total block gas usage. I believe this is a bug in geth implementation and for traceTransaction should be equal to the gas used purely by a traced transaction, not the whole block.
- [x] Putting main "capture_state" into "finally" block allows us to catch both error and success cases in one place. Also it's the only way I've found to store a valid "cost" amount. In the future we shall move "cost" estimation calculations into Opcode's "gas_cost" method (see "operation's gasCost gasFunc" implementation in geth: https://github.com/ethereum/go-ethereum/blob/master/core/vm/jump_table.go).


eth/vm/computation.py
```diff
@@ -546,25 +565,51 @@ def apply_computation(cls,
...
                 )
 
+                if computation.state.tracer is not None:
+                    """
+                    TODO: once we are able to predict gas_used, we can capture_state here,
+                          no dumping of stack and memory required
+                    """
+                    gas_remaining_prev = computation._gas_meter.gas_remaining
+                    stack_prev = deepcopy(computation._stack)
+                    memory_prev = deepcopy(computation._memory)
```
rayrapetyan: Once we can predict gas usage beforehand, we can avoid storing computation state before execution.

eth/tools/tracing.py
```diff
@@ -0,0 +1,187 @@
...
+    (tx_block_num, tx_ix) = chain.chaindb.get_transaction_index(decode_hex(tx_hash))
+    tx_block = chain.get_canonical_block_by_number(tx_block_num)
+    parent_block = chain.get_canonical_block_by_number(tx_block_num - 1)
+    parent_header = chain.ensure_header(parent_block.header)  # type: ignore
+    vm = chain.get_vm(parent_header)
+    vm.apply_all_transactions(transactions=tx_block.transactions[0:tx_ix],
+                              base_header=parent_header)
+    vm.state.tracer = StructLogger(conf)
+    _, receipt, comp = vm.apply_transaction(vm.block.header, tx_block.transactions[tx_ix])
+    return ExecutionResult(gas=receipt.gas_used,
```
rayrapetyan: Here, receipt.gas_used is calculated as:
```
gas_used = base_header.gas_used + tx_gas_used
```
in make_frontier_receipt(). Shouldn't it be tx_gas_used as we were tracing only one transaction in a block?

eth/tools/tracing.py
```diff
@@ -0,0 +1,187 @@
...
+    from eth.vm.computation import BaseComputation  # noqa: F401
+    from trinity.chains.base import BaseAsyncChain  # noqa: F401
+
+
+class TraceConfig(NamedTuple):
+    debug: bool  # print output during capture end
+    disable_memory: bool  # disable memory capture
+    disable_stack: bool  # disable stack capture
+    disable_storage: bool  # disable storage capture
+    limit: int  # maximum length of output, but zero means unlimited
```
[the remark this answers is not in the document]
rayrapetyan: Right, their docs are outdated, see actual supported options in LogConfig definition: https://github.com/ethereum/go-ethereum/blob/master/core/vm/logger.go#L46. Zero was copied from there, but we can use None.

rayrapetyan:
Thanks @pipermerriam. Do you want me to merge your changes into my branch or should we drop mine completely? Thanks.

rayrapetyan:
Thanks @pipermerriam, I'll continue here using your changes.

pipermerriam:
@rayrapetyan :+1: let me know if you need anything from me.

eth/tools/tracing.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class StructLogEntry(NamedTuple):
+    depth: int
```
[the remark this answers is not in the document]
rayrapetyan: @pipermerriam, btw any reason to maintain Python 3.5 in this project? Thanks.

eth/vm/computation.py
```diff
@@ -125,7 +106,8 @@ class BaseComputation(Configurable, ABC):
     def __init__(self,
                  state: BaseState,
                  message: Message,
-                 transaction_context: BaseTransactionContext) -> None:
+                 transaction_context: BaseTransactionContext,
+                 tracer: BaseTracer = None) -> None:
```
pipermerriam: I've been struggling with whether the `tracer` should be a required parameter and I'm leaning towards yes.  My thought is that I already had a situation where my tests were failing because I *thought* I was using a real tracer but a code path hadn't been updated properly so it was defaulting to the `NoopTracer`.  By making this a required argument we can't make that mistake.

rayrapetyan: @pipermerriam, I've made it required. As per tests - do you think keeping expected JSON-trace and then simply comparing it with a result will be a good test for that?

p.s. py35-lint went crazy lol - it blames to one line which passes checks in master... 
It behaves really strange, e.g. we have a lot of formatted strings like f"...", which "...are only supported in Python 3.6 and greater", how does master passes py35-lint checks then?

pipermerriam: The `eth` module still has to support python 3.5 so we don't use f-strings there.

rayrapetyan: @pipermerriam, so why py35-lint is launching on a trinity module then? It fails on:
```__version__: str``` at the top of __init__.py.
It somehow passes same on master branch...

Comment (pipermerriam) [reply eth/vm/computation.py]: Ah! `py35-lint` should **not** be run on the trinity module.  That is an error/mistake.

Comment (rayrapetyan) [reply eth/vm/computation.py]: @pipermerriam. Weird, I don't see any diff for tox.ini in master\tracer branches, but somehow master passes all tests, and tracer fails one py35-lint test here.

**Comment by pipermerriam** [comment]
I took some additional liberties here: https://github.com/rayrapetyan/py-evm/pull/2

File: eth/vm/tracing.py [hunk 6]
```diff
@@ -0,0 +1,246 @@
...
+)
+
+from eth.vm.opcode import (
+    Opcode,
+)
+
+if TYPE_CHECKING:
+    # avoid circular import
+    from eth.vm.computation import BaseComputation  # noqa: F401
+    from trinity.chains.base import BaseAsyncChain   # noqa: F401
```
carver: This is why the lint is failing: the code imports the `trinity` module from inside `eth`. Not sure the right solution here. I can't think of a reason off-hand why the `BaseAsyncChain` couldn't live in `eth` but I don't know why it was moved out in the first place...

eth/vm/tracing.py
```diff
@@ -0,0 +1,246 @@
...
+            gas_cost=gas_cost,
+            memory=memory,
+            op=op,
+            pc=pc,
+            stack=stack,
+            storage=storage,
+        ))
+
+
+def trace_transaction(chain: 'BaseAsyncChain', tx_hash: Hash32, memory: bool, stack: bool,
```
pipermerriam: This utility should just be moved into the `trinity` module somewhere near the `debug_traceTransaction` implementation.

eth/vm/tracing.py
pipermerriam: See my comment below about moving the utility (and thus not needing this import)

rayrapetyan:
@pipermerriam, awaiting your responses in https://github.com/rayrapetyan/py-evm/pull/2. Thanks.

rayrapetyan:
@pipermerriam, please review and merge.

carver:
This commit history is pretty messy. Are you up for doing rebase/squash and force push?

Also, best to remove `[WIP] ` from the PR title if you're ready for the final review before merging. That is one signal we use for readiness.

rayrapetyan:
@carver, removed WIP, please review and if everything is OK I'll squash. Thanks.

eth/vm/computation.py
```diff
@@ -248,7 +277,7 @@ def memory_write(self, start_position: int, size: int, value: bytes) -> None:
 
     def memory_read(self, start_position: int, size: int) -> memoryview:
         """
-        Read and return a view of ``size`` bytes from memory starting at ``start_position``.
+        Read and return ``size`` bytes from memory starting at ``start_position``.
```
carver:
```suggestion
        Read and return a view of ``size`` bytes from memory starting at ``start_position``.
```

This looks like a mistaken revert.

eth/vm/computation.py
```diff
@@ -258,6 +287,9 @@ def memory_read_bytes(self, start_position: int, size: int) -> bytes:
         """
         return self._memory.read_bytes(start_position, size)
 
+    def dump_memory(self) -> bytes:
+        return bytearray(self._memory)
```
carver: Maybe use the new `read_bytes` like:

```suggestion
        return self._memory.read_bytes()
```

eth/vm/computation.py
```diff
@@ -402,12 +437,14 @@ def generate_child_computation(self, child_msg: Message) -> 'BaseComputation':
                 self.state,
                 child_msg,
                 self.transaction_context,
+                self.tracer
```
carver: *nitpick* trailing commas are nice, because adding a new argument only diffs one line instead of two. (like here, line 439 didn't have to change).

eth/vm/computation.py
```diff
@@ -402,12 +437,14 @@ def generate_child_computation(self, child_msg: Message) -> 'BaseComputation':
...
             ).apply_create_message()
         else:
             child_computation = self.__class__(
                 self.state,
                 child_msg,
                 self.transaction_context,
+                self.tracer
```
carver: *nitpick* trailing commas are nice

eth/vm/memory.py
```diff
@@ -23,6 +26,9 @@ class Memory(object):
     def __init__(self) -> None:
         self._bytes = bytearray()
 
+    def __iter__(self) -> Iterator[int]:
+        return iter(self._bytes)
```
carver: This can be dropped if you use the new `read_bytes()`

eth/vm/stack.py
```diff
@@ -35,6 +36,18 @@ def __init__(self) -> None:
...
+    def __iter__(self) -> Iterable[int]:
+        for item in self.values:
+            if isinstance(item, int):
+                yield item
+            elif isinstance(item, bytes):
+                yield big_endian_to_int(item)
+            else:
+                raise TypeError(
+                    "Invariant: stack should only contain `int` and `bytes` "
+                    "types.  Got: {0}".format(type(item))
```
carver: *nitpick* it can be nice to see the value in addition to the type. One way to do that would be something like:

```suggestion
                    "types.  Got: {!r}".format(item)
```

which usually is enough to determine both.

pipermerriam:
@carver I'm a little too close to this issue (as I did some significant rewriting along the way) so I'm good delegating to you for final merge.

trinity/rpc/modules/debug.py
```diff
@@ -0,0 +1,158 @@
...
+                                  pc=e.pc,
+                                  stack=[],
+                                  storage={})
+        if e.err:
+            entry['error'] = str(e.err)
+
+        i = 0
+        while i + 32 <= len(e.memory):
+            entry['memory'].append(remove_0x_prefix(encode_hex(e.memory[i: i + 32])))
+            i += 32
```
pipermerriam: This is a section I'd like to cleanup but it doesn't have to be in this PR.  I'm actually :+1: on removing all of the JSON-RPC stuff from this PR and doing it in a subsequent one (and just merging the core tracer API in this PR.

rayrapetyan:
@carver, if possible - please merge on your side. I swear never merge master into dev branches anymore :)

pipermerriam:
@rayrapetyan in case you aren't aware: https://help.github.com/articles/about-git-rebase/

`git rebase master` is what you're looking for but it takes a little bit of a learning curve to deal with merge conflicts properly.

rayrapetyan:
@pipermerriam, I would be glad to merge, but most of the changes there are yours so I can miss\overlook something. As for trinity parts - I guess you mean formatting function, I would leave it as is with a TODO mark. `while i + 32 <= len(e.memory):` part is exactly how it was done in geth. And btw unassign me from bounty - this was primarily your job done here. Thanks.

pipermerriam:
@rayrapetyan I'm fine with you collecting the bounty.

1) you laid the foundation for what I did
2) you still did a significant part of the work
3) I like contributors being paid though I appreciate you willingness to contribute for free.

pipermerriam:
@rayrapetyan it's less about that specific line of code and more 1) wanting to give the whole implementation of the RPC endpoint another once-over and 2) wanting to convert that entire normalization function into something more inline with our code style (functional and immutable).

PR #1520, opened 2018-11-27
test_lightchain_integration: Alternative method to wait for lightchain sync.

cryptomental:
### What was wrong?

Test case test_lightchain_integration uses low level calls to verify that the syncer has synced.

### How was it fixed?

Use LightChainSyncer logger to verify that the sync completed.

### DIFF
--- a/tests/trinity/integration/test_lightchain_integration.py
+++ b/tests/trinity/integration/test_lightchain_integration.py
@@ -1,6 +1,8 @@
 import asyncio
 import logging
 from pathlib import Path
+import os
+import re
 import socket
 import subprocess
 import time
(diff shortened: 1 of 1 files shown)


cryptomental:
Hi @cburgdorf it took me a few hours to experiment with pexpect and the lightchain integration test, it was actually very entertaining and I came up with the approach above. Feel free to approve or decline, I did not find anything like on the web that but this approach may be used for other tests as well.

The reason I did not use pexpect was that pexpect works great when a new  a new process is spawned. I tried to use !(fdexpect)[https://pexpect.readthedocs.io/en/stable/api/fdpexpect.html] to connect to a known logger file descriptor and spawn a new pipe to intercept the log but pytest seems to swallow the output. Then I looked at the problem from a different angle. This test relies on LightChainSyncer which communicates with the outside world using `"Imported %d headers in %0.2f seconds, new head: %s"` logger message. I went down in the logger tree to BaseHeaderChainSyncer => BaseService and noticed the TraceLogger logs go the root logger which is intercepted by pytest during tests execution and exposed in the "caplog" test fixture. I modified the test to wait for 5 seconds until logger message contains "Imported * headers" and the number of headers is compared to the required number. If the message is found, test resumes, otherwise it fails with a timeout.

Issue #1510, opened 2018-11-23
Rework helper functions


hwwhww:
### What is wrong?
Sync with the spec: https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#helper-functions
- [x] `def get_active_validator_indices(validators)`
- [x] `def shuffle(values: List[Any], seed: Hash32) -> List[Any]`
- [x] `def split(seq: List[Any], split_count: int) -> List[Any]`
- [x] `def clamp(minval: int, maxval: int, x: int) -> int`
- [x] `def get_new_shuffling(seed: Hash32, validators: List[ValidatorRecord],  crosslinking_start_shard: int) -> List[List[ShardAndCommittee]]`
- [x] `def get_block_hash(state: BeaconState, current_block: BeaconBlock, slot: int) -> Hash32`
- [x] `def get_beacon_proposer(state:BeaconState, slot: int) -> ValidatorRecord`
- [x] `def add_validator_set_change_record(state: BeaconState, index: int, pubkey: int, flag: int) -> None`
- [x] `def int_sqrt(n: int) -> int`

### How can it be fixed
1. Go through the code and see if anything require updates.
2. Turn on the test (`tests/beacon/test_helpers.py`) or write new tests.

hwwhww:
@djrtwo I'm starting to rework/review the first 4 functions. If you'd like to start with it, leave a comment under this issue then I'll know! :)

hwwhww:
Update: finished the first 5 functions in this branch: https://github.com/hwwhww/py-evm/commits/rework_helper, will start the other 4 functions.

PR #1521, opened 2018-11-28
Rework helper functions - part 1

hwwhww:
### What was wrong?
#1510 - part 1
Some helper functions that are more independent to beacon chain logics.

### How was it fixed?
* `eth/beacon/helpers.py`
    * (refactor) `def get_active_validator_indices(validators: Sequence['ValidatorRecord']) -> Iterable[int]`
    * (refactor) `def get_new_shuffling(*, seed: Hash32, validators: Sequence['ValidatorRecord'], crosslinking_start_shard: int, cycle_length: int, target_committee_size: int, shard_count: int) -> Iterable[Iterable[ShardAndCommittee]])`
* `eth/beacon/utils/random.py`
    * (refactor) `def shuffle(values: Sequence[Any], seed: Hash32) -> Iterable[Any]`
    * (refactor) `def split(seq: Sequence[TItem], split_count: int) -> Iterable[Any]`
* `eth/utils/numeric.py`
    * **(new!)** `def int_sqrt(value: int) -> int`

#### Cute Animal Picture
cat_chicken_alison groves


eth/beacon/utils/random.py
```diff
@@ -17,70 +17,80 @@
...
     Returns the shuffled ``values`` with seed as entropy.
     Mainly for shuffling active validators in-protocol.
 
-    Spec: https://github.com/ethereum/eth2.0-specs/blob/0941d592de7546a428066c0473fd1000a7e3e3af/specs/beacon-chain.md#helper-functions  # noqa: E501
+    Spec: https://github.com/ethereum/eth2.0-specs/blob/70cef14a08de70e7bd0455d75cf380eb69694bfb/specs/core/0_beacon-chain.md#helper-functions  # noqa: E501
     """
     values_count = len(values)
 
-    if values_count > SAMPLE_RANGE:
+    if values_count > RAND_MAX:
```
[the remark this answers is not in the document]
hwwhww: Ahhhh nice catches!!!

eth/beacon/utils/random.py
```diff
@@ -17,70 +17,80 @@
...
     return output
 
 
 @to_tuple
-def split(seq: Sequence[TItem], pieces: int) -> Iterable[Any]:
+def split(seq: Sequence[TItem], split_count: int) -> Iterable[Any]:
     """
-    Returns the split ``seq`` in ``pieces`` pieces in protocol.
-    Spec: https://github.com/ethereum/eth2.0-specs/blob/0941d592de7546a428066c0473fd1000a7e3e3af/specs/beacon-chain.md#helper-functions  # noqa: E501
+    Returns the split ``seq`` in ``split_count`` pieces in protocol.
```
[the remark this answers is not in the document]
hwwhww: It was pointed out by Piper that there are different ways to do some kind of "splitting a list". So here we noted that this `split` function is required and defined in beacon chain protocol.

hwwhww:
@djrtwo The whole Justinized clean-up will be reflected in other PRs.

eth/beacon/utils/random.py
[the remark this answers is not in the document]
pipermerriam: Wondering if `partition` wouldn't be a better term for this.

eth/beacon/helpers.py
```diff
@@ -208,63 +215,30 @@ def get_attestation_indices(crystallized_state: 'CrystallizedState',
...
+    Gets indices of active validators from ``validators``.
     """
-    o = []
-    for index, validator in enumerate(validators):
-        if (validator.start_dynasty <= dynasty and dynasty < validator.end_dynasty):
-            o.append(index)
-    return o
+    return [
+        i for i, v in enumerate(validators)
+        if v.status in [ValidatorStatusCode.ACTIVE, ValidatorStatusCode.PENDING_EXIT]
```
pipermerriam: When doing inclusion checks, you should use a `set`.

eth/beacon/helpers.py
```diff
@@ -208,63 +215,30 @@ def get_attestation_indices(crystallized_state: 'CrystallizedState',
...
-    rotation logic.
-    https://github.com/ethereum/eth2.0-specs/commit/52cf7f943dc99cfd27db9fb2c03c692858e2a789#diff-a08ecec277db4a6ed0b3635cfadc9af1  # noqa: E501
+    Gets indices of active validators from ``validators``.
     """
-    o = []
-    for index, validator in enumerate(validators):
-        if (validator.start_dynasty <= dynasty and dynasty < validator.end_dynasty):
-            o.append(index)
-    return o
+    return [
```
pipermerriam: You should unroll this comprehension to use a `yield` pattern since you're using the `to_tuple` decorator.  Otherwise, you should convert this inner comprehension to be `tuple(i for thing in things)` so that we don't waste the overhead of constructiing the list and then converting to a tuple.

eth/beacon/utils/random.py
```diff
@@ -17,70 +17,82 @@
...
+def split(values: Sequence[TItem], split_count: int) -> Iterable[Any]:
     """
-    Returns the split ``seq`` in ``pieces`` pieces in protocol.
-    Spec: https://github.com/ethereum/eth2.0-specs/blob/0941d592de7546a428066c0473fd1000a7e3e3af/specs/beacon-chain.md#helper-functions  # noqa: E501
+    Returns the split ``values`` in ``split_count`` pieces in protocol.
+    Spec: https://github.com/ethereum/eth2.0-specs/blob/70cef14a08de70e7bd0455d75cf380eb69694bfb/specs/core/0_beacon-chain.md#helper-functions  # noqa: E501
     """
-    list_length = len(seq)
+    list_length = len(values)
     return [
```
pipermerriam: You should unroll this comprehension to use a `yield` pattern since you're using the `to_tuple` decorator.  Otherwise, you should convert this inner comprehension to be `tuple(i for thing in things)` so that we don't waste the overhead of constructiing the list and then converting to a tuple.

eth/utils/numeric.py
```diff
@@ -99,3 +99,29 @@ def clamp(inclusive_lower_bound: int,
...
+            "Value cannot be negative: Got: {0}".format(
+                value,
+            )
+        )
+
+    x = value
+    y = (x + 1) // 2
+    while y < x:
+        x = y
+        y = (x + value // x) // 2
```
pipermerriam: Maybe worth profiling, but I think the body of this function could be replaced by `return int(math.log2(value))` which *should* produce the correct result.

tests/beacon/conftest.py
```diff
@@ -349,3 +363,25 @@ def max_attestation_count():
...
+
+#
+# genesis
+#
+@pytest.fixture
+@to_tuple
+def genesis_validators(init_validator_keys,
+                       init_randao,
+                       deposit_size):
+    return [
```
pipermerriam: Same here, the `to_tuple` decorator should either be dropped in favor of just using a tuple comprehension **or** you should convert the inner body to use a `yield` pattern if you want to keep the `to_tuple` decorator.

tests/beacon/conftest.py
```diff
@@ -349,3 +363,25 @@ def max_attestation_count():
...
+        ValidatorRecord(
+            pubkey=pub,
+            withdrawal_credentials=ZERO_HASH32,
+            randao_commitment=init_randao,
+            randao_skips=0,
+            balance=deposit_size * denoms.gwei,
```
pipermerriam: question:  why is the deposit balance in gwei?  It seems like using `wei` would be the least error prone approach.

tests/core/numeric-utils/test_integer_squareroot.py
```diff
@@ -0,0 +1,30 @@
...
+        (4, 2),
+        (27, 5),
+        (65535, 255),
+        (65536, 256),
+        (18446744073709551615, 4294967295),
+        (1.5, ValueError()),
+        (-1, ValueError()),
+    )
+)
+def test_integer_squareroot(value, expected):
```
pipermerriam: Noting that the `integer_squareroot` function is a good candidate for property based testing since it has a very clearly defined property between input/output `output ** 2 <= input`.

eth/beacon/utils/random.py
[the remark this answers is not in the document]
lithp: `partition` might be confusing, because itertoolz already has a [partition](https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.partition) which does something different; clojure [also uses "partition"](https://clojuredocs.org/clojure.core/partition) to refer to the same operation itertoolz uses it for.

Maybe something like: `partition_into_n_pieces`, where the `split_count` parameter is renamed to `n`?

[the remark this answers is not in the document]
lithp: This probably isn't the best name but nim calls it [distribute](https://nim-lang.org/docs/sequtils.html#distribute%2Cseq%5BT%5D%2CPositive)

tests/beacon/conftest.py
hwwhww: Because we currently use `uint64` (~1.8e+19) for almost all number fields. The `MAX_DEPOSIT` is `32`, plus the validator will get rewards, so we utilize gwei (1e+18) as the unit of balance.

eth/utils/numeric.py
hwwhww: Oh, interesting.
I'm inclined to open an issue to the backlog and keep the algorithm as a reference of the spec for a while.

eth/beacon/helpers.py
hwwhww: So sorry for that I totally forgot to Py-EVM-ize this time(!!!). Inclined to use the second pattern (`tuple(i for thing in things)`) for syncing with the spec.

eth/utils/numeric.py
hwwhww: backlog: #1550

tests/beacon/conftest.py
pipermerriam: So this lets the number fit into a `uint64` where-as if it was denominated in wei it would not?

hwwhww: Correct. Per homogenizing types to int64 (ethereum/eth2.0-specs#36, ethereum/eth2.0-specs#48).

pipermerriam: Also, I assume your previous comment should have been `MIN_DEPOSIT`?

eth/beacon/utils/random.py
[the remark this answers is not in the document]
hwwhww: @lithp yeah that may be better. `divide` sounds fine to me.

PR #1523, opened 2018-11-29
Use pre-built leveldb fixtures in a temporary directory

carver:
### What was wrong?

We were re-calculating the proof-of-work for 20 headers, three times over, when running `p2p-proto/test_sync.py` tests. This was a waste, and extending it to more headers would have been impractical. (copy/paste from #1512)

A solution was attempted in #1512, but it the special DB class in the PR became unnecessary after the temporary directory solution was added in. Also, this new solution avoids accidentally running against the checked-in db, and doesn't check in as many little leveldb files.

### How was it fixed?

- upgraded chain to byzantium
- zipped up the generated leveldb files for 20 headers
- add a loader to unzip into a temp directory for tests
- switch to using chain builder API for more succinct code

#### Cute Animal Picture


tests/trinity/core/integration_test_helpers.py
```diff
@@ -96,3 +109,57 @@ class FakeAsyncChain(MiningChain):
...
+def load_fixture_db(db_fixture, db_class=LevelDB):
+    """
+    Extract the database from the zip file to a temp directory, which has two benefits and one cost:
+    - B1. works with xdist, multiple access to the ldb files at the same time
+    - B2. prevents dirty-ing the git index
+    - C1. slows down test, because of time to extract files
+    """
+    assert isinstance(db_fixture, DBFixture)
+    zip_filename = (db_fixture.value + '.zip')
+    zipped_path = Path(__file__).parent.parent / 'integration' / 'fixtures' / zip_filename
```
pipermerriam: I'd recommend we place the first part of this as a top-level constant in this module.  With it embedded in the function it seems easier to forget to update it if this were to move somewhere else.

tests/trinity/core/integration_test_helpers.py
```diff
@@ -96,3 +109,57 @@ class FakeAsyncChain(MiningChain):
...
+
+def load_fixture_db(db_fixture, db_class=LevelDB):
+    """
+    Extract the database from the zip file to a temp directory, which has two benefits and one cost:
+    - B1. works with xdist, multiple access to the ldb files at the same time
+    - B2. prevents dirty-ing the git index
+    - C1. slows down test, because of time to extract files
+    """
+    assert isinstance(db_fixture, DBFixture)
+    zip_filename = (db_fixture.value + '.zip')
```
pipermerriam: f-string?

tests/trinity/core/integration_test_helpers.py
```diff
@@ -96,3 +109,57 @@ class FakeAsyncChain(MiningChain):
...
+
+    GENESIS_STATE = {
+        FUNDED_ACCT.public_key.to_canonical_address(): {
+            "balance": 100000000000000000,
+        }
+    }
+
+    return build(
+        FakeAsyncChain,
+        fork_at(POWByzantiumVM, 0),
```
pipermerriam: You can do this as:

```python
return build(
  FakeAsyncChain,
  byzantium_at(0),
  enable_pow_mining(),
  ...,
)
```

Minor improvement but removes the need for the local `POWByzantiumVM` class.

carver:
Everything passed, CircleCI rate-limited.

PR #1524, opened 2018-11-29
Support light client header validation during sync

carver:
### What was wrong?

The light client couldn't validate header chains, in sync or async contexts.

Interestingly, it appears that our light sync is not currently running full header validations :(

On the up-side, we will be getting those for "free" in #1392 (which is where this PR was stripped from)

### How was it fixed?

- Move `validate_chain` to a classmethod on the base class
- Inherit `validate_chain` in the light chain
- Add the async call using `async_method("validate_chain")`.

#### Cute Animal Picture

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -315,13 +315,31 @@ class BaseChain(Configurable, ABC):
     def validate_uncles(self, block: BaseBlock) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
-    @abstractmethod
+    @classmethod
     def validate_chain(
-            self,
+            cls,
             parent: BlockHeader,
             chain: Tuple[BlockHeader, ...],
             seal_check_random_sample_rate: int = 1) -> None:
-        raise NotImplementedError("Chain classes must implement this method")
+
+        all_indices = list(range(len(chain)))
+        if seal_check_random_sample_rate == 1:
+            headers_to_check_seal = set(all_indices)
+        else:
+            sample_size = len(all_indices) // seal_check_random_sample_rate
+            headers_to_check_seal = set(random.sample(all_indices, sample_size))
+
+        for i, header in enumerate(chain):
+            if header.parent_hash != parent.hash:
+                raise ValidationError(
+                    "Invalid header chain; {} has parent {}, but expected {}".format(
+                        header, header.parent_hash, parent.hash))
+            vm_class = cls.get_vm_class_for_block_number(header.block_number)
+            if i in headers_to_check_seal:
+                vm_class.validate_header(header, parent, check_seal=True)
+            else:
+                vm_class.validate_header(header, parent, check_seal=False)
+            parent = header
 
 
 class Chain(BaseChain):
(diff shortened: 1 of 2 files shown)


eth/chains/base.py
```diff
@@ -315,13 +315,31 @@ def validate_gaslimit(self, header: BlockHeader) -> None:
...
+    @classmethod
     def validate_chain(
-            self,
+            cls,
             parent: BlockHeader,
             chain: Tuple[BlockHeader, ...],
             seal_check_random_sample_rate: int = 1) -> None:
-        raise NotImplementedError("Chain classes must implement this method")
+
+        all_indices = list(range(len(chain)))
```
pipermerriam: Some brief experimentation suggests that the `list` cast isn't needed here.  `range` objects support all of the operations you do below.

eth/chains/base.py
```diff
@@ -315,13 +315,31 @@ def validate_gaslimit(self, header: BlockHeader) -> None:
...
+        if seal_check_random_sample_rate == 1:
+            headers_to_check_seal = set(all_indices)
+        else:
+            sample_size = len(all_indices) // seal_check_random_sample_rate
+            headers_to_check_seal = set(random.sample(all_indices, sample_size))
+
+        for i, header in enumerate(chain):
```
pipermerriam: I think we can clean this loop up by doing:

```python
for i, (parent, header) in enumerate(sliding_window(2, concatv([parent], chain))):
    ...
```

This gives us pairs of `parent/child` allowing us to eliminate the `parent` tracking.

eth/chains/base.py
```diff
@@ -315,13 +315,31 @@ def validate_gaslimit(self, header: BlockHeader) -> None:
     def validate_uncles(self, block: BaseBlock) -> None:
         raise NotImplementedError("Chain classes must implement this method")
 
-    @abstractmethod
+    @classmethod
     def validate_chain(
-            self,
+            cls,
             parent: BlockHeader,
```
pipermerriam: What do you think about the name `anchor` for this to free up the name `parent` name for the ephemeral loop variable.

eth/chains/base.py
```diff
@@ -315,13 +315,31 @@ def validate_gaslimit(self, header: BlockHeader) -> None:
...
+        for i, header in enumerate(chain):
+            if header.parent_hash != parent.hash:
+                raise ValidationError(
+                    "Invalid header chain; {} has parent {}, but expected {}".format(
+                        header, header.parent_hash, parent.hash))
+            vm_class = cls.get_vm_class_for_block_number(header.block_number)
+            if i in headers_to_check_seal:
+                vm_class.validate_header(header, parent, check_seal=True)
+            else:
+                vm_class.validate_header(header, parent, check_seal=False)
```
pipermerriam: Do you think this is better readability:

```python
should_check_seal = i in headers_to_check_seal
vm_class.validate_header(header, parent, check_seal=should_check_seal)
```

eth/chains/base.py
carver: How about `root` and `descendants` instead of `parent` and `children` for the method args?

PR #1528, opened 2018-11-29
re-organize Computation methods

pipermerriam:
### What was wrong?

The organization of the methods for our `Computation` object wasn't consistent and had room for improvement.

### How was it fixed?

Moved things around to better group them by functionality.

#### Cute Animal Picture

article-2094738-118a758b000005dc-950_634x436

### DIFF
--- a/eth/vm/computation.py
+++ b/eth/vm/computation.py
@@ -149,6 +149,9 @@ class BaseComputation(Configurable, ABC):
         """
         return self.msg.sender == self.transaction_context.origin
 
+    #
+    # Error handling
+    #
     @property
     def is_success(self) -> bool:
         """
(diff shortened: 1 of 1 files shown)


eth/vm/computation.py
```diff
@@ -516,7 +525,7 @@ def __exit__(self, exc_type: None, exc_value: None, traceback: None) -> None:
                 self.msg.value,
                 self.msg.depth,
                 "y" if self.msg.is_static else "n",
-                self.msg.gas - self._gas_meter.gas_remaining,
+                self.get_gas_used(),
```
pipermerriam: This should be the only functionality change.

cburgdorf: So, the previous implementation did not account for `should_burn_gas`? Was that wrong? I can't tell but I guess you know what you're doing :)

pipermerriam: This is a logging message so it was likely logging the wrong value in cases where gas was not burned.

Issue #1526, opened 2018-11-29
rename `TRACE` log level to `DEBUG2`


pipermerriam:
### What is wrong?

In the original design for Py-EVM I imagined using the built-in python logging facilities to do EVM tracing #62 .  Since then the `TRACE` log level has evolved to just mean *lower level debug logs*.

Second, we've ended up needing to tune the logging level of certain *noisy* modules when ideally we'd be able to do this with finer grained logging levels.

### How can it be fixed

Rename the `TRACE` to `DEBUG2` and change the numeric value to be `8`.  The idea here is that we can extend this to `DEBUG3` and onward.

- The `eth.tools.logging.TraceLogger` should be renamed to `EthereumLogger`
- The `eth.tools.setup_trace_logging` should be renamed to `setup_ethereum_logging`
- We should add a call `eth.tools.logging.setup_ethereum_logging` should be updated to call [`logging.setLoggerClass`](https://docs.python.org/3/library/logging.html#logging.setLoggerClass) so that the `logging.getLogger` will always return an instance of our logger.
- The `TraceLogger.trace` method should be renamed to `TraceLogger.debug2`
- Update the Trinity CLI parser component that handles logging configuration.

Next steps would be to create `DEBUG3` and to take our historically *noisy* loggers and set them to use this new lower level logging.

rayrapetyan:
@pipermerriam, please confirm no one is working on it so I can start working on this one.

pipermerriam:
@rayrapetyan :+1: go for it.

veox:
> The idea here is that we can extend this to DEBUG3 and onward.

I'd rather have `TRACE2` than `DEBUG3`. :|

PR #1530, opened 2018-11-29
Rename TRACE log level to DEBUG2

rayrapetyan:
### What was wrong?

See https://github.com/ethereum/py-evm/issues/1526

### How was it fixed?

See https://github.com/ethereum/py-evm/issues/1526

### DIFF
--- a/eth/__init__.py
+++ b/eth/__init__.py
@@ -2,14 +2,14 @@ import pkg_resources
 import sys
 
 from eth.tools.logging import (
-    setup_trace_logging
+    setup_extended_logging
 )
 
 #
 #  Setup TRACE level logging.
 #
 # This needs to be done before the other imports
-setup_trace_logging()
+setup_extended_logging()
 
 from eth.chains import (  # noqa: F401
     Chain,
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -49,7 +49,7 @@ from eth.validation import (
     validate_canonical_address,
 )
 from eth.tools.logging import (
-    TraceLogger
+    ExtendedDebugLogger
 )
 from eth.utils.padding import (
     pad32,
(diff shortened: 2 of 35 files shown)


eth/vm/computation.py
```diff
@@ -463,7 +463,7 @@ def get_gas_remaining(self) -> int:
     # Context Manager API
     #
     def __enter__(self) -> 'BaseComputation':
-        self.logger.trace(
+        self.logger.debug2(
```
cburgdorf: I have to admit that I'm not entirely sold on the naming but if the majority wants it I'm not opposed.

eth/db/account.py
```diff
@@ -49,7 +49,7 @@
     validate_canonical_address,
 )
 from eth.tools.logging import (
-    TraceLogger
+    EthereumLogger
```
cburgdorf: I think I'm even less sold on this naming ;) Basically because there's nothing Ethereum specific with this logger. So I'd rather name this `[Custom|Extended|FancyPants]DebugLogger` or something like that.

rayrapetyan: @pipermerriam - is CustomDebugLogger looking ok?

pipermerriam: I think I like just plain `DebugLogger` or maybe `ExtendedDebugLogger`.  "Custom" is a bit too generic.

eth/__init__.py
```diff
@@ -2,14 +2,14 @@
...
-    setup_trace_logging
+    setup_ethereum_logging
 )
 
 #
 #  Setup TRACE level logging.
 #
 # This needs to be done before the other imports
-setup_trace_logging()
+setup_ethereum_logging()
```
pipermerriam: Along with the rename, maybe this should just be `setup_logging` or `setup_extended_logging`

eth/vm/computation.py
veox: Me, too. Would prefer `trace2` to `debug3` if numbering's the approach to be taken.

pipermerriam:
Rebased locally, waiting for tests.

pipermerriam:
yellow test was github rate limit thing

PR #1533, opened 2018-11-29
Always use a full chain instance in local server

carver:
### What was wrong?

Can't get block body from light peers.

An initialization cycle was introduced by [this line](https://github.com/ethereum/py-evm/commit/75be226ddd2da06066f1c7a3b7fe0e3826e201a9#diff-49f4c2ce3b6081ee05a5a002420f0446R70), which causes `_peer_chain` to be `None` in the `LightDispatchChain`, which prevents the light dispatch chain from making any requests to peers

(Symptom is something like: `None object doesn't have attribute coro_get_block_by_header`)

### How was it fixed?

Always use a full chain instance in the server, even a `LightServer`. It should always be asking for data locally, instead of trying to ask other peers.

#### Cute Animal Picture

### DIFF
--- a/trinity/nodes/base.py
+++ b/trinity/nodes/base.py
@@ -16,6 +16,7 @@ from p2p.service import (
     BaseService,
 )
 
+from trinity.chains.full import FullChain
 from trinity.db.header import (
     AsyncHeaderDB,
 )
@@ -40,6 +41,8 @@ class Node(BaseService):
     Create usable nodes by adding subclasses that define the following
     unset attributes.
     """
+    _full_chain: FullChain = None
+
     def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
         super().__init__()
         self.trinity_config = trinity_config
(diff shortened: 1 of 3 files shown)


carver:
@hoi -- I haven't been able to connect to a light server (even on Ropsten) long enough to test it out live, but so far this branch fixes the specific error message we were looking at today.

carver:
(All tests green, just rate-limited by github)

carver:
> we do in fact still rely on the LightPeerChain?

Right, the `LightPeerChain` is still broadcast here: https://github.com/ethereum/py-evm/blob/f0d2ca6426cc8bd1c4bae0b898020654d79d2e68/trinity/nodes/base.py#L127-L131

The `LightServer` however, is responsible for handling incoming requests from other peers. If we can't respond to a data request, then we should just give them an empty response (instead of trying to dispatch to another peer).

This was happening before the genesis re-org, because it was using the `manager.get_chain()` which was always a full chain.

carver:
Okay @hoi you should be able to rebase on master now for your test

PR #1536, opened 2018-11-30
Add support for trinity-beacon command

cburgdorf:
### What was wrong?

We currently do not have any bootstrapping for the beacon node.

### How was it fixed?

There are several routes we can take to add maintain an entirely different node type in the same code base.

One pragmatic solution is to maintain multiple main files and expose multiple top level commands

E.g. There would not only be `trinity` but also `trinity-beacon` (name to be discussed).

While this comes with the slight disadvantage of occuping an entirely different command, it gives us the most freedom for experimentation.

Also, it should be relatively straight forward to merge the bootstrapping at a later point in time. So this is just me trying to be pragmatic.

Steps:

1. I moved everything from `main.py` to `bootstrap.py` that I consider generic enough to be useful for any type of program that we want to support in this code base. This stuff is mostly dealing with database access (which needs to be tunneled through one process for now), setting up the event bus, plugins and logging

2. I  left the actual `trinity_boot` method to be defined by each different program. This is mainly because current Trinity uses this shared `networking` process to kickoff the actual node logic and it might be that we don't need this shared process at all for the beacon node.

3. I modified the `setup.py` and `setup_trinity.py` files to expose the other entry point as `trinity-beacon`. Currently, this just spins up the db process and then winds it down gracefully again. 

#### Cute Animal Picture

### DIFF
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -1,0 +1,282 @@
+from argparse import ArgumentParser, Namespace
+import logging
+import multiprocessing
+import os
+import signal
+from typing import (
+    Any,
+    Callable,
+    Dict,
+    Iterable,
+    Type,
+)
+
+from lahja import (
+    EventBus,
+    Endpoint,
+)
+
+from eth.db.backends.base import BaseDB
+
+from trinity.db.manager import get_chaindb_manager
+from trinity.exceptions import (
+    AmbigiousFileSystem,
+    MissingPath,
+)
+from trinity.initialization import (
+    initialize_data_dir,
+    is_data_dir_initialized,
+)
+from trinity.cli_parser import (
+    parser,
+    subparser,
+)
+from trinity.config import (
+    TrinityConfig,
+)
+from trinity.constants import (
+    MAINNET_NETWORK_ID,
+    MAIN_EVENTBUS_ENDPOINT,
+    ROPSTEN_NETWORK_ID,
+)
+from trinity.extensibility import (
+    BaseManagerProcessScope,
+    MainAndIsolatedProcessScope,
+    PluginManager,
+)
+from trinity.plugins.registry import (
+    get_all_plugins,
+)
+from trinity.utils.ipc import (
+    kill_process_gracefully,
+)
+from trinity.utils.logging import (
+    enable_warnings_by_default,
+    setup_log_levels,
+    setup_trinity_stderr_logging,
+    setup_trinity_file_and_queue_logging,
+    with_queued_logging,
+)
+from trinity.utils.mp import (
+    ctx,
+)
+from trinity.utils.profiling import (
+    setup_cprofiler,
+)
+from trinity.utils.version import (
+    construct_trinity_client_identifier,
+    is_prerelease,
+)
+
+
+PRECONFIGURED_NETWORKS = {MAINNET_NETWORK_ID, ROPSTEN_NETWORK_ID}
+
+
+TRINITY_HEADER = "\n".join((
+    "\n"
+    r"      ______     _       _ __       ",
+    r"     /_  __/____(_)___  (_) /___  __",
+    r"      / / / ___/ / __ \/ / __/ / / /",
+    r"     / / / /  / / / / / / /_/ /_/ / ",
+    r"    /_/ /_/  /_/_/ /_/_/\__/\__, /  ",
+    r"                           /____/   ",
+))
+
+TRINITY_AMBIGIOUS_FILESYSTEM_INFO = (
+    "Could not initialize data directory\n\n"
+    "   One of these conditions must be met:\n"
+    "   * HOME environment variable set\n"
+    "   * XDG_TRINITY_ROOT environment variable set\n"
+    "   * TRINITY_DATA_DIR environment variable set\n"
+    "   * --data-dir command line argument is passed\n"
+    "\n"
+    "   In case the data directory is outside of the trinity root directory\n"
+    "   Make sure all paths are pre-initialized as Trinity won't attempt\n"
+    "   to create directories outside of the trinity root directory\n"
+)
+
+
+BootFn = Callable[[
+    Namespace,
+    TrinityConfig,
+    Dict[str, Any],
+    PluginManager,
+    logging.handlers.QueueListener,
+    EventBus,
+    Endpoint,
+    logging.Logger
+], None]
+
+
+def main_entry(trinity_boot: BootFn) -> None:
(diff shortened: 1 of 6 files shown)


cburgdorf:
Btw, we *could* use the exact same approach to also add a `trinity-validator` program. Just to be clear: This would still be an entirely separate piece of software that e.g. can against a prysmatic/lighthouse/whatever beacon node.

The main advantage here is just that:

- we share the multiprocessing infrastructure (e.g event bus)
- we share our plugin infrastructure
- we share db bootstrapping
- we share basic logging bootstrapping
- and obviously we have access to all other primitives that the Trinity code base provides

Basically, just like the `eth` modules provides building blocks for both Ethereum 1 and 2, Trinity provides building blocks and executables for all the different node types of the Ethereum space.

They are still different pieces of software from an OS point of view. It's more about code organization and sharing.

/cc @djrtwo

hwwhww:
Thank you so much for making a kickoff of Trinity side! :D

The isolation looks good to me.

> I left the actual trinity_boot method to be defined by each different program. This is mainly because current Trinity uses this shared networking process to kickoff the actual node logic and it might be that we don't need this shared process at all for the beacon node.

To make sure, do you mean we can start with running one single beacon node that proposes all beacon blocks? Makes sense to me.

> Btw, we could use the exact same approach to also add a trinity-validator program.

Ah,  see how many reusable functions here for the program side. I believe the snake charmers have more knowledge to balance the module isolation and reducing repetition, or maybe more refactoring if needs. For starting prototyping, IMO having a simple `trinity-validator` looks reasonable.

cburgdorf:
>Thank you so much for making a kickoff of Trinity side!

It's a small step forward but I'm quite excited to dip my toe into this. I guess I'll have to pick your brain a lot to build a deeper understanding of all the things that are coming :)

>To make sure, do you mean we can start with running one single beacon node that proposes all beacon blocks?

What I meant is: In the current Ethereum 1.0 trinity node, we spawn a process that we call `networking`

https://github.com/ethereum/py-evm/blob/f2d0d5d187400ba46a6b8f5b1f1c9997dc7fbb5a/trinity/main.py#L237-L241

This process hosts the `Node` and a *secondary* `PluginManager`

https://github.com/ethereum/py-evm/blob/f2d0d5d187400ba46a6b8f5b1f1c9997dc7fbb5a/trinity/main.py#L340-L357

Basically, the role of the current `Node` class is to host the `PeerPool` and all the networking as well as the syncing process.

However, we've taken a path in which we aim to separate these things to run in more fine grained dedicated processes and communicate through events with each other.

I'm not making a definitive statement that the `Node` class will entirely disappear but even if it stays I bet it will look a whole lot different from today.

With the beacon node, we can take these learnings and separate these things right from the start.

I guess we will need at least the following things for the beacon node:

- PeerPool
- Syncing
- JSON RPC API for validators
- ???

All of these should be written as isolated plugins and we have a communication pattern in place to have them talk to each other.

What are the things that you'd like to see happen first? I guess it's a bit of a chicken and egg situation but we can start stubbing out things. E.g. we can produce some fake beacon chain, sketch out a JSON-RPC API and kick off a new `trinity-validator` node that consumes the API. That should be relatively straight forward.

I guess the most complicated stuff (for me at least) would be the actual `PeerPool` because this will build on py-libp2p which I guess is still in alpha stages.

What would be the most important first milestone in regard to Trinity integration that you'd like to see happen soon? If you can teach me which milestones would be important for you, then I'm sure we can work out a viable path to codify it :)

cburgdorf:
Paging in @carver and @pipermerriam? Are you ok with the direction this PR takes to kickstart a `trinity-beacon` node. Any other comments?

djrtwo:
I'd say the "JSON RPC API for validators" is the most important milestone at the moment rather than the peering and syncing. In a test implementation, a validator can control _all_ keys and build/attest to the entire chain via a JSON RPC connection to a single node.

cburgdorf:
@djrtwo cool! The existing [JSON-RPC API](https://github.com/ethereum/py-evm/blob/master/trinity/plugins/builtin/json_rpc/plugin.py) for Trinity is already implemented as an isolated plugin which we can treat as a rough blueprint for this.

I can sketch out some code for this fairly soon.

trinity/bootstrap.py
```diff
@@ -0,0 +1,282 @@
...
+
+    try:
+        trinity_config = TrinityConfig.from_parser_args(args)
+    except AmbigiousFileSystem:
+        parser.error(TRINITY_AMBIGIOUS_FILESYSTEM_INFO)
+
+    if not is_data_dir_initialized(trinity_config):
+        # TODO: this will only work as is for chains with known genesis
+        # parameters.  Need to flesh out how genesis parameters for custom
+        # chains are defined and passed around.
```
pipermerriam: Is this still accurate?

pipermerriam: Is this still accurate?

trinity/bootstrap.py
```diff
@@ -0,0 +1,282 @@
...
+
+def display_launch_logs(trinity_config: TrinityConfig) -> None:
+    logger = logging.getLogger('trinity')
+    logger.info(TRINITY_HEADER)
+    logger.info("Started main process (pid=%d)", os.getpid())
+    logger.info(construct_trinity_client_identifier())
+    logger.info("Trinity DEBUG log file is created at %s", str(trinity_config.logfile_path))
+
+
+@setup_cprofiler('run_database_process')
```
pipermerriam: side-note.

I think this is junk and we could drop it.  The way it works I don't believe it provides us with very useful profiling data (but my memory may be wrong here).

cburgdorf: I just tried running `trinity --profile` and it produces two profile dumps and after installing `cprofilev` I can simply run `cprofilev -f <name-of-dump-file` to start inspecting.

image

So, I actually think this is very valuable and should be documented better. Or is there something else that makes you believe the data itself isn't useful?

cburgdorf: I created a separate issue for this to not block this PR. https://github.com/ethereum/py-evm/issues/1548

trinity/bootstrap.py
cburgdorf: I believe it is. I will soon dig deeper into this.

trinity/bootstrap.py
pipermerriam: I seem to recall that one deficiency is that it doesn't cover things which run in our process executors.  Otherwise :+1: that it does what we want.

PR #1537, opened 2018-11-30
Remove obsolete lahja hack

cburgdorf:
### What was wrong?

Code used a hack that isn't needed anymore.

### How was it fixed?

Removed hack.

#### Cute Animal Picture

### DIFF
--- a/trinity/nodes/base.py
+++ b/trinity/nodes/base.py
@@ -54,20 +54,15 @@ class Node(BaseService):
         self.event_bus = event_bus
 
     async def handle_network_id_requests(self) -> None:
-        async def f() -> None:
-            # FIXME: There must be a way to cancel event_bus.stream() when our token is triggered,
-            # but for the time being we just wrap everything in self.wait().
-            async for req in self.event_bus.stream(NetworkIdRequest):
-                # We are listening for all `NetworkIdRequest` events but we ensure to only send a
-                # `NetworkIdResponse` to the callsite that made the request.  We do that by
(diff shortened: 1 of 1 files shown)


trinity/nodes/base.py
```diff
@@ -54,20 +54,15 @@ def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
...
-                # `NetworkIdResponse` to the callsite that made the request.  We do that by
-                # retrieving a `BroadcastConfig` from the request via the
-                # `event.broadcast_config()` API.
-                self.event_bus.broadcast(
-                    NetworkIdResponse(self._network_id),
-                    req.broadcast_config()
-                )
-
-        await self.wait(f())
+        async for req in self.event_bus.stream(NetworkIdRequest):
```
carver: Ah, that's why I've been seeing `handle_network_id_requests` failures on shutdown. :smile: 

The event bus is never getting cancelled, so it hangs waiting for this asynchronous stream. I think this would do the trick:

```
        async for req in self.wait_iter(self.event_bus.stream(NetworkIdRequest)):
```

cburgdorf: Lahja does handle the cancellations but there was a bug that re-raised the cancellation. That was fixed here.

https://github.com/ethereum/lahja/pull/24

And it should now be fixed that we upgraded to lahja `0.10.0`.

https://github.com/ethereum/py-evm/pull/1567

Issue #1484, opened 2018-11-20
Conflicts within dev requirements installation.


pipermerriam:
### What is wrong?

A `pip install -e .[dev]` from the current master (https://github.com/ethereum/py-evm/tree/6fca8f723244c2e7f920a6aa18689ae61e257d03) results in the following errors:

```
eth-abi 1.2.2 has requirement eth-typing<2, but you'll have eth-typing 2.0.0 which is incompatible.
pytest 3.6.4 has requirement pluggy<0.8,>=0.5, but you'll have pluggy 0.8.0 which is incompatible.
```

### How can it be fixed

- The `pytest/pluggy` conflict comes from `tox`
- The `eth-abi/typing` conflict comes from `web3.py`

We should be able to fix the `pytest/pluggy/tox` issue by futzing until we find a compatable pair of `pytest/tox` versions.  Note that both have had major version bumps so we should work to upgrade them.

We should be able to fix the `eth-abi` one by making a `1.x` release of `eth-abi` that supports `eth-typing>=2` as well as a web3.py release that allows for this if needed.

javimaravillas:
I will continue with this issue to fix: 
```sh
eth-abi 1.2.2 has requirement eth-typing<2, but you'll have eth-typing 2.0.0 which is incompatible.
```

rayrapetyan:
@javimaravillas, could you also upgrade coincurve to v10 as part of this task? Outdated v8 produces some issues on certain OSes (FreeBSD).
Basically replace:
```
coincurve>=8.0.0,<9.0.0
```
with
```
coincurve>=10.0.0,<11.0.0
```
in two places. I don't think it worth a separate PR. Thanks.

javimaravillas:
Done

javimaravillas:
Waiting for https://github.com/ethereum/eth-abi/pull/110 to be merged and a new version released from eth-abi 1.x for updating the PR and fixing this issue.

javimaravillas:
The previous two error has been fixed, but another error has come:
```d
requests 2.20.1 has requirement idna<2.8,>=2.5, but you'll have idna 2.8 which is incompatible.
```
Fixing idna to 2.7 is making this error to disappear.

javimaravillas:
This issue is fixed now in: https://github.com/ethereum/py-evm/pull/1538

javimaravillas:
This one could be closed now

carver:
Thanks @javimaravillas !

PR #1538, opened 2018-12-01
eth-abi 1.2.2 has requirement eth-typing<2, but you'll have eth-typing 2.0.0 which is incompatible.but you'll ha…

javimaravillas:
https://github.com/ethereum/py-evm/issues/1484

### What was wrong?

1. pytest 3.6.4 has requirement pluggy<0.8,>=0.5, but you'll have pluggy 0.8.0 which is incompatible.
2. eth-abi 1.2.2 has requirement eth-typing<2, but you'll have eth-typing 2.0.0 which is incompatible.
3. requests 2.20.1 has requirement idna<2.8,>=2.5, but you'll have idna 2.8 which is incompatible.

### How was it fixed?

1. tox was installing the latest version of pluggy 0.8 that is incompatible with pytest so pytest was not working. Fixing a version of pluggy minor to 0.8 makes pytest work without an issue for tox

2. Releasing a new version of eth-abi using eth-typing 2.0.0 (waiting for a PR in eth-abi to be ready)
3. Fixing idna to use 2.7 is fixing the issue

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ deps = {
     # Installing these libraries may make the evm perform better than
     # using the default fallbacks though.
     'eth-extra': [
-        "coincurve>=8.0.0,<9.0.0",
+        "coincurve>=10.0.0,<11.0.0",
         "eth-hash[pysha3];implementation_name=='cpython'",
         "eth-hash[pycryptodome];implementation_name=='pypy'",
         "plyvel==1.0.5",
@@ -40,7 +40,7 @@ deps = {
         "async-generator==1.10",
         "bloom-filter==1.3",
         "cachetools>=2.1.0,<3.0.0",
-        "coincurve>=8.0.0,<9.0.0",
+        "coincurve>=10.0.0,<11.0.0",
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.5",
         "web3==4.4.1",
--- a/setup.py
+++ b/setup.py
@@ -83,6 +83,7 @@ deps = {
         "bumpversion>=0.5.3,<1",
         "wheel",
         "setuptools>=36.2.0",
+        "pluggy==0.7.1",
         "tox==2.7.0",
         "twine",
     ],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -40,7 +40,7 @@
         "async-generator==1.10",
         "bloom-filter==1.3",
         "cachetools>=2.1.0,<3.0.0",
-        "coincurve>=8.0.0,<9.0.0",
+        "coincurve>=10.0.0,<11.0.0",
```
cburgdorf: @javimaravillas what's the reason for bumping these?

cburgdorf: Ah, ok I just read it was suggested [here](https://github.com/ethereum/py-evm/issues/1484) as lower versions have issues with FreeBSD. Sounds fair.

setup.py
```diff
@@ -83,6 +83,7 @@
         "bumpversion>=0.5.3,<1",
         "wheel",
         "setuptools>=36.2.0",
+        "pluggy==0.7.1",
```
cburgdorf: I think if we add this as a standalone dependency to workaround the issue, we should put a comment above that line as an explainer. Something like:

>Pinning pluggy version to workaround incompatibility issues between certain `pytest` and `tox` versions.

javimaravillas: Done

javimaravillas:
This PR is waiting for https://github.com/ethereum/eth-abi/pull/110

javimaravillas:
This PR is complete to review and merge

setup.py
```diff
@@ -83,6 +83,10 @@
         "bumpversion>=0.5.3,<1",
         "wheel",
         "setuptools>=36.2.0",
+        #Fixing this dependency due to: pytest 3.6.4 has requirement pluggy<0.8,>=0.5, but you'll have pluggy 0.8.0 which is incompatible.
+        "pluggy==0.7.1",
+        #Fixing this dependency due to: requests 2.20.1 has requirement idna<2.8,>=2.5, but you'll have idna 2.8 which is incompatible.
+        "idna==2.7",
```
carver: Just as an FYI, I think it would be *ideal* to do `>=2.5,<2.8` here, but since it's only a dev dependency, I'm okay pinning it.

PR #1540, opened 2018-12-01
update readme development setup guide

javimaravillas:
### What was wrong?

Nothing really wrong, a small update in the readme for development setup

### How was it fixed?

update readme

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,7 @@ Then install the required python packages via:
 pip install -e .[dev]
 ```
 
+Note: For OSX, if you have an error during the build of ['plyvel'](https://github.com/wbolster/plyvel/issues/34#issuecomment-46703144), you will need to install leveldb with homebrew: brew install leveldb
 
 ### Running the tests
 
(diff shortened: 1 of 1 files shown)


cburgdorf:
Went ahead and restructured the README entirely: https://github.com/ethereum/py-evm/pull/1541
I'm closing this as I prefer to move us more towards a very concise README and refer people to the official website for more infos.

carver:
> I actually think that we should just drop this whole Development section here just delegate the user to our docs site.

FWIW: we have a clear split in docs for most/all the other py repositories where the README is for development/contribution, and RTD is for usage. I like that in general, and like consistency, but don't feel strongly enough to push the issue on this repo. :)

cburgdorf:
I understand that concert but I think at some point there's just too much information to cram into a Readme and eventually it gets to a point where it just looks messy. 

I also think that people would expect some of the developer information in the docs and then we are at a point again where we would have to maintain it in two places (which already was the case before this PR).

PR #1541, opened 2018-12-03
Simplify and deduplicate README

cburgdorf:
### What was wrong?

Our Readme feels a bit overloaded and it contains information that is better covered in the docs.

### How was it fixed?

- cleaned up Readme
- moved some things into the docs
- fixed some broken links
- updated the introduction texts

Here is how that looks like now:

image

To be honest, I could be convinced to even go a step further and remove this whole *goals* stuff and just go with a very short intro on Py-EVM and Trinity.

#### Cute Animal Picture


cburgdorf:
@javimaravillas your PR #1540 triggered me to cleanup our README as well as some other parts as I think it was still very messy. Would love to hear your feedback on this.

PR #1543, opened 2018-12-03
Define beacon chain network commands

### LINKED ISSUE #1542: Implement becaon chain MVP network protocol
We need the following commands:

### For the handshake

- `Status`: Exchange protocol version, network id, and head hash

### For chain syncing

- `GetBeaconBlock`: Get a specific block by hash or number
- `BeaconBlock`: Reply with a requested block

### For building the chain

- `BeaconBlock`: Same command as above, but sent with out request
- `AttestationRecord`: Send attestations

### What we ignore for now to keep it simple

- Requesting/sending multiple blocks
- State syncing
- `SpecialRecord`s

jannikluhn:
Closes #1542

#### Cute Animal Picture

### DIFF
--- a/trinity/protocol/bcc/commands.py
+++ b/trinity/protocol/bcc/commands.py
@@ -1,0 +1,36 @@
+from rlp import sedes
+
+from p2p.protocol import (
+    Command,
+)
+
+from trinity.rlp.sedes import HashOrNumber
+from eth.beacon.types.blocks import BaseBeaconBlock
+from eth.beacon.types.attestation_records import AttestationRecord
+
+
+class Status(Command):
+    _cmd_id = 0
+    structure = [
+        ('protocol_version', sedes.big_endian_int),
+        ('network_id', sedes.big_endian_int),
+        ('best_hash', sedes.binary),
+    ]
+
+
+class GetBeaconBlocks(Command):
+    _cmd_id = 1
+    structure = [
+        ('block_number_or_hash', HashOrNumber()),
+        ('max_headers', sedes.big_endian_int),
+    ]
+
+
+class BeaconBlocks(Command):
+    _cmd_id = 2
+    structure = sedes.CountableList(BaseBeaconBlock)
+
+
+class AttestationRecords(Command):
+    _cmd_id = 3
+    structure = sedes.CountableList(AttestationRecord)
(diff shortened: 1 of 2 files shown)


trinity/protocol/bcc/proto.py
```diff
@@ -0,0 +1,18 @@
...
+
+from trinity.protocol.bcc.commands import (
+    Status,
+    GetBeaconBlocks,
+    BeaconBlocks,
+    AttestationRecords,
+)
+
+
+class BCCProtocol(Protocol):
```
hwwhww: Hey hey hey... BitConnect! 😂

trinity/protocol/bcc/commands.py
```diff
@@ -0,0 +1,36 @@
...
+        ('network_id', sedes.big_endian_int),
+        ('best_hash', sedes.binary),
+    ]
+
+
+class GetBeaconBlocks(Command):
+    _cmd_id = 1
+    structure = [
+        ('block_number_or_hash', HashOrNumber()),
+        ('max_headers', sedes.big_endian_int),
```
hwwhww: Not beacon block headers. Maybe `max_blocks`?
```suggestion
        ('max_blocks', sedes.big_endian_int),
```

trinity/protocol/bcc/commands.py
```diff
@@ -0,0 +1,36 @@
...
+        ('protocol_version', sedes.big_endian_int),
+        ('network_id', sedes.big_endian_int),
+        ('best_hash', sedes.binary),
+    ]
+
+
+class GetBeaconBlocks(Command):
+    _cmd_id = 1
+    structure = [
+        ('block_number_or_hash', HashOrNumber()),
```
hwwhww: Note that it's actually `block slot`. Maybe:
```suggestion
        ('block_slot_or_hash', HashOrNumber()),
```

trinity/protocol/bcc/proto.py
```diff
@@ -0,0 +1,18 @@
...
+    name = "bcc"
+    version = 0
+    _commands = [Status, GetBeaconBlocks, BeaconBlocks, AttestationRecords]
+    cmd_length = 4
```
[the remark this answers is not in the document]
jannikluhn: Semi-confident :)

If I understand correctly, the `cmd_length` is only used for protocols higher up the stack so that they can choose command ids that don't have collisions with lower level protocols (by offsetting them with the sum of all `cmd_length`s below). I assume `eth` and `les` use a larger number than necessary to reserve some commands for future use.

Issue #1545, opened 2018-12-04
Introduce label system for plugins


cburgdorf:
### What is wrong?

Today, trinity supports two different node types:

- Ethereum 1.0 full node
- Ethereum 1.0 light node

In the future I believe that we want to support at least two more node types with this code base:

- Ethereum 2.0 / Serenity beacon node
- Ethereum 2.0 / Serenity validator node
- ???

There's already PR #1536 that adds support for a `trinity-beacon` command to demonstrate one possible path to maintain an entirely different node type in this code base.

The way I see our client architecture evolving Trinity is basically layered in:

- Reusable APIs / Primitives
- Plugins
- Core (DB Access, Event Bus, Plugin Infrastructure, Logging)

We end up with a relatively lightweight core, a set of libraries and plugins that wire things together.

In this kind of architecture, not every plugin ends up being useful for every node type. Some of our plugins may be useful for every node type e.g.

- `trinity attach`
- `trinity fix-unclean-shutdown` (also, in the long term I hope we can obsolete that one)

Some of them may only be useful for an Ethereum 1 light node e.g.

- `LightPeerChainBridge`

Even today already ensure that some plugins would only run in some context e.g.

https://github.com/ethereum/py-evm/blob/8ba5ed751926ae993487aed10b3b7c65aa6a8cde/trinity/plugins/builtin/light_peer_chain_bridge/plugin.py#L43-L45

However, we not only need to keep plugins from running in the wrong context, some plugins simply should not be loaded at all so that they don't even show up in the context of `--help`.

### How can it be fixed

Introduce some form of label system for plugins. Similar how plugins advertise their name, they could advertise labels e.g.

```python
    @property
    def labels(self) -> Tuple[str,..]:
        return (PluginLabels.ETHEREUM1_FULL_NODE, PluginLabels.ETHEREUM1_LIGHT_NODE)
```

This would give nodes (even 3rd party nodes that others may build) the flexibility to simply filter the plugins that are registered at a `PluginManager` instance for specific labels.

cburgdorf:
This is flagged as "Good First Issue" but it needs some discussion before we decide if we want to go that way. /cc @carver @pipermerriam

pipermerriam:
I'm :-1: on "Good First Issue".

pipermerriam:
I'd like to put this into context with the boot process.  It looks something like:

1. initial plugin setup.
2. plugins modify CLI parser
3. CLI parser parses arguments
[...]

PR #1549, opened 2018-12-05
Separate plugins between trinity and trinity-beacon

cburgdorf:
### What was wrong?

trinity and trinity-beacon are two different pieces
of software that just share a common code base.

Hence they need to be bootstrapped with different
built-in plugins.

Relates to: #1545

### How was it fixed?

Make `setup_plugins` take an `Iterable[BasePlugin]` and separate the plugins in each bootstrapping code.

#### Cute Animal Picture

### DIFF
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -40,12 +40,10 @@ from trinity.constants import (
     ROPSTEN_NETWORK_ID,
 )
 from trinity.extensibility import (
+    BasePlugin,
     BaseManagerProcessScope,
     MainAndIsolatedProcessScope,
     PluginManager,
-)
-from trinity.plugins.registry import (
-    get_all_plugins,
 )
 from trinity.utils.ipc import (
     kill_process_gracefully,
@@ -108,13 +106,14 @@ BootFn = Callable[[
 ], None]
 
 
-def main_entry(trinity_boot: BootFn) -> None:
+def main_entry(trinity_boot: BootFn, plugins: Iterable[BasePlugin]) -> None:
     event_bus = EventBus(ctx)
     main_endpoint = event_bus.create_endpoint(MAIN_EVENTBUS_ENDPOINT)
     main_endpoint.connect_no_wait()
 
     plugin_manager = setup_plugins(
-        MainAndIsolatedProcessScope(event_bus, main_endpoint)
+        MainAndIsolatedProcessScope(event_bus, main_endpoint),
+        plugins
     )
     plugin_manager.amend_argparser_config(parser, subparser)
     args = parser.parse_args()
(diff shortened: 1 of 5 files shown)


trinity/plugins/registry.py
```diff
@@ -35,10 +35,14 @@ def is_ipython_available() -> bool:
         return True
 
 
-BUILTIN_PLUGINS = (
+COMMON_BUILTIN_PLUGINS: Tuple[BasePlugin, ...] = (
     AttachPlugin() if is_ipython_available() else AttachPlugin(use_ipython=False),
```
pipermerriam: I just realized this can be simplified to `AttachPlugin(use_ipython=is_ipython_available())`

trinity/plugins/registry.py
```diff
@@ -35,10 +35,14 @@ def is_ipython_available() -> bool:
         return True
 
 
-BUILTIN_PLUGINS = (
+COMMON_BUILTIN_PLUGINS: Tuple[BasePlugin, ...] = (
```
pipermerriam: Alternate name ideas:

- `CORE_PLUGINS`
- `BASE_PLUGINS`
- `SYSTEM_PLUGINS`

carver: +1 to core or base

trinity/bootstrap.py
```diff
@@ -108,13 +106,14 @@
 ], None]
 
 
-def main_entry(trinity_boot: BootFn) -> None:
+def main_entry(trinity_boot: BootFn, plugins: Iterable[BasePlugin]) -> None:
```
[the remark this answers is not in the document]
cburgdorf: The rule of thumb (which we also put into the [tactical manual](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#function-variance)) is to always use the most abstract type for **function parameters** that qualifies the function body.

And for **function return values**, usually use the concrete type that is actually being returned.

This btw mimics the default behavior that you'd also get if you do not use type hints at all. If you don't apply any type hints, you'll be able to pass in any parameter that qualifies what is needed by the function body.

In this particular case, there is nothing why we would need to restrict it to `Sequence[BasePlugin]` (e.g. we do not need to access `len()`) and every `Sequence[BasePlugin]` implicitly qualifies as an `Iterable[BasePlugin]` (but not the reverse).

trinity/plugins/registry.py
cburgdorf: Ah, yes. I think at one point that were two different classes and then later got refactored to a constructor parameter again but left the `if/else` in place. Will change it.

trinity/plugins/registry.py
cburgdorf: `BASE_PLUGINS` has my vote.

PR #1552, opened 2018-12-05
Implement bcc protocol class and handshake

jannikluhn:
This implements the protocol class and the basis for the peer class for the beacon chain network. It's mostly boilerplate, tests, and a little handshake logic.

#### Cute Animal Picture


jannikluhn:
> Currently, this is all RLP serialized. This will all move to SSZ eventually, right?

Eventually yes. But probably not for the devp2p testnet, as the current devp2p implementation is (presumably) not designed for mixing different encodings.

PR #1554, opened 2018-12-05
New cancellable asynchronous iterators

carver:
### What was wrong?

I wanted a clean way to iterate through an asynchronous generator, cancelling early if the service exits. Didn't find any good built-in options.

### How was it fixed?

Added `wait_iter` used like:

```py
async for val in service.wait_iter(my_async_iterator()):
  do_stuff(val)
```

(The intent is to roughly match the `await self.wait(async_method())` API)

#### Cute Animal Picture

### DIFF
--- a/p2p/cancellable.py
+++ b/p2p/cancellable.py
@@ -1,27 +1,28 @@
 from typing import (
+    AsyncIterator,
     Awaitable,
     TypeVar,
 )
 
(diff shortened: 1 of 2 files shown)


carver:
Working on adding tests. Otherwise good to review. (Another strip-out from #1392 )

PR #1556, opened 2018-12-06
Move peer discovery into an isolated plugin

cburgdorf:
### What was wrong?

Currently the peer discovery runs in the networking process among a lot of other things. Moving it into an isolated plugin should:

- increase efficiency of syncing as the peer discovery probably takes a fair share of computing power
- increase decoupling as it breaks up the networking code into smaller pieces 
- improve reusability as the new discovery can be accessed from any other interested party since it is integrated with the event bus 

### How was it fixed?

- inverted the model in the `DiscoveryService` so that it doesn't know about the `PeerPool` and instead just exposes an API for consumers to request peer candidates from
- the API is exposed over the EventBus so that it can be requested from any other process
- created an isolated plugin to run the new discovery service in a separate process
- removed discovery code from the server where it previously lived

#### Cute Animal Picture

### DIFF
--- a/p2p/discovery.py
+++ b/p2p/discovery.py
@@ -32,6 +32,10 @@ from typing import (
 
 import cytoolz
 
+from lahja import (
+    Endpoint,
+)
+
 import rlp
 
 from eth_typing import Hash32
@@ -56,10 +60,11 @@ from eth.tools.logging import ExtendedDebugLogger, DEBUG2_LEVEL_NUM
 
 from cancel_token import CancelToken, OperationCancelled
 
+from p2p.events import PeerCandidatesRequest, RandomBootnodeRequest
 from p2p.exceptions import AlreadyWaitingDiscoveryResponse, NoEligibleNodes, UnableToGetDiscV5Ticket
+from p2p.kademlia import to_uris
 from p2p import kademlia
 from p2p import protocol
-from p2p.peer import BasePeerPool
 from p2p.service import BaseService
 
 if TYPE_CHECKING:
@@ -951,21 +956,48 @@ class DiscoveryService(BaseService):
     _last_lookup: float = 0
     _lookup_interval: int = 30
 
-    def __init__(self, proto: DiscoveryProtocol, peer_pool: BasePeerPool,
-                 port: int, token: CancelToken = None) -> None:
+    def __init__(self,
+                 proto: DiscoveryProtocol,
+                 port: int,
+                 event_bus: Endpoint,
+                 token: CancelToken = None) -> None:
         super().__init__(token)
         self.proto = proto
-        self.peer_pool = peer_pool
         self.port = port
+        self._event_bus = event_bus
         self._lookup_running = asyncio.Lock()
+
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
+        self.run_daemon_task(self.handle_get_random_bootnode_requests())
+
+    async def handle_get_peer_candidates_requests(self) -> None:
+        async for event in self._event_bus.stream(PeerCandidatesRequest):
+
+            self.run_task(self.maybe_lookup_random_node())
+
+            nodes = tuple(to_uris(self.proto.get_nodes_to_connect(event.max_candidates)))
+
+            self.logger.debug2("Broadcasting peer candidates (%s)", nodes)
+            self._event_bus.broadcast(
+                event.expected_response_type()(nodes),
+                event.broadcast_config()
+            )
+
+    async def handle_get_random_bootnode_requests(self) -> None:
+        async for event in self._event_bus.stream(RandomBootnodeRequest):
+
+            nodes = tuple(to_uris(self.proto.get_random_bootnode()))
+
+            self.logger.debug2("Broadcasting random boot nodes (%s)", nodes)
+            self._event_bus.broadcast(
+                event.expected_response_type()(nodes),
+                event.broadcast_config()
+            )
 
     async def _run(self) -> None:
         await self._start_udp_listener()
-        connect_loop_sleep = 2
         self.run_task(self.proto.bootstrap())
-        while self.is_operational:
-            await self.maybe_connect_to_more_peers()
-            await self.sleep(connect_loop_sleep)
+        await self.cancel_token.wait()
 
     async def _start_udp_listener(self) -> None:
         loop = asyncio.get_event_loop()
@@ -974,22 +1006,6 @@ class DiscoveryService(BaseService):
             lambda: self.proto,
             local_addr=('0.0.0.0', self.port),
             family=socket.AF_INET)
-
-    async def maybe_connect_to_more_peers(self) -> None:
-        """Connect to more peers if we're not yet maxed out to max_peers"""
-        if self.peer_pool.is_full:
-            self.logger.debug("Already connected to %s peers; sleeping", len(self.peer_pool))
-            return
-
-        self.run_task(self.maybe_lookup_random_node())
-
-        await self.peer_pool.connect_to_nodes(
-            self.proto.get_nodes_to_connect(self.peer_pool.max_peers))
-
-        # In some cases (e.g ROPSTEN or private testnets), the discovery table might be full of
-        # bad peers so if we can't connect to any peers we try a random bootstrap node as well.
-        if not len(self.peer_pool):
-            await self.peer_pool.connect_to_nodes(self.proto.get_random_bootnode())
 
     async def maybe_lookup_random_node(self) -> None:
         if self._last_lookup + self._lookup_interval > time.time():
--- a/p2p/events.py
+++ b/p2p/events.py
@@ -1,4 +1,5 @@
 from typing import (
+    Tuple,
     Type,
 )
 
@@ -6,6 +7,36 @@ from lahja import (
     BaseEvent,
     BaseRequestResponseEvent,
 )
+
+
+class BaseDiscoveryServiceResponse(BaseEvent):
+
+    def __init__(self, error: Exception) -> None:
+        self.error = error
+
+
+class PeerCandidatesResponse(BaseDiscoveryServiceResponse):
+
+    def __init__(self, candidates: Tuple[str, ...], error: Exception=None) -> None:
+        super().__init__(error)
+        self.candidates = candidates
+
+
+class PeerCandidatesRequest(BaseRequestResponseEvent[PeerCandidatesResponse]):
+
+    def __init__(self, max_candidates: int) -> None:
+        self.max_candidates = max_candidates
+
+    @staticmethod
+    def expected_response_type() -> Type[PeerCandidatesResponse]:
+        return PeerCandidatesResponse
+
+
+class RandomBootnodeRequest(BaseRequestResponseEvent[PeerCandidatesResponse]):
+
+    @staticmethod
+    def expected_response_type() -> Type[PeerCandidatesResponse]:
+        return PeerCandidatesResponse
 
 
 class PeerCountResponse(BaseEvent):
--- a/p2p/peer.py
+++ b/p2p/peer.py
@@ -48,7 +48,10 @@ from lahja import Endpoint
 
 from p2p import auth
 from p2p import protocol
-from p2p.kademlia import Node
+from p2p.kademlia import (
+    from_uris,
+    Node,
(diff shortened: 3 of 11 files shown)


p2p/peer.py
```diff
@@ -799,6 +804,40 @@ def __init__(self,
...
+    async def maybe_connect_more_peers(self) -> None:
+        discovery_interval = 2
+            
+        while self.is_operational:
+            await self.sleep(discovery_interval)
+
+            left_peer_count = self.max_peers - len(self)
+            if left_peer_count > 0:
+                try:
+                    response = await self.wait(
```
pipermerriam: I wanted to point something out that may not be obvious.

The previous implementation of this started with `Discovery.maybe_connect_to_more_peers` which calls `DiscoveryProto.get_nodes_to_connect` which depending on the protocol calls `kademlia.RoutingTable.get_random_nodes()` or in the case of v5, uses the topics table to return nodes.

https://github.com/ethereum/py-evm/blob/master/p2p/kademlia.py#L236-L257

https://github.com/ethereum/py-evm/blob/master/p2p/discovery.py#L917-L923

You'll notice that both of these implementations use generators to return the list of candidates, and that these generators are lazily consumed and sometimes discarded before being fully consumed.

If I'm reading this correctly, this PR will change this functionality so that the full iterable of candidates is no longer lazily generated.  I don't think that it's a very big issue but I wanted to make note of it.

trinity/plugins/builtin/peer_discovery/events.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class PeerCandidatesResponse(BaseDiscoveryServiceResponse):
+
+    def __init__(self, candidates: Tuple[str, ...], error: Exception=None) -> None:
+        super().__init__(error)
+        self.candidates = candidates
+
+
+class GetPeerCandidatesRequest(BaseRequestResponseEvent[PeerCandidatesResponse]):
```
pipermerriam: Mirroring my comment here.  I think these events and the actual `DiscoveryService` can still live in the `p2p` codebase.

p2p/peer.py
cburgdorf: Thanks for pointing this out.

Previously `connect_to_nodes` was given a generator

https://github.com/ethereum/py-evm/blob/b10f1edc4a9a316f72ce3905a62219ba6ca79b90/p2p/discovery.py#L983-L984

 and consumed it lazily as seen here.

https://github.com/ethereum/py-evm/blob/b10f1edc4a9a316f72ce3905a62219ba6ca79b90/p2p/peer.py#L938-L948

The *lazy* aspect of this basically comes into play if the pool becomes *full* e.g. after connecting to the first node. Then, even if the generator was meant to yield e.g. 5 nodes, it would stop being consumed after the first one.

We can not have this under the new design (unless we want to make lazy lahja based cross process generators a thing but honestly that will end up being inefficient as hell) because we are requesting e.g. 5 nodes from another process so that other process needs to consume all 5 of them, wrap them in a tuple and send them over the to that other process that requested them.

But as you said, I don't think that will be an issue. Getting a bunch of nodes out of the routing table shouldn't be a costly operation *I guess*.

cburgdorf:
Another reason to wait on a fresh Lahja release before this gets merged (apart from it still being WIP)

https://github.com/ethereum/lahja/pull/24

trinity/config.py
```diff
@@ -468,3 +468,7 @@ def node_class(self) -> Type['Node']:
     def process_id_file(self, process_name: str):  # type: ignore
         with PidFile(process_name, self.data_dir):
             yield
+
+
+class TrinityBeaconConfig(TrinityConfig):
```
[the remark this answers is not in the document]
pipermerriam: I'd rather not do derivative/inheritance base configurations.  What do you think about instead having some concept of sub-configs and an API something like:

- `TrinityConfig` -> Base configuration, only handles things that are truly *global*
- mechanism for adding *sub-config* classes which the `TrinityConfig` would be responsible for instantiating as part of the `TrinityConfig.__init__`
- Each sub-config class would have an API for initializing itself from the cli args: `BaseSubConfig.from_cli_args(cls, trinity_config, cli_args)`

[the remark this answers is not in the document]
cburgdorf: Yep, sounds good. Will give it a shot. Maybe I move Beacon node support into a follow up PR because this is mostly finished apart from that.

trinity/plugins/builtin/peer_discovery/plugin.py
```diff
@@ -0,0 +1,151 @@
...
+    Bootstrap discovery to provide a parent ``CancellationToken``
+    """
+
+    def __init__(self, event_bus: Endpoint, trinity_config: TrinityConfig) -> None:
+        super().__init__()
+        self.event_bus = event_bus
+        self.trinity_config = trinity_config
+
+    async def _run(self) -> None:
+        external_ip = "0.0.0.0"
```
cburgdorf: @pipermerriam in current `master` we use the (experimental) upnpservice to *maybe* resolve a different external IP first. I'm honestly not so sure how important that is or if it ever worked. I'm pretty sure I never had this working locally here but I guess it just doesn't apply to the local home network that I usually work from.

trinity/config.py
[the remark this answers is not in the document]
cburgdorf: I think your proposal has some nice properties that I just became aware of so I'm posting them here.

So, initially it may seem like having a `BaseSubConfig` and then derived classes from there doesn't buy us much from subclassing a `BaseTrinityConfig` directly. However the benefits become more obvious when we think about the possible use cases.

E.g Let's say we would have a `BaseTrinityConfig` and then another `Eth1TrinityConfig` for the core Ethereum 1 configs. Now lets say we want to add configs specific to light mode and lets assume they would be reusable across Ethereum 1 and 2. In this model, we would be forced to derive both `Eth1TrinityLightConfig` and `Eth2TrinityLightConfig` and duplicate the light node settings.

In constrast, with the proposed model, it works more like a mixin where we can just reuse the `LightSubConfig` to a compositon of `TrinityConfig`+ `Eth1SubConfig` + `LightSubConfig` as well as `TrinityConfig` + `Eth2SubConfig` + `LightSubConfig`.

So, when I first thought about this, I thought that straight top level inheritance would be just fine here as we are mostly talking about data with barely any behavior. But turns out, this has some nice extra properties (it also has some other nice properties) that just became obvious to me now.

So, thanks for pointing it out @pipermerriam

p2p/discovery.py
```diff
@@ -948,21 +953,49 @@ class DiscoveryService(BaseService):
...
         self._lookup_running = asyncio.Lock()
 
+        self.run_daemon_task(self.handle_get_peer_candidates_requests())
+        self.run_daemon_task(self.handle_get_random_bootnode_requests())
+
+    async def handle_get_peer_candidates_requests(self) -> None:
+        async for event in self._event_bus.stream(PeerCandidatesRequest):
+
+            # Not sure if we should be doing this here
+            self.run_task(self.maybe_lookup_random_node())
```
cburgdorf: @pipermerriam a controversial part I wanted to highlight. This is placed more or less randomly here. The idea was to somewhat mimic the old behavior. We could also just do that on a given interval. Not sure.

pipermerriam: I think this is fine.

p2p/discovery.py
```diff
@@ -948,21 +953,49 @@ class DiscoveryService(BaseService):
...
+
+            nodes = tuple(to_uris(self.proto.get_nodes_to_connect(event.max_candidates)))
```
pipermerriam: I assume the use of `to_uris` here is due to the inability to pickle the `Node` class?  If so, maybe this should be marked with a comment indicating why and link to a github issue for it (which probably needs to be created).

p2p/events.py
```diff
@@ -8,6 +9,36 @@
...
 
+class BaseDiscoveryServiceResponse(BaseEvent):
+
+    def __init__(self, error: Exception) -> None:
+        self.error = error
+
+
+class PeerCandidatesResponse(BaseDiscoveryServiceResponse):
+
+    def __init__(self, candidates: Tuple[str, ...], error: Exception=None) -> None:
```
pipermerriam: *might* be nice to use a custom `TypeVar` to represent the `str` representation of a `Node`

trinity/plugins/builtin/peer_discovery/plugin.py
pipermerriam: I believe it does indeed work, it just isn't highly reliable and when it fails it is a little noisy towards the user.

pipermerriam:
Needs to be rebased now that https://github.com/ethereum/py-evm/pull/1530 is merged to use `logger.debug2` instead of `logger.trace`

p2p/discovery.py
cburgdorf: Correct. Will create an issue and link to it.

p2p/events.py
cburgdorf: I guess you mean a `NewType` . Yes we *could* do that given how "short-lived" and "fenced" these `str` are I'm not sure if the extra `NewType` is worth it. It's really just a very small path that these travel as strings and really only because of the serialization issue.

trinity/plugins/builtin/peer_discovery/plugin.py
cburgdorf: But do you think it is ok to leave this hardcoded with `0.0.0.0`? The way I understand the upnpservice, it runs continuously to do port mapping so it is not just there to potentially give us a different external IP for discovery but also for later stages. Basically, I'm trying to figure out if this would be something that needs to be resolved for the discovery or if it is acceptable to go without it (and maybe create an issue to follow up on it)

carver: Can confirm: `upnpservice` does provide an external IP/port pair for me on a network.

Not sure about hard-coding to `0.0.0.0`. It sounds bad, but I won't be able to dig in today. So I guess... try it? :stuck_out_tongue:

cburgdorf:
This should definitely wait until we have more clarity around #1583

trinity/plugins/builtin/peer_discovery/plugin.py
cburgdorf: @carver do you prefer to keep this open until you have tried this on your upnp network? I also think we should move the UpnpService in an isolated plugin as well and then we can also put back that first external ip lookup. I don't think that service does much right now though. Apart from the initial lookup of the external ip, it is supposed to continuously provide a port mapping but I don't see how that would be used anywhere.

carver: No it's fine, I'll end up testing it in master :+1:

carver: > Apart from the initial lookup of the external ip, it is supposed to continuously provide a port mapping but I don't see how that would be used anywhere.

I could be mistaken, but I think the idea is that you have to keep pinging the router so that incoming requests to that port get directed to your computer. So it would be necessary for other peers to create incoming connections. Incoming peer connections is not a thing that seems to be working well (at all?) already, so definitely not worth holding up the PR.

trinity/plugins/builtin/peer_discovery/plugin.py
```diff
@@ -0,0 +1,143 @@
...
+from .discovery_service import (
+    DiscoveryService,
+)
+
+
+def get_protocol(trinity_config: TrinityConfig) -> Protocol:
+    # For now DiscoveryByTopicProtocol supports a single topic, so we use the latest
+    # version of our supported protocols. Maybe this could be more generic?
+    if isinstance(trinity_config, TrinityBeaconConfig):
+        return BCCProtocol
```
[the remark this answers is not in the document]
cburgdorf: @jannikluhn @hwwhww This just got merged. In theory this means that the `trinity-beacon` node can use the existing discovery now (including discv5). This is, because this plugin is currently provided as a base plugin for `trinity` as well as `trinity-beacon` (https://github.com/ethereum/py-evm/pull/1556/files#diff-bcb3f875bb745a964ec62cfb0d9ac188R44)

In practice though, the `trinity-beacon` command is just a stub at this point that just spins up and then kills itself right after bootstrap. Just pointing this out to give you a sense of how things are starting to get wired together.

PR #1560, opened 2018-12-07
eth/vm/logic/system: fix CREATE2 failures on BadOpcode test fixture

veox:
### What was wrong?

`CREATE2` does not match the spec, [https://eips.ethereum.org/EIPS/eip-1014](EIP-1014).

PR #1181 was therefore failing fixtures.

### How was it fixed?

Use missing gas amount of 32000 (as `CREATE` does); increment nonce.

### Cute Animal Picture

Source: unknown; found [here](https://gr8poems.blogspot.com/2010/12/two-parrots.html)

### DIFF
--- a/eth/vm/logic/system.py
+++ b/eth/vm/logic/system.py
@@ -227,13 +227,14 @@ class Create2(CreateByzantium):
         return CreateOpcodeStackData(endowment, memory_start, memory_length, salt)
 
     def get_gas_cost(self, data: CreateOpcodeStackData) -> int:
-        return constants.GAS_SHA3WORD * ceil32(data.memory_length) // 32
+        return constants.GAS_CREATE + constants.GAS_SHA3WORD * ceil32(data.memory_length) // 32
 
     def generate_contract_address(self,
                                   stack_data: CreateOpcodeStackData,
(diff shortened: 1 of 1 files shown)


eth/vm/logic/system.py
```diff
@@ -227,13 +227,14 @@ def get_stack_data(self, computation: BaseComputation) -> CreateOpcodeStackData:
...
     def get_gas_cost(self, data: CreateOpcodeStackData) -> int:
-        return constants.GAS_SHA3WORD * ceil32(data.memory_length) // 32
+        return constants.GAS_CREATE + constants.GAS_SHA3WORD * ceil32(data.memory_length) // 32
 
     def generate_contract_address(self,
                                   stack_data: CreateOpcodeStackData,
                                   call_data: bytes,
                                   computation: BaseComputation) -> Address:
 
+        computation.state.account_db.increment_nonce(computation.msg.storage_address)
```
cburgdorf: You are the man! :+1: I must have missed that detail in the spec. Thanks for helping out with this :pray:

veox: Heh. Sorry for making so much noise out of one missing line. ^_^

PR #1563, opened 2018-12-08
Rework helper functions - part 2

hwwhww:
### What was wrong?
#1510

### How was it fixed?
- Fix `get_shard_committees_at_slot`
    - The reason why I add `_get_shard_committees_at_slot` is for making the tests easier.
- Fix `get_block_hash`
- Add `get_beacon_proposer_index`
- Add `get_attestation_participants`
- Add `get_effective_balance`
- Add `get_new_validator_registry_delta_chain_tip`

Also, make the tests for these functions not depend on genesis state or block.

### DIFF
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -18,6 +18,10 @@ from eth_typing import (
     Hash32,
 )
 
+from eth.utils.bitfield import (
+    get_bitfield_length,
+    has_voted,
+)
 from eth.utils.numeric import (
     clamp,
 )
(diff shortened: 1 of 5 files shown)


eth/beacon/helpers.py
```diff
@@ -362,3 +340,104 @@ def get_block_committees_info(parent_block: 'BaseBeaconBlock',
...
+    )
+    try:
+        first_shard_committee = shard_committees[0]
+    except IndexError:
+        raise ValueError("shard_committees should not be empty.")
+
+    proposer_committee_size = len(first_shard_committee.committee)
+
+    if proposer_committee_size <= 0:
+        raise ValueError(
```
pipermerriam: Do you think `ValidationError` would be more appropriate for these?  My loose guideline has been:

- `ValidationError` for things that we *might* encounter in live network conditions (like someone sending us a transaction with an invalid signature).
- `ValueError` for things that indicate broken code (and that when they occur we address them directly rather than by `try/except` catching the `ValueError`)

eth/beacon/helpers.py
```diff
@@ -362,3 +340,104 @@ def get_block_committees_info(parent_block: 'BaseBeaconBlock',
...
+                                               pubkey: int,
+                                               flag: int) -> Hash32:
+    """
+    Compute the next hash in the validator registry delta hash chain.
+    """
+    return blake(
+        current_validator_registry_delta_chain_tip +
+        flag.to_bytes(1, 'big') +
+        index.to_bytes(3, 'big') +
+        # TODO: currently, we use 256-bit pubkey which is different form the spec
```
pipermerriam: Not for this PR but something I think we should address.

We have tended to use `eth_keys` data types for private and public key representations.  I suspect that we'll have to make some modifications to `eth_keys` to support new key types?  If this is accurate, do you think you can open issues both here and in `eth_keys` to track those needs?

tests/beacon/conftest.py
```diff
@@ -75,12 +75,12 @@ def sample_attestation_data_params():
...
 
 @pytest.fixture
-def sample_beacon_block_params():
+def sample_beacon_block_params(epoch_length):
     return {
         'slot': 10,
         'randao_reveal': b'\x55' * 32,
         'candidate_pow_receipt_root': b'\x55' * 32,
-        'ancestor_hashes': (),
+        'ancestor_hashes': tuple([ZERO_HASH32 for _ in range(2 * epoch_length)]),
```
pipermerriam: Maybe more idiomatic.

```python
ancestor_hashes = tuple(itertools.repeat(ZERO_HASH32, 2 * epoch_length))
```

eth/beacon/helpers.py
```diff
@@ -362,3 +340,104 @@ def get_block_committees_info(parent_block: 'BaseBeaconBlock',
...
+        ]
+    )
+
+    try:
+        shard_committee = shard_committees[0]
+    except IndexError:
+        raise ValueError("shard_committees should not be empty.")
+
+    if len(participation_bitfield) != get_bitfield_length(len(shard_committee.committee)):
+        raise ValueError(
```
djrtwo: Similarly, a user could trigger this by adding malformed attestations to a block.
Throw `ValidationError`

tests/beacon/test_helpers.py
```diff
@@ -112,160 +143,163 @@ def test_get_block_hash(
...
         from_slot,
         to_slot,
         epoch_length,
     )
     assert len(result) == to_slot - from_slot + 1
 
 
-@pytest.mark.xfail(reason="Need to be fixed")
-def test_get_hashes_to_sign(genesis_block, epoch_length):
+def test_get_hashes_to_sign(sample_block, epoch_length):
```
djrtwo: No longer needed

eth/beacon/helpers.py
```diff
@@ -170,54 +157,45 @@ def get_new_recent_block_hashes(old_block_hashes: Sequence[Hash32],
 # Get shards_committees or indices
 #
 @to_tuple
-def get_shards_committees_for_slot(
-        crystallized_state: 'CrystallizedState',
+def _get_shard_committees_at_slot(
```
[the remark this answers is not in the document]
hwwhww: Opened #1575 to address it.

PR #1566, opened 2018-12-10
Teach TrinityConfig to handle sub configs

cburgdorf:
### What was wrong?

As we aim to support more and more node types using one common `TrinityConfig` becomes seems to become less sustainable as it would mean that e.g. we need to add things for the beacon node into the `TrinityConfig` that will just be noise for other node types. In language theory speak, that would lower the cohesion of the `TrinityConfig`which is a bad thing.

### How was it fixed?

Related discussion: https://github.com/ethereum/py-evm/pull/1556#discussion_r240014047

1.This introduces a concept of sub configurations which need to be derived from `BaseSubConfig`. 

```python
class BaseSubConfig:

    def __init__(self, args: argparse.Namespace, base_config: TrinityConfig) -> None:
        self.args = args
        self.base_config = base_config
```

2. `TrinityConfig.from_parser_args` now accepts an additional `Iterable[BaseSubConfig` to bootstrap any number of given sub configs.

3. `TrinityConfig` now has APIs `has_sub_config()` and `get_sub_config` to work with sub configs 

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/core/configuration/test_trinity_config_object.py
+++ b/tests/trinity/core/configuration/test_trinity_config_object.py
@@ -15,6 +15,7 @@ from trinity.utils.chains import (
 )
 from trinity.config import (
     TrinityConfig,
+    BeaconAppConfig,
     DATABASE_DIR_NAME,
 )
 from trinity.utils.filesystem import (
@@ -90,3 +91,12 @@ def test_trinity_config_explictely_provided_nodekey(nodekey_bytes, as_bytes):
     )
 
     assert trinity_config.nodekey.to_bytes() == nodekey_bytes
+
+
+def test_trinity_config_sub_configs():
+    trinity_config = TrinityConfig(network_id=1)
+    trinity_config.initialize_app_configs(None, (BeaconAppConfig,))
+
+    assert trinity_config.has_app_config(BeaconAppConfig)
+    beacon_config = trinity_config.get_app_config(BeaconAppConfig)
+    assert type(beacon_config) is BeaconAppConfig
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -1,3 +1,7 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
 import argparse
 from contextlib import contextmanager
 import json
@@ -6,9 +10,11 @@ from typing import (
     Any,
     cast,
     Dict,
+    Iterable,
     TYPE_CHECKING,
     Tuple,
     Type,
+    TypeVar,
     Union,
 )
 
@@ -185,6 +191,9 @@ class ChainConfig:
         return self.genesis_data.vm_configuration
 
 
+TAppConfig = TypeVar('TAppConfig', bound='BaseAppConfig')
+
+
 class TrinityConfig:
     _trinity_root_dir: Path = None
 
@@ -202,6 +211,8 @@ class TrinityConfig:
     bootstrap_nodes: Tuple[KademliaNode, ...] = None
 
     _genesis_config: Dict[str, Any] = None
+
+    _app_configs: Dict[Type['BaseAppConfig'], 'BaseAppConfig'] = None
 
     def __init__(self,
                  network_id: int,
@@ -222,6 +233,7 @@ class TrinityConfig:
         self.sync_mode = sync_mode
         self.port = port
         self.use_discv5 = use_discv5
+        self._app_configs = {}
 
         if genesis_config is not None:
             self.genesis_config = genesis_config
@@ -284,14 +296,6 @@ class TrinityConfig:
         if value not in {SYNC_FULL, SYNC_LIGHT}:
             raise ValueError(f"Unknown sync mode: {value}")
         self._sync_mode = value
-
-    @property
-    def is_light_mode(self) -> bool:
-        return self.sync_mode == SYNC_LIGHT
-
-    @property
-    def is_full_mode(self) -> bool:
-        return self.sync_mode == SYNC_FULL
 
     @property
     def logfile_path(self) -> Path:
@@ -440,14 +444,89 @@ class TrinityConfig:
                 f"`PrivateKey` instance: got {type(self._nodekey)}"
             )
 
+    @contextmanager
+    def process_id_file(self, process_name: str):  # type: ignore
+        with PidFile(process_name, self.data_dir):
+            yield
+
     @classmethod
-    def from_parser_args(cls, parser_args: argparse.Namespace) -> 'TrinityConfig':
+    def from_parser_args(cls,
+                         parser_args: argparse.Namespace,
+                         app_config_types: Iterable[Type['BaseAppConfig']]) -> 'TrinityConfig':
         """
         Helper function for initializing from the namespace object produced by
         an ``argparse.ArgumentParser``
         """
         constructor_kwargs = construct_trinity_config_params(parser_args)
-        return cls(**constructor_kwargs)
+        trinity_config = cls(**constructor_kwargs)
+
+        trinity_config.initialize_app_configs(parser_args, app_config_types)
+
+        return trinity_config
+
+    def initialize_app_configs(self,
+                               parser_args: argparse.Namespace,
+                               app_config_types: Iterable[Type['BaseAppConfig']]) -> None:
+        """
+        Initialize all available ``BaseAppConfig`` based on their concrete types,
+        the ``parser_args`` and the existing ``TrinityConfig`` instance.
+        """
+        for app_config_type in app_config_types:
+            self.add_app_config(app_config_type.from_parser_args(parser_args, self))
+
+    def add_app_config(self, app_config: 'BaseAppConfig') -> None:
+        """
+        Register the given ``app_config``.
+        """
+        self._app_configs[type(app_config)] = app_config
+
+    def has_app_config(self, app_config_type: Type['BaseAppConfig']) -> bool:
+        """
+        Check if a ``BaseAppConfig`` exists based on the given ``app_config_type``.
+        """
+        return app_config_type in self._app_configs.keys()
+
+    def get_app_config(self, app_config_type: Type[TAppConfig]) -> TAppConfig:
+        """
+        Return the ``BaseAppConfig`` derived instance based on the given ``app_config_type``.
+        """
+        # We want this API to return the specific type of the app config that is requested.
+        # Our backing field only knows that it is holding `BaseAppConfig`'s but not concrete types
+        return cast(TAppConfig, self._app_configs[app_config_type])
+
+
+class BaseAppConfig(ABC):
+
+    def __init__(self, trinity_config: TrinityConfig):
+        self.trinity_config = trinity_config
+
+    @classmethod
+    @abstractmethod
+    def from_parser_args(cls,
+                         args: argparse.Namespace,
+                         trinity_config: TrinityConfig) -> 'BaseAppConfig':
+        """
+        Initialize from the namespace object produced by
+        an ``argparse.ArgumentParser`` and the ``TrinityConfig``
+        """
+        pass
+
+
+class Eth1AppConfig(BaseAppConfig):
+
+    @classmethod
+    def from_parser_args(cls,
+                         args: argparse.Namespace,
+                         trinity_config: TrinityConfig) -> 'BaseAppConfig':
+        return cls(trinity_config)
+
+    @property
+    def is_light_mode(self) -> bool:
+        return self.trinity_config.sync_mode == SYNC_LIGHT
+
+    @property
+    def is_full_mode(self) -> bool:
+        return self.trinity_config.sync_mode == SYNC_FULL
 
     @property
     def node_class(self) -> Type['Node']:
@@ -464,7 +543,11 @@ class TrinityConfig:
         else:
             raise NotImplementedError("Only full and light sync modes are supported")
 
-    @contextmanager
-    def process_id_file(self, process_name: str):  # type: ignore
-        with PidFile(process_name, self.data_dir):
-            yield
+
+class BeaconAppConfig(BaseAppConfig):
+
+    @classmethod
+    def from_parser_args(cls,
+                         args: argparse.Namespace,
+                         trinity_config: TrinityConfig) -> 'BaseAppConfig':
+        return cls(trinity_config)
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -32,6 +32,7 @@ from trinity.cli_parser import (
     subparser,
 )
 from trinity.config import (
+    BaseAppConfig,
     TrinityConfig,
 )
 from trinity.constants import (
@@ -106,7 +107,9 @@ BootFn = Callable[[
 ], None]
 
 
-def main_entry(trinity_boot: BootFn, plugins: Iterable[BasePlugin]) -> None:
+def main_entry(trinity_boot: BootFn,
+               plugins: Iterable[BasePlugin],
+               sub_configs: Iterable[Type[BaseAppConfig]]) -> None:
     event_bus = EventBus(ctx)
     main_endpoint = event_bus.create_endpoint(MAIN_EVENTBUS_ENDPOINT)
     main_endpoint.connect_no_wait()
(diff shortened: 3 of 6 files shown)


trinity/config.py
```diff
@@ -457,14 +500,13 @@ def node_class(self) -> Type['Node']:
         from trinity.nodes.full import FullNode
         from trinity.nodes.light import LightNode
 
-        if self.is_full_mode:
+        if self.base_config.is_full_mode:
```
cburgdorf: I moved the `node_class` into the `TrinityEth1SubConfig` because I personally think it is unclear at this point if the beacon node will actually end up with a concept of a `Node` class and even if we do it may be totally different from the existing one. More on that here: https://github.com/ethereum/py-evm/pull/1536#issuecomment-443317636

Maybe the same could be said about `get_chain_config` but I would defer that decision to a later point when we get our hands on that.

trinity/config.py
```diff
@@ -457,14 +500,13 @@ def node_class(self) -> Type['Node']:
...
             return FullNode
-        elif self.is_light_mode:
```
cburgdorf: I'm not sure if the sync modes itself should also move into the `TrinityEth1SubConfig`. They probably should? Thoughts?

pipermerriam: I think they should be on the `Eth1SubConfig`

trinity/config.py
```diff
@@ -440,14 +448,49 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
...
         """
         constructor_kwargs = construct_trinity_config_params(parser_args)
-        return cls(**constructor_kwargs)
+        trinity_config = cls(**constructor_kwargs)
+
+        trinity_config.initialize_sub_configs(parser_args, sub_config_types)
+
+        return trinity_config
+
+    def initialize_sub_configs(self,
```
pipermerriam: I think I'm :-1: on this API (for now while we work out the shape of this).  What do you think about the following.

```python
trinity_config = cls(**constructor_kwargs)

for app_config_type in app_config_types:
    app_config = app_config_type.from_parser_args(parser_args, trinity_config)
    trinity_config.add_app_config(app_config)
```

The code above also implicitely suggests that the `AppConfig` classes may each have their own unique `__init__` implementations but that they each *know* how to construct themselves from `parser_args`.  I like this distinction because it removes tight coupling between the `AppConfig` and the `parser_args` with-respect-to instantiation which I think is a better API to support testing.

Thoughts?

cburgdorf: I like that! :heart: :+1:

cburgdorf: The only downside of this though is that then, it is up to the `BaseSubConfig` what to do with the passed `trinity_config`. That means, there is no defined way how to access the `base_config` from the sub config.

E.g.

```python
class TrinityEth1SubConfig(BaseSubConfig):

    @property
    def node_class(self) -> Type['Node']:
        from trinity.nodes.full import FullNode
        from trinity.nodes.light import LightNode

        if self.base_config.is_full_mode:                   # <--- NEEDED HERE
            return FullNode
        elif self.base_config.is_light_mode:
            return LightNode
        else:
            raise NotImplementedError("Only full and light sync modes are supported")
```

Unless the `BaseSubConfig` dictates these as properties but even then it is more like a *suggestion* rather than a guarantee that the base config will be accessible at `base_config`.

trinity/config.py
```diff
@@ -440,14 +452,60 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
...
+
+    def get_sub_config(self, sub_config_type: Type[TSubConfig]) -> TSubConfig:
+        # We want this API to return the specific type of the sub config that is requested.
+        # Our backing field only knows that it is holding `BaseSubConfig`'s but not concrete types
+        return cast(TSubConfig, self._sub_configs[sub_config_type])
+
+
+class BaseSubConfig:
+
+    base_config: BaseApplicationConfig = None
```
cburgdorf: @pipermerriam this feels a bit weird but maybe I was misunderstanding your proposal. Now I created these here because if we don't define the `__init__` how do we standardize where to access the `base_config` from sub configs?

trinity/config.py
```diff
@@ -440,14 +452,60 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
...
+    args: argparse.Namespace = None
+
+    @classmethod
+    @abstractmethod
+    def from_parser_args(cls, args: argparse.Namespace, base_config: BaseApplicationConfig) -> 'BaseSubConfig':
+        sub_config = cls()
```
cburgdorf: This feels weird to. So, now if you want define an alternative constructor (which is the thing we want to enable by introducing the `from_parser_args` API), then you would also have to also overwrite `from_parser_args` because the default one would just call the constructor without arguments before it sets the properties.

Not sure if you had something different in mind?

trinity/config.py
cburgdorf: Ok, cool. Will move that once we've sorted the core constructor/from_parser_args issue out.

trinity/config.py
```diff
@@ -185,7 +191,10 @@ def vm_configuration(self) -> VMConfiguration:
         return self.genesis_data.vm_configuration
 
 
-class TrinityConfig:
+TSubConfig = TypeVar('TSubConfig', bound='BaseSubConfig')
+
+
+class BaseApplicationConfig:
```
pipermerriam: Ah! I see where my suggestion went sideways.

I was suggesting we **leave** the `TrinityConfig` name as is, and then call what is currently referred to as `SubConfig` an `ApplicationConfig` or `AppConfig`.  That lets the "trinity" application use the `TrinityConfig` and each other application that wants to run in trinity have their own `AppConfig`

trinity/config.py
pipermerriam: WRT to how an `AppConfig` gets access to the `TrinityConfig`, I think that's something up to the `AppConfig` to determine and I don't think it needs to be standardized.  Also, in your example above, I would expect the `is_full_mode` and `is_light_mode` to live in the `Eth1AppConfig` class.

trinity/config.py
pipermerriam: I think that each `AppConfig` **must** implement their own `from_parser_args` and that this function body should just be a `pass` statement.  The individual app configs should follow the same structure as the `TrinityConfig` to convert the parser args into a set of constructor args and then construct the class.  In theory we can generalize this if we had an expectation for a `classmethod` or `staticmethod` that did the conversion from parser args to constructor args which would allow this method to be standardized.

cburgdorf: Ok, but even then we are still shifting the responsibility of where to store the `base_config` and `args` references into the sub classes. Unless, we do what I did in this PR and at least define `base_config` and `args` as properties as *suggestions*. But still, since each subclass would be free to just bypass this, there would be no guarantee that from every sub config you can access the base config at `base_config`.

trinity/config.py
cburgdorf: Ah, I just made a comment before reading this one.

>I think that's something up to the AppConfig to determine and I don't think it needs to be standardized

Mmh...ok. I think I would probably favor the initial PR design where we do end up with a standardized access to `base_config` and `args` in each `AppConfig` and rather ditch the "let them be free with the `__init__` method idea but I don't feel strongly about it.

It just strikes me as something annoying to be forced to rely on conventions with this, when we could rely on the code itself. Your call.

trinity/config.py
cburgdorf: Gotcha. Changed the naming. Hope I got it right this time :)

trinity/config.py
```diff
@@ -440,14 +444,77 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
...
+
+    @classmethod
+    @abstractmethod
+    def from_parser_args(cls,
+                         args: argparse.Namespace,
+                         base_config: TrinityConfig) -> 'BaseAppConfig':
+        pass
+
+
+class DefaultBaseAppConfig(BaseAppConfig):
```
cburgdorf: @pipermerriam would this be an acceptable compromise? From the perspective of the `TrinityConfig` there exists only `BaseAppConfig` as an abstraction and it only dictates the `from_parser_args` class method.

However, additionally we introduce a `DefaultBaseAppConfig` as a convenience abstraction that dictates the `args` / `base_config` convention. So we maintain the full flexibility from the perspective of the `TrinityConfig` but standardize the common scheme.

trinity/config.py
```diff
@@ -440,14 +444,77 @@ def nodekey(self, value: Union[bytes, PrivateKey]) -> None:
...
+
+class Eth1AppConfig(DefaultBaseAppConfig):
+
+    @property
+    def is_light_mode(self) -> bool:
+        return self.base_config.sync_mode == SYNC_LIGHT
+
+    @property
+    def is_full_mode(self) -> bool:
+        return self.base_config.sync_mode == SYNC_FULL
```
cburgdorf: @pipermerriam One could argue that `sync_mode` should move into the `Eth1AppConfig` as well but the problem is that this spreads like wild fire. Because the `database_dir` depends on the `sync_mode` but I don't think we want to move the `database_dir` as it really is a core thing that will be needed for any client that trinity would support.

So, my current syncing is, let `sync_mode` continue to be a `str` in `TrinityConfig` but leave it up to the specific app config to interpret that string (there's currently a hard validation against `full` / `light` but it we could make that more flexible)

The bigger problem really is that everything in our current scheme is modeled around the idea that we are always talking about an Ethereum 1 implementation.

```
trinity/<network>/chain/<sync-mode>
trinity/<network>/logs
...
```

I guess we don't want to abuse the network identifier for the purpose of supporting the beacon node ;)

So, it seems to boil down to two options:

1. Add another top level hierarchy `trinity/<client|protocol|whatever>/<network>/chain/<sync-mode>`

2. Use an entirely different `trinity_root_dir` so that it becomes `trinity-beacon/<network>/chain/<sync-mode>`

I'm in favor of the second option for multiple reasons:

A.) It is backwards compatible. The current trinity does not need to migrate its dirs

B.) It feels natural **if** we continue to treat the beacon node as an entirely different executable (it currently is `trinity-beacon`.

I might go ahead and give B a shot to see where that falls apart :)

cburgdorf: Went ahead with that idea and added a commit on top https://github.com/ethereum/py-evm/pull/1566/commits/2df39dbfd6eddcb74b8d330c9651d57fb6f0e63c

To recap:

- we use a different root path per application
- the `TrinityConfig` only holds what is anticipated to be used across all applications (`SYNC_MODE` is debatable but since it influences very low level things like `database_dir` I feel we should just keep that a core config) 
- everything else goes into additionally registered sub configs

tests/trinity/core/configuration/test_trinity_config_object.py
```diff
@@ -90,3 +91,12 @@ def test_trinity_config_explictely_provided_nodekey(nodekey_bytes, as_bytes):
     )
 
     assert trinity_config.nodekey.to_bytes() == nodekey_bytes
+
+
+def test_trinity_config_sub_configs():
+    trinity_config = TrinityConfig(network_id=1)
+    trinity_config.initialize_app_configs(None, (BeaconAppConfig,))
+
+    assert trinity_config.has_app_config(BeaconAppConfig)
```
pipermerriam: Curious what others think about this.  I've often done `assert thing is True` when I am actually testing a boolean value to ensure that some other "truthy" value cannot slip through.

trinity/config.py
pipermerriam: I'm leaning towards :-1: 

- I don't see a clear benefit (if it isn't truly standardized then it's not standard)
- it's an attempt at establishing an abstraction when I don't think we really *know* what the abstraction looks like...
- it results in the `__init__` being tied to an `argparse.Namespace` shaped object which likely makes testing harder.

Thoughts?

trinity/config.py
pipermerriam: What about:

- `trinity/mainnet/eth1/<sync-mode>`
- `trinity/mainnet/beacon/...'
- 'trinity/mainnet/shard-01/...'

It seems beneficial to co-locate all of the various *chains* under a single root directory for the network they are part of.

pipermerriam: Regarding `database_dir`.

I think this is actually an `eth1` specific thing and **should-not** be shared by any of the other applications (beacon/shard), as each of them should have their own databases.

trinity/config.py
cburgdorf: >I don't see a clear benefit

The benefit would be not having to implement `from_parser_args` and `__init__` in every `AppConfig`

>it's an attempt at establishing an abstraction when I don't think we really know what the abstraction looks like...

Mmmh...I think most `AppConfig` implementations will just take the `args` and the `trinity_config` in their `__init__` anyway or what are the cases where you see them taking other things (especially considering, that they all origin in a top level `from_parser_args` call from the `TrinityConfig`

>t results in the `__init__` being tied to an `argparse.Namespace` shaped object which likely makes testing harder.

Well, without this abstraction, we will just have the exact same `__init__` with ` argsparse.Namespace` on every implementation of `AppConfig`

pipermerriam: > I think most `AppConfig` implementations will just take the `args` and the `trinity_config` in their `__init__` anyway

I think this suggests otherwise:

https://github.com/ethereum/py-evm/blob/751c856953060d68ae7be2c62306d11935e32345/trinity/utils/chains.py#L123-L163

In the case of  `__init__(args, trinity_config)` the result is that the `__init__` method must parse and normalize the values from the `argparse.Namespace` object which I would argue is a bad coupling,  binding it to the CLI

In the case of `__init__(trinity_config, arg_1, arg_2, ...)` and `Config.from_parser_args(args, trinity_config)` we still have *coupling* but it is looser and keeps the `cli_args -> init_args` conversion from bleeding into the actual configuration objects implementation.

A **lot** of my reasoning is centered around how the `Config` classes are tested so admittedly this doesn't have much effect on the end result as far as production code goes.

I'm :+1: on whatever you decide.

trinity/config.py
cburgdorf: >or the network they are part of.

Can we really say that the beacon node and the Ethereum are in the same network? All the different shards that a beacon node would download could sure all live under one `trinity-beacon/mainnet/shard-0|shard-1|...` space but since shards do not exist in Ethereum 1 it feels natural to me to just use `/<network>/<beacon|shard-n>` as a new schema for the beacon node.

I think the two primary reasons why I'd be in favor of a different `<root-dir>` rather than changing the existing schema are:

1. It currently looks like `trinity-beacon` will be a different executable, a different program that just builds on a shared code base. So in that sense, a different root dir makes most sense to me.

2. Changing the existing schema is a hard breaking change (sure we could detect and migrate existing data directories but still)

trinity/config.py
cburgdorf: >A lot of my reasoning is centered around how the Config classes are tested so admittedly this doesn't have much effect on the end result as far as production code goes.

Ah, I missed that goal to spread out the `args` into individual parameters for better testing. That sounds reasonable.

I would however, still favor to at least provide an `__init__(trinity_config: TrinityConfig)__` and have child classes overwrite and extend that. Not having the `trinity_config` access standardized just doesn't sit so well with me.

trinity/config.py
cburgdorf: I will move this into another PR so that we can discuss this separately. I'd also be interested to hear what @carver thinks about it. Something in me really likes to treat the beacon node as a separate program here. E.g what about the `NodeKey` would the regular trinity node share that with the beacon node? It seems cleaner to me to keep the root directories of each of them clearly separated..also in terms of logs etc.

carver: > Plus, isn't the litmus test test that if we ever would decide to move the beacon client into its entirely own project (instead of just maintaining a separate entry point / bootstrapping in the same code base) then we would for sure use a different root directory anyway? So, this would make a potential migrating easier long term by treating them as two different programs right from the start.

This seems like a fairly strong argument. If it runs as a separate project, maybe it belongs in its own root dir.

OTOH: *maybe* `trinity` becomes the all-in-one command, and what we currently call `trinity` moves to `trinity-eth1` or something. Then implicitly `trinity` runs both `trinity-beacon` and `trinity-eth1`. So maybe the directory structure is something like `trinity/beacon-mainnet/...` and `trinity/eth1-mainnet/...`.

Can you link the other issue/PR to move the discussion to?

cburgdorf: Yep, it's in this PR (that proposes a solution) https://github.com/ethereum/py-evm/pull/1603

PR #1567, opened 2018-12-10
Upgrade Lahja and make requests less wasteful

cburgdorf:
### What was wrong?

Prior to this change, whenever we made a `request(SomeEvent())` on the `EventBus`, the event traveled out to every connected endpoint(read: different processes) and waited for *someone* to respond.

This is great in a way because it really embraces loose coupling. However, for many request we really know that there should only be one single party to receive the request and give us a response. Spreading these requests across each and every connected process seems like an unnecessary waste of resources. 

For some critical events it could even be considered a security issue as someone could trick people into installing a plugin which listens for specific events and answers them in a malicious way.

### How was it fixed?

With the latest Lahja release the `request` API accepts a `BroadcastConfig` (similar to the `broadcast` API) which is being used to restrict the number of endpoints that the request goes out to to the single one of them that should be receiving the request.

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,7 @@ deps = {
         "ipython>=6.2.1,<7.0.0",
         "plyvel==1.0.5",
         "web3==4.4.1",
-        "lahja==0.9.0",
+        "lahja==0.10.0",
         "termcolor>=1.1.0,<2.0.0",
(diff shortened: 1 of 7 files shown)


cburgdorf:
Ah..forgot the tests. Will look at those tomorrow.

PR #1569, opened 2018-12-10
Peer discovery on high-latency connection

carver:
### What was wrong?

Wasn't connecting to nodes on bad wifi

### How was it fixed?

Add some more logging info to discovery, and increase the general timeout (which ping/pong uses).

#### Cute Animal Picture


pipermerriam:
I wonder if this is another realm where we could use percentile + stddev to tune the ping timeout to user network conditions so that we still get prompt timeouts in the case that something is *truly* taking too long.

Issue #1562, opened 2018-12-08
test_blockchain_fixtures() works poorly


veox:
* py-evm Version: current `master` (commit b6865d0d615f2eb6d53cdf514e1d8bd2e62f89d2)
* OS: Arch Linux
* Python Version (python --version): 3.7.1
* Environment (output of `pip freeze`): MISSING (sorry!..)

### What is wrong?

In testing for PR #1181 - which spawned PR #1560, and https://github.com/cburgdorf/py-evm/pull/2 in particular - I've done the following ([quote from gitter](https://gitter.im/ethereum/py-evm?at=5c0aee738b656e2b04dc221c)):

> (...) there's a very handy `verify_account_db()` invocation in `test_blockchain_fixtures()` of `tests/json-fixtures/test_blockchain.py`, which - with a little bit of commenting-out of asserts and checks through a couple files - can be used to easily compare the difference between the upstream test and `py-evm's` result.
> 
> Normally, `GeneralStateTests` are skipped in these test runs, to avoid duplication with `tests/json-fixtures/test_state.py`. However, the latter tests are less detailed in their JSON specification - so the comparison trick can't be used.
> 
> Using this, [along with increasing the logging level to `5` (trace), ] helps circumvent some guesswork [in debugging externally-provided tests].

[The test function](https://github.com/ethereum/py-evm/blob/b6865d0d615f2eb6d53cdf514e1d8bd2e62f89d2/tests/json-fixtures/test_blockchain.py#L92), however, works "poorly" in general, since it applies `verify_account_db()` unconditionally, and only after it has already applied the blocks (specified in the fixture) that have produced this account-DB-to-be-checked, and has compared the produced blocks (with those in the fixture).

In other words:

> (...) there's no sense in doing these fine-grained state comparisons when [the state root in] the block header already matches. It just slows down the test even more.

Also, the `assert` in

https://github.com/ethereum/py-evm/blob/b6865d0d615f2eb6d53cdf514e1d8bd2e62f89d2/tests/json-fixtures/test_blockchain.py#L123-L125

is never hit, because `apply_fixture_block_to_chain()` has a very similar nested validity check that gets hit in the same conditions - shadowing the `assert`, ~~and terrifying the poor sob debugging this with a single `cytoolz`-/`functools`-obfuscated generic function (aren't these supposed to go on _forever, all the way down_?..)~~.

### How can it be fixed

Not sure about the latter issue, and what "fixing" it might break. (Didn't do a `git blame`.)

[...]

PR #1573, opened 2018-12-11
Failing blockchain tests now emit a state diff

lithp:
Intended to address #1562 

    - Instead of failing with "these block hashes do not match", blockchain
      tests now fail with "this part of the state tree does not match".
    
    - The tests were inefficient, they first checked that the state hash
      matched and then checked that the states matched. It's unlikely the
      states will ever not match!
    
    - Some validations needed to be deferred until after state-tree
      comparison, or else they would fail the test early with an unhelpful
      message.

### What was wrong?

Failures used to look like this:

```
E       eth_utils.exceptions.ValidationError: Mismatch between block and mined block on 3 fields:
E        - header.state_root  :
E           (actual)  : b"R\xafB',,\xd3\xcb\x8b#W\x98)wL\xc4fLm\x90m\x08\xbb\xed\xc0\x03\x8e\x7f\xfcO:X"
E           (expected): b"pv\xa0\x01\xd5\xc2\xab]\x94\xdbj\xcac\x17\x84\x1fW2QZ\xf7\x1f>:\x00\xb3\x8f\xe2'Sy\xfc"
E        - header.receipt_root:
E           (actual)  : b'\n\x106\x18\xe1\xe1)0\xf1%\xef2\x94\xd9\x0fQ\xed\x12\x1d+\x19\xae\xd7\t\xcf\x92\x8c4]\xd9zy'
E           (expected): b'\x0e\xf4\xbaL\x0be\xce\xcau\xb1\xb6~i\xf5\xaf\xb1\xabd\xe7_~\x96!`m\xe5\x17M\xce\x9eD\x05'
E        - header.gas_used    :
E           (actual)  : 128058
E           (expected): 113058
```

And now they look like this:

```
E           AssertionError: State DB did not match expected state on 3 values:
E           0x1000000000000000000000000000000000000000(storage[1]) | Actual: 859287 | Expected: 849245
E            - 0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba(balance) | Actual: 2000000000000160716 | Expected: 2000000000000170758 | Delta: 10042
E            - 0xa94f5374fce5edbc8e2a8697c15331677e6ebf0b(balance) | Actual: 999999999999739284 | Expected: 999999999999729242 | Delta: -10042
```

### How was it fixed?

I'm running with all the hard work @veox did, this is an attempt to put it into a more-committable form.

#### Cute Animal Picture

a-fluffy-cat-looking-funny-surprised-or-concerned

### DIFF
--- a/eth/tools/fixtures/helpers.py
+++ b/eth/tools/fixtures/helpers.py
@@ -193,8 +193,10 @@ def new_chain_from_fixture(fixture: Dict[str, Any],
     )
 
 
-def apply_fixture_block_to_chain(block_fixture: Dict[str, Any],
-                                 chain: BaseChain) -> Tuple[BaseBlock, BaseBlock, BaseBlock]:
+def apply_fixture_block_to_chain(
+        block_fixture: Dict[str, Any],
+        chain: BaseChain,
+        perform_validation: bool=True) -> Tuple[BaseBlock, BaseBlock, BaseBlock]:
     '''
     :return: (premined_block, mined_block, rlp_encoded_mined_block)
     '''
@@ -209,7 +211,7 @@ def apply_fixture_block_to_chain(block_fixture: Dict[str, Any],
 
     block = rlp.decode(block_fixture['rlp'], sedes=block_class)
 
-    mined_block, _, _ = chain.import_block(block)
+    mined_block, _, _ = chain.import_block(block, perform_validation=perform_validation)
 
     rlp_encoded_mined_block = rlp.encode(mined_block, sedes=block_class)
 
--- a/tests/json-fixtures/test_blockchain.py
+++ b/tests/json-fixtures/test_blockchain.py
@@ -111,8 +111,10 @@ def test_blockchain_fixtures(fixture_data, fixture):
     # 2 - loop over blocks:
     #     - apply transactions
     #     - mine block
-    # 4 - profit!!
+    # 3 - diff resulting state with expected state
+    # 4 - check that all previous blocks were valid
 
+    mined_blocks = list()
     for block_fixture in fixture['blocks']:
         should_be_good_block = 'blockHeader' in block_fixture
 
@@ -121,8 +123,12 @@ def test_blockchain_fixtures(fixture_data, fixture):
             continue
 
         if should_be_good_block:
-            (block, mined_block, block_rlp) = apply_fixture_block_to_chain(block_fixture, chain)
-            assert_mined_block_unchanged(block, mined_block)
+            (block, mined_block, block_rlp) = apply_fixture_block_to_chain(
+                block_fixture,
+                chain,
+                perform_validation=False  # we manually validate below
+            )
+            mined_blocks.append((block, mined_block))
         else:
             try:
                 apply_fixture_block_to_chain(block_fixture, chain)
@@ -133,6 +139,10 @@ def test_blockchain_fixtures(fixture_data, fixture):
                 raise AssertionError("Block should have caused a validation error")
 
     latest_block_hash = chain.get_canonical_block_by_number(chain.get_block().number - 1).hash
-    assert latest_block_hash == fixture['lastblockhash']
+    if latest_block_hash != fixture['lastblockhash']:
+        verify_account_db(fixture['postState'], chain.get_vm().state.account_db)
+        assert False, 'the state must be different if the hashes are'
 
-    verify_account_db(fixture['postState'], chain.get_vm().state.account_db)
+    for block, mined_block in mined_blocks:
+        assert_mined_block_unchanged(block, mined_block)
+        chain.validate_block(block)


lithp:
These test failures happened because `Slave 'gwXX' crashed while running`, when I run each of these failing tests locally they work, which makes me thing it might be something transient, maybe a timeout? I don't have permissions to rerun the build though.

carver:
> It may mean the tests need to be split into smaller chunks.

Even if it's not the cause of the crash, it would be best to split the tests.

Note how we had split the state tests into 5 parts, in order to get some balance on the length of the test jobs. With this PR, all 5 of those, and the blockchain tests, are combined into a single job.

See how they were split before here:
https://github.com/ethereum/py-evm/blob/f6162987c03ae6f6c52724061b2ad80710faf694/tox.ini#L38-L45

(which may or may not help)

eth/tools/fixtures/helpers.py
```diff
@@ -40,6 +40,7 @@
     BaseVM,
 )
 from eth.vm.forks import (
+    ConstantinopleVM,
```
veox: Some of these changes - extracted from [this gist](https://gist.github.com/veox/5874314b11cb731e69256c0bd451c88c), via issue #1562 - are already pending inclusion as part of https://github.com/cburgdorf/py-evm/pull/2.

Since you're not including the portion that enables more `Constantinople` tests, this change (and the other one from same file) would be unrelated to the rest of this PR.

carver:
> it looks like some of the SLOWEST_TESTS are so slow they can't even be run in circleci?

Yeah, or maybe too memory-intensive? Haven't had a chance to dig in.

> Don't know what your development process looks like, I think what I'll do is turn the `BlockchainTests/GeneralStateTests` off again, and enable them in a different PR so this one can stay focused?

Yeah, that sounds right. The more focused the better. Probably just rebase away the state test changes here, and move those to another PR.

> I think the next step here is to copy over some of the tools state tests have:

Yup, all three LGTM.

> Comments on #1562 would be appreciated! I think it would make the test a lot easier to debug if state tests were dropped in favor of their BlockchainTest equivalents.

Which open questions on 1562 are you looking for feedback on? On my quick scan, I didn't see any glaring unsolved problems.

eth/tools/fixtures/helpers.py
lithp: You're right, removed the Constantinople parts.

veox:
> > it looks like some of the SLOWEST_TESTS are so slow they can't even be run in circleci?
> 
> Yeah, or maybe too memory-intensive? Haven't had a chance to dig in.

Anecdotal: a local test run (still on-going) is currently consuming 3 GiB.

lithp:
>

lithp:
Accidental close :cry:

> Which open questions on 1562 are you looking for feedback on? On my quick scan, I didn't see any glaring unsolved problems.

The one I'm concerned about is whether it's a good idea to drop the state tests and instead use the one embedded within the blockchain tests

lithp:
The tests are all passing but:

> GitHub rate limited us while adding a commit status of success: You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.

veox:
While you're at it, could you also reverse the calculation on this line?.. (I've [mentioned this in passing elsewhere](https://github.com/ethereum/py-evm/pull/1181#issuecomment-445986061).)

https://github.com/ethereum/py-evm/pull/1573/files#diff-38aa0bc8e408266959b43002a08f0cd4R81

For example, as things stand, if a fixture-expected value is 1000, and the actual produced by `py-evm` is 500, the printed message will show a delta of 500 - that is, a _net positive_, whereas what truly happened is `py-evm` having a _value that's too low_. This is confusing: say, a miner's account balance showing "Delta: 500" makes one think that the miner received 500 wei more than expected (which is not true).

If not, I can open a separate PR.

carver:
> The tests are all passing but:
> 
> > GitHub rate limited us while adding a commit status of success: You have triggered an abuse detection mechanism. Please wait a few minutes before you try again.

Yeah, that issue is a recent-ish, known, and annoying.

veox:
Blame Microsoft.

carver:
> While you're at it, could you also reverse the calculation on this line?.. (I've [mentioned this in passing elsewhere](https://github.com/ethereum/py-evm/pull/1181#issuecomment-445986061).)
> 
> https://github.com/ethereum/py-evm/pull/1573/files#diff-38aa0bc8e408266959b43002a08f0cd4R81

I was juuust about to merge this, and it would be nice to rebase #1577 on this. Maybe @lithp can sneak the change into #1577

lithp:
> While you're at it, could you also reverse the calculation on this line?
> Maybe @lithp can sneak the change into #1577

I didn't want to sneak an unrelated change into #1577 but made a tiny PR for it: https://github.com/ethereum/py-evm/pull/1580

PR #1574, opened 2018-12-11
Beacon block request handler

jannikluhn:
Implements handling of and responding to beacon chain `GetBeaconBlocks` messages.

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/core/p2p-proto/bcc/helpers.py
+++ b/tests/trinity/core/p2p-proto/bcc/helpers.py
@@ -1,3 +1,5 @@
+import asyncio
+
 from cancel_token import CancelToken
 
 from eth.db.atomic import AtomicDB
@@ -98,3 +100,23 @@ async def get_directly_linked_peers(request, event_loop, alice_chain_db=None, bo
         alice_factory=alice_factory,
         bob_factory=bob_factory,
     )
+
+
+async def get_directly_linked_peers_in_peer_pools(request, event_loop, chain_db=None):
+    alice, bob = await get_directly_linked_peers(request, event_loop, bob_chain_db=chain_db)
+    alice_peer_pool = BCCPeerPool(alice.privkey, alice.context)
+    bob_peer_pool = BCCPeerPool(bob.privkey, bob.context)
+
+    asyncio.ensure_future(alice_peer_pool.run())
+    asyncio.ensure_future(bob_peer_pool.run())
+
+    def finalizer():
+        event_loop.run_until_complete(alice_peer_pool.cancel())
+        event_loop.run_until_complete(bob_peer_pool.cancel())
+
+    request.addfinalizer(finalizer)
+
+    alice_peer_pool._add_peer(alice, [])
+    bob_peer_pool._add_peer(bob, [])
+
+    return alice, alice_peer_pool, bob, bob_peer_pool
--- a/tests/trinity/core/p2p-proto/bcc/helpers.py
+++ b/tests/trinity/core/p2p-proto/bcc/helpers.py
@@ -1,13 +1,12 @@
 import asyncio
 
-from cytoolz import (
-    merge,
-)
-
 from cancel_token import CancelToken
 
 from eth_utils import (
     to_tuple,
+)
+from eth_utils.toolz import (
+    merge,
 )
 
 from eth.db.atomic import AtomicDB
--- a/tests/trinity/core/p2p-proto/bcc/helpers.py
+++ b/tests/trinity/core/p2p-proto/bcc/helpers.py
@@ -1,0 +1,100 @@
+from cancel_token import CancelToken
+
+from eth.db.atomic import AtomicDB
+from eth.beacon.db.chain import BeaconChainDB
+from eth.beacon.types.blocks import BaseBeaconBlock
+from eth.constants import (
+    ZERO_HASH32,
+)
+
+from trinity.protocol.bcc.context import BeaconContext
+from trinity.protocol.bcc.peer import (
+    BCCPeerFactory,
+)
+
+from p2p import ecies
+from p2p.tools.paragon.helpers import (
+    get_directly_linked_peers_without_handshake as _get_directly_linked_peers_without_handshake,
+    get_directly_linked_peers as _get_directly_linked_peers,
+)
+
+
+def get_fresh_chain_db():
+    db = AtomicDB()
+    genesis_block = BaseBeaconBlock(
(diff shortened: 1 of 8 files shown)


jannikluhn:
Note that in the last commit, I made a minor change to the `BaseRequestServer`: I changed the type hint for the peer pool from `BaseChainPeerPool` to `BasePeerPool` so that it works with the `BCCPeerPool` (which is not a `BaseChainPeerPool` because `BCCPeer` is not a `BaseChainPeer` because `BaseChainPeer` has some PoW specific details).

trinity/protocol/bcc/servers.py
```diff
@@ -0,0 +1,126 @@
...
+from eth.beacon.types.blocks import BaseBeaconBlock
+
+from trinity.protocol.common.servers import BaseRequestServer
+from trinity.protocol.bcc import commands
+from trinity.protocol.bcc.peer import (
+    BCCPeer,
+    BCCPeerPool,
+)
+
+from typing import (
```
cburgdorf: nitpick but this should move under itertools as it is part of the standardlib.

Ordering should be:

- standard libs
- installed libs
- local

See also: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#overall-ordering

trinity/protocol/bcc/servers.py
```diff
@@ -0,0 +1,126 @@
...
+    cast,
+    Any,
+    Dict,
+    Iterable,
+    Set,
+    Type,
+)
+from eth_typing import (
```
cburgdorf: And this should move between `cytoolz` and `eth_utils` (installed libs)

trinity/protocol/bcc/servers.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if isinstance(block_slot_or_hash, int):
+            get_start_block = functools.partial(
+                self.db.get_canonical_block_by_slot,
+                cast(int, block_slot_or_hash),
+            )
+        elif isinstance(block_slot_or_hash, bytes):
+            get_start_block = functools.partial(
+                self.db.get_block_by_hash,
+                cast(Hash32, block_slot_or_hash),
```
cburgdorf: In general `Hash32(block_slot_or_hash)` is to be preferred as this checks the type and e.g. `Hash32("foo")` will be rejected by mypy. In contrast nonsense such as `cast(Hash32, "foo")` is allowed as a cast will accept anything you feed it.

However, in this specific case, unfortunately the type hints don't have any affect at all. That is because `functool.partial` erases any type hints, everything becomes `Any` and we are left without any type safety. E.g. you can try replacing `cast(Hash32, block_slot_or_hash)` with `"foo"` and it will go through even though `get_block_by_hash` should not accept a string.

My preference would be to rewrite this part in a less dynamic way (reject wrong types, then execute one or the other)

tests/trinity/core/p2p-proto/bcc/helpers.py
```diff
@@ -0,0 +1,156 @@
+import asyncio
+
+from cytoolz import (
```
pipermerriam: In an attempt to maintain good pypy support (though we still don't actually support pypy) we've shifted to importing these from `eth_utils.toolz` which appropriately imports from `cytoolz` if it is installed or `toolz` otherwise.

trinity/protocol/bcc/servers.py
```diff
@@ -0,0 +1,124 @@
...
+    async def _handle_msg(self, base_peer: BasePeer, cmd: Command,
+                          msg: protocol._DecodedMsgType) -> None:
+        peer = cast(BCCPeer, base_peer)
+
+        if isinstance(cmd, commands.GetBeaconBlocks):
+            await self._handle_get_beacon_blocks(peer, cast(Dict[str, Any], msg))
+        else:
+            raise Exception("Invariant: Only subscribed to GetBeaconBlocks")
+
+    async def _handle_get_beacon_blocks(self, peer: BCCPeer, msg: Dict[str, Any]) -> None:
```
pipermerriam: Can you use a `mypy_extensions.TypedDict` for the `msg` to give us stronger typing and remove the need for the `cast` call.

trinity/protocol/bcc/servers.py
```diff
@@ -0,0 +1,124 @@
...
+            blocks = ()
+
+        self.logger.debug2("Replying to %s with %d blocks", peer, len(blocks))
+        peer.sub_proto.send_blocks(blocks)
+
+    @to_tuple
+    def _get_blocks(self,
+                    start_block: BaseBeaconBlock,
+                    max_blocks: int) -> Iterable[BaseBeaconBlock]:
+        if max_blocks <= 0:
```
pipermerriam: Seems like a negative `max_blocks` should be an error case.

trinity/protocol/bcc/servers.py
jannikluhn: Nice, didn't know about `TypeDict`, seems pretty useful! I added it, but here it didn't help a lot unfortunately: It only got rid of the cast for `max_blocks`, we still need to first cast the message to the `TypeDict`.

jannikluhn:
> Also noticed there is no skip parameter for the GetBeaconBlocks command, nor a reverse. Is this intentional?

Yes, I'd like to keep everything as simple as possible for as long as possible, and neither `reverse` nor `skip` seem particular important for an MVP testnet. If it turns out that we need it, we can always add it later.

>  I would highly recommend adding a request_id parameter to any of the GetThing/Thing command pairs so that clients can have strong linkage between a request/response.

Basically left this out for simplicity as well. But I agree that request ids are most likely going to be quite helpful. So I'll probably going to add them in the course of beacon chain syncing which I'm going to work on next.

trinity/protocol/bcc/servers.py
```diff
@@ -0,0 +1,126 @@
...
+                 token: CancelToken = None) -> None:
+        super().__init__(peer_pool, token)
+        self.db = db
+
+    async def _handle_msg(self, base_peer: BasePeer, cmd: Command,
+                          msg: protocol._DecodedMsgType) -> None:
+        peer = cast(BCCPeer, base_peer)
+
+        if isinstance(cmd, GetBeaconBlocks):
+            await self._handle_get_beacon_blocks(peer, cast(GetBeaconBlocksMessage, msg))
```
pipermerriam: I'm *much* more comfortable with this type of `cast` than with the other so :+1:

PR #1577, opened 2018-12-11
Use the state tests embedded in the blockchain tests

lithp:
When state tests fail they tell you the root state hash has changed, but there's no easy way of telling which parts of the state have changed. However, a copy of the state tests is *also* available under the blockchain tests, and that copy includes the full expected state tree.

This PR changes which version of the state tests are run. In combination with #1573 this should make diagnosing test failures much easier. 

- test_state.py skipped some tests which take too long to run, those changes are now reproduced in test_blockchain.py
- test_blockchain.py now also supports the `--fork` flag, to allow running just a subset of the tests
- the tox commands have been updated to only run blockchain tests, not the state tests.
- circleci has been updated in-step with tox
- `SLOWEST_TESTS` has been regenerated. There are now more tests (each of the indexes appears as a separate blockchain test) so the list is a little longer now. The 60ish tests skipped here (out of 10851) account for 15 of the 21 minutes it takes to run the test suite on my machine.

#### Cute Animal Picture

kitten


tests/json-fixtures/test_blockchain.py
```diff
@@ -34,6 +36,84 @@
 BASE_FIXTURE_PATH = os.path.join(ROOT_PROJECT_DIR, 'fixtures', 'BlockchainTests')
 
 
+# These are the slowest tests from the full blockchain test run. This list
+# should be regenerated occasionally using `--durations 100`.
+SLOWEST_TESTS = {
```
cburgdorf: Not for this PR but just as an idea: Isn't there something in Circle CI that we can differentiate between CI runs for PRs and CI runs on master? I think it would be nice if we could at least achieve that all these tests run from time to time. So that, even if we don't get immediate feedback that our PR broke one of these, we at least get notified about it some time after.

tox.ini
```diff
@@ -1,10 +1,10 @@
 [tox]
 envlist=
-    py{35,36}-{core,database,transactions,vm,native-blockchain}
```
cburgdorf: I'm still a bit confused why the `GeneralStateTests` also exists outside of the `BlockchainTest` and if there is something we might be forgetting about dropping those. Do you know why they exist in the first place? Did you check if the number of tests that are run still matches after this change?

.circleci/config.yml
```diff
@@ -339,17 +333,16 @@ workflows:
       - py37-trinity-integration
       - py37-beacon
 
-      - py36-native-state-byzantium
-      - py36-native-state-frontier
-      - py36-native-state-homestead
-      - py36-native-state-eip150
-      - py36-native-state-eip158
```
carver: These are still state tests, they just happen to be in a different folder. I think the job name could stay the same.

tox.ini
veox: Upstream docs:

https://ethereum-tests.readthedocs.io/en/latest/test_types/state_tests.html
https://ethereum-tests.readthedocs.io/en/latest/test_types/blockchain_tests.html

So, you should be confused as to why `GeneralStateTests` exist inside `BlockchainTests`, not the other way around. ^_^

Anyway, according to the docs, all of the former should eventually make their way into the latter:

https://ethereum-tests.readthedocs.io/en/latest/generating-tests.html#advanced-converting-a-generalstatetest-case-into-a-blockchaintest-case

veox:
@carver Good question. You mean like this one?..

```
pytest -k 'test_blockchain_fixtures[/home/veox/src/py-evm/fixtures/BlockchainTests/bcUncleTest/uncleWithSameBlockNumber.json:uncleWithSameBlockNumber_Homestead]'
```

They are in the same submodule that gets pulled in as `fixtures`.

(I haven't tested this particular PR yet, and whether it has inadvertently removed something.)

tests/json-fixtures/test_blockchain.py
lithp: It's possible with Travis and I would be surprised if it wasn't also possible with Circle CI. It could be nice to have a nightly build which takes master and runs all the tests, and maybe even does benchmarking?

tox.ini
```diff
@@ -35,14 +35,13 @@ commands=
     rpc-state-quadratic: pytest {posargs:tests/trinity/json-fixtures-over-rpc/test_rpc_fixtures.py -k 'GeneralStateTests and stQuadraticComplexityTest'}
     transactions: pytest {posargs:tests/json-fixtures/test_transactions.py}
     vm: pytest {posargs:tests/json-fixtures/test_virtual_machine.py}
-    native-blockchain: pytest {posargs:tests/json-fixtures/test_blockchain.py}
```
carver: > Good question. You mean like this one?..
> 
> ```
> pytest -k 'test_blockchain_fixtures[/home/veox/src/py-evm/fixtures/BlockchainTests/bcUncleTest/uncleWithSameBlockNumber.json:uncleWithSameBlockNumber_Homestead]'
> ```
> They are in the same submodule that gets pulled in as `fixtures`.
> 
> (I haven't tested this particular PR yet, and whether it has inadvertently removed something.)

@veox right, the complete removal of this test (non-state blockchain tests) is what threw me off. Just wanted to make sure we had confirmation that nothing got dropped in the refactor.

tests/json-fixtures/test_blockchain.py
pipermerriam: I dug around a bit and couldn't find anything in the CircleCI interface to do this

.circleci/config.yml
lithp: They're blockchain tests now! Blockchain tests which happen to also contain state tests. The command for this specific one is now: 

> native-blockchain-eip158: pytest {posargs:tests/json-fixtures/test_blockchain.py --fork EIP158}

tox.ini
lithp: The only way we would be missing some tests would be if this collection of forks doesn't cover the full set of blockchain tests: `frontier, homestead, eip150, eip158, byzantium, constantinople, metropolis`. Let me think of a way to test whether that's true.

lithp: Well shucks, looks like we're missing some tests:

```
(py-evm-ojdIf4Yr) brian@rhythm:~/ef/py-evm$ pytest tests/json-fixtures/test_blockchain.py --collect-only 2>/dev/null | grep -oP '\[\K.*(?=\])' | cut -d':' -f3 | sort | uniq
Byzantium
Constantinople
EIP150
EIP158
EIP158ToByzantiumAt5
Frontier
FrontierToHomesteadAt5
Homestead
HomesteadToDaoAt5
HomesteadToEIP150At5
```

lithp: Okay, thanks for catching this Carver! I've pushed [a commit](https://github.com/ethereum/py-evm/pull/1577/commits/9d28e837fb7722ab46858342fbd22024cf088cc1) which adds the missing tests.

.circleci/config.yml
pipermerriam: Somewhat orthogonal but it would be good to rename the `EIPXXX` to their long form fork names.

lithp: What are their long form fork names? The EIP's page just [calls it 158](https://eips.ethereum.org/EIPS/eip-158) and doesn't say when it was scheduled.

lithp: 150 is apparently Tangerine Whistle and yeah, that name is 200% more exciting than EIP-150 :grinning:

lithp: Do you want me to do that in this PR?

pipermerriam: I think "Spurious Dragon" is the other one.  Look in `eth.vm.forks` and you can find them (and there should be things within each module to hint at which EIPXXX they correspond to.

tox.ini
carver: Haha, well I pretty much just pointed and grumbled. Thanks for hunting it down!

tests/json-fixtures/test_blockchain.py
lithp: CircleCI exposes [some useful environment variables](https://circleci.com/docs/2.0/env-vars/#built-in-environment-variables) such as CIRCLE_PULL_REQUEST and CIRCLE_BRANCH you could use to detect whether you're building a branch or not.

I've [opened a PR](https://github.com/ethereum/py-evm/pull/1588) which attempts to schedule a nightly build which runs all the tests, even the slowest ones.

PR #1576, opened 2018-12-11
Upgrade py-ecc past a buggy release, to 1.4.6

carver:
### What was wrong?

py-ecc was pinned to an old version to resolve a bug.

### How was it fixed?

Upgrade py-ecc past the buggy release.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

Issue #1565, opened 2018-12-09
Update BLS signature module


hwwhww:
### What is wrong?

Update it with the latest [spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/bls_verify.md).

### How can it be fixed

Currently, the module is here: https://github.com/ethereum/py-evm/blob/master/eth/utils/bls.py

1. Replace BN128 with [BLS-12-381](https://github.com/ethereum/py_ecc/tree/master/py_ecc/optimized_bn128) curve.
2. Implement and `BLSVerify` and `BLSMultiVerify` from spec. Pseudocode:
    ```python
    def multi_verify(pubs, msgs, sig):
        len_msgs = len(msgs)
        assert len(pubs) == len_msgs
        o = FQ12(1)
        for m in set(msgs):
            group_pub = G1.zero()
            for i in range(len_msgs):
                if msgs[i] == m:
                    group_pub = add(group_pub, pubs[i])
            o *= pairing(group_pub, m, final_exponentiate=False)
        o *= pairing(neg(G1), sig, final_exponentiate=False)
        return o ** ((field_modulus ** 12 - 1) // curve_order) == 1

    ```
3. Plus `aggregate_sigs`, pseudocode:
    ```python
    def aggregate_sigs(sigs):
        o = FQ12(1)
        for s in sigs:
            o *= s
        return o
    ```
4. Replace the current blake hashing function with keccak-256.

hwwhww:
closing via #1581

PR #1581, opened 2018-12-12
Update BLS signature module

hwwhww:
### What was wrong?
#1565 

### How was it fixed?
1. Follow the logic from [the BLS signature verification
](https://github.com/ethereum/eth2.0-specs/blob/master/specs/bls_verify.md)
2. Replace BN128 with BLS-12-381 curve.
3. Replace the current blake hashing function with keccak-256.

### Note

* Please ignore the `eth/beacon/aggregation.py` for now. It's supposed to be updated in other PRs.
* This PR is opened for reviewing the reference implementation. The current CI error is due to other type hinting issues. @carver, feel free to solve that and merge #1576 before it, I'll resolve the conflicts. :)

#### Cute Animal Picture

saiga_antelope

### DIFF
--- a/eth/utils/bls.py
+++ b/eth/utils/bls.py
@@ -34,9 +34,9 @@ from eth.beacon.utils.hash import hash_
 
 
 G2_cofactor = 305502333931268344200999753193121504214466019254188142667664032982267604182971884026507427359259977847832272839041616661285803823378372096355777062779109  # noqa: E501
-qmod = q ** 2 - 1
+FQ2_order = q ** 2 - 1
 eighth_roots_of_unity = [
-    FQ2([1, 1]) ** ((qmod * k) // 8)
+    FQ2([1, 1]) ** ((FQ2_order * k) // 8)
     for k in range(8)
 ]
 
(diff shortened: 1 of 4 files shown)


hwwhww:
@pipermerriam yes, I'll investigate it.

eth/utils/bls.py
```diff
@@ -18,117 +21,163 @@
...
+def verify_multiple(pubs: Sequence[int],
+                    msgs: Sequence[bytes],
+                    sig: bytes,
+                    domain: int) -> bool:
+    len_msgs = len(msgs)
+    assert len(pubs) == len_msgs
+
+    o = FQ12([1] + [0] * 11)
+    for m in set(msgs):
+        # aggregate the pubs
```
djrtwo: I would gather the `pubs` related to `m` in a var maybe called `m_pubs` and then use `aggregate_pubs(m_pubs)` to get the `group_pub` rather than doing the individual calls to `add`.

carver:
> * @carver, feel free to solve that and merge #1576 before it, I'll resolve the conflicts. :)

Done. Thanks for doing the type cleanups! :)

eth/utils/bls.py
```diff
@@ -18,117 +21,163 @@
...
+    if both solutions have equal imaginary component the value with higher real
+    component is favored.
+    """
+    candidate_squareroot = value ** ((qmod + 8) // 16)
+    check = candidate_squareroot ** 2 / value
+    if check in eighth_roots_of_unity[::2]:
+        x1 = candidate_squareroot / eighth_roots_of_unity[eighth_roots_of_unity.index(check) // 2]
+        x2 = FQ2([-x1.coeffs[0], -x1.coeffs[1]])
+        # x2 = - x2
+        return x1 if (x1.coeffs[1], x1.coeffs[0]) > (x2.coeffs[1], x2.coeffs[0]) else x2
```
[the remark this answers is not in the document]
hwwhww: Oh nice catch 👀
The `x1` and `x2` are `FQ2` objects, and `coeffs` is `Sequence[Union[int, "FQ"]]`. So
1. If `FQ2` is initialized with `int` -> no `FQ2_object.coeffs[1].n` element
2. If `FQ2` is initialized wtih `FQ` -> there is `FQ2_object.coeffs[1].n` element

Since it is `int` in our usage, without `.n` is correct now.

I'll note it in the spec repo.

eth/utils/bls.py
```diff
@@ -18,117 +21,163 @@
...
+        x_cubed_plus_b2 = x_coordinate ** 3 + FQ2([4, 4])
+        y_coordinate = modular_squareroot(x_cubed_plus_b2)
+        if y_coordinate is not None:
+            break
+        x_coordinate += FQ2([1, 0])  # Add one until we get a quadratic residue
 
-assert HEX_ROOT ** 8 != FQ2([1, 0])
-assert HEX_ROOT ** 16 == FQ2([1, 0])
+    return multiply(
+        (x_coordinate, y_coordinate, FQ2([1, 0])),
```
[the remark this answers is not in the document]
hwwhww: Yes, I'll create a fix in spec.

eth/utils/bls.py
```diff
@@ -20,124 +24,200 @@
...
+        raise ValidationError(
+            "len(pubkeys) (%s) should be equal to len(messages) (%s)" % (
+                len(pubkeys), len_msgs
+            )
+        )
+
+    o = FQ12([1] + [0] * 11)
+    for m_pubs in set(messages):
+        # aggregate the pubs
+        group_pub = Z1
```
djrtwo: I think I was a bit unclear in the last comment. I meant something like the following which utilizes `aggregate_pubs` rather than directly calling `add`

```python
for message in set(messages):
    pubkeys_for_message = [
        pubkey for i, pubkey in enumerate(pubkeys)
        if messages[i] == message
    ]
    group_pub = aggregate_pubkeys(pubkeys_for_message)
```

hwwhww: Oh I see! I think I've tried to do that before, if we want to utilize `aggregate_pubkeys`, we will need:
1. `len(messages)` times extra `compress_G1()` calling (inside `aggregate_pubkeys`)
2. `len(messages)` times extra `decompress_G1()` calling for the second argument of `pairing`.

Alternatively, we might refactor `aggregate_signatures(signatures: Sequence[bytes]) -> Tuple[int, int]` to:
```python
def aggregate_pubkeys(pubkeys: Sequence[int]) -> int:
    return compress_G1(_aggregate_pubkeys(pubkeys))

def _aggregate_pubkeys(pubkeys: Sequence[int]) -> Tuple[FQ, FQ, FQ]:
    o = Z1
    for p in pubkeys:
        o = add(o, decompress_G1(p))
    return o
```

And then make `verify_multiple` call `_aggregate_pubkeys()`.


What do you think of it?

djrtwo: oh crap, you're right. 

I'm okay either way. It's not a huge gain in code reuse and this code won't change much once in place.

djrtwo: Do whichever you want. I'm going to use the updated work in #1631 in the morning so merge whenever you're ready.

hwwhww:
@djrtwo merging it now! We can optimize the BLS APIs while moving it to `py-ecc` later (#1592). :)

PR #1588, opened 2018-12-13
Add a nightly build which runs even the slow blockchain tests

lithp:
Working off https://github.com/ethereum/py-evm/pull/1577/files#r240997519, it might be nice to run the SLOWEST_TESTS more often than never. I'm not sure if this works, I wasn't able to get it to run locally, `circleci build --job py36-all-blockchain-tests` kept failing with out of disk space errors.

Does anyone know of a better way to test it than merging and waiting for nightfall like this adorable little owl?

#### Cute Animal Picture

owl

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -205,6 +205,12 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: py36-native-blockchain-transition
+  py36-all-blockchain-tests:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-all-blockchain-tests
   py36-rpc-state-byzantium:
     <<: *common
     docker:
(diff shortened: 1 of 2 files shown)


carver:
> Does anyone know of a better way to test it than merging and waiting for nightfall like this adorable little owl?

I do not.

Here is a potentially bad idea: what if we *only* ran the slowest tests in the nightly job? Even if we do that, I'm a bit worried that the tests will fail based on memory issues. I feel mildly confident that they will if we try to run everything *plus* the slow tests. Even if we only run slow tests, we might have to break them up further into smaller jobs to convince circle to run them.

pipermerriam:
Anyone have experience setting up and running a Jenkins server?  I think our best route to accomplish this would be having the devops team provisioning a more powerful machine that we can run these more intensive tests on (and maybe the benchmarks too since that machine is *likely* to have a more stable CPU profile.

pipermerriam:
Might also be worth looking into whether https://github.com/features/actions can help us with this.

veox:
Unrelated: note that there are other tests that aren't run, which are _not_ in `SLOWEST_TESTS`:

https://github.com/ethereum/py-evm/blob/7e3b313dbb485fe16729cc40b442310afd750dc2/tests/json-fixtures/test_blockchain.py#L131-L148

lithp:
Opened #1606 to help reduce memory pressure. Hopefully it's enough to stop CircleCI from crashing!

PR #1590, opened 2018-12-13
Just for fun: Minor speedup on CodeStream.

jacqueswww:
### What was wrong?

Nothing really, I actually was curious what numba @jit would do in terms of VM speed on all the function of `arithmetic.py` - it had no effect. Then ended up finding this, and seems to be faster, Interested to find if this is just a margin of error (I gave it x3 runs each) ?

Before:
```
------------------------------------------------------------------------------------------------------------------------------------------------
|       Total       |    389.244     |     12504      |       -        |      9012      |         -          | 2,337,825,400  |       -        |
|        Avg        |     4.990      |      160       |     32.124     |      116       |       23.153       |   29,972,121   | 6,006,064.601  |
================================================================================================================================================


```

After:
```

------------------------------------------------------------------------------------------------------------------------------------------------
|       Total       |    366.143     |     12504      |       -        |      9012      |         -          | 2,337,825,400  |       -        |
|        Avg        |     4.694      |      160       |     34.151     |      116       |       24.613       |   29,972,121   | 6,385,013.504  |
================================================================================================================================================

```

### How was it fixed?

Basically caching len & value[x] instead of getvalue[i].

### DIFF
--- a/eth/vm/code_stream.py
+++ b/eth/vm/code_stream.py
@@ -15,12 +15,16 @@ from eth.vm import opcode_values
 class CodeStream(object):
     stream = None
     depth_processed = None
+    _len = None
+    _value = None
 
     logger = logging.getLogger('eth.vm.CodeStream')
 
     def __init__(self, code_bytes: bytes) -> None:
         validate_is_bytes(code_bytes, title="CodeStream bytes")
         self.stream = io.BytesIO(code_bytes)
+        self._value = code_bytes
+        self._len = len(code_bytes)
         self.invalid_positions = set()  # type: Set[int]
         self.depth_processed = 0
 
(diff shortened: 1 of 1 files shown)


pipermerriam:
That's actually quite a nice easy optimization and produced what appears to be a non-trivial VM speedup.  :+1:

eth/vm/code_stream.py
```diff
@@ -15,20 +15,24 @@
...
+    _len = None
+    _value = None
 
     logger = logging.getLogger('eth.vm.CodeStream')
 
     def __init__(self, code_bytes: bytes) -> None:
         validate_is_bytes(code_bytes, title="CodeStream bytes")
         self.stream = io.BytesIO(code_bytes)
+        self._value = code_bytes
+        self._len = len(code_bytes)
```
pipermerriam: Thoughts on renaming these to:

- `_len > _length_cache`
- `_value > _raw_code_bytes` or `_code_bytes`

jacqueswww: way more descriptive yes :)

Issue #1383, opened 2018-10-10
Add beacon chain / eth2.0 related NewType


hwwhww:
### What is wrong?
https://github.com/ethereum/py-evm/pull/1373#discussion_r224107745
To make the types more explict!

### How can it be fixed
1. Time unit - `Second` - and also can be used for PoW chain?
2. For beacon chain, some other `NewType` candidates: `Slot`, `ShardID`. And [with the latest spec](https://github.com/ethereum/eth2.0-specs/blob/cbd254a3090771e79ccc8eb8e21c7b32dba93c45/specs/beacon-chain.md#constants): `ValidatorStatus`, `SpecialRecordType`, `ValidatorSetDeltaFlag`.
3. **EDIT:** Move to [ethereum/eth-typing/](https://github.com/ethereum/eth-typing/)

hwwhww:
@cburgdorf is there any known solution to define: (i) the length of an integer (ii) unsigned/signed integer via type hinting?

cburgdorf:
> is there any known solution to define: (i) the length of an integer (ii) unsigned/signed integer via type hinting?

No, `mypy` can not do these things. What it *can* do is to allow you different types for different *intents* and ensure to trace the flow of these to prevent the wrong one ending up at a wrong place.

Consider this:

```python

def create_some_thing_unsigned() -> int:
    ...

def create_some_thing_signed() -> int:
    ...

def consume_something_unsigned(unsigned: int):
    ...
```

clearly we want to prevent the following situation:

```python
x = create_something_signed()
consume_something_unsigned(x)  # runtime crash!
```

Mypy can help you with this if you create a `NewType`.

```python
UnsignedInt = NewType('UnsignedInt', int)

def create_some_thing_unsigned() -> UnsignedInt:
    ...

def create_some_thing_signed() -> int:
    ...

def consume_something_unsigned(unsigned: UnsignedInt):
    ...
```

Now, if you'd try this again, mypy would yell at you because it knows that `x` is not of type `UnsignedInt`.

```python
x = create_something_signed()
consume_something_unsigned(x)  # development time mypy error!
```

hwwhww:
@cburgdorf thanks for the reply!

Re: > Move to ethereum/eth-typing/

What do you think about making the beacon chain new types in `eth/beacon/typing.py` temporarily, and then move to `eth-typing` once we settle down? I want to sort out what are the new types we need.

The types that I think we'd need:
```python
[...]

PR #1591, opened 2018-12-13
Add `get_fork_version` and `get_domain` helpers

hwwhww:
### What was wrong?
#1510 part 3

### How was it fixed?
1. Add `get_fork_version` and `get_domain` and tests. Spec: https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#get_fork_version
2. Put `ValidatorStatusCode`, `ValidatorRegistryDeltaFlag`, and `SignatureDomain` in `eth/beacon/enums.py`.

#### Cute Animal Picture

fox-in-snow

### DIFF
--- a/eth/beacon/enums.py
+++ b/eth/beacon/enums.py
@@ -1,0 +1,21 @@
+from enum import IntEnum
+
+
+class ValidatorStatusCode(IntEnum):
+    PENDING_ACTIVATION = 0
+    ACTIVE = 1
+    PENDING_EXIT = 2
+    EXITED_WITHOUT_PENALTY = 3
+    EXITED_WITH_PENALTY = 4
+
+
+class ValidatorRegistryDeltaFlag(IntEnum):
+    ACTIVATION = 0
+    EXIT = 1
+
+
+class SignatureDomain(IntEnum):
+    DOMAIN_DEPOSIT = 0
+    DOMAIN_ATTESTATION = 1
+    DOMAIN_PROPOSAL = 2
+    DOMAIN_LOGOUT = 3
--- a/eth/beacon/enums.py
+++ b/eth/beacon/enums.py
@@ -4,7 +4,7 @@ from enum import IntEnum
 class ValidatorStatusCode(IntEnum):
     PENDING_ACTIVATION = 0
     ACTIVE = 1
-    PENDING_EXIT = 2
+    ACTIVE_PENDING_EXIT = 2
     EXITED_WITHOUT_PENALTY = 3
     EXITED_WITH_PENALTY = 4
 
@@ -18,4 +18,4 @@ class SignatureDomain(IntEnum):
     DOMAIN_DEPOSIT = 0
     DOMAIN_ATTESTATION = 1
     DOMAIN_PROPOSAL = 2
-    DOMAIN_LOGOUT = 3
+    DOMAIN_EXIT = 3
--- a/eth/beacon/enums/signature_domain.py
+++ b/eth/beacon/enums/signature_domain.py
@@ -1,7 +1,7 @@
 from enum import IntEnum
 
 
-class BLSDomain(IntEnum):
+class SignatureDomain(IntEnum):
     DOMAIN_DEPOSIT = 0
     DOMAIN_ATTESTATION = 1
     DOMAIN_PROPOSAL = 2
--- a/eth/beacon/enums/signature_domain.py
+++ b/eth/beacon/enums/signature_domain.py
@@ -1,8 +1,0 @@
-from enum import IntEnum
-
-
-class SignatureDomain(IntEnum):
-    DOMAIN_DEPOSIT = 0
-    DOMAIN_ATTESTATION = 1
-    DOMAIN_PROPOSAL = 2
-    DOMAIN_LOGOUT = 3
--- a/eth/beacon/enums/special_record_types.py
+++ b/eth/beacon/enums/special_record_types.py
@@ -1,8 +1,0 @@
-from enum import IntEnum
-
-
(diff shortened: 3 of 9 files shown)


eth/beacon/enums/signature_domain.py
```diff
@@ -1,7 +1,7 @@
 from enum import IntEnum
 
 
-class BLSDomain(IntEnum):
+class SignatureDomain(IntEnum):
```
cburgdorf: @hwwhww how would you feel about creating just a single file at `eth/beacon/types/enum.py` and then collecting all enums in that single file. Given the simplicity of enums I feel we don't need one file per enum and keeping them all in one files slightly cleans up imports.

I'd also recommend we rename `eth/beacon/types` to `eth/beacon/typing` because that is more inline with the rest of our naming (e.g. we have a library `eth_typing` and we also have an `eth.typing` module with is like a gateway for things that eventually may be extracted into `eth_typing).

In the long run (read: when ETH 2.0 / Serenity is stable), I believe these types would move into the `eth_typing` library anyway to be reusable across many different projects.

hwwhww: > How would you feel about creating just a single file at eth/beacon/types/enum.py and then collecting all enums in that single file.

sounds good to me.

> I'd also recommend we rename `eth/beacon/types` to `eth/beacon/typing` because that is more inline with the rest of our naming

I stole the folder name `types` from `go-ethereum`. 😂
The role is similar to `eth/rlp/` instead these data structures will be SSZ serialization objects after we finish `py-ssz`. `typing` might be confused with simple type definition from `eth/typing.py`? The "correct" folder name might be `ssz`, although it's not ready, yet.

> In the long run (read: when ETH 2.0 / Serenity is stable), I believe these types would move into the eth_typing library anyway to be reusable across many different projects.

Agreed! ✨
I just updated https://github.com/ethereum/py-evm/issues/1383 description.

cburgdorf: >might be confused with simple type definition

The `eth_typing` as well as it's little sister `eth.typing` aren't exclusively for `NewType` or *type aliases* if that's what you mean by "type definitions".

`eth_typing` holds all types that are generic enough to be useful for other Ethereum related Python libraries. These types are usually very shallow and contain no or only minimal logic. They often fall into the category of `Enum`, `NewType`, `NamedTuple`, `TypedDict` as well as type aliases to specific `Tuple` or `Union` of some types.

The only difference between `eth_typing` and `eth.typing` is that the latter isn't entirely proven to be useful outside of the scope of this repository and so they reside within the repository unless someones screams "Hey, why are we duplicating all these types in project x that also exist in `eth.typing`". Then, we would move them into the `eth_typing` library.

>The role is similar to `eth/rlp/` instead these data structures will be SSZ serialization objects after we finish `py-ssz`

Right, I missed that. It kinda opens up the broader question how does `eth.rlp` relate to this. I guess the fact that they are all RLP serialized and contain some specific functionality makes them not a good fit to go into the broader scope of  `eth_typing`. But apart from that, there may be valid arguments for restructuring our modules to be more like:

```
eth/
    typing/
        rlp/
           eth1
              accounts.py
              blocks.py
              ...
           beacon
              attestation_data.py
              attestation_records.py
              ...
        misc.py (everything that is currently in eth.typing.py)
```

But this opens up an even bigger code organization question. Right now, everything ETH 2.0 related is sitting shielded in `eth/beacon` and while that is great for keeping it isolated from the rest of the library (especially while you keep iterating on the spec heavily) it does make some things less ergonomic I think.

Looking at the following two imports one can easily see that the beacon chain is something that was added later to this library.

```
from eth.rlp.accounts import Account
from eth.beacon.rlp.attestation_data import AttestationData
```

It's just not symmetric. To make it symmetric (and treating `rlp` as a submodule of `typing`) it would either be.

```
from eth.eth1.typing.rlp.accounts import Account
from eth.beacon.typing.rlp.attestation_data import AttestationData
```

or

```
from eth.typing.rlp.eth1.accounts import Account
from eth.typing.rlp.beacon.attestation_data import AttestationData
```

But this opens up a much bigger box of what I initially meant to discuss here (sorry). I think we touched on this topic in https://github.com/ethereum/py-evm/issues/930 before. Maybe it would be a good idea to continue the discussion with this specific example at hand. I think that at least *long* term we want to get rid of the asymmetry to not make the beacon support standout alien. /cc @djrtwo @pipermerriam @carver

pipermerriam: In generally I'm :+1: on figuring this out but I'm "meh" :neutral_face: on taking any major action towards it at this time.  Namespacing the `eth1` data structures and generally the `eth1` *stuff* seems appropriate at some point but low priority.

cburgdorf: Yes, we don't have to address this now. I guess it would be nice to address this at least before any release that declares stable beacon support though.

hwwhww: @cburgdorf I put `ValidatorStatusCode`, `ValidatorRegistryDeltaFlag`, and `SignatureDomain` in `eth/beacon/enums.py` now. (using plural for being aligned with `constants.py`)
I hope that we will use SSZ before the stable beacon chain release. 😊

eth/beacon/enums.py
```diff
@@ -0,0 +1,21 @@
+from enum import IntEnum
+
+
+class ValidatorStatusCode(IntEnum):
+    PENDING_ACTIVATION = 0
+    ACTIVE = 1
+    PENDING_EXIT = 2
```
ChihChengLiang:
```suggestion
    ACTIVE_PENDING_EXIT = 2
```

PR #1598, opened 2018-12-13
Pretty hash display utility

carver:
### What was wrong?

We're likely to write little bits of hash display code here and there. Better to have a common utility.

Inspired by https://github.com/ethereum/py-evm/pull/1392#discussion_r240773225

### How was it fixed?

Decided to keep it super simple by only supporting the long form.

The short form seems just so short to implement, that it's better not to worry about the utility, like:

```py
f"{header.hash[:4].hex()}"
```

I don't feel super strongly though.

#### Cute Animal Picture

screenshot_2018-12-13_15-25-22
_"hash browns"_

### DIFF
--- a/tests/trinity/core/utils/test_humanize.py
+++ b/tests/trinity/core/utils/test_humanize.py
@@ -44,7 +44,7 @@ def test_humanize_elapsed(seconds, expected):
 @pytest.mark.parametrize(
     'hash32,expected',
     (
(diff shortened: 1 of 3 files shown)


pipermerriam:
You could switch this to use it:

https://github.com/ethereum/py-evm/blob/master/trinity/sync/full/chain.py#L540

carver:
Will do, and use 2 bytes instead of 3.

PR #1603, opened 2018-12-14
Add support for different trinity root dirs

cburgdorf:
### What was wrong?

This is a spin off and a continuation of the discussion that started in https://github.com/ethereum/py-evm/pull/1566#discussion_r241366836

Now that we have support for the `trinity-beacon`  command, we still need to sort out how we would like to handle all the files that Trinity saves to discs. Today, that looks roughly like this:

```
/home/cburgdorf/.local/share/trinity
├── mainnet
│   ├── chain
│   │   ├── full
│   │   │   └── db files
│   │   └── light
│   │       └── db files
│   ├── logs
│   │   ├── trinity.log
│   │   ├── ...
│   │   └── trinity.log.7
│   └── nodekey
└── ropsten
    ├── chain
    │   ├── full
    │   │   └── db files
    │   └── light
    │       └── db files
    ├── logs
    │   └── trinity.log
    │   ├── ...
    │   └── trinity.log.7
    └── nodekey
```

For the beacon node, I guess we need a different `nodekey`, we want to keep logs separate and we need different directories for beacon and shard dbs.

My preferred solution, would be to treat the beacon node as an entirely different program and hence maintain an entirely separate root directory which would then roughly look like this.

```
/home/cburgdorf/.local/share/trinity-beacon   <-- notice the different root path
├── mainnet
│   ├── chain
│   │   ├── full
|   |   |     ├── beacon
│   │   │     |     └── db files
|   |   |     ├── shard-1
│   │   │     |     └── db files
|   |   |     └── shard-n
│   │   │           └── db files
│   │   ├── light
|   |   |     ├── beacon
│   │   │     |     └── db files
|   |   |     ├── shard-1
│   │   │     |     └── db files
|   |   |     └── shard-n
│   │   │           └── db files
│   ├── logs
│   │   ├── trinity.log
│   │   ├── ...
│   │   └── trinity.log.7
│   └── nodekey
└── some_future_testnet
     ├── chain
     │   ├── full
     |    |     ├── beacon
     │    │     |     └── db files
     |    |     ├── shard-1
     │    │     |     └── db files
     |    |     └── shard-n
     │    │           └── db files
     │    ├── light
     |    |     ├── beacon
     │    │     |     └── db files
     |    |     ├── shard-1
     │    │     |     └── db files
     |    |     └── shard-n
     │    │           └── db files
     ├── logs
     │   └── trinity.log
     │   ├── ...
     │   └── trinity.log.7
     └── nodekey
```

The alternative to that would be to figure out a new layout that places the beacon node files side by side inside the existing trinity root directory.

However, that would most likely also mean that existing data directories need to be migrated.

I personally think the cleanest thing to do is to treat the beacon node as an entirely different program which we already do by exposing it via a separate `trinity-beacon` command and providing a separate bootstrapping with different plugins etc.

### How was it fixed?

This implements the proposed solution by slightly changing our utils to allow setting using different root directories.

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/conftest.py
+++ b/tests/trinity/conftest.py
@@ -15,6 +15,7 @@ from eth.chains import (
 
 from trinity.constants import (
     NETWORKING_EVENTBUS_ENDPOINT,
+    APP_NAME_ETH1,
 )
 from trinity.chains.coro import (
     AsyncChainMixin,
@@ -53,7 +54,7 @@ def xdg_trinity_root(monkeypatch, tmpdir):
     dir_path = tmpdir.mkdir('trinity')
     monkeypatch.setenv('XDG_TRINITY_ROOT', str(dir_path))
 
-    assert not is_under_path(os.path.expandvars('$HOME'), get_xdg_trinity_root())
+    assert not is_under_path(os.path.expandvars('$HOME'), get_xdg_trinity_root(APP_NAME_ETH1))
 
     return Path(str(dir_path))
 
--- a/trinity/bootstrap.py
+++ b/trinity/bootstrap.py
@@ -109,6 +109,7 @@ BootFn = Callable[[
 
 def main_entry(trinity_boot: BootFn,
                plugins: Iterable[BasePlugin],
+               app_name: str,
                sub_configs: Iterable[Type[BaseAppConfig]]) -> None:
     event_bus = EventBus(ctx)
     main_endpoint = event_bus.create_endpoint(MAIN_EVENTBUS_ENDPOINT)
@@ -152,7 +153,7 @@ def main_entry(trinity_boot: BootFn,
         setup_log_levels(args.log_levels)
 
     try:
-        trinity_config = TrinityConfig.from_parser_args(args, sub_configs)
+        trinity_config = TrinityConfig.from_parser_args(args, app_name, sub_configs)
     except AmbigiousFileSystem:
         parser.error(TRINITY_AMBIGIOUS_FILESYSTEM_INFO)
 
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -35,6 +35,7 @@ from p2p.constants import (
 )
 
 from trinity.constants import (
+    APP_NAME_ETH1,
     ASSETS_DIR,
     DEFAULT_PREFERRED_NODES,
     MAINNET_NETWORK_ID,
@@ -216,6 +217,7 @@ class TrinityConfig:
 
     def __init__(self,
                  network_id: int,
+                 app_name: str = APP_NAME_ETH1,
                  genesis_config: Dict[str, Any]=None,
                  max_peers: int=25,
                  trinity_root_dir: str=None,
@@ -228,6 +230,7 @@ class TrinityConfig:
                  use_discv5: bool = False,
                  preferred_nodes: Tuple[KademliaNode, ...]=None,
                  bootstrap_nodes: Tuple[KademliaNode, ...]=None) -> None:
+        self.app_name = app_name
         self.network_id = network_id
         self.max_peers = max_peers
         self.sync_mode = sync_mode
@@ -330,7 +333,7 @@ class TrinityConfig:
         if self._trinity_root_dir is not None:
             return self._trinity_root_dir
         else:
-            return get_xdg_trinity_root()
+            return get_xdg_trinity_root(self.app_name)
 
     @trinity_root_dir.setter
     def trinity_root_dir(self, value: str) -> None:
@@ -348,7 +351,7 @@ class TrinityConfig:
         if self._trinity_root_dir is not None:
             return self._trinity_root_dir
         else:
-            return get_xdg_trinity_root()
+            return get_xdg_trinity_root(self.app_name)
 
     @trinity_root_dir.setter
     def trinity_root_dir(self, value: str) -> None:
@@ -452,6 +455,7 @@ class TrinityConfig:
     @classmethod
     def from_parser_args(cls,
                          parser_args: argparse.Namespace,
+                         app_name: str,
                          app_config_types: Iterable[Type['BaseAppConfig']]) -> 'TrinityConfig':
         """
         Helper function for initializing from the namespace object produced by
--- a/trinity/constants.py
+++ b/trinity/constants.py
@@ -15,6 +15,8 @@ from eth_keys import (
 
 from p2p.kademlia import Address, Node
 
+APP_NAME_ETH1 = 'trinity'
+APP_NAME_BEACON = 'trinity-beacon'
 
 # The file path to the non-python assets
 ASSETS_DIR = Path(__file__).parent / "assets"
--- a/trinity/main.py
+++ b/trinity/main.py
@@ -23,6 +23,9 @@ from trinity.bootstrap import (
     run_database_process,
     setup_plugins,
 )
+from trinity.constants import (
+    APP_NAME_ETH1,
+)
 from trinity.config import (
     TrinityConfig,
     Eth1AppConfig,
@@ -66,7 +69,7 @@ def get_all_plugins() -> Iterable[BasePlugin]:
 
 
 def main() -> None:
-    main_entry(trinity_boot, get_all_plugins(), (Eth1AppConfig,))
+    main_entry(trinity_boot, get_all_plugins(), APP_NAME_ETH1, (Eth1AppConfig,))
 
 
 def trinity_boot(args: Namespace,
--- a/trinity/main_beacon.py
+++ b/trinity/main_beacon.py
@@ -23,6 +23,9 @@ from trinity.config import (
     TrinityConfig,
     BeaconAppConfig
 )
+from trinity.constants import (
+    APP_NAME_BEACON,
+)
 from trinity.events import (
     ShutdownRequest
 )
@@ -42,7 +45,7 @@ from trinity.utils.mp import (
 
 
 def main_beacon() -> None:
-    main_entry(trinity_boot, BASE_PLUGINS, (BeaconAppConfig,))
+    main_entry(trinity_boot, BASE_PLUGINS, APP_NAME_BEACON, (BeaconAppConfig,))
 
 
 def trinity_boot(args: Namespace,
--- a/trinity/utils/xdg.py
+++ b/trinity/utils/xdg.py
@@ -35,11 +35,12 @@ def get_xdg_data_home() -> Path:
         return get_home() / '.local' / 'share'
 
 
-def get_xdg_trinity_root() -> Path:
+def get_xdg_trinity_root(app_name: str) -> Path:
     """
     Returns the base directory under which trinity will store all data.
     """
+    xdg_lookup_key = f'XDG_{app_name.upper()}_ROOT'
     try:
-        return Path(os.environ['XDG_TRINITY_ROOT'])
+        return Path(os.environ[xdg_lookup_key])
     except KeyError:
-        return get_xdg_data_home() / 'trinity'
+        return get_xdg_data_home() / app_name


carver:
On alternative is that maybe `trinity` becomes the all-in-one command for running both the beacon and mainnet. What we currently call `trinity` moves to `trinity-eth1` or something. Then implicitly `trinity` runs both `trinity-beacon` and `trinity-eth1`. So maybe the directory structure is something like `trinity/beacon-mainnet/...` and `trinity/eth1-mainnet/...`.

cburgdorf:
Yes, we could introduce this *sub-client* level but it's also a strong breaking change which means either:

- everyone needs to resync
- we tell people how to move their files to handle the change
- we auto-migrate existing directories

I'd be in favor to go down the simpler path and just treat the beacon node (and potentially the validator client, too) as entirely different programs with different root paths. In fact, maybe long term what we know today as `trinity` the client will just become a library with different clients being built on top of this (maybe similar to parity substrate even though I don't know anything beyond the marketing info about it)

carver:
Eh, we could go down the "it's an alpha suck-it-up" route, which we've decided before on this same basic issue.

Or we could name the eth1 mainnet as `mainnet`, and then it is backwards compatible with the current system.

cburgdorf:
Sure we can. But I would ask: what for? I think I just favor the idea of treating these as different applications. More flexibility, better separation of concerns, better isolation, less complexity. I guess I just don't see a strong argument why we should try treat this as one application or even make trinity the `all-in-one` command. What's with the linux philosophy of things? :)

pipermerriam:
@cburgdorf it should be trivial for us to write a quick plugin to do the migration for people so I'm :+1: on breaking change approach.  alpha is alpha...  We can also document the manual approach as well and really highlight the change in the release notes.

carver:
> I guess I just don't see a strong argument why we should try treat this as one application or even make trinity the `all-in-one` command.

AIUI the beacon chain will have ultimately have checkpoints that allow you to sync mainnet from the middle of the chain. From a user's point of view, they probably don't want to start the mainnet and the beacon chain and then wire them together, just to get that benefit. So the concept of what `trinity` is to the end user might reasonably expand to include several "internal" applications that we wire together for them.

pipermerriam:
I'm :-1: on a whole new directory.  I think 1) beacon chain is going to be linked to main chain because it will provide finality for main chain so they aren't completely separate and 2) I like the proposal to have `trinity` implicitly run the **full everything** (beacon + main + whatever-shards).   and 3) separating them doesn't feel right since sharing a nodekey across all of these should be fine and there are likely other resources that are acceptable to be shared/colocated.

cburgdorf:
>beacon chain is going to be linked to main chain because it will provide finality for main chain so they aren't completely separate

Ok, but I would argue that this kind of connection should rely on API access and not on both chains sharing the same application directory. Or put another way, I guess it should be totally reasonable for a trinity beacon node to receive any Ethereum 1 related information from a geth node.

All things considered I'm :-1: on mixing the application directories but I'm :+1: on democracy and if both of you prefer going down that route, so be it.

Let's dream up the application directory layout you'd like to have. How about this one. That one could even be backwards compatible by just having adding a `chain-beacon` directory next to the existing `chain`?

```
/home/cburgdorf/.local/share/trinity
├── mainnet                       <-- one mainnet directory, but different chain directories
│   ├── chain
│   │   ├── full
|   |   |     └── db files
│   │   └── light
|   |         └── db files
│   ├── chain-beacon
│   │   ├── full
|   |   |     ├── beacon
│   │   │     |     └── db files
|   |   |     ├── shard-1
│   │   │     |     └── db files
|   |   |     └── shard-n
│   │   │           └── db files
│   │   └── light
|   |         ├── beacon
│   │         |     └── db files
|   |         ├── shard-1
│   │         |     └── db files
|   |         └── shard-n
│   │               └── db files
│   ├── logs                      <-- one log directory per network but different file names
│   │   ├── trinity.log
│   │   ├── trinity-beacon.log
│   │   ├── ...
│   │   ├── trinity.log.7
│   │   └── trinity-beacon.log.7
│   └── nodekey
|
└── ropsten
    ├── chain
    │   ├── full
    |   |     └── db files
    │   └── light
    |         └── db files
    ├── chain-beacon
    │   ├── full
    |   |     ├── beacon
    │   │     |     └── db files
    |   |     ├── shard-1
    │   │     |     └── db files
    |   |     └── shard-n
    │   │           └── db files
    │   └── light
    |         ├── beacon
    │         |     └── db files
    |         ├── shard-1
    │         |     └── db files
    |         └── shard-n
    │               └── db files
    ├── logs
    │   ├── trinity.log
    │   ├── trinity-beacon.log
    │   ├── ...
    │   ├── trinity.log.7
    │   └── trinity-beacon.log.7
    └── nodekey
```

cburgdorf:
Ok, let me challenge this with one last line of thought before I give up. 

Where do you see the ETH 2.0 / Serenity validator client fit in to this?

I've been telling @djrtwo and @hwwhww that my preference would be to build the validator client on top of Trinity because there is a lot of nice infrastructure to reuse (mostly plugin / event bus / multiprocessing architecture)

I know that @djrtwo wants the validator as separate as possible and I agree to that and made the point that the validator client would be a different piece of software from the perspective of the operation system (different command, different process(es) etc). It would be able to connect to any kind of different but compatible software by prysmatic, lighthouse, nimbus, `<you name it>`.

Now if you are saying that you also want to put the validator client into the same app directory as well, then clearly the line gets blurred. From the perspective of the operating system a clearly distinct program should absolutely use its own application directory.

If you are saying that the validator client should use its own application directory but the beacon node should share the application directory with the Ethereum 1 node then I would argue that this makes things even worse (consistency etc).

And on 

>I think 1) beacon chain is going to be linked to main chain because it will provide finality for main chain so they aren't completely separate

Most of the other ETH 2.0 clients that are currently being build do not even have an existing ETH 1.0 sister project that they could share the application directory with. I think any kind of connection between the ETH 2.0 and the ETH 1.0 client should rely on clear API boundaries and not on a shared application directory.

To visualize how I see the future vision of Trinity, actually think parts of the code base can be treated as a library with just different clients being build on top of it.

```
trinity-library
├── eth 1 cient   <- own command, process(es) and application directory
│
├── beacon client  <- own command, process(es) and application directory
│
└── validator client  <- own command, process(es) and application directory
```

I like that unix philosophy more than the *all-in-one* vision. It is easy enough to build a thin *all-in-one* wrapper atop, it is difficult to go the opposite direction and tear apart things that have been mangled together.

Sorry to be so stubborn about it but I feel that there's little benefit from sharing one application directory and I just want to make sure we think this through.

pipermerriam:
I don't have a strong opinion on the validator stuff *yet* but I'm inclined to have it live elsewhere from the main trinity directory.

I see your arguments on separating eth1 and eth2 but I'm not convinced *at-this-time*.  If you feel strongly, lets open an issue to re-visit this in a few months, but for now, I'm still somewhat *strongly* :-1: on separating them.

cburgdorf:
@pipermerriam alright. I'll come up with a different PR to address this then. In the process of creating #1627 I just noticed that the above proposed layout still won't work under that model though because we currently track pidfiles in `trinity/<network>/<process>.pid`. So right now the `trinity` and the `trinity-beacon` command would overwrite each others pidfiles (e.g. both would create a `database.pid` or a `discovery-plugin.pid` etc.) To me this is yet another signal that we are really attempting to mix concerns here but it feels like beating a dead horse at this point.

The `TrinityConfig` could be made aware of an `app_prefix` which would then be used for logfiles and pidfiles to have both the eth1 and the beacon files live side by side. Or would you suggest something else?

carver:
> So right now the `trinity` and the `trinity-beacon` command would overwrite each others pidfiles (e.g. both would create a `database.pid` or a `discovery-plugin.pid` etc.) To me this is yet another signal that we are really attempting to mix concerns here but it feels like beating a dead horse at this point.

Here's how I'm thinking about it, so that the services are separate, but still under the trinity umbrella:
- right now, `trinity` runs a single core service: the eth1 mainnet
- sometime in the future `trinity` will run several core services: eth1 mainnet, beacon chain, etc

In that future, hard-core developers will be able to download and run the services for eth1 mainnet or beacon chain separately, to mix and match them with other services as desired. When run independently, they might default to something like `~/.local/eth1-mainnet`, `~/.local/eth-beacon`.

End users can always download and run `trinity`, which is meant to do all the coordination for them in a single local setup. When the services are run together, orchestrated by `trinity`, they are run in an combined data directory. `trinity` sets up the configuration for each service so that they are run in something like `~/.local/trinity/eth1-mainnet` and `~/.local/trinity/eth-beacon`. `trinity` also configures each service to talk to each other, across whatever ports or IPC files are necessary for the standard API communication.

cburgdorf:
@pipermerriam I think that we should create subdirectories for pidfiles and ipc files and then make the `TrinityConfig` aware of an `app_identifier` which e.g. could be `beacon` for the beacon node and then arrive at the following layout (assuming that the `app_identifier` for the eth1 node is `""`).

```
/home/cburgdorf/.local/share/trinity
├── mainnet                       <-- one mainnet directory, but sub directories
│   ├── chain
│   │   ├── full
|   |   |     └── db files
│   │   └── light
|   |         └── db files
│   ├── chain-beacon
│   │   ├── full
|   |   |     ├── beacon
│   │   │     |     └── db files
|   |   |     ├── shard-1
│   │   │     |     └── db files
|   |   |     └── shard-n
│   │   │           └── db files
│   │   └── light
|   |         ├── beacon
│   │         |     └── db files
|   |         ├── shard-1
│   │         |     └── db files
|   |         └── shard-n
│   │               └── db files
│   ├── logs
│   │   ├── trinity.log
│   │   ├── ...
│   │   ├── trinity.log.7
│   ├── logs-beacon
│   │   └── trinity.log
│   │   ├── ...
│   │   └── trinity.log.7
│   ├── ipc
│   │   ├── db.ipc
│   │   └── ...
│   ├── ipc-beacon
│   │   ├── db.ipc
│   │   └── ...
│   ├── pid
│   │   ├── database.pid
│   │   └── ...
│   ├── pid-beacon
│   │   ├── database.pid
│   │   └── ...
│   └── nodekey
|
└── ropsten
    ├── chain
    │   ├── full
    |   |     └── db files
    │   └── light
    |         └── db files
    ├── chain-beacon
    │   ├── full
    |   |     ├── beacon
    │   │     |     └── db files
    |   |     ├── shard-1
    │   │     |     └── db files
    |   |     └── shard-n
    │   │           └── db files
    │   └── light
    |         ├── beacon
    │         |     └── db files
    |         ├── shard-1
    │         |     └── db files
    |         └── shard-n
    │               └── db files
    ├── logs
    │   ├── trinity.log
    │   ├── ...
    │   ├── trinity.log.7
    ├── logs-beacon
    │   └── trinity.log
    │   ├── ...
    │   └── trinity.log.7
    ├── ipc
    │   ├── db.ipc
    │   └── ...
    ├── ipc-beacon
    │   ├── db.ipc
    │   └── ...
    ├── pid
    │   ├── database.pid
    │   └── ...
    ├── pid-beacon
    │   ├── database.pid
    │   └── ...
    └── nodekey
```

Because we don't want to bleed the multi app support into various parts of the application like the `fix-unclean-shutdown` plugin that currently globs all pidfiles in the data directory.

https://github.com/ethereum/py-evm/blob/629a4e47adb232f5143aa390a581f270e078b7bf/trinity/plugins/builtin/fix_unclean_shutdown/plugin.py#L38

So, with the proposed layout this would become:

```
pidfiles = tuple(trinity_config.pidfiles.glob('*.pid'))
```

And then `trinity.pidfiles` would just resolve to
`<trinity_root_dir>/<data-dir>/<pidfiles-{app-identifier}>`.

Or put another way, every directory `data-dir` is suffixed with the `app-identifer` and everything that is not already in a directory (currently ipc and pidfiles) moves into a directory. Does that make sense to you?

PR #1623, opened 2018-12-17
Fix #1564, rename hash_ to hash_eth2

### LINKED ISSUE #1564: Change Hashing Function from Blake2b to Sha3
### What is wrong?

Per https://github.com/ethereum/eth2.0-specs/pull/227

### How can it be fixed

1. Use this one:
    ```
    from eth_hash.auto import keccak
    ```
2. Remove `eth/utils/blake.py` and related tests.

ChihChengLiang:
### What was wrong?

- `hash_` is unclear name for the specific hash that used everywhere in eth2.0

### How was it fixed?

- rename `hash_` to `hash_eth2`

#### Cute Animal Picture

### DIFF
--- a/eth/beacon/utils/hash.py
+++ b/eth/beacon/utils/hash.py
@@ -3,4 +3,9 @@ from eth_hash.auto import keccak
 
 
 def hash_eth2(data: bytes) -> Hash32:
+    """
+    Return Keccak-256 hashed result.
+    Note: it's a placeholder and we aim to migrate to a S[T/N]ARK-friendly hash function in
+    a future Ethereum 2.0 deployment phase.
+    """
     return Hash32(keccak(data))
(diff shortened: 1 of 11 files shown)


eth/beacon/utils/hash.py
```diff
@@ -2,5 +2,5 @@
 from eth_hash.auto import keccak
 
 
-def hash_(data: bytes) -> Hash32:
+def hash_eth2(data: bytes) -> Hash32:
```
hwwhww: Add some docstring here, for example:
```suggestion
def hash_eth2(data: bytes) -> Hash32:
    """
    Return Keccak-256 hashed result.
    Note: it's a placeholder and we aim to migrate to a S[T/N]ARK-friendly hash function in
    a future Ethereum 2.0 deployment phase.
    """

PR #1626, opened 2018-12-17
[WIP] Extend round trip API to GetReceipts LES protocol command

hoi:
### What was wrong?

 * [ ]  `GetReceipts -> Receipts`

### How was it fixed?

#### Cute Animal Picture


carver:
Once you get this working, I'd say it would be best to go ahead and find and replace the API for receipts (like you did with block bodies here https://github.com/ethereum/py-evm/pull/1372/commits/bc5c2200db09c621d20be5c9fc38fe90c4983b38 ) and smoke test it in web3 (using `w3.eth.getTransactionReceipt(...)`)

PR #1627, opened 2018-12-17
Ensure isolated plugins create pidfiles

cburgdorf:
### What was wrong?

Without this, when an isolated plugin creates a
process, we do not track it with a pidfile.
That in turn can lead to dangling processes
which do not get cleaned up with the
`fix-unclean-shutdown` command.

### How was it fixed?

- add a `normalized_name` property to `BasePlugin` (btw, this should be used for Endpoint names as well)
- wrap `do_start` in a  `with ... process_pid_file()` block 

#### Cute Animal Picture


trinity/utils/os.py
```diff
@@ -7,3 +9,15 @@ def get_open_fd_limit() -> int:
...
     return soft_limit
+
+
+def friendly_filename_or_url(value: str) -> str:
+    """
+    Normalize any string to be file name and URL friendly.
+    Convert to lowercase, remove non-alpha characters,
+    and convert spaces to hyphens.
+    """
+    value = str(unicodedata.normalize('NFKD', value).encode('ascii', 'ignore'))
```
cburgdorf: This is from https://stackoverflow.com/questions/295135/turn-a-string-into-a-valid-filename/295466#295466

pipermerriam: Can you embed that link into the comment so that this is documented in the code.

cburgdorf:
>IUC were we to have two plugins with the same name then trinity would fail during startup because they would both try to occupy the same PID file?

In fact, you wouldn't even get as far because we currently use the plugin name as the `Endpoint` name and endpoint names need to be unique. So, even in current master plugin names need to be unique as well or else there will be an (probably unhandled) exception.

Maybe instead of just normalizing the name we should also put a `uuid` in there and call it `canonical_name` or `unique_name` or `normalized_identifier` or something like that? And then we can also use that property as the name for the `Endpoint` to get rid of that bug where two plugins with the same name would cause a crash.

pipermerriam:
> Maybe instead of just normalizing the name we should also put a uuid in there

Has to be deterministic so we'd need to use something like a UUID5 with some local source of randomness...

PR #1628, opened 2018-12-17
[WIP] Extend round trip API to GetNodeData LES protocol command

hoi:
### What was wrong?

 * [ ]  `GetNodeData -> NodeData`

### How was it fixed?

#### Cute Animal Picture


pipermerriam:

Issue #1607, opened 2018-12-15
Reduce benchmark suite running time


carver:
### What is wrong?

Benchmarking is regularly the slowest test. We aren't using the benchmarking data for anything, so the slowness has no benefit.

### How can it be fixed

Reduce the load factor on all the benchmark tests by ~100x (with a minimum of 2-3 runs, just to exercise the test a couple times).

Probably by setting custom run counts here:
https://github.com/ethereum/py-evm/blob/86297509afb50ccae75b9d2edc6aef23637556f1/scripts/benchmark/run.py#L78-L92

jmcph4:
@carver Apologies if I've overlooked something as I'm new to the codebase, but would it not be better to go straight to the source? For example with `MineEmptyBlocksBenchmark`:

https://github.com/ethereum/py-evm/blob/86297509afb50ccae75b9d2edc6aef23637556f1/scripts/benchmark/checks/mine_empty_blocks.py#L21-L25

carver:
The reason I suggested modifying in the test script is that when you *actually* run the benchmark (instead of just test that it works), you probably do want a bunch of runs (say, 500 in this case).

ashish10677:
Hello. I am new to open source and I want to contribute to this issue. Can someone please guide me?

carver:
@ashish10677 thanks for reaching out. This one is in progress in #1629 and fairly close to completion. Feel free to ping on another `Good First Issue` if you find another one you're keen to work on.

carver:
If anyone wants to pick up where #1629 left off (and update for recent benchmark changes), it looks like it's not being actively worked on anymore. @ashish10677 are you interested?

carver:
It's fast enough for now, closing.

PR #1629, opened 2018-12-18
Reduce benchmark suite running time

rayrapetyan:
### What was wrong?
https://github.com/ethereum/py-evm/issues/1607

### How was it fixed?
Reduced benchmark suite running time

### DIFF
--- a/scripts/benchmark/checks/deploy_dos.py
+++ b/scripts/benchmark/checks/deploy_dos.py
@@ -53,10 +53,13 @@ CONTRACT_NAME = 'DOSContract'
 
 W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
 
+DEFAULT_NUM_BLOCKS = 20
+DEFAULT_NUM_TX = 10
+
 
 class BaseDOSContractBenchmark(BaseBenchmark):
 
-    def __init__(self, num_blocks: int = 20, num_tx: int = 10) -> None:
+    def __init__(self, num_blocks: int = DEFAULT_NUM_BLOCKS, num_tx: int = DEFAULT_NUM_TX) -> None:
         super().__init__()
 
         self.num_blocks = num_blocks
@@ -227,6 +230,8 @@ class BaseDOSContractBenchmark(BaseBenchmark):
 
 
 class DOSContractDeployBenchmark(BaseDOSContractBenchmark):
+    def __init__(self, num_blocks: int = DEFAULT_NUM_BLOCKS, num_tx: int = DEFAULT_NUM_TX) -> None:
+        super().__init__(num_blocks, num_tx)
 
     @property
     def name(self) -> str:
@@ -237,6 +242,8 @@ class DOSContractDeployBenchmark(BaseDOSContractBenchmark):
 
 
 class DOSContractSstoreUint64Benchmark(BaseDOSContractBenchmark):
+    def __init__(self, num_blocks: int = DEFAULT_NUM_BLOCKS, num_tx: int = DEFAULT_NUM_TX) -> None:
+        super().__init__(num_blocks, num_tx)
 
     @property
     def name(self) -> str:
@@ -251,6 +258,8 @@ class DOSContractSstoreUint64Benchmark(BaseDOSContractBenchmark):
 
 
 class DOSContractCreateEmptyContractBenchmark(BaseDOSContractBenchmark):
+    def __init__(self, num_blocks: int = DEFAULT_NUM_BLOCKS, num_tx: int = DEFAULT_NUM_TX) -> None:
+        super().__init__(num_blocks, num_tx)
 
     @property
     def name(self) -> str:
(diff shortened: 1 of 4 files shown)


scripts/benchmark/checks/simple_value_transfers.py
```diff
@@ -94,15 +83,14 @@ def execute(self) -> DefaultStat:
 
         return total_stat
 
-    def mine_blocks(self, chain: MiningChain, num_blocks: int) -> Tuple[int, int]:
+    def mine_blocks(self, chain: MiningChain, num_blocks: int, num_tx: int) -> Tuple[int, int]:
         total_gas_used = 0
         total_num_tx = 0
 
         for i in range(1, num_blocks + 1):
-            num_tx = chain.get_block().header.gas_limit // SIMPLE_VALUE_TRANSFER_GAS_COST
```
carver: Hm, I see where you're going, but this hard-coding approach for the number of transactions seems problematic. It requires the configuration to be precisely aware of the gas block limit in order to fill it up. Maybe the configuration parameter could be the percentage of the block to fill up? (defaulting to 1.0)

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -227,6 +230,8 @@ def create_empty_contract_revert(self, chain: MiningChain) -> None:
 
 
 class DOSContractDeployBenchmark(BaseDOSContractBenchmark):
+    def __init__(self, num_blocks: int = DEFAULT_NUM_BLOCKS, num_tx: int = DEFAULT_NUM_TX) -> None:
+        super().__init__(num_blocks, num_tx)
```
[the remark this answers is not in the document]
rayrapetyan: @carver, oh wow... In Python derived class inherits base class "constructor", good to know!

scripts/benchmark/checks/simple_value_transfers.py
```diff
@@ -94,15 +83,14 @@ def execute(self) -> DefaultStat:
...
             block = self.mine_block(chain, i, num_tx)
-            total_num_tx = total_num_tx + len(block.transactions)
             total_gas_used = total_gas_used + block.header.gas_used
+            total_num_tx = total_num_tx + len(block.transactions)
```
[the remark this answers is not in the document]
rayrapetyan: @carver, this caused the longest delay in this benchmark.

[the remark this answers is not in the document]
rayrapetyan: @carver, to keep similarity with mine_blocks from erc20_interact and deploy_dos.

scripts/benchmark/checks/simple_value_transfers.py
rayrapetyan: @carver, do we really need to fill it up completely? Why don't we care about same in erc20_interact and deploy_dos?

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -53,10 +53,13 @@
 
 W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
 
+NUM_BLOCKS = 20
+NUM_TX = 10
```
carver: Heh, so these ones *are* default values, and I think the `DEFAULT_` prefix makes them clearer.
```suggestion
DEFAULT_NUM_TX = 10
```

scripts/benchmark/checks/simple_value_transfers.py
carver: Yup, totally fair point. This is a fundamental change to how the benchmark works, but it's a change *toward* consistency with the other benchmarks. :+1: no change needed.

scripts/benchmark/run.py
```diff
@@ -48,6 +50,24 @@
     bold_red,
 )
 
+NUM_BLOCKS_TO_BENCHMARK = 2
+NUM_TX_TO_BENCHMARK = 1
```
carver: Maybe use `2` transactions here, just to make sure that there are no crashes while looping over the transactions.

```suggestion
NUM_TX_TO_BENCHMARK = 2
```

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -57,10 +57,12 @@
 
 W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
 
+NUM_BLOCKS = 100
```
carver:
```suggestion
DEFAULT_NUM_BLOCKS = 100
```

scripts/benchmark/checks/erc20_interact.py
```diff
@@ -57,10 +57,12 @@
...
+NUM_TX = 2
```
carver:
```suggestion
DEFAULT_NUM_TX = 2
```

carver:
```suggestion
DEFAULT_NUM_TX = 2
```

scripts/benchmark/checks/deploy_dos.py
```diff
@@ -53,10 +53,13 @@
 
 W3_TX_DEFAULTS = {'gas': 0, 'gasPrice': 0}
 
+NUM_BLOCKS = 20
```
carver:
```suggestion
DEFAULT_NUM_BLOCKS = 20
```

rayrapetyan:
@carver, is there anything awaiting any changes from me? I think this patch should significantly decrease tests time so earlier we merge it is better for devs here. Thanks.

pipermerriam:
@rayrapetyan looks like @carver left a number of suggested changes.  If you can review/implement/respond to those I think we can move this forward.

carver:
Let's reopen this later, if it becomes necessary again. It's not a huge change, the PR has been quiet a long time, and benchmarking has been fast enough lately.

Issue #1608, opened 2018-12-15
[WIP] Postmortem of trinity v0.1.0-alpha.20 release


carver:
### What is wrong?

There was a bug in the code that only affected *released* code (not the project when tested locally). In the process of fixing it and re-releasing, a lot of other problems occurred.

- The first problem was mostly inconsequential, but set the mood: the first release crashed mid flow with:
```
twine upload dist/*
ContextualVersionConflict: idna 2.7
```
- The core problem is that Ropsten and Mainnet now relied on genesis file assets which did not get included in the release
- This ^ was discovered after the alpha 18 release in a smoke test of the pip release, immediately after upload
- Attempted to panic-patch: 8d902b4f3b56b903a788a82548a77a4537ab46f2 and 30779883444ff508c7975c5dbd5e52221f7876ec (hooray for stack overflow)
- some local garbage like the two patches above ^ got pushed and couldn't be cleaned up because master is locked from force-push (not a bad idea, but it was also news to me)
- because things were crashing, and I wanted to test in the middle of the build, I had opened the `Makefile` and was essentially copying it line by line, but missed the clean step since it is written differently from the others. so an earlier test build of trinity alpha19 got uploaded when I had only meant to upload py-evm
- As part of ^ I tried to ctrl-c in the middle which was just me being on-tilt really. It turns out nothing terrible happened from the ctrl-c, but I thought it did because:
- One time that I did a local test of py-evm release, things looked *super* broken and I thought it was because of ^. It turns out it was because I forgot to `--pre` in my pip install. Weirdly there is currently a stable v0.1.0 release on pip (that is super busted).
- I thought I could simply release `trinity` with the new assets, but was (of course, in hindsight) wrong. Even though the `trinity` build *does* package up all the assets, those aren't the ones actually used by trinity. (trinity uses the assets packaged with the py-evm release)
- Finally, I "broke the build" (red :x: in CI) on master when I hot-patched the live release. It looks ugly, but turned out to simply be a warning in the docs. Nothing obvious to fix about that particular issue. (Except maybe that the underlines under doc titles can be much longer than the title to avoid being a couple characters short after a change)

This writeup is still a WIP. I just wanted to get the core stuff written before the memory starts fading over the weekend.

### How can it be fixed

- [ ] split the Makefile targets in two:
  - [ ] one to produce a local release wheel for testing
  - [ ] another to push to git (esp since master is not force-push-able) and upload to pypi via twine
[...]

PR #1630, opened 2018-12-18
Add `make package` target to test packaged release

carver:
### What was wrong?

#1608 release issues - hard/slow/error-prone to test a local package of a release

### How was it fixed?

Add a new script and `Makefile` target.

Works locally, in fact I think I found a bug that would affect the next release. PR for that will show up tomorrow.

#### Cute Animal Picture

### DIFF
--- a/scripts/release/test_package.py
+++ b/scripts/release/test_package.py
@@ -1,0 +1,53 @@
+from pathlib import Path
+import subprocess
+from tempfile import TemporaryDirectory
+import venv
+
+
+def create_venv(parent_path):
+    venv_path = parent_path / 'package-smoke-test'
+    venv.create(venv_path, with_pip=True)
+    subprocess.run([venv_path / 'bin' / 'pip', 'install', '-U', 'pip', 'setuptools'], check=True)
+    return venv_path
+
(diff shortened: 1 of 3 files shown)


carver:
> So the idea is to run make package first and test locally if the package works?

Exactly. I'll take a first stab at the docs and merge.

scripts/release/test_package.py
```diff
@@ -0,0 +1,53 @@
...
+            venv_path / 'bin' / 'pip',
+            'install',
+            f"{wheel_path}{extra_suffix}"
+        ],
+        check=True,
+    )
+
+
+def test_install_local_wheel():
+    temporary_dir = TemporaryDirectory()
```
pipermerriam: If you use the context manager API here then the tempdir will still get cleaned up even if/when this script errors out.  If you're intentionally keeping it around it might be good to add a comment indicating as much.

carver: Ah, thanks. I was originally doing a `yes/no` prompt to keep it around for inspection. When I refactored back to this current approach, I forgot to go back to the context manager.

Issue #1616, opened 2018-12-17
Operate attestations


hwwhww:
### What is wrong?

There *were* some validation helpers in [`eth/beacon/state_machines/validation.py`](https://github.com/ethereum/py-evm/blob/1dabf4657ad78bfbd40bd29adb560197bb839616/eth/beacon/state_machines/validation.py). Now we need to implement the new ones for the latest spec. Start from `Attestion` validation.

### How can it be fixed
Implement `operate_attestations` based on [the spec - operating attestations](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#attestations-1):

```python
def operate_attestations (state: BeaconChain, attestations: Sequence[Attestation]) -> BeaconState:
    validate_attestations(state, attestations)
    state = update_latest_attestations(state, attesations)
    return state
```

1. Implement the `validate_attestations` function in `eth/beacon/state_machines/validation.py`
    * Implement the "Verify ..." logic from spec.
    * Raise `ValidationError` exception if the check fails.
2. Implement an immutable function `update_latest_attestations` for the last line - Append `PendingAttestationRecord(data=attestation.data, participation_bitfield=attestation.participation_bitfield, custody_bitfield=attestation.custody_bitfield, slot_included=state.slot)` to `state.latest_attestations`.
    * (STUB) Implement it in `eth/beacon/state_machines/state_update.py`.  <-- may need a better name...

djrtwo:
I think I'm in favor of using `process_` to describe what we do to the beacon chain operations. "operate ___" reads a bit strange.

- `process_attestations`
- `process_exits`
- etc...

djrtwo:
Where should these new `operations` live? maybe `eth/beacon/state_machines/operations.py`?

I'm thinking thinking that most of the work will happen in a file like `eth/beacon/state_machines/state_transition.py` that has the following functions (or something similar):
* `state_transition`
* `per_slot_transition`
* `per_block_transition`
* `per_epoch_transition`

The operation functions (`process_attestations`, `process_exits`, etc) could live in there or be imported from some specific `operations.py` file.

hwwhww:
Having `operations.py` for the operation functions sounds good to me. 👍 

The layout (`eth/beacon/state_machines/state_transition.py`) above makes sense. So we can follow the spec tightly at the beginning. (And yay we have better spec structure now!), then the `eth/beacon/state_machines/base.py` would be lighter (yay no more `CrystallizedState` now!).

[...]

PR #1633, opened 2018-12-18
`StateMachine` and `StateTransition` layout

hwwhww:
### What was wrong?
Per follow-up https://github.com/ethereum/py-evm/issues/1616#issuecomment-448078307 and https://github.com/ethereum/py-evm/issues/1616#issuecomment-448094498, create a stub architecture of `StateMachiane` and `StateTransition`.

### How was it fixed?

1. Directory would be like this:
```
eth/beacon/state_machines
├── __init__.py
├── base.py                         # BaseStateMachine
├── configs.py
├── forks
│   ├── __init__.py
│   └── serenity
│       ├── __init__.py
│       ├── blocks.py
│       ├── configs.py
│       ├── operations.py
│       ├── state_transitions.py
│       └── states.py
└── state_transitions.py            # BaseStateTransition
```

2. The class configuration will be at `StateMachine`, and we can swap the `BeaconConfig` config (actually, they are more like constants, but they can still be configurable during tests) easily.

3. The `tests/beacon/state_machines/test_demo.py` is just a demo of how to use `BeaconConfig` within `StateTransition` class. The fake `process_attestations` sets `state.slot` to `config.ZERO_BALANCE_VALIDATOR_TTL`.

Feedback would be appreciated. 🙂 

#### Cute Animal Picture

sheep


djrtwo:
oh,  I just notice this!

eth/beacon/state_machines/base.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+        if cls.state_class is None:
+            raise AttributeError("No `state_class` has been set for this StateMachine")
+        else:
+            return cls.state_class
+
+    @classmethod
+    def get_state_transiton_class(cls) -> Type[BaseStateTransition]:
+        """
+        Return the :class:`~eth.beacon.state_machines.state_transition.state_transition_class`
```
cburgdorf: The `:class:` directive should point to the class itself, not some property.

eth/beacon/state_machines/forks/serenity/state_transitions.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def apply_state_transition(self, state: BeaconState, block: BaseBeaconBlock) -> BeaconState:
+        state = self.per_slot_transition(state, block)
+        state = self.per_block_transition(state, block)
+        state = self.per_epoch_transition(state, block)
+
+        return state
+
+    def per_slot_transition(self, state: BeaconState, block: BaseBeaconBlock) -> BeaconState:
+        state = process_attestations(state, block, self.config)
```
djrtwo: `process_attestations` will go in `per_block_transition`

eth/beacon/state_machines/base.py
```diff
@@ -0,0 +1,95 @@
...
+class BeaconStateMachine(BaseBeaconStateMachine):
+    def __init__(self,
+                 chaindb: BaseBeaconChainDB,
+                 block: BaseBeaconBlock,
+                 state: BeaconState) -> None:
+        # TODO: get state from DB, now it's just a stub!
+        self.state = state
+
+    @classmethod
+    def get_block_class(cls) -> Type[BaseBeaconBlock]:
```
cburgdorf: @hwwhww @djrtwo I think we should put all these methods as `abstract` methods on the `BaseBeaconStateMachine` just to have a formal interace without implementation. Similar to what I stubbed in https://github.com/ethereum/py-evm/issues/1616#issuecomment-448297218

hwwhww: @cburgdorf 
👍 on having @abstract methods on `BaseBeaconStateMachine`!

I suppose we still want to have `class StateMachine(BaseStateMachine)` (copied from `class VM(BaseVM)`) ? And  then define `class SerenityStateMachine(StateMachine)`? <-- different from https://github.com/ethereum/py-evm/issues/1616#issuecomment-448297218

cburgdorf: Yes, the `StateMachine(BaseStateMachine)` can implement everything that is generic enough to be implemented on that level. :+1:

eth/beacon/state_machines/forks/serenity/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+from .blocks import SerenityBeaconBlock
+from .states import SerenityBeaconState
+from .state_transitions import SerenityStateTransition
+from .configs import SERENITY_CONFIG
+
+
+class SerenityStateMachine(BeaconStateMachine):
+    # fork name
+    fork = 'serenity'  # type: str
```
pipermerriam: @cburgdorf am I correct that we should start using [`ClassVar`](https://docs.python.org/3.5/library/typing.html#typing.ClassVar) for things like this?

eth/beacon/state_machines/forks/serenity/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+    # classes
+    block_class = SerenityBeaconBlock  # type: Type[BaseBeaconBlock]
```
pipermerriam: Similarly, these would all be `ClassVar` right?

eth/beacon/state_machines/state_transitions.py
```diff
@@ -0,0 +1,36 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
+
+from eth.utils.datatypes import (
+    Configurable,
```
pipermerriam: @cburgdorf this has been problematic for typing right?  Thoughts on deprecating this pattern other than in cases where we require dynamic class construction (like the `Chain` API which needs to be able to be dynamically configured at runtime to support custom genesis configs)

eth/beacon/state_machines/forks/serenity/__init__.py
cburgdorf: I think that should continue to be `Type[BaseBeaconBlock]` (https://github.com/ethereum/py-evm/pull/1633#discussion_r243098530)

eth/beacon/state_machines/state_transitions.py
cburgdorf: I have thought about it as well but I wasn't sure how much we still rely on the benefits of the pattern. E.g. is it needed for the sweet builder system that use to construct chains for testing?

hwwhww: > Thoughts on deprecating this pattern other than in cases where we require dynamic class construction (like the Chain API which needs to be able to be dynamically configured at runtime to support custom genesis configs)

Do you mean `StateTransition` doesn't need to be `Configurable` since we can configure it in `StateMachine` level?

hwwhww:
@cburgdorf @pipermerriam I'm merging it since it's blocking #1631. Feel free to continue the discussion about design pattern here or creating a new issue. :)

eth/beacon/state_machines/state_transitions.py
cburgdorf: The idea would be to get rid of  `Configurable` entirely. Here's some previous discussion about it https://github.com/ethereum/py-evm/issues/622

PR #1637, opened 2018-12-18
Fix #1635, update pubkey and signature length: `uint256` to `uint384`

### LINKED ISSUE #1635: Update pubkey and signature length: `uint256` to `uint384`
### What is wrong?
We have BLS12-381 curve now!

### How can it be fixed
1. Update all `pubkey` fields from `uint256` to `uint384`.
2. Update `signature` fields from `CountableList(uint256)` to `CountableList(uint384)`
3. Find all related tests and functions and then modify them. 👀

ChihChengLiang:
### What was wrong?

#1635

### How was it fixed?

- Update whenever `uint384` is in the spec.
- `get_new_validator_registry_delta_chain_tip` untouched, since it will be replaced with tree hash soon.

#### Cute Animal Picture

### DIFF
--- a/eth/beacon/types/attestations.py
+++ b/eth/beacon/types/attestations.py
@@ -9,7 +9,7 @@ from rlp.sedes import (
 )
 
 from eth.rlp.sedes import (
-    uint256,
+    uint384,
 )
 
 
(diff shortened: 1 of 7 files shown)


eth/beacon/types/attestations.py
```diff
@@ -29,7 +29,7 @@ class Attestation(rlp.Serializable):
         # Proof of custody bitfield
         ('custody_bitfield', binary),
         # BLS aggregate signature
-        ('aggregate_sig', CountableList(uint256)),
+        ('aggregate_sig', CountableList(uint384)),
```
hwwhww: note: renamed in #1631

PR #1638, opened 2018-12-18
Add missing __ini__.py file

cburgdorf:
### What was wrong?

The discovery plugin does not have an `__init__.py` file. While this does not have any effect during development, it will blow up once we cut a new release (because directories without `__init__.py` files won't be included in the distribution package)

### How was it fixed?

Added `__init__.py` file.

#### Cute Animal Picture


cburgdorf:
@carver sure, no problem

pipermerriam:
Needs to be migrated to `ethereum/trinity`

cburgdorf:
Moved to https://github.com/ethereum/trinity/pull/9

PR #1640, opened 2018-12-18
Routine for processing deposits

hwwhww:
### What was wrong?
Implement [routine for processing deposits](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#routine-for-processing-deposits)

### How was it fixed?
1. Add `BeaconState.update_validator(validator_index, validator)`
2. Fix `DepositInput`
3. Fix `get_new_validator_registry_delta_chain_tip`
4. Add `eth/beacon/deposits_helpers.py` <-- might be put to somewhere under state machine
    - Implement `def min_empty_validator_index` as spec
    - Impelment `def process_deposit` as spec

**Question to the reviewers**:
1. The state transition function will update `state: BeaconState` RLP/SSZ object frequently. Currently, I'm using `changeset()` for each snippet, but I'm not sure if that's the best practice.
2. Also add an instance function `BeaconState.update_validator(validator_index, validator)` for updating `BeaconState.validator_registry: Sequence[ValidatorRecord]` - immutably. Do you think it's okay for readability or any other concerns?

#### Cute Animal Picture

shark


eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+        if validator.withdrawal_credentials != validator.withdrawal_credentials:
+            raise ValidationError(
+                "`withdrawal_credentials` are incorrect:\n"
+                "\texpected: %s, found: %s" % (
+                    validator.withdrawal_credentials,
+                    validator.withdrawal_credentials,
+                )
+            )
+
+        # Update validator's balance and state
```
djrtwo: Might be worth considering creating two state helpers -- `increase_balance`, `decrease_balance`--  that handle this common `validator_changset` + `update_validator` flow.

tests/beacon/test_deposit_helpers.py
```diff
@@ -0,0 +1,221 @@
...
+def test_process_deposit(sample_beacon_state_params,
+                         zero_balance_validator_ttl,
+                         privkeys,
+                         pubkeys):
+    state = BeaconState(**sample_beacon_state_params).copy(
+        slot=zero_balance_validator_ttl + 1,
+        validator_registry=(),
+    )
+
+    privkey_1 = privkeys[0]
```
djrtwo: maybe should do a loop

```python
for index in range(3):
    privkey = privkeys[index]
    pubkey = pubkeys[index]
    ....
    result_state, validator_index = process_deposit()
    ...
    check_results()
```

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+        pubkey,
+        proof_of_possession,
+        withdrawal_credentials,
+        randao_commitment,
+    )
+
+    validator_pubkeys = tuple([v.pubkey for v in state.validator_registry])
+    if pubkey not in validator_pubkeys:
+        # Add new validator
+        validator = ValidatorRecord(
```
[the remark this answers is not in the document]
djrtwo: hmm not sure. Could make isolated testing a little easier, but not sold on it.

djrtwo:
(2) seems okay to me.
Not sure about (1). We're going to be updating the shit out of `state` in the transition function.

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+                "\texpected: %s, found: %s" % (
+                    validator.withdrawal_credentials,
+                    validator.withdrawal_credentials,
+                )
+            )
+
+        # Update validator's balance and state
+        with validator.build_changeset() as validator_changeset:
+            validator_changeset.balance = validator.balance + deposit
+            validator = validator_changeset.commit()
```
pipermerriam: This can be simplified to: `validator = validator.copy(balance=validator.balance + deposit)`.  The `changeset` API is better purposed for when you need to do multiple updates, potentially with some conditional logic.

hwwhww: Oh, you're right.

And I just benchmarked between using `copy()` and `changeset()` - I thought using `copy` would be slower since it's `deepcopy`, but, it turns out using `copy` *is 30-40% faster*(!)

hwwhww: p.s. I haven't tried a hugh RLP data like `BeaconState`. But the performance for `ValidatorRecord` should be better with using `copy()`.

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,142 @@
...
+        ):
+            return index
+    return None
+
+
+def validate_proof_of_possession(state: BeaconState,
+                                 pubkey: int,
+                                 proof_of_possession: bytes,
+                                 withdrawal_credentials: Hash32,
+                                 randao_commitment: Hash32) -> bool:
```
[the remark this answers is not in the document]
hwwhww: Will fixed it! As @pipermerriam also pointed out on https://github.com/ethereum/py-evm/pull/1640#discussion_r242729504

hwwhww:
### Changelog
1. Rebase for resolving conflicts
2. Add `eth/beacon/exceptions.py` - `MinEmptyValidatorIndexNotFound`
3. Refactors
	1. Update `eth.utils.bls`
		1. Currently, inside the the pairing functions, there are some `assert ...` thing that will be fixed in https://github.com/ethereum/py_ecc/pull/35. Currently I add something to make sure the `verify` function returns `bool`. 
	2. Add `eth.beacon.deposit_helpers.add_pending_validator`
	3. Add `ValidatorRecord.get_pending_validator()` classmethod for generating new pending validator
4. Resolve PR feedback

tests/beacon/test_deposit_helpers.py
djrtwo: Oh I see! Maybe update comment 

```
# Add the third validator.
# Should overwrite previously exited validator
```

or something. Would be good to make it more explicit than just the `assert len(result_state.validator_registry) == 2`

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,160 @@
...
+        state,
+        pubkey,
+        proof_of_possession,
+        withdrawal_credentials,
+        randao_commitment,
+    )
+
+    validator_pubkeys = tuple([v.pubkey for v in state.validator_registry])
+    if pubkey not in validator_pubkeys:
+        validator = ValidatorRecord.get_pending_validator(
```
djrtwo: Would this be clearer as `create_pending_validator` rather than `get_`?
`get` implies to me that it is trying to retrieve.

Not sure what the codebase naming conventions are in general. Will defer to @pipermerriam

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,160 @@
...
+from eth.beacon.helpers import (
+    get_domain,
+)
+
+
+def get_min_empty_validator_index(validators: Sequence[ValidatorRecord],
+                                  current_slot: int,
+                                  zero_balance_validator_ttl: int) -> int:
+    for index, validator in enumerate(validators):
+        if (
```
pipermerriam: small formatting nitpick.  I'd capture the contidional in a variable and then use it in a single line `if`

```python
is_empty = (validator.balance == 0 and ...)
if is_empty:
    return index
```

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,160 @@
...
+    try:
+        index = get_min_empty_validator_index(
+            state.validator_registry,
+            state.slot,
+            zero_balance_validator_ttl,
+        )
+    except MinEmptyValidatorIndexNotFound:
+        index = None
+
+    if index is None:
```
pipermerriam: I think this would all be a little better readability if you had all of this in the `try/except/else`

```python
try:
    index = ...
except MinEmptyValidatorIndexNotFound:
    ...  # logic from the `if index is None`
else:
    ...  # other logic.

return state, index
```

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    validate_proof_of_possession(
+        state,
+        pubkey,
+        proof_of_possession,
+        withdrawal_credentials,
+        randao_commitment,
+    )
+
+    validator_pubkeys = tuple([v.pubkey for v in state.validator_registry])
```
pipermerriam: The inner `[...]` brackets can be removed as `tuple(generator)` works just fine.

eth/beacon/deposit_helpers.py
```diff
@@ -0,0 +1,160 @@
...
+            state,
+            validator,
+            zero_balance_validator_ttl,
+        )
+    else:
+        # Top-up - increase balance by deposit
+        index = validator_pubkeys.index(pubkey)
+        validator = state.validator_registry[index]
+
+        if validator.withdrawal_credentials != validator.withdrawal_credentials:
```
pipermerriam: This conditional looks broken.  Seems like you are comparing the same value with itself.  Is one of these supposed to be just `withdrawal_credentials`.

Maybe add a test that covers this code path since this code path probably isn't being exercised in the tests right now?

eth/beacon/types/states.py
```diff
@@ -157,3 +157,13 @@ def num_validators(self) -> int:
     @property
     def num_crosslinks(self) -> int:
         return len(self.latest_crosslinks)
+
+    def update_validator(self,
+                         validator_index: int,
+                         validator: ValidatorRecord) -> 'BeaconState':
+        validator_registry = list(self.validator_registry)
+        validator_registry[validator_index] = validator
+        self = self.copy(
```
pipermerriam: I'm :-1: on overriding `self` in a method context.  Seems like an antipattern and easily able to lead to confusion.  Maybe just inline the `copy` in the `return` statement or call the variable something like `updated_state`

eth/utils/bls.py
```diff
@@ -170,15 +170,18 @@ def privtopub(k: int) -> int:
...
-            FQP_point_to_FQ2_point(hash_to_G2(message, domain)),
-            neg(decompress_G1(pubkey)),
-            False
+    try:
+        final_exponentiation = final_exponentiate(
+            pairing(FQP_point_to_FQ2_point(decompress_G2(signature)), G1, False) *
+            pairing(
+                FQP_point_to_FQ2_point(hash_to_G2(message, domain)),
+                neg(decompress_G1(pubkey)),
+                False
```
pipermerriam: I find positional booleans to be confusing since they give no context on what is being `False`'d..  Maybe convert to either a keyword argument, or just use a comment to denote what argument this is for if performance is an issue.

eth/utils/bls.py
```diff
@@ -208,16 +211,19 @@ def verify_multiple(pubkeys: Sequence[int],
...
-    o *= pairing(decompress_G2(signature), neg(G1), False)
-
-    final_exponentiation = final_exponentiate(o)
-    return final_exponentiation == FQ12.one()
+    try:
+        o = FQ12([1] + [0] * 11)
+        for m_pubs in set(messages):
+            # aggregate the pubs
+            group_pub = Z1
+            for i in range(len_msgs):
```
pipermerriam: This looks like you could do `for pubkey, message in zip(pubkeys, messages):` which removes the indexing.

Alternatively, I think the following is a functional equivalent implementation.

```python
groups = (
    pairing(hash_to_G2(m_pubs, domain), reduce(add, (decompress_G1(pubkey), Z1)), False)
    for (pubkey, message) 
    in zip(pubkeys, messages) if message == m_pubs
    for m_pubs in set(messages)
)
final_exponentiation = final_exponentiate(
    reduce(
        operator.mul, 
        groups,
        FQ12([1] + [0] * 11),
    ) * pairing(decompress_G2(signature), neg(G1), False)
)
return final_exponentiation == FQ12.one()
```

Also, there are a few constants in there that could be lifted out such as:

- `FQ12.one()`
- `neg(G1)`

The above is just hand-written code without being tested so you should verify if you choose to switch.  I *think* the above code would be more performant... but that's just intuition.

eth/beacon/deposit_helpers.py
djrtwo: Yeah, should be checking that the existing validator with the pubkey `pubkey` has the same `withdrawal_credentials` that were passed in with the deposit

PR #1643, opened 2018-12-18
Some tox.ini cleanups

carver:
### What was wrong?

#1608 - no CI test of the packaged release

### How was it fixed?

Add a new tox and circle run against the packaged release.

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -320,6 +320,12 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: py36-beacon
+  py36-wheel-cli:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-wheel-cli
 
   py37-core:
     <<: *common
@@ -345,6 +351,12 @@ jobs:
       - image: circleci/python:3.7
         environment:
           TOXENV: py37-beacon
+  py37-wheel-cli:
+    <<: *common
+    docker:
+      - image: circleci/python:3.7
+        environment:
+          TOXENV: py37-wheel-cli
 
 workflows:
   version: 2
@@ -355,6 +367,7 @@ workflows:
       - py37-core
       - py37-trinity
       - py37-trinity-integration
+      - py37-wheel-cli
       - py37-beacon
 
       - py36-native-blockchain-byzantium
@@ -382,6 +395,7 @@ workflows:
       - py36-database
       - py36-rpc-state-quadratic
       - py36-beacon
+      - py36-wheel-cli
 
       - py35-native-blockchain-byzantium
       - py35-native-blockchain-constantinople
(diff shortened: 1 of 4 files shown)


carver:
Hm, I was actually expecting the py36 one to fail (it does locally), because it doesn't collect the `peer_discovery` plugin. (which is missing the `__init__.py` file) https://github.com/ethereum/py-evm/tree/master/trinity/plugins/builtin/peer_discovery

Investigating...

pipermerriam:
The `xdist` thing in circle is because I have it configured using environment variables.  Let me know if you'd like me to remove that.

carver:
Might need some back and forth @pipermerriam -- moving to chat

carver:
To summarize some sidelined chat:

`-n` is being added to the `pytest` command in CI. I can't figure out where. It's not in `$PYTEST_ADDOPTS` on the circle server (`PYTEST_ADDOPTS="4 --durations 50 --maxfail 10"`), it's not in `tox.ini`, it's not in `pytest.ini`, not in `.circleci/config.yml`, there is no `setup.cfg`. If I could remove the `-n` then it would work correctly (at least it does locally) . Note that `ADDOPTS` starts with `4` which ends up being used as the argument to this phantom prefix `-n`.

This is the lovely failure from running the test when `pytest-xdist` is not installed:
```
py36-wheel-cli runtests: commands[3] | pytest tests/trinity/integration/ -k test_full_boot
usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: -n tests/trinity/integration/
  inifile: /home/circleci/repo/pytest.ini
  rootdir: /home/circleci/repo
ERROR: InvocationError for command '/home/circleci/repo/.tox/py36-wheel-cli/bin/pytest tests/trinity/integration/ -k test_full_boot' (exited with code 4)
```
Note that pytest lists the ini files that should be affecting it, but none seem to have the `-n`
```
$ cat pytest.ini 
[pytest]
addopts= --showlocals
python_paths= .
xfail_strict=true
log_format = %(levelname)8s  %(asctime)s  %(filename)20s  %(message)s
log_date_format = %m-%d %H:%M:%S

$ echo $PYTEST_ADDOPTS
4 --durations 50 --maxfail 10
```

---

I also tried it with `xdist` enabled, but it seems to be using a different build of trinity than the one that regular `pytest` runs do. If I run `pytest-xdist` locally or remotely, then the test passes (it shouldn't, because the package is broken right now). If there was some way to make xdist run the same build than non-xdist is using, that would be okay, too.

carver:
:tada: it crashed for the right reason now!

carver:
Blocked until the py-eth & trinity split is complete.

pipermerriam:

carver:
Already copied over to https://github.com/ethereum/trinity/pull/106

Most of this can be nuked, but a few things should be kept, like the tox.ini.

carver:
Eh, if we really need to make the tox change later, we can come back to it. No immediate need. Closing...

PR #1644, opened 2018-12-18
Update fixtures to the v6.0.0-beta.3 release

lithp:
[A new release of the tests](https://github.com/ethereum/tests/releases/tag/v6.0.0-beta.3) has just been published! This branch updates the tests but does not yet resolve the 7 errors that causes.

### DIFF
--- a/eth/vm/logic/context.py
+++ b/eth/vm/logic/context.py
@@ -151,7 +151,7 @@ def extcodehash(computation: BaseComputation) -> None:
     account = force_bytes_to_address(computation.stack_pop(type_hint=constants.BYTES))
     account_db = computation.state.account_db
 
-    if not account_db.account_exists(account):
+    if account_db.account_is_empty(account):
         computation.stack_push(constants.NULL_BYTE)
     else:
         computation.stack_push(account_db.get_code_hash(account))
--- a/tests/json-fixtures/test_blockchain.py
+++ b/tests/json-fixtures/test_blockchain.py
@@ -274,7 +274,6 @@ def test_blockchain_fixtures(fixture_data, fixture):
     latest_block_hash = chain.get_canonical_block_by_number(chain.get_block().number - 1).hash
     if latest_block_hash != fixture['lastblockhash']:
         verify_account_db(fixture['postState'], chain.get_vm().state.account_db)
-        assert False, 'the state must be different if the hashes are'
 
     for block, mined_block in mined_blocks:
         assert_mined_block_unchanged(block, mined_block)
--- a/tests/json-fixtures/test_blockchain.py
+++ b/tests/json-fixtures/test_blockchain.py
@@ -226,7 +226,7 @@ def test_blockchain_fixtures(fixture_data, fixture):
     try:
         chain = new_chain_from_fixture(fixture)
     except ValueError as e:
-        raise AssertionError("could not load chain for %r" % (fixture_data,)) from e
+        raise AssertionError("could not load chain for {}".format((fixture_data,))) from e
 
     genesis_params = genesis_params_from_fixture(fixture)
     expected_genesis_header = BlockHeader(**genesis_params)
--- a/tests/json-fixtures/test_blockchain.py
+++ b/tests/json-fixtures/test_blockchain.py
@@ -226,7 +226,7 @@ def test_blockchain_fixtures(fixture_data, fixture):
     try:
         chain = new_chain_from_fixture(fixture)
     except ValueError as e:
-        raise AssertionError("could not load chain for %r" % fixture_data) from e
+        raise AssertionError("could not load chain for %r" % (fixture_data,)) from e
 
     genesis_params = genesis_params_from_fixture(fixture)
     expected_genesis_header = BlockHeader(**genesis_params)
--- a/eth/tools/fixtures/helpers.py
+++ b/eth/tools/fixtures/helpers.py
@@ -153,6 +153,11 @@ def chain_vm_configuration(fixture: Dict[str, Any]) -> Iterable[Tuple[int, Type[
             (0, SpuriousDragonVM),
             (5, ByzantiumVM),
         )
+    elif network == 'ByzantiumToConstantinopleAt5':
+        return (
+            (0, ByzantiumVM),
+            (5, ConstantinopleVM),
+        )
     else:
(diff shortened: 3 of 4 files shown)


lithp:
Despite my optimism in the release notes https://github.com/ethereum/tests/pull/561 appears to have broken some trinity tests, reverting https://github.com/ethereum/tests/pull/561/commits/f8a8bcf762c538e8ec1f760d7535f377e73ae7a2 causes `test_blockchain_fixtures[/home/brian/ef/py-evm/fixtures/BlockchainTests/TransitionTests/bcHomesteadToDao/DaoTransactions_UncleExtradata.json:DaoTransactions_UncleExtradata:HomesteadToDaoAt5]` to pass.

lithp:
Note to self, the failing block was changed [here](https://github.com/ethereum/tests/pull/561/files#diff-cca3ea305df2f52261293da7412a19baR161). The undocumented `expectExceptionAll` key refers to the exception which aleth calls while processing the block. In this case, `ExtraDataIncorrect`, it's something do to with expecting specific extradata during the dao hardfork: https://github.com/ethereum/aleth/blob/master/libethcore/SealEngine.cpp#L98

carver:
> it's something do to with expecting specific extradata during the dao hardfork: https://github.com/ethereum/aleth/blob/master/libethcore/SealEngine.cpp#L98

Presumably, it's expecting the `extra_data` set to `'dao-hard-fork'`, as is checked here: https://github.com/ethereum/py-evm/blob/7c4630dff1b5ec6270b3383029883b4ff23712a9/eth/chains/mainnet/__init__.py#L55-L70

Note that we currently only check for this header `extra_data` in Mainnet, not every homestead transition.

lithp:
@carver thanks for the tip. Just pushed https://github.com/ethereum/py-evm/pull/1644/commits/2d386dccea309204f5c282fbcb6af1e9d6024174 which flips on the check for the correct extradata around the fork. `MainnetDAOValidatorVM` looks like it might have been intended to be private to the `eth.chains.mainnet` namespace though, do you think this is the right patch or do you think I should move MainnetDAOValidatorVM to somewhere in the `eth.vm.*` namespace?

lithp:
[Mana is also failing](https://github.com/mana-ethereum/mana/pull/655) on the fuzzer-generated test we're failing on.

lithp:
I ran `lllc` over the contracts found in the test to get an idea of what's being run: https://github.com/ethereum/tests/pull/551/files#diff-7f63422c4913e385f6ba379f9d952f77R47

one contract is `PUSH3 0xABCDEF SELFDESTRUCT`
and the other `PUSH2 0xDEAD PUSH1 0x0 PUSH1 0x0 PUSH1 0x0 PUSH1 0x0 PUSH1 0x0 PUSH2 0xDEAD GAS CALL PUSH3 0xABCDEF EXTCODEHASH PUSH1 0x1 SSTORE`

tests/json-fixtures/test_blockchain.py
```diff
@@ -226,7 +226,7 @@ def test_blockchain_fixtures(fixture_data, fixture):
     try:
         chain = new_chain_from_fixture(fixture)
     except ValueError as e:
-        raise AssertionError("could not load chain for %r" % fixture_data) from e
+        raise AssertionError("could not load chain for %r" % (fixture_data,)) from e
```
pipermerriam: *if* you get back in here this might as well be updated to use `.format(...)` to be consistent with the rest of the codebase.

lithp: Yeah, that's probably better. I'll change it to use `format()`

lithp:
After some quality time with the yellowpaper I think trinity's behavior is wrong. Trinity correctly implements `EIP-161` as specified but the yellowpaper says something slightly different. A return from a nested call frame also counts as the end of a transaction, so cleaning such as clearing empty accounts needs to happen then. I have a patch which fixes this test but makes some others fail, still trying to get it into a working state.

carver:
> Trinity correctly implements `EIP-161` as specified but the yellowpaper says something slightly different.

Just to confirm what I think you're saying: `EIP-161` under-specifies the situation, and the yellowpaper has a clear stance? Then yeah, the conclusion makes sense to me, although the yellowpaper is generally more fluid, so I wouldn't *assume* that there's buy-in in AllCoreDevs about that subtlety. The question to Arachnid on the EIP seems totally reasonable.

lithp:
> Just to confirm what I think you're saying: EIP-161 under-specifies the situation, and the yellowpaper has a clear stance? Then yeah, the conclusion makes sense to me, although the yellowpaper is generally more fluid, so I wouldn't assume that there's buy-in in AllCoreDevs about that subtlety. The question to Arachnid on the EIP seems totally reasonable.

This is embarrassing, it turned out to be a one-line change. The whitepaper definitely doesn't take a clear stance but page 8 says things like:

> After the message call or contract creation is processed,
the refund counter has to be incremented for the accounts
that were self-destructed throughout its invocation

which made it sound like the finalization needed to also happen at the end of nested transaction calls. And doing this fixed the test! But then getting the other tests to pass took forever, and eventually I looked at geth and parity and discovered they also weren't doing any finalization until the end of a transaction.

I suppose that in the future I should refer to, in descending order of authoritativeness: geth/parity, then the tests, then the EIP, then the yellowpaper

carver:
> I suppose that in the future I should refer to, in descending order of authoritativeness: geth/parity, then the tests, then the EIP, then the yellowpaper

Hopefully, someday, the spec will be more authoritative than the implementations :grin:

eth/vm/logic/context.py
```diff
@@ -151,7 +151,7 @@ def extcodehash(computation: BaseComputation) -> None:
     account = force_bytes_to_address(computation.stack_pop(type_hint=constants.BYTES))
     account_db = computation.state.account_db
 
-    if not account_db.account_exists(account):
+    if account_db.account_is_empty(account):
```
carver: Ahah, right, makes sense.

carver:
> This will need to be duplicated for the new `trinity` repo

Thinking out loud, some of the steps:
- cut an release on py-evm (probably `v0.3.0-alpha.1` since it's such a significant change)
- open a PR on trinity with an:
  - upgrade to the tests submodule
  - upgrade the py-evm dependency

Maybe that's it?

Issue #1646, opened 2018-12-19
Fresh clone of repo followed by `pytest` fails


ralexstokes:
* py-evm Version: `master` at `48be6ca6bea5c6349ff3e00e477ce0db8305b43f`
* OS: osx
* Python Version (python --version): 3.6.5
* Environment (output of `pip freeze`): 

alabaster==0.7.12
apipkg==1.5
appnope==0.1.0
argh==0.26.2
asn1crypto==0.24.0
async-generator==1.10
async-lru==0.1.0
asyncio-cancel-token==0.1.0a2
atomicwrites==1.2.1
attrdict==2.0.0
attrs==18.2.0
Babel==2.6.0
backcall==0.1.0
bleach==3.0.2
bloom-filter==1.3
bumpversion==0.5.3
cachetools==2.1.0
certifi==2018.11.29
cffi==1.11.5
chardet==3.0.4
coincurve==10.0.0
colorama==0.4.1
coverage==4.5.2
cryptography==2.4.2
cytoolz==0.9.0.1
decorator==4.3.0
docopt==0.6.2
docutils==0.14
eth-abi==1.3.0
eth-account==0.2.3
eth-bloom==1.0.1
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.0b3
eth-rlp==0.1.2
eth-typing==2.0.0
eth-utils==1.4.1
execnet==1.5.0
flake8==3.5.0
hexbytes==0.1.0
hypothesis==3.69.5
idna==2.7
ifaddr==0.1.6
imagesize==1.1.0
ipython==6.5.0
ipython-genutils==0.2.0
jedi==0.13.2
Jinja2==2.10
jsonschema==2.6.0
lahja==0.10.0
lru-dict==1.1.6
lxml==4.2.5
MarkupSafe==1.1.0
mccabe==0.6.1
more-itertools==4.3.0
mypy==0.641
mypy-extensions==0.4.1
netdisco==2.2.0
netifaces==0.10.7
packaging==18.0
parsimonious==0.8.1
parso==0.3.1
pathtools==0.1.2
pexpect==4.6.0
pickleshare==0.7.5
pkginfo==1.4.2
pluggy==0.7.1
plyvel==1.0.5
prompt-toolkit==1.0.15
ptyprocess==0.6.0
py==1.7.0
py-ecc==1.4.7
[...]

Issue #1547, opened 2018-12-05
Installation of external plugins broken


cburgdorf:
### What is wrong?

```
$ pip install ./trinity-external-plugins/examples/peer_count_reporter/
Processing ./trinity-external-plugins/examples/peer_count_reporter
Building wheels for collected packages: trinity-peer-count-reporter-plugin
  Running setup.py bdist_wheel for trinity-peer-count-reporter-plugin ... done
  Stored in directory: /tmp/pip-ephem-wheel-cache-4feh3p99/wheels/81/fb/9e/ab0d1ff12d1589a26e07fef18a6481ab4abcf7e888f0a5f9ad
Successfully built trinity-peer-count-reporter-plugin
Installing collected packages: trinity-peer-count-reporter-plugin
Successfully installed trinity-peer-count-reporter-plugin-0.0.0
(venv) cburgdorf at craftbot in ~/Documents/hacking/ef/py-evm2 on master
$ trinity --help
Traceback (most recent call last):
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/venv/bin/trinity", line 11, in <module>
    load_entry_point('py-evm', 'console_scripts', 'trinity')()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/trinity/main.py", line 115, in main
    MainAndIsolatedProcessScope(event_bus, main_endpoint)
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/trinity/main.py", line 380, in setup_plugins
    plugin_manager.register(get_all_plugins())
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/trinity/plugins/registry.py", line 58, in get_all_plugins
    return BUILTIN_PLUGINS + discover_plugins()
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/trinity/plugins/registry.py", line 53, in discover_plugins
    entry_point.load()() for entry_point in pkg_resources.iter_entry_points('trinity.plugins')
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/trinity/plugins/registry.py", line 53, in <genexpr>
    entry_point.load()() for entry_point in pkg_resources.iter_entry_points('trinity.plugins')
  File "/home/cburgdorf/Documents/hacking/ef/py-evm2/venv/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2346, in load
    return self.resolve()
[...]

Issue #1539, opened 2018-12-01
pytest run is failing


javimaravillas:
* py-evm Version: master (with https://github.com/ethereum/py-evm/pull/1538)
* OS: osx
* Python Version 3.6.7, 3.7.1
* Environment (output of `pip freeze`):
```sh
(venv36) javi@franciscos-mbp ~/e/py-evm> pip freeze
alabaster==0.7.12
apipkg==1.5
appnope==0.1.0
argh==0.26.2
asn1crypto==0.24.0
async-generator==1.10
async-lru==0.1.0
asyncio-cancel-token==0.1.0a2
atomicwrites==1.2.1
attrdict==2.0.0
attrs==18.2.0
Babel==2.6.0
backcall==0.1.0
bleach==3.0.2
bloom-filter==1.3
bumpversion==0.5.3
cachetools==2.1.0
certifi==2018.11.29
cffi==1.11.5
chardet==3.0.4
coincurve==8.0.2
colorama==0.4.1
coverage==4.5.2
cryptography==2.4.2
cytoolz==0.9.0.1
decorator==4.3.0
docopt==0.6.2
docutils==0.14
eth-abi==1.2.2
eth-account==0.2.3
eth-bloom==1.0.1
eth-hash==0.2.0
eth-keyfile==0.5.1
eth-keys==0.2.0b3
eth-rlp==0.1.2
eth-typing==2.0.0
eth-utils==1.3.0
execnet==1.5.0
flake8==3.5.0
hexbytes==0.1.0
hypothesis==3.69.5
idna==2.7
ifaddr==0.1.5
imagesize==1.1.0
ipython==6.5.0
ipython-genutils==0.2.0
jedi==0.13.1
Jinja2==2.10
jsonschema==2.6.0
lahja==0.9.0
lru-dict==1.1.6
lxml==4.2.5
MarkupSafe==1.1.0
mccabe==0.6.1
more-itertools==4.3.0
mypy==0.641
mypy-extensions==0.4.1
netdisco==2.2.0
netifaces==0.10.7
packaging==18.0
parsimonious==0.8.1
parso==0.3.1
pathtools==0.1.2
pexpect==4.6.0
pickleshare==0.7.5
pkginfo==1.4.2
pluggy==0.7.1
plyvel==1.0.5
prompt-toolkit==1.0.15
ptyprocess==0.6.0
py==1.7.0
py-ecc==1.4.3
[...]

PR #1649, opened 2018-12-19
Run plugin discovery test only with --integration option

cburgdorf:
### What was wrong?

Our integration test that the installation of external plugins actually works is getting is often getting in the way for people new to the code base as seen in #1646, #1547 and #1539

The problem with this test is that nobody yet has found a way to do the plugin installation from right within pytest in a way that allows to dynamically import the plugin after the installation. That's why for that test, we rely on `tox` to install the plugin.

https://github.com/ethereum/py-evm/blob/48be6ca6bea5c6349ff3e00e477ce0db8305b43f/tox.ini#L75

But for people unaware of this, it means they try to run `pytest` after a fresh clone of the repo and the tests just breaks because of the missing import.

### How was it fixed?

Only run this test when the `--integration` parameter is given, so effectively only run it during CI.

#### Cute Animal Picture

### DIFF
--- a/tests/trinity/integration/test_plugin_discovery.py
+++ b/tests/trinity/integration/test_plugin_discovery.py
@@ -1,12 +1,17 @@
+import pytest
 from trinity.plugins.registry import (
     discover_plugins
 )
-# This plugin is external to this code base and installed by tox
-# In order to install it locally run:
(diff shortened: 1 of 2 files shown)


cburgdorf:
moved to https://github.com/ethereum/trinity/pull/10

Issue #845, opened 2018-05-31
Use `NamedTuple` instead of `Dict` for Command messages


pipermerriam:
### What is wrong?

We currently are using dictionaries for the `msg` portion for commands like `GetBlockHeaders`.  It's hard to specify types for these, as well as them being inherently mutable structures.

### How can it be fixed

Commands should decode into a `NamedTuple` instead when they have key/value data.

jannikluhn:
Related question: Is there a reason why `cmd` and `msg` are separate? If not one could also think about making `Command` an `rlp.Serializable` and instantiate it for every message.

javimaravillas:
I will take a look on this one.

carver:
Here is an example (maybe the only place) where the message dict is created: https://github.com/ethereum/py-evm/blob/b10f1edc4a9a316f72ce3905a62219ba6ca79b90/p2p/protocol.py#L98-L102

The RLP is already enforcing the number of elements there. So I guess what we could do is define a `NamedTuple` on each command, and then return an instance of that `NamedTuple` instead of the `dict`. It would be nice if we could figure out a way to not have to define the field names twice. Something like:

```py
class GetBlockBodiesMessage(NamedTuple):
    request_id: int
    block_hashes: List[bytes]

class GetBlockBodies(Command):
    _cmd_id = 4
    message_class = GetBlockBodiesMessage
    deserialize_types = [
        sedes.big_endian_int,
        sedes.CountableList(sedes.binary),
    ]

```

The thing I like least about that approach is that the name/type connection is implicit (based on ordering). Definitely happy to see alternatives.

PR #1650, opened 2018-12-19
BCC syncing

jannikluhn:
This implements syncing of the beacon chain according to the following basic strategy:

- Select the peer with the highest announced slot number.
- Request blocks starting with the last finalized block.
- Stop when the peer does not provide any more blocks (i.e. we've reached their head or something went wrong).

Non-beacon chain code I touched: I've added an empty `TypedDict` as a fourth option for p2p payload types in order to use my `StatusMessage`, `BeaconBlocksMessage`, and `GetBeaconBlocksMessage` properly. For some reason, this meant that I had to add explicit type hints for "normal" dictionaries in a few instances (e.g. [here](https://github.com/ethereum/py-evm/compare/ethereum:48be6ca...jannikluhn:724bd22#diff-2fe3625b5324b16f88b7a92761437638R107)). Otherwise mypy expects dictionaries without entries (even though `Dict[str, Any]` is still allowed). Referencing #845 

#### Cute Animal Picture


trinity/protocol/bcc/validators.py
```diff
@@ -0,0 +1,86 @@
...
+                f"Requested up to {self.max_blocks} blocks but received {len(blocks)}"
+            )
+
+    def _validate_sequence(self, blocks: Tuple[BaseBeaconBlock, ...]) -> None:
+        # workaround for https://github.com/pytoolz/cytoolz/issues/123#issuecomment-432905716
+        if not blocks:
+            return
+
+        for parent, child in sliding_window(2, blocks):
+            if child.parent_root != parent.hash:
```
pipermerriam: This appears to be different from the `eth` protocol, requiring that responses be contiguous (no gaps allowed).  Is that accurate?  Might be nice to add a comment indicating what exactly is being validated at each of these steps in human readable terms like we do here: https://github.com/ethereum/py-evm/blob/master/trinity/protocol/common/validators.py#L130

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,145 @@
...
+            cast(BCCPeer, peer)
+            for peer in self.peer_pool.connected_nodes.values()
+        ]
+
+        # chooses the peer with the highest head slot
+        sorted_peers = sorted(peers, key=lambda peer: peer.head_slot, reverse=True)
+        best_peer = first(sorted_peers)
+        return best_peer
+
+    def sync_peer_useful(self) -> bool:
```
pipermerriam: I think this method is a bit odd.

1) I'd expect this logic to live in `select_sync_peer`
2) **if** kept separate from `select_sync_peer`, I'd expect it to take a `peer` as an argument and operate on that, rather than relying on internal state.  This is particularly important if there will be functionality to switch which peer we're syncing against mid-sync.

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,145 @@
...
+
+        await self.sync()
+
+        new_head = self.chain_db.get_canonical_head()
+        self.logger.info(f"Sync with {self.sync_peer} finished, new head: {new_head}")
+
+    def select_sync_peer(self) -> BCCPeer:
+        if len(self.peer_pool) == 0:
+            raise ValueError("No peers to sync with")
+        peers = [
```
pipermerriam: `sorted` works fine with generators.  So you can convert this to be something like `(peer for peer in self.peer_pool.connected_nodes.values())`

Also, it might be a little cleaner to do `cast(Iterable[BCCPeer], peers)` assuming that works well enough with `mypy`.

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,145 @@
...
+            cast(BCCPeer, peer)
+            for peer in self.peer_pool.connected_nodes.values()
+        ]
+
+        # chooses the peer with the highest head slot
+        sorted_peers = sorted(peers, key=lambda peer: peer.head_slot, reverse=True)
```
pipermerriam: Some quick profiling suggest using `operator.attrgetter` is about 20% faster than the `lambda`

trinity/sync/beacon/chain.py
```diff
@@ -0,0 +1,145 @@
...
+            self.sync_peer.head_slot,
+            self.chain_db.get_finalized_head().slot,
+        )
+
+        start_slot = self.chain_db.get_finalized_head().slot + 1
+        batches = self.request_batches(start_slot)
+
+        last_block = None
+        async for batch in batches:
+            if last_block is None:
```
pipermerriam: This seems like a hack to do something special on the *first* item.  Maybe change to something like:

```python
async def async_enumerate(iterable, start_idx=None):
    counter = itertools.counter(start_idx)
    async for value in iterable:
        yield next(counter), value

    async for (idx, batch) in async_enumerate(batches):
        if idx == 0:
            ....
```

A more ideal solution ignoring the async stuff would be:

```python
for previous_batch, batch in sliding_window(2, concatv([None], batches)):
    if previous_batch is None:
        self.validate_first_batch(batch)
    elif batch[0].parent_root != previous_batch[-1].hash:
        ...
```

But the async nature of `batches` complicates this approach.  I suspect we could generate some basic machinery to handle that though and I think the resulting body of this `for` loop could be cleaner.  I'll leave it to you to decide what (if anything) you'd like to do here.

p2p/protocol.py
```diff
@@ -28,10 +32,12 @@
 if TYPE_CHECKING:
     from p2p.peer import BasePeer  # noqa: F401
 
+TypedDictPayloadType = TypedDict("TypedDictPayloadType", {})
```
cburgdorf: nitpick: `TypedPayloadDict` would be more inline with the rest of the names in the code base. In general, we don't use a `Type` suffix for type names except we try to express that something is modelling the *kind* / *category* / *type* of something.

e.g.

```python
class CarType(enum):
    SportsCar = auto()
    Van = auto()
```

But apart from such cases we would have classes such as `Animal`, `Building`, `Food` and not `AnimalType`, `BuildingType` or `FoodType`.

trinity/protocol/bcc/validators.py
jannikluhn: Yes, it's because we don't have `reverse` and `skip`, so we can be a bit more strict here. I've added some comments.

pipermerriam:

jannikluhn:
Moved to https://github.com/ethereum/trinity/pull/163

PR #1654, opened 2018-12-20
Share the app directory across multiple programs

cburgdorf:
### What was wrong?

The `trinity-beacon` adds support for a new type of client that we need to support running side by side with the existing Ethereum 1 client. This client will have entirely different data, logs, ipc and pid files. 

Based on the discussion in #1603 we want to support running this client out of the same application directory as the Ethereum 1 client which means the application directory layout needs to be modified to support this.

### How was it fixed?

This change introduces multi-app support to trinity by introducing a new `app_identifier` config to the `TrinityConfig`.

Assuming there Ethereum 1 client uses an `app_identifier` of `""`(empty string)  and the Ethereum 2 beacon node uses an `app_identifier` of `beacon` the directory layout looks as follows.

```
/home/cburgdorf/.local/share/trinity
├── mainnet                       <-- one mainnet directory, but app specific sub directories
│   ├── chain
│   │   ├── full
|   |   |     └── db files
│   │   └── light
|   |         └── db files
│   ├── chain-beacon
│   │   ├── full
|   |   |     ├── beacon
│   │   │     |     └── db files
|   |   |     ├── shard-1
│   │   │     |     └── db files
|   |   |     └── shard-n
│   │   │           └── db files
│   │   └── light
|   |         ├── beacon
│   │         |     └── db files
|   |         ├── shard-1
│   │         |     └── db files
|   |         └── shard-n
│   │               └── db files
│   ├── logs
│   │   ├── trinity.log
│   │   ├── ...
│   │   ├── trinity.log.7
│   ├── logs-beacon
│   │   └── trinity.log
│   │   ├── ...
│   │   └── trinity.log.7
│   ├── ipc
│   │   ├── db.ipc
│   │   └── ...
│   ├── ipc-beacon
│   │   ├── db.ipc
│   │   └── ...
│   ├── pid
│   │   ├── database.pid
│   │   └── ...
│   ├── pid-beacon
│   │   ├── database.pid
│   │   └── ...
│   └── nodekey
|
└── ropsten
    ├── chain
    │   ├── full
    |   |     └── db files
    │   └── light
    |         └── db files
    ├── chain-beacon
    │   ├── full
    |   |     ├── beacon
    │   │     |     └── db files
    |   |     ├── shard-1
    │   │     |     └── db files
    |   |     └── shard-n
    │   │           └── db files
    │   └── light
    |         ├── beacon
    │         |     └── db files
    |         ├── shard-1
    │         |     └── db files
    |         └── shard-n
    │               └── db files
    ├── logs
    │   ├── trinity.log
    │   ├── ...
    │   ├── trinity.log.7
    ├── logs-beacon
    │   └── trinity.log
    │   ├── ...
    │   └── trinity.log.7
    ├── ipc
    │   ├── db.ipc
    │   └── ...
    ├── ipc-beacon
    │   ├── db.ipc
    │   └── ...
    ├── pid
    │   ├── database.pid
    │   └── ...
    ├── pid-beacon
    │   ├── database.pid
    │   └── ...
    └── nodekey
```

#### Cute Animal Picture

### DIFF
--- a/trinity/config.py
+++ b/trinity/config.py
@@ -37,7 +37,11 @@ from p2p.constants import (
 from trinity.constants import (
     ASSETS_DIR,
     DEFAULT_PREFERRED_NODES,
+    IPC_DIR,
+    LOG_DIR,
+    LOG_FILE,
     MAINNET_NETWORK_ID,
+    PID_DIR,
     ROPSTEN_NETWORK_ID,
     SYNC_FULL,
     SYNC_LIGHT,
@@ -47,7 +51,6 @@ from trinity._utils.chains import (
     get_data_dir_for_network_id,
     get_database_socket_path,
     get_jsonrpc_socket_path,
-    get_logfile_path,
     get_nodekey_path,
     load_nodekey,
 )
@@ -72,7 +75,6 @@ if TYPE_CHECKING:
     from trinity.chains.light import LightDispatchChain  # noqa: F401
 
 DATABASE_DIR_NAME = 'chain'
-
 
 MAINNET_EIP1085_PATH = ASSETS_DIR / 'eip1085' / 'mainnet.json'
 ROPSTEN_EIP1085_PATH = ASSETS_DIR / 'eip1085' / 'ropsten.json'
@@ -216,18 +218,19 @@ class TrinityConfig:
 
     def __init__(self,
                  network_id: int,
+                 app_identifier: str="",
                  genesis_config: Dict[str, Any]=None,
                  max_peers: int=25,
                  trinity_root_dir: str=None,
                  data_dir: str=None,
                  nodekey_path: str=None,
-                 logfile_path: str=None,
                  nodekey: PrivateKey=None,
                  sync_mode: str=SYNC_FULL,
                  port: int=30303,
                  use_discv5: bool = False,
                  preferred_nodes: Tuple[KademliaNode, ...]=None,
                  bootstrap_nodes: Tuple[KademliaNode, ...]=None) -> None:
+        self.app_identifier = app_identifier
         self.network_id = network_id
         self.max_peers = max_peers
         self.sync_mode = sync_mode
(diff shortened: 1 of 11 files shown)


trinity/config.py
```diff
@@ -216,18 +218,19 @@ class TrinityConfig:
 
     def __init__(self,
                  network_id: int,
+                 app_identifier: str="",
                  genesis_config: Dict[str, Any]=None,
                  max_peers: int=25,
                  trinity_root_dir: str=None,
                  data_dir: str=None,
                  nodekey_path: str=None,
-                 logfile_path: str=None,
```
cburgdorf: Maybe at one point we allowed setting the `logfile_path` vial CLI but it doesn't seem like that functionality still exists. It seems to be dead code that was never called from anywhere.

trinity/constants.py
```diff
@@ -16,9 +16,16 @@
 from p2p.kademlia import Address, Node
 
 
+# application identifier
+APP_IDENTIFIER_ETH1 = ""  # An empty identifier ensures backwards compatibility of data directory
```
pipermerriam: What are the implications of not preserving this backwards compat?  Just that previous PID files won't be in the *right* location or more than that?

Assuming the former, I'd prefer we go ahead and make this `eth1` rather than accumulate debt in an alpha codebase.

cburgdorf: The `chain` directory would turn into `chain-eth1` and the logs directory would turn into `logs-eth1`. So, not a big deal. We can put that in the release notes and people can move files manually if they care. That's fine with me if you prefer that.

pipermerriam: > That's fine with me if you prefer that.

:+1:

trinity/config.py
carver: Then we should definitely remove the code that crashes if the log folder is missing

cburgdorf:
This has been moved to https://github.com/ethereum/trinity/pull/8

PR #1658, opened 2018-12-20
[WIP] Add slashable helper

### LINKED ISSUE #1636: Implement `verify_slashable_vote_data` helper
### What is wrong?
See spec: https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#verify_slashable_vote_data

### How can it be fixed
1. Hint:
    * `aggregate_pubkey`:  `eth.utils.bls.aggregate_pubkey`
    * `bls_verify_multiple`: `eth.utils.bls.verify_multiple`
2. Implement `SlashableVoteData.root` instead of using WIP `hash_tree_root`.
    * Leave a note: `TODO: use SSZ hash_tree_root() funtion`
3. Py-EVM-ize it. ✨
4. Write tests.

ralexstokes:
### What was wrong?

Fixes #1636.

Missing a helper function to verify a `SlashableVoteData` instance.

### How was it fixed?

Implemented the function and required auxillary functionality to get it working with current codebase. In particular, we are using a flat hash in lieu of the yet-to-be-stabilized `ssz` tree hash.

#### Cute Animal Picture


ralexstokes:
This is a WIP PR and lacks any tests of the new functionality which will be arriving shortly.

Figured I could open in the meantime to get review (@hwwhww @djrtwo)

eth/beacon/helpers.py
```diff
@@ -402,6 +404,32 @@ def get_domain(fork_data: 'ForkData',
         slot,
     ) * 4294967296 + domain_type
 
+def verify_slashable_vote_data(state: BeaconState, vote_data: SlashableVoteData) -> bool:
```
hwwhww: Since you're using `TYPE_CHECKING`, here should be:
```suggestion
def verify_slashable_vote_data(state: BeaconState, vote_data: 'SlashableVoteData') -> bool:
```

eth/beacon/helpers.py
```diff
@@ -402,6 +404,32 @@ def get_domain(fork_data: 'ForkData',
...
+        vote_data_root + bytes1(1)
+    ]
+    return bls.verify_multiple(
+        pubkeys=pubkeys,
+        messages=messages,
+        signature=vote_data.aggregate_signature,
+        domain=get_domain(
+            state.fork_data,
+            state.slot,
+            DOMAIN_ATTESTATION,
```
hwwhww: Needs to import `SignatureDomain`, which is inside `if TYPE_CHECKING:` currently. Change to:
```python
from eth.beacon.enums import (
    SignatureDomain,
)
```

And here we do:
```suggestion
            SignatureDomain.DOMAIN_ATTESTATION,
```

eth/beacon/helpers.py
```diff
@@ -402,6 +404,32 @@ def get_domain(fork_data: 'ForkData',
...
 
+def verify_slashable_vote_data(state: BeaconState, vote_data: SlashableVoteData) -> bool:
+    vote_count = len(vote_data.aggregate_signature_poc_0_indices) + len(vote_data.aggregate_signature_poc_1_indices)
+    if vote_count > MAX_CASPER_VOTES:
+        return False
+
+    pubkeys = [
+        bls.aggregate_pubkeys([state.validators[i].pubkey for i in vote_data.aggregate_signature_poc_0_indices]),
+        bls.aggregate_pubkeys([state.validators[i].pubkey for i in vote_data.aggregate_signature_poc_1_indices])
+    ]
```
hwwhww: In Py-EVM we like `tuple` more than `list` for immutability. 

Refactoring idea 1 - Formatting
```python
pubkeys = tuple(
    tuple(
        state.validators[i].pubkey
        for i in vote_data.aggregate_signature_poc_0_indices
    ),
    tuple(
        state.validators[i].pubkey
        for i in vote_data.aggregate_signature_poc_1_indices
    ),
)
```

Refactoring idea 2 - Adding new helper function
```python
@to_tuple
def get_pubkey_from_indices(validators: Sequence[ValidatorRecord], indices: Sequence[int]) -> Iterable[int]:
    for index in vote_data.aggregate_signature_poc_0_indices]:
        yield validators[index]
```

eth/beacon/helpers.py
```diff
@@ -402,6 +404,32 @@ def get_domain(fork_data: 'ForkData',
...
+
+    # TODO: change to hash_tree_root(vote_data) when we have SSZ tree hashing
+    vote_data_root = vote_data.root
+    messages = [
```
hwwhww:
```python
messages = (
    ...
)
```

eth/beacon/helpers.py
ralexstokes: awesome! i was actually right in the middle of adding it as a parameter to the function... like this:

```python
def verify_slashable_vote_data(state: BeaconState,
                               vote_data: 'SlashableVoteData',
                               max_casper_votes: int,
                               signature_domain_attestation: 'SignatureDomain') -> bool:
```

Which do we prefer?

eth/beacon/helpers.py
ralexstokes: i've seen this `@to_tuple` decorator... i'll look into this method

eth/beacon/helpers.py
hwwhww: Oh, right right, we need to pass constants like `max_casper_votes` for testing. For example, we can set `shard_count` to 10 instead of 1024 for lightening the tests.

On the other hand, `SignatureDomain` is one of the enum constants that (i) we use as flags
(ii) similar the ID of the opcode, so I *think* it's unlikely to be changed in future forks.

So use:
```python
def verify_slashable_vote_data(state: BeaconState,
                               vote_data: 'SlashableVoteData',
                               max_casper_votes: int) -> bool:
```

eth/beacon/helpers.py
```diff
@@ -408,38 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+
+
+def verify_vote_count(vote_data: 'SlashableVoteData', max_casper_votes: int) -> bool:
+    """
+    Ensures we have no more than ``max_casper_votes`` in the ``vote_data``.
+    """
+    return vote_data.vote_count <= max_casper_votes
+
+
+def verify_signature(state: BeaconState, vote_data: 'SlashableVoteData') -> bool:
```
ralexstokes: i feel like this function name is fairly generic given what it is doing -- either we should specialize the function name or generalize the arguments (which really just gets us to `bls.verify_multiple` so not ideal); this function was named `verify_signature_for_slashable_vote_data` but the linter did not like the line length on 479.

anyone have any preference or ideas on this?

eth/beacon/helpers.py
```diff
@@ -408,38 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+    """
     proof_of_custody_0_indices = vote_data.aggregate_signature_poc_0_indices
     proof_of_custody_1_indices = vote_data.aggregate_signature_poc_1_indices
-    vote_count = len(proof_of_custody_0_indices) + len(proof_of_custody_1_indices)
-    if vote_count > max_casper_votes:
-        return False
+    all_indices = (proof_of_custody_0_indices, proof_of_custody_1_indices)
+    get_pubkeys = functools.partial(get_pubkey_for_indices, state.validators)
+    return map(bls.aggregate_pubkeys,
+               map(get_pubkeys, all_indices))
```
ralexstokes: general note: i'll lean towards a functional style where i can -- if any contributor wants to step in and nudge me towards what some consider a more pythonic style, i'm all ears :) i think it's important to follow the established repo guidelines but will bend the rules where i can to get what imo is more expressive code ;)

eth/beacon/helpers.py
```diff
@@ -403,6 +408,77 @@ def get_domain(fork_data: 'ForkData',
     ) * 4294967296 + domain_type
 
 
+@to_tuple
+def get_pubkey_for_indices(validators: Sequence[ValidatorRecord],
```
hwwhww:
```suggestion
def get_pubkey_for_indices(validators: Sequence['ValidatorRecord'],
```

eth/beacon/helpers.py
```diff
@@ -403,6 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+    vote_data_root = vote_data.root
+    return (
+        vote_data_root + (0).to_bytes(1, 'big'),
+        vote_data_root + (1).to_bytes(1, 'big'),
+    )
+
+
+def verify_vote_count(vote_data: 'SlashableVoteData', max_casper_votes: int) -> bool:
+    """
+    Ensures we have no more than ``max_casper_votes`` in the ``vote_data``.
```
hwwhww: Nitpick: [We use imperative, present tense to describe APIs](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/documentation.md#grammar). So:
```suggestion
    Ensure we have no more than ``max_casper_votes`` in the ``vote_data``.
```

eth/beacon/helpers.py
```diff
@@ -403,6 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+    """
+    return vote_data.vote_count <= max_casper_votes
+
+
+def verify_signature(state: BeaconState, vote_data: 'SlashableVoteData') -> bool:
+    """
+    Ensures we have a valid aggregate signature for the ``vote_data``.
```
hwwhww: ditto: Ensure

eth/beacon/helpers.py
```diff
@@ -403,6 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+    """
+    proof_of_custody_0_indices = vote_data.aggregate_signature_poc_0_indices
+    proof_of_custody_1_indices = vote_data.aggregate_signature_poc_1_indices
+    all_indices = (proof_of_custody_0_indices, proof_of_custody_1_indices)
+    get_pubkeys = functools.partial(get_pubkey_for_indices, state.validators)
+    return map(bls.aggregate_pubkeys,
+               map(get_pubkeys, all_indices))
+
+
+def build_vote_data_messages(vote_data: 'SlashableVoteData') -> Tuple[bytes, bytes]:
```
hwwhww: I think it can be moved to be a property of `SlashableVoteData`:

In `class SlashableVoteData`, add:
```python
@property
def create_vote_data_messages(self) -> Tuple[bytes, bytes]:
    """
    Builds the messages that validators are expected to sign for a ``CasperSlashing`` operation.
    """
    # TODO: change to hash_tree_root(vote_data) when we have SSZ tree hashing
    vote_data_root = self.root
    return (
        vote_data_root + (0).to_bytes(1, 'big'),
        vote_data_root + (1).to_bytes(1, 'big'),
    )
```

eth/beacon/helpers.py
```diff
@@ -403,6 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+        signature=signature,
+        domain=domain,
+    )
+
+
+def verify_slashable_vote_data(state: BeaconState,
+                               vote_data: 'SlashableVoteData',
+                               max_casper_votes: int) -> bool:
+    """
+    Ensures that the ``vote_data`` is properly assembled and contains the signature
```
hwwhww: ditto: Ensure

eth/beacon/helpers.py
```diff
@@ -403,6 +408,77 @@ def get_domain(fork_data: 'ForkData',
...
+    """
+    Computes the aggregate pubkey we expect based on
+    the proof-of-custody indices found in the ``vote_data``.
+    """
+    proof_of_custody_0_indices = vote_data.aggregate_signature_poc_0_indices
+    proof_of_custody_1_indices = vote_data.aggregate_signature_poc_1_indices
+    all_indices = (proof_of_custody_0_indices, proof_of_custody_1_indices)
+    get_pubkeys = functools.partial(get_pubkey_for_indices, state.validators)
+    return map(bls.aggregate_pubkeys,
+               map(get_pubkeys, all_indices))
```
hwwhww: 👍 on using functools. Just nitpick the formatting:

```python
    return map(
        bls.aggregate_pubkeys,
        map(get_pubkeys, all_indices),
    )
```

eth/beacon/helpers.py
hwwhww: `verify_slashable_vote_data_signature`?
For line 479, you can do:
```python
    return (
        verify_vote_count(vote_data, max_casper_votes) and
        verify_slashable_vote_data_signature(state, vote_data)
    )
```

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+
+    num_validators = len(validators)
+    num_indices = 5
+
+    key = "aggregate_signature_poc_0_indices"
+    (poc_0_indices, poc_0_signatures) = _get_indices_and_signatures(num_validators,
+                                                                    num_indices,
+                                                                    validators,
+                                                                    messages[0],
+                                                                    privkeys)
```
hwwhww: nitpick of formatting: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#closing-parenthesisbracketbrace

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+    valid_params[key] = poc_0_indices
+
+    key = "aggregate_signature_poc_1_indices"
+    # NOTE: does not guarantee non-empty intersection
+    (poc_1_indices, poc_1_signatures) = _get_indices_and_signatures(num_validators,
+                                                                    num_indices,
+                                                                    validators,
+                                                                    messages[1],
+                                                                    privkeys)
```
hwwhww: nitpick of formatting: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#closing-parenthesisbracketbrace

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+    state = BeaconState(**sample_beacon_state_params)
+
+    # NOTE: we can do this before "correcting" the params as they
+    # touch disjoint subsets of the provided params
+    messages = _create_slashable_vote_data_messages(sample_slashable_vote_data_params)
+
+    valid_params = _correct_slashable_vote_data_params(sample_slashable_vote_data_params,
+                                                       genesis_validators,
+                                                       messages,
+                                                       privkeys)
```
hwwhww: nitpick of formatting: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#closing-parenthesisbracketbrace

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+    state = BeaconState(**sample_beacon_state_params)
+
+    # NOTE: we can do this before "correcting" the params as they
+    # touch disjoint subsets of the provided params
+    messages = _create_slashable_vote_data_messages(sample_slashable_vote_data_params)
+
+    params = _correct_slashable_vote_data_params(sample_slashable_vote_data_params,
+                                                 genesis_validators,
+                                                 messages,
+                                                 privkeys)
```
hwwhww: nitpick of formatting: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#closing-parenthesisbracketbrace

eth/beacon/types/slashable_vote_data.py
```diff
@@ -38,3 +43,39 @@ def __init__(self,
...
+
+    _vote_count = None
+
+    @property
+    def vote_count(self) -> int:
+        if self._vote_count is None:
+            count_zero_indices = len(self.aggregate_signature_poc_0_indices)
+            count_one_indices = len(self.aggregate_signature_poc_1_indices)
+            self._vote_count = count_zero_indices + count_one_indices
+        return self._vote_count
```
hwwhww: I like it. 👍

eth/beacon/types/states.py
```diff
@@ -158,6 +158,10 @@ def num_validators(self) -> int:
     def num_crosslinks(self) -> int:
         return len(self.latest_crosslinks)
 
+    @property
+    def validators(self) -> Sequence[ValidatorRecord]:
+        return self.validator_registry
```
hwwhww: I don't think we need this alias. It might be confusing for the readers about which is the RLP/SSZ field and which is settable.

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
     )
 
 
+@given(st.data())
+def test_get_pubkey_for_indices(genesis_validators, data):
+    max_value_for_list = len(genesis_validators) - 1
+    indices = data.draw(st.lists(st.integers(min_value=0,
+                                             max_value=max_value_for_list)))
```
hwwhww: nitpick formatting:
```python
    indices = data.draw(
        st.lists(
            st.integers(
                min_value=0,
                max_value=max_value_for_list,
            )
        )
    )
```

tests/beacon/test_helpers.py
```diff
@@ -34,10 +39,20 @@
...
+    verify_slashable_vote_data,
     is_double_vote,
     is_surround_vote,
 )
+import eth._utils.bls as bls
 
+from hypothesis import (
+    given,
+    strategies as st,
+)
```
hwwhww: nitpick: move `hypothesis` importing to about L6 (between `import random` and `from eth_utils import (...)` because it's a third-party package.  (https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#overall-ordering)

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+    i = data.draw(st.integers(min_value=0, max_value=max(len(xs) - 1, 0)))
+    return (xs, i)
+
+
+@given(st.data())
+def test_generate_aggregate_pubkeys(genesis_validators, sample_slashable_vote_data_params, data):
+    max_value_for_list = len(genesis_validators) - 1
+    (indices, some_index) = _list_and_index(data,
+                                            elements=st.integers(min_value=0,
+                                                                 max_value=max_value_for_list))
```
hwwhww: formatting:
```python
    (indices, some_index) = _list_and_index(
        data,
        elements=st.integers(
            min_value=0,
            max_value=max_value_for_list,
        )
    )
```

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+                                             max_value=max_value_for_list)))
+    pubkeys = get_pubkey_for_indices(genesis_validators, indices)
+    all_pubkeys = tuple(
+        map(lambda validator: validator.pubkey, genesis_validators)
+    )
+
+    assert len(indices) == len(pubkeys)
+
+    for pubkey in pubkeys:
+        assert pubkey in all_pubkeys
```
hwwhww: This might be more meaningful:
```python
for index, pubkey in enumerate(pubkeys):
    assert genesis_validators[index].pubkey == pubkey
```

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+    '''
+    Randomly select some validator indices
+    '''
+    indices = []
+    for i in range(count):
+        next_index = random.randint(0, max - 1)
+        while next_index in indices:
+            next_index = random.randint(0, max - 1)
+        indices.append(next_index)
+    return indices
```
hwwhww: It seems you can use `random.sample` or [`hypothesis.strategies.random.sample`](https://www.programcreek.com/python/example/434/random.sample) to return unique elements in some use cases.

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+
+
+def _get_indices_and_signatures(num_validators,
+                                num_indices,
+                                validators,
+                                message,
+                                privkeys):
+    indices = _select_indices(num_validators, num_indices)
```
hwwhww: could be replaced by:
```python
assert num_validators >= num_indices
indices = random.sample(range(num_validators), num_indices)
```

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,224 @@ def test_get_domain(pre_fork_version,
...
+    assert not verify_slashable_vote_data_signature(state, invalid_votes)
+
+
+def _run_verify_slashable_vote(params, state, max_casper_votes, should_fail):
+    votes = SlashableVoteData(**params)
+    result = verify_slashable_vote_data(state, votes, max_casper_votes)
+    if should_fail:
+        assert not result
+    else:
+        assert result
```
hwwhww: I *think* using something like `success` or `expected` would be more clear than using `should_fail` to assert the result?

djrtwo: Agreed. Should generally avoid negative boolean names.

eth/beacon/types/states.py
ralexstokes: fair enough

tests/beacon/test_helpers.py
ralexstokes: gotcha good call

tests/beacon/test_helpers.py
ralexstokes: yeah this is much better! like it

tests/beacon/test_helpers.py
ralexstokes: ah nice -- thanks for calling out `random.sample`. it is what i would use although to your other point i'll use `hypothesis` to be consistent although i can't find `hypothesis.strategies.random.sample` so i'll rely on the indices generation used in other places in the test file

tests/beacon/test_helpers.py
ralexstokes: will change!

tests/beacon/test_helpers.py
ralexstokes: ended up going this route -- using hypothesis to sample param space *and* process signatures takes too long

ralexstokes:
@hwwhww included changes prompted by the comments! thanks!!

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,228 @@ def test_get_domain(pre_fork_version,
...
+
+
+@pytest.mark.parametrize(
+    (
+        'param_mapper,'
+        'should_succeed'
+    ),
+    [
+        (lambda params: params, True),
+        (lambda params: _corrupt_vote_count(params), False),
```
pipermerriam: this `lambda` isn't needed as I am pretty sure you could just provide `_corrupt_vote_count` directly.

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,228 @@ def test_get_domain(pre_fork_version,
...
+        (lambda params: _corrupt_signature(params), False),
```
pipermerriam: this `lambda` isn't needed as I am pretty sure you could just provide `_corrupt_signature` directly.

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,228 @@ def test_get_domain(pre_fork_version,
...
+    params["aggregate_signature"] = bls.sign(message, privkey, domain)
+    return params
+
+
+def _corrupt_vote_count(params):
+    params = copy.deepcopy(params)
+    key = "aggregate_signature_poc_0_indices"
+    for i in itertools.count():
+        if i not in params[key]:
+            params[key].append(i)
```
pipermerriam: Here's how you can do this without the mutation.

```python
from toolz import assoc

def _corrupt_vote_count(params):
    for i in itertools.count():
        if i not in params[key]:
           return assoc(params, key, params[key] + [i])
    else:
        raise SomeException("Unreachable code path")
```

It's minor but it's nice to not mutate things.  If the `deepcopy` call somehow got dropped this would begin modifying the original `params` which isn't ideal.

tests/beacon/test_helpers.py
```diff
@@ -707,6 +722,228 @@ def test_get_domain(pre_fork_version,
...
+
+    return valid_params
+
+
+def _corrupt_signature(params):
+    params = copy.deepcopy(params)
+    message = bytes.fromhex("deadbeefcafe")
+    privkey = 42
+    domain = SignatureDomain.DOMAIN_ATTESTATION
+    params["aggregate_signature"] = bls.sign(message, privkey, domain)
```
pipermerriam: See other example for how this mutation can be removed.

pipermerriam:
@ralexstokes I opened https://github.com/ethereum/py-evm/issues/1696 to track the cleanup issues I just commented on.  Not high priority, just wanted to track this so it doesn't get lost.

ralexstokes:
@pipermerriam i'll take a look now

eth/beacon/types/slashable_vote_data.py
```diff
@@ -38,3 +43,39 @@ def __init__(self,
...
+
+    @property
+    def messages(self) -> Tuple[bytes, bytes]:
+        """
+        Build the messages that validators are expected to sign for a ``CasperSlashing`` operation.
+        """
+        # TODO: change to hash_tree_root(vote_data) when we have SSZ tree hashing
+        vote_data_root = self.root
+        return (
+            vote_data_root + (0).to_bytes(1, 'big'),
```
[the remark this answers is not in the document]
ralexstokes: agreed -- this was outdated with a more recent spec change already implemented here though: https://github.com/ethereum/py-evm/commit/f1a0b206904f1baf8528b88bf60d84cd7a989621

tests/beacon/test_helpers.py
ralexstokes: we can -- i think it breaks the symmetry of the expression but there is some gain to be had by avoiding allocation of the `lambda` object.

tests/beacon/test_helpers.py
ralexstokes: ditto above

tests/beacon/test_helpers.py
ralexstokes: yeah this is nice and i agree safer than what was there w/ `copy.deepcopy`.

eth/beacon/helpers.py
pipermerriam: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md

If you haven't seen that yet it outlines a lot of our style guides but we also are all about rule bending when the situation calls for it.

:heart: functional style

PR #1660, opened 2018-12-20
rename tests to avoid naming collision that was causing pytest to fail

ralexstokes:
### What was wrong?

Fixes remainder of #1646.

Name collision was causing `pytest` to fail.

### How was it fixed?

Rename the tests as was suggested here:

https://github.com/ethereum/py-evm/issues/1646#issuecomment-448686503

#### Cute Animal Picture


pipermerriam:
yellow CI is rate limit issue.

Issue #1657, opened 2018-12-20
Trinity prints noisy error message when being port-scanned in "light client" mode


veox:
* py-evm Version: `master` (commit 8841a6980254b93294c80a889ed3901fd6baf9c6)
* OS: Arch Linux
* Python Version (python --version): 3.7.1
* Environment (output of `pip freeze`): MISSING

### What is wrong?

Running `trinity` as

``` sh
% nice ionice -c3 trinity --network-id=1 --sync-mode=light --discv5 $ETHSTATS_OPTIONS
```

and then checking if the port is accessible with

``` sh
% nc -vz localhost 30303
```

results in this error message being printed:

``` python
   ERROR  12-20 17:58:10           LightServer  Unexpected error handling handshake
Traceback (most recent call last):
  File "/home/veox/src/py-evm/trinity/server.py", line 185, in receive_handshake
    await self._receive_handshake(reader, writer)
  File "/home/veox/src/py-evm/trinity/server.py", line 199, in _receive_handshake
    ip, socket, *_ = writer.get_extra_info("peername")
TypeError: cannot unpack non-iterable NoneType object
```

Not sure if fresh or regression.

### How can it be fixed

Not sure.

Checking how "regular" sync mode reacts to this and handles it would be a good start.

veox:
The "fast" sync variant has:

``` python
   ERROR  12-20 18:54:08            FullServer  Unexpected error handling handshake
Traceback (most recent call last):
  File "/home/veox/src/py-evm/trinity/server.py", line 185, in receive_handshake
    await self._receive_handshake(reader, writer)
  File "/home/veox/src/py-evm/trinity/server.py", line 199, in _receive_handshake
    ip, socket, *_ = writer.get_extra_info("peername")
TypeError: cannot unpack non-iterable NoneType object
```

(Same code path.)

pipermerriam:
So we use this API:

https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info

Which delegates to this API

https://docs.python.org/3/library/socket.html#socket.socket.getpeername

It looks like we just need to implement some error handling there for the case where that returns `None` and probably log some information about the connection for debug purposes.

veox:
@pipermerriam Works for me. ^_^ (Replied in PR.)

PR #1661, opened 2018-12-20
Handle errors on incoming connections

pipermerriam:
### What was wrong?

fixes: https://github.com/ethereum/py-evm/issues/1657

When an incoming connection's socket doesn't have address information an exception escapes the Trinity process.

Also, we end up leaving the TCP transports open.

### How was it fixed?

- Adds handling for the case where the incoming connection doesn't have the proper address info and log what data we can about the connection.
- Adds handling for the error cases while receiving the handshake to close the TCP transport.

#### Cute Animal Picture

441ab30cbf83ea412ec0ebb34bec6fc0

### DIFF
--- a/trinity/server.py
+++ b/trinity/server.py
@@ -181,14 +181,22 @@ class BaseServer(BaseService, Generic[TPeerPool]):
             HandshakeFailure,
             asyncio.IncompleteReadError,
         )
+
+        def cleanup_reader_and_writer():
+            if not reader.at_eof():
+                reader.feed_eof()
+            writer.close()
+
         try:
             await self._receive_handshake(reader, writer)
         except expected_exceptions as e:
             self.logger.debug("Could not complete handshake: %s", e)
+            cleanup_reader_and_writer()
         except OperationCancelled:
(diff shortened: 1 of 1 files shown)


veox:
I've tried the PR as a patch manually in a `trinity` repo (since the `py-evm` repo no longer distributes the `trinity` executable in an easily-accessible manner).

It works fine: there are no more messages printed to the log, especially scary red errors.

trinity/server.py
```diff
@@ -181,25 +181,42 @@ def _make_request_server(self) -> BaseRequestServer:
             HandshakeFailure,
             asyncio.IncompleteReadError,
         )
+
+        def cleanup_reader_and_writer():
```
veox: [CI linting error](https://circleci.com/gh/ethereum/py-evm/119237?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link): function is missing a type annotation.

pipermerriam:
Replaced by https://github.com/ethereum/trinity/pull/104

PR #1662, opened 2018-12-20
Move SpoofTransaction out of private namespace

pipermerriam:
### What was wrong?

The `SpoofTransaction` data type was located in a private namespace.

### How was it fixed?

Move it to a public namespace so that it can be used by 3rd party libraries.

#### Cute Animal Picture

cat-in-a-bread-basket


pipermerriam:
Will address any PR feedback post-merge (need this to move #1663 forward)

PR #1664, opened 2018-12-20
Remove the `trinity` and `eth` modules as well as any related docs, tests, etc.

pipermerriam:
### What is this?

Sister issue to https://github.com/ethereum/trinity/pull/6

Removes the `p2p` and `trinity` modules now that they are in their own repository.

#### Cute Animal Picture

img_1087


.circleci/config.yml
```diff
@@ -271,116 +187,43 @@ jobs:
...
 workflows:
   version: 2
   test:
     jobs:
       - py36-docs
 
-      - py37-core
-      - py37-trinity
-      - py37-trinity-integration
-      - py37-beacon
```
pipermerriam: This needs to get added back

pipermerriam:
> Only surprise was that `stubs/eth_keys/*` was removed. It's not used anywhere in `eth/`?

That is a leftover from before type hinting worked for installed packages, however I also realize now that we need to cut a new `eth-keys` release that includes this... and update the local dependency here.

Issue #1656, opened 2018-12-20
Discussion : Remove support for multiple types in the EVM stack


Bhargavasomu:
### What is wrong?
As part of issue #1398 this new issue arised from this https://github.com/ethereum/py-evm/issues/1398#issuecomment-432303381. It was proposed to remove the different types of elements in the `stack` (`bytes` and `int`), and put only one element type in the `Stack`

### How can it be fixed

Fill this in if you know how to fix it.

Bhargavasomu:
@cburgdorf @pipermerriam @carver I see a problem with this proposed approach. Previously (current live version scenario) we are storing both `bytes` and `int`, and most of the time, it is expected that if the user requests the popped data to be in the `bytes` format, the popped data would be in the `bytes` format. But now we are entirely converting the data stored in the `stack` into a single `type`, and hence if the user asks for the other `type`, we would have to convert it into that `type` and present it to the user. I believe that this would take a lot more time. Please correct me if I am wrong.

carver:
Silly question inbound: is there any time that leading 0s matter in bytes? Like is there ever a difference between `b'\0' and b'\0' *32` on the stack? I assume not, if the EVM is simulating a fixed 32-byte instruction width.

But if so, converting all values to `int` would be problematic, of course. Having everything be `bytes` would be fine.

> I believe that this would take a lot more time.

You're talking about the time to convert the value from `int` to `bytes`? Seems relatively small compared to other costs, but the only way to know for sure is to benchmark.

pipermerriam:
> and hence if the user asks for the other type, we would have to convert it into that type and present it to the user.

Part of this change would be removing this type of conversion from the `Stack` API and requiring call sites to 1) expect to get an `int` back and 2) do any necessary conversions to `bytes` types locally.

kclowes:
Closing, not planned for current py-evm scope.

PR #1666, opened 2018-12-21
[WIP] Refactor Stack to eliminate mix types

Bhargavasomu:
### What was wrong?
As part of Issue #1656 

### How was it fixed?
The `Stack` now stores only objects of type `int`. It previously also used to store `bytes`. Now the `bytes` are converted to `int` before storing on the stack, and any the `Stack` issues out the popped elements in the form of `int`. It is expected to convert the `integers` to `bytes` at the call site, wherever necessary.

#### Cute Animal Picture

### DIFF
--- a/eth/vm/computation.py
+++ b/eth/vm/computation.py
@@ -19,11 +19,15 @@ from eth_typing import (
 )
 from eth_utils import (
     encode_hex,
+    int_to_big_endian,
 )
 
 from eth.constants import (
+    ANY,
+    BYTES,
     GAS_MEMORY,
     GAS_MEMORY_QUADRATIC_DENOMINATOR,
+    UINT256,
 )
 from eth.exceptions import (
     Halt,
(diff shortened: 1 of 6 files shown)


Bhargavasomu:
@cburgdorf  We can see the `Stack` performance comparison for the `old` version and the `new` version under `py36-stack-benchmark`. But I feel that just comparing the `Stack Times` is not a good benchmark, because we can see that performance decrease while pushing and increases while popping, when compared to the old one. So we might not still get a full idea of the performance change.

A better benchmark would be comparing the times taken to run a `computation` (like a benchmark bytecode) between the old and the new version. This way we would get an overall performance feel caused by the changes. Hence, where can I find a bunch of `sample bytecodes` to run the benchmarks against them? Are they somewhere in the `fixtures directory`?

pipermerriam:
> I just implemented `pop` and `pop_n` functions instead of `pop1, pop2, pop3` because there are places in the codebase which requires the `popping of 5 elements`, and I felt that this approach was not so feasible.

I'm still leaning pretty heavily towards just having `pop1`, `pop2`, for every number of elements that we need to pop.  It may not be very DRY but it will be highly performant and shouldn't result in much of any code maintenance since it's unlikely to change.  Each of these APIs would need to be replicated on the `Computation` object as well.

eth/vm/computation.py
```diff
@@ -319,7 +323,41 @@ def stack_pop(self, num_items: int=1, type_hint: str=None) -> Any:
         Raise `eth.exceptions.InsufficientStack` if there are not enough items on
         the stack.
         """
-        return self._stack.pop(num_items, type_hint)
+        if num_items == 1:
```
pipermerriam: I'd like to completely get rid of the type hinting and any conversion between `int <-> bytes` in both the `Computation` and `Stack` APIs, moving **all** conversion down into the opcode functions themselves.

For example:

https://github.com/ethereum/py-evm/blob/dfe43ae8031767837eaf04316c0052f7af02db4c/eth/vm/logic/memory.py#L6-L8

This would become:

```python
def mstore(computation: BaseComputation) -> None:
    start_position = computation.stack_pop(type_hint=constants.UINT256)
    raw_value = computation.stack_pop1()
    value = int_to_big_endian(raw_value)
    ...
```

kclowes:
Closing, stale.

PR #1669, opened 2018-12-22
Implements validate_proposer_signature function

### LINKED ISSUE #1639: Implement `validate_proposer_signature`
### What is wrong?
Per-slot processing - [validate proposer signature](
https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#proposer-signature)

### How can it be fixed
1. Implement the `validate_proposer_signature` logic in `eth/beacon/state_machines/validation.py`.
    * Use flat hash instead of tree hashing for now
2. Add tests for it!

Bhargavasomu:
### What was wrong?
Fixes #1639 

### How was it fixed?
Followed the `specs` and most of the `helper` functions were used. Further added a `TODO` for replacing `flat hash` with `Tree Hash` in the future.

#### Cute Animal Picture

### DIFF
--- a/eth/beacon/types/blocks.py
+++ b/eth/beacon/types/blocks.py
@@ -114,10 +114,8 @@ class BaseBeaconBlock(rlp.Serializable):
     def num_attestations(self) -> int:
         return len(self.body.attestations)
 
-    @classmethod
-    def get_block_without_signature_root(
-            cls,
-            block: 'BaseBeaconBlock') -> Hash32:
-        return block.copy(
+    @property
+    def block_without_signature_root(self) -> Hash32:
+        return self.copy(
             signature=EMPTY_SIGNATURE
         ).root
--- a/eth/beacon/types/blocks.py
+++ b/eth/beacon/types/blocks.py
@@ -118,12 +118,6 @@ class BaseBeaconBlock(rlp.Serializable):
     def get_block_without_signature_root(
             cls,
             block: 'BaseBeaconBlock') -> Hash32:
-        return cls(
-            slot=block.slot,
-            parent_root=block.parent_root,
-            state_root=block.state_root,
-            randao_reveal=block.randao_reveal,
-            candidate_pow_receipt_root=block.candidate_pow_receipt_root,
-            signature=EMPTY_SIGNATURE,
-            body=block.body,
+        return block.copy(
+            signature=EMPTY_SIGNATURE
         ).root
--- a/eth/beacon/types/blocks.py
+++ b/eth/beacon/types/blocks.py
@@ -117,7 +117,7 @@ class BaseBeaconBlock(rlp.Serializable):
     @classmethod
     def get_block_without_signature_root(
             cls,
-            block: 'BaseBeaconBlock') -> 'BaseBeaconBlock':
+            block: 'BaseBeaconBlock') -> Hash32:
         return cls(
             slot=block.slot,
             parent_root=block.parent_root,
--- a/eth/beacon/types/blocks.py
+++ b/eth/beacon/types/blocks.py
@@ -19,6 +19,8 @@ from eth.rlp.sedes import (
     uint64,
     uint384,
 )
+
+from eth.beacon.constants import EMPTY_SIGNATURE
 from eth.beacon._utils.hash import hash_eth2
 
 from .attestations import Attestation
@@ -111,3 +113,17 @@ class BaseBeaconBlock(rlp.Serializable):
     @property
     def num_attestations(self) -> int:
         return len(self.body.attestations)
+
+    @classmethod
+    def create_block_without_signature_root(
+            cls,
+            block: 'BaseBeaconBlock') -> 'BaseBeaconBlock':
+        return cls(
+            slot=block.slot,
+            parent_root=block.parent_root,
+            state_root=block.state_root,
+            randao_reveal=block.randao_reveal,
+            candidate_pow_receipt_root=block.candidate_pow_receipt_root,
+            signature=EMPTY_SIGNATURE,
+            body=block.body,
+        )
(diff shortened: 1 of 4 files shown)


Bhargavasomu:
@hwwhww I have the following exhaustive issues from the implementation I did.
* I have implemented the function assuming that the `use case` for this function is to validate the `block` which comes `unsigned`. In other words, I am assuming that this function is called only on `Yet not signed by proposer blocks`.
> Let block_without_signature_root be the hash_tree_root of block where block.signature is set to EMPTY_SIGNATURE

I came up with this assumption because, from the specs the above point was not clear enough. Would be happy to make the corresponding changes, if cleared upon that point.
 * Also I was thinking that `BEACON_CHAIN_SHARD_NUMBER` and `DOMAIN_PROPOSAL`(Including others) constants should be included in `eth/beacon/constants.py`, because it doesn't feel correct that `BEACON_CHAIN_SHARD_NUMBER` should be binded to only `Serenity` fork (or should it? Please correct me if wrong.)
* I see that `epoch_length` is an extra argument which is not mentioned in the `specs`. What should be the value of `epoch_length` that should be used for the `get_beacon_proposer_index` function?

hwwhww:
> * I have implemented the function assuming that the `use case` for this function is to validate the `block` which comes `unsigned`. In other words, I am assuming that this function is called only on `Yet not signed by proposer blocks`.
>> Let block_without_signature_root be the hash_tree_root of block where block.signature is set to EMPTY_SIGNATURE
> I came up with this assumption because, from the specs the above point was not clear enough. Would be happy to make the corresponding changes, if cleared upon that point.

`validate_proposer_signature` function is called only on "processing a signed block".

I'd create classmethod `BaseBeaconBlock.create_block_without_signature_root` that returns a `BaseBeaconBlock` object that the signature is set to `EMPTY_SIGNATURE`. Then I can use `block.create_block_without_signature_root().root` as `block_without_signature_root`.


> * Also I was thinking that `BEACON_CHAIN_SHARD_NUMBER` and `DOMAIN_PROPOSAL`(Including others) constants should be included in `eth/beacon/constants.py`, because it doesn't feel correct that `BEACON_CHAIN_SHARD_NUMBER` should be binded to only `Serenity` fork (or should it? Please correct me if wrong.)

`BEACON_CHAIN_SHARD_NUMBER` would be passed by parameter, like:
```python
def validate_proposer_signature(state: BeaconState,
                                block: BaseBeaconBlock,
                                beacon_chain_shard_number: int) -> None:
```
I suppose we will clean the constants up after the spec is even more stable. i.e., not all constants should be passed via parameters for testing.

`DOMAIN_PROPOSAL` is `eth.beacon.enums.SignatureDomain.DOMAIN_PROPOSAL`.


> * I see that `epoch_length` is an extra argument which is not mentioned in the `specs`. What should be the value of `epoch_length` that should be used for the `get_beacon_proposer_index` function?

It needs to pass `epoch_length` to `validate_proposer_signature` as well for testing; become:
```python
def validate_proposer_signature(state: BeaconState,
                                block: BaseBeaconBlock,
                                beacon_chain_shard_number: int,
                                epoch_length: int) -> None:
```

Bhargavasomu:
@hwwhww please verify.

tests/beacon/state_machines/test_validation.py
```diff
@@ -0,0 +1,41 @@
...
+            for _ in range(10)
+        ],
+        shard_committees_at_slots=get_sample_shard_committees_at_slots(
+            num_slot=128,
+            num_shard_committee_per_slot=10,
+            sample_shard_committee_params=sample_shard_committee_params,
+        ),
+    )
+
+    validate_proposer_signature(
```
hwwhww: Could you add another case that `validate_proposer_signature` would raise `ValidationError`?
```python
with pytest.raises(ValidationError):
    validate_proposer_signature(...)
```

eth/beacon/types/blocks.py
```diff
@@ -111,3 +113,17 @@ def root(self) -> Hash32:
     @property
     def num_attestations(self) -> int:
         return len(self.body.attestations)
+
+    @classmethod
+    def create_block_without_signature_root(
```
hwwhww: Whoops, sorry, now I think this function could just return the root as its name says.
```python
    @classmethod
    def get_block_without_signature_root(
            cls,
            block: 'BaseBeaconBlock') -> Hash32:
        return cls(
            slot=block.slot,
            parent_root=block.parent_root,
            state_root=block.state_root,
            randao_reveal=block.randao_reveal,
            candidate_pow_receipt_root=block.candidate_pow_receipt_root,
            signature=EMPTY_SIGNATURE,
            body=block.body,
        ).root
```

Bhargavasomu:
@hwwhww could you please take a look now. The `tests` have been `parametrized`. Thankyou.

eth/beacon/types/blocks.py
```diff
@@ -111,3 +113,17 @@ def root(self) -> Hash32:
     @property
     def num_attestations(self) -> int:
         return len(self.body.attestations)
+
+    @classmethod
+    def get_block_without_signature_root(
+            cls,
+            block: 'BaseBeaconBlock') -> Hash32:
+        return cls(
```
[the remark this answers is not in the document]
hwwhww: Ah right, using `copy` looks more concise. 👍 
If we use `copy`, it could be a `property`:
```python
    @property
    def block_without_signature_root(self) -> Hash32:
        return self.copy(
            signature=EMPTY_SIGNATURE,
        ).root
```

it assumes the type of "block_without_signature" is as same as the given `block`

Bhargavasomu:
@hwwhww @djrtwo made the changes. Please review again. Thankyou.

PR #1670, opened 2018-12-22
Lazily evaluate vm.previous_hashes

### LINKED ISSUE #648: Lazily evaluate vm.previous_hashes
### What is wrong?

`vm.previous_hashes` hits the database a lot. Caching helps, but it still involves ~`255 * n` DB reads when importing n blocks. There are probably a lot of blocks that never use the `BLOCKHASH` evm opcode, and so these reads are wasted.

### How can it be fixed

Send a callable (like `lambda: self.previous_hashes`, though probably not literally that) to execution_context that it will invoke, as needed.

I would put a [Good First Issue], but the API is changing a lot, and I don't want to ask a new contributor to do a bunch of ugly rebases.

Bhargavasomu:
### What was wrong?
Fixes #648 

### How was it fixed?
The `execution_context` variable and also the `ExecutionContext` class now stores the variable `prev_hashes` or `_prev_hases` as a `Generator` object. Even the function `get_prev_hashes` of `eth/vm/base.py` returns a `generator` object instead of `tuple`.

The `Generator` object is converted to tuple only if `get_ancestor_hash` function of `eth/vm/state.py` is called, which is in turn called only by the `BLOCKHASH` opcode functionality.

#### Cute Animal Picture

### DIFF
--- a/eth/_utils/generator.py
+++ b/eth/_utils/generator.py
@@ -1,5 +1,5 @@
 import itertools
-from typing import (
+from typing import (  # noqa: F401
     Generic,
     Iterable,
     Iterator,
@@ -12,14 +12,14 @@ TItem = TypeVar('TItem')
 
 
 class CachedIterable(Generic[TItem], Iterable[TItem]):
-    def __init__(self, iterator: Iterable[TItem]) -> None:
-        self._cached_results: List[TItem] = []
-        self._iterator = iterator
+    def __init__(self, iterable: Iterable[TItem]) -> None:
+        self._cached_results = []  # type: List[TItem]
+        self._iterator = iter(iterable)
 
     def __iter__(self) -> Iterator[TItem]:
         return itertools.chain(self._cached_results, self._cache_and_yield())
 
-    def _cache_and_yield(self) -> Iterable[TItem]:
+    def _cache_and_yield(self) -> Iterator[TItem]:
         for item in self._iterator:
             self._cached_results.append(item)
             yield item
--- a/eth/_utils/generator.py
+++ b/eth/_utils/generator.py
@@ -1,0 +1,25 @@
+import itertools
+from typing import (
+    Generic,
+    Iterable,
+    Iterator,
+    List,
+    TypeVar,
+)
+
+
+TItem = TypeVar('TItem')
+
+
+class CachedIterable(Generic[TItem], Iterable[TItem]):
+    def __init__(self, iterator: Iterable[TItem]) -> None:
+        self._cached_results: List[TItem] = []
+        self._iterator = iterator
+
+    def __iter__(self) -> Iterator[TItem]:
+        return itertools.chain(self._cached_results, self._cache_and_yield())
+
+    def _cache_and_yield(self) -> Iterable[TItem]:
+        for item in self._iterator:
+            self._cached_results.append(item)
+            yield item
--- a/eth/rlp/headers.py
+++ b/eth/rlp/headers.py
@@ -1,7 +1,7 @@
 import time
 from typing import (
+    Iterable,
     Optional,
-    Tuple,
     Union,
     overload,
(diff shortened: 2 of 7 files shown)


Bhargavasomu:
@carver should any of the tests be changed so as to suit the changes. This is because most of the cases are failing on `storage` and `balance`. I feel that changing this logic may cause the state variables like `storage` to be different. But not sure why the `balance` after mining is changed.

Bhargavasomu:
Now I get it, instead of storing the whole `tuple of previous 255 block hashes`, we are just storing the `generator of the 255 previous block hashes` in the block header. Hence the overall `block hash` would change, and this `if condition` becomes `True` and leads to a `storm of errors`.
https://github.com/ethereum/py-evm/blob/d82b10ae361cde6abbac62f171fcea7809c4e3cf/tests/json-fixtures/test_blockchain.py#L275-L277

We should modify the `JSON Fixtures` if we want to merge this logic I guess.

carver:
> Is it the case that these changes have to be made in `Fixtures` directory?

The `fixtures` directory is a submodule of `ethereum/tests` which is used by the other clients as well. Let's assume they are not changeable, for now.

Bhargavasomu:
@carver I will dig further into this as to why the tests are failing. But is the approach I have taken right?

carver:
> But is the approach I have taken right?

No, putting an lru cache on a generator will return the same generator which has already reached the end of iteration.

carver:
Also, if you only need the first ancestor, this code is looking up all of them anyway in `get_ancestor_hash`. We can do a more significant rewrite of the `get_ancestor_hash` method so that it iterates over only the ancestors you need.

Bhargavasomu:
@carver I have modified the code. Now all the testcases are passing and now we have a pro and a con.
* PRO: if we need only the `nth ancestor`, this code will now traverse only till a depth of `n`.
* CON: Previously we had this part of the code
https://github.com/ethereum/py-evm/blob/100b8c7d5267a6a6c9ae9c0472963db217b390db/eth/vm/state.py#L204
But now since we are using a generator, this part of the code is invalid and obsolete. The disadvantage of that is, we would end up traversing the whole `previous hashes` and then come to know that we don't have the `hash of the required ancestor`. We could store another variable called `number_prev_hashes` recorded and pass that variable around, but it seems to me that it would make maintenance hard. Please let me know how to proceed.

/cc @pipermerriam

carver:
> * But now since we are using a generator, this part of the code is invalid and obsolete. The disadvantage of that is, we would end up traversing the whole `previous hashes` and then come to know that we don't have the `hash of the required ancestor`. We could store another variable called `number_prev_hashes_recorded` and pass that variable around, but it seems to me that it would make maintenance hard. Please let me know how to proceed.

Hm, at first glance, I don't see what that validation adds above and beyond double-checking that the block number is >=0. I think it would be fine to do something like:

```py
        ancestor_depth = self.block_number - block_number - 1
        is_ancestor_depth_out_of_range = (
            ancestor_depth >= MAX_PREV_HEADER_DEPTH or
            ancestor_depth < 0 or
            block_number < 0
        )
        if is_ancestor_depth_out_of_range:
            return Hash32(b'')
        ancestor_hash = self.execution_context.prev_hashes[ancestor_depth]
```
We could optionally do something like catching key errors and return the empty hash, but since it's entirely unexpected, a crash seems like a reasonable response.

Bhargavasomu:
@carver I have done the changes. Could you please review? Thankyou.

carver:
Just pushed a concept for a utility that caches the generated values, which saves us from some of the generator-copying bookkeeping. Look okay to you?

Bhargavasomu:
@carver this looks **fantastic** !!! This makes the code lot cleaner and so easy to read. Thankyou.

PR #1677, opened 2018-12-26
Put validator balances into a separate list in the state

### LINKED ISSUE #1647: Put validator balances into a separate list in the state
### What is wrong?

https://github.com/ethereum/eth2.0-specs/pull/317

### How can it be fixed

Update related modules.

hwwhww:
### What was wrong?
Fixes #1647

### How was it fixed?
Put `ValidatorRecord.balance: int` to `BeaconState.validator_balances: Sequence[int]`.

#### Cute Animal Picture

iguana_wish_you_a_merry_christmas_card-r47b27aa002bc4616a76125640f7648d1_xvuak_8byvr_540

### DIFF
--- a/tests/beacon/types/test_states.py
+++ b/tests/beacon/types/test_states.py
@@ -65,15 +65,17 @@ def test_defaults(sample_beacon_state_params):
 def test_num_validators(expected,
                         max_deposit,
                         empty_beacon_state):
-    validator_registry = [
-        mock_validator_record(
-            pubkey,
-            max_deposit,
-        )
-        for pubkey in range(expected)
-    ]
     state = empty_beacon_state.copy(
-        validator_registry=validator_registry,
+        validator_registry=[
+            mock_validator_record(
+                pubkey,
+            )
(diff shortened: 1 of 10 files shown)


tests/beacon/types/test_states.py
```diff
@@ -65,15 +65,17 @@ def test_defaults(sample_beacon_state_params):
...
-        validator_registry=validator_registry,
+        validator_registry=[
+            mock_validator_record(
+                pubkey,
+            )
+            for pubkey in range(expected)
+        ],
+        validator_balances=(
+            max_deposit
+            for _ in range(10)
```
ChihChengLiang: Though it might not matter, should this be `expected`? The parameters in this test look strange to me, why we are using `[(0), (1)]` not `[0, 1]` if we are using the variable `expected` as integers.

hwwhww: `[(0), (1)]` means testing twice, not a list/tuple of integers.

PR #1679, opened 2018-12-27
Update StateMachine.__init__() and add `BeaconBlock`

hwwhww:
### What was wrong?
Make `StateMachine` better.

### How was it fixed?
1. Add `BeaconBlock` layer between `BaseBeaconBlock` and
`SerenityBeaconBlock`.
    - IMO the rule is:
        - In `eth.beacon.state_machines.forks.serenity` - use `SerenityBeaconBlock`.
        - In type hinting - use `BaseBeaconBlock`.
        - In other cases - use `BeaconBlock`.
2. Update `StateMachine.__init__()`
3. Add `BeaconChainDB.block_class` and `BeaconChainDB.set_block_class(block_class)`: now it's forkable.
    - TBD: It's for `rlp.decode(block_rlp, sedes=sedes)` with different subclasses of `BaseBeaconBlock`. Now I use set_block_class per `StateMachine` initialization, but I do think it doesn't a beautiful thing to do.

#### Cute Animal Picture

penguin2

### DIFF
--- a/eth/beacon/state_machines/base.py
+++ b/eth/beacon/state_machines/base.py
@@ -7,6 +7,10 @@ from typing import (
     Type,
 )
 
+from eth_typing import (
+    Hash32,
+)
+
 from eth._utils.datatypes import (
     Configurable,
 )
@@ -30,7 +34,7 @@ class BaseBeaconStateMachine(Configurable, ABC):
     config = None  # type: BeaconConfig
 
     block = None  # type: BaseBeaconBlock
-    state = None  # type: BeaconState
+    _state = None  # type: BeaconState
 
     block_class = None  # type: Type[BaseBeaconBlock]
     state_class = None  # type: Type[BeaconState]
@@ -67,10 +71,16 @@ class BaseBeaconStateMachine(Configurable, ABC):
 class BeaconStateMachine(BaseBeaconStateMachine):
     def __init__(self,
                  chaindb: BaseBeaconChainDB,
-                 block: BaseBeaconBlock,
-                 state: BeaconState) -> None:
-        # TODO: get state from DB, now it's just a stub!
-        self.state = state
+                 block_root: Hash32) -> None:
+        self.chaindb = chaindb
+        self.chaindb.set_block_class(self.get_block_class())
+        self.block = self.chaindb.get_block_by_root(block_root)
+
+    @property
+    def state(self) -> BeaconState:
+        if self._state is None:
+            self._state = self.chaindb.get_state_by_root(self.block.state_root)
+        return self._state
 
     @classmethod
     def get_block_class(cls) -> Type[BaseBeaconBlock]:
--- a/eth/beacon/db/chain.py
+++ b/eth/beacon/db/chain.py
@@ -4,6 +4,7 @@ import functools
 from typing import (
     Iterable,
     Tuple,
+    Type,
 )
 from cytoolz import (
     first,
@@ -38,7 +39,10 @@ from eth.validation import (
 )
 
 from eth.beacon.types.states import BeaconState  # noqa: F401
-from eth.beacon.types.blocks import BaseBeaconBlock  # noqa: F401
+from eth.beacon.types.blocks import (  # noqa: F401
+    BaseBeaconBlock,
+    BeaconBlock,
+)
 from eth.beacon.validation import (
     validate_slot,
 )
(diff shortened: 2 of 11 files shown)


hwwhww:
> Question: Does the base logic (whatever ships with the initial fork) just go into `BeaconBlock` and the `SerenityBeaconBlock` is just a shell around it?

I think yes.

Compare to Py-EVM Eth1.0 chain, 1.0 chain uses `BlockHeader`; 2.0 chain uses *forks subclasses* `BeaconBlock` for serialization:
* There's no per-fork `BlockHeader` implementation like `FrontierBlockHeader`.
* It somehow assumes that no RLP fields changes in the future, and just adjust the architecture when the header-fields-change-fork happen.

This PR assumes that the `BeaconBlock` RLP/SSZ fields *might* change. @djrtwo do you think that makes sense, at this stage?

djrtwo:
@hwwhww I think so. Although we are planning for the future, there are a lot of unknowns..

eth/beacon/state_machines/base.py
```diff
@@ -67,10 +71,16 @@ def import_block(self, block: BaseBeaconBlock) -> Tuple[BeaconState, BaseBeaconB
 class BeaconStateMachine(BaseBeaconStateMachine):
     def __init__(self,
                  chaindb: BaseBeaconChainDB,
-                 block: BaseBeaconBlock,
-                 state: BeaconState) -> None:
-        # TODO: get state from DB, now it's just a stub!
-        self.state = state
+                 block_root: Hash32) -> None:
+        self.chaindb = chaindb
+        self.chaindb.set_block_class(self.get_block_class())
```
pipermerriam: First, I'm worried about the statefulness this introduces to the `chaindb` instance which is a shared resource.  This code is very likely to become problematic within the `trinity` context since there will be multiple processes/coroutines interacting with the `chaindb` and thus if this value changes mid-stream we'll get weird bugs.

I'd like to explore how we can get away from this pattern.  

Properties we need from the API:

- the `chaindb` *needs* to know the proper `BlockClass` to use.  
- Ideally we avoid having to specify a new `ChainDB` class for every fork since it's likely that the internal logic won't need to be changed.

And last, we've been wanting to move away from the inheritance based model that we are currently using in the current `eth.vm` namespace in favor of something that is composition based.

> Forgive 

### Idea A: pass `block_class` as a parameter into the `ChainDB.__init__` class.

* Benefits:
  - compositional
* Cons
  - requires the `chaindb` to exist at the fork level rather than the chain level, meaning it will need to be re-instantiated for each fork.  This may result in some larger refactors.

### Idea B: pass `block_class` into any method that needs it.

* Benefits
  - makes the `ChainDB` agnostic about the rich encoding classes.
* Cons
  - requires much more explicit code

### Idea C: create an `Encoder` class which acts as the translator between the `ChainDB` and the `StateMachine`.  Each `StateMachine` would require an `Encoder` class like this:

```python
BlockClass = TypeVar("BlockClass", bound=BaseBlock)

class Encoder(Generic[BlockClass]):
    @classmethod
    def get_block_class(cls) -> BlockClass:
        ...
    def encode_block(block: BlockClass) -> bytes:
        ...
    def decode_block(raw_block: bytes) -> BlockClass:
        ...
```

Then each `StateMachine` would need to be told which `Encoder` class to use which ideally isn't done with inheritance but a class variable seems like the ideal way to do this...

And then the `ChainDB` would need to be given the `Encoder` instance via some API, potentially by passing it into any method that needs it.

### Closing thoughts

I think I'm leaning towards either B or C, with both being different approaches to the same concept.  I think passing in the data to `chaindb.__init__` may prove problematic as it would move the `ChainDB` from being a `Chain` level object down to being a `StateMachine` level object, thus I think it may be appropriate to require each method which needs this data to require it as a parameter.  Note that option C allows a smooth transition from `RLP` to `SSZ` since the `Encoder` class abstracts away any encoding details.

hwwhww:
Thanks a lot for the feedback! ❤️ 

I hate the `set_block_class` pattern too, and I’m happy to get rid of it!

Idea A and B came to my mind:
- For (A), the cons were that I hope we don’t make the architecture *too* different from Ethereum 1.0 chains.
- For (B), I’m not sure how much work it would be in Trinity side as well? Will we need to pass `block_class` through all the APIs in Trinity level? Or mostly in `eth` `BeaconChain` class? For example:
    - https://github.com/ethereum/py-evm/blob/8336a9ebebfed9391b4483f43d8edbb4369dcdf1/eth/beacon/chains/base.py#L194
    - https://github.com/ethereum/py-evm/blob/8336a9ebebfed9391b4483f43d8edbb4369dcdf1/eth/beacon/chains/base.py#L217

I’m leaning to C > B > A currently.

pipermerriam:
I don't think you should concern yourself too much with how `trinity` comes into play since the Eth2.0 stuff in `trinity` will very likely be independent from the `eth1` constructs.

PR #1681, opened 2018-12-28
Add beacon chain related new type

### LINKED ISSUE #1383: Add beacon chain / eth2.0 related NewType

ChihChengLiang:
### What was wrong?

This is an attempt to fix #1383

### How was it fixed?

- Add `NewType`s in `eth/beacon/typing.py`

### Discussion

I added some additional new types:

1. Differentiate signatures in `bytes` and `Tuple[int, int]` form.
```python
BLSSignature = NewType('BLSSignature', bytes)  
BLSSignatureAggregated = NewType('BLSSignatureAggregated', Tuple[int, int]) # Tuple[uint384, uint384]
```
2. Add `Timestamp` and `Second` types.
3. Add a `ValidatorIndex` type.

#### Also, some issues need to resolve

- Are 5th slot and 5 slots considered same `SlotNumber` type?
- Do we need `uint64` and `uint384` types to differentiate different integers? The purpose is to check the intention of the variable rather than to check integer length or signed/unsigned.

#### Suggested types

- CommitteeIndex

### Cute Animal Picture

image

### DIFF
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -33,6 +33,7 @@ from eth.beacon.block_committees_info import (
 )
 from eth.beacon.enums import (
     SignatureDomain,
+    ValidatorRegistryDeltaFlag,
 )
 from eth.beacon.types.shard_committees import (
     ShardCommittee,
@@ -44,13 +45,13 @@ from eth.beacon._utils.random import (
 import functools
 
 from eth.beacon.typing import (
-    ShardNumber,
+    Bitfield,
     BLSPubkey,
+    Ether,
+    Gwei,
+    ShardNumber,
     SlotNumber,
     ValidatorIndex,
-    Bitfield,
-    Gwei,
-    Ether,
 )
 
 if TYPE_CHECKING:
@@ -162,7 +163,7 @@ def get_active_validator_indices(
 #
 @to_tuple
 def _get_shards_committees_for_shard_indices(
-        shard_indices: Sequence[Sequence[ShardNumber]],
+        shard_indices: Sequence[Sequence[ValidatorIndex]],
         start_shard: ShardNumber,
         total_validator_count: int,
         shard_count: int) -> Iterable[ShardCommittee]:
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -385,7 +385,7 @@ def get_effective_balance(
 def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_chain_tip: Hash32,
                                                validator_index: ValidatorIndex,
                                                pubkey: BLSPubkey,
-                                               flag: int) -> Hash32:
+                                               flag: ValidatorRegistryDeltaFlag) -> Hash32:
     """
     Compute the next hash in the validator registry delta hash chain.
     """
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -43,6 +43,10 @@ from eth.beacon._utils.random import (
 )
 import functools
 
+from eth.beacon.typing import (
+    ShardNumber,
+    BLSPubkey,
+)
 
 if TYPE_CHECKING:
     from eth.beacon.types.attestation_data import AttestationData  # noqa: F401
@@ -51,10 +55,6 @@ if TYPE_CHECKING:
     from eth.beacon.types.fork_data import ForkData  # noqa: F401
     from eth.beacon.types.slashable_vote_data import SlashableVoteData  # noqa: F401
     from eth.beacon.types.validator_records import ValidatorRecord  # noqa: F401
-    from eth.beacon.typing import (
-        ShardNumber,
-        BLSPubkey,
-    )
 
 
 def _get_element_from_recent_list(
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -296,7 +296,7 @@ def get_block_committees_info(parent_block: 'BaseBeaconBlock',
 
 def get_beacon_proposer_index(state: 'BeaconState',
                               slot: SlotNumber,
-                              epoch_length: int) -> int:
+                              epoch_length: int) -> ValidatorIndex:
     """
     Return the beacon proposer index for the ``slot``.
     """
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -56,6 +56,7 @@ if TYPE_CHECKING:
         BLSPubkey,
     )
 
+
 def _get_element_from_recent_list(
         target_list: Sequence[Any],
         target_slot: int,
--- a/eth/beacon/helpers.py
+++ b/eth/beacon/helpers.py
@@ -46,6 +46,11 @@ import functools
 from eth.beacon.typing import (
     ShardNumber,
     BLSPubkey,
+    SlotNumber,
+    ValidatorIndex,
+    Bitfield,
+    Gwei,
+    Ether,
 )
 
 if TYPE_CHECKING:
@@ -59,8 +64,8 @@ if TYPE_CHECKING:
 
 def _get_element_from_recent_list(
         target_list: Sequence[Any],
-        target_slot: int,
-        slot_relative_position: int) -> Any:
+        target_slot: SlotNumber,
+        slot_relative_position: SlotNumber) -> Any:
     """
     Return the element from ``target_list`` by the ``target_slot`` number,
     where the the element should be at ``target_slot - slot_relative_position``th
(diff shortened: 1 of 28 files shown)


ChihChengLiang:
After discussion with @hwwhww and @mhchia. Since the nature that all types defined in this PR are only used for the time being, we can just focus on the short-term gain. We conclude that

- `BLSSignature`: Though the distinction between `bytes` and `Tuple[int, int]` only happens in py_ecc. , defining two types for both cases is beneficial in py-evm now. Rename them to `BLSSignatureBytes` and `BLSSignatureIntegers`.
- `uint64` etc. are not urgent recently, and a better decision could be made when https://github.com/ethereum/eth2.0-specs/issues/308 resolves.
- `CommitteeIndex` is good to be added.

For the last question, `Are 5th slot and 5 slots considered same SlotNumber type?`. Should we use a type `ShardCount` for variable `SHARD_COUNT`, or should it share the same type with `ShardNumber`. We would like to request the wisdom from @cburgdorf.

eth/beacon/types/exits.py
```diff
@@ -24,9 +28,9 @@ class Exit(rlp.Serializable):
     ]
 
     def __init__(self,
-                 slot: int,
+                 slot: SlotNumber,
                  validator_index: int,
```
hwwhww:
```suggestion
                 validator_index: ValidatorIndex,
```

eth/beacon/types/shard_committees.py
```diff
@@ -27,7 +30,7 @@ class ShardCommittee(rlp.Serializable):
     ]
 
     def __init__(self,
-                 shard: int,
+                 shard: ShardNumber,
                  committee: Sequence[int],
```
hwwhww:
```suggestion
                 committee: Sequence[ValidatorIndex],
```

eth/beacon/types/shard_reassignment_records.py
```diff
@@ -21,8 +25,8 @@ class ShardReassignmentRecord(rlp.Serializable):
 
     def __init__(self,
                  validator_index: int,
```
hwwhww:
```suggestion
                 validator_index: ValidatorIndex,
```

eth/beacon/types/states.py
```diff
@@ -76,27 +83,27 @@ class BeaconState(rlp.Serializable):
...
+            justification_bitfield: Bitfield,
+            finalized_slot: SlotNumber,
             processed_pow_receipt_root: Hash32,
             validator_registry: Sequence[ValidatorRecord]=(),
-            validator_balances: Sequence[int]=(),
+            validator_balances: Sequence[Gwei]=(),
             latest_randao_mixes: Sequence[Hash32]=(),
             latest_vdf_outputs: Sequence[Hash32]=(),
             shard_committees_at_slots: Sequence[Sequence[ShardCommittee]]=(),
             persistent_committees: Sequence[Sequence[int]]=(),
```
hwwhww:
```suggestion
            persistent_committees: Sequence[Sequence[ValidatorIndex]]=(),
```

eth/beacon/types/validator_registry_delta_block.py
```diff
@@ -28,7 +32,7 @@ class ValidatorRegistryDeltaBlock(rlp.Serializable):
     def __init__(self,
                  latest_registry_delta_root: Hash32,
                  validator_index: int,
```
hwwhww:
```suggestion
                 validator_index: ValidatorIndex,
```

eth/beacon/types/validator_registry_delta_block.py
```diff
@@ -28,7 +32,7 @@ class ValidatorRegistryDeltaBlock(rlp.Serializable):
...
-                 pubkey: int,
+                 pubkey: BLSPubkey,
                  flag: int) -> None:
```
hwwhww:
```suggestion
                 flag: ValidatorRegistryDeltaFlag) -> None:
```

eth/beacon/helpers.py
```diff
@@ -370,8 +383,8 @@ def get_effective_balance(validator_balances: Sequence[int], index: int, max_dep
 
 
 def get_new_validator_registry_delta_chain_tip(current_validator_registry_delta_chain_tip: Hash32,
-                                               validator_index: int,
-                                               pubkey: int,
+                                               validator_index: ValidatorIndex,
+                                               pubkey: BLSPubkey,
                                                flag: int) -> Hash32:
```
hwwhww:
```suggestion
                                               flag: ValidatorRegistryDeltaFlag) -> Hash32:
```

hwwhww: FYI: I'm removing this function in other PR.

ChihChengLiang:
Thank you @hwwhww  for the review ❤️

eth/beacon/types/states.py
```diff
@@ -76,27 +83,27 @@ class BeaconState(rlp.Serializable):
...
+            validator_registry_latest_change_slot: SlotNumber,
             validator_registry_exit_count: int,
             validator_registry_delta_chain_tip: Hash32,
-            previous_justified_slot: int,
-            justified_slot: int,
-            justification_bitfield: int,
-            finalized_slot: int,
+            previous_justified_slot: SlotNumber,
+            justified_slot: SlotNumber,
+            justification_bitfield: Bitfield,
```
hwwhww: Maybe leave a comment: "TODO: check if `justification_bitfield` is `bytes` or `int`".

eth/beacon/types/validator_records.py
```diff
@@ -41,12 +45,12 @@ class ValidatorRecord(rlp.Serializable):
...
 
     def __init__(self,
-                 pubkey: int,
+                 pubkey: BLSPubkey,
                  withdrawal_credentials: Hash32,
                  randao_commitment: Hash32,
-                 randao_layers: int,
-                 status: int,
-                 latest_status_change_slot: int,
+                 randao_layers: SlotNumber,
```
hwwhww: Ah, `randao_layers` is the "5 slots" thing.
```suggestion
                 randao_layers: int,
```

eth/beacon/types/validator_records.py
```diff
@@ -67,18 +71,18 @@ def is_active(self) -> bool:
...
+                              latest_status_change_slot: SlotNumber) -> 'ValidatorRecord':
         """
         Return a new pending ``ValidatorRecord`` with the given fields.
         """
         return cls(
             pubkey=pubkey,
             withdrawal_credentials=withdrawal_credentials,
             randao_commitment=randao_commitment,
-            randao_layers=0,
+            randao_layers=SlotNumber(0),
```
hwwhww:
```suggestion
            randao_layers=0,
```

eth/beacon/types/proposer_slashings.py
```diff
@@ -27,13 +30,14 @@ class ProposerSlashing(rlp.Serializable):
     def __init__(self,
                  proposer_index: int,
```
hwwhww:
```suggestion
                 proposer_index: ValidatorIndex,
```

eth/beacon/helpers.py
```diff
@@ -284,7 +295,7 @@ def get_block_committees_info(parent_block: 'BaseBeaconBlock',
 
 
 def get_beacon_proposer_index(state: 'BeaconState',
-                              slot: int,
+                              slot: SlotNumber,
                               epoch_length: int) -> int:
```
hwwhww:
```suggestion
                              epoch_length: int) -> ValidatorIndex:
```

ChihChengLiang:
Merge first and address the issues later

cburgdorf:
>Should we use a type `ShardCount` for variable `SHARD_COUNT`, or should it share the same type with `ShardNumber`.

I would not make `SHARD_COUNT` be of type `ShardNumber` as the intend is different imho. However, I would consider just using `int` for `SHARD_COUNT`.

But that's my gut feeling response without having too much insight how it ends up in code. Like, do we have a lot of APIs accepting or returning shard counts? If yes, a `ShardCount` type might make sense, if not, just using `int` strikes me as just fine.

PR #1684, opened 2018-12-28
Reorg `validate_proposer_signature`

hwwhww:
### What was wrong?

Arrange the `StateMachine`.

### How was it fixed?
1. Move `eth.beacon.state_machines.validation.validate_proposer_signature` to ` eth.beacon.state_machines.forks.serenity.validation.validate_serenity_proposer_signature`.
2. Call `validate_serenity_proposer_signature` in `per_block_transition`.
3. Add something so that `test_demo.py` can pass.

#### Cute Animal Picture
penguin_in_antarctica_jumping_out_of_the_water


eth/beacon/state_machines/forks/serenity/validation.py
```diff
@@ -21,12 +21,46 @@
...
+
+    # Get the public key of proposer
+    beacon_proposer_index = get_beacon_proposer_index(state, state.slot, epoch_length)
+    proposer_pubkey = state.validator_registry[beacon_proposer_index].pubkey
+
+    is_valid_signature = bls.verify(
+        pubkey=proposer_pubkey,
+        message=proposal_root,
+        signature=block.signature,
+        domain=get_domain(state.fork_data, state.slot, SignatureDomain.DOMAIN_PROPOSAL)
```
djrtwo: add comma
```suggestion
         domain=get_domain(state.fork_data, state.slot, SignatureDomain.DOMAIN_PROPOSAL),
```

eth/beacon/state_machines/forks/serenity/validation.py
```diff
@@ -21,12 +21,46 @@
...
+                                         block: BaseBeaconBlock,
+                                         beacon_chain_shard_number: int,
+                                         epoch_length: int) -> None:
+    block_without_signature_root = block.block_without_signature_root
+
+    # TODO: Replace this root with tree hash root
+    proposal_root = ProposalSignedData(
+        state.slot,
+        beacon_chain_shard_number,
+        block_without_signature_root
```
djrtwo: add comma
```suggestion
         block_without_signature_root,
```

tests/beacon/state_machines/test_demo.py
```diff
@@ -1,17 +1,74 @@
...
+              config,
+              privkeys,
+              pubkeys):
     chaindb = BeaconChainDB(base_db)
-    state = BeaconState(**sample_beacon_state_params)
-    block = BaseBeaconBlock(**sample_beacon_block_params).copy(
-        slot=state.slot + 2,
+    state = genesis_state
+    block = SerenityBeaconBlock(**sample_beacon_block_params).copy(
+        slot=state.slot + 1,
```
djrtwo: I had `+ 2` here to quickly demo that the state transition would process up to the block slot.
I think it is worth leaving that sanity check in here for the time being

tests/beacon/state_machines/test_demo.py
```diff
@@ -1,17 +1,74 @@
...
+    )
+
+    # Sign block
+    beacon_proposer_index = get_beacon_proposer_index(
+        state.copy(
```
djrtwo: You can just use `state` here. The current `state.slot` will give the look ahead on the proposer slot as long as it's not past an epoch transition.

hwwhww: Ahhh! We need to fix bug https://github.com/ethereum/eth2.0-specs/issues/352 first so that we can use `state`. 😭

hwwhww: Leave a TODO comment for now:
```
# TODO: use `state` when the bug of `get_shard_committees_at_slot` is fixed.
```

eth/beacon/state_machines/forks/serenity/state_transitions.py
```diff
@@ -40,7 +43,15 @@ def per_slot_transition(self,
 
     def per_block_transition(self, state: BeaconState, block: BaseBeaconBlock) -> BeaconState:
         # TODO
```
pipermerriam: Is this `TODO` a leftover?

eth/beacon/state_machines/forks/serenity/validation.py
```diff
@@ -21,12 +21,49 @@
...
+    # TODO: Replace this root with tree hash root
+    proposal_root = ProposalSignedData(
+        state.slot,
+        beacon_chain_shard_number,
+        block_without_signature_root,
+    ).root
+
+    # Get the public key of proposer
+    beacon_proposer_index = get_beacon_proposer_index(state, state.slot, epoch_length)
+    proposer_pubkey = state.validator_registry[beacon_proposer_index].pubkey
```
pipermerriam: I'm always a little nervous with indexing into iterables naively like this.  I understand/recognize that the previous `get_beacon_proposer_index` provides some level of guarantee that the `beacon_proposer_index` is indeed present and accurate in the `validator_registry` iterable.

What if this was abstracted away on the `BeaconState` class as something like `state.get_beacon_proposer(epoch_length)`:

```python
proposer = state.get_beacon_proposer(epoch_length)
proposer_pubkey = proposer.pubkey
```

eth/beacon/state_machines/forks/serenity/validation.py
```diff
@@ -21,12 +21,49 @@
...
+
+    is_valid_signature = bls.verify(
+        pubkey=proposer_pubkey,
+        message=proposal_root,
+        signature=block.signature,
+        domain=get_domain(state.fork_data, state.slot, SignatureDomain.DOMAIN_PROPOSAL),
+    )
+
+    if not is_valid_signature:
+        raise ValidationError("Invalid Proposer Signature on block")
```
pipermerriam: Is there any more detailed information we could include in this message?

hwwhww: We can output the arguments of `bls.verify`. Or do you mean throwing the exceptions from `bls.verify` function?

eth/beacon/state_machines/forks/serenity/state_transitions.py
pipermerriam: Mostly just that anyone reading the code will have no context for what is left `TODO`..

eth/beacon/state_machines/forks/serenity/validation.py
djrtwo: We can put the expected proposer that should have signed/proposed at this slot based on the shuffling.
We _cannot_ figure out who signed it if it's invalid.

pipermerriam: @hwwhww I like what @djrtwo suggested, If we add the proposer that we expected the signature to come from as well as the signature details then I think someone encountering this exception should have some basic context information to start digging into "why".

Issue #1315, opened 2018-09-22
Add more tests for rewards


cburgdorf:
### What is wrong?

We recently added unit tests (additional to the existing state tests) that cover the rewards that are paid to the miners.

We should expand the following cases ([citing from](https://github.com/ethereum/py-evm/pull/1303#pullrequestreview-157854461))

>- calculating uncle/nephew reward when the uncle is on a different VM (like: uncle in frontier at 0, nephew in constantinople at 1)
>- uncles that are 2 -> 8 layers back
>- validating that uncles > 8 back are invalid

>(I'm saying the last two without having reviewed the uncles section of the YP recently, so probably have some detail wrong)

### How can it be fixed

Add test [here](https://github.com/ethereum/py-evm/blob/0cc5d59c92c4c8947d67f9462c00c17d39089cc0/tests/core/vm/test_rewards.py)

pipermerriam:
Tests should be *reasonably* easy to write using the new chain builder tools.

Bhargavasomu:
I'm taking this up.

Bhargavasomu:
This can be closed with #1685

PR #1685, opened 2018-12-29
Add some more tests for Rewards

Bhargavasomu:
### What was wrong?
Fixes Issue #1315 

#### Cute Animal Picture

### DIFF
--- a/tests/core/vm/test_rewards.py
+++ b/tests/core/vm/test_rewards.py
@@ -1,7 +1,8 @@
 import pytest
 
 from eth_utils import (
-    to_wei
+    to_wei,
+    ValidationError,
 )
 
 from eth.chains.base import (
@@ -12,6 +13,7 @@ from eth.tools.builder.chain import (
     build,
     disable_pow_check,
     mine_block,
+    mine_blocks,
     byzantium_at,
     frontier_at,
     homestead_at,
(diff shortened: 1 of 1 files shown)


Bhargavasomu:
@cburgdorf please review. Thankyou.

pipermerriam:
I assume that all of the numbers here were simply derived from what the existing code produces (which is fine, just checking).  This strikes me as the type of thing that we could contribute upstream.  cc @lithp as a use case for upstream contributions to the `ethereum/tests` repo.

Bhargavasomu:
@pipermerriam yes these numbers were generated by the current existing code. But I did verify a **few** based on the `block rewards` as to see whether the numbers are matching are not.

pipermerriam:
Cool, I'd like to wait for @lithp to review this and give his thoughts on whether these tests would be better suited for a fixture based approach.  If that was the case this will be an ideal time for us to streamline the manner in which such fixtures are generated and I expect the existing code you've written will get re-used in some form.

lithp:
Sorry for leaving you hanging for so long @Bhargavasomu! These tests are useful to have, thanks for writing them.

What did you mean by `But I did verify a few based on the block rewards as to see whether the numbers are matching are not`? How did you verify the rewards?

The tests repo isn't easy to read but I think there's a good chance the fixtures already cover these new tests.
- I'm 80% that `test_rewards_uncle_created_at_different_generations` is already covered by tests such as [oneUncleGeneration2](https://github.com/ethereum/tests/blob/develop/BlockchainTests/bcUncleTest/oneUncleGeneration2.json). 
- It's very likely that `test_uncle_block_inclusion_validity` is duplicated by [oneUncleGeneration7](https://github.com/ethereum/tests/blob/develop/BlockchainTests/bcUncleTest/oneUncleGeneration7.json)
- I'm much less sure on this one, but `test_rewards_nephew_uncle_different_vm` is plausibly duplicated by [the transition tests](https://github.com/ethereum/tests/blob/develop/src/BlockchainTestsFiller/TransitionTests/bcByzantiumToConstantinople/ConstantinopleTransitionFiller.json)?

I don't think this is a reason to reject the PR though, it's nice to have some quick sanity checks that nothing has been broken before spending an hour running the fixtures. It's probably worth checking more closely to see that these tests really aren't novel but it's unlikely enough that I don't think the more thorough check should block these changes.

pipermerriam:
@lithp in generally I'm :+1: on us having tests which are duplicated in the `ethereum/tests` under the condition that we are reasonably confident they check most of the following.

- easier to debug failures
- easy to maintain and read the tests

Bhargavasomu:
@lithp by verifying, I mean that I did some paper work taking the formula being used and the numbers, for some of the test cases I wrote.

But in overall, is there something more that has to be done as part of this PR?
/cc @pipermerriam

cburgdorf:
@Bhargavasomu I'll admit that I did not check every case to the bone but it seems you've done some basic   checks :)

I'm pretty confident that this improves on the status quo which is why I went ahead and merged it. Thank you very much for another solid contribution! :pray:

PR #1686, opened 2018-12-30
[WIP] Implement validate_randao function 

prateekreddy:
### What was wrong?
Implement `validate_randao` as per Eth 2.0 spec https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#randao

#### Cute Animal Picture

### DIFF
--- a/eth/_utils/numeric.py
+++ b/eth/_utils/numeric.py
@@ -134,4 +134,4 @@ def bitwise_xor(a: Hash32, b: Hash32) -> Hash32:
     c = b''
     for i in range(32):
         c += bytes([a[i] ^ b[i]])
-    return c
+    return Hash32(c)
--- a/eth/_utils/numeric.py
+++ b/eth/_utils/numeric.py
@@ -131,6 +131,18 @@ def bitwise_xor(a: Hash32, b: Hash32) -> Hash32:
     """
     Return the xor of hash ``a`` and hash ``b``
     """
+    if len(a) != 32:
+        raise ValueError(
+            "Arguments must be of size 32: Got: {0}".format(
+                len(a)
+            )
+        )
+    if len(b) != 32:
+        raise ValueError(
+            "Arguments must be of size 32: Got: {0}".format(
+                len(b)
+            )
+        )
     c = b''
     for i in range(32):
         c += bytes([a[i] ^ b[i]])
--- a/eth/_utils/numeric.py
+++ b/eth/_utils/numeric.py
@@ -125,3 +125,13 @@ def integer_squareroot(value: int) -> int:
         x = y
         y = (x + value // x) // 2
     return x
+
+
+def bitwise_xor(a: Hash32, b: Hash32) -> Hash32:
+    """
+    Return the xor of hash ``a`` and hash ``b``
+    """
+    c = b''
+    for i in range(32):
+        c += bytes([a[i] ^ b[i]])
+    return c
--- a/eth/beacon/_utils/hash.py
+++ b/eth/beacon/_utils/hash.py
@@ -9,3 +9,13 @@ def hash_eth2(data: bytes) -> Hash32:
     a future Ethereum 2.0 deployment phase.
     """
     return Hash32(keccak(data))
+
+
+def repeat_hash_eth2(data: bytes, repeats: int) -> Hash32:
+    """
+    Return Hash onion of ``data`` with ``repeats`` layers
+    """
+    if repeats == 0:
+        return Hash32(data)
+    else:
+        return repeat_hash_eth2(bytes(hash_eth2(data)), repeats - 1)
--- a/eth/beacon/state_machines/forks/serenity/operations.py
+++ b/eth/beacon/state_machines/forks/serenity/operations.py
@@ -30,7 +30,7 @@ def validate_randao(state: BeaconState,
     proposer = validator_registry[proposer_index]
 
     validate_serenity_randao_reveal(block, proposer)
-    
+
     latest_randao_mix = latest_randao_mixes[state.slot % config.LATEST_RANDAO_MIXES_LENGTH]
     latest_randao_mix = bitwise_xor(latest_randao_mix, block.randao_reveal)
 
--- a/eth/beacon/state_machines/forks/serenity/operations.py
+++ b/eth/beacon/state_machines/forks/serenity/operations.py
@@ -22,14 +22,28 @@ def validate_randao(state: BeaconState,
     ``randao_commitment`` in the ``state``
 
     """
-    proposer = state.validator_registry[get_beacon_proposer_index(state,
-                                                                  state.slot,
-                                                                  config.EPOCH_LENGTH)]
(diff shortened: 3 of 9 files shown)


eth/beacon/state_machines/forks/serenity/operations.py
```diff
@@ -2,12 +2,51 @@
...
+from eth.beacon.helpers import get_beacon_proposer_index
+from eth._utils.numeric import bitwise_xor
 
 from .validation import (
     validate_serenity_attestation,
+    validate_serenity_randao_reveal,
 )
 
 
+def validate_randao(state: BeaconState,
```
hwwhww: 1. Move this function to new file `eth/beacon/state_machines/forks/serenity/block_transtions.py`.
2. Rename it to `process_randao`.

eth/_utils/numeric.py
```diff
@@ -125,3 +125,25 @@ def integer_squareroot(value: int) -> int:
...
+    if len(b) != 32:
+        raise ValueError(
+            "Arguments must be of size 32: Got: {0}".format(
+                len(b)
+            )
+        )
+    c = b''
+    for i in range(32):
+        c += bytes([a[i] ^ b[i]])
+    return Hash32(c)
```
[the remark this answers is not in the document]
pipermerriam: I think this should be updated to:

```python
result = bytes(bit_a, bit_b for bit_a, bit_b in zip(a, b)
return Hash32(result)
```

Also can you update the `a, b` variables to be `left, right` to avoid single character names.

eth/_utils/numeric.py
```diff
@@ -125,3 +125,25 @@ def integer_squareroot(value: int) -> int:
         x = y
         y = (x + value // x) // 2
     return x
+
+
+def bitwise_xor(a: Hash32, b: Hash32) -> Hash32:
+    """
+    Return the xor of hash ``a`` and hash ``b``
+    """
+    if len(a) != 32:
```
pipermerriam: Depending on the performance needs for this function we could drop these two validation checks since the `Hash32` type checks give us *some* assurances..  I'll leave it to @hwwhww to decide if she thinks this is necessary/beneficial.

eth/beacon/_utils/hash.py
```diff
@@ -9,3 +9,13 @@ def hash_eth2(data: bytes) -> Hash32:
...
+
+
+def repeat_hash_eth2(data: bytes, repeats: int) -> Hash32:
+    """
+    Return Hash onion of ``data`` with ``repeats`` layers
+    """
+    if repeats == 0:
+        return Hash32(data)
+    else:
+        return repeat_hash_eth2(bytes(hash_eth2(data)), repeats - 1)
```
pipermerriam: Can you update this to the following implementation to get rid of the recursion.

```python
from eth_utils.toolz import pipe
import itertools

def repeat_hash_eth2(data, iterations):
    if iterations > 0:
        return pipe(data, *itertools.repeat(hash_eth2, iterations))
    elif iterations == 0:
        return data
    else:
        raise ValidationError("Iterations must be positive: Got {0}".format(iterations))
```

Also, is `iterations=0` a valid use case we expect to encounter?  If not, that code path should be removed.

eth/beacon/state_machines/forks/serenity/operations.py
```diff
@@ -2,12 +2,51 @@
...
+    validate_serenity_randao_reveal(block, proposer)
+
+    latest_randao_mix = latest_randao_mixes[state.slot % config.LATEST_RANDAO_MIXES_LENGTH]
+    latest_randao_mix = bitwise_xor(latest_randao_mix, block.randao_reveal)
+
+    proposer = proposer.copy(
+        randao_commitment=block.randao_reveal,
+        randao_layers=0,
+    )
+    latest_randao_mixes[state.slot % config.LATEST_RANDAO_MIXES_LENGTH] = latest_randao_mix
```
pipermerriam: What do you think about removing this mutation in favor of something like:

idx = state.slot % config.LATEST_RANDAO_MIXES_LENGTH
latest_randao_mixes = tuple(state.latest_randao_mixes[:idx] + (latest_randao_mix,) + tuple(state.latest_randao_mixes[idx + 1:]
```

Removes the mutation and keeps everything as a tuple.

File: eth/beacon/state_machines/forks/serenity/operations.py [hunk 6]
```diff
@@ -2,12 +2,51 @@
...
+    Implements 'per-block-processing.RANDAO' portion of Phase 0 spec:
+    https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#randao
+
+    Verify the ``randao_reveal`` contained within the ``block`` in the context of
+    ``randao_commitment`` in the ``state``
+
+    """
+    proposer_index = get_beacon_proposer_index(state, state.slot, config.EPOCH_LENGTH)
+
+    validator_registry = list(state.validator_registry)
```
pipermerriam: It looks like ths `list` cast isn't needed unless I missed somewhere that `validator_registry` gets mutated.

eth/beacon/state_machines/forks/serenity/operations.py
```diff
@@ -2,12 +2,51 @@
...
+    latest_randao_mixes = list(state.latest_randao_mixes)
+
+    proposer = validator_registry[proposer_index]
+
+    validate_serenity_randao_reveal(block, proposer)
+
+    latest_randao_mix = latest_randao_mixes[state.slot % config.LATEST_RANDAO_MIXES_LENGTH]
```
pipermerriam: We try to avoid overwriting variables.  Can you rename this first varaible to something like `original_randao_mix`

tests/beacon/_utils/test_hash.py
```diff
@@ -9,3 +14,16 @@ def test_hash():
 
 def test_hash_is_keccak256():
     assert hash_eth2(b'foo') == keccak(b'foo')
+
+
+def test_repeat_hash():
+    output = repeat_hash_eth2(b'helloworld', 5)
+    assert len(output) == 32
```
pipermerriam: This type of test is well suited for property based testing.  You could write one using hypothesis like this:

```python
from hypothesis import strategies as st, given

@given(value=st.binary(), iterations=st.integers(min_value=1))
def test_repeat_hash_always_produces_32_bytes(value, iterations)
    output = repeat_hash_eth2(value, iterations)
    assert len(output) == 32
```

eth/_utils/numeric.py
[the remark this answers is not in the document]
hwwhww: @pipermerriam oh yes that's better! 👍 

nitpick:
```python
result = bytes(bit_a ^ bit_b for bit_a, bit_b in zip(a, b))
return Hash32(result)
```

eth/_utils/numeric.py
hwwhww: It's "thoughtful" to have the checks for "utility function", but we definitely don't need it for RANDAO use case. Remove +1.

jannikluhn:
Closing in favor of https://github.com/ethereum/trinity/pull/214

Issue #597, opened 2018-04-25
Expand Chain object tests to use all mainnet VM rules


pipermerriam:
### What is wrong?

The tests in `tests/core/chain-object` are all using only the `SpuriousDragonVM`

https://github.com/ethereum/py-evm/blob/48e94db187ed0b6166a6572f499ef624d5f987ce/tests/core/fixtures.py#L13

We really should be running these tests across all of the mainnet VMs.

### How can it be fixed

Change the tests so that they run against each set of VM rules.  This is probably best done by parametrizing the chain fixtures that are used.

Bhargavasomu:
I'm on this.

PR #1687, opened 2018-12-30
Chain object tests to use all mainnet VM rules

Bhargavasomu:
### What was wrong?
Fixes Issue #597 

### How was it fixed?
A separate fixture was added to create a `Chain` specific to the `VM` being used. This is used to pass the `VM's` as parameters and generate the `chains` from them. This is needed because some properties change from one `VM` to another `VM`. One such example is the signal `doRevert()` raises a `InvalidInstruction` for `FrontierVM`; but the same signal raises `Revert` for `ByzantiumVM`.

Also the `chain_without_validation` fixture now requests all the `VM's` to create all the chains, which are subsequently used in the tests over the whole tests.

#### Cute Animal Picture


Bhargavasomu:
@pipermerriam @cburgdorf review please.

Bhargavasomu:
@pipermerriam made the suggested change. Please take a look once `CircleCI` goes green. Thankyou.

PR #1690, opened 2018-12-31
[WIP] Fix #1683, Implement process_justification

### LINKED ISSUE #1683: Implement `process_justification`
### What is wrong?

- Implement per-epoch processing: justification logic as [spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#justification)
- See if we can set `justification_bitfield` as `bytes`.

### How can it be fixed
- Implement `total_balance`
    - In the spec: Let `total_balance(shard_committee) = sum([get_effective_balance(state, i) for i in shard_committee.committee])`...looks buggy.
    - It can be put in `eth/beacon/helpers.py` for now.
- Implement `process_justification(state: BeaconState, config: BeaconConfig) -> BeaconState` in `eth/beacon/state_machines/forks/serenity/epoch_transitions.py`
- Call `process_justification` in `eth.beacon.state_machines.forks.serenity.per_epoch_transition`.

ChihChengLiang:
#### Cute Animal Picture


eth/beacon/state_machines/forks/serenity/epoch_transitions.py
```diff
@@ -0,0 +1,224 @@
...
+            attestation.data.shard,
+            attestation.participation_bitfield,
+            EPOCH_LENGTH,
+        ))
+        for attestation in previous_epoch_boundary_attestations
+    )
+    previous_epoch_boundary_attester_indices = (
+        tuple()
+        if len(sets_of_previous_epoch_boundary_participants) == 0
+        else frozenset.union(*sets_of_previous_epoch_boundary_participants)
```
pipermerriam: *if* it is possible for `set_of_previous_epoch_boundary_participants` to be empty then this line will fail.

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
```diff
@@ -0,0 +1,224 @@
...
+            state,
+            attestation.data.slot,
+            attestation.data.shard,
+            attestation.participation_bitfield,
+            EPOCH_LENGTH,
+        ))
+        for attestation in previous_epoch_boundary_attestations
+    )
+    previous_epoch_boundary_attester_indices = (
+        tuple()
```
pipermerriam: Is there a reason this is a `tuple()` instead of an empty `frozenset()`

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
```diff
@@ -0,0 +1,224 @@
...
+    )
+    previous_epoch_boundary_attesting_balance = Gwei(sum(
+        get_effective_balance(state, index, MAX_DEPOSIT)
+        for index in previous_epoch_boundary_attester_indices
+    ))
+
+    current_epoch_boundary_root = get_block_root(
+        state,
+        state.slot - EPOCH_LENGTH,
+        LATEST_BLOCK_ROOTS_LENGTH)
```
pipermerriam: *style nitpick*

We normally drop the closing `)` to the next line and leave a dangling `,` in cases like this.

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
```diff
@@ -0,0 +1,224 @@
...
+            attestation.data.slot,
+            attestation.data.shard,
+            attestation.participation_bitfield,
+            EPOCH_LENGTH,
+        ))
+        for attestation in current_epoch_boundary_attestations
+    )
+
+    current_epoch_boundary_attester_indices = (
+        tuple()
```
pipermerriam: Same here with the type mismatch.  Maybe this should be an empty `frozenset()` to keep things homogeneous.

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
```diff
@@ -0,0 +1,224 @@
...
+    justification_bitfield = int.from_bytes(state.justification_bitfield, 'big')
+    finalized_slot = state.finalized_slot
+
+    # Add two bits at the right of justification_bitfield. Cap it at 64 bits.
+    justification_bitfield = (justification_bitfield * 2) % TWO_POWER_64
+
+    if 3 * previous_epoch_boundary_attesting_balance >= 2 * total_balance:
+        justification_bitfield |= 2
+        justified_slot = state.slot - 2 * EPOCH_LENGTH
+    if 3 * current_epoch_boundary_attesting_balance >= 2 * total_balance:
```
pipermerriam: Assuming this isn't intended to be an `elif` then I think it *might* be more appropriate to reverse these two `if` blocks so that we can make them connected using an `elif`.  The goal would be to re-structure this code so that `justification_bitfield` and `justification_slot` (as well as any other variables) are not being overwritten in any locations.

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
ChihChengLiang: That's why I awkwardly protect it by checking `if len(sets_of_previous_epoch_boundary_participants) == 0`
This works on my Python:
```python
>>> sets_of_previous_epoch_boundary_participants = tuple()
>>> frozenset() if len(sets_of_previous_epoch_boundary_participants) == 0 else frozenset.union(*sets_of_previous_epoch_boundary_participants)
frozenset()
```
Is there a better way to do that union?

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
ChihChengLiang: Thanks, will fix

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
ChihChengLiang: Thanks, will fix this.

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
ChihChengLiang: Will fix

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
ChihChengLiang: Would the desired code look like the follows?
```python
one_epoch_ago_justifiable = 3 * current_epoch_boundary_attesting_balance >= 2 * total_balance
two_epochs_ago_justifiable = 3 * previous_epoch_boundary_attesting_balance >= 2 * total_balance

if one_epoch_ago_justifiable and two_epochs_ago_justifiable:
    justification_bitfield |= 3
    justified_slot = state.slot - EPOCH_LENGTH
elif one_epoch_ago_justifiable and not two_epochs_ago_justifiable :
    justification_bitfield |= 1
    justified_slot = state.slot - EPOCH_LENGTH
elif not one_epoch_ago_justifiable and two_epochs_ago_justifiable:
    justification_bitfield |= 2
    justified_slot = state.slot - 2* EPOCH_LENGTH
```
or maybe
```python
if one_epoch_ago_justifiable:
    justified_slot = state.slot - EPOCH_LENGTH
elif two_epochs_ago_justifiable:
    justified_slot = state.slot - 2* EPOCH_LENGTH

if one_epoch_ago_justifiable and two_epochs_ago_justifiable:
    justification_bitfield |= 3
elif one_epoch_ago_justifiable and not two_epochs_ago_justifiable :
    justification_bitfield |= 1
elif not one_epoch_ago_justifiable and two_epochs_ago_justifiable:
    justification_bitfield |= 2

```

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
pipermerriam: :man_facepalming: sorry, previous comment was misguided but now I see why and thus the awkwardness.

I'd change this to be a `try/except`.  Removes the *checking* overhead and is more *pythonic* (depending on who you ask).

```python
try:
    previous_epoch_boundary_attester_indices = frozenset.union(*sets_of_previous_epoch_boundaries)
except:
    previous_epoch_boundary_attester_indices = frozenset()
```

Alternatively you could do this:

```python
previous_epoch_boundary_attester_indices = frozenset.union(frozenset(), *sets_of_previous_epoch_boundaries)
# or more simply
previous_epoch_boundary_attester_indices = frozenset().union(*sets_of_previous_epoch_boundaries)
```

Both of these seems to work correctly with an empty `sets_of_previous_epoch_boundaries`.

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
pipermerriam: Yes, either of these would be :+1:

eth/beacon/state_machines/forks/serenity/epoch_transitions.py
ChihChengLiang: A note for what I learned from the implementation.
I like this one for its expresiveness: 

```python
previous_epoch_boundary_attester_indices = frozenset().union(*sets_of_previous_epoch_boundaries)
```

but it's trickier than I've imagined.
Since the expression starts with an empty frozenset, mypy asked the type of the variable to be explicitly annotated. So it become this:

```python
previous_epoch_boundary_attester_indices: Iterable[ValidatorIndex]  = frozenset().union(*sets_of_previous_epoch_boundaries)
```
However, the variable annotation feature is not supported in Python3.5.

At the end of the day, this makes CI happy.

```python
    previous_epoch_boundary_attester_indices = frozenset.union(
        frozenset(),
        *sets_of_previous_epoch_boundary_participants,
    )
```

ChihChengLiang:
Move to Trinity

PR #1697, opened 2019-01-05
Slashable vote data cleanup

### LINKED ISSUE #1696: Cleanup slashable condition helpers.
### What is wrong?

See https://github.com/ethereum/py-evm/pull/1658#pullrequestreview-189543979

There are a few things that could be cleaned up in the linked pull request.

### How can it be fixed

Do the cleanup that is suggested in the linked review.

ralexstokes:
### What was wrong?

Fixes #1696.

### How was it fixed?

Implemented the suggested changes from the review here: https://github.com/ethereum/py-evm/pull/1658#pullrequestreview-189543979

#### Cute Animal Picture


ralexstokes:
i also found a bug in the test code so good thing you took a second look!

and while normally we should add a test for the broken test case, i want to discuss updating the BLS spec before making any further changes

tests/beacon/test_helpers.py
```diff
@@ -61,6 +63,10 @@
 )
 
 
+class UnreachableCodePathError(Exception):
```
pipermerriam: We've tended to just use `Exception` for this as it's more of a visual indication that the code path can't be reached and were we to encounter the actual exception the fix should be quite straight forward.

ralexstokes: sounds good -- was electing to be as explicit as possible but what you say makes sense

tests/beacon/test_helpers.py
```diff
@@ -874,24 +893,27 @@ def _create_slashable_vote_data_messages(params):
...
     # touch disjoint subsets of the provided params
     messages = _create_slashable_vote_data_messages(sample_slashable_vote_data_params)
 
+    fork_data = ForkData(**sample_fork_data_params)
     valid_params = _correct_slashable_vote_data_params(
         sample_slashable_vote_data_params,
         genesis_validators,
         messages,
         privkeys,
+        fork_data,
```
hwwhww: The conftest setting `state.fork_data == fork_data = ForkData(**sample_fork_data_params)` is not clear in the tests. I think we should use `state.fork_data` here:
```python
    state = BeaconState(**sample_beacon_state_params).copy(
        validator_registry=genesis_validators,
        fork_data=ForkData(**sample_fork_data_params),
    )

    ...

    valid_params = _correct_slashable_vote_data_params(
        sample_slashable_vote_data_params,
        genesis_validators,
        messages,
        privkeys,
        state.fork_data,
    )
```

tests/beacon/test_helpers.py
```diff
@@ -931,13 +958,18 @@ def test_verify_slashable_vote_data(param_mapper,
     # touch disjoint subsets of the provided params
     messages = _create_slashable_vote_data_messages(sample_slashable_vote_data_params)
 
+    fork_data = ForkData(**sample_fork_data_params)
     params = _correct_slashable_vote_data_params(
         sample_slashable_vote_data_params,
         genesis_validators,
         messages,
         privkeys,
+        fork_data,
```
hwwhww: ditto, use `state.fork_data`.

tests/beacon/test_helpers.py
```diff
@@ -874,24 +893,27 @@ def _create_slashable_vote_data_messages(params):
...
     state = BeaconState(**sample_beacon_state_params)
 
     # NOTE: we can do this before "correcting" the params as they
     # touch disjoint subsets of the provided params
     messages = _create_slashable_vote_data_messages(sample_slashable_vote_data_params)
 
+    fork_data = ForkData(**sample_fork_data_params)
     valid_params = _correct_slashable_vote_data_params(
         sample_slashable_vote_data_params,
         genesis_validators,
```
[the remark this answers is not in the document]
ralexstokes: agreed

tests/beacon/test_helpers.py
ralexstokes: gotcha

ralexstokes:
@hwwhww i made those updates :) 

the related tests pass locally -- trying to sort why the CI failed... seems unrelated to these edits...

Issue #1550, opened 2018-12-05
Speed up `integer_squareroot` with `math`


hwwhww:
### What is wrong?
The [`integer_squareroot`](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#integer_squareroot) function could be logically equal to `math.floor(math.sqrt(value))`.

### How can it be fixed
I simply tried to use `math.floor(math.sqrt(value))`. However, it didn't pass the test case of big number:
```
actual     = 4294967296
expected   = 4294967295
value      = 18446744073709551615
```

We can investigate it later.

pipermerriam:
This works:

```python
def fast_integer_sqrt(value: int) -> int:
    with decimal.localcontext() as ctx:
        ctx.prec = 100
        return int(decimal.Decimal(value).sqrt())
```

Looks like it's about 4x faster as profiled using:

```python
import random
import timeit

def integer_sqrt(value: int) -> int:
    ...

def fast_integer_sqrt(value: int) -> int:
    ...

def test_time_base():
    print('base', timeit.timeit("integer_sqrt(random.randint(0, 2**256 - 1))", globals={"integer_sqrt": integer_sqrt, 'random': random}))

def test_time_fast():
    print('fast', timeit.timeit("fast_integer_sqrt(random.randint(0, 2**256 - 1))", globals={"fast_integer_sqrt": fast_integer_sqrt, 'random': random}))
```

pipermerriam:
You might experiment with the precision as 100 was chosen arbitrarily and is likely overkill though i don't know how much it will effect performance.

pipermerriam:
> Would you like me to submit a PR?

That would be great

hwwhww:
closing in favor of #1698, thanks to @onyb!

PR #1698, opened 2019-01-06
Implement fast integer square root

onyb:
### What was wrong?

See #1550.

### How was it fixed?

Used `decimal` module of Python to calculate the square root, which is ~4 times faster than the implementation described in Ethereum 2.0 specs.

#### Cute Animal Picture

### DIFF
--- a/eth/_utils/numeric.py
+++ b/eth/_utils/numeric.py
@@ -1,3 +1,4 @@
+import decimal
 import functools
 import itertools
 from typing import (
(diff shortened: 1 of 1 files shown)


eth/_utils/numeric.py
```diff
@@ -119,9 +119,8 @@ def integer_squareroot(value: int) -> int:
...
-    y = (x + 1) // 2
-    while y < x:
-        x = y
-        y = (x + value // x) // 2
-    return x
+    with decimal.localcontext() as ctx:
+        # Set precision to 128, since the largest square root of a
+        # 256-bit integer is a 128-bit integer.
+        ctx.prec = 128
+        return int(decimal.Decimal(value).sqrt())
```
hwwhww: Looks cooool...!
I think it's better to note the "256-bit integer" assumption in the docstring.

hwwhww:
defer to @pipermerriam on this.

PR #1700, opened 2019-01-07
Sync ethereum/eth2.0-specs#374 part1: types, constants, deposit_helpers

hwwhww:
### What was wrong?
ethereum/eth2.0-specs#374 updated a loooot logic of validator status.

### How was it fixed?
1. Sync the constants. Also solve partial #1688: remove `SHARD_PERSISTENT_COMMITTEE_CHANGE_PERIOD`.
2. Sync related data structures:
    - `DepositData`
	    - Rename field `value` to `amount`, amount in Gwei.
    - `ValidatorRecord`
	    - Add new fields:
            ```python
            # Slot when validator activated
           ('activation_slot', uint64),
            # Slot when validator exited
            ('exit_slot', uint64),
            # Slot when validator withdrew
            ('withdrawal_slot', uint64),
            # Slot when validator was penalized
            ('penalized_slot', uint64),
            # Status flags
           ('status_flags', uint64),
            ```
	        - `activation_slot`, `exit_slot`, `penalized_slot`, `withdrawal_slot` are usedto determine if a validator is active.
	        - `status_flags` is for indicating that the validator has initialized exist or being withdrawable.
	    - Remove old fields:
            ```python
           # Status code
            ('status', uint64),
            # Slot when validator last changed status (or 0)
            ('latest_status_change_slot', uint64),    
            ```
   
    - `ValidatorRegistryDeltaBlock`
3. `eth.beacon.helpers`
	1. Update `get_new_shuffing`
		1. Rename to `get_shuffling`
		2. Add `slot: SlotNumber` parameter: Invariant - if `get_shuffling(seed, validators, shard, slot)` returns some value `x`, it should return the same value `x` for the same seed and shard and possible future modifications of validators forever in phase 0, and until the ~1 year deletion delay in phase 2 and in the future.
	2. Update `get_active_validator_indices`: add `slot: SlotNumber` parameter.
4. Refactor
	1. `BeaconState`
		1. Update `BeaconState.update_validator(validator_index: ValidatorIndex, validator: ValidatorRecord, balance: Gwei)` API.
		2. Check `len(validator_registry) == len(validator_balances)`
5. Bugix: `denoms.gwei != GWEI_PER_ETH` 🙈 
    - Define `GWEI_PER_ETH` in `eth/beacon/constants.py` and replace `denoms.gwei` with it.
6. Apply PR#1677 [feedback](https://github.com/ethereum/py-evm/pull/1677#discussion_r245436688)

#### Cute Animal Picture
dog-2983021_640


djrtwo:
I think `denoms.gwei` does equal `GWEI_PER_ETH`..

I just ran this locally:
```python
>>> import eth_utils
>>> eth_utils.denoms
<class 'eth_utils.currency.denoms'>
>>> eth_utils.denoms.gwei
1000000000
>>> eth_utils.denoms.gwei == 10**9
True
```

hwwhww:
@djrtwo 
I did that months ago! But I think that's just a coincidence. `eth_utils.denoms.gwei` is "wei per gwei" (see https://github.com/ethereum/eth-utils/blob/master/eth_utils/units.py), conceptionally different. 😅

djrtwo:
aaahh. I see
Tricky!

Reviewing PR now

tests/beacon/conftest.py
```diff
@@ -553,22 +621,42 @@ def genesis_state(sample_beacon_state_params,
...
+
+
+@pytest.fixture
+def activated_genesis_validators(initial_validators, genesis_slot):
+    """
+    Active
+    """
+    validators = tuple()
+    for validator in initial_validators:
+        validators += (validator.copy(activation_slot=genesis_slot),)
```
pipermerriam: *nitpick

This function *appears* to be well suited for using `eth_utils.to_tuple` and converting this to a `yield` pattern

tests/beacon/types/test_states.py
```diff
@@ -20,62 +20,39 @@
...
-
 def test_defaults(sample_beacon_state_params):
     state = BeaconState(**sample_beacon_state_params)
     assert state.validator_registry == sample_beacon_state_params['validator_registry']
     assert state.validator_registry_latest_change_slot == sample_beacon_state_params['validator_registry_latest_change_slot']  # noqa: E501
 
 
+def test_validator_registry_and_balances_length(sample_beacon_state_params, far_future_slot):
+    # When len(BeaconState.validator_registry) != len(BeaconState.validtor_balances)
+    with pytest.raises(ValueError):
```
pipermerriam: Is `ValidationError` a more appropriate error here?

tests/beacon/helpers.py
```diff
@@ -1,22 +1,36 @@
...
 from eth.beacon.types.validator_records import (
     ValidatorRecord,
 )
 
-from eth_utils import to_tuple
-
 
-def mock_validator_record(pubkey):
+def mock_validator_record(pubkey,
+                          far_future_slot,
```
djrtwo: I'm starting to think passing around `far_future_slot` doesn't really make much sense. It is more akin to `GWEI_PER_ETH` and `EMPTY_SIGNATURE` than some of the other configurable constants. Specifically, there is no reason that `far_future_slot` will ever be changed in value unless the bit length of slot is increased or decreased in some deployment.

My opinion is to define as `FAR_FUTURE_SLOT` and remove it from all function params and instead import and use where needed.

Curious to hear both of your thoughts on this @hwwhww and @pipermerriam 

context for @pipermerriam -- `far_future_slot` is a default value used for many of the validator fields and is defined as the max value in `uint64` (`64**2 - 1`). Unlike constants things like `epoch_length` it will never be configured differently depending on the deployment context.

tests/beacon/helpers.py
```diff
@@ -36,3 +50,17 @@ def get_pseudo_chain(length, genesis_block):
             parent_root=block.root
         )
         yield block
+
+
+def sign_proof_of_possession(deposit_input, privkey, domain):
+    return bls.sign(deposit_input.root, privkey, domain)
```
djrtwo: This is a bit out of the context of this PR, but this probably should take `fork_data` and calculate the `domain` locally using `DOMAIN_DEPOSIT` along with the `fork_data` param. This is instead of forcing the user of this function to pull in `DOMAIN_DEPOSIT` and calculate the domain before passing it.

eth/beacon/types/validator_records.py
```diff
@@ -65,28 +66,30 @@ def __init__(self,
...
+    def is_active(self, slot: int) -> bool:
         """
-        Returns ``True`` if the validator is active.
+        Return ``True`` if the validator is active.
         """
-        return self.status in VALIDATOR_RECORD_ACTIVE_STATUSES
+        return self.activation_slot <= slot < self.exit_slot
 
     @classmethod
     def get_pending_validator(cls,
```
djrtwo: @pipermerriam what are you thoughts on using `get_` naming convention when it is never a retrieval and only a creation? I personally think `create_` or `new_` is clearer in this context.

Is there a previously established convention around this in py-evm?

tests/beacon/test_helpers.py
```diff
@@ -480,31 +483,25 @@ def mock_get_shard_committees_at_slot(state,
...
         for i in range(3)
     ]
-    active_validator_indices = get_active_validator_indices(validators)
+    active_validator_indices = get_active_validator_indices(validators, current_slot)
     assert len(active_validator_indices) == 3
 
-    # Make one validator becomes ACTIVE_PENDING_EXIT.
     validators[0] = validators[0].copy(
-        status=ValidatorStatusCode.ACTIVE_PENDING_EXIT,
+        activation_slot=2,  # activation_slot < current_slot
```
djrtwo: The comment here is incorrect -- "activation_slot > current_slot"

I think the following will be clearer
```suggestion
        activation_slot=current_slot + 1,  # activation_slot > current_slot
```

tests/beacon/test_helpers.py
```diff
@@ -873,18 +843,21 @@ def _create_slashable_vote_data_messages(params):
...
                                               sample_beacon_state_params,
-                                              genesis_validators,
+                                              activated_genesis_validators,
+                                              genesis_balances,
                                               sample_slashable_vote_data_params):
-    sample_beacon_state_params["validator_registry"] = genesis_validators
-    state = BeaconState(**sample_beacon_state_params)
+    state = BeaconState(**sample_beacon_state_params).copy(
+        validator_registry=activated_genesis_validators,
+        validator_balances=activated_genesis_validators,
```
djrtwo: should assign `genesis_balances` here

```suggestion
        validator_balances=genesis_balances,
```

tests/beacon/types/test_states.py
djrtwo: The _user_ should never be able to provide data/input that would cause this to trigger. A _developer_ could. Does that make it a `ValidationError`?

hwwhww: @djrtwo yes, that's also what I thought! We should ensure `len(BeaconState.validator_registry) == len(BeaconState.validtor_balances)` during writing  `eth` APIs.

tests/beacon/helpers.py
hwwhww: @djrtwo Agreed! I think `far_future_slot` could be moved to `eth.beacon.contants`.

tests/beacon/types/test_states.py
pipermerriam: @djrtwo 

- `ValidationError` **if** it is *possible* for this code path to be encountered under some form of live conditions like *somehow* ending up with an invalid state where the registry and balances get out of sync.
- `ValueError` for lower level errors, normally indicative of us having some kind of broken code which is passing in the wrong type of value or a malformed value.

Admittedly these two definitions are not fully independent.  Maybe rule of thumb is: 1) A user should *never* see a `ValueError`, nor should anything like the `ethereum/tests` end up triggering a `ValueError`.  `ValidationError` **may** occasionally show up for users and is likely a regular result for many of the `ethereum/tests` style tests which test edge cases.

eth/beacon/types/validator_records.py
```diff
@@ -79,7 +79,8 @@ def __init__(self,
 
     def is_active(self, slot: int) -> bool:
         """
-        Return ``True`` if the validator is active.
+        Return ``True`` if the validator is active.Return ``True``
```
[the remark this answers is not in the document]
hwwhww: whoops!!! thanks for the catch!!!

eth/beacon/types/validator_records.py
hwwhww: `create_` sounds good to me as it has been used in `eth.rlp.BaseTransaction.create_unsigned_transaction`.

tests/beacon/types/test_states.py
djrtwo: Hmmm. I'm still a bit unclear on this.
There is an invariant that and addition to or removal from one of these arrays should always require an symmetric addition/removal from the other array. All helper functions such as `add_pending_validator` should enforce this invariant.

The state would only be out of sync if one of these helper functions was broken or if a user decided to manipulate/create a state directly with mismatched arrays. Because of the assumed atomicity of operations affecting the length of these two arrays, the network should never be able to provide data that gets them out of sync. This is different than say, the network providing an invalid block that throws a `ValidationError` when processing.

I somewhat think `ValueError`, but I am still trying to grok the difference and hope for some more clarity :)

hwwhww: @pipermerriam thanks for clarifying! FYI the reason why we put validator balances into a separate list in BeaconState (#1647) was clarified [here](https://github.com/ethereum/eth2.0-specs/issues/386#issuecomment-451033847) by Justin:
> It's an optimisation to minimise the amount of hashing when computing the beacon state root. The `validator_registry` (which contains most of the state) changes slowly and requires little hashing effort to maintain. The `validator_balances` is a smaller part of the state with much higher churn, hence why it is segregated.

So I do think this `ValueError` here fits "*for lower level errors, normally indicative of us having some kind of broken code which is passing in the wrong type of value or a malformed value*" since it's for avoiding reckless coding.

pipermerriam: :+1: seems like this could fit either case and that's expected since the two cases do have some conceptual overlap.  No strong opinion of what should be done if anything here.

eth/beacon/helpers.py
```diff
@@ -189,13 +186,14 @@ def _get_shards_committees_for_shard_indices(
...
-                      target_committee_size: int,
-                      shard_count: int) -> Iterable[Iterable[ShardCommittee]]:
+def get_shuffling(*,
+                  seed: Hash32,
+                  validators: Sequence['ValidatorRecord'],
+                  crosslinking_start_shard: ShardNumber,
+                  slot: SlotNumber,
+                  epoch_length: int,
+                  target_committee_size: int,
+                  shard_count: int) -> Iterable[Iterable[ShardCommittee]]:
```
pipermerriam: *minor* nitpick

Return type here would be more correct as a `Tuple` type.

eth/beacon/types/validator_records.py
```diff
@@ -95,9 +100,12 @@ def get_pending_validator(cls,
             withdrawal_credentials=withdrawal_credentials,
             randao_commitment=randao_commitment,
             randao_layers=0,
-            status=ValidatorStatusCode.PENDING_ACTIVATION,
-            latest_status_change_slot=latest_status_change_slot,
+            activation_slot=FAR_FUTURE_SLOT,
+            exit_slot=FAR_FUTURE_SLOT,
+            withdrawal_slot=FAR_FUTURE_SLOT,
```
pipermerriam: I assume that we are using `FAR_FUTURE_SLOT` to represent the concept of *undefined* or `None` since it works out to something like 5 billion centuries in the future?

djrtwo: Yes, it is essentially an "unset" flag on the var. It works nicely because we can use it in status checks.'
example: is_active -- `v.activation_slot <= state.slot < v.exit_slot`
* in the case that neither `activation_slot` nor `exit_slot` have been updated, the validator is not active.
* in the case that `activation_slot` is set but is in the future, the validator is non active
* in the case that `activation_slot` is in the past, but `exit_slot` is not yet set, then the validator is active
* in the case that `activation_slot` is in the past` and `exit_slot` has been set but is still in the future, the validator is active
* in the case that `activation_slot` and `exit_slot` are in the past, then the validator is not active and is instead exited.

pipermerriam: Is there any way that a validator could set any of these three fields to this value themselves?  If so, it might be useful to explicitely define `2**64 - 1` as a disallowed value so that we have strong assurances that when any of these values is equal to `2**64 -1` we **know** that it means unset.

djrtwo: Yeah, that seems reasonable.

These values can only change via state transitions that a validator can trigger but not control the exact value of (e.g. requesting and `exit` which would set their `exit_slot` as `state.slot + ENTRY_EXIT_DELAY`, or getting penalized which would set their`penalization_slot` to `state.slot`). These state transitions all change the value from `FAR_FUTURE_SLOT` to a value that is a function of the current `state.slot`

hwwhww:
@djrtwo I'm going to merge it. Feel free to give post-merge review if any, I'll amend in the part2 PR. :)

Issue #1648, opened 2018-12-19
Persisting an iterable of headers or beacon blocks fails


jannikluhn:
* py-evm Version: current master

### What is wrong?

`HeaderDB.persist_header_chain` iterates twice over the headers which fails if they are given in the form of an iterator.

First loop: https://github.com/ethereum/py-evm/blob/master/eth/db/header.py#L219-L227
Second loop: https://github.com/ethereum/py-evm/blob/master/eth/db/header.py#L240-L251
It also extracts the first element: https://github.com/ethereum/py-evm/blob/master/eth/db/header.py#L214

The same is bug is present in `eth.beacon.db.chain._persist_block_chain` where I originally encountered it.

### How can it be fixed

I can see three simple fixes:
- Handle everything in a single loop
- Convert the iterable to a tuple in the beginning
- Change the type hint to only accept sequences

cburgdorf:
Restricting that to `Sequence[BlockHeader]` to me feels like the best choice here.

cburgdorf:
Well, if we *can* handle everything in a single loop without sacrificing readability than maybe **that** would be the best choice as it would also be a slight performance improvement but I haven't given that much thought. I'm not a fan of converting it to a tuple in the beginning because then we are effectively looping over the data three times (converting to tuple + two existing loops)

pipermerriam:
My preference:

1) look into handling things in a single loop.  If can be cleanly done, do the change **and**  add a test that uses a generator and ensure that it works as expected
2) if single loop isn't viable for whatever reason, use `Sequence` approach.

PR #1701, opened 2019-01-07
[WIP] Refactor DB persistence logic to work with iterators

onyb:
### What was wrong?

See #1648 for details.

1. `HeaderDB.persist_header_chain` was failing when passed an iterator of `BlockHeader` objects.
2. We were looping over the headers multiple timesm which had a performance cost. This was actually also the cause for the above failure.

### How was it fixed?

I eliminated the need to loop on the sequence of headers twice. Essentially, I'm handling chain validation and setting of scores inside a single loop. I did some small refactoring as well to make the function more flat and readable.

**TODO**

- [ ] Fix a similar bug in the `eth.beacon.db.chain._persist_block_chain` module.
- [ ] Implement unit tests

#### Cute Animal Picture

### DIFF
--- a/eth/db/header.py
+++ b/eth/db/header.py
@@ -5,6 +5,7 @@ from typing import Iterable, Tuple
 import rlp
 
 from cytoolz import (
+    concat,
     first,
     sliding_window,
 )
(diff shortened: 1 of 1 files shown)


onyb:
@pipermerriam For implementing tests, should I duplicate all the tests in `tests/database/test_header_db.py` that use the `HeaderDB.persist_header_chain` method to test the same thing but using an iterator of `BlockHeader` objects?

eth/db/header.py
```diff
@@ -204,62 +205,78 @@ def persist_header_chain(self,
...
+    def _set_hash_scores_to_db(
+            cls,
+            db: BaseDB,
+            header: BlockHeader,
+            score: int
+    ) -> int:
+        db.set(
+            header.hash,
+            rlp.encode(header),
+        )
```
pipermerriam: This feels odd.  Why does this function also persist the header itself?

onyb: To give a context, the code is coming from this line: https://github.com/ethereum/py-evm/pull/1701/files#diff-ad4737f864351f38187efa966585825eL241

pipermerriam: With the current name of the function `_set_hash_scores_to_db` this code doesn't belong because persisting the header itself isn't the same as persisting the header's total difficulty/score.


I'd be fine with either 1) renaming this function to something like `_persist_header_to_db` or 2) just inlining this functionality in the two places it is needed within the `persist_header_chain` function.

Preference is currently for option 2

pipermerriam:
> For implementing tests, should I duplicate all the tests in `tests/database/test_header_db.py` that use the `HeaderDB.persist_header_chain` method to test the same thing but using an iterator of BlockHeader objects?

I don't think they all need to be duplicated.  Probably just need two tests.

- iterator of at least 3 headers
- iterator of a single header

That should be sufficient to show that it works as expected.

eth/db/header.py
onyb: Thanks for your input. I went for option 2.

onyb:
@pipermerriam I implemented the requested unit test for checking if `HeaderDB.persist_header_chain` works with an iterator. Please review this.

I would also like to do with same for `eth.beacon.db.chain.persist_block_chain`, but I couldn't find any example tests that create a chain of Beacon blocks (I was looking here: `tests/beacon/db/test_beacon_chaindb.py`). It seems that the `persist_block_chain` function is not well tested even for list of `BeaconBlock` objects. Could you please tell me how to go about this?

eth/beacon/db/chain.py
```diff
@@ -289,61 +290,79 @@ def persist_block_chain(
...
+            if parent.root != child.parent_root:
+                raise ValidationError(
+                    "Non-contiguous chain. Expected {} to have {} as parent but was {}".format(
+                        encode_hex(child.root),
+                        encode_hex(parent.root),
+                        encode_hex(child.parent_root),
+                    )
+                )
 
+            curr_block_head = child
```
pipermerriam: If I'm reading this correctly we *should* be able to drop this assignment and just use `child` in all of the places where `curr_block_head` is used from here through the end of the function, with the possible need to do some minor additional handling for the calls to `cls._set_as_canonical_chain_head`.

eth/beacon/db/chain.py
```diff
@@ -289,61 +290,79 @@ def persist_block_chain(
         with self.db.atomic_batch() as db:
             return self._persist_block_chain(db, blocks, self.block_class)
 
+    @classmethod
+    def _set_block_scores_to_db(
+            cls,
+            db: BaseDB,
+            block: BaseBeaconBlock
+    ) -> None:
+        # TODO: It's a stub before we implement fork choice rule
```
pipermerriam: I'm confused about this comment.  Leftover?

eth/beacon/db/chain.py
onyb: That's right, though I'm not sure there is a better way to achieve this than what I already tried to do here. The biggest constraints we have here that prevented me from coming up with a better approach were -
1. The chain can be an iterator.
2. We want to loop only once

I'm open to ideas.

eth/beacon/db/chain.py
onyb: [This](https://github.com/ethereum/py-evm/commit/ed324233c66d5722e27f9fd92266661af2562934#diff-f153130c7072bcb8bd89fe3da54eddb9R297) is the source of the comment.

Maybe this comment should be moved [to this line](https://github.com/ethereum/py-evm/pull/1701/files/53de28a49c6926e7827680113d012dc0a1fa2d23..4e5ffba6c331e31b39cec3e49e8d95efcaf90d89#diff-f153130c7072bcb8bd89fe3da54eddb9R356). What do you think?

hwwhww: This was saying that we haven't implemented the fork choice rule as [the spec](https://github.com/ethereum/eth2.0-specs/blob/master/specs/core/0_beacon-chain.md#beacon-chain-fork-choice-rule) (Latest Message Driven (LMD) Greediest Heaviest Observed SubTree (GHOST)). We currently just set the `score` to the block slot number as next line (`score = block.slot`) of this comment, and by that time we were not sure if we are going to use IMD (Immediate Message Driven (IMD) or LMD.

Could add more context:
```suggestion
        # TODO: To implement the new beacon fork choice rule as the spec.
        # Current workaround is setting score to the block slot number.
```

PR #1705, opened 2019-01-09
Implement get_transaction_receipt function as part of ChainDB

Bhargavasomu:
### What was wrong?
The ChainDB API could use the `get_transaction_receipt` method. This function is aimed to get the `Receipt` of the `Transaction` specified by the `Transaction Index` in the block of the provided `BlockHeader`. This function would also be useful to implement the `eth_getTransactionReceipt` JSON RPC function.

### How was it fixed?
The `Transaction Index` is RLP encoded and searched in the `receipt_db` whose root would be present in the `Header`.

#### Cute Animal Picture

### DIFF
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -398,8 +398,7 @@ class ChainDB(HeaderDB, BaseChainDB):
 
     def get_receipt_by_index(self,
                              block_number: BlockNumber,
-                             receipt_index: int,
-                             receipt_class: Type[Receipt]) -> Receipt:
+                             receipt_index: int) -> Receipt:
         """
         Returns the Receipt of the transaction at specified index
         for the block header obtained by the specified block number
@@ -413,7 +412,7 @@ class ChainDB(HeaderDB, BaseChainDB):
         receipt_key = rlp.encode(receipt_index)
         if receipt_key in receipt_db:
             receipt_data = receipt_db[receipt_key]
-            return rlp.decode(receipt_data, sedes=receipt_class)
+            return rlp.decode(receipt_data, sedes=Receipt)
         else:
             raise ReceiptNotFound(
                 "Receipt with index {} not found in block".format(receipt_index))
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -28,6 +28,7 @@ from eth.constants import (
 )
 from eth.exceptions import (
     HeaderNotFound,
+    ReceiptNotFound,
     TransactionNotFound,
 )
 from eth.db.header import BaseHeaderDB, HeaderDB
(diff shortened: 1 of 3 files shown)


Bhargavasomu:
@pipermerriam @cburgdorf review please.

eth/db/chain.py
```diff
@@ -395,6 +395,22 @@ def get_transaction_index(self, transaction_hash: Hash32) -> Tuple[BlockNumber,
         transaction_key = rlp.decode(encoded_key, sedes=TransactionKey)
         return (transaction_key.block_number, transaction_key.index)
 
+    def get_transaction_receipt(self,
```
pipermerriam: I think this function name and signature should mirror the existing `ChainDB.get_transaction_by_index` function, meaning that rather than a `header` it should operate on a `block_number` and `receipt_index`.

Bhargavasomu:
@pipermerriam in the latest push, the function signature has been changed to make use of the 
`Block Number` instead of `Block Header` and also I have changed the function signature to `get_receipt_of_transaction`. Hence to suit with this function name, I thought it would be better to just use `transaction_index` instead of `receipt_index`. But if these names are confusing, will change them. Please take a look. Thankyou.

eth/db/chain.py
```diff
@@ -395,6 +395,28 @@ def get_transaction_index(self, transaction_hash: Hash32) -> Tuple[BlockNumber,
         transaction_key = rlp.decode(encoded_key, sedes=TransactionKey)
         return (transaction_key.block_number, transaction_key.index)
 
+    def get_receipt_of_transaction(self,
+                                   block_number: BlockNumber,
+                                   transaction_index: int,
```
pipermerriam: Yeah, `receipt_index` would be better.  :+1:

eth/db/chain.py
```diff
@@ -395,6 +395,28 @@ def get_transaction_index(self, transaction_hash: Hash32) -> Tuple[BlockNumber,
         transaction_key = rlp.decode(encoded_key, sedes=TransactionKey)
         return (transaction_key.block_number, transaction_key.index)
 
+    def get_receipt_of_transaction(self,
```
pipermerriam: Should be `get_receipt_by_index` to mirror the naming convention established by `get_transaction_by_index`

Bhargavasomu:
@pipermerriam could you merge this if possible, so that I could do the `JSON RPC` work in the `trinity` repo. (If there are no further changes that I have to do on this)

eth/db/chain.py
```diff
@@ -395,6 +396,28 @@ def get_transaction_index(self, transaction_hash: Hash32) -> Tuple[BlockNumber,
...
+            raise ReceiptNotFound("Block {} is not in the canonical chain".format(block_number))
+
+        receipt_db = HexaryTrie(db=self.db, root_hash=block_header.receipt_root)
+        receipt_key = rlp.encode(receipt_index)
+        if receipt_key in receipt_db:
+            receipt_data = receipt_db[receipt_key]
+            return rlp.decode(receipt_data, sedes=receipt_class)
+        else:
+            raise ReceiptNotFound(
+                "Transaction with index {} not found in block".format(receipt_index))
```
pipermerriam: `Transaction -> Receipt`

Bhargavasomu:
@pipermerriam I have added the tests. Please note that I have not used `pytest.mark.parametrize`, as I thought that adding more test cases would take a longer time. But nevertheless, let me know if more tests are needed.

eth/db/chain.py
```diff
@@ -395,6 +396,28 @@ def get_transaction_index(self, transaction_hash: Hash32) -> Tuple[BlockNumber,
         transaction_key = rlp.decode(encoded_key, sedes=TransactionKey)
         return (transaction_key.block_number, transaction_key.index)
 
+    def get_receipt_by_index(self,
+                             block_number: BlockNumber,
+                             receipt_index: int,
+                             receipt_class: Type[Receipt]) -> Receipt:
```
pipermerriam: Is there a reason this needs to be provided by the user?  Currently all vms use the same `Receipt` class.

Bhargavasomu: @pipermerriam I just put that up, so as to mimic the `get_transaction_by_index` function. Guess I will add a default value which is `Receipt` to this function.

Bhargavasomu:
@pipermerriam please take a look now.

PR #1707, opened 2019-01-09
[WIP] Add Stack and Computation Benchmark

Bhargavasomu:
### What was wrong?
As part of the discussion [here](https://github.com/ethereum/py-evm/pull/1666#issuecomment-451587778), a new `Stack Benchmark` is needed.

### How was it fixed?
A `TestStack.sol` contract was written which would serve as the benchmark.

#### Cute Animal Picture


Bhargavasomu:
@pipermerriam I still haven't made this a benchmark as of yet, but this is my first time where I am running a contract on the `EVM`, and I just had a small doubt. 

* For the above `TestStack Contract` which you mentioned, I ran that code through our `EVM`, but the final stack values that I am seeing is `[2729995958]`. As per the above code shouldn't the expected value be `[10,000]`.
* If at all this is stored in `Memory` (which is in bytes), how could I obtain the final value of `v`, because I would like to put an `assert` statement there.

Or is there anything else I am missing? Could you please clarify? Thankyou

Bhargavasomu:
Also you can find the above stated values in `py36-stack-benchmark`

Bhargavasomu:
@pipermerriam I am extremely sorry, I think I messed up with `git` and it is showing some of the previous commits as mine. Will fix it immediately.

pipermerriam:
For the benchmarking of this I think the final value doesn't matter very much.  It's more about determining how many times the stack is pushed/popped from (and maybe even in what quantities).  

I only just now realized that https://github.com/ethereum/py-evm/pull/1515 still hasn't been completed as it would provide an ideal API for this.  Without that I think you're limited to manually measuring it using something like `unittest.mock` and then just hard coding the numbers in the benchmark.

Then, suppose the contract code results in 150 stack pushes and 150 stack pops, then you can output some metrics like `N pops / second`. You should probably run the computation 100 times or something like that to get a sufficiently large sample size.

Then later when #1515 is merged we can adjust the code to dynamically measure the number of push/pops.

kclowes:
Closing, stale.

PR #1709, opened 2019-01-09
[WIP] debug_traceTransaction endpoint implementation

pipermerriam:
Replaces: https://github.com/ethereum/py-evm/pull/1515 , removing all of the `trinity` components which will be added in a separate PR to the `trinity` repo once this is merged and released.

### What was wrong?

https://github.com/ethereum/go-ethereum/wiki/Tracing:-Introduction

Adds support for tracing EVM transactions.

### How was it fixed?

This adds a new API that can be used to trace transactions by supplying a `tracer` to `VM.apply_transaction`

#### Cute Animal Picture

23-hilarious-photos-of-surprised-animals-19

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -470,7 +470,7 @@ class VM(BaseVM):
 
         # Execute it in the VM
         with self.state.trace(tracer):
-            return self.state.get_computation(message, transaction_context).apply_computation(
+            return self.state.get_computation_class().apply_computation(
                 self.state,
                 message,
                 transaction_context,
--- a/eth/vm/state.py
+++ b/eth/vm/state.py
@@ -225,17 +225,20 @@ class BaseState(Configurable, ABC):
     #
     # Computation
     #
+    @classmethod
+    def get_computation_class(cls) -> 'BaseComputation':
+        if cls.computation_class is None:
+            raise AttributeError("No `computation_class` has been set for this State")
+        else:
+            return cls.computation_class
+
     def get_computation(self,
                         message: Message,
                         transaction_context: 'BaseTransactionContext') -> 'BaseComputation':
         """
         Return a computation instance for the given `message` and `transaction_context`
         """
-        if self.computation_class is None:
-            raise AttributeError("No `computation_class` has been set for this State")
-        else:
-            computation = self.computation_class(self, message, transaction_context, self.tracer)
-        return computation
+        return self.get_computation_class()(self, message, transaction_context, self.tracer)
 
     #
     # Transaction context


eth/vm/computation.py
```diff
@@ -559,30 +577,37 @@ def apply_create_message(self) -> 'BaseComputation':
...
                     "OPCODE: 0x%x (%s) | pc: %s",
                     opcode,
                     opcode_fn.mnemonic,
-                    max(0, computation.code.pc - 1),
+                    pc,
                 )
 
                 try:
-                    opcode_fn(computation=computation)
+                    with computation.tracer.capture(computation, opcode_fn):
```
pipermerriam: I'm not thrilled with the overhead this incurs for EVM execution but I'm inclined to leave this be and optimize later when we get around to actual EVM optimizations.

eth/vm/computation.py
```diff
@@ -1,6 +1,6 @@
 from abc import (
     ABC,
-    abstractmethod
+    abstractmethod,
```
lithp: Did you catch this manually or you have a tool you run to manage your imports?

eth/vm/base.py
```diff
@@ -111,8 +115,9 @@ def logger(self) -> logging.Logger:
     @abstractmethod
     def apply_transaction(self,
                           header: BlockHeader,
-                          transaction: BaseTransaction
-                          ) -> Tuple[BlockHeader, Receipt, BaseComputation]:
+                          transaction: BaseTransaction,
+                          *,
```
lithp: Most of my python experience is with 2.7. I'm still a little bitter about `reduce` and [3113](https://www.python.org/dev/peps/pep-3113/) but wow, this is a nice feature.

eth/vm/base.py
```diff
@@ -455,11 +469,13 @@ def execute_bytecode(self,
         )
 
         # Execute it in the VM
-        return self.state.get_computation(message, transaction_context).apply_computation(
-            self.state,
-            message,
-            transaction_context,
-        )
+        with self.state.trace(tracer):
```
lithp: I think this call to `trace` is unnecessary. Neither `get_computation_class` nor `apply_computation` use `state`'s tracer attribute.

eth/vm/state.py
```diff
@@ -99,9 +103,19 @@ def logger(self) -> ExtendedDebugLogger:
         return cast(ExtendedDebugLogger, normal_logger)
 
     #
-    # Block Object Properties (in opcodes)
+    # Tracing
     #
+    @contextlib.contextmanager
+    def trace(self, tracer: BaseTracer) -> Iterator[None]:
```
lithp: I think the idea behind this context manager is that `BaseState` instances will outlive transactions, so tracer should only be set while it's running a transaction. But I think that's really a sign that tracer shouldn't be an attribute of `BaseState` at all.

Using a context manager for this seems error-prone and unnecessary. It's error-prone because you have to remember to use this whenever you're calling a method which might need the tracer. If tracer was passed in as a parameter then the mistake of not passing one would be much harder to make.

It's unnecessary because this code seems to take two approaches simultaneously. It both adds the tracer to vm_state, making it available nearly everywhere, and it also manually threads tracer through all the calls. Because the threading has already been done, it's simpler to use that tracer everywhere.

eth/vm/state.py
```diff
@@ -299,7 +316,7 @@ def __init__(self, vm_state: BaseState) -> None:
     def __call__(self, transaction: BaseOrSpoofTransaction) -> 'BaseComputation':
         valid_transaction = self.validate_transaction(transaction)
         message = self.build_evm_message(valid_transaction)
-        computation = self.build_computation(message, valid_transaction)
+        computation = self.build_computation(message, valid_transaction, self.vm_state.tracer)
```
lithp: Here's somewhere it'd be difficult to thread the tracer too, this might be the origin of the context manager? I've exhausted my timebox for looking into this PR so I don't have an answer but I expect that getting a tracer here isn't insurmountable.

Looking at this now, BaseState is long-lived but Message isn't, maybe it's a better place to attach the tracer?

eth/vm/forks/frontier/state.py
```diff
@@ -128,14 +136,16 @@ def build_computation(self,
...
                     message,
-                    transaction_context,
-                ).apply_create_message()
-        else:
-            computation = self.vm_state.get_computation(
-                message,
-                transaction_context).apply_message()
+                    transaction_context).apply_message()
 
         return computation
```
lithp: `apply_message()` calls `apply_computation` which throws an exception if the tracer does not exist. This function returns `BaseComputation`'s which have an `apply_message` method but if that message is ever called an exception is thrown. The caller of this method will *probably* never call `apply_message` but this is still a smell.

eth/vm/forks/frontier/state.py
```diff
@@ -128,14 +136,16 @@ def build_computation(self,
                     encode_hex(message.storage_address),
                 )
             else:
+                with self.vm_state.trace(tracer):
+                    computation = self.vm_state.get_computation(
+                        message,
+                        transaction_context,
+                    ).apply_create_message()
```
lithp: An alternative might be to pass the tracer into `get_computation` or `apply_create_message`

eth/vm/computation.py
```diff
@@ -559,30 +577,37 @@ def apply_create_message(self) -> 'BaseComputation':
...
 
                 try:
-                    opcode_fn(computation=computation)
+                    with computation.tracer.capture(computation, opcode_fn):
+                        opcode_fn(computation=computation)
                 except Halt:
                     break
+
+        # Allow the tracer to record final values.
+        computation.tracer.finalize(computation)
```
lithp: I'm exhausted my timebox but I was planning at looking at this next. I think this fails if there are any child computations? When the child's apply_computation finishes the tracer is finalized. It then returns to the parent which attempts to continue tracing and the tracer will complain that it's already been finalized.

eth/vm/stack.py
```diff
@@ -35,6 +36,18 @@ def __init__(self) -> None:
     def __len__(self) -> int:
         return len(self.values)
 
+    def __iter__(self) -> Iterable[int]:
```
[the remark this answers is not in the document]
lithp: Also, why cast everything to int instead of returning `Iterator[Union[int, bytes]]`? It doesn't look like tracer relies on receiving just ints.

RaoulSchaffranek:
Are there any plans to continue the work on tracing?

Issue #1659, opened 2018-12-20
Beacon chain evacuation plan


hwwhww:
### What is wrong?

Per #1594 - `trinity` and `p2p` modules are moving to a new home - but the beacon chain is still here and being developed rapidly.

There are other tradeoffs if we use new branches + submodules.

### How can it be fixed

Follow `trinity` to the new repo like a stalker (!!!)
1. Collect the "eth2-only" thing into `eth/beacon/` or `eth/shard/`.
    - `eth/rlp/sedes.py`: e.g., `uint384 = BigEndianInt(384)`
    - `eth/_utils/bls.py`: will be moved to `ethereum/py-ecc` repo
    - `eth/_utils/blobs.py`: blobs and chunks, merkle root tool <- sharding
    - `eth/constants.py`: chunks <- sharding

    I probably missed something, but it's not much.
2. Add an `eth2` module in `trinity` repo, where `eth2` module has `eth2/beacon` - like what `eth/beacon` is now.
3. Hope `eth2` will be strong enough to go back to `py-evm` one day.

hwwhww:
@pipermerriam @cburgdorf @carver What do you think of it?

cburgdorf:
>but I think it might be worth exploring options that keep the beacon code in the `py-evm` repository, or to at least try them before this approach.

I agree that we should give it a try. I know it is annoying because now PRs need to be reviewed in tandem ("Please review PR X in Trinity but make sure to test it on top of PR Y in Py-EVM") but that's the main disadvantage of standalone repos vs mono repos. With just two repositories I think the overhead is still manageable.

If it turns out to be too annoying to work under that model I agree that we should do what @hwwhww proposes and introduce an `eth2` module in the trinity repo until things have stabilized. The main danger here that I see is that the `eth2` namespace could turn into a quickfix area of the code base. E.g. consider that we need to change an API that trinity uses from `eth` and this API is used in both the beacon node and the eth1 node. The proper fix obviously is to make the change in the `Py-EVM` repo, cut a release and then make the change in Trinity. But since that's a lot of hassle one might be tempted to just copy and paste the API to the `eth2` namespace and make the changes there.

jannikluhn:
To be honest, I suspect that the Trinity/py-evm split will be quite annoying for all work related to the beacon chain in Trinity. But I'm fine with trying it out for a couple of weeks and reevaluating when we know how it turned out in practice.

cburgdorf:
[...]

PR #1713, opened 2019-01-10
Create `eth2` module

hwwhww:
### What was wrong?
#1659 - part 1: Create `eth2` module

### How was it fixed?

1. Add `eth2` module and CI setting
2. Move eth2 specific utils from `eth._utils` to `eth2._utils`.
3. Move `eth.beacon` to `eth2.beacon`
4. Move Blobs and Chunks constants to `eth2/_utils/blobs.py`
5. Remove `eth.beacon` from tox and CI
6. Create `eth2/beacon/sedes.py` for beacon chain specific sedes

Note: To pass CI, I just pin the dependencies now before #1710 is fixed.

#### Cute Animal Picture

images


setup.py
```diff
@@ -7,17 +7,26 @@
     'eth': [
         "cryptography>=2.0.3,<3.0.0",
         "cytoolz>=0.9.0,<1.0.0",
-        "eth-bloom>=1.0.0,<2.0.0",
+        "eth-bloom==1.0.1",
```
cburgdorf: Any specific reason why the upper bound was removed?

setup.py
```diff
@@ -7,17 +7,26 @@
...
         "eth-keys>=0.2.0b3,<1.0.0",
         "eth-typing>=2.0.0,<3.0.0",
         "eth-utils>=1.3.0b0,<2.0.0",
         "lru-dict>=1.1.6",
         "mypy_extensions>=0.4.1,<1.0.0",
         "py-ecc>=1.4.7,<2.0.0",
         "pyethash>=0.1.27,<1.0.0",
-        "rlp>=1.0.3,<2.0.0",
```
cburgdorf: same here

setup.py
hwwhww: When using `eth-bloom>=1.0.0,<2.0.0`, the `mypy` check will fail as how #1710 is right now:
https://circleci.com/gh/ethereum/py-evm/124106?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
py35-lint runtests: commands[2] | mypy --follow-imports=silent --warn-unused-ignores --ignore-missing-imports --no-strict-optional --check-untyped-defs --disallow-incomplete-defs --disallow-untyped-defs --disallow-any-generics -p eth
eth/rlp/receipts.py:49: error: Cannot instantiate abstract class 'BloomFilter' with abstract attribute '__hash__'
eth/vm/base.py:415: error: Cannot instantiate abstract class 'BloomFilter' with abstract attribute '__hash__'
eth/vm/forks/frontier/blocks.py:67: error: Cannot instantiate abstract class 'BloomFilter' with abstract attribute '__hash__'
```

I think the reason why `master` branch *looks* fine now is due to Circle CI cache.

cburgdorf: Ah, just noticed. It's because of https://github.com/ethereum/py-evm/pull/1710

setup.py
hwwhww: ditto

Issue #705, opened 2018-05-16
After Byzantium, stop calculating state root after each transaction at all


carver:
### What is wrong?

Persisting all account changes to the trie and calculating the new state root is expensive. After Byzantium, we no longer need to calculate the state root after each transaction (because it has been removed from the receipt).

### How can it be fixed

Stop returning state root after every `apply_transaction()`, instead generating it only if the receipt requires it.

(Depends on PR #704 )

rayrapetyan:
@carver, state root is being updated in function "make_state_root" inside:
```
self._journaldb.persist()
self._journaltrie.persist()
```
calls. 
"stop calculating state root" also mean stop calling persist() (which I think is wrong?)
So do you mean "make_state_root" shouldn't be called from apply_transaction at all? 
Otherwise I don't see a way to separate state storing (persisting) and state root recalculation, this is an internal logic in DB and trie.hexary (inside persist() call).
Please clarify. Thanks.

pipermerriam:
@rayrapetyan this is going to be an incomplete answer, but I hope it points you in the right direction.

When `py-evm` was originally written, we didn't have the additional layers between the lowest level `HexaryTrie` implementation and the base `db` object that the EVM was using, and thus, we were re-calculating the state root on **every** db write.

This was improved by wrapping the base `HexaryTrie` database in a few layers which you can find by digging into the `eth.db.account.AccountDB` object and seeing the various journals and whatnot that wrap it.  As you pointed out, the `make_state_root` function causes all of the pending changes to be written to the database and then we retrieve the resulting state root.

One way I *think* this issue could be addressed would be to add one additional layer between the base `HexaryTrie` database and the `AccountDB` such that in between transactions we can still make a `persist` call (since we need that for journaling and rolling back changes for failed transactions) but such that it doesn't actually result in writing to the underlying `HexaryTrie` database.  Then at the end of all of the block's transactions we would make an additional call to some new API which persisted all of the changes down into the `HexaryTrie` database and computed the state root.

It'll take some investigation to dig up exactly where the changes need to go as well as figuring out a clean way for this new API to be implemented.

rayrapetyan:
[acc_dbs.pdf](https://github.com/ethereum/py-evm/files/2730226/acc_dbs.pdf)
Attached AccountDB schema.

pipermerriam:
[...]

PR #1716, opened 2019-01-15
Stop persisting state after Byzantium fork

rayrapetyan:
### What was wrong?

State was persisting after every apply_transaction call in all forks (https://github.com/ethereum/py-evm/issues/705)

### How was it fixed?

State is not persisting after every apply_transaction call after Byzantium fork

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -16,9 +16,6 @@ from typing import (
 )
 from typing import Set  # noqa: F401
 
-from eth_bloom import (
-    BloomFilter,
-)
 from eth_hash.auto import keccak
 from eth_typing import (
     Address,
@@ -111,7 +108,7 @@ class BaseVM(Configurable, ABC):
     def apply_transaction(self,
                           header: BlockHeader,
                           transaction: BaseTransaction
-                          ) -> Tuple[BlockHeader, Receipt, BaseComputation]:
+                          ) -> Tuple[Receipt, BaseComputation]:
         raise NotImplementedError("VM classes must implement this method")
 
     @abstractmethod
@@ -186,6 +183,15 @@ class BaseVM(Configurable, ABC):
     #
     # Headers
     #
+    @abstractmethod
+    def add_receipt_to_header(self, old_header: BlockHeader, receipt: Receipt) -> BlockHeader:
+        """
+        Apply the receipt to the old header, and return the resulting header. This may have
+        storage-related side-effects. For example, pre-Byzantium, the state root hash
+        is included in the receipt, and so must be stored into the database.
+        """
+        pass
+
     @classmethod
     @abstractmethod
     def compute_difficulty(cls, parent_header: BlockHeader, timestamp: int) -> int:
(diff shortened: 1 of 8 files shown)


eth/vm/forks/constantinople/state.py
```diff
@@ -7,3 +19,10 @@
 
 class ConstantinopleState(ByzantiumState):
     computation_class = ConstantinopleComputation
+
+    def apply_transaction(self, transaction: BaseOrSpoofTransaction) -> \
+            Tuple[bytes, BaseComputation]:
+        if self.state_root != BLANK_ROOT_HASH and not self.account_db.has_root(self.state_root):
+            raise StateRootNotFound(self.state_root)
+        computation = self.execute_transaction(transaction)
+        return EMPTY_SHA3, computation
```
carver: This approach of sometimes returning an empty root hash seems surprising from a callers perspective, and potentially error prone. I'll look at a bigger refactor to get it done with a new API.

carver: > This approach of sometimes returning an empty root hash seems surprising from a callers perspective, and potentially error prone. I'll look at a bigger refactor to get it done with a new API.

I added a commit showing the direction I was imagining. If you have more you want to add, let me know.

Otherwise, I'll ask for a last review and we should be able to get it merged as is.

carver:
This ends up being relevant for beam sync. It's the easiest way to make sure we only ask for nodes from the trie root before the block started (rather than from intermediate tries).

carver:
ping @lithp

eth/vm/base.py
```diff
@@ -553,10 +552,7 @@ def mine_block(self, *args: Any, **kwargs: Any) -> BaseBlock:
         """
         packed_block = self.pack_block(self.block, *args, **kwargs)
 
-        if packed_block.number == 0:
-            final_block = packed_block
-        else:
-            final_block = self.finalize_block(packed_block)
+        final_block = self.finalize_block(packed_block)
```
lithp: Is the idea here that we'll never try to mine the genesis block?

eth/vm/state.py
```diff
@@ -244,18 +244,19 @@ def get_transaction_context_class(cls) -> Type['BaseTransactionContext']:
...
-    def apply_transaction(self, transaction: 'BaseTransaction') -> Tuple[bytes, 'BaseComputation']:
+    def apply_transaction(
+            self,
+            transaction: BaseOrSpoofTransaction) -> 'BaseComputation':
         """
         Apply transaction to the vm state
 
         :param transaction: the transaction to apply
-        :return: the new state root, and the computation
+        :return: the new state root (before Constantinople), and the computation
```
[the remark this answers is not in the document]
carver: Yup, thanks! Just some doc cruft left over.

eth/vm/base.py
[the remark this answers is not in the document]
carver: Yup! In fact it's important to explicitly calculate the state root now and persist it to disk, after applying all the genesis state. That is part of the reason for the change.

carver:
> If this PR accidentally didn't persist state between each transactions, even during frontier, would that cause any tests to fail?

Yup, because the blockchain tests test the transaction receipts, which include the intermediate state roots pre-Byzantium.

> I think I'm not the right person to review this code. Everything looks good to me but without a deeper understanding of how this part of the codebase works I might be missing something. ... It might be nice to include some testing.

It's okay, I think that's enough for now. I'm counting on the test fixtures to confirm correctness about the state root calculation itself. Thanks for looking it over, even if it's just a gut check for code that is transparently dumb implementation-wise :).

PR #1718, opened 2019-01-17
enable account_cache

rayrapetyan:
### What was wrong?

account_cache was declared but not used causing ~30% of performance degradation when accessing accounts.

### How was it fixed?

Starting to use account_cache in get\set_account functions


cburgdorf:
This looks good to me. Should we include it in the planned release today/tomorrow? /cc @carver @pipermerriam

pipermerriam:
@cburgdorf I'd say :+1: on inclusion.

PR #1719, opened 2019-01-18
Add mysterious Petersburg fork

cburgdorf:
### What was wrong?

It looks like we are going forward with another fork that is like Constantinople but without EIP-1283. Since the name is still up for debate, I'm using "Petersburg" as a placeholder for the time being.

### How was it fixed?

Notice that implementation wise "Petersburg" isn't a child of Constantinople but a child of Byzantium which will make it easier to remove Constantinople entirely should we ever decide to remove support for it.

#### Cute Animal Picture


veox:
> Notice that implementation wise "Petersburg" isn't a child of Constantinople but a child of Byzantium (...)

Won't this skip a large set of upstream transition tests ("XToYAt5")?.. Those will most likely have have "Byzantium to Constantinople" and "Constantinople to Petersburg" transitions, and (possibly) no explicit "Byzantium to Petersburg".

cburgdorf:
The geth team went forward with the name `Petersburg` and included it in their [latest release](https://github.com/ethereum/go-ethereum/releases/tag/v1.8.22) so we might want to merge this as well?

cburgdorf:
Ok, after upgrading the fixtures to the latest commit `geth` uses this still needs some work. Seems it even fails the Byzantium tests with the new fixtures (cc @veox sounds like a riddle you would enjoy).

Also, `ethereum/tests` calls `Petersburg` `ConstantinopleFix` and tests for this fork still need to be configured.

veox:
The failing `BlockchainTests/GeneralStateTests/stRevertTest/RevertPrecompiledTouch_storage*` tests (a series of four) are new; this is not a regression in the strict sense of the word.

Taking a look. Will open an issue if there's too much text to fit here.

veox:
The `RevertPrecompiledTouch_storage` tests describe that precompiles at addresses from `0x01` to `0x08` [have storage set](https://github.com/ethereum/tests/blob/7d66cbfff1e6561d1046e45df8b7918d186b136f/src/GeneralStateTestsFiller/stRevertTest/RevertPrecompiledTouch_storageFiller.json#L199).

cburgdorf:
@veox thanks for digging up all this info! Do I understand that right that our Spurious Dragon implementation is wrong then?

I'm currently knee deep in a different Trinity issue so I'm hesitant to chase down a different rabbit whole. If you have a good idea about what needs to be done and are willing to make the change that would be superb!

Otherwise, I'd propose I cut out the test upgrade for now so that we can merge this and cut a new release before the fork happens. @pipermerriam @carver @lithp

veox:
> Do I understand that right that our Spurious Dragon implementation is wrong then?

It's a little more complicated. `REVERT` came in Byzantium, after Spurious Dragon, - is seems like the issue is in the interaction of the two.

> If you have a good idea about what needs to be done and are willing to make the change that would be superb!

I'm also a bit pre-occupied right now, but will try to look into remedies later today.

carver:
> Otherwise, I'd propose I cut out the test upgrade for now so that we can merge this and cut a new release before the fork happens.

Fine by me, if Spurious dragon is already broken, we don't need to delay ConstantiNOPEl implementation.

veox:
[Current failures](https://circleci.com/gh/ethereum/py-evm/126231?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) are from tests already in `INCORRECT_UPSTREAM_TESTS` (of PR #1224 fame), except for `ConstantinopleFix`.

I'll submit a quick fix.

tests/json-fixtures/test_blockchain.py
```diff
@@ -144,17 +144,24 @@
...
     # All four variants have been specifically added to test a collision type
     # like the above; therefore, they fail in the same manner.
     # * https://github.com/ethereum/py-evm/pull/1579#issuecomment-446591118
     ('GeneralStateTests/stSStoreTest/InitCollision_d0g0v0.json', 'InitCollision_d0g0v0_Constantinople'),  # noqa: E501
     ('GeneralStateTests/stSStoreTest/InitCollision_d1g0v0.json', 'InitCollision_d1g0v0_Constantinople'),  # noqa: E501
     ('GeneralStateTests/stSStoreTest/InitCollision_d2g0v0.json', 'InitCollision_d2g0v0_Constantinople'),  # noqa: E501
     ('GeneralStateTests/stSStoreTest/InitCollision_d3g0v0.json', 'InitCollision_d3g0v0_Constantinople'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d0g0v0.json', 'InitCollision_d0g0v0_ConstantinopleFix'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d1g0v0.json', 'InitCollision_d1g0v0_ConstantinopleFix'),  # noqa: E501
+    ('GeneralStateTests/stSStoreTest/InitCollision_d3g0v0.json', 'InitCollision_d3g0v0_ConstantinopleFix'),  # noqa: E501
```
veox: NB: `InitCollision_d2g0v0` not present here, since (it seems) in that particular case, the failure relates indirectly to net gas metering of EIP-1283.

Since that was disabled in Petersburg, it no longer surfaces.

cburgdorf:
>if Spurious dragon is already broken, we don't need to delay ConstantiNOPEl implementation.

Better! @veox fixed Spurious Dragon [with](https://github.com/ethereum/py-evm/pull/1719/commits/47c521f263833df03517d1126068e87b28d09c7b)

This is now complete with CI tests etc and can be given a full review.

eth/vm/forks/petersburg/computation.py
```diff
@@ -0,0 +1,29 @@
...
+from eth.vm.forks.byzantium.computation import (
+    ByzantiumComputation
+)
+
+from .opcodes import PETERSBURG_OPCODES
+
+PETERSBURG_PRECOMPILES = merge(
+    BYZANTIUM_PRECOMPILES,
+    {
+        # TODO: add new precompiles
```
carver: Both here and Constantinople can be simplified to:
```py
PETERSBURG_PRECOMPILES = BYZANTIUM_PRECOMPILES
```

eth/vm/forks/petersburg/headers.py
```diff
@@ -0,0 +1,13 @@
+from eth.vm.forks.byzantium.headers import (
+    configure_header,
+    create_header_from_parent,
+    compute_difficulty,
+)
+
+
+compute_petersburg_difficulty = compute_difficulty(5000000)
```
carver: Maybe `compute_difficulty(bomb_delay=5000000)` as a way to document what this magic number is.

tests/core/tester/test_generate_vm_configuration.py
```diff
@@ -121,9 +139,9 @@ def test_generate_vm_configuration(args, kwargs, expected):
 
         assert left_block == right_block
 
-        if right_vm == Forks.Frontier:
-            assert 'Frontier' in left_vm.__name__
-        elif right_vm == Forks.Homestead:
+        assert right_vm.value in left_vm.__name__
+
+        if right_vm == Forks.Homestead:
             assert 'Homestead' in left_vm.__name__
```
carver: I guess this assertion can be dropped now, too.

tox.ini
```diff
@@ -28,6 +28,7 @@ commands=
     native-blockchain-spurious_dragon: pytest {posargs:tests/json-fixtures/test_blockchain.py --fork EIP158}
     native-blockchain-byzantium: pytest {posargs:tests/json-fixtures/test_blockchain.py --fork Byzantium}
     native-blockchain-constantinople: pytest {posargs:tests/json-fixtures/test_blockchain.py --fork Constantinople}
+    native-blockchain-petersburg: pytest {posargs:tests/json-fixtures/test_blockchain.py --fork ConstantinopleFix}
```
carver: Hm, seems unfortunate to call it petersburg in one place and ConstantinopleFix in another. Obviously this isn't the first fork to have that problem, but I can't think of a good reason we'd want to continue the divergence.

cburgdorf: It's unfortunate that we didn't manage to reach consensus among all dev teams how to name this fork. Geth and Py-EVM both call it Petersburg but `ethereum/tests` calls it `ConstantinopleFix` which is why we have `ConstantinopleFix` here. I believe this should be fixed upstream because `ConstantinopleFix` imho is visually too close to `Constantinople` and makes `Constantinople` less grepable.

eth/vm/forks/petersburg/headers.py
cburgdorf: For some reason, this fails the tests with:

>TypeError: compute_difficulty() got multiple values for argument 'bomb_delay' ([CI job](https://circleci.com/gh/ethereum/py-evm/126329?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link))

Leaving it as is for now.

tox.ini
pipermerriam: cc @holgerd77 and @winsvega what do you two think about changing the name.

@cburgdorf you might surface this in the `ethereum/tests` repo as an issue

carver: I'd even prefer calling the old one `ConstantiNOPEl` and the new one `Constantinople` :smile: -- caps make it visually apparent, and it's greppable

cburgdorf: @carver I remember I proposed in the AllCoreDevs that we should consider renaming the failed fork to free up the name for the new fork but there were concerns that we can't do that since it is already live on various nets including xDai. Technically, we can do whatever we want in this code base but I guess deciding on a name in `ethereum/tests` opens up a broader discussion again ;)

winsvega:
Didn't we decide on a core dev call to name Constantinople and ConstantinopleFix ?

cburgdorf:
@winsvega Geth is more ambiguous than I thought. It uses both terms `ConstantinopleFix` and `Petersburg` https://github.com/holiman/go-ethereum/blob/bb75b656929e5688d2e6756461a51b7aecb66c9d/cmd/puppeth/genesis.go#L352-L354

The [announcement calls it Petersburg](https://github.com/ethereum/go-ethereum/releases/tag/v1.8.22) though.

tox.ini
veox: If the issue is the CLI arg being named as-upstream instead of as-codebase, could there be a "convenience" mapping of names for the `--fork` arg?..

holgerd77:
We are also using ``Petersburg`` since I read out of the conversations that this would be the settled name. For the moment we do some not to beautiful name mapping from ``constantinopleFix`` -> ``petersburg``, so I would be supportive on renaming.

tox.ini
cburgdorf: @veox we could do that but it just moves the `ConstantinopleFix` to another place. With geth, EthereumJS, Py-EVM all using `Petersburg` as a name, maybe an upstream change would be appropriate.

pipermerriam:
@winsvega I'm strongly in favor of renaming to `Petersburg`.  I don't have a full grasp of the *cost* but with advent of `ethereum/tests` publishing releases (and release notes) I think we have a framework in place to let people know this change happened and I suspect every client has some form of their own internal mapping from the `ethereum/tests` names to their own internal names.

Would explicit approval from Geth and Parity teams be sufficient for you to be comfortable making this change?  cc @karalabe and @5chdn

karalabe:
We just use ConstantinopleFix as a log label. I'd gladly drop it and use
Petersburg evertwhere. Our chain config API's use Petersburg in the genesis
spec, do that's our current "official" stance.

pipermerriam:
cc @folsen

holiman:
Yeah I used `ConstantinopleFix` internally, since it's quite obvious for us what it means. I considered it totally useless for non-developers though, since it's highly confusing unless you know the full backstory. Hence put 'Petersburg' in the genesis spec.  

> Didn't we decide on a core dev call to name Constantinople and ConstantinopleFix ?

No, I'm 99% certain we never did any such thing (I would have strongly objected). I think we said "let's decide over allcoredev gitter channel". 

I don't really care what we call it, but I think ConstantinopleFix is a horrible 'official' name, and in the name fo consistency it might be better to use 'Petersburg" everywhere.

PR #1722, opened 2019-02-12
Add get_transaction_receipt method to Chain API

Bhargavasomu:
### What was wrong?
Currently, if we want to implement the `getTransactionReceipt` JSON RPC functionality, then we would have to go all the way to the `ChainDB` part of the `Chain API`. This PR aims to make this functionality available as part of the `Chain API`, so that the convention of JSON RPC implementations using only the Chain API and not the `chaindb` part of the `Chain`, is followed.

### How was it fixed?
The `get_receipt_by_index` of `chaindb` was used in the `Chain API`, so as to abstract the usage.

#### Cute Animal Picture

### DIFF
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -126,6 +126,12 @@ class BaseChainDB(BaseHeaderDB):
         raise NotImplementedError("ChainDB classes must implement this method")
 
     @abstractmethod
+    def get_receipt_by_index(self,
+                             block_number: BlockNumber,
+                             receipt_index: int) -> Receipt:
+        raise NotImplementedError("ChainDB classes must implement this method")
+
+    @abstractmethod
     def get_receipts(self,
                      header: BlockHeader,
                      receipt_class: Type[Receipt]) -> Iterable[Receipt]:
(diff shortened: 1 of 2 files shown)


Bhargavasomu:
@cburgdorf could you please review this? This is a part of the PR https://github.com/ethereum/trinity/pull/122

eth/db/chain.py
```diff
@@ -143,6 +143,12 @@ def get_transaction_by_index(
     def get_transaction_index(self, transaction_hash: Hash32) -> Tuple[BlockNumber, int]:
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
```
cburgdorf: nitpick: would be nice to have this directly next to `get_receipts`.

Bhargavasomu: Done :+1:

Bhargavasomu:
@pipermerriam review please. Thankyou.

cburgdorf:
@Bhargavasomu went ahead and merged it in preparation of a new release being cut within the next few days.

PR #1723, opened 2019-02-18
Change inheritance order for chain_class_without_seal_validation

Bhargavasomu:
### What was wrong?
Previously the order in which `chain_class_without_seal_validation` would cause `MRO` errors when called on an instance of `NoChainSealValidationMixin` class. (Because `chain_class` is superset of `NoChainSealValidationMixin` class)

### How was it fixed?
The order was changed so that the superset class is added in the front. This prevents the `MRO` errors.

#### Cute Animal Picture

### DIFF
--- a/eth/tools/builder/chain/builders.py
+++ b/eth/tools/builder/chain/builders.py
@@ -290,11 +290,15 @@ class NoVMSealValidationMixin:
 @to_tuple
 def _mix_in_disable_seal_validation(vm_configuration: VMConfiguration) -> Iterable[VMFork]:
     for fork_block, vm_class in vm_configuration:
-        vm_class_without_seal_validation = type(
-            vm_class.__name__,
-            (NoVMSealValidationMixin, vm_class),
-            {},
-        )
+        if issubclass(vm_class, NoVMSealValidationMixin):
+            # Seal validation already disabled, hence nothing to change
+            vm_class_without_seal_validation = vm_class
+        else:
+            vm_class_without_seal_validation = type(
+                vm_class.__name__,
+                (NoVMSealValidationMixin, vm_class),
+                {},
+            )
         yield fork_block, vm_class_without_seal_validation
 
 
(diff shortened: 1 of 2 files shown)


Bhargavasomu:
@pipermerriam could you please review this? This came up when I was trying to reset the `Genesis` multiple times using `evm_resetToGenesisFixture`.

pipermerriam:
Is there a test we could add that would have exposed this under the chain builder test suite?

eth/tools/builder/chain/builders.py
```diff
@@ -290,11 +290,18 @@ def validate_seal(cls, header: BlockHeader) -> None:
...
+        if issubclass(vm_class, NoVMSealValidationMixin):
+            vm_class_without_seal_validation = type(
+                vm_class.__name__,
+                (vm_class,),
+                {},
+            )
+        else:
+            vm_class_without_seal_validation = type(
+                vm_class.__name__,
+                (NoVMSealValidationMixin, vm_class),
```
Bhargavasomu: When I changed this to `(vm_class, NoVMSealValidationMixin)`, it resulted in a bunch of errors, which I had no idea of how to resolve (some `hash mismatch errors`). Hence had to resent to this method.

eth/tools/builder/chain/builders.py
```diff
@@ -290,11 +290,18 @@ def validate_seal(cls, header: BlockHeader) -> None:
 @to_tuple
 def _mix_in_disable_seal_validation(vm_configuration: VMConfiguration) -> Iterable[VMFork]:
     for fork_block, vm_class in vm_configuration:
-        vm_class_without_seal_validation = type(
-            vm_class.__name__,
-            (NoVMSealValidationMixin, vm_class),
-            {},
-        )
+        if issubclass(vm_class, NoVMSealValidationMixin):
+            vm_class_without_seal_validation = type(
```
pipermerriam: How about we just return `vm_class` unmodified since this is effectively just adding another layer to the inheritance chain which doesn't actually add or modify any functionality.

eth/tools/builder/chain/builders.py
```diff
@@ -312,11 +319,18 @@ def disable_pow_check(chain_class: Type[BaseChain]) -> Type[BaseChain]:
...
         raise ValidationError("Chain class has no vm_configuration")
 
-    chain_class_without_seal_validation = type(
-        chain_class.__name__,
-        (NoChainSealValidationMixin, chain_class),
-        {},
-    )
+    if issubclass(chain_class, NoChainSealValidationMixin):
+        chain_class_without_seal_validation = type(
+            chain_class.__name__,
```
pipermerriam: Same here, instead of doing the additional subclass, lets just return `chain_class` unmodified.

Bhargavasomu:
@pipermerriam ready to be merged.

PR #1727, opened 2019-02-25
Flake8 bugbear

pipermerriam:
### What was wrong?

There are some linting checks that we want like:

- no mutable values for default arguments
- `getattr(x, 'value)` since we can use `x.value` since the attribute string is a constant value
- see more at https://github.com/PyCQA/flake8-bugbear

### How was it fixed?

Added `flake8-bugbear` to the `[lint]` installs.

#### Cute Animal Picture

71da8415374d3a4c766e32e40a7b1f33


pipermerriam:
CI failure fixed here: https://github.com/ethereum/py-evm/pull/1728

I suspect that may show up in other repos soon too.

jacqueswww:
@pipermerriam niiice! ;)

PR #1728, opened 2019-02-25
Remove pandoc and setuptools markdown requirement

pipermerriam:
### What was wrong?

A while back we started using `setuptools-markdown` to correctly render markdown files in pypi.  

We just started getting CI failures that seem related to this requirement: https://circleci.com/gh/ethereum/py-evm/126552?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Also, the requirement requires the user to have a system requirement installed (`pandoc`)

Now pypi supports markdown directly so we can remove it 

### How was it fixed?

Emulated this project: https://pypi.org/project/markdown-description-example/

#### Cute Animal Picture

https://pypi.org/project/markdown-description-example/

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -58,9 +58,11 @@ deps = {
         "bumpversion>=0.5.3,<1",
         "wheel",
         "setuptools>=36.2.0",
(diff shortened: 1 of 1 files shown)


pipermerriam:
If this goes in before releasing tomorrow, whoever makes that release should double check that everything still renders correctly on pypi.

PR #1729, opened 2019-03-04
Add pysha as dependency for docs

cburgdorf:
### What was wrong?

Our live API docs are broken because `pysha` is needed for autodoc to be able to build and extract the docs.

### How was it fixed?

Added pysha as a `docs` dependency.

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,8 @@ deps = {
     ],
     'doc': [
         "py-evm>=0.2.0-alpha.14",
+        # We need to have pysha for autodoc to be able to extract API docs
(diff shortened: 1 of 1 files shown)


cburgdorf:
I'm going to merge this without review as soon as it goes through CI because our live docs are broken without it. If anyone gets upset by this fix feel free to propose a different one or keep silent :stuck_out_tongue_closed_eyes:

pipermerriam:
Is there a way we could have caught this in circle-ci?

cburgdorf:
I was wondering the same but didn't follow up. Damn multi tasking ;)

PR #1730, opened 2019-03-06
remove superfluous raise Halt from selfdestruct

voith:
### What was wrong?

`selfdestruct` calls `_selfdestruct` which already does `raise Halt('SELFDESTRUCT')`.
The `raise Halt('SELFDESTRUCT')` in `selfdestruct` is unreachable as a result.

### How was it fixed?
Removed the superfluous `raise Halt` from `selfdestruct`.
`selfdestruct_eip150` and `selfdestruct_eip161` use `_selfdestruct`, hence removed the superfluos code from `selfdestruct` and not  from `_selfdestruct`.

#### Cute Animal Picture

### DIFF
--- a/eth/vm/logic/system.py
+++ b/eth/vm/logic/system.py
@@ -48,7 +48,6 @@ def revert(computation: BaseComputation) -> None:
 def selfdestruct(computation: BaseComputation) -> None:
     beneficiary = force_bytes_to_address(computation.stack_pop(type_hint=constants.BYTES))
     _selfdestruct(computation, beneficiary)
-    raise Halt('SELFDESTRUCT')
 
 
 def selfdestruct_eip150(computation: BaseComputation) -> None:


eth/vm/logic/system.py
```diff
@@ -48,7 +48,6 @@ def revert(computation: BaseComputation) -> None:
 def selfdestruct(computation: BaseComputation) -> None:
     beneficiary = force_bytes_to_address(computation.stack_pop(type_hint=constants.BYTES))
     _selfdestruct(computation, beneficiary)
-    raise Halt('SELFDESTRUCT')
```
pipermerriam: It might be good to replace this with a `raise Exception("UNREACHABLE CODE PATH")` so that if the behavior of `_selfdestruct` were to ever change this would not break.

voith: I think the tests should tell us if some behavior changed. For me, an exception is to be raised when something wrong happens. To check expected behavior, we use tests, just my thoughts.

Also, there'll be so many other places where we would want to add `raise Exception("UNREACHABLE CODE PATH")`, if that's the pattern we decide to follow. I rather keep the code clean and keep only what is needed.

PR #1731, opened 2019-03-07
fix ecpairing for py_ecc upgrade

ChihChengLiang:
### What was wrong?

Fix py-evm to adapt for upgrades in py_ecc 

### How was it fixed?

import all field elements from `py_ecc.fields`

#### Cute Animal Picture


ChihChengLiang:
Fixed in https://github.com/ethereum/py_ecc/pull/63

PR #1732, opened 2019-03-11
Bugfix ethash cache

carver:
### What was wrong?

After release and trinity upgrade, will fix https://github.com/ethereum/trinity/issues/340

If you generate the proof of work cache in multiple threads, the cache seeds can get created out of order (or duplicated).

### How was it fixed?

Switch the cache from a list to a dict that references the epoch by index. Then if a new cache is added twice, or out of order, it's not a problem because it always gets specifically inserted to the expected index.

This *could* be solved without switching to a dict, but this seemed easy to implement and reason about. It doesn't come with any meaningful performance hit that I'm aware of.

#### Cute Animal Picture


eth/consensus/pow.py
```diff
@@ -30,23 +31,33 @@
...
 
 
 def get_cache(block_number: int) -> bytes:
-    while len(cache_seeds) <= block_number // EPOCH_LENGTH:
-        cache_seeds.append(keccak(cache_seeds[-1]))
-    seed = cache_seeds[block_number // EPOCH_LENGTH]
+    # Generate the seed for the cache if it's missing, aim for thread-safety
+    epoch_index = block_number // EPOCH_LENGTH
+    while epoch_index not in cache_seeds:
+        nearest_index = min(epoch_index - 1, max(cache_seeds.keys()))
```
carver: This `min` check seems totally unnecessary, I think `nearest_index = max(cache_seeds.keys())` is sufficient.

eth/consensus/pow.py
```diff
@@ -30,23 +31,33 @@
 
 
 # Type annotation here is to ensure we don't accidentally use strings instead of bytes.
-cache_seeds = [b'\x00' * 32]  # type: List[bytes]
+cache_seeds = {0: b'\x00' * 32}  # type: Dict[int, bytes]
```
pipermerriam: Probably not worth it unless it turns out to be easy, but it appears that `cache_by_seeds` naturally regulates it's maximum size.  Is there a way we can use a `weakref.WeakValueDict` to have this automatically clean itself out based on what is stored in the `cache_by_seed` data structure.

Given these are just 32 byte values and epochs are long, not a big deal at all.

carver: Hm, not sure where the strong reference would come from. Also, in case we want to go verify an old seed, we might trigger a regen of seeds from 0. It opens a lot of questions, and I don't *think* the upside is big enough to justify it.

eth/consensus/pow.py
```diff
@@ -30,23 +31,33 @@
...
-        cache_seeds.append(keccak(cache_seeds[-1]))
-    seed = cache_seeds[block_number // EPOCH_LENGTH]
+    # Generate the seed for the cache if it's missing, aim for thread-safety
+    epoch_index = block_number // EPOCH_LENGTH
+    while epoch_index not in cache_seeds:
+        highest_index = max(cache_seeds.keys())
+        cache_seeds[highest_index + 1] = keccak(cache_seeds[highest_index])
+    seed = cache_seeds[epoch_index]
+
+    # Get the cache if already generated, marking it as recently used
```
pipermerriam: Can we drop the `cache_by_seed` data structure (and the need to manage it's size) by just decorating `get_cache` with `@functools.lru_cache(max_size=10)`?  I feel like I must be missing something that prevents that.

eth/consensus/pow.py
```diff
@@ -62,8 +73,19 @@ def check_pow(block_number: int,
     mining_output = hashimoto_light(
         block_number, cache, mining_hash, big_endian_to_int(nonce))
     if mining_output[b'mix digest'] != mix_hash:
-        raise ValidationError("mix hash mismatch; {0} != {1}".format(
-            encode_hex(mining_output[b'mix digest']), encode_hex(mix_hash)))
+        raise ValidationError(
+            "mix hash mismatch; expected: {} != actual: {}. "
```
pipermerriam: f-string?

tests/core/fixtures.py
```diff
@@ -29,3 +29,48 @@
...
+# blocks: 1, 30001, 60001, etc
+ropsten_epoch_headers = [
+    '0xf90210a041941023680923e0fe4d74a34bdac8141f2540e3ae90623718e47d66d1ca4a2da01dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d4934794d1aeb42885a43b72b518182ef893125814811048a0c7b01007a10da045eacb90385887dd0c38fcb5db7393006bdde24b93873c334ba056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421a056e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421b9010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000830f3a000183ffc001808458318da299d883010503846765746887676f312e372e318664617277696ea00f98b15f1a4901a7e9204f3c500a7bd527b3fb2c3340e12176a44b83e414a69e880ece08ea8c49dfd9',  # noqa: E501
```
pipermerriam: Plopping these in a flat text file with one per line is actually quite easy to load > decode.  That reduces these large blocks of hex in python files (at the cost of obscuring them in some `tests/core/ropsten_epoch_headers.rlp` file).  easy then to wrap them up in a fixture for easy use.

Don't feel very strongly on the subject, mostly just getting bothered here and there by large hex blobs in our python files but unsure if the situation can be improved.

eth/consensus/pow.py
carver: py-evm is still supporting py3.5 unfortunately, afaik

eth/consensus/pow.py
carver: We want to reuse one cache entry for all blocks in the same epoch.

Maybe you mean that we could split this into two functions, so that `get_cache(block_number)` calls `get_cached_pow_cache(seed)`, then `get_cached_pow_cache()` could have a simple lru cache.

carver: Actually, I'm super glad you asked because *most* of this code is unnecessary. We're not even using the seed for anything! I went back in history to see why, and don't see a good reason: https://github.com/ethereum/py-evm/commit/78ab70915e8cdd500ba8dffd710a7465ff1e8bd9

Looks like it was just a bug. This is about to become two 2-line functions :man_facepalming:

carver: Well, I made it a lot simpler, like just:

```py
def get_cache(block_number: int) -> bytes:                                                                             
    epoch_index = block_number // EPOCH_LENGTH                                                                         
    return _get_cached_pow_cache(epoch_index)                                                                          
                                                                                                                       
                                                                                                                       
@functools.lru_cache(maxsize=CACHE_MAX_ITEMS)                                                                          
def _get_cached_pow_cache(epoch_index: int) -> bytes:                                                                  
    # simulate requesting mkcache by block number: multiply index by epoch length                                      
    return mkcache_bytes(epoch_index * EPOCH_LENGTH)  
```

That caused the tests to run 70% slower, which was a bit of a surprise to me. We didn't have to do *any* keccak hashing. But I guess the lru_cache overhead is higher than I thought and/or is especially ineffective in a heavy threaded environment.

eth/consensus/pow.py
pipermerriam: I didn't see what codebase I was in :disappointed:

eth/consensus/pow.py
carver: Seemed like too high a perf penalty to pay. I did remove the pointless hashing, but kept the manual caching, plus a note explaining why.

PR #1735, opened 2019-03-28
Move storage lookups into new StorageDB

carver:
### What was wrong?

We are currently calculating storage roots at every storage change.

We need to request state from other nodes at the previous block boundary, even when running the query from in the middle of block execution.

It's also a waste of disk space, and a waste of cycles on `keccak` hashing for the trie reorgs.

### How was it fixed?

- Moved the storage logic out of `AccountDB` into new `StorageDB`
- Set up a similar pipeline, with a special database adapter that looks up storage roots on demand
- The storage db now keeps a journal of changes in their decoded state (slot number and int value) rather than keeping a journal of trie changes.
- Moved the storage API used by VM execution out to `state.set_storage(...)` instead of `state.account_db.set_storage(...)`, which seems like an improvement. Going forward, it lets us tweak the Storage DB API while keeping the `VMState` API the same.

TODO:
- [x] docstrings
- [x] squash commits
- [x] split out any prerequisite changes and rebase on top after merge, like
  - [x] #1716 - made some updates, to avoid persisting state before the miner reward
  - [x] #1737 
  - [x] #1738
  - [x] #1739
  - [x] #1740
  - [x] #1741
  - [x] #1745
  - [x] #1746
  - [x] #1749
  - [x] #1755
  - [x] #1756
  - on request: `Account.__repr__` 8918923990b61c70edcf2b740b757557e3cb4aca
  - [x] ~~probably more to go~~
- [x] pass all state tests
- [x] type annotations
- [x] lint
- [x] also batch writes in accountdb ~~(probably separate PR)~~ (but it's in a separate commit if reviewers want to split it)

#### Cute Animal Picture

### DIFF
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -390,19 +390,19 @@ class AccountDB(BaseAccountDB):
     #
     # Record and discard API
     #
-    def record(self) -> Tuple[UUID, UUID]:
-        return (self._journaldb.record(), self._journaltrie.record())
+    def record(self) -> UUID:
+        changeset_id = self._journaldb.record()
+        self._journaltrie.record(changeset_id)
+        return changeset_id
 
-    def discard(self, changeset: Tuple[UUID, UUID]) -> None:
-        db_changeset, trie_changeset = changeset
-        self._journaldb.discard(db_changeset)
-        self._journaltrie.discard(trie_changeset)
+    def discard(self, changeset: UUID) -> None:
+        self._journaldb.discard(changeset)
+        self._journaltrie.discard(changeset)
         self._account_cache.clear()
 
-    def commit(self, changeset: Tuple[UUID, UUID]) -> None:
-        db_changeset, trie_changeset = changeset
-        self._journaldb.commit(db_changeset)
-        self._journaltrie.commit(trie_changeset)
+    def commit(self, changeset: UUID) -> None:
+        self._journaldb.commit(changeset)
+        self._journaltrie.commit(changeset)
 
     def make_state_root(self) -> Hash32:
         self.logger.debug2("Generating AccountDB trie")
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -5,23 +5,27 @@ from abc import (
 from uuid import UUID
 import logging
 from lru import LRU
-from typing import cast, Set, Tuple  # noqa: F401
+from typing import (  # noqa: F401
+    cast,
+    Dict,
+    Iterable,
+    Set,
+    Tuple,
+)
 
+from eth_hash.auto import keccak
 from eth_typing import (
     Address,
     Hash32
 )
-
-import rlp
-
-from trie import (
-    HexaryTrie,
-)
-
-from eth_hash.auto import keccak
 from eth_utils import (
     encode_hex,
-    int_to_big_endian,
+    to_tuple,
+    ValidationError,
+)
+import rlp
+from trie import (
+    HexaryTrie,
 )
 
 from eth.constants import (
@@ -30,7 +34,6 @@ from eth.constants import (
 )
 from eth.db.backends.base import (
     BaseAtomicDB,
-    BaseDB,
 )
 from eth.db.batch import (
     BatchDB,
@@ -40,6 +43,9 @@ from eth.db.cache import (
 )
 from eth.db.journal import (
     JournalDB,
+)
+from eth.db.storage import (
+    AccountStorageDB,
 )
 from eth.rlp.accounts import (
     Account,
@@ -51,9 +57,6 @@ from eth.validation import (
 )
 from eth.tools.logging import (
     ExtendedDebugLogger
-)
-from eth._utils.padding import (
-    pad32,
 )
 
 from .hash_trie import HashTrie
@@ -80,11 +83,22 @@ class BaseAccountDB(ABC):
     # Storage
     #
     @abstractmethod
-    def get_storage(self, address: Address, slot: int) -> int:
+    def get_storage(self, address: Address, slot: int, from_journal: bool=True) -> int:
         raise NotImplementedError("Must be implemented by subclasses")
 
     @abstractmethod
     def set_storage(self, address: Address, slot: int, value: int) -> None:
+        raise NotImplementedError("Must be implemented by subclasses")
+
+    @abstractmethod
+    def delete_storage(self, address: Address) -> None:
+        """
+        Delete *all* storage values in this account. This action is journaled, like set() actions.
+
+        Unlike set(), deleting storage will not cause :meth:`make_storage_roots` to emit a new
+        storage root (and therefore will not persist deletes to the base database).
+        The account's storage root must be explicitly set to the empty root.
+        """
         raise NotImplementedError("Must be implemented by subclasses")
 
     #
@@ -139,21 +153,29 @@ class BaseAccountDB(ABC):
     # Record and discard API
     #
     @abstractmethod
-    def record(self) -> Tuple[UUID, UUID]:
+    def record(self) -> UUID:
         raise NotImplementedError("Must be implemented by subclass")
 
     @abstractmethod
-    def discard(self, changeset: Tuple[UUID, UUID]) -> None:
+    def discard(self, changeset: UUID) -> None:
         raise NotImplementedError("Must be implemented by subclass")
 
     @abstractmethod
-    def commit(self, changeset: Tuple[UUID, UUID]) -> None:
+    def commit(self, changeset: UUID) -> None:
         raise NotImplementedError("Must be implemented by subclass")
 
     @abstractmethod
     def make_state_root(self) -> Hash32:
         """
         Generate the state root with all the current changes in AccountDB
+
+        Current changes include every pending change to storage, as well as all account changes.
+        After generating all the required tries, the final account state root is returned.
+
+        This is an expensive operation, so should be called as little as possible. For example,
+        pre-Byzantium, this is called after every transaction, because we need the state root
+        in each receipt. Byzantium+, we only need state roots at the end of the block,
+        so we *only* call it right before persistance.
 
         :return: the new state root
         """
@@ -164,6 +186,9 @@ class BaseAccountDB(ABC):
         """
         Send changes to underlying database, including the trie state
         so that it will forever be possible to read the trie from this checkpoint.
+
+        :meth:`make_state_root` must be explicitly called before this method.
+        Otherwise persist will raise a ValidationError.
         """
         raise NotImplementedError("Must be implemented by subclass")
 
@@ -179,8 +204,6 @@ class AccountDB(BaseAccountDB):
 
         .. code::
 
-                                                                    -> hash-trie -> storage lookups
-                                                                  /
             db > _batchdb ---------------------------> _journaldb ----------------> code lookups
              \
               -> _batchtrie -> _trie -> _trie_cache -> _journaltrie --------------> account lookups
@@ -207,7 +230,7 @@ class AccountDB(BaseAccountDB):
         rather than the nodes stored by the trie). This enables
         a squashing of all account changes before pushing them into the trie.
 
-        .. NOTE:: There is an opportunity to do something similar for storage
+        .. NOTE:: StorageDB works similarly
 
         AccountDB synchronizes the snapshot/revert/persist of both of the
         journals.
@@ -220,6 +243,8 @@ class AccountDB(BaseAccountDB):
         self._trie_cache = CacheDB(self._trie)
         self._journaltrie = JournalDB(self._trie_cache)
         self._account_cache = LRU(2048)
+        self._account_stores = {}  # type: Dict[Address, AccountStorageDB]
+        self._dirty_accounts = set()  # type: Set[Address]
 
     @property
     def state_root(self) -> Hash32:
@@ -227,8 +252,9 @@ class AccountDB(BaseAccountDB):
 
     @state_root.setter
     def state_root(self, value: Hash32) -> None:
-        self._trie_cache.reset_cache()
-        self._trie.root_hash = value
+        if self._trie.root_hash != value:
+            self._trie_cache.reset_cache()
+            self._trie.root_hash = value
 
     def has_root(self, state_root: bytes) -> bool:
         return state_root in self._batchtrie
@@ -240,40 +266,74 @@ class AccountDB(BaseAccountDB):
         validate_canonical_address(address, title="Storage Address")
         validate_uint256(slot, title="Storage Slot")
 
-        account = self._get_account(address, from_journal)
-        storage = HashTrie(HexaryTrie(self._journaldb, account.storage_root))
-
-        slot_as_key = pad32(int_to_big_endian(slot))
-
-        if slot_as_key in storage:
-            encoded_value = storage[slot_as_key]
-            return rlp.decode(encoded_value, sedes=rlp.sedes.big_endian_int)
-        else:
-            return 0
+        account_store = self._get_address_store(address)
+        return account_store.get(slot, from_journal)
 
     def set_storage(self, address: Address, slot: int, value: int) -> None:
         validate_uint256(value, title="Storage Value")
         validate_uint256(slot, title="Storage Slot")
         validate_canonical_address(address, title="Storage Address")
 
-        account = self._get_account(address)
-        storage = HashTrie(HexaryTrie(self._journaldb, account.storage_root))
-
-        slot_as_key = pad32(int_to_big_endian(slot))
-
-        if value:
-            encoded_value = rlp.encode(value)
-            storage[slot_as_key] = encoded_value
-        else:
-            del storage[slot_as_key]
-
-        self._set_account(address, account.copy(storage_root=storage.root_hash))
+        account_store = self._get_address_store(address)
+        self._dirty_accounts.add(address)
+        account_store.set(slot, value)
 
     def delete_storage(self, address: Address) -> None:
         validate_canonical_address(address, title="Storage Address")
 
+        self._set_storage_root(address, BLANK_ROOT_HASH)
+        self._wipe_storage(address)
+
+    def _wipe_storage(self, address: Address) -> None:
+        """
+        Wipe out the storage, without explicitly handling the storage root update
+        """
+        account_store = self._get_address_store(address)
+        self._dirty_accounts.add(address)
+        account_store.delete()
+
+    def _get_address_store(self, address: Address) -> AccountStorageDB:
+        if address in self._account_stores:
+            store = self._account_stores[address]
+        else:
+            storage_root = self._get_storage_root(address)
+            store = AccountStorageDB(self._raw_store_db, storage_root, address)
+            self._account_stores[address] = store
+        return store
+
+    def _dirty_account_stores(self) -> Iterable[Tuple[Address, AccountStorageDB]]:
+        for address in self._dirty_accounts:
+            store = self._account_stores[address]
+            yield address, store
+
+    @to_tuple
+    def _get_changed_roots(self) -> Iterable[Tuple[Address, Hash32]]:
+        # list all the accounts that were changed, and their new storage roots
+        for address, store in self._dirty_account_stores():
+            if store.has_changed_root:
+                yield address, store.get_changed_root()
+
+    def _get_storage_root(self, address: Address) -> Hash32:
         account = self._get_account(address)
-        self._set_account(address, account.copy(storage_root=BLANK_ROOT_HASH))
+        return account.storage_root
+
+    def _set_storage_root(self, address: Address, new_storage_root: Hash32) -> None:
+        account = self._get_account(address)
+        self._set_account(address, account.copy(storage_root=new_storage_root))
+
+    def _validate_flushed_storage(self, address: Address, store: AccountStorageDB) -> None:
+        if store.has_changed_root:
+            actual_storage_root = self._get_storage_root(address)
+            expected_storage_root = store.get_changed_root()
+            if expected_storage_root != actual_storage_root:
+                raise ValidationError(
+                    "Storage root was not saved to account before trying to persist roots. "
+                    "Account %r had storage %r, but should be %r." % (
+                        address,
+                        actual_storage_root,
+                        expected_storage_root,
+                    )
+                )
 
     #
     # Balance
@@ -352,9 +412,12 @@ class AccountDB(BaseAccountDB):
 
     def delete_account(self, address: Address) -> None:
         validate_canonical_address(address, title="Storage Address")
+
         if address in self._account_cache:
             del self._account_cache[address]
         del self._journaltrie[address]
+
+        self._wipe_storage(address)
 
     def account_exists(self, address: Address) -> bool:
         validate_canonical_address(address, title="Storage Address")
@@ -395,28 +458,80 @@ class AccountDB(BaseAccountDB):
     def record(self) -> UUID:
         changeset_id = self._journaldb.record()
         self._journaltrie.record(changeset_id)
+
+        for _, store in self._dirty_account_stores():
+            store.record(changeset_id)
         return changeset_id
 
     def discard(self, changeset: UUID) -> None:
         self._journaldb.discard(changeset)
         self._journaltrie.discard(changeset)
         self._account_cache.clear()
+        for _, store in self._dirty_account_stores():
+            store.discard(changeset)
 
     def commit(self, changeset: UUID) -> None:
         self._journaldb.commit(changeset)
         self._journaltrie.commit(changeset)
+        for _, store in self._dirty_account_stores():
+            store.commit(changeset)
 
     def make_state_root(self) -> Hash32:
-        self.logger.debug2("Generating AccountDB trie")
+        for _, store in self._dirty_account_stores():
+            store.make_storage_root()
+
+        for address, storage_root in self._get_changed_roots():
+            self.logger.debug2(
+                "Updating account 0x%s to storage root 0x%s",
+                address.hex(),
+                storage_root.hex(),
+            )
+            self._set_storage_root(address, storage_root)
+
         self._journaldb.persist()
         self._journaltrie.persist()
         return self.state_root
 
     def persist(self) -> None:
         self.make_state_root()
+
+        # persist storage
+        with self._raw_store_db.atomic_batch() as write_batch:
+            for address, store in self._dirty_account_stores():
+                self._validate_flushed_storage(address, store)
+                store.persist(write_batch)
+
+        for address, new_root in self._get_changed_roots():
+            if new_root not in self._raw_store_db and new_root != BLANK_ROOT_HASH:
+                raise ValidationError(
+                    "After persisting storage trie, a root node was not found. "
+                    "State root for account 0x%s is missing for hash 0x%s." % (
+                        address.hex(),
+                        new_root.hex(),
+                    )
+                )
+
+        # reset local storage trackers
+        self._account_stores = {}
+        self._dirty_accounts = set()
+
+        # persist accounts
+        self._validate_generated_root()
         with self._raw_store_db.atomic_batch() as write_batch:
             self._batchtrie.commit_to(write_batch, apply_deletes=False)
             self._batchdb.commit_to(write_batch, apply_deletes=False)
+
+    def _validate_generated_root(self) -> None:
+        db_diff = self._journaldb.diff()
+        if len(db_diff):
+            raise ValidationError(
+                "AccountDB had a dirty db when it needed to be clean: %r" % db_diff
+            )
+        trie_diff = self._journaltrie.diff()
+        if len(trie_diff):
+            raise ValidationError(
+                "AccountDB had a dirty trie when it needed to be clean: %r" % trie_diff
+            )
 
     def _log_pending_accounts(self) -> None:
         accounts_displayed = set()  # type: Set[bytes]
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -29,6 +29,7 @@ from eth.constants import (
     EMPTY_SHA3,
 )
 from eth.db.backends.base import (
+    BaseAtomicDB,
     BaseDB,
 )
 from eth.db.batch import (
@@ -171,7 +172,7 @@ class AccountDB(BaseAccountDB):
 
     logger = cast(ExtendedDebugLogger, logging.getLogger('eth.db.account.AccountDB'))
 
-    def __init__(self, db: BaseDB, state_root: Hash32=BLANK_ROOT_HASH) -> None:
+    def __init__(self, db: BaseAtomicDB, state_root: Hash32=BLANK_ROOT_HASH) -> None:
         r"""
         Internal implementation details (subject to rapid change):
         Database entries go through several pipes, like so...
@@ -211,6 +212,7 @@ class AccountDB(BaseAccountDB):
         AccountDB synchronizes the snapshot/revert/persist of both of the
         journals.
         """
+        self._raw_store_db = db
         self._batchdb = BatchDB(db)
         self._batchtrie = BatchDB(db)
         self._journaldb = JournalDB(self._batchdb)
@@ -412,8 +414,9 @@ class AccountDB(BaseAccountDB):
 
     def persist(self) -> None:
         self.make_state_root()
-        self._batchtrie.commit(apply_deletes=False)
-        self._batchdb.commit(apply_deletes=True)
+        with self._raw_store_db.atomic_batch() as write_batch:
+            self._batchtrie.commit_to(write_batch, apply_deletes=False)
+            self._batchdb.commit_to(write_batch, apply_deletes=False)
 
     def _log_pending_accounts(self) -> None:
         accounts_displayed = set()  # type: Set[bytes]
--- a/eth/db/journal.py
+++ b/eth/db/journal.py
@@ -384,6 +384,8 @@ class JournalDB(BaseDB):
 
         If this is the base changeset then all changes will be written to
         the underlying database and the Journal starts a new recording.
+        Typically, callers won't have access to the base changeset, because
+        it is dropped during .reset() which is called in JournalDB().
         """
         self._validate_changeset(changeset_id)
         journal_data = self.journal.commit_changeset(changeset_id)
--- a/eth/db/storage.py
+++ b/eth/db/storage.py
@@ -1,0 +1,264 @@
+import logging
+from typing import (  # noqa: F401
+    cast,
+    Dict,
+    Iterable,
+    Set,
+    Tuple,
+)
+from uuid import UUID
+
+from eth_hash.auto import keccak
+from eth_typing import (
+    Address,
+    Hash32
+)
+from eth_utils import (
+    ValidationError,
+    int_to_big_endian,
+)
+import rlp
+from trie import (
+    HexaryTrie,
+    exceptions as trie_exceptions,
+)
+
+from eth._utils.padding import (
+    pad32,
+)
+from eth.db.backends.base import (
+    BaseAtomicDB,
+    BaseDB,
+)
+from eth.db.batch import (
+    BatchDB,
+)
+from eth.db.cache import (
+    CacheDB,
+)
+from eth.db.journal import (
+    JournalDB,
+)
+from eth.tools.logging import (
+    ExtendedDebugLogger
+)
+
+
+class StorageLookup(BaseDB):
+    """
+    This lookup converts lookups of storage slot integers into the appropriate trie lookup.
+    Similarly, it persists changes to the appropriate trie at write time.
+
+    StorageLookup also tracks the state roots changed since the last persist.
+    """
+    logger = cast(ExtendedDebugLogger, logging.getLogger("eth.db.storage.StorageLookup"))
+
+    def __init__(self, db: BaseDB, storage_root: Hash32, address: Address) -> None:
+        self._db = db
+        self._starting_root_hash = storage_root
+        self._address = address
+        self._write_trie = None
+        self._trie_nodes_batch = None  # type: BatchDB
+
+    def _get_write_trie(self) -> HexaryTrie:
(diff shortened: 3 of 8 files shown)


carver:
Neat, the only state failure left is an xfail in `INCORRECT_UPSTREAM_TESTS`, time to go decide if it's really incorrect upstream...

eth/db/storage.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def commit(self, changeset: UUID) -> None:
+        self.logger.debug2('commit checkpoint %r', changeset)
+        self._journal_storage.commit(changeset)
+
+    def make_storage_roots(self) -> Tuple[Tuple[Address, Hash32], ...]:
+        """
+        Force calculation of all storage roots
+
+        MUST immediately update account storage roots for reads to be consistent
```
pipermerriam: Probably doesn't need to be done in this PR but this seems like a place for future optimization since we shouldn't actually need the storage root for each account to do EVM execution right?  Just an implementation detail.  In theory we should be able to use a simple mapping of `slot -> value` during execution and then write it to a trie once execution finishes, we'd just need a way to maintain that simplistic data view of account storage during the execution lifetime.

carver: Sorry, I know this is hard to follow without some docs. I'll write it up in the docstring before I remove the `[WIP]`.

The short answer is: that's what it's doing! :)  All the changes are in a slot/val uncommitted mapping until the storage roots are needed, when it's pushed into the trie.

This particular method is only called at the specific moments when we *do* need to calculate the state root (end-of-transaction for pre-Byzantium, or end-of-block afterward).

carver:
#1749 (BatchDB `read_through_deletes`) is the only one left I am planning on merging separately and rebasing. But it's only a few lines and should be pretty easy to ignore in review. If you think more needs to come out, let me know.

carver:
Ah, just realized I need add docstring comments to help with reviewing.  I'll remove draft status when the comments are added.

carver:
I had to do one more round of updates to the `JournalDB`. I just left them in, but will split them out if anyone feels it will make the review unmanageable.

lithp:
Quick update, it's taking me a while but I'm still looking at this. I'm trying to figure out all the moving pieces here, this is a part of the codebase I haven't looked at before. I'll have a real review for you tomorrow!

eth/db/journal.py
```diff
@@ -384,6 +421,8 @@ def diff(self) -> DBDiff:
                     continue
```
lithp: There's a comment above this line which I think is no longer correct: 

`# This is purely for performance, not correctness.`

That changesets are iterated in reverse means that `ERASE_CREATED_ENTRY` has a chance to set `visited_keys` and prevent the earlier operation which set this key from showing up in the diff.

eth/db/journal.py
```diff
@@ -247,7 +260,9 @@ def __getitem__(self, key: bytes) -> bytes:
 
         val = self.journal[key]
         if val is DELETED_ENTRY:
-            raise KeyError(key, "item was explicitly deleted in JournalDB")
+            raise KeyError(key, "item to be deleted from wrapped DB, by JournalDB")
```
lithp: nit: maybe something like "this item's deletion has been journaled" or "this item is pending deletion"

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+        storage_root = self._account_db.get_storage_root(self._address)
+        if storage_root == BLANK_ROOT_HASH:
+            return HexaryTrie({})
+        elif self._write_trie is not None:
+            return self._write_trie
+        else:
+            if self._starting_root_hash is None:
+                self._starting_root_hash = storage_root
+            # Creating "HexaryTrie" is a pretty light operation, so not a huge cost
+            # to create a new one at every ready, but we could
```
pipermerriam: s/ready/read

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+            # cache the read trie, if this becomes a bottleneck.
+            return HexaryTrie(self._db, root_hash=storage_root)
+
+    def _decode_key(self, key: bytes) -> Tuple[Address, bytes]:
+        padded_slot = pad32(key)
+        return keccak(padded_slot)
```
pipermerriam: return value and function signature don't seem to match.

eth/vm/computation.py
```diff
@@ -114,7 +114,7 @@ class BaseComputation(Configurable, ABC):
     _error = None  # type: VMError
 
     _log_entries = None  # type: List[Tuple[int, bytes, List[int], bytes]]
```
pipermerriam: if you get back in here, the above should be 

```
# type: List[Tuple[int, Address, List[int], bytes]]
```

eth/db/journal.py
```diff
@@ -330,24 +349,42 @@ def commit(self, changeset_id: uuid.UUID) -> None:
...
             self.reset()
 
+            for key, value in journal_data.items():
+                try:
+                    if value is DELETED_ENTRY:
+                        del self.wrapped_db[key]
+                    elif value is ERASE_CREATED_ENTRY:
+                        pass
+                    else:
+                        self.wrapped_db[key] = value  # type: ignore # value can only be bytes here
```
lithp: nit: instead of a `#type: ignore` I think a `cast(bytes, value)` would do the trick here.

lithp: alternatively, `assert isinstance(value, bytes)` also makes mypy happy, for a performance penalty.

eth/db/journal.py
```diff
@@ -330,24 +349,42 @@ def commit(self, changeset_id: uuid.UUID) -> None:
...
+            changeset_id: uuid.UUID,
+            journal_data: Dict[bytes, Union[bytes, DeletedEntry]]) -> None:
+        self.record(changeset_id)
+        for key, value in journal_data.items():
+            if value is DELETED_ENTRY:
+                del self.journal[key]
+            elif value is ERASE_CREATED_ENTRY:
+                self.journal.local_delete(key)
+            else:
+                self.journal[key] = value  # type: ignore # value can only be bytes here
```
lithp: the same nitpick applies here, `# type: ignore` isn't necessary.

eth/db/journal.py
carver: Yeah, I'll do it for now. I seem to remember `JournalDB` being a place that was fairly sensitive to performance, but it could probably use a lot of performance love anyway, so I won't try to optimize without benchmarks. :)

carver:
Send it back to WIP until the next couple of spin-off PRs are merged

eth/db/account.py
```diff
@@ -393,30 +359,42 @@ def _set_account(self, address: Address, account: Account) -> None:
...
-        return (self._journaldb.record(), self._journaltrie.record())
-
-    def discard(self, changeset: Tuple[UUID, UUID]) -> None:
-        db_changeset, trie_changeset = changeset
-        self._journaldb.discard(db_changeset)
-        self._journaltrie.discard(trie_changeset)
+    def record(self) -> UUID:
+        changeset_id = self._journaldb.record()
+        self._journaltrie.record(changeset_id)
+        return changeset_id
```
lithp: Yeah, this is clearly better :)

eth/db/account.py
```diff
@@ -393,30 +359,42 @@ def _set_account(self, address: Address, account: Account) -> None:
...
         return self.state_root
 
     def persist(self) -> None:
-        self.make_state_root()
-        self._batchtrie.commit(apply_deletes=False)
-        self._batchdb.commit(apply_deletes=True)
+        self._validate_generated_root()
+        with self._raw_store_db.atomic_batch() as write_batch:
+            self._batchtrie.commit_to(write_batch, apply_deletes=False)
+            self._batchdb.commit_to(write_batch, apply_deletes=False)
```
lithp: I know I'm picking on a lot of decisions which came before your PRs but this is my first time really interacting with these files :sweat_smile: I don't think `_batchtrie` and `_batchdb` need to be kept separate, since `commit_to` is always called at the same time changes to the trie and db are always tracked together.

eth/db/account.py
```diff
@@ -393,30 +359,42 @@ def _set_account(self, address: Address, account: Account) -> None:
...
+
+    def _validate_generated_root(self) -> None:
```
lithp: *nit*: I guess this means "validate that the root has been generated". I think it'd be slightly more clear if it was something like `_validate_root_is_generated()` or `_validate_journal_dbs_not_dirty()`

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+
+        .. code::
+
+            db -> _storage_lookup -> _storage_cache -> _journal_storage
+
+        db is the raw database, we can assume it hits disk when written to.
+        Keys are stored as node hashes and rlp-encoded node values.
+
+        _storage_lookup is itself a pair of databases: (BatchDB -> HexaryTrie),
+        writes to storage lookup *are* immeditaely applied to a trie, generating
```
lithp: `s/immeditaely/immediately/`

eth/vm/state.py
```diff
@@ -218,35 +254,42 @@ def account_is_empty(self, address: Address) -> bool:
...
-    def snapshot(self) -> Tuple[bytes, Tuple[UUID, UUID]]:
+    def snapshot(self) -> Tuple[Hash32, UUID, UUID]:
         """
         Perform a full snapshot of the current state.
 
         Snapshots are a combination of the :attr:`~state_root` at the time of the
         snapshot and the id of the changeset from the journaled DB.
         """
-        return (self.state_root, self._account_db.record())
+        return self.state_root, self._account_db.record(), self._storage_db.record()
```
lithp: The same trick of sharing snapshot ids might be possible here. No problem if you don't want to do it! But if you don't it might be worth opening a ticket for.

eth/db/account.py
```diff
@@ -225,53 +221,23 @@ def state_root(self) -> Hash32:
...
     # Storage
     #
-    def get_storage(self, address: Address, slot: int, from_journal: bool=True) -> int:
-        validate_canonical_address(address, title="Storage Address")
-        validate_uint256(slot, title="Storage Slot")
-
-        account = self._get_account(address, from_journal)
-        storage = HashTrie(HexaryTrie(self._journaldb, account.storage_root))
-
-        slot_as_key = pad32(int_to_big_endian(slot))
```
lithp: This line used to be succinct documentation of how int slots are turned into bytes keys. This PR scatters that knowledge across a couple methods in two different classes. What do you think of moving the `int_to_big_endian` part into `_decode_key` so that these can continue living side-by-side? It seems reasonable for `StorageLookup` to be the sole owner of how slots are read and written.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+        # list all the accounts that were changed, and their new storage roots
+        for address, store in self._written_account_stores():
+            if store.has_changed_root:
+                yield address, store.get_changed_root()
+
+    def persist(self) -> Tuple[Tuple[Address, Hash32], ...]:
+        self.logger.debug2('persisting storage roots')
+
+        # must generate changed_roots before persisting
+        changed_roots = self._get_changed_roots()
```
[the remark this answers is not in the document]
lithp: Looking at this some more, I was surprised that this didn't work!

```python
    storage_db.set(ADDRESS, 1, 1)
    storage_db.make_storage_roots()
    storage_db.persist()  # throws `ValidationError`
```

In retrospect, I now see that the error message was telling me to do this:

```python
    storage_db.set(ADDRESS, 1, 1)

    changed_roots = storage_db.make_storage_roots()
    for addr, storage_root in changed_roots:
        account_db.set_storage_root(addr, storage_root)
    storage_db.persist()
```

But reading the code wasn't enough for me to figure this out. Eventually I had to look at the tests. I think the reason `make_storage_roots()` can't be merged into `persist()` is that in frontier we want to make a state root for each receipt but don't necessarily want to persist each time. However, given that `StorageDB` is already aware of `AccountDB`, what do you think of having `StorageDB.make_storage_roots()` do the `set_storage_root()` part itself? I think that would lead to more straightforward code, and save the programmer from having to remember something `StorageDB` already knows.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+)
+from eth.db.journal import (
+    JournalDB,
+)
+from eth.tools.logging import (
+    ExtendedDebugLogger
+)
+
+
+class BaseStorageDB(ABC):
```
lithp: I guess it's `py-evm`s convention to make interfaces like this, but I wonder why? It doesn't seem likely another class will  implement this interface any time soon, and the primary effect seems to be separating these useful docstrings from the code they document.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def __init__(self, db: BaseDB, account_db: BaseAccountDB, address: Address) -> None:
+        self._db = db
+        self._account_db = account_db
+        self._address = address
+        self._write_trie = None
+        self._starting_root_hash = None  # type: Hash32
+        self._trie_nodes_batch = None  # type: BatchDB
+
+    def get_storage_root_at_last_persist(self) -> Hash32:
```
lithp: This function is unused

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+    """
+    logger = cast(ExtendedDebugLogger, logging.getLogger("eth.db.storage.StorageLookup"))
+
+    def __init__(self, db: BaseDB, account_db: BaseAccountDB, address: Address) -> None:
+        self._db = db
+        self._account_db = account_db
+        self._address = address
+        self._write_trie = None
+        self._starting_root_hash = None  # type: Hash32
+        self._trie_nodes_batch = None  # type: BatchDB
```
lithp: Creating a `BatchDB` is a pretty lightweight operation, what do you think of simplifying the code by creating it here? As-is, you'll cause an exception if you call `commit_to()` too soon.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+            self.logger.warning(exc)
+            self.logger.debug("expected to get slot, but:", exc_info=True)
+            raise exc
+
+    def __setitem__(self, key: bytes, value: bytes) -> None:
+        hashed_slot = self._decode_key(key)
+        write_trie = self._get_write_trie()
+        write_trie[hashed_slot] = value
+
+    def _exists(self, key: bytes) -> bool:
```
lithp: This method is also unused.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+        the key in _journal_storage, because the cache is only invalidated
+        after a state root change. Otherwise, you will see data since the last
+        storage root was calculated.
+        Keys are set/retrieved as the big_endian encoding of the slot integer, and the rlp-encoded
+        value.
+
+        Journaling batches writes at the _journal_storage layer, until persist is called.
+        It manages all the checkpointing and rollbacks that happen during EVM execution.
+        Keys are set/retrieved as the big_endian encoding of the slot integer, and the rlp-encoded
+        value.
```
lithp: This line: "Keys are set/retrieved as the big_endian encoding of the slot integer, and the rlp-encoded value." appears twice in the above docstring.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+        and so on.
+
+        We generate the pipelines on the fly, as they are needed for reading or writing.
+        We explicitly keep track of which accounts were written to, so that we can identify
+        which accounts might have changed storage roots.
+        """
+        self._account_db = account_db
+        self._raw_store_db = db
+        self._account_stores = {}  # type: Dict[Address, AccountStorageDB]
+        self._dirty_accounts = set()  # type: Set[Address]
```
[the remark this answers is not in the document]
carver: Yeah, in practice a new VM and StorageDB is initialized for every block. But I agree that it's probably cleaner to explicitly drop it at persist time.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+            self._journal_storage.flatten()
+
+    def make_storage_root(self) -> None:
+        """
+        Force calculation of all storage roots
+
+        MUST immediately update account storage roots for reads to be consistent
+        SHOULD make sure that there are pending changes first
+
+        :return: the addresses and storage roots that were updated during storage persistance
```
lithp: I think this line is a copy-paste error?

eth/db/storage.py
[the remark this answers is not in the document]
carver: Oh yeah, somehow a change to that comment didn't make it in here... :eyes:  yup, it was in the next patch-list.

`AccountStorageDB` actually enforces that there is no changed data since the last root was calculated, on each `AccountStorageDB.persist()` call, so I'll just remove the comment. I think the docstring on `persist()` covers it.

> Is there any harm in turning this into a call to `make_storage_roots`? Or maybe it could validate that the state roots have been generated, or maybe `make_storage_roots()` and `persist()` could be combined. Currently there's only one caller so the knowledge that one method must be called before the other isn't exactly distributed but it still feels like it belongs in `StorageDB`.

`VMState` always needs to know the output of `make_storage_roots()` so it can update the `AccountDB`. We could have persist return the updated storage roots, but it feels easy to miss the necessary step of capturing the storage roots coming out of `persist()`. Instead requiring that the caller explicitly handle the new storage roots before persisting seemed the cleanest. Especially since the storage roots are sometimes generated without persisting.

eth/db/storage.py
```diff
@@ -0,0 +1,477 @@
...
+    def set(self, address: Address, slot: int, value: int) -> None:
+        account_store = self._get_address_store(address)
+        if address not in self._dirty_accounts:
+            self._trigger_writeable_storage(address)
+        account_store.set(slot, value)
+
+    def delete(self, address: Address) -> None:
+        account_store = self._get_address_store(address)
+        if address not in self._dirty_accounts:
+            self._trigger_writeable_storage(address)
```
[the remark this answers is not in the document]
carver: It was actually a little bit faster based on a quick `%timeit` run, and I was sort of ambivalent on the syntax. (Other than inlining, which I'm already an board with) If you think it's easier to read as one line, then I'll go ahead and do that. I'll assume you think so unless you comment otherwise.

eth/db/account.py
carver: Yeah, I feel it.

Unfortunately, the `BaseDB` API assumes `bytes` keys and values. We could *maybe* genericize it, but since the whole concept may get a rework in #1759 I'm hesitant to spend much time fiddling.

Luckily the int to bytes conversion is pretty straightforward. The "interesting" stuff like the pad32 and the keccak are still together.

eth/db/storage.py
carver: Some general rationals are:
- easier backend swapping during testing (don't think we use it here, but I could maybe imagine it
- preparing for things that change from one fork to the next
- easier fiddling with swapping in 
- a stronger guarantee about which methods are accessed publicly, since mypy will complain if you access anything only defined on the implementing class.

Not necessarily a good reason, but it's a reason. :)

eth/db/storage.py
carver: It's not like a race condition or anything, it would just be wrong to call `commit` when there are no pending writes. I'll update the docstring and raise an explicit `ValidationError`.

eth/db/storage.py
carver: This one is a little tricky, it's actually used by `BaseDB` for `__contains__` checks. I can add an inline comment.

carver: I suppose we're not doing any explicit `in` checks right now, but I'm comfortable just having this implemented so that it will work when someone needs it.

eth/db/storage.py
carver: Hm, one refers to `_storage_cache` and the other refers to `_journal_storage`. Any idea how to make that clearer?

eth/db/storage.py
lithp: Yeah, it's definitely not a race-condition! And certainly the reason why this bug exists is that users aren't expected to call `commit_to()` with no changes. But, I'd still argue that it's a bug: there's nothing really wrong with calling `commit_to()` when there are no changes. My theory is that if `BatchDB` always exists then this class of bug is impossible in future code.

I'm also kind of surprised that mypy doesn't catch this. Even if I change the type annotation to `Optional[BatchDB]` it still doesn't complain about the null dereference :cry:

eth/db/storage.py
lithp: Ahh, good point, I checked for `in` but missed that `BaseDB` actually uses this. `BaseDB` doesn't just call `__contains__` because then everyone would need to add `# type: ignore`?

lithp: Thanks for the comment! That will help future Brian's :)

eth/db/storage.py
```diff
@@ -222,7 +213,16 @@ def clear_changed_root(self) -> None:
         self._starting_root_hash = None
 
     def commit_to(self, db: BaseDB) -> None:
+        """
+        Trying to commit changes when nothing has been written will raise a
+        ValidationError
+        """
         self.logger.debug2('persist storage root to data store')
+        if self._trie_nodes_batch is None:
```
lithp: Sorry for critiquing even the things you add in response to my critiques :sweat_smile: 

But this check isn't quite right. It's possible for `_trie_nodes_batch` to be not none at the same time as there are no pending changes, `commit_to()` doesn't set `_trie_nodes_batch` to `None` so two calls to `commit_to()` right after each other wouldn't trigger this error.

eth/db/storage.py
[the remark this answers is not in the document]
lithp: > VMState always needs to know the output of make_storage_roots() so it can update the AccountDB.

Do you mean in `BaseState.make_state_root()`? It looks like there it takes the state roots and immediately gives them to `_account_db` and then they go out of scope. Since `StorageDB` already has a reference to `AccountDB` I think it could become easier to work with by calling `set_storage_root()` itself in `make_storage_roots`.

eth/db/storage.py
lithp: Ahh, gotcha, I guess it's okay. Two ideas:
- since `_storage_cache` is never written to by this class and readers should know that all operations to `_journal_storage` will pass through to it, maybe it's fine to drop the first duplication of this line and not specify how keys are encoded.
- Maybe it could become "In both _storage_cache and _journal_storage, keys are set/retreived ..."

eth/db/storage.py
```diff
@@ -0,0 +1,470 @@
...
+        write_trie = self._get_write_trie()
+        try:
+            del write_trie[hashed_slot]
+        except trie_exceptions.MissingTrieNode as exc:
+            self.logger.warning(exc)
+            self.logger.debug("expected to be able to delete slot, but:", exc_info=True)
+
+    @property
+    def has_changed_root(self) -> bool:
+        return self._write_trie and self._write_trie.root_hash != self._starting_root_hash
```
lithp: *nitpick*: There's no null pointer exception here but there's the possibility that some refactor some day might expose one. This would be ever so slightly safer, and it would support making this method simpler, if `self._starting_root_hash` became a boolean: `self.has_changed_root`

eth/db/storage.py
```diff
@@ -0,0 +1,470 @@
...
+
+    def get_changed_root(self) -> Hash32:
+        if self._write_trie is not None:
+            return self._write_trie.root_hash
+        else:
+            raise ValidationError("Asked for changed root when no writes have been made")
+
+    def clear_changed_root(self) -> None:
+        self._write_trie = None
+        self._starting_root_hash = None
```
lithp: This still leaves data behind if the caller had already made changes, here's a failing test:

```python
def test_storage_lookup_clear_clears_changes(account_db):
    base_db = AtomicDB()
    lookup = StorageLookup(base_db, account_db, ADDRESS)

    target = dict()

    lookup[b'key'] = b'val' 
    lookup.clear_changed_root()
    assert not lookup.has_changed_root

    lookup.commit_to(target)
    assert target == {}  # fails here, because _trie_nodes_batch wasn't cleared by clear_changed_root
```

It might be a simpler user-interface if this method was dropped and `commit_to()` set these fields to None.

eth/db/account.py
lithp: Alright, makes sense!

eth/db/journal.py
```diff
@@ -384,6 +384,8 @@ def commit(self, changeset_id: uuid.UUID) -> None:
 
         If this is the base changeset then all changes will be written to
         the underlying database and the Journal starts a new recording.
+        Typically, callers won't have access to the base changeset, because
+        it is dropped during .reset() which is called in JournalDB().
```
lithp: Nice addition :) I thought I had found a bug until I saw this comment, we probably went down the same path :laughing:

eth/vm/base.py
```diff
@@ -841,7 +841,7 @@ def validate_block(self, block: BaseBlock) -> None:
             validate_length_lte(block.header.extra_data, 32, title="BlockHeader.extra_data")
         else:
             parent_header = get_parent_header(block.header, self.chaindb)
-            self.validate_header(block.header, parent_header)
+            self.validate_header(block.header, parent_header, check_seal=False)
```
lithp: What is this line for?

eth/vm/state.py
lithp: The issue: https://github.com/ethereum/py-evm/issues/1760

eth/db/journal.py
lithp: I'm starting to think that `commit()` does the wrong thing. `commit()` should only ever collapse changesets together, and the entire `self.journal.is_empty()` block should move into `persist()`. That would be more straightforward.

eth/vm/state.py
```diff
@@ -159,19 +172,42 @@ def state_root(self) -> Hash32:
...
+        validate_canonical_address(address, title="Storage Address")
+
+        self._storage_db.delete(address)
+        self._account_db.set_storage_root(address, BLANK_ROOT_HASH)
 
     def delete_account(self, address: Address) -> None:
+        validate_canonical_address(address, title="Storage Address")
+
         self._account_db.delete_account(address)
+        self._storage_db.delete(address)
```
lithp: I think that this orphans data. I'm not sure but, `_storage_db.delete` doesn't do anything, it reduces to a `JournalDB.clear()` which isn't persisted. So the account is deleted but the state trie is left behind in the database, which would lead to ever growing disk usage.

eth/db/account.py
```diff
@@ -390,30 +356,42 @@ def _set_account(self, address: Address, account: Account) -> None:
...
         return self.state_root
 
     def persist(self) -> None:
-        self.make_state_root()
-        self._batchtrie.commit(apply_deletes=False)
-        self._batchdb.commit(apply_deletes=True)
+        self._validate_generated_root()
+        with self._raw_store_db.atomic_batch() as write_batch:
+            self._batchtrie.commit_to(write_batch, apply_deletes=False)
+            self._batchdb.commit_to(write_batch, apply_deletes=False)
```
lithp: Also, why was this `apply_deletes=True` turned into `apply_deletes=False`?

eth/db/account.py
```diff
@@ -390,30 +356,42 @@ def _set_account(self, address: Address, account: Account) -> None:
...
         self._journaltrie.persist()
         return self.state_root
 
     def persist(self) -> None:
-        self.make_state_root()
-        self._batchtrie.commit(apply_deletes=False)
-        self._batchdb.commit(apply_deletes=True)
+        self._validate_generated_root()
+        with self._raw_store_db.atomic_batch() as write_batch:
+            self._batchtrie.commit_to(write_batch, apply_deletes=False)
```
lithp: Won't `apply_deletes=False` orphan a lot of intermediate nodes?

eth/vm/state.py
lithp: Maybe this also needs a call to `self._account_db.set_storage_root(address, BLANK_ROOT_HASH)`? I'm not sure

eth/db/storage.py
lithp: Yeah, that makes sense, thanks for the answer!

lithp:
Okay, I think that wraps up the review. I found a bunch of tiny things and some small refactors which I think would be improvements, but it took me a long time to review and I still don't have many thoughts on the larger architecture. I think the biggest reason for both is that there are a lot of interlocking stateful pieces which could get out of sync.

I'm not sure what to say here because I don't have any suggestions of my own but it feels like this might not be the best factoring. For one, these classes know a lot about each other, so much that it's difficult to reason about them in isolation:
- In order to understand that it was safe for `StorageDB.discard()` to call `discard()` on *all* dirty accounts, even ones which didn't exist when the savepoint was created, I needed to know how `AccountStorageDB.discard()` worked.
- I still don't know whether it's okay for `AccountStorageDB.clear()` to call `self._journal_storage_clear()`, clears aren't persisted so I don't think an account someone called `delete()` on would show up in `_get_changed_roots` so it skips the `_validate_flushed` check even though that seems important. The tests pass so *something* makes this okay and I think might be `BaseState`(?), a class in a completely different file(!)
- StorageLookup builds different tries for reads and writes, writes go to a batch db while reads are routed to the underlying db. This seems like it would prevent us from reading our own writes but the tests pass so I guess I'm missing something, maybe `AccountStorageDB._storage_cache` papers over the difference?

It's also a bit strange that `AccountDB` has to be threaded so deeply into `storage.py`, and that storage cares so much about exactly when AccountDB updates the state root, and that nearly every `BaseState` method which manipulates the storage_db does something with the account_db right after. These seem like signs that account storage isn't so separate from the account_db.

I'm not sure that factoring classes out by which part of the database is touched is the right idea. It seems like a lot of this code and complication comes from keeping all of our journals and caches in sync. That might be easier if we had fewer journals and caches!

eth/db/storage.py
carver: Haha, no apologizing for pointing out code that looks bad to you! If I disagree that it's worth taking care of right away, we can always punt to an issue.

But yeah, I agree :). So the next change is:
- set `_trie_nodes_batch = None` in `clear_changed_root()`
- call `clear_changed_root()` at the end of `commit_to()`, instead of having the caller to manage that. No reason the caller needs to handle it AFAICT
- move `clear_changed_root()` to `_clear_changed_root()` since it's never called outside the class anymore

eth/db/storage.py
carver: I can make the `None` check explicit, which should make a broken refactor more obvious. I'm not as keen on changing `_starting_root_hash` to be a `bool` since it is possible to change to a different root and change back. In that case I expect `has_changed_root` to be `False`. The most obvious way I see to check that is to keep the original root hash around.

eth/db/storage.py
carver: Hah, yeah, I same to the same conclusion while handling the other comment above.

eth/db/journal.py
carver: > Nice addition :) I thought I had found a bug until I saw this comment, we probably went down the same path

Haha, yeah, I remember that surprising me a bit.

> I'm starting to think that `commit()` does the wrong thing. `commit()` should only ever collapse changesets together, and the entire `self.journal.is_empty()` block should move into `persist()`. That would be more straightforward.

I think I'm on board, but probably belongs in a different PR. Opening an issue.

eth/vm/base.py
carver: Oops X) Not sure how that slipped in.

eth/vm/state.py
carver: Pruning old state tries is something we're not attempting to do at this point, for the account or storage tries. Trinity is essentially an archive node at this point.

eth/db/account.py
carver: `_batchdb` used to have storage tries in it, but now it's just the code. There is no concept of deleting code, so this was flipped off.

eth/db/account.py
carver: Yes, and yes. But we don't know which recent tries still need access to those intermediate nodes. Also, we don't know what other accounts are still using the code with that hash. For now this is the easy/safe approach.

carver:
> * In order to understand that it was safe for `StorageDB.discard()` to call `discard()` on _all_ dirty accounts, even ones which didn't exist when the savepoint was created, I needed to know how `AccountStorageDB.discard()` worked.

Not just *safe*, but required! We must support handling checkpoint actions on items that are dirty. I could take another stab at an alternative I had considered, which was to immediately record any new account stores with the most recent checkpoint. It was a way messier implementation the last time I attempted it, but I think that was before I added the ability to define your own checkpoint in `JournalDB.record`. It seems it might be cleaner now.

> * I still don't know whether it's okay for `AccountStorageDB.clear()` to call `self._journal_storage_clear()`, clears aren't persisted so I don't think an account someone called `delete()` on would show up in `_get_changed_roots` so it skips the `_validate_flushed` check even though that seems important. The tests pass so _something_ makes this okay and I think might be `BaseState`(?), a class in a completely different file(!)

Yeah, so I made this warning to try to explain:
```
Unlike set(), deleting storage will not cause :meth:`make_storage_roots` to emit a new                         
storage root (and therefore will not persist deletes to the base database).                                    
The account's storage root must be explicitly set to the empty root.     
```

So the caller has to change the account's storage root hash to the empty root explicitly. (in a journaled way, on the account db). Yup, this caller happens to be `BaseState` right now. I agree this is an ugly part of the current implementation. I don't have any better ideas.

> * StorageLookup builds different tries for reads and writes, writes go to a batch db while reads are routed to the underlying db. This seems like it would prevent us from reading our own writes but the tests pass so I guess I'm missing something, maybe `AccountStorageDB._storage_cache` papers over the difference?

We read our own writes if we have made any writes:
```py
    def _get_read_trie(self) -> HexaryTrie:
        storage_root = self._account_db.get_storage_root(self._address)
        if storage_root == BLANK_ROOT_HASH:
            return HexaryTrie({})
        elif self._write_trie is not None:
            return self._write_trie
        else:
            ...
```

> It's also a bit strange that `AccountDB` has to be threaded so deeply into `storage.py`, and that storage cares so much about exactly when AccountDB updates the state root, and that nearly every `BaseState` method which manipulates the storage_db does something with the account_db right after. These seem like signs that account storage isn't so separate from the account_db.

Yeah, so *most* of this is related to deletes. In the original design, it had to look up the current account's storage root on demand, because it could have been arbitrarily reset elsewhere (due to a delete, or a delete reversal because of a rollback).

Although this design came before the `JournalDB.clear()` concept, so it's possible that it could be simplified a bit now. Since now, a `clear()` explicitly intercepts calls to other keys and treats them as missing.

I can take a stab at reading out the storage root hash immediately on initialization, again. If that all works correctly (because of the way `clear()` works now), then I think we could get rid of the `AccountDB` dependency altogether. That would make `BaseState` responsible for passing in a pair of `account: Address, storage_root: Hash32` instead of just `Address`, for any `get()` calls, but that seems like a worthwhile tradeoff.

I'll explore and report back.

carver:
> > * In order to understand that it was safe for `StorageDB.discard()` to call `discard()` on _all_ dirty accounts, even ones which didn't exist when the savepoint was created, I needed to know how `AccountStorageDB.discard()` worked.
> 
> Not just _safe_, but required! We must support handling checkpoint actions on items that are dirty. I could take another stab at an alternative I had considered, which was to immediately record any new account stores with the most recent checkpoint. It was a way messier implementation the last time I attempted it, but I think that was before I added the ability to define your own checkpoint in `JournalDB.record`. It seems it might be cleaner now.

Nope, I remember now. It could be *any* checkpoint in the past that is being committed or reverted. I don't have a better solution than the current one.

carver:
> I can take a stab at reading out the storage root hash immediately on initialization, again. If that all works correctly (because of the way `clear()` works now), then I think we could get rid of the `AccountDB` dependency altogether. That would make `BaseState` responsible for passing in a pair of `account: Address, storage_root: Hash32` instead of just `Address`, for any `get()` calls, but that seems like a worthwhile tradeoff.
> 
> I'll explore and report back.

Yeah, so this worked about like expected. I think the code got clearer, so I followed through and made AccountDB responsible for all the storage again. This got rid of the all-account `StorageDB`, and now `AccountDB` does that orchestration, and uses `AccountStorageDB` directly, passing in the storage root when needed.

So now the control is flipped: accounts access storage and not the other way around.

@lithp all the changes since we chatted are in 8f53ce3. Good enough for a :+1: for now?

eth/db/account.py
lithp: Ahh, really good point re multiple contracts might be sharing that code!

eth/db/storage.py
```diff
@@ -151,7 +78,7 @@ def _get_write_trie(self) -> HexaryTrie:
         return self._write_trie
 
     def _get_read_trie(self) -> HexaryTrie:
-        storage_root = self._account_db.get_storage_root(self._address)
+        storage_root = self._storage_root
```
lithp: I *think* that this first conditional means that if we start with a state_root of `BLANK_ROOT_HASH` then we'll never read our own writes. Though, that must be wrong, I'm surprised that none of the state tests would exercise that!

eth/db/storage.py
```diff
@@ -129,9 +56,9 @@ class StorageLookup(BaseDB):
     """
     logger = cast(ExtendedDebugLogger, logging.getLogger("eth.db.storage.StorageLookup"))
 
-    def __init__(self, db: BaseDB, account_db: BaseAccountDB, address: Address) -> None:
+    def __init__(self, db: BaseDB, storage_root: Hash32, address: Address) -> None:
         self._db = db
-        self._account_db = account_db
+        self._storage_root = storage_root
```
lithp: Now that `_storage_root` is constant, I think that `_storage_root` can be renamed to `_starting_root_hash`, and some of the manipulation of `_starting_root_hash` can be dropped.

eth/db/storage.py
carver: Hm, not sure why that wouldn't be exercised. I guess it's possible. The test would have to run on a pre-Byzantium chain, and modify state in a second transaction (we don't ever read out storage after trie writes, since Byzantium).

Anyhow, it seems vestigial from the old handling of account deletes, and I think it's cleaner to just drop it... pushed.

carver:
All that's left is to squash and merge :+1:

PR #1738, opened 2019-03-29
Commit batch to target db

carver:
### What was wrong?

I want to commit queued changes to a target database (specifically, I want to commit changes to an atomic write batch)

### How was it fixed?

Added new `BatchDB.commit_to()` method and tests.

#### Cute Animal Picture


carver:
> Effectively a wrapper around the equivalent.
> 
> ```python
> db_a.commit_to(db_b)
> # same as
> db_a.diff().apply_to(db_b)
> ```

Yup, the only other notable difference is that `apply_to()` does *not* clear the diff, so you would reapply the changes if you run `diff.apply_to()` twice.

PR #1739, opened 2019-03-29
Get access to all the pending JournalDB changes

carver:
### What was wrong?

Wanted an easy way to get access to all pending changes. (for #1735 )

### How was it fixed?

- Added a new `JournalDB.diff()` which reuses the `DBDiff` class used elsewhere to represent a single diff
- Added test that `JournalDB.diff().apply_to(db)` is equivalent to `JournalDB(db).persist()`
- Added a new `DBDiff` equality test to help test its use in the journal
- Added `__iter__` and `__len__` to MemoryDB to enable equality and length checking

Several of these changes seem only a bit related. The final goal was the journal db change, so all the other changes support building and testing that. Still, if you would like me to split it out, I can. It is already split into separate commits, so it will be easy.

#### Cute Animal Picture

### DIFF
--- a/eth/db/journal.py
+++ b/eth/db/journal.py
@@ -1,5 +1,5 @@
 import collections
-from typing import cast, Dict, Union  # noqa: F401
+from typing import cast, Dict, Set, Union  # noqa: F401
 import uuid
 
 from eth_utils.toolz import (
@@ -12,6 +12,7 @@ from eth_utils import (
 )
 
 from eth.db.backends.base import BaseDB
+from eth.db.diff import DBDiff, DBDiffTracker
 
 
 class DeletedEntry:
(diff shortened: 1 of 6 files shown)


tests/database/test_journal_db.py
```diff
@@ -194,3 +196,23 @@ def test_returns_key_from_underlying_db_if_missing(journal_db, memory_db):
     assert memory_db.exists(b'1')
 
     assert journal_db.get(b'1') == b'test-a'
+
+
+def test_journal_db_diff_application_mimics_persist(journal_db, memory_db):
```
pipermerriam: This seems like something that is ideal for a property based test which does more in depth excercising of the various orders of key deletions, additions and updates.

eth/db/journal.py
```diff
@@ -262,3 +263,27 @@ def reset(self) -> None:
         """
         self.journal = Journal()
         self.record()
+
+    def diff(self) -> DBDiff:
+        """
+        Generate a DBDiff of all pending changes.
+        These are the changes that would occur if you called :meth:`persist()`.
```
cburgdorf: *Nitpick*, We try to avoid *you* form, so this should rather be:

>These are the changes that would occur if one called :meth:`persist()`.

Or maybe 

>These are the changes that would occur :meth:`persist()` was called

PR #1740, opened 2019-03-29
DBDiff inspection

carver:
### What was wrong?

Wanted to get information about a `DBDiff` in #1735 -- specifically a list of keys that have been updated.

### How was it fixed?

Added `DBDiff.new_keys()` to return all keys that have been updated. While there, it made sense to add the companion methods:
- `DBDiff.new_items()` (the (key, value) pairs of all updated values)
- `DBDiff.deleted_keys()`

Also, added some basic tests that touch on all these.

#### Cute Animal Picture

### DIFF
--- a/eth/db/diff.py
+++ b/eth/db/diff.py
@@ -6,11 +6,13 @@ from typing import (
     Dict,
     Iterable,
     Union,
+    Tuple,
     TYPE_CHECKING,
 )
 
 from eth_utils import (
     encode_hex,
+    to_tuple,
 )
 
 from eth.db.backends.base import BaseDB
@@ -128,7 +130,8 @@ class DBDiff(ABC_Mapping):
 
     def __iter__(self) -> None:
         raise NotImplementedError(
-            "Cannot iterate through changes, use apply_to(db) to update a database"
+            "Cannot iterate through changes, use apply_to(db) to update a database. "
+            "Also, pending_keys(), deleted_keys(), and pending_items() might be of interest."
         )
 
     def __eq__(self, other: object) -> bool:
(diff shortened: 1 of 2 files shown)


carver:
This is related but pretty separable from #1739 -- thought it would be easiest for you to look at after reviewing that other one @cburgdorf

eth/db/diff.py
```diff
@@ -124,12 +129,42 @@ def __getitem__(self, key: bytes) -> bytes:
...
+    def deleted_keys(self) -> Iterable[bytes]:
+        """
+        List all the keys that have been deleted.
+        """
+        for key, value in self._changes.items():
+            if value is DELETED:
+                yield key
+
+    @to_tuple
+    def new_keys(self) -> Iterable[bytes]:
```
cburgdorf: Just noting that the name is slightly confusing because at first sight it makes it look as if this (and similarly `new_items`) would only contain *additions*  when in reality the `DiffTracker` simply does not differentiate between *updated*  or *inserted* keys. I think it can not differentiate between these two cases but maybe the API name can be clearer to reflect that?

I'm not the best with naming suggestions but I'll try nonetheless:

- `set_keys` / `set_items`
- `touched_keys` / `touched_items`
- `new_and_updated_keys` / `new_and_updated_items`

carver: Agreed. Do you think that `updated_keys()` has the opposite problem?

If you don't like `updated_keys()` or `changed_keys()`, how about `pending_keys()`, `dirty_keys()` or `new_value_keys()`?

`set` has the problem that it sounds like a verb. `touched` could also apply to deleted keys, and `new_and_updated` is long. Just hoping we can find a magic name :)

cburgdorf: >Do you think that `updated_keys()` has the opposite problem?

Yes, kind of. I think I like `pending` and `dirty` better :thumbsup:

carver: `pending` it is :+1:

PR #1745, opened 2019-04-10
Add JournalDB.clear()

carver:
### What was wrong?

When storage is deleted, I want a quick way to immediately clear all data, but still be committable/rollback-able as normal. (We didn't have to deal with this before, because we simply calculated and kept *all* intermediate storage states mid-block.)

Now I want to journal the changes, but it wasn't built for the idea of clearing out all underlying data, then rolling that back. So I added `JournalDB.clear()`.

### How was it fixed?

Added `clear()` method and tests.

Also note c8c1643 which fails on master. It seems like this should be the desired behavior: when a key is deleted (but not persisted) then `journal.exists(key)` should return `False`. (Master returns `True`)

#### Cute Animal Picture

### DIFF
--- a/eth/db/journal.py
+++ b/eth/db/journal.py
@@ -36,6 +36,7 @@ class Journal(BaseDB):
         # to a dictionary of key:value pairs with the recorded changes
         # that belong to the changeset
         self.journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, Union[bytes, DeletedEntry]]]  # noqa E501
+        self._clears_at = set()  # type: Set[uuid.UUID]
 
     @property
     def root_changeset_id(self) -> uuid.UUID:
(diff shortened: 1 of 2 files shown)


carver:
cc @cburgdorf I'm setting a reminder to ask you to review this after you get back, since you probably know this code best.

eth/db/journal.py
```diff
@@ -80,43 +81,86 @@ def record_changeset(self) -> uuid.UUID:
...
         later changesets.
         """
         if changeset_id not in self.journal_data:
-            raise KeyError("Unknown changeset: {0}".format(changeset_id))
+            raise KeyError(changeset_id, "Unknown changeset in JournalDB")
 
         all_ids = tuple(self.journal_data.keys())
         changeset_idx = all_ids.index(changeset_id)
         changesets_to_pop = all_ids[changeset_idx:]
+        popped_clears = tuple(idx for idx in changesets_to_pop if idx in self._clears_at)
```
pipermerriam: It *appears* that we only want the last one based on line 98.  Can we invert this to iterate over `changesets_to_pop` in reverse order and stop at the first one we encounter that is truthy for `in self._clears_at`

tests/database/test_journal_db.py
```diff
@@ -141,6 +199,49 @@ def test_commit_merges_changeset_into_previous(journal_db):
...
+
+    journal_db.commit(changeset3)
+
+    assert not journal_db.exists(b'only-in-wrapped')
+    assert not journal_db.exists(b'wrapped-and-journal')
+    assert not journal_db.exists(b'only-in-journal')
+    assert not journal_db.exists(b'in-merged-snapshot')
+    assert not journal_db.exists(b'in-unmerged-snapshot')
+    assert not journal_db.exists(b'just-before-clear')
+    assert not journal_db.exists(b'just-before-clear')
```
pipermerriam: this assertion is done twice.  Is this intentional?

eth/db/journal.py
carver: Unfortunately, we also use it here:
https://github.com/ethereum/py-evm/pull/1745/files#diff-4691bcc3798d57741a237f81721f0b81R118

```py
        self._clears_at.difference_update(popped_clears)
```

I can't think of better way to remove the old clears. Although it does remind me that we should empty the set on a `reset()`.

tests/database/test_journal_db.py
carver: Nope! Thanks :)

PR #1746, opened 2019-04-10
JournalDB.squash() and custom checkpoint ID

carver:
### What was wrong?

We will maintain a list of JournalDB's (one for each active storage slot). 

#### Squashing
When we create a JournalDB in the middle of a block, it may not have been keeping up will all previous checkpoints. If an earlier checkpoint is discarded, we can simply reset the entire JournalDB. If an earlier checkpoint is committed, then we want to commit everything but we don't want to persist it.

We could track the first checkpoint seen and commit back to that, but this public API on JournalDB seemed cleaner: Hence the new `squash()`.

#### Custom checkpoint IDs
It is ideal for a number of reasons to reuse the same checkpoint id for each of the different journals. This makes for a short and sweet checkpoint ID for the whole set of journals.

### How was it fixed?

Added implementations and tests for `JournalDB.squash` and `.record(custom_checkpoint_id)`

#### Cute Animal Picture

### DIFF
--- a/tests/database/test_journal_db.py
+++ b/tests/database/test_journal_db.py
@@ -163,6 +163,78 @@ def test_committing_middle_changeset_merges_in_subsequent_changesets(journal_db)
     assert journal_db.journal.has_changeset(changeset_c) is False
 
 
+def test_squash_does_not_persist_0_checkpoints(journal_db, memory_db):
+    journal_db.set(b'1', b'test-a')
+
+    # should have no effect
+    journal_db.squash()
+
+    assert b'1' not in memory_db
+    assert b'1' in journal_db
+
+    journal_db.persist()
+
+    assert b'1' in memory_db
+
+
+def test_squash_does_not_persist_1_checkpoint(journal_db, memory_db):
+    journal_db.set(b'1', b'test-a')
+
+    checkpoint = journal_db.record()
+
+    journal_db.set(b'2', b'test-b')
+
+    # should only remove this checkpoint, but b'2' still be available
+    assert journal_db.has_changeset(checkpoint)
+    journal_db.squash()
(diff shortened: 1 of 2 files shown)


tests/database/test_journal_db.py
```diff
@@ -163,6 +201,78 @@ def test_committing_middle_changeset_merges_in_subsequent_changesets(journal_db)
...
+
+    assert b'1' in memory_db
+
+
+def test_squash_does_not_persist_1_checkpoint(journal_db, memory_db):
+    journal_db.set(b'1', b'test-a')
+
+    checkpoint = journal_db.record()
+
+    journal_db.set(b'2', b'test-b')
```
pipermerriam: I really liked you're other pr that used semantically meaningful keys.

```python
journal_db.set((b'in-base-checkpoint', ...)

checkpoint = journal_db.record()

journal_db.set(b'in-first-checkpoint', ...)

...
```

It made the later assertions a lot easier to read without having to re-evaluate the whole test body.

tests/database/test_journal_db.py
```diff
@@ -46,6 +49,41 @@ def test_snapshot_and_revert_with_set(journal_db):
     assert journal_db.get(b'1') == b'test-a'
 
 
+def test_custom_snapshot_and_revert_with_set(journal_db):
+    journal_db.set(b'1', b'test-a')
```
pipermerriam: Same here with using semantically meaningful values.

```python
journal_db.set(b'k', b'original-journaled-value')

changeset = ...

...

journal_db.set(b'k', b'value-in-custom-changeset')

...

journal_db.discard(changeset)

assert journal_db.get(b'k') == b'original-journaled-value')
```

PR #1747, opened 2019-04-11
Lazy VM.block generation

carver:
This makes it easier to inspect data, even if the block transactions are
missing (for example, during light sync or parts of beam sync)

### What was wrong?

Loading the VM immediately tries to look up the block in the database, but we don't always have the block ready. (Also, it's slower, for cases when we don't need the block)

### How was it fixed?

Lazily generate `vm.block` when it's needed. Often all we need is the header, so make that directly accessible on the VM, and prefer its usage `vm.header` over `vm.block.header`.

#### Cute Animal Picture

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -87,10 +87,20 @@ class BaseVM(Configurable, ABC):
     @property
     @abstractmethod
     def state(self) -> BaseState:
-        raise NotImplementedError("VM classes must implement this property")
+        pass
 
     @abstractmethod
     def __init__(self, header: BlockHeader, chaindb: BaseChainDB) -> None:
+        pass
+
+    @property
+    @abstractmethod
+    def header(self) -> BlockHeader:
+        pass
+
+    @property
+    @abstractmethod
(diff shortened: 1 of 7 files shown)


eth/vm/base.py
```diff
@@ -374,15 +384,34 @@ class VM(BaseVM):
...
     _state = None
+    _block = None
 
     def __init__(self, header: BlockHeader, chaindb: BaseChainDB) -> None:
         self.chaindb = chaindb
-        self.block = self.get_block_class().from_header(header=header, chaindb=self.chaindb)
+        self._initial_header = header
+
+    @property
+    def header(self) -> BlockHeader:
```
pipermerriam: What do you think about going ahead and converting these to methods.

```python
class VM:
    def get_header(self):
        ...
    def get_block(self):
        ...
```

Makes it a lot clearer that they have side effects.

carver: I am :+1: on it, but the API is used in a *lot* of places right now and I'd like to get this live for beam sync. So I'll add a good first issue.

PR #1749, opened 2019-04-11
Batch db optional read through delete

carver:
### What was wrong?

I have a scenario where I want to delete all intermediate data, but leave the underlying data immutable. (Specifically: adding several items to a trie, which generates a bunch of nodes we don't need)

Unfortunately, that means that in memory, the trie tries to "delete" the old root hash (and other old nodes). We don't have to worry about this getting propagated to the DB, because we use `BatchDB.commit(apply_deletes=False)`. But, sometimes we need to read out of those nodes that we deleted during the process, while we are in the process of reading.

### How was it fixed?

Added a new init flag to `BatchDB` that allows you to read through deletes. If a delete happens in batch, you fall back to getting whatever the wrapped database tells you the value is.

`read_through_deletes` and `apply_deletes` seem heavily linked. It *might* make sense do do something like `NonDeletingBatchDB` (ugh, that name) and a regular `BatchDB`. Open to thoughts.

#### Cute Animal Picture

### DIFF
--- a/eth/db/batch.py
+++ b/eth/db/batch.py
@@ -22,9 +22,10 @@ class BatchDB(BaseDB):
     wrapped_db = None  # type: BaseDB
     _track_diff = None  # type: DBDiffTracker
 
-    def __init__(self, wrapped_db: BaseDB) -> None:
+    def __init__(self, wrapped_db: BaseDB, read_through_deletes: bool = False) -> None:
         self.wrapped_db = wrapped_db
         self._track_diff = DBDiffTracker()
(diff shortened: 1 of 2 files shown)


carver:
Ping @pipermerriam for :eyes:

carver:
> This feels very awkward and spider sense says it may not be the right solution... but I don't really have a better idea right now.

Yup, I'm in the same boat with all that.

lithp:
This is one of the cuter animal pictures py-evm has used :heart_eyes:

PR #1750, opened 2019-04-11
Bring in HasExtendedDebugLogger into this codebase

pipermerriam:
### What was wrong?

The `HasExtendedDebugLogger` from the `trinity` codebase is useful beyond that codebase.

https://github.com/ethereum/trinity/blob/3b52bdeef6e145f9d079de1dfdf0e246bcf62b54/p2p/persistence.py#L9

### How was it fixed?

Added it to `eth.tools.logging` and used it in places where it made sense.

#### Cute Animal Picture

68312a6e5df3945fc22ac808fb0a7b55

### DIFF
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -3,7 +3,6 @@ from abc import (
     abstractmethod
 )
 from uuid import UUID
-import logging
 from lru import LRU
 from typing import cast, Set, Tuple  # noqa: F401
 
@@ -49,7 +48,7 @@ from eth.validation import (
     validate_canonical_address,
 )
 from eth.tools.logging import (
-    ExtendedDebugLogger
+    HasExtendedDebugLogger,
 )
 from eth._utils.padding import (
     pad32,
@@ -58,7 +57,7 @@ from eth._utils.padding import (
 from .hash_trie import HashTrie
 
 
-class BaseAccountDB(ABC):
+class BaseAccountDB(HasExtendedDebugLogger, ABC):
 
     @abstractmethod
     def __init__(self) -> None:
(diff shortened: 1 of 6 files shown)


carver:
Hm, maybe this is the kind of thing web3.py wants to use too. Does it belong in a new `eth-utils.logging`?

pipermerriam:
Moving to `eth_utils.logging`

pipermerriam:
> This whole mechanism makes me a little nervous

@lithp I am leaning somewhere in that direction now too.  The logging library [is of the opinion that adding custom levels is *likely* an anti pattern](https://docs.python.org/3/howto/logging.html#custom-levels).

I'm thinking about re-working this whole thing to use the logging namespace to replace this mechanism since you can set levels at the namespace level and we can just control the verbosity of output by using a different logger instance with a different namespace for more verbose logging.

```
class Something:
    logger = logging.getLogger('trinity.discovery")
    logger2 = logging.getLogger("trinity.discovery.verbose")
```

Not sure how I feel about this yet, but I'm now sure that this isn't functionality that's ready to live in eth-utils.

Does this spark any additional thoughts from anyone?

carver:
> ```
> class Something:
>     logger = logging.getLogger('trinity.discovery")
>     logger2 = logging.getLogger("trinity.discovery.verbose")
> ```

Hm, the weird thing here is: what does `logger2.warning()` mean? Should that show up in the client or the debug logs or neither (by default)?

---

> > This whole mechanism makes me a little nervous
> 
> @lithp I am leaning somewhere in that direction now too. The logging library [is of the opinion that adding custom levels is _likely_ an anti pattern](https://docs.python.org/3/howto/logging.html#custom-levels).

I don't really see any way around having more levels: we definitely want at least two levels lower than INFO. So that leaves us with how to implement it.

Being worried about a different package calling `setLoggerClass()` seems reasonable. I think we can reasonably handle the case by not using `setLoggerClass()` ourselves, and wrapping the logger instead, like:
```py
if self._logger is None:
  self._logger = ExtendedDebugLogger(logging.getLogger(...))
```


wanting to execute

pipermerriam:
just correcting code snipped for `Logger.__init__` signature:

```
if self._logger is None:
  base_logger = logging.getLogger(...)
  self._logger = ExtendedDebugLogger(base_logger.name, base_logger.level)
```

pipermerriam:
Replaced by https://github.com/ethereum/eth-utils/pull/158

PR #1751, opened 2019-04-12
Prepare for Istanbul

fubuloubu:
### What was wrong?
Istanbul is coming! Multiple EIPs incoming!

### How was it fixed?
This PR just adds the necessary boilerplate to start Istanbul updates. It does not integrate with anything at all yet. Basically dead code.

#### Cute Animal Picture
play dead

### DIFF
--- a/eth/tools/builder/chain/__init__.py
+++ b/eth/tools/builder/chain/__init__.py
@@ -23,6 +23,7 @@ from .builders import (  # noqa: F401
     spurious_dragon_at,
     tangerine_whistle_at,
     constantinople_at,
+    istanbul_at,
 )
 
 
@@ -33,6 +34,7 @@ mainnet_fork_at_fns = (
     spurious_dragon_at,
     tangerine_whistle_at,
     constantinople_at,
+    istanbul_at,
 )
 
 
@@ -60,6 +62,7 @@ class API:
     spurious_dragon_at = staticmethod(spurious_dragon_at)
     byzantium_at = staticmethod(byzantium_at)
     constantinople_at = staticmethod(constantinople_at)
+    istanbul_at = staticmethod(istanbul_at)
 
     # iterable of the fork specific functions
     mainnet_fork_at_fns = mainnet_fork_at_fns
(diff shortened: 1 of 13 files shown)


carver:
Seems worth adding a test to make sure that we can create a chain and execute a transaction with it, I guess with a Constantinople opcode. It's okay if this test will fail later as Istanbul gets updated, we'll just upgrade the test.

For lint, it looks like there are some imports to remove from opcodes.py

fubuloubu:
@carver The linting error doesn't seem right since I am using all of those imports in `as_opcode()`

Am I doing something incorrectly where it sees this as dead code or something? Seem strange!

carver:
> @carver The linting error doesn't seem right since I am using all of those imports in `as_opcode()`
> 
> Am I doing something incorrectly where it sees this as dead code or something? Seem strange!

Do you have some local changes that didn't get pushed or something? `as_opcode()` isn't being called in `eth/vm/forks/istanbul/opcodes.py`

fubuloubu:
Added tests for what I could see that copies Constantinople. The ethereum tests repo would probably be updated for the newer opcodes, but everything works as is now.

fubuloubu:
Ahhh, I see what I was doing wrong! I was looking at the branch where I added the opcode for eip-1344.

Will push a fix

fubuloubu:
Crap, it really doesn't seem to like an empty `UPDATED_OPCODES`, so I guess I'll just remove the rest of the boilerplate for that.

fubuloubu:
Note: a later PR will have to add back commits https://github.com/ethereum/py-evm/pull/1751/commits/225c119fc19b921fe1d20bafa967cc83d333ac24 and https://github.com/ethereum/py-evm/pull/1751/commits/d53a14dbb44732375ab762e3cfc0ea3509792ee8

carver:
Adding it here should get Istanbul included in quite a few tests:
https://github.com/ethereum/py-evm/blob/30eb1ae4a2d77db0f63e9a635345cc3ac718d9b8/tests/conftest.py#L71-L83

fubuloubu:
Oh nice, I missed that one. Will add!

pipermerriam:
We should create a checklist and put it somewhere in the developer documentation for creating a new VM.  This PR seems like a good template for that.

Issue #1753, opened 2019-04-18
Update to latest `eth-utils` and `mypy`


pipermerriam:
### What is wrong?

Latest `eth-utils==0.5.1` has a bugfix for some typing issues with `to_dict` which is used in a number of places here.

Mypy `>=0.701` is faster and contains bugfixes.

### How can it be fixed

Update these two dependencies and then fix any new type checking issues that surface.

PR #1754, opened 2019-04-18
Remove the cryptography library from dependencies

pipermerriam:
### What was wrong?

This library specifies a dependency on `cryptography` but it doesn't actually appear to be used.

### How was it fixed?

Removed it.

#### Cute Animal Picture

zebra1


pipermerriam:
Looks like this is blocked by #1753 since the typing errors are now being thrown in CI

PR #1755, opened 2019-04-18
Move several methods from AccountDB to VMState

carver:
- set_storage
- get_storage
- delete_storage
- delete_account
- make_state_root
- persist

### What was wrong?

#1735 will cause some churn in the `AccountDB` API, which would cause effects all over the code. It makes sense for external callers to just call `VMState` and ask the state class to figure out the right way to delegate the responsibility (like how to get storage values).

### How was it fixed?

Moved all the listed methods from `AccountDB` to `VMState` (they just pass through the implementations, until #1735)

#### Cute Animal Picture

### DIFF
--- a/eth/_utils/db.py
+++ b/eth/_utils/db.py
@@ -6,15 +6,14 @@ from eth_typing import (
     Hash32,
 )
 
-from eth.db.account import (
-    BaseAccountDB,
-)
-
 from eth.rlp.headers import (
     BlockHeader,
 )
 from eth.typing import (
     AccountState,
+)
+from eth.vm.state import (
+    BaseState,
 )
 
 if TYPE_CHECKING:
@@ -35,7 +34,8 @@ def get_block_header_by_hash(block_hash: Hash32, db: 'BaseChainDB') -> BlockHead
     return db.get_block_header_by_hash(block_hash)
 
 
-def apply_state_dict(account_db: BaseAccountDB, state_dict: AccountState) -> BaseAccountDB:
+def apply_state_dict(state: BaseState, state_dict: AccountState) -> None:
+    account_db = state.account_db
 
     for account, account_data in state_dict.items():
         account_db.set_balance(account, account_data["balance"])
@@ -43,6 +43,4 @@ def apply_state_dict(account_db: BaseAccountDB, state_dict: AccountState) -> Bas
         account_db.set_code(account, account_data["code"])
 
         for slot, value in account_data["storage"].items():
-            account_db.set_storage(account, slot, value)
-
-    return account_db
+            state.set_storage(account, slot, value)
--- a/eth/_utils/db.py
+++ b/eth/_utils/db.py
@@ -35,12 +35,10 @@ def get_block_header_by_hash(block_hash: Hash32, db: 'BaseChainDB') -> BlockHead
 
 
 def apply_state_dict(state: BaseState, state_dict: AccountState) -> None:
-    account_db = state.account_db
-
     for account, account_data in state_dict.items():
-        account_db.set_balance(account, account_data["balance"])
-        account_db.set_nonce(account, account_data["nonce"])
-        account_db.set_code(account, account_data["code"])
+        state.set_balance(account, account_data["balance"])
+        state.set_nonce(account, account_data["nonce"])
+        state.set_code(account, account_data["code"])
 
         for slot, value in account_data["storage"].items():
             state.set_storage(account, slot, value)
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -413,7 +413,6 @@ class Chain(BaseChain):
 
         pre_genesis_header = BlockHeader(difficulty=0, block_number=-1, gas_limit=0)
         state = genesis_vm_class.build_state(base_db, pre_genesis_header)
-        account_db = state.account_db
 
         if genesis_state is None:
             genesis_state = {}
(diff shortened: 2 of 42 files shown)


eth/_utils/db.py
```diff
@@ -35,14 +34,13 @@ def get_block_header_by_hash(block_hash: Hash32, db: 'BaseChainDB') -> BlockHead
     return db.get_block_header_by_hash(block_hash)
 
 
-def apply_state_dict(account_db: BaseAccountDB, state_dict: AccountState) -> BaseAccountDB:
+def apply_state_dict(state: BaseState, state_dict: AccountState) -> None:
+    account_db = state.account_db
```
lithp: `eth/tools/fixtures/fillers/_utils.py::calc_state_root` also calls this function but hasn't been updated to use the new signature.

eth/vm/state.py
```diff
@@ -189,6 +204,9 @@ def commit(self, snapshot: Tuple[bytes, Tuple[UUID, UUID]]) -> None:
         _, checkpoint_id = snapshot
         self.account_db.commit(checkpoint_id)
 
+    def persist(self) -> None:
+        self.account_db.persist()
```
pipermerriam: Would it be accurate to say that nobody should be directly accessing the `state.account_db` (and thus calling methods on it?).  Maybe we should move it to `self._acccount_db`.  If there's a valid use case for pulling the db off the state object, we can add a `get_account_db()` so that people have to choose between:

- `state.persist()`
- `state._account_db.persist()`
- `state.get_account_db().persist()`

And the later two will stand out where-as `state.account_db.persist()`  would just blend in.

tests/core/opcodes/test_opcodes.py
```diff
@@ -618,8 +618,8 @@ def test_sstore(vm_class, code, gas_used, refund, original):
     computation = setup_computation(vm_class, CANONICAL_ADDRESS_B, decode_hex(code))
 
     computation.state.account_db.set_balance(CANONICAL_ADDRESS_B, 100000000000)
```
pipermerriam: With my comment above about moving to `state._account_db` , it *might* make sense to mirror these APIs as well...?

eth/vm/computation.py
```diff
@@ -113,8 +113,8 @@ class BaseComputation(Configurable, ABC):
     return_data = b''
     _error = None  # type: VMError
 
-    _log_entries = None  # type: List[Tuple[int, bytes, List[int], bytes]]
-    accounts_to_delete = None  # type: Dict[bytes, bytes]
+    _log_entries = None  # type: List[Tuple[int, Address, List[int], bytes]]
+    accounts_to_delete = None  # type: Dict[Address, Address]
```
lithp: Why did you have to change this? My guess is that the previous type was already wrong but somehow mypy didn't notice until you made this PR?

eth/chains/base.py
```diff
@@ -412,29 +411,27 @@ def from_genesis(cls,
         """
         genesis_vm_class = cls.get_vm_class_for_block_number(BlockNumber(0))
 
-        account_db = genesis_vm_class.get_state_class().get_account_db_class()(
-            base_db,
-            BLANK_ROOT_HASH,
-        )
+        pre_genesis_header = BlockHeader(difficulty=0, block_number=-1, gas_limit=0)
```
lithp: This seems a little weird to me, rather than making a fake object it might better to change `bulid_state`'s signature to accept a `state_root`, then you can pass the state root in directly.

eth/vm/state.py
carver: Yup, I think that could be good. It will make the PR significantly bigger, because of all the current calls to `state.account_db.get_balance()`, etc. But I'll give it a shot.

eth/vm/computation.py
carver: > Why did you have to change this?

mypy started complaining

> My guess is that the previous type was already wrong but somehow mypy didn't notice until you made this PR?

Yup, it can be hard to tell which things are our fault for misconfiguration, and which things are mypy bugs. So I tend not to dig in until it complains in a way that makes sense to me. I also read it as "mypy didn't notice before"

eth/chains/base.py
carver: Yeah, I do agree this is a little funky. But adding an explicit API where `ExecutionContext` is `None` is a serious footgun (any calls to `state.timestamp` etc will crash hard). Really the only scenario where it makes sense to generate state via a state root and no execution context is when you know there won't be any need to access state besides the database, which should only happen in the genesis block. I was a little hesitant to build a custom API method just for that.

I guess we could set `ExecutionContext` to `None` here in the genesis builder, and add a giant warning for other people not to copy that code elsewhere. I like that a tiny bit less than this, but a bit more than adding a footgun API.

Any other alternatives you see?

lithp: Ahh, somehow I missed that `header` was being used for more than just its state root. Sure, so we could do the `genesis_vm_class.get_state_class()(db, None, BLANK_ROOT_HASH)` thing, but that's bad form since state holds more than the account data.

Maybe the problem is that this function isn't really using a state! It's using a state which happens to wrap an AccountDB and that's why more objects are getting pulled in. Maybe this function alone could continue using AccountDB directly?

carver: Except it also gets and sets storage, which won't be in `account_db` as of #1735 :confused:

carver: At that point we are saying we want something like a `StateDB` that's distinct from `State` which really becomes something like the old `AccountDB` was...

carver: Or maybe the opposite, and we extract `ExecutionContext` out as something that's adjacent to `State` instead of embedded in it. Then if you want to execute, you run it through the `VM` which combines state with the execution context generated with the header in order to actually build a computation.

I'm starting to think something like:

- move `BaseState.execution_context` -> `BaseVM.execution_context`
- move `BaseState.get_computation` -> `BaseVM.get_computation` (which builds with its local execution context)
- rename `ExecutionContext` -> `BlockContext`
- Computation learns about `BlockContext`, like it knows about `TransactionContext`
- `BaseState` now only needs a db and state root to be generated

I'll capture this into an issue for follow-up since I don't think it belongs here. I will also note that this hacky code should be cleaned up when `State` is easier to initialize.

lithp: For future readers here's the issue: https://github.com/ethereum/py-evm/issues/1758

Issue #1752, opened 2019-04-18
Convert VM.block and .header to .get_block() and .get_header()


carver:
Remove the `@property` values for `vm.header` and `vm.block`, and replace with a `get_header()` and `get_block()`

```python
class VM:
    def get_header(self):
        ...
    def get_block(self):
        ...
```

Makes it a lot clearer that they have side effects.

_Originally posted by @pipermerriam in https://github.com/ethereum/py-evm/pull/1747_

carver:
This API is used in a lot of places, so it will take some work to hunt them all down. But there shouldn't be any subtleties. As long as the tests pass, we should be good. So this seems like a good first issue.

dherykw:
Hi,

I have this issue solved with all test passed, I also had to add set_block() function in order to remove the property properly.

My doubt is if you want me to create a new branch before to do the pull request or you want me to add the modification to a current branch.

pipermerriam:
New branch and new pull request please.

csurfer:
Looks like this issue is already [addressed and completed](https://github.com/ethereum/py-evm/blob/1af151ab218b905f4fdf7a285cbe14ebf094a7c4/eth/vm/base.py#L113-L123). Suggesting to close this issue for cleaner issue queue.

carver:
Thanks for taking a look!

PR #1761, opened 2019-04-22
[WIP] Make VM block and header properties real methods

jmcph4:
This commit changes the direct definitions in the `VM` class in
`eth/vm/base.py`.

### What was wrong?

See #1752.

### How was it fixed?

By changing the definitions in the `VM` class and updating any other modules dependent on that API.

#### Cute Animal Picture

Domestic cat kneading (or making biscuits)

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -399,22 +399,19 @@ class VM(BaseVM):
         self.chaindb = chaindb
         self._initial_header = header
 
-    @property
-    def header(self) -> BlockHeader:
+    def get_header(self) -> BlockHeader:
         if self._block is None:
             return self._initial_header
         else:
             return self._block.header
 
-    @property
-    def block(self) -> BaseBlock:
+    def get_block(self) -> BaseBlock:
         if self._block is None:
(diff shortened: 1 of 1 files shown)


eth/vm/base.py
```diff
@@ -399,22 +399,19 @@ def __init__(self, header: BlockHeader, chaindb: BaseChainDB) -> None:
...
-    def block(self) -> BaseBlock:
+    def get_block(self) -> BaseBlock:
         if self._block is None:
             block_class = self.get_block_class()
             self._block = block_class.from_header(header=self._initial_header, chaindb=self.chaindb)
         return self._block
 
-    @block.setter
-    def block(self, block: BaseBlock) -> None:
+    def set_block(self, block: BaseBlock) -> None:
```
pipermerriam: Noting that we probably want/need a test to ensure that `self._block.header` matches `self._initial_header` (which we currently are not enforcing so this would be a new test)

carver:
This got cleaned up elsewhere. Thanks for taking a look. Closing...

PR #1765, opened 2019-04-24
performance: avoid debug2 calls in high-traffic paths

carver:
A mainnet test showed a 15% `import_block()` speedup.

(re-)importing these blocks takes 20.0s without this change.
Took 17.0s with this change.

#### Cute Animal Picture


pipermerriam:
After writing that last snippet with the `ExtendedDebugLogger` itself caching whether debug2 should be logged I'm inclined to push for it since it would turn this PR into a much smaller diff with a very simple mechanism and prevent us from adding additional state and complexity to the EVM implementation.

carver:
About to have a call about it, but going in: the reason I didn't do it this way is that there can be non-trivial costs for generating the arguments to the call, which this other approach doesn't resolve.

carver:
On the call, I think we got close enough: we use something like django's `@cached_property` on the logger so that we don't add the overhead of a method call before every debug log.

carver:
@pipermerriam weird, I recognize some of these linting errors as things you've been fixing up. I have no idea why they just started showing up on my last force-push though.

carver:
Ah, I didn't have the latest eth-utils locally. Rather than trying to fix those in this PR, I'll just pin eth-utils down and then we can fix it in a separate one and unpin it.

PR #1766, opened 2019-04-24
Cache transaction.get_sender()

carver:
A mainnet test previously showed this method took 3.8% of the total
import_block() time. After this change it was 0.5% of the total.

#### Cute Animal Picture

### DIFF
--- a/eth/rlp/transactions.py
+++ b/eth/rlp/transactions.py
@@ -91,7 +91,7 @@ class BaseTransaction(BaseTransactionFields, BaseTransactionMethods):
     @cached_property
     def sender(self) -> Address:
         """
-        Convenience property for the return value of `get_sender`
+        Convenience and performance property for the return value of `get_sender`
         """
         return self.get_sender()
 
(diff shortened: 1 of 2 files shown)


eth/rlp/transactions.py
```diff
@@ -83,6 +83,8 @@ def hash(self) -> bytes:
 
 
 class BaseTransaction(BaseTransactionFields, BaseTransactionMethods):
+    _sender_cache = None
```
cburgdorf: If this is a simple backing field, maybe just calling it `_sender` would be more inline with the rest of the code? When I read `cache` I think of a bag of things such as a `Dict[K,V]` or so.

eth/rlp/transactions.py
```diff
@@ -92,7 +94,9 @@ def sender(self) -> Address:
         """
         Convenience property for the return value of `get_sender`
         """
-        return self.get_sender()
+        if self._sender_cache is None:
```
pipermerriam: Can we just go ahead and steal [`cached_property`](https://github.com/django/django/blob/7e6b214ed34f5562dbd83cf54924a5b589a29715/django/utils/functional.py#L16-L58) from django....  It's actually better than this since future lookups pull straight from `__dict__` and bypass the descriptor entirely.....

carver: Sounds good. I'll wait for #1765 to land, which adds `cached_property`

PR #1767, opened 2019-04-25
Migrate to mypy.ini file

cburgdorf:
### What was wrong?

Our `mypy` invocation was getting rather lengthy and hence discouraged adding further checks which is bad as we all love to be punished and bow to our `mypy` master as often as we can.

### How was it fixed?

- migrated to `mypy.ini` file
- added more checks and fixed some issues in the process

[//]: # (For important changes, please add a new entry to the release notes file)
[//]: # (See: https://trinity-client.readthedocs.io/en/latest/contributing.html#pull-requests)

#### Cute Animal Picture

### DIFF
--- a/eth/tools/_utils/normalization.py
+++ b/eth/tools/_utils/normalization.py
@@ -252,7 +252,7 @@ normalize_state = compose(
 )
 
 
-normalize_main_transaction = cast(Normalizer, dict_normalizer({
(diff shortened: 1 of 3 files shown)


cburgdorf:
This can wait and rebase once #1763 landed because then we can enable even more checks

pipermerriam:
probably warrants being contributed upstream to the project template.

PR #1769, opened 2019-04-30
Separate JournalDB's commit() and persist()

### LINKED ISSUE #1762: JournalDB.commit() should no longer "auto-persist" at the base commit
> I'm starting to think that `commit()` does the wrong thing. `commit()` should only ever collapse changesets together, and the entire `self.journal.is_empty()` block should move into `persist()`. That would be more straightforward.

_Originally posted by @lithp in https://github.com/ethereum/py-evm/pull/1735_

I agree, I think it's easier to understand that `persist` and `commit` are two separate concepts. Since external callers shouldn't ever end up with the base recorded checkpoint, there's not really any point in exposing the capability of persisting via commit.

So the base checkpoint would only ever be popped via `persist()`, and trying to commit the base checkpoint would be a `ValidationError`.

lithp:
A tiny refactor to JournalDB. The active part of `commit()` was only ever called by `persist()`, so that logic now lives in `persist()` where it will cause less confusion.

Closes #1762 

This leaves one hole open: If you call `commit()` with the root changeset you'll end up in a bad state. All your data is now lost and calling methods will probably give you exceptions until you call `JournalDB.reset()`. There are a few things we could do here:
1. Ignore the problem. Users shouldn't have access to `root_changeset_id` and in fact trinity never uses it.
2. Change `Journal.commit()` such that this is okay. If you're `Journal.commit()`ing the root changeset then squash changes like before but put them back into the Journal under a new root changeset.
3. Throw a `ValidationError` when the user calls `JournalDB.commit()` with the root changeset.

I'm kind of leaning toward 2 but which would you prefer?

#### Cute Animal Picture

tiny-turtle

### DIFF
--- a/eth/db/journal.py
+++ b/eth/db/journal.py
@@ -111,7 +111,7 @@ class Journal(BaseDB):
     def pop_changeset(self, changeset_id: uuid.UUID) -> Dict[bytes, Union[bytes, DeletedEntry]]:
         """
         Returns all changes from the given changeset.  This includes all of
-        the changes from any subsequent changeset, giving precidence to
+        the changes from any subsequent changeset, giving precedence to
         later changesets.
         """
         if changeset_id not in self.journal_data:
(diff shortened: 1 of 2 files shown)


carver:
> 2\. Change `Journal.commit()` such that this is okay. If you're `Journal.commit()`ing the root changeset then squash changes like before but put them back into the Journal under a new root changeset.
> 3\. Throw a `ValidationError` when the user calls `JournalDB.commit()` with the root changeset.
> 
> I'm kind of leaning toward 2 but which would you prefer?

Since we're never actually doing this, it seems easier to test & build # 3. We can add the feature later if we want it. (but seems unlikely)

lithp:
Added the validation error if you want to give it one last look before merging

eth/db/journal.py
```diff
@@ -381,31 +381,14 @@ def commit(self, changeset_id: uuid.UUID) -> None:
...
-                        pass
-                    else:
-                        self.wrapped_db[key] = cast(bytes, value)
-                except Exception:
-                    self._reapply_changeset_to_journal(changeset_id, journal_data)
-                    raise
+        if changeset_id == self.journal.root_changeset_id:
+            raise ValidationError(
+                "Tried to commit the root changeset. Callers should not keep references "
+                "to the root changeset"
```
carver: Maybe add a note about what to do instead of calling this. Something like: "You may want to call `persist()` instead."

PR #1770, opened 2019-05-02
Optimize code stream iteration

carver:
### What was wrong?

`CodeStream.__next__()` is called a lot, and is slow

A mainnet test previously showed this method took 5.0% of the total
import_block() time. After this change it was 2.6% of the total.

### How was it fixed?

Removed a couple attribute lookups

#### Cute Animal Picture

### DIFF
--- a/eth/vm/code_stream.py
+++ b/eth/vm/code_stream.py
@@ -39,11 +39,17 @@ class CodeStream(object):
     def __len__(self) -> int:
         return self._length_cache
 
-    def __iter__(self) -> 'CodeStream':
-        return self
-
     def __getitem__(self, i: int) -> int:
         return self._raw_code_bytes[i]
+
+    def __iter__(self) -> Iterator[int]:
+        # a very performance-sensitive method
+        read = self.read
+        try:
+            while True:
+                yield ord(read(1))
+        except TypeError:
+            yield STOP
 
     def __next__(self) -> int:
         # a very performance-sensitive method
--- a/eth/vm/code_stream.py
+++ b/eth/vm/code_stream.py
@@ -11,7 +11,7 @@ from eth.validation import (
 )
 from eth.vm import opcode_values
 
-PUSH1, PUSH32 = opcode_values.PUSH1, opcode_values.PUSH32
+PUSH1, PUSH32, STOP = opcode_values.PUSH1, opcode_values.PUSH32, opcode_values.STOP
 
 
 class CodeStream(object):
@@ -25,14 +25,16 @@ class CodeStream(object):
 
     def __init__(self, code_bytes: bytes) -> None:
         validate_is_bytes(code_bytes, title="CodeStream bytes")
-        self.stream = io.BytesIO(code_bytes)
+        stream = io.BytesIO(code_bytes)
+        self.stream = stream
+        self._bound_stream_read = stream.read
         self._raw_code_bytes = code_bytes
         self._length_cache = len(code_bytes)
         self.invalid_positions = set()
         self.valid_positions = set()
 
     def read(self, size: int) -> bytes:
-        return self.stream.read(size)
+        return self._bound_stream_read(size)
 
     def __len__(self) -> int:
         return self._length_cache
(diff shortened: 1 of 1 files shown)


eth/vm/code_stream.py
```diff
@@ -25,34 +25,34 @@ class CodeStream(object):
...
 
-    def __next__(self) -> int:
-        return self._next()
-
     def __getitem__(self, i: int) -> int:
         return self._raw_code_bytes[i]
 
-    def _next(self) -> int:
-        next_opcode_as_byte = self.read(1)
+    def __next__(self) -> int:
```
pipermerriam: Maybe instead of having `CodeStream` implement a `__next__` we could just do this in the `__iter__` and make it a generator function that does this in a loop.  That way we don't have to do the `self._bound_stream_read` property lookup and it could also be faster by not having to create a new call stack every time assuming generators are able to re-use it?

eth/vm/code_stream.py
```diff
@@ -25,34 +25,34 @@ class CodeStream(object):
...
+        # a very performance-sensitive method
+        next_opcode_as_byte = self._bound_stream_read(1)
 
         if next_opcode_as_byte:
```
pipermerriam: In theory if this was done as a generator function we could do something like this to get rid of this `if` statement.

```python
def __iter__(self) -> Generator[int, None, None]:
    read = self.stream.read
    try:
        while True:
            yield ord(read(1))
    except TypeError?  # whatever `ord(b'')` raises
        yield STOP
```

cburgdorf:
I skimmed over the benchmark to see if one can observe a speedup there but I think the benchmark runs are just too much influenced by other factors to be reliable for this kind of thing. In fact, the benchmark shows slightly worse numbers between https://circleci.com/gh/ethereum/py-evm/130728?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (this PR)

and https://circleci.com/gh/ethereum/py-evm/130792?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (previous merge in `master`) depending on which concrete benchmark you pick. 

However, on a macro level, if you zoom out a bit and compare benchmark results from now with those from Dec 2018 it seems to suggest that we have improved quite a lot.

https://circleci.com/gh/ethereum/py-evm/112490?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (old run)

carver:
> I skimmed over the benchmark to see if one can observe a speedup there but I think the benchmark runs are just too much influenced by other factors to be reliable for this kind of thing. In fact, the benchmark shows slightly worse numbers between https://circleci.com/gh/ethereum/py-evm/130728?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (this PR)
> 
> and https://circleci.com/gh/ethereum/py-evm/130792?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (previous merge in `master`) depending on which concrete benchmark you pick.
> 
> However, on a macro level, if you zoom out a bit and compare benchmark results from now with those from Dec 2018 it seems to suggest that we have improved quite a lot.
> 
> https://circleci.com/gh/ethereum/py-evm/112490?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link (old run)

Yeah, even on my local machine, running the same code twice can easily have a 2.5% total time variation. But the % that each method takes of the total run is often much more stable than the total time. So I've been focusing on driving down the percent of any particular method.

The confirmation of total time improvement doesn't really show up until we stacking a few of these different performance changes together, which overcomes the variation. Roughly, before I started in on these, the total `import_block()` run for the four blocks was something like 21-22s. Now, it's often clocking in at ~12s. I'm hoping to roughly break the 50% barrier (~10-11s) before moving onto other things again.

eth/vm/code_stream.py
carver: Huh, I'm quite surprised that a speedup is nearly undetectable from this change. It's within tolerance for variation from run to run (eg~ 2.6->2.4%). However, it's not worse, and I'm so convinced that it *should* be better that I'll go ahead and make the change.

pipermerriam: looks like this changed didn't make it in (just making a note of it)

carver: Ah, so `__next__` isn't getting called by the iterator anymore, but is still used by `.peek()` so I left it in.

A silly example to show that `__next__` isn't being used anymore:
```
In [1]: class IterMe:
   ...:     def __iter__(self):
   ...:         yield 1
   ...:         yield 2
   ...:     def __next__(self):
   ...:         return 'a'
   ...:     

In [3]: i = IterMe()

In [4]: next(i)
Out[4]: 'a'

In [5]: [v for v in i]
Out[5]: [1, 2]

pipermerriam: If it's only called by `peek` then how about in-lining it there?

PR #1771, opened 2019-05-02
Apply computation speedup

carver:
Shaved off 1.5% of total import_block() time.

### How was it fixed?

- inline some hot code
- avoid some attribute lookups
- appease mypy without an ignore or cast

#### Cute Animal Picture

### DIFF
--- a/eth/vm/computation.py
+++ b/eth/vm/computation.py
@@ -19,6 +19,7 @@ from eth_typing import (
 )
 from eth_utils import (
     encode_hex,
+    ValidationError,
 )
 
 from eth.constants import (
@@ -74,7 +75,13 @@ from eth.vm.transaction_context import (
     BaseTransactionContext
 )
 
-NO_RESULT = object()
+
+def NO_RESULT(computation: 'BaseComputation') -> None:
+    """
+    This is a special method intended for usage as the "no precompile found" result.
+    The type signature is designed to match the other precompiles.
+    """
+    raise ValidationError("This method is never intended to be executed")
 
 
 def memory_gas_cost(size_in_bytes: int) -> int:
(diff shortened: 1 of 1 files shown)


eth/vm/computation.py
```diff
@@ -568,14 +577,19 @@ def apply_computation(cls,
...
+                precompile(computation)
                 return computation
 
             show_debug2 = computation.logger.show_debug2
 
+            opcode_lookup = computation.opcodes
             for opcode in computation.code:
-                opcode_fn = computation.get_opcode_fn(opcode)
+                try:
+                    opcode_fn = opcode_lookup[opcode]
```
pipermerriam: Wondering if you checked if this is faster/slower `opcode_lookup.get(opcode, INVALID)`?  Where `INVALID` uses the `computation.code_stream` to lookup what the `opcode` is at the current code position?

Not needed for this pr, but just stood out to me as *maybe* another similar styled speedup.

eth/vm/computation.py
```diff
@@ -75,6 +76,14 @@
 )
 
 
+def NO_RESULT(computation: 'BaseComputation') -> None:
+    """
+    This is a special method intended for usage as the "no precompile found" result.
+    The type signature is designed to match the other precompiles.
+    """
+    raise ValidationError("This method is never intended to be executed")
```
pipermerriam: Maybe just use `Exception` here since we shouldn't touch this code path?

eth/vm/computation.py
carver: Hm, yeah I don't remember exactly, but it sounds like an alternative I tried and rejected. I think `get` had more overhead that `__getitem__` and the exception rarely happens, plus the `try` is cheap to run in the happy case. I'll see if I can find anything in a quick re-run.

carver: Tried a couple variations without getting any speedup

PR #1773, opened 2019-05-03
EVM stack operation speedup, ~9% speedup

carver:
```
    EVM stack push, pop, and dup ~8.9% speedup
    
    Percent of total import_block() time
    ====================================
    
    In eth/vm/stack.py:
    
          old      new     saved
    push  5.99% -> 3.69% = 2.30%
    pop   8.63% -> 4.90% = 3.73%
    dup   1.76% -> 0.69% = 1.07%
    ----------------------------------
    Total lift             7.1%
    
    In eth/vm/computation.py:
    
                old      new    saved
    stack_push  0.98% -> ~0% =  0.98%
    stack_pop   0.87% -> ~0% =  0.87%
    ----------------------------------
    Total lift                  1.85%
    
    Tested import_block() against mainnet blocks:
    (7620447, 7620450, 7620453, 7620454)
```

(Note: performance measured on top of #1777)

### How was it fixed?

There are a few important changes here:
- switch push and pop to use explicit type signatures, which avoids a lot of dynamic checks
  - split out a single pop from multiple popped values
  - split out int/bytes/any into separate methods
- skip some explicit type checks, count on mypy to catch any issues
- store the value type explicitly to do a much faster `is` check (rather than calling `isinstance()` so much)
- ~~push was causing a lot of `CodeStream` reads that were slow. Skipped the method overhead and attribute lookup by exposing the internal `read`.~~

This seemed to surface some typing issues that I went ahead and fixed up (one of the method signatures had `Any` as a return value, for example -- with that removed, mypy spit some warnings).

TODO
- [x] drop `CodeStream` changes and rebase on #1777
- [x] re-measure performance lift
- [x] squash
- [x] more docs (some are probably stale now)
- [x] ~~(maybe) update some more of the opcode tests to use the `VM` fixture instead of explicitly listing vms.~~ (or maybe not if this PR is already feeling too big)
- [x] resolve `CodeStream` conflicts with #1770

#### Cute Animal Picture

### DIFF
--- a/eth/validation.py
+++ b/eth/validation.py
@@ -190,14 +190,21 @@ def validate_uint256(value: int, title: str="Value") -> None:
         )
 
 
-def validate_stack_item(value: Union[int, bytes]) -> None:
-    if isinstance(value, bytes) and len(value) <= 32:
-        return
-    elif isinstance(value, int) and 0 <= value <= UINT_256_MAX:
+def validate_stack_int(value: int) -> None:
+    if 0 <= value <= UINT_256_MAX:
         return
     raise ValidationError(
         "Invalid Stack Item: Must be either a length 32 byte "
-        "string or a 256 bit integer. Got {0}".format(value)
+        "string or a 256 bit integer. Got {!r}".format(value)
+    )
+
+
+def validate_stack_bytes(value: bytes) -> None:
+    if len(value) <= 32:
+        return
+    raise ValidationError(
+        "Invalid Stack Item: Must be either a length 32 byte "
+        "string or a 256 bit integer. Got {!r}".format(value)
     )
 
 
--- a/eth/vm/computation.py
+++ b/eth/vm/computation.py
@@ -1,6 +1,6 @@
 from abc import (
     ABC,
-    abstractmethod
+    abstractmethod,
 )
 import itertools
 import logging
@@ -14,6 +14,7 @@ from typing import (  # noqa: F401
     Union,
 )
 
+from cached_property import cached_property
 from eth_typing import (
     Address,
 )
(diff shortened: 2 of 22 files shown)


eth/validation.py
```diff
@@ -190,14 +190,21 @@ def validate_uint256(value: int, title: str="Value") -> None:
...
+    if 0 <= value <= UINT_256_MAX:
         return
-    elif isinstance(value, int) and 0 <= value <= UINT_256_MAX:
+    raise ValidationError(
+        "Invalid Stack Item: Must be either a length 32 byte "
+        "string or a 256 bit integer. Got {!r}".format(value)
+    )
+
+
+def validate_stack_bytes(value: bytes) -> None:
```
pipermerriam: I'd advocate for going ahead and dropping support for heterogenous stack and update so everything uses an `int` but that doesn't have to happen here, but I suspect there's more easy performance to be gained there.

eth/vm/computation.py
```diff
@@ -92,7 +92,86 @@ def memory_gas_cost(size_in_bytes: int) -> int:
...
+        """
+        raise Exception("Must be overridden by subclass, maybe at init time")
+
+    def stack_pop1_int(self) -> int:
+        """
+        Pop and return an integer from the stack.
+
+        Raise `eth.exceptions.InsufficientStack` if the stack was empty.
+        """
+        raise Exception("Must be overridden by subclass, maybe at init time")
```
pipermerriam: Why this instead of using an `ABC` + `abstractmethod`?

eth/vm/computation.py
```diff
@@ -150,6 +229,17 @@ def __init__(self,
         code = message.code
         self.code = CodeStream(code)
 
+        # Stack manipulation is performance-sensitive code
+        # Avoid method call overhead by assigning stack method directly to object
+        self.stack_push_int = self._stack.push_int  # type: ignore # method assignment unsupported
```
pipermerriam: Can you make these `cached_property` instaed so that it's lazy and fast.  That potentially bypasses the `type: ignore` as well.

eth/vm/stack.py
```diff
@@ -22,69 +23,213 @@
...
+    def pop_ints(self, num_items: int) -> Tuple[int, ...]:
+        """
+        Pop and return a tuple of integers of length ``num_items`` from the stack.
+
+        Raise `eth.exceptions.InsufficientStack` if there are not enough items on
+        the stack.
+
+        Items are ordered with the top of the stack as the first item in the tuple.
+
+        Note: This function is optimized for speed over readability.
```
pipermerriam: Maybe the function could have more comments in it to explain what is hppening and why.  Also, This type of comment should be very visable like:

```
def thing(...):
    """
    ###################################
    #           !!!!! STOP !!!!!      #
    ###################################
```

Something hard to miss when someone decides to edit a line or two.

eth/vm/stack.py
```diff
@@ -22,69 +23,213 @@
...
-            else:
-                raise TypeError(
-                    "Unknown type_hint: {0}.  Must be one of {1}".format(
-                        type_hint,
-                        ", ".join((constants.UINT256, constants.BYTES)),
-                    )
-                )
+                    raise _busted_type(item_type, popped)
+
+            return tuple(type_cast_popped)
```
pipermerriam: Is this actually faster than `to_tuple` and yield?

tests/core/stack/test_stack.py
```diff
@@ -63,59 +71,79 @@ def test_dup_does_not_allow_stack_to_exceed_1024_items(stack):
...
     (
-        (1, UINT256, int, True),
-        (b'101', BYTES, bytes, True),
-        (1, SECPK1_N, int, False)
+        (1, 'push_int', 'pop_ints', (1, )),
+        (1, 'push_int', 'pop_any', (1, )),
+        (1, 'push_int', 'pop_bytes', (b'\x01', )),
+        (1, 'push_int', 'pop1_int', 1),
+        (1, 'push_int', 'pop1_any', 1),
+        (1, 'push_int', 'pop1_bytes', b'\x01'),
```
pipermerriam: I don't know how I feel about this, but I realized you could do this as:


```
@pytest.parametrize(
    "method",
    (
        SomeClass.method_a,
        SomeClass.method_b,
    ),
)
de test_some_methods_on_a_class(method):
    instance = SomeClass(...)
    result = method(instance, ...)
    assert result ...
```

Leverage providing the `self` argument and call the method on the class itself.

eth/validation.py
carver: Hah, so that's a tangent I went on and ended up dropping. I got some early indications that the extra cost of converting and de-converting the elements when both the push and pop are the same type (say, both `bytes`) is meaningful.

eth/vm/computation.py
carver: Because `abstractmethod` doesn't recognize methods overridden in `__init__` by assignment.

eth/vm/computation.py
carver: Interesting, so a cached property that is an underlying method. Sure, I don't see any problems :+1: .

eth/vm/stack.py
carver: Yup

eth/vm/computation.py
carver: Yup, I'm happy with this. It's still a tiny bit surprising, but much less than assigning the method in init.

tests/core/stack/test_stack.py
carver: Eh, seems about the same to me. The slight gain in avoiding the `getattr()` is offset by the longer parameterization boilerplate. Not opposed, but probably not worth doing the change now.

eth/validation.py
pipermerriam: Maybe you could code comment that in the main class so that it's documented that a heterogeneous stack is more performant.

tests/core/stack/test_stack.py
pipermerriam: Wasn't really advocating for it as much as realizing it was possible.  :+1:

PR #1776, opened 2019-05-07
Faster JournalDB committing

### LINKED ISSUE #1775: Optimize JournalDB More
### What is wrong?

My latest mainnet test shows `JournalDB` being a major contributor to `import_block()` time:

- `.commit()` taking 13.5% of total time
- `.__getitem__` taking another >2% of time
- `.latest` and `.latest_id` ~1.1%
- `.record()` ~0.85%

A few other methods that are in the 0.5-1% range, but add up to another couple of percent. It looks like a major reason for each of them is `uuid` which is spending 4.3% of total blocktime in `__hash__` and another 2.7% in `__eq__`.

### How can it be fixed

- Look into uuid alternatives
- refactor the data store to always keep the latest data and have the changesets be backwards-facing diffs (since gets get is called 17k times, commit called 43k times, and `discard` only called 274 times, we should optimize for the "happy path" where everything gets committed).

I think this will wrap up the `import_block()` performance sprint for now (`HexaryTrie` is the next major contributor, but that will take some serious new features, and I think it's fast enough to make Beam Sync progress)

carver:
### What was wrong?

JournalDB was taking a lot of time to handle the checkpoint/rollback machinery. ~8.3% of the import block timing

Fixes #1775

### How was it fixed?

- Switch from uuid -> int checkpoints
- Switch underlying data structure to a single "latest" data, with dicts for rollback
- Use `__slots__` for faster attribute and method lookups
- Move some validation down from `JournalDB` into `Journal`, where it can often validate almost for free
- Some targeted in-lining
- Don't re-initialize `Journal`, just reset it, which enables...
- Pass through a couple method calls direct from `JournalDB` to `Journal`
(Unfortunately, the neat `@cached_property` trick doesn't work with slots, so this goes back to the `self.method = self.journal.method` approach)

So far, total time spent in journal dropped to 1.4% of the run.

This 6.9% speedup *seems* to be under-reporting the total speedup, which looks more like a 20% speedup based on eyeballing total runtime. But the variance is too high to really get a feel, so I'll use the more conservative number.

TODO
- [x] lint
- [x] some refactors
- [x] maybe some more low-hanging fruit
- [x] hypothesis test against old `JournalDB`
- [x] if ^ doesn't work, hunt down consensus failure manually
- [x] commit history cleanup

#### Cute Animal Picture

### DIFF
--- a/eth/db/journal.py
+++ b/eth/db/journal.py
@@ -54,29 +54,30 @@ class Journal(BaseDB):
 
     Changesets are referenced by an internally-generated integer.
     """
+    __slots__ = ['_journal_data', '_clears_at', '_current_values', '_ignore_wrapped_db']
 
     def __init__(self) -> None:
         # contains a mapping from all of the `uuid4` changeset_ids
         # to a dictionary of key:value pairs that describe how to rewind from the current values
         # to the given checkpoint
-        self.journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, Union[bytes, DeletedEntry]]]  # noqa E501
+        self._journal_data = collections.OrderedDict()  # type: collections.OrderedDict[uuid.UUID, Dict[bytes, Union[bytes, DeletedEntry]]]  # noqa E501
         self._clears_at = set()  # type: Set[uuid.UUID]
         self._current_values = {}  # Dict[bytes, Union[bytes, DeletedEntry]]
-        self._is_cleared = False
+        self._ignore_wrapped_db = False  # if there is a clear journaled, then treat wrapped DB as empty
 
     @property
     def root_changeset_id(self) -> uuid.UUID:
         """
         Returns the id of the root changeset
         """
-        return first(self.journal_data.keys())
+        return first(self._journal_data.keys())
 
     @property
     def is_flattened(self) -> bool:
         """
         :return: whether there are any explicitly committed checkpoints
         """
-        return len(self.journal_data) < 2
+        return len(self._journal_data) < 2
 
     @property
     def latest_id(self) -> uuid.UUID:
@@ -85,27 +86,27 @@ class Journal(BaseDB):
         """
         # last() was iterating through all values, so first(reversed()) gives a 12.5x speedup
         # Interestingly, an attempt to cache this value caused a slowdown.
-        return first(reversed(self.journal_data.keys()))
+        return first(reversed(self._journal_data.keys()))
 
     @property
     def latest(self) -> Dict[bytes, Union[bytes, DeletedEntry]]:
         """
         Returns the dictionary of db keys and values for the latest changeset.
         """
-        return self.journal_data[self.latest_id]
+        return self._journal_data[self.latest_id]
 
     @latest.setter
     def latest(self, value: Dict[bytes, Union[bytes, DeletedEntry]]) -> None:
         """
         Setter for updating the *latest* changeset.
         """
-        self.journal_data[self.latest_id] = value
+        self._journal_data[self.latest_id] = value
 
     def is_empty(self) -> bool:
-        return len(self.journal_data) == 0
+        return len(self._journal_data) == 0
 
     def has_changeset(self, changeset_id: uuid.UUID) -> bool:
-        return changeset_id in self.journal_data
+        return changeset_id in self._journal_data
 
     def record_changeset(self, custom_changeset_id: uuid.UUID = None) -> uuid.UUID:
         """
@@ -113,7 +114,7 @@ class Journal(BaseDB):
         to prevent collisions between multiple changesets.
         """
         if custom_changeset_id is not None:
-            if custom_changeset_id in self.journal_data:
+            if custom_changeset_id in self._journal_data:
                 raise ValidationError(
                     "Tried to record with an existing changeset id: %r" % custom_changeset_id
                 )
@@ -122,7 +123,7 @@ class Journal(BaseDB):
         else:
             changeset_id = next(id_generator)
 
-        self.journal_data[changeset_id] = {}
+        self._journal_data[changeset_id] = {}
         return changeset_id
 
     def discard(self, discard_through_checkpoint_id):
@@ -133,19 +134,19 @@ class Journal(BaseDB):
                 else:
                     self._current_values[old_key] = old_value
 
-            del self.journal_data[checkpoint_id]
+            del self._journal_data[checkpoint_id]
 
             if checkpoint_id in self._clears_at:
                 self._clears_at.remove(checkpoint_id)
-                self._is_cleared = False
+                self._ignore_wrapped_db = False
 
         if self._clears_at:
             # if there is still a clear in older locations, then reinitiate the clear flag
-            self._is_cleared = True
+            self._ignore_wrapped_db = True
 
     @to_tuple
     def _rollbacks_through(self, through_checkpoint_id):
-        for checkpoint_id, rollback_data in reversed(self.journal_data.items()):
+        for checkpoint_id, rollback_data in reversed(self._journal_data.items()):
             yield checkpoint_id, rollback_data
             if checkpoint_id == through_checkpoint_id:
                 break
@@ -156,7 +157,7 @@ class Journal(BaseDB):
     def _drop_rollbacks(self, drop_through_checkpoint_id: uuid.UUID) -> Dict[bytes, Union[bytes, DeletedEntry]]:
         had_clear = False
         for checkpoint_id, _ in self._rollbacks_through(drop_through_checkpoint_id):
-            del self.journal_data[checkpoint_id]
+            del self._journal_data[checkpoint_id]
 
             if checkpoint_id in self._clears_at:
                 self._clears_at.remove(checkpoint_id)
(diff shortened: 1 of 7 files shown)


eth/db/journal.py
```diff
@@ -32,21 +36,33 @@ class DeletedEntry:
...
 class Journal(BaseDB):
     """
     A Journal is an ordered list of changesets.  A changeset is a dictionary
-    of database keys and values.  The values are tracked changes that were
-    written after the changeset was created
+    of database keys and values.  The values are tracked changes which give
+    the information needed to revert a changeset to an old checkpoint.
 
-    Changesets are referenced by a random uuid4.
+    Changesets are referenced by an internally-generated integer.
```
cburgdorf: Maybe even explicitly mention that the `Journal` isn't thread safe.

eth/db/journal.py
```diff
@@ -24,237 +26,230 @@ class DeletedEntry:
...
 # 1. key not in wrapped
 # 2. key created in journal
 # 3. key deleted
-ERASE_CREATED_ENTRY = DeletedEntry()
+REVERT_TO_WRAPPED = DeletedEntry()
+
+ChangesetValue = Union[bytes, DeletedEntry]
+ChangesetDict = Dict[bytes, ChangesetValue]
+
+id_generator = count()
```
pipermerriam: Maybe `get_checkpoint_id = count().__next__`?  Might look a little more natural in the places where it's used.

eth/db/journal.py
```diff
@@ -24,237 +26,230 @@ class DeletedEntry:
...
+            checkpoint_id, rollback_data = self._journal_data.popitem()
+
+            for old_key, old_value in rollback_data.items():
+                if old_value is REVERT_TO_WRAPPED:
+                    # The current value may not exist, if it was a delete followed by a clear,
+                    # so pop it off, or ignore if it is already missing
+                    self._current_values.pop(old_key, None)
+                elif old_value is DELETE_WRAPPED:
+                    self._current_values[old_key] = old_value
+                else:
```
pipermerriam: implicit else?  I recognize this is performance code so no big deal.  Might still be worth doing `elif type(old_value) is bytes` which should be decently quick (maybe as-fast as the `cast` call which might be able to be done away with if `mypy` is smart enough.

eth/db/typing.py
```diff
@@ -0,0 +1 @@
+JournalDBCheckpoint = int
```
pipermerriam: Shouldn't  this use [`typing.NewType`](https://docs.python.org/3/library/typing.html#newtype)?  My impression is that is the *correct* way to do this.

cburgdorf: `JournalDBCheckpoint = int` is just an alias usually used to get a more descriptive name to a type that would otherwise be cumbersome to use repetitively. From the perspective of the type system the alias is identical to the type it is aliased to.

On the other hand [`typing.NewType`](https://docs.python.org/3/library/typing.html#newtype), really creates a different kind of type so that the type system can enforce it isn't mixed up with a different type somewhere else. In other words `JournalDBCheckpoint` would become entirely different from `int` and hence can `JournalDBCheckpoint(1) + 1` would be rejected from the type system.

I guess a [`typing.NewType`](https://docs.python.org/3/library/typing.html#newtype) was meant to be used here.

eth/db/journal.py
carver: Discard is used rarely enough that I'm good adding this validation back in. The `diff()` is way too sensitive, so I'll leave it out there.

tests/database/test_journal_db.py
```diff
@@ -548,21 +561,189 @@ def test_journal_persist_set_KeyError_then_persist():
...
+    journal_db[1] = b'to-be-removed'
+
+    journal_db.commit(commit_to)
+    assert journal_db[1] == b'to-be-removed'
+
+    journal_db.discard(discard_to)
+    assert 1 not in journal_db
+
+
+class JournalComparison(RuleBasedStateMachine):
```
carver: Thanks @lithp for tipping me off about this construction. It saved me from having to debug a mystery bad state root from the consensus tests.

eth/db/journal.py
```diff
@@ -24,237 +26,258 @@ class DeletedEntry:
...
 
-    Changesets are referenced by a random uuid4.
+    Changesets are referenced by an internally-generated integer. This is *not* threadsafe.
     """
+    __slots__ = [
+        '_journal_data',
+        '_clears_at',
+        '_current_values',
+        '_ignore_wrapped_db',
+        '_checkpoint_stack',
```
cburgdorf: Not for this PR but maybe worth checking out something like https://github.com/cjrh/autoslot
If we use `__slots__` at more places I would feel increasingly uneasy about having to keep this "stringly typed" array in sync with the actual attributes.

eth/db/journal.py
```diff
@@ -24,237 +26,258 @@ class DeletedEntry:
...
-            if custom_changeset_id in self.journal_data:
+            if custom_changeset_id in self._journal_data:
                 raise ValidationError(
                     "Tried to record with an existing changeset id: %r" % custom_changeset_id
                 )
             else:
                 changeset_id = custom_changeset_id
         else:
-            changeset_id = uuid.uuid4()
+            changeset_id = get_next_checkpoint_id()
```
cburgdorf: We used to call the recorded changes `changeset` and the ids referencing them `changeset_id`. The new number generator calls them `checkpoint_id` which introduces a slight inconsistency of terminology. I think we should make the following renames:

`get_next_checkpoint_id` -> `get_next_changeset_id`
`JournalDBCheckpoint` -> `JournalChangesetID` (notice the dropped `DB` because the class is also called `Journal` not `JournalDB`

We could also go the other way around and use `checkpoint` everywhere but I think I have a slight preference for `changeset` and `changeset_id` because it maps better to the actual concepts in my head. A `changeset`  (and `changeset_id`) implies that we are holding a *set*  of *changes* that we can reference by some assigned *id*  which is exactly what is happening.

eth/db/journal.py
```diff
@@ -24,237 +26,258 @@ class DeletedEntry:
...
 
-        self.journal_data[changeset_id] = {}
+        self._journal_data[changeset_id] = {}
+        self._checkpoint_stack.append(changeset_id)
```
cburgdorf: Just to back my previous comment, here we are appending a `changeset_id` to the `checkpoint_stack`

eth/db/journal.py
```diff
@@ -24,237 +26,258 @@ class DeletedEntry:
...
-
-        all_ids = tuple(self.journal_data.keys())
-        changeset_idx = all_ids.index(changeset_id)
-        changesets_to_pop = all_ids[changeset_idx:]
-        popped_clears = tuple(idx for idx in changesets_to_pop if idx in self._clears_at)
-        if popped_clears:
-            last_clear_idx = changesets_to_pop.index(popped_clears[-1])
-            changesets_to_drop = changesets_to_pop[:last_clear_idx]
-            changesets_to_merge = changesets_to_pop[last_clear_idx:]
+    def discard(self, through_checkpoint_id: JournalDBCheckpoint) -> None:
```
cburgdorf: To back it further, this method talks about `checkpoint_id` what is elsewhere referred to as `changeset_id`

tests/database/test_journal_db.py
```diff
@@ -1,14 +1,15 @@
...
     given,
+    settings,
     strategies as st,
 )
+from hypothesis.stateful import Bundle, RuleBasedStateMachine, rule
 import pytest
 
 from eth.db.backends.memory import MemoryDB
 from eth.db.journal import JournalDB
+from eth.db.slow_journal import JournalDB as SlowJournalDB
```
cburgdorf: I see that you are using some hypothesis magic to test the new implementation against the old one but I'm still wondering...what's the intended process here, do we plan to keep `eth.db.slow_journal` as a reference point or was this just something used for debugging that can now be dropped? :thinking:

eth/db/journal.py
carver: Okay, onboard with getting the naming unified, but with the opposite preference. Here's why checkpoint makes more sense to me:
- `checkpoint` is already an ID. So we are comparing `checkpoint` to `changeset_id`
- there are other places (like tests) that the name is shortened to `changeset` (presumably because it's shorter/easier to type)
- Dropping from `changeset_id` to `changeset` dramatically changes the meaning. I expect a `changeset` to be something like a `DBDiff`, a set of all the changes (between two checkpoints).
- committing or discarding a changeset (ID) seems like we're only committing or discarding the changes that happened in that specific changeset
- committing a checkpoint is more intuitively committing everything that happened since the checkpoint.
- put another way: it's like the changeset for every single changeset ID is changing every time you add or delete to the journal (committing the first changeset has a different meaning after every touch of the journal)

Re: dropping `DB`, all of the classes that refer to the checkpoint (like `VMState`, `AccountDB`, etc) are using the `JournalDB` class and theoretically don't know/care that the `Journal` implementation exists. Of course I get that it's strange that `Journal` would reference `JournalDB`. I guess theoretically there should be two checkpoint classes, one that would only be used internally. That's kind of more babysitting of mypy than I feel like doing. I've come close to dropping `Journal` altogether and inlining it to `JournalDB` since we're not using `Journal` elsewhere and it's so performance sensitive. So I really didn't feel like wrangling mypy if that's where we're headed.

tests/database/test_journal_db.py
carver: Yeah, I am sort of imagining keeping it around until a little after the release, and then dropping it. Just wanted to make sure there aren't any more hidden bugs in the new one.

eth/db/journal.py
cburgdorf: I just skimmed over this really fast, so no in-depth reply here. Just wanted to signal that I don't  have a strong preference here so whatever resolves the inconsistency is :+1: for me.

carver:
Ok, I'll go ahead and merge because as I wrote it out I only got more confident about the name change to `checkpoint`. I kept the name change as a separate commit 8b3bd4a -- so reverting the name change will be easy if we decide to.

PR #1777, opened 2019-05-14
Drop BytesIO from CodeStream, ~7.5% import block speedup

carver:
### What was wrong?

While inlining `CodeStream.__next __()`, it started to make sense to drop BytesIO altogether. We were basically doing all the same work anyway, to avoid any slowdowns caused by dispatching to CodeStream.

### How was it fixed?

- Use slots to speed up attribute/method access
- Drop `io.BytesIO` and just keep a `pc` and bytes buffer explicitly
- Added `SlowCodeStream` as a reference implementation for testing, until this new one "burns in" for a while
- (Bonus) try to skip the `ljust` in `push_xx` opcodes because it's typically not needed.

All added to a nice speed bump:

```
__iter__  4.53% -> 1.58% = 2.95%
read      2.31% -> 0.56% = 1.75%
pc()      1.12% -> 0%    = 1.12%
peek      1.07% -> 0.09% = 0.98%

push_xx   5.40% -> 4.57% = 0.83%
--------------------------------
Total lift                 7.6%
```

TODO:
- [x] see how bad it would be to drop the jump-dest `min()` (it was proactive, since pc is not being explicitly fenced anymore)
- [x] add notes about why performance decisions were made
- [x] squash
- [x] More explicit commit message about speedup

#### Cute Animal Picture


eth/vm/code_stream.py
```diff
@@ -14,27 +13,27 @@
...
 
     def read(self, size: int) -> bytes:
-        return self._bound_stream_read(size)
+        old_pc = self.pc
+        target_pc = old_pc + size
+        if target_pc >= self._length_cache:
+            self.pc = self._length_cache
+        else:
+            self.pc = target_pc
+        return self._raw_code_bytes[old_pc:target_pc]
```
pipermerriam: Did you try using a `memoryview` here to avoid the copy?

carver: Nice, will try

carver: Wow, quite surprised that it doesn't speed things up. I stumbled on a 0.85% speedup in `push_XX` though (since switching to a `memoryview` broke things in there). Pushing...

eth/vm/logic/stack.py
```diff
@@ -11,9 +11,9 @@ def pop(computation: BaseComputation) -> None:
 
 def push_XX(computation: BaseComputation, size: int) -> None:
     raw_value = computation.code.read(size)
-
-    if not raw_value.strip(b'\x00'):
```
carver: This strip was spending time to on a conversion that we might never need. Just let 0-bytes get pushed into the stack.

eth/vm/logic/stack.py
```diff
@@ -11,9 +11,9 @@ def pop(computation: BaseComputation) -> None:
...
-        computation.stack_push(0)
+    raw_len = len(raw_value)
+    if raw_len == size:
```
carver: Calling `len()` is shorter than calling an `ljust()` that takes no action (according to profiling), so we try to avoid the `ljust()`. (In fact, the `import_block()` calls in the four mainnet blocks that I'm testing *never* call `ljust()`, over the course of 154k push opcodes).

carver:
@pipermerriam okay, actually ready for review now

PR #1778, opened 2019-05-15
Exception with more context when trie data missing

carver:
### What was wrong?

Beam Sync involves intentionally executing transactions without the full state, including some state that might be necessary to execute. There are currently issues when we fail out of execution in the middle with `MissingTrieNode` (like leaving behind partial state). Also, Beam Sync is if we only learn about the one particular missing trie node. Including data about which account and storage slot are requested speeds things up considerably.

### How was it fixed?

- Add new exceptions in `eth.vm.interrupt`. They are intentionally not in `eth.exceptions` because they are aware of (and import from) other `eth` modules. These are full-blown classes with references back to data, rather than simple exceptions.
  - `MissingAccountTrieNode`
  - `MissingStorageTrieNode`
  - `MissingBytecode`
- Add `HashTrie.squash_changes()`, which proxies through to `HexaryTrie` (used for atomicity of trie update)
- Bonus minor optimization: only load parent header for validation, instead of full parent block
- More context about transaction failures due to being unable to afford gas
- More context about failures during PoW verification

TODO:
- [x] hunt down consensus error
- [x] drop state root from `MissingStorageTrieNode`, since `StorageLookup` no longer has access to `AccountDB`
- [x] pass the account address instead of the hash of the address, in the interrupt exceptions
- [x] ~~stop silently swallowing `KeyError` in `DBDiff`~~ issue #1779 
- [x] ~~add issue to rlp to have a default `rlp.Serializeable.__repr__` that does something reasonable~~ -- opened PR https://github.com/ethereum/pyrlp/pull/117
- [x] type annotations
- [x] test exceptions are raised with the expected values when a trie node is missing
- [x] squash
- [x] ~~document journal batch~~ *Removed it instead*
- [x] document and remove warning when handling `MissingTrieNode` due to adding a new account
- [x] remove TODO that's superseded by new issue
- [x] fix typo in MissingAccountTrieNode docs
- [ ] re-squash

#### Cute Animal Picture

### DIFF
--- a/eth/db/diff.py
+++ b/eth/db/diff.py
@@ -199,7 +199,6 @@ class DBDiff(ABC_Mapping):
                     except EVMMissingData:
                         raise
                     except KeyError:
-                        # TODO stop silently swallowing this key error
                         pass
                 else:
                     pass
--- a/eth/db/diff.py
+++ b/eth/db/diff.py
@@ -17,6 +17,7 @@ from eth_utils import (
 )
 
 from eth.db.backends.base import BaseDB
+from eth.vm.interrupt import EVMMissingData
 
 if TYPE_CHECKING:
     ABC_Mutable_Mapping = MutableMapping[bytes, Union[bytes, 'MissingReason']]
@@ -195,7 +196,10 @@ class DBDiff(ABC_Mapping):
                 if apply_deletes:
                     try:
                         del db[key]
+                    except EVMMissingData:
+                        raise
                     except KeyError:
+                        # TODO stop silently swallowing this key error
                         pass
                 else:
                     pass
--- a/eth/db/hash_trie.py
+++ b/eth/db/hash_trie.py
@@ -1,4 +1,11 @@
+import contextlib
+from typing import (
+    cast,
+    Iterator,
+)
+
 from eth_hash.auto import keccak
+from trie import HexaryTrie
 
 from eth.db.keymap import (
     KeyMapDB,
@@ -7,3 +14,8 @@ from eth.db.keymap import (
 
 class HashTrie(KeyMapDB):
     keymap = keccak
+
+    @contextlib.contextmanager
+    def squash_changes(self) -> Iterator['HashTrie']:
+        with cast(HexaryTrie, self._db).squash_changes() as memory_trie:
+            yield type(self)(memory_trie)
--- a/eth/db/storage.py
+++ b/eth/db/storage.py
@@ -41,6 +41,9 @@ from eth.db.journal import (
 from eth.db.typing import (
     JournalDBCheckpoint,
 )
+from eth.vm.interrupt import (
+    MissingStorageTrieNode,
+)
 from eth.tools.logging import (
     ExtendedDebugLogger
 )
@@ -91,9 +94,12 @@ class StorageLookup(BaseDB):
         try:
             return read_trie[hashed_slot]
         except trie_exceptions.MissingTrieNode as exc:
-            self.logger.warning(exc)
-            self.logger.debug("expected to get slot, but:", exc_info=True)
-            raise exc
+            raise MissingStorageTrieNode(
+                exc.missing_node_hash,
+                self._starting_root_hash,
+                exc.requested_key,
+                self._address,
+            ) from exc
 
     def __setitem__(self, key: bytes, value: bytes) -> None:
         hashed_slot = self._decode_key(key)
@@ -112,8 +118,12 @@ class StorageLookup(BaseDB):
         try:
             del write_trie[hashed_slot]
         except trie_exceptions.MissingTrieNode as exc:
-            self.logger.warning(exc)
-            self.logger.debug("expected to be able to delete slot, but:", exc_info=True)
+            raise MissingStorageTrieNode(
+                exc.missing_node_hash,
+                self._starting_root_hash,
+                exc.requested_key,
+                self._address,
+            ) from exc
 
     @property
     def has_changed_root(self) -> bool:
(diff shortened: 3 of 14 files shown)


eth/db/diff.py
```diff
@@ -195,7 +196,10 @@ def apply_to(self,
                 if apply_deletes:
                     try:
                         del db[key]
+                    except EVMMissingData:
+                        raise
                     except KeyError:
+                        # TODO stop silently swallowing this key error
```
carver: I guess I can remove this comment because I added #1779

eth/db/hash_trie.py
```diff
@@ -7,3 +14,8 @@
 
 class HashTrie(KeyMapDB):
     keymap = keccak
+
+    @contextlib.contextmanager
+    def squash_changes(self) -> Iterator['HashTrie']:
+        with cast(HexaryTrie, self._db).squash_changes() as memory_trie:
+            yield type(self)(memory_trie)
```
carver: Let me know if you think it's worth the time to split this out.

eth/db/storage.py
```diff
@@ -112,8 +118,12 @@ def __delitem__(self, key: bytes) -> None:
         try:
             del write_trie[hashed_slot]
         except trie_exceptions.MissingTrieNode as exc:
-            self.logger.warning(exc)
-            self.logger.debug("expected to be able to delete slot, but:", exc_info=True)
+            raise MissingStorageTrieNode(
+                exc.missing_node_hash,
+                self._starting_root_hash,
+                exc.requested_key,
+                self._address,
```
carver: Note that we don't actually *need* the address pre-image in Beam Sync. But it seemed silly to send the hash here when we have the address in natural form already.

eth/vm/base.py
```diff
@@ -380,6 +383,7 @@ def state_in_temp_block(self) -> Iterator[BaseState]:
 
 
 class VM(BaseVM):
+    cls_logger = logging.getLogger('eth.vm.base.VM')
```
carver: Did the new class logger get released in eth-utils yet?

eth/vm/base.py
```diff
@@ -912,7 +928,14 @@ def validate_header(cls,
                 )
 
             if check_seal:
-                cls.validate_seal(header)
+                try:
+                    cls.validate_seal(header)
+                except ValidationError:
+                    cls.cls_logger.warning(
+                        "Failed to validate header proof of work on header: %r",
+                        header.as_dict()
```
carver: This is an example of somewhere that I'd like to use the built-in `RLP.Serializeable.__repr__`

eth/vm/interrupt.py
```diff
@@ -0,0 +1,118 @@
...
+)
+
+
+class EVMMissingData(PyEVMError):
+    pass
+
+
+class MissingAccountTrieNode(EVMMissingData, MissingTrieNode):
+    """
+    Raised when a storage trie node is missing from the DB
```
carver: Oops, copy-pasta
```suggestion
    Raised when an account trie node is missing from the DB
```

tests/core/opcodes/test_opcodes.py
```diff
@@ -657,7 +657,10 @@ def test_sstore(vm_class, code, gas_used, refund, original):
 
     computation.state.set_balance(CANONICAL_ADDRESS_B, 100000000000)
     computation.state.set_storage(CANONICAL_ADDRESS_B, 0, original)
+    assert computation.state.get_storage(CANONICAL_ADDRESS_B, 0) == original
     computation.state.persist()
+    assert computation.state.get_storage(CANONICAL_ADDRESS_B, 0, from_journal=True) == original
+    assert computation.state.get_storage(CANONICAL_ADDRESS_B, 0, from_journal=False) == original
```
carver: These ended up being helpful when debugging funny things happening to storage.

eth/vm/base.py
carver: Pending in #1750 -- if that gets merged first then this can be updated.

carver:
Ping @pipermerriam for review

tests/core/vm/test_interrupt.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@pytest.fixture
+def bytecode_hash(bytecode):
+    return keccak(bytecode)
+
+
+@pytest.fixture
+def address_with_storage():
+    return b'3' * 20
```
pipermerriam: Note-to-self, we really should implement `evm.tools.factories` and make some `factory-boy` factories for these things.  I think it would help readability a lot.

carver:
> I think the fact that this has no comments is somewhat indicative of the complexity of what you're doing here. I don't really have any suggestions, partially because the overall shape of this is sort of opaque without being steeped in it. I'm  for this to get merged/moved-forward and I suspect that until some of the rest of us really get our hands dirty with this stuff it's gonna stay hard to review.

Totally fair. We'll assume all of this API is still high-velocity. I do want to do a py-evm release after merging, so that we can all start getting our hands dirty on the trinity side.

> Is it possible to write tests (eventually) that demonstrate what you're doing in beam-sync, repeated iterative running of transactions, catching failure, populating database, repeat? That might help understand how this stuff is actually being used as well as establishing this functionality as a core/supported/tested API of py-evm.

Yeah, so I've written some beam sync tests on the trinity side that are a reasonable start at this. It should be reasonable to back-port the data I built for those trinity tests into some py-evm tests.

I'm thinking I'll go ahead and post the PRs that do Beam Sync the current way (raise exception and restart execution). Then everyone can see the direction it's headed, even though sizeable chunks of it will be replaced, when we switch to a pause and resume approach. (Another big change that's coming is making sync work in a different process)

PR #1780, opened 2019-05-16
Switch to eth_utils.HasXXXLogger

pipermerriam:
### How was it fixed?

#### Cute Animal Picture


carver:
Seems like this ought to block on https://github.com/ethereum/eth-utils/issues/159 because the logging check  has a significant EVM performance impact.

pipermerriam:
Looks like this doesn't play nicely with `abc.ABC`.  Ends up with a metaclass error that I think needs to be resolved at the eth-utils level.  Might opt to re-work it using brian's descriptor approach which resolves both `abc.ABC` and `typing.Generic`.

PR #1783, opened 2019-05-20
[Hacked Prototype] Beam Sync Support

carver:
This is a change that supports the first hacked version of Beam Sync. That version, which will soon be posted to trinity (and linked here) allows the exception about a missing node exception to be bubbled up and then re-imports the block.

Key bits:
- Uncle validation doesn't work because we don't have the bodies of the uncles at the first step of beam sync, so I added a mechanism to turn it off (I think I'll just download the last 6 bodies to get around this problem in the next beam sync version
- Restarting block execution from the beginning of the block after a missing node was *really* slow, so I added a way to restart execution starting from the previously completed transaction. I expect a very different approach in the next Beam Sync, so this will mostly be moot (and probably never merged).
- There are a lot of debugging statements that would surely be dropped if this was going to be a real PR
- no py3.5 support

Sorry the commit names are mostly useless

#### Cute Animal Picture

### DIFF
--- a/eth/__init__.py
+++ b/eth/__init__.py
@@ -21,7 +21,7 @@ from eth.chains import (  # noqa: F401
 #
 #  Ensure we can reach 1024 frames of recursion
 #
-EVM_RECURSION_LIMIT = 1024 * 12
+EVM_RECURSION_LIMIT = 1024 * 12 * 1000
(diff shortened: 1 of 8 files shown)


carver:
This approach (restarting block imports from the middle) was abandoned for: pausing VM execution while waiting for the missing nodes to come in. (much faster!)

PR #1785, opened 2019-05-22
Drop python3.5 support

carver:
### What was wrong?

python3.5 prevents us from using f-strings, native `pathlib.Path()` objects in some places, type annotations, etc. python3.6+ should be available on enough targets, and the most important dependency (trinity) already requires 3.6+.

### How was it fixed?

- Dropped support from pypi classifiers
- Dropped 3.5 tests
- Some test reordering to aim to run the longest tests up-front (to minimize joint test time)
- Update docs links to point to python3.6
- Added py3.7 tests (not really related, just maintenance while touching CI)

### TODO

- [ ] add to release notes

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,10 +1,9 @@
 [tox]
 envlist=
-    py{35,36}-{core,database,transactions,vm}
-    py{36}-{benchmark}
-    py{35,36}-native-blockchain-{frontier,homestead,tangerine_whistle,spurious_dragon,byzantium,constantinople,petersburg,metropolis,transition}
-    py37-{core}
-    py{35,36}-lint
+    py{36,37}-{core,database,transactions,vm}
+    py36-benchmark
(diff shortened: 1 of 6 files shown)


carver:
no rush

tox.ini
```diff
@@ -31,13 +30,18 @@ commands=
...
+deps =
+    .[eth-extra,test]
+    lint: .[eth,lint]
 
 basepython =
-    py35: python3.5
     py36: python3.6
     py37: python3.7
+setenv=
+    lint: MYPYPATH={toxinidir}:{toxinidir}/stubs
```
pipermerriam: just noting that we *should* be able to move away from this but need to double check what stubs we are using.

carver: Huh, I moved this over pretty blindly. There is no stubs folder... I'm just going to try removing it and see what happens.

PR #1787, opened 2019-06-04
Expose and test petersburg_at and latest_mainnet_at

carver:
### What was wrong?

`petersburg_at` wasn't exposed in `eth.tools.builder.chain`.

No easy way to use the chain builder to always test the latest VM. (Every network upgrade would require updating the tests explicitly)

### How was it fixed?

Exposed `petersburg_at()` and `latest_mainnet_at()`.

#### Cute Animal Picture


eth/tools/builder/chain/builders.py
```diff
@@ -234,6 +237,8 @@ def dao_fork_at(dao_fork_block_number: BlockNumber,
 petersburg_at = fork_at(PetersburgVM)
 istanbul_at = fork_at(IstanbulVM)
 
+latest_mainnet_at = petersburg_at
```
pipermerriam: :+1: to not making this magic so tests don't suddenly start failing without this being updated explicitly.

PR #1797, opened 2019-07-11
Improve layout of API docs

cburgdorf:
### What was wrong?

This builds on top of #1796 to get the changes into the release notes, too.

Now

image

### How was it fixed?

Then:

image

#### Cute Animal Picture

### DIFF
--- a/docs/_static/css/custom.css
+++ b/docs/_static/css/custom.css
@@ -18,3 +18,46 @@ tt, code, pre {
   margin: 0;
   padding: 0;
 }
+
+/* We instruct sphinx to order members groupwise and use the following css to add headlines.
+Notice that while the css differentiates between methods, classmethods and staticmethods,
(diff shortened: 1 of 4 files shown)


pipermerriam:
Am I correct that the grouping isn't working as expected?

cburgdorf:
>Am I correct that the grouping isn't working as expected?

It groups by *attributes*  / *methods*. I would like to group by *attributes*, *methods*, *class methods* and *static methods* but I believe grouping by *attributes*  and *methods*  is the most essential as it decides if you need to call something with or without parenthesis.

cburgdorf:
@pipermerriam I rebased this and pinged you for review again. I think it's a solid improvement. It also occured to me that on our road to Trinity beta, we'll probably cut a beta release of `py-evm` some time before that. I'm taking a quick pass through the docs to improve as much as I can in a reasonable short time frame before I direct my focus on the Trinity tasks.

PR #1803, opened 2019-07-27
Add EIP-1344 Chain ID

fubuloubu:
### What was wrong?

Fixes https://github.com/ethereum/py-evm/issues/1817

### How was it fixed?
Added a new `ChainContext` class, which allows the chain configuration (and other external parameters) to inform the VM class of that outside data. Routed the Chain ID configuration parameter through this new class.

#### Cute Animal Picture

cute penguin


eth/vm/logic/context.py
```diff
@@ -76,6 +76,10 @@ def calldatacopy(computation: BaseComputation) -> None:
     computation.memory_write(mem_start_position, size, padded_value)
 
 
+def chain_id(computation: BaseComputation) -> None:
+    computation.stack_push(computation.transaction_context.chain_id)
```
fubuloubu: This line needs to change to not use the transaction context value

pipermerriam: Where do you foresee the `chain_id` being retrieved from?

fubuloubu: The `BaseChain` class and all subclasses define the correct configuration we wish to use.

Unfortunately, using the transaction context would not be valid because technically it is still valid to forgo using EIP-155 when signing a transaction, and we cannot have the opcode returning a different value based on the message signer's choice as it could be a security bug where the L2 message signer abd the L1 message signer disagree, and the messages are considered invalid when they were correctly signed.

pipermerriam: Wherever it gets stored, we should probably ensure we have some rigorous tests for this in place to be extra sure that we don't end up leaking the wrong value.

I'm thinking that this might justify a different data storage object, something like `ChainContext` that might only house this value for now but would give us an appropriate place to store it.

fubuloubu: Still WIP, but I added that `ChainContext` class to `BaseVM`/`VM`

Unsure of how to pass it to `BaseVM` classes though...

fubuloubu:
@pipermerriam got it working, just missing test cases, and the updated test suite which includes EIP-1344 tests (which is WIP IIRC) How do you feel about this approach?

tests/json-fixtures/test_virtual_machine.py
```diff
@@ -187,7 +188,8 @@ def test_vm_fixtures(fixture, vm_class, computation_getter):
         gas_limit=fixture['env']['currentGasLimit'],
         timestamp=fixture['env']['currentTimestamp'],
     )
-    vm = vm_class(header=header, chaindb=chaindb)
+    chain_context = ChainContext(chain_id=42)  # TODO is there a better way to do this?
```
carver: IIUC, this will have to partially be answered by the test fixture builders. They will have to supply a chainID in the `env` if they want to test different ones. But presumably we need to handle the default case where nothing is supplied. Choosing an arbitrary chain ID 42 when nothing is supplied seems fine to me.

eth/vm/base.py
```diff
@@ -438,7 +452,7 @@ def build_state(cls,
         even if you don't have the VM initialized. This is a convenience method to do that.
         """
 
-        execution_context = header.create_execution_context(previous_hashes)
+        execution_context = header.create_execution_context(previous_hashes, chain_context)
```
carver: Hm, by the name `ChainContext`, it seems like `previous_hashes` could be a part of it. What does it look like if you try to combine the two?

carver: Other than that question, approach seems fine. :+1:

tests/json-fixtures/test_virtual_machine.py
fubuloubu: I think the value where nothing is supplied should be `0` or `None` because 42 is actually a valid chain ID for the Kovan network: https://chainid.network/

eth/vm/base.py
fubuloubu: Unclear to me what previous hashes is supposed to be for other than a pass-through to `ExecutionContext` but I do agree that `ChainContext` should inform `ExecutionContext` with any valid data from outside the VM.

Is `previous_hashes` used to obtain the result of calling the `BLOCKHASH` opcode?

tests/json-fixtures/test_virtual_machine.py
carver: Since it's not defined for `None`, chain ID should probably use some actual integer. Zero seems fine, or something random and a bit big, like `0xa1fa1fa = 169845242`

fubuloubu: Currently, pre-`EIP-155` tranasctions (and actually post too, they're still valid) does not have to define a chain ID since one didn't exist. How is this handled?

fubuloubu: Also, 0 is probably the best choice since I don't think it is defined as being valid with 0 as the value.

eth/vm/base.py
pipermerriam: :+1: to moving `previous_hashes` to `ChainContext`

tests/json-fixtures/test_virtual_machine.py
carver: > Currently, pre-`EIP-155` tranasctions (and actually post too, they're still valid) does not have to define a chain ID since one didn't exist. How is this handled?

How do you mean? `EIP-155` isn't validated at all, yet.

fubuloubu: oh, I think I understand now why you made this comment. `chain_id` is a mandatory parameter of `ChainContext`, I just used the kwarg syntax to make it clearer what that number was for

eth/vm/base.py
fubuloubu: So, this doesn't seem like a good idea because the list of previous hashes needs to be generated on the spot when an execution context is started because the block height may have changed.

Perhaps `ChainContext` isn't the right way to keep track of this, maybe something more like `BlockContext` or something would be better? Also, would be ideal for `chain_id` because there is some small likelihood that this number would change at a particular block height for a contentious fork, or if a process is invented to cycle chain ID on any hard fork.

carver: Fair enough, making `ChainContext` something that is immutable through the lifetime of a chain is a reasonable choice. Let's write that explicitly in the class docstring.

fubuloubu: Well, that's my point... chain ID _might not_ be the same for the entirety of the chain, but I think it's something that is liable to be immutable per VM. But it's _defined_ from the chain config... so confusing, lol.

fubuloubu:
Refactored and it at least passes MyPy now. Will look into failures later.

carver:
Rather than wait on state tests to merge, can you add some local tests, like #1829 does?

Then we can get it merged up, and add in the state tests as they come in. Don't forget to add a release note to mention that the EIP is added to Istanbul (there's also an example of that in #1829).

fubuloubu:
@carver I added a simple test and release notes for the feature. Any further suggestions to get this merged? Also, any suggestions on what to clean up with the commit history, if necessary?

eth/vm/chain_context.py
```diff
@@ -0,0 +1,25 @@
...
+    This immutable object houses chain information that remains constant for the entire context of
+    the VM execution.
+    """
+    __slots__ = ['_chain_id']
+
+    def __init__(self, chain_id: Optional[int]) -> None:
+
+        if chain_id is None:
+            chain_id = 0  # Default value (invalid for public networks)
+        validate_uint256(chain_id)
```
fubuloubu: Note: update this to ensure `uint64`, which is the suggested maximum due to the mathematical configuration of EIP-155

carver:
In the interest of time, I just added a few little tweaks (and handled the rebase). If anything looks off to you, let me know, and we can do a follow-up.

fubuloubu:
The tweaks look good!

PR #1805, opened 2019-07-31
Implement eth.abc module with abstract base classes for public APIs

pipermerriam:
### What was wrong?

While working with the database components of the library in the trinity codebase I found myself wanting real abstract base class APIs to develop against.

### How was it fixed?

Created `eth.abc` and created propert `ABC` base classes for all of the main APIs from the codebase.

This should have near zero API changes.  The places where APIs changed are minor and have to do with some juggling to define the ABC classes for the `rlp.Serializable` objects.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/db/__init__.py
+++ b/eth/db/__init__.py
@@ -7,21 +7,22 @@ from typing import (
 
 from eth_utils import import_string
 
-from eth.db.backends.base import BaseAtomicDB
+from eth.abc import AtomicDatabaseAPI
 
 
 DEFAULT_DB_BACKEND = 'eth.db.atomic.AtomicDB'
 
 
-def get_db_backend_class(import_path: str = None) -> Type[BaseAtomicDB]:
+def get_db_backend_class(import_path: str = None) -> Type[AtomicDatabaseAPI]:
     if import_path is None:
         import_path = os.environ.get(
             'CHAIN_DB_BACKEND_CLASS',
             DEFAULT_DB_BACKEND,
         )
-    return cast(Type[BaseAtomicDB], import_string(import_path))
+    return cast(Type[AtomicDatabaseAPI], import_string(import_path))
 
 
-def get_db_backend(import_path: str = None, **init_kwargs: Any) -> BaseAtomicDB:
+def get_db_backend(import_path: str = None, **init_kwargs: Any) -> AtomicDatabaseAPI:
     backend_class = get_db_backend_class(import_path)
-    return backend_class(**init_kwargs)
+    # mypy doesn't understand the constructor  of AtomicDatabaseAPI
+    return backend_class(**init_kwargs)  # type: ignore
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -1,7 +1,3 @@
-from abc import (
-    ABC,
-    abstractmethod
-)
 import logging
 from lru import LRU
 from typing import (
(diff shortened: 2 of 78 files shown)


eth/db/account.py
```diff
@@ -75,142 +73,11 @@
...
-
-        Current changes include every pending change to storage, as well as all account changes.
-        After generating all the required tries, the final account state root is returned.
-
-        This is an expensive operation, so should be called as little as possible. For example,
-        pre-Byzantium, this is called after every transaction, because we need the state root
-        in each receipt. Byzantium+, we only need state roots at the end of the block,
-        so we *only* call it right before persistance.
-
-        :return: the new state root
```
pipermerriam: Anywhere like this where the original abc method had a docstring, the docstring was migrated to the new abc base class.

eth/db/chain.py
```diff
@@ -239,8 +139,8 @@ def persist_block(self,
     @classmethod
     def _persist_block(
             cls,
-            db: 'BaseDB',
```
pipermerriam: This resulted in a lot of cleanups like this where we no longer have circular import issues with type hints.

eth/db/chain.py
```diff
@@ -314,27 +216,27 @@ def add_transaction(self,
...
-            transaction_class: Type['BaseTransaction']) -> Iterable['BaseTransaction']:
+            header: BlockHeaderAPI,
+            transaction_class: Type[SignedTransactionAPI]) -> Tuple[SignedTransactionAPI, ...]:
         """
         Returns an iterable of transactions for the block speficied by the
         given block header.
         """
         return self._get_block_transactions(header.transaction_root, transaction_class)
 
-    def get_block_transaction_hashes(self, block_header: BlockHeader) -> Iterable[Hash32]:
```
pipermerriam: And I found a few places like this where our return types were allowing for things like returning a generator when that is typically a foot-gun.

pipermerriam:
Looks like I need to do some cleanup to get the fixture tests passing.

eth/db/__init__.py
```diff
@@ -7,21 +7,22 @@
...
         )
-    return cast(Type[BaseAtomicDB], import_string(import_path))
+    return cast(Type[AtomicDatabaseAPI], import_string(import_path))
 
 
-def get_db_backend(import_path: str = None, **init_kwargs: Any) -> BaseAtomicDB:
+def get_db_backend(import_path: str = None, **init_kwargs: Any) -> AtomicDatabaseAPI:
     backend_class = get_db_backend_class(import_path)
-    return backend_class(**init_kwargs)
+    # mypy doesn't understand the constructor  of AtomicDatabaseAPI
```
carver: Does it work if you add back the explicit `@abstractmethod` with `__init__` that was in `BaseDB`?

```py
    @abstractmethod	
    def __init__(self) -> None:	
        raise NotImplementedError(	
            "The `init` method must be implemented by subclasses of BaseDB"	
        )
```

pipermerriam: Yes, but I don't believe there is a generic `__init__` method for `AtomicDatabaseAPI` implementations.  We have one right now that wraps another database but I don't think that's the only constructor.

Also, I'm inclined to remove this functionality entirely because I don't think we actually use it in any meaningful way.

PR #1808, opened 2019-08-05
Update to eth-utils 1.6.4

pipermerriam:
### What was wrong?

we've ended up needing `ExtendedDebugLogger` in other codebases (primarily the `trinity` one).

### How was it fixed?

As of `eth-utils >= 1.6.4` there is now an implementation of the `ExtendedDebugLogger` available from `eth-utils`.  I removed the local implementation and updated all of the code to use the one from `eth-utils`.

This includes some metaclass hoop-jumping in the `eth.abc` module.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth/vm/state.py
```diff
@@ -218,7 +205,7 @@ def revert(self, snapshot: Tuple[Hash32, UUID]) -> None:
         # now roll the underlying database back
         self._account_db.discard(account_snapshot)
 
-    def commit(self, snapshot: Tuple[Hash32, UUID]) -> None:
+    def commit(self, snapshot: Tuple[Hash32, JournalDBCheckpoint]) -> None:
```
cburgdorf: I'm curious, is this change because `mypy` `0.720` started noticing that this can not be `UUID` anymore? If so, hooray for improved type safety :)

eth/abc_compat.py
```diff
@@ -0,0 +1,53 @@
...
+    from .abc import (  # noqa: F401
+        ComputationAPI,
+    )
+
+
+ABCWithLoggerMeta = HasExtendedDebugLoggerMeta.meta_compat(ABCMeta)
+
+
+# mypy doesn't recognize ABCWithLoggerMeta as a valid metaclass
+class ABCWithLogger(metaclass=ABCWithLoggerMeta):  # type: ignore
```
cburgdorf: Can you elaborate why you went for this? Why not continue to set the `logger` in the class that needs the logger as it was before? Considering how low-level the  `ABC` class it, it doesn't feel natural to me to push the logger in there. Is there another reason for this other than just: `ABC` is usually the lowest level thing that many our classes derive from and now there's `ABCWithLogger` to save you an extra line for that one thing you usually need?

eth/vm/state.py
pipermerriam: Yes, new errors showed up when I upgraded `mypy`

eth/abc_compat.py
pipermerriam: I'm in agreement that I don't love the `ABCWithLogger` base class and the semi-complex juggling that is required to make it work, but I'm in favor of it **over** having the following line all over our code.

```python
logger = cast(ExtendedDebugLogger, logging.getLogger('path.to.MyLogger')
```

If it was *only* the `cast` I might be ok with it but it's also that there is no guarantee that `logging.getLogger` will return an actual `ExtendedDebugLogger` since that depends on `logging.setLoggerClass` having been last called with `ExtendedDebugLogger`.

So my reasoning is, isolated complexity here to remove lots of fragile feeling code across most of our code.

However, in saying this, I realize there's an allowable 3rd option that we should consider.  We can write a `get_extended_debug_logger(name)` function which uses the following pattern from `eth-utils` to reliably return an `ExtendedDebugLogger` instance.

https://github.com/ethereum/eth-utils/blob/e8aea732d5115a94d82268dd88c7d2ef1c73d794/eth_utils/logging.py#L81-L82

This is probably better than the fancy metaclass approach and I'm annoyed at myself that I didn't see this before.....

cburgdorf: >but it's also that there is no guarantee that `logging.getLogger` will return an actual `ExtendedDebugLogger`

That's a good reason indeed.

> realize there's an allowable 3rd option that we should consider. We can write a `get_extended_debug_logger(name)` function 

I like that one more :+1: 

>I'm annoyed at myself that I didn't see this before

Don't be! I'm happy my review could contribute constructive feedback :)

carver: So the pattern would change like:

```diff
- logger = cast(ExtendedDebugLogger, logging.getLogger('path.to.MyLogger')
+ logger = get_extended_debug_logger('path.to.MyLogger')
```

It's a small improvement, but I have seen copy-pasta errors when the class name or path gets injected incorrectly. What if we add the logger attribute using a class decorator? Can that be made friendly to mypy? (Assuming mypy was the reason extended debug logger was being made into a metaclass instead of a traditional mixin class, which seems like the easiest solution)

pipermerriam: Metaclass was chosen because mix-in class (without a metaclass) doesn't expose the logger on the class itself and it's a little nicer when it isn't a computed property.  The proposed `get_extended_debug_logger` approach would still be subject to copy/pasta errors but I don't really see that as a very big drawback since the consequences are pretty small and it's easy to fix when it does happen.

pipermerriam:
superceded by #1846

PR #1813, opened 2019-08-06
Package up base db and atomic db test suites into re-usable classes

### LINKED ISSUE #1812: Make test suites for `DatabaseAPI` and `AtomicDatabaseAPI` re-usable
### What is wrong?

The `trinity` codebase now has an implementation of `AtomicDatabaseAPI` that needs to be tested.  This would require duplicating all of the tests we have here in `py-evm` that verify the correctness of different implementations of the API

### How can it be fixed

The tests can be packaged up in a re-usable manner and stowed somewhere like `eth.tools.testing` so they can be re-used by other libraries.

pipermerriam:
fixes #1812 

### What was wrong?

The atomic and base DB test suites are useful outside of the py-evm library but they were not re-usable in the form they were written.  Also most of the base database test suite was duplicated.

### How was it fixed?

Packaged the test suites up into class-based tests and put them under `eth.tools.db`.

I also expanded the test suite to cover the `batch` object for the `AtomicDB` as well as and the base db suite to cover the `CacheDB`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth/tools/db/base.py
```diff
@@ -0,0 +1,76 @@
...
+        assert b'key-1' not in db
+
+    def test_database_api_delete_missing_key(self, db: DatabaseAPI) -> None:
+        assert b'key-1' not in db
+        db.delete(b'key-1')
+
+    def test_database_api_item_delete_missing_key(self, db: DatabaseAPI) -> None:
+        assert b'key-1' not in db
+        with pytest.raises(KeyError):
+            del db[b'key-1']
```
lithp: I'm surprised that `.delete()` doesn't throw an exception but `__delitem__` does! It seems like those two should have the same behavior.

pipermerriam: I just realized that `MutableMapping` doesn't have a `set`, `delete` or `exists` API....

Issue #1815, opened 2019-08-08
Can't install py-evm dev setup with latest `brew install leveldb`


davesque:
* py-evm Version: current master (d2ccc1e8a9be1cb92cb15a8bb117bb6a1972636e)
* OS: osx
* Python Version (python --version): 3.7.4
* Environment (output of `pip freeze`):
```
alabaster==0.7.12
attrs==19.1.0
Babel==2.7.0
bumpversion==0.5.3
cached-property==1.5.1
certifi==2019.6.16
chardet==3.0.4
cytoolz==0.10.0
docutils==0.15.2
eth-bloom==1.0.3
eth-hash==0.2.0
eth-keys==0.2.4
eth-typing==2.1.0
eth-utils==1.6.4
idna==2.7
imagesize==1.1.0
Jinja2==2.10.1
lru-dict==1.1.6
MarkupSafe==1.1.1
mypy-extensions==0.4.1
packaging==19.1
py-ecc==1.7.1
pyethash==0.1.27
Pygments==2.4.2
pyparsing==2.4.2
pysha3==1.0.2
pytz==2019.2
requests==2.22.0
rlp==1.1.0
six==1.12.0
snowballstemmer==1.9.0
Sphinx==1.7.9
sphinxcontrib-asyncio==0.2.0
sphinxcontrib-websupport==1.1.2
toolz==0.10.0
trie==1.4.0
urllib3==1.25.3
```

### What is wrong?

It seems the required version of `plyvel` (1.0.5) is incompatible with the latest version of leveldb on installable with homebrew (stable 1.22):
```
$ pip install --upgrade pip setuptools wheel
Collecting pip
  Using cached https://files.pythonhosted.org/packages/62/ca/94d32a6516ed197a491d17d46595ce58a83cbb2fca280414e57cd86b84dc/pip-19.2.1-py2.py3-none-any.whl
Collecting setuptools
  Using cached https://files.pythonhosted.org/packages/ec/51/f45cea425fd5cb0b0380f5b0f048ebc1da5b417e48d304838c02d6288a1e/setuptools-41.0.1-py2.py3-none-any.whl
Collecting wheel
  Using cached https://files.pythonhosted.org/packages/bb/10/44230dd6bf3563b8f227dbf344c908d412ad2ff48066476672f3a72e174e/wheel-0.33.4-py2.py3-none-any.whl
Installing collected packages: pip, setuptools, wheel
  Found existing installation: pip 19.0.3
[...]

PR #1816, opened 2019-08-08
Update dev plyvel requirement

davesque:
### What was wrong?

See #1815 .

### How was it fixed?

Updated `plyvel` dev requirement to `1.1.0`.

### To-Do

- [x] Clean up commit history

- [ ] Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md) (I think this is N/A?)

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ deps = {
         "coincurve>=10.0.0,<11.0.0",
         "eth-hash[pysha3];implementation_name=='cpython'",
         "eth-hash[pycryptodome];implementation_name=='pypy'",
-        "plyvel==1.0.5",
+        "plyvel>=1.0.5,<1.2.0",
     ],
     'test': [
         "hypothesis==3.69.5",


setup.py
```diff
@@ -25,7 +25,7 @@
         "coincurve>=10.0.0,<11.0.0",
         "eth-hash[pysha3];implementation_name=='cpython'",
         "eth-hash[pycryptodome];implementation_name=='pypy'",
-        "plyvel==1.0.5",
```
cburgdorf: I wonder if we should change this into a range to not break things for those that are stuck on something lower than `1.1.0` which is known to work well.

davesque: Yeah, probably.  Just made that change.

cburgdorf: Ah wait, what I meant was to include everything from `1.0.5` up to whatever you say is working fine. Because `1.0.5` is the currently used version and I don't see a good reason to *require*  a higher version when `1.0.5` is known to work well. Especially because there may be people who can not easily upgrade to a higher version.

davesque: As a reminder, `plyvel` is just the python interface library to LevelDB.  If we require a more recent `plyvel` version, it doesn't necessarily mean we're requiring a higher LevelDB version.  And it's a stronger guarantee that macOS users won't encounter the linked issue.

Also, `plyvel` has no sub-dependencies.  So that also makes it seem less risky to make this change.

What do you think?

davesque: @cburgdorf By the way, with my comment above, you didn't mention being concerned about the level db version, but I figured I'd mention my thoughts on that just in case you were.

davesque: Anyway, I'll just go with your suggestion.  I don't actually feel that strongly about it. 😄

cburgdorf: >As a reminder, plyvel is just the python interface library to LevelDB

Ah, that's a good point. That weakens my concerns. :+1: 

With that in mind, I guess a range of `plyvel>=1.1.0, <1.2.0` is fine.

>Anyway, I'll just go with your suggestion. I don't actually feel that strongly about it.

Sorry, for not making myself clear enough. I didn't mean to suggest a range without an upper bound such as `plyvel>=1.0.5"`.

Because I think that one can in fact become problematic. Current plyvel works under the assumption of having a `leveldb` of version `1.2` or higher on the system. An unbounded `plyvel` dependency might even install a `plyvel` version `2.x` which may depend on a `leveldb` `2.x` (hypothetical future versions) and with breaking changes things might fall apart then.

I'd pick one of these:

- `plyvel>=1.1.0, <2` (if we trust plyvel does semver right)
- `plyvel>=1.0.5, <2` (if we trust plyvel does semver right)
- `plyvel>=1.1.0, <1.2.0` (more conservative)
- `plyvel>=1.0.5, <1.2.0` (more conservative)

davesque: I'll go with the last suggestion since it basically solves the macOS issue and also potentially avoids other issues with `plyvel` being incompatible with certain LevelDB versions.

cburgdorf:
>Is there a reason to think that plyvel is not using semver? Anyway, no objection to the PR as-is.

No idea. Previously, the dependency was pinned to a single specific version which might serve as a hint that there were issues about this in the past but it's pure speculation ;)

PR #1823, opened 2019-08-14
Allow persisting headers that are disconnected from genesis

cburgdorf:
### What was wrong?

We want to allow persisting headers that are disconnected from genesis (with the intend to connect them back later). This is to allow syncing from a trusted checkpoint.

### How was it fixed?

Add an *optional* `genesis_parent_hash` parameter to
`eth.db.header.HeaderDB.persist_header_chain` and
`eth.db.chain.ChainDB.persist_block` that allows to overwrite the hash that is used
to identify the genesis header. This allows persisting headers / blocks that aren't (yet)
connected back to the true genesis header.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth/db/header.py
```diff
@@ -83,6 +83,10 @@ def persist_header(self,
                        ) -> Tuple[Tuple[BlockHeader, ...], Tuple[BlockHeader, ...]]:
         raise NotImplementedError("ChainDB classes must implement this method")
 
+    @abstractmethod
+    def persist_checkpoint_header(self, header: BlockHeader, score: int) -> None:
+        raise NotImplementedError("ChainDB classes must implement this method")
```
pipermerriam: New paradigm uses `...` for this.

cburgdorf:
>I have a very minor spider-sense reaction to the overloading of genesis_parent_hash to trick the database into doing the right thing

Yes, I agree. But at the same time, it's strikingly simple so it may not be that bad after all :D

>tests failing, waiting to see what tests say when conflicts are gone.

Yep, I started off on the previous release. But now that you upgraded Trinity to latest py-evm code (thanks for taking care of that!) I have this rebased and polished.

cburgdorf:
I added another commit to kill the `AsyncHeaderDB`. It wasn't used anywhere (no, not even in Trinity). Trinity defines it's own async class and I think that may be the better place after all. So, this became just a burden to update for no real reason. I can move that out if it turns out to be controversial.

PR #1824, opened 2019-08-15
Update understanding_the_mining_process.rst

tczpl:
func: get_nonce is transferred from VM.state to VM.state.account_db

### What was wrong?

get_nonce() can not be called from VM.state.

### How was it fixed?

get_nonce() has been transferred from VM.state to VM.state.account_db

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/guides/understanding_the_mining_process.rst
+++ b/docs/guides/understanding_the_mining_process.rst
@@ -269,7 +269,7 @@ With sender and receiver prepared, let's create the actual transaction.
 ::
 
     vm = chain.get_vm()
-    nonce = vm.state.get_nonce(SENDER)
+    nonce = vm.state.account_db.get_nonce(SENDER)
 
     tx = vm.create_unsigned_transaction(
         nonce=nonce,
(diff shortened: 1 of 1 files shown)


cburgdorf:
Hey @tczpl 

thanks for reading through the guide (I hope you liked it!) and opening up the PR. I don't follow the proposed changes though. The `get_nonce()` API **is available** on the state.

https://github.com/ethereum/py-evm/blob/d9a5b94a2f218ad922a503876fde4faeecc9b6de/eth/vm/state.py#L165-L166

Also, notice that the bottom of the guide has the entire example written with `doctest` syntax, which means it runs against the real python interpreter on every CI run. It should work :)

docs/guides/understanding_the_mining_process.rst
```diff
@@ -356,7 +356,7 @@ zero value transfer transaction.
   >>> chain = klass.from_genesis(AtomicDB(), GENESIS_PARAMS)
   >>> vm = chain.get_vm()
 
-  >>> nonce = vm.state.get_nonce(SENDER)
+  >>> nonce = vm.state.account_db.get_nonce(SENDER)
```
cburgdorf: @tczpl  Everything in this example is actual code that is executed on every CI run. If you just copy this out and replace `>>> ` and `... ` with an empty string, you can put it in a file and run it and it should just work.

tczpl: Thank you. I found it in state.py in the github. But when I execute it as it is in the guide, it throws `AttributeError: 'ByzantiumState' object has no attribute 'get_nonce'`.  So I check the state.py in my local package in python that it does not exist. It seems that I have installed an old version of py-evm. So I close this issue~

PR #1830, opened 2019-08-21
Make checkpoint header the canonical head

cburgdorf:
### What was wrong?

We recently added support to persist checkpoint headers, in other words, headers, not connected to genesis, persisted with a score that we trust.

However, previously, when we set `persist_checkpoint_header` it did not make that checkpoint header the canonical head.

The feature is used for checkpointed beam sync (https://github.com/ethereum/trinity/pull/921) 
and in a previous version of that PR it didn't mattered if that API call would make the checkpoint header the new canonical head or not.

The reason for that is because eventually a header on top would be persisted with `persist_header_chain(headers, genesis_parent_hash)` and then that would move the canonical head to the latest header.

But with a refactored https://github.com/ethereum/trinity/pull/921 this turned out to be a problem because we now assume that calling `persist_checkpoint_header` would also make that checkpoint header the new canonical head (Which strikes me as a logical thing).

### How was it fixed?

This makes the persisted header the new canonical head without any further checking. I'm not entirely sure but I can't think of a case where we would **not** want to make that checkpoint header the new canonical head.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/db/header.py
+++ b/eth/db/header.py
@@ -167,6 +167,10 @@ class HeaderDB(HeaderDatabaseAPI):
             return self._persist_header_chain(db, headers, genesis_parent_hash)
 
     def persist_checkpoint_header(self, header: BlockHeaderAPI, score: int) -> None:
+        """
+        Persist a checkpoint header with a trusted score. Persisting the checkpoint header
+        automatically sets it as the new canonical head.
+        """
         with self.db.atomic_batch() as db:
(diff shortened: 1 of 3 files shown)


cburgdorf:
@carver Good point! I have some additional questions here. I created an issue to not hold back the release + Trinity PR. https://github.com/ethereum/py-evm/issues/1833

PR #1829, opened 2019-08-20
EIP-1884: Reprice trie-size dependent opcodes

carver:
### What was wrong?

Fixes #1826 

### How was it fixed?

Changed pricing, and added selfbalance opcode, added tests.

Bonus simplification in the insufficient stack exception.

### To-Do

- [x] Add tests defined in the EIP
- [x] Rebase on #1828 

[//]: # (Stay ahead of things, add list items here!)
- [x] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md)

[//]: # (See: https://py-evm.readthedocs.io/en/latest/contributing.html#pull-requests)
- [x] Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

Issue #1825, opened 2019-08-15
EIP-2200: Rebalance net-metered SSTORE gas cost


carver:
For Istanbul: https://github.com/ethereum/EIPs/pull/2200/files?short_path=3aefdb0#diff-3aefdb0e903ce495687d8c3fbc101c1d

PR #1831, opened 2019-08-21
Refactor net gas metering as prep for EIP-2200

carver:
### What was wrong?

There is a lot of shared logic between the original 1283 net gas metering, and the new 2200 net gas metering, but not an easy way to reuse it with a few variables updated.

Related to #1825 

### How was it fixed?

Generalized net gas metering logic to accept a gas schedule as an argument.

You can see where this is headed with something like an Istanbul version of:

```py
from eth.exceptions import OutOfGas
from eth.vm.computation import BaseComputation
from eth.vm.forks.constantinople.storage import (
    GAS_SCHEDULE_EIP1283,
)
from eth.vm.forks.istanbul import (
    constants
)
from eth.vm.logic.storage import (
    net_sstore,
)

GAS_SCHEDULE_EIP2200 = GAS_SCHEDULE_EIP1283._replace(base=constants.GAS_SLOAD_EIP1884)


def sstore_eip2200(computation: BaseComputation) -> None:
    gas_remaining = computation.get_gas_remaining()
    if gas_remaining <= 2300:
        raise OutOfGas(
            "Net-metered SSTORE always fails below 2300 gas, per EIP-2200",
            gas_remaining,
        )
    else:
        return net_sstore(GAS_SCHEDULE_EIP2200, computation)
```

Since adding this ^ depends on merging #1829 , I'm splitting this in two parts. Each one should be a pretty quick review.

### To-Do

- [x] Clean up commit history

- [x] ~~Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md)~~

#### Cute Animal Picture

### DIFF
--- a/eth/vm/logic/storage.py
+++ b/eth/vm/logic/storage.py
@@ -1,3 +1,5 @@
+from typing import NamedTuple
+
 from eth_utils import (
     encode_hex,
 )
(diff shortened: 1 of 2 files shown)


eth/vm/logic/storage.py
```diff
@@ -58,3 +60,73 @@ def sload(computation: BaseComputation) -> None:
         slot=slot,
     )
     computation.stack_push_int(value)
+
+
+class NetSStoreGasSchedule(NamedTuple):
+    base: int    # the gas cost when nothing changes (eg~ dirty->dirty, clean->clean, etc)
+    create: int  # a brand new value, where none previously existed, aka init or set
+    update: int  # a change to a value when the value was previously unchanged, aka clean, reset
+    remove_refund: int  # the refund for removing a value, aka: clear_refund
```
cburgdorf: Great to have :point_up: here!

PR #1832, opened 2019-08-22
EIP-2028: Transaction data gas cost reduction

### LINKED ISSUE #1820: EIP-2028: Calldata gas cost reduction
For Istanbul: https://eips.ethereum.org/EIPS/eip-2028

voith:
### What was wrong?
closes #1820

### How was it fixed?
changed tx data nonzero byte cost to 16 in Istanbul

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture


eth/vm/forks/istanbul/transactions.py
```diff
@@ -40,3 +54,21 @@ def as_signed_transaction(self,
             r=r,
             s=s,
         )
+
+    def get_intrinsic_gas(self) -> int:
+        return _get_instanbul_intrinsic_gas(self)
+
+
+def _get_instanbul_intrinsic_gas(transaction: BaseTransaction) -> int:
```
voith: This fn is a duplicate of `_get_homestead_intrinsic_gas` but with a different `GAS_TXDATANONZERO` value . Should I make this a utility function with `GAS_TXDATANONZERO` as a parameter?

pipermerriam: Yes, better to make the original function re-usable.

voith:
@carver ready for review

tests/core/chain-object/test_gas_estimation.py
```diff
@@ -23,18 +30,18 @@ def chain(chain_without_block_validation):
     'should_sign_tx', (True, False),
 )
 @pytest.mark.parametrize(
-    'data, gas_estimator, to, on_pending, expected',
+    'data, gas_estimator, to, on_pending, pre_instanbul_expected, post_constantinople_expected',
```
carver: This part is tripping me up, but I didn't quickly think of a better option. This approach going to get ugly quickly, after one or two more changes to this gas schedule.

Maybe we can try pairing the expected result with the VM, like:
```
( ... (ConstantinopleVM, 21680), (PetersburgVM, 21680), (IstanbulVM, 21160), ...)
```
(Also note that the last one would be post petersburg, instead of constantinople)

voith:
@carver ready for round 2.

tests/core/chain-object/test_gas_estimation.py
```diff
@@ -14,49 +24,194 @@
...
     ids=[
-        'simple default pending',
-        'simple default',
-        '10 bytes default pending',
-        '10 bytes default',
-        'sha3 precompile 32 bytes default pending',
-        'sha3 precompile 32 bytes default',
-        'sha3 precompile 320 bytes default pending',
-        'sha3 precompile 32 bytes 1000_tolerance binary pending',
+        'simple default pending for FrontierVM',
```
pipermerriam: I think that with such a large list, we should inline the ids using this format:

```python
pytest.param(
            b'', None, ADDR_1010, True, FrontierVM, 21000, id='simple default pending for FrontierVM',
        ),
```

PR #1842, opened 2019-09-02
Make Istanbul inherit from Petersburg instead of Constantinople

voith:
### What was wrong?

`Istanbul` was inheriting everything from `Constantinople` instead of `Petersburg`.
This is incorrect as `Istanbul` is a descendant of `Petersburg` and not `Constantinople`.

### How was it fixed?

Made Istanbul inherit from Petersburg instead of Constantinople

### To-Do

- [x] Clean up commit history

- [x] <strike>Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md)</strike> (Doesn't need release notes IMO as this hasn't been released yet)

#### Cute Animal Picture

### DIFF
--- a/eth/vm/forks/istanbul/computation.py
+++ b/eth/vm/forks/istanbul/computation.py
@@ -6,11 +6,15 @@ from eth import precompiles
 from eth._utils.address import (
     force_bytes_to_address,
 )
-from eth.vm.forks.constantinople.computation import (
-    CONSTANTINOPLE_PRECOMPILES
+from eth.vm.forks.petersburg.computation import (
+    PETERSBURG_PRECOMPILES
 )
-from eth.vm.forks.constantinople.computation import (
-    ConstantinopleComputation
+from eth.vm.forks.petersburg.computation import (
+    PetersburgComputation,
+)
+from eth.vm.gas_meter import (
+    allow_negative_refund_strategy,
+    GasMeter,
 )
 
 from .constants import (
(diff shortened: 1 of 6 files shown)


eth/vm/forks/istanbul/computation.py
```diff
@@ -35,11 +39,17 @@
...
     """
     A class for all execution computations in the ``Istanbul`` fork.
-    Inherits from :class:`~eth.vm.forks.constantinople.computation.ConstantinopleComputation`
+    Inherits from :class:`~eth.vm.forks.constantinople.petersburg.PetersburgComputation`
     """
     # Override
     opcodes = ISTANBUL_OPCODES
     _precompiles = ISTANBUL_PRECOMPILES
+
+    def get_gas_meter(self) -> GasMeter:
```
voith: I copied this from `ConstantinopleComputation`. Not sure what the right thing to do here is. I have two thoughts in my mind:
(1)  Instead of inheriting from `PetersburgComputation`, inherit from `ConstantinopleComputation`. This could be interpreted as reverting the computation object back to Constantinople. This way `get_gas_meter` won't have to be re-defined.
(2) `Istanbul` is the descendant of `Petersburg` and so it should only inherit from `Petersburg`.  Although (1) is DRY, it might not be semantically correct.

voith:
cc @carver

eth/vm/forks/istanbul/computation.py
carver: I don't have a better solution. This seems better than alternative 1.

carver:
Thanks!

PR #1845, opened 2019-09-05
Expose type data to other packages via PEP561

pipermerriam:
### What was wrong?

From a 3rd party library perspective, all of the values from py-evm are `Any` types.

### How was it fixed?

Expose type information via PEP561

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

ratsoff

### DIFF
--- a/newsfragments/1845.feature.rst
+++ b/newsfragments/1845.feature.rst
@@ -1,0 +1,1 @@
+Expose type hint information via PEP561
--- a/setup.py
+++ b/setup.py
@@ -109,6 +109,7 @@ setup(
     zip_safe=False,
     keywords='ethereum blockchain evm',
     packages=find_packages(exclude=["tests", "tests.*"]),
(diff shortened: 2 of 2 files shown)


pipermerriam:
> do we get any fun mypy errors in trinity now?

Yeah, I'm betting we get alot

cburgdorf:
Uh, who would have thought we forgot about that haha

pipermerriam:
@cburgdorf I was thoroughly surprised.  Took me a while to chase down why all these things were showing up as `Any` types.

cburgdorf:
Another thing contributing to `Any` are decorators. Those from `eth_utils` do already preserve type hints but stuff like `@curry` transforms a nicely typed method into an `Any` mess. We should reevaluate from time to time if we can increase some of the extra `mypy` checks to lead us out of `Any`land.

PR #1850, opened 2019-09-17
More consistent use of BlockNumber type

### LINKED ISSUE #1847: `eth.constants.GENESIS_BLOCK_NUMBER` is not a BlockNumber type.
### What is wrong?

The constant value `eth.constants.GENESIS_BLOCK_NUMBER` is of type `int` when it should be of type `eth_typing.BlockNumber`

### How can it be fixed

Wrap the value as `BlockNumber(0)` and then update any places in the codebase where `mypy` complains.

pipermerriam:
fixes #1847

### What was wrong?

The `eth.constants.GENESIS_BLOCK_NUMBER` was not typed as `eth_typing.BlockNumber`.  Nor were many other variables that should have been `BlockNumber` types.

### How was it fixed?

Updated all the places I could find where block numbers are used and made them into proper `BlockNumber` types.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

pets-halloween-willjpg-5d1761894278a2952

### DIFF
--- a/eth/tools/builder/chain/builders.py
+++ b/eth/tools/builder/chain/builders.py
@@ -108,7 +108,7 @@ def chain_id(chain_id: int, chain_class: Type[ChainAPI]) -> Type[ChainAPI]:
 
 @curry
 def fork_at(vm_class: Type[VirtualMachineAPI],
-            at_block: int,
+            at_block: Union[int, BlockNumber],
             chain_class: Type[ChainAPI]) -> Type[ChainAPI]:
     """
     Adds the ``vm_class`` to the chain's ``vm_configuration``.
@@ -146,7 +146,7 @@ def fork_at(vm_class: Type[VirtualMachineAPI],
     else:
         base_configuration = tuple()
 
-    vm_configuration = base_configuration + ((at_block, vm_class),)
+    vm_configuration = base_configuration + ((BlockNumber(at_block), vm_class),)
     validate_vm_configuration(vm_configuration)
     return chain_class.configure(vm_configuration=vm_configuration)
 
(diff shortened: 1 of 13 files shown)


eth/tools/builder/chain/builders.py
```diff
@@ -257,7 +257,7 @@ def dao_fork_at(dao_fork_block_number: BlockNumber,
 
 @to_dict
 def _get_default_genesis_params(genesis_state: AccountState,
-                                ) -> Iterable[Tuple[str, Union[int, None, bytes, Address, Hash32]]]:
+                                ) -> Iterable[Tuple[str, Union[BlockNumber, int, None, bytes, Address, Hash32]]]:  # noqa: E501
```
cburgdorf: *nitpick*: I have a slight preference to reformat (e.g. put the parameter on the next line) instead of putting a `noqa: E501` on it. Especially if we consider that one day we'll be using black anyway.

newsfragments/1850.bugfix.rst
```diff
@@ -0,0 +1 @@
+Update codebase to more consistently use the ``eth_typing.BlockNumber`` type.
```
cburgdorf: *Nitpick*, this might be better served as `misc` as it's more of a refactoring than a bugfix. At least from the user perspective it's not fixing anything worth showing up in the release notes. Now that we use towncrier we get much longer [release notes](https://trinity-client.readthedocs.io/en/latest/release_notes.html#trinity-0-1-0-alpha-28-2019-09-12) than before which is great but we should also try to keep them as meaningful and focused as we can to not overwhelm readers with less relevant information.

pipermerriam: I marked it as bugfix because in other libraries if you do `from eth.constants import GENESIS_BLOCK_NUMBER` you end up having to do `BlockNumber(GENESIS_BLOCK_NUMBER)` to have it work appropriately with properly typed interfaces.  This fixes that.

eth/tools/builder/chain/builders.py
pipermerriam: generally agreed.  I chose this after having done it the other way and finding the readability to be poor.

PR #1853, opened 2019-09-20
Revert premature optimization for EIP-2294

fubuloubu:
### What was wrong?
There was a suggestion to limit the size of Chain ID to 64 bits, however this was a premature proposal and has since been punted to [EIP-2294](https://GitHub.com/Ethereum/EIPs/issues/2294).

### How was it fixed?
Reverted back to `uint256` validation

#### Cute Animal Picture

koala


carver:
:+1: after updating the tests to check the boundary at the uint256 maximum

PR #1854, opened 2019-09-23
Use new fast blake2 implementation

### LINKED ISSUE #1836: Blake2b Compression Function F is too slow
### What is wrong?

Blake2b compression takes way too long in the all-python implementation. It's now trivial to DoS trinity & py-evm by filling a block with ~~calls to the precompile~~ one call to the precompile, specifying millions of rounds.

See #1818 for rough analysis:
> It takes 25 seconds to run 2 million rounds (which is 2M gas). At that rate, it will take:
> 
> * 100 seconds to run a full 8 million gas block _(just confirmed, it took 91s)_
> * 14 hours to run test vector 8, which has 4 billion rounds
> 
> So it's about 6 times too slow, at a minimum, assuming 15s blocks.

### How can it be fixed

Ideally, write a python wrapper around the C implementation that is already distributed the stdlib blake2b module.

It should meet the same API as the current implementation:
https://github.com/ethereum/py-evm/blob/7696779d2f17e4000198d14d5c017f705bb6cf67/eth/_utils/blake2/compression.py#L71-L76

It is *not* a requirement that everything beneath that ^ API is in C. If you want a python layer that prepares & returns the appropriate types, that's fine/great. The primary requirement is that it runs fast enough.

It should run 8 million rounds in under 8 seconds (of course, faster is better).

A working PR should include a hypothesis test that shows equivalence between the all-python implementation and the new API. It should also include a test showing that it runs fast enough, maybe run 100k rounds to make sure they complete in under 0.1s.

*Note that some of the nomenclature on the types and variable names may be a little funky, since the author (@carver) was not familiar with the internals of Blake2. Bonus points for identifying better names, with citations. (Especially for variables and types like message, block, and state)*

carver:
### What was wrong?

Closes #1836 

### How was it fixed?

- Upgrade pytest.
- Add new blake2b library
- Modify parameter expansion for slightly changed blake2 compression input (it changed from `bytes` to decoded integers)
- Ran CI test to show the long test working
- Took the test back out

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


carver:
It's good to know that the 4B round (the maximum) blake2 example works with the new library. It does take >90s to run, so I'm thinking I'll probably remove it again before merging it. Now I can update the skip notes to say: that it has been tested, which version of `blake2b-py` it was tested against, how long it took, etc.

carver:
Given that the whole tox run takes about a minute, I'll pull the 90s test back out for now...

PR #1855, opened 2019-09-24
Add Clique Consensus Engine

cburgdorf:
### What was wrong?

Py-EVM should be able to sync chains that use *Clique* consensus (e.g. used by Görli and Rinkeby)

### How was it fixed?

This adds a new class `CliqueConsensus` with a `validate_seal` method that validates headers according to the clique consensus rules.

Notice, that some details of this feel a little hacky and I expect this to take some rounds of review to got to an architecture that feels sound.

I'll add some comments in line.

### Post-merge Todo:

- add issue regarding pyristent libary
- add issue regarding hypthesis tests

#### Cute Animal Picture


eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,189 @@
...
+            block_number, vm = pair
+            vm_class = vm.configure(
+                block_extra_data_max_bytes=65535,
+                validate_seal=lambda header: self.validate_seal(header),
+                configure_header=self.configure_header
+            )
+
+            yield block_number, vm_class
+
+    def construct_wrong_turn_message(self,
```
pipermerriam: I think "out_of_turn" is more intuitive verbage for this (wrong turn in english conveys more like you took a wrong turn in your car).

eth/consensus/clique/encoding.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+
+def decode_vote(vote: bytes) -> Vote:
+
+    container = rlp.decode(vote, sedes=rlp.sedes.CountableList(rlp.sedes.binary))
+    signer = rlp.decode(container[0], sedes=rlp.sedes.binary)
+    block_number = rlp.decode(container[1], sedes=uint256)
+    subject = rlp.decode(container[2], sedes=rlp.sedes.binary)
+    action = rlp.decode(container[0], sedes=rlp.sedes.binary)
```
pipermerriam: is the `0` index here repeated on purpose?

Also, do you know about using `rlp.decode(some_bytes, sedes.List((sedes.binary, sedes.big_endian_int)))` to decode something into the rich values in one-go.

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,189 @@
...
+    if import_block is not None:
+        try:
+            coinbase = get_signer_address(import_block.header)
+        except Exception:
+            coinbase = self.coinbase
+    else:
+        coinbase = self.coinbase
+
+    return ExecutionContext(
+        coinbase=coinbase,
```
pipermerriam: rather than overload the `coinbase` I think it would be better to add a new field `ExecutionContext.fee_recipient`.  This value would be optional, defaulting to the `coinbase` unless explicitely provided.  I think you'll run into problems since what you're doing will change the value returned by the `COINBASE` opcode during EVM execution which may or may not be the desired behavior.

eth/consensus/clique/encoding.py
cburgdorf: Uh, no that is a bug. Notice that I don't actually use the rlp encoding for the snapshots that much yet which is probably the reason why this went unnoticed so far.

>Also, do you know about using rlp.decode(some_bytes, sedes.List((sedes.binary, sedes.big_endian_int))) to decode something into the rich values in one-go.

Sweet, I did not!

eth/consensus/clique/clique.py
cburgdorf: >I think you'll run into problems since what you're doing will change the value returned by the COINBASE opcode during EVM execution which may or may not be the desired behavior.

That's a good point :+1:

eth/consensus/clique/clique.py
cburgdorf: The Clique protocol has a concept of blocks that are coming "in turn" or "out of turn" and both are valid (just getting a different difficulty assigned). This function is constructing an error message for when difficulty and "turness" don't match up. So using "out of turn" might be confusing here as "out of turn" isn't necessarily wrong in Clique consensus.

Would `construct_turn_error_message` feel natural as well?

eth/consensus/clique/_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    NONCE_DROP,
+    SIGNATURE_LENGTH,
+    VANITY_LENGTH,
+)
+from .datatypes import (
+    Snapshot,
+)
+
+
+@to_list
```
pipermerriam: Is this intentionally using a mutable data structure?

eth/consensus/clique/_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    return signature_header.hash
+
+
+def get_block_signer(header: BlockHeaderAPI) -> Address:
+    """
+    Return the address of the signer of the ``header``.
+    """
+
+    if len(header.extra_data) < SIGNATURE_LENGTH:
+        raise Exception("header.extra_data too short to contain signature")
```
pipermerriam: Seems like this check should also be in place within `get_signature_hash`, otherwise if the `extra_data` is too short, the signature hash will just be the header with empty `extra_data`.

eth/consensus/clique/_utils.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def is_in_turn(signer: Address, snapshot: Snapshot, header: BlockHeaderAPI) -> bool:
+    """
+    Return ``True`` if the block was produced *in turn*, otherwise return ``False``.
+    """
+    sorted_signers = snapshot.get_sorted_signers()
+    offset = 0
+    while offset < len(sorted_signers) and sorted_signers[offset] != signer:
+        offset += 1
```
pipermerriam: Is this the same as

```try
    offset = sorted_signers.index(signer)
except ValueError:
    return False
else:
    return header.block_number % len(sorted_signers) == offset
```

pipermerriam: I suspect `Sequence.index` is pretty fast, but the `bisect` module also has tools for finding the right index in pre-sorted data structures.

eth/consensus/clique/_utils.py
```diff
@@ -0,0 +1,154 @@
...
+def sign_block_header(header: BlockHeaderAPI, private_key: PrivateKey) -> BlockHeaderAPI:
+    signature_hash = get_signature_hash(header)
+    signature = private_key.sign_msg_hash(signature_hash)
+
+    signers = get_signers_at_checkpoint(header)
+    signed_extra_data = header.extra_data[:VANITY_LENGTH]
+    for signer in signers:
+        signed_extra_data += signer
+
+    signed_extra_data += signature.to_bytes()
```
pipermerriam: I don't know if it actually saves the object copies, but you can remove the mutability here with:

```python
signed_extra_data = b''.join(
    header.extra_data[...],
    b''.join(signers),
    signature.to_bytes(),
)
```

Alternatively you could use a `bytearray` which is inherently mutable and then cast it to `bytes` once's you've built up the state.

eth/consensus/clique/_utils.py
```diff
@@ -0,0 +1,154 @@
...
+    if header.timestamp > int(time.time()):
+        raise ValidationError(f"Invalid future timestamp: {header.timestamp}")
+
+    at_checkpoint = is_checkpoint(header.block_number, epoch_length)
+
+    if at_checkpoint and header.coinbase != ZERO_ADDRESS:
+        raise ValidationError(f"Coinbase must be {encode_hex(ZERO_ADDRESS)} on checkpoints")
+
+    if header.nonce != NONCE_AUTH and header.nonce != NONCE_DROP:
+        raise ValidationError("Invalid nonce")
```
pipermerriam: Worth including the actual value in most of these error messages.

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,173 @@
...
+        block_number=self.block_number,
+        difficulty=self.difficulty,
+        gas_limit=self.gas_limit,
+        prev_hashes=prev_hashes,
+        chain_id=chain_context.chain_id,
+        fee_recipient=fee_recipient
+    )
+
+
+BlockHeader.create_execution_context = create_execution_context  # type: ignore
```
pipermerriam: eeek!  just leaving comment here because this will be problematic if left in.

eth/consensus/clique/encoding.py
```diff
@@ -0,0 +1,136 @@
...
+        [
+            snapshot.block_hash,
+            [signer for signer in snapshot.signers],
+            [encode_vote(vote) for vote in snapshot.votes],
+            [
+                encode_address_tally_pair((address, tally))
+                for address, tally in snapshot.tallies.items()
+            ],
+        ],
+        sedes=rlp.sedes.List(
```
pipermerriam: this code should be easier to read if these `sedes` values are set as constants at the module level.

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,173 @@
...
+    @to_tuple
+    def amend_vm_configuration(self, config: VMConfiguration) -> Iterable[VMFork]:
+        for pair in config:
+            block_number, vm = pair
+            vm_class = vm.configure(
+                block_extra_data_max_bytes=65535,
+                validate_seal=lambda header: self.validate_seal(header),
+                configure_header=self.configure_header,
+                _assign_block_rewards=lambda _, __: None,
+            )
```
cburgdorf: @pipermerriam while you are looking at this, can I drive your attention towards this :point_up: This is basically the heart of how the clique consensus currently fits into the existing architecture. We basically just overwrite a few methods on each VM. This needs to happen early in the process. Notice that these patched methods are operating on an *instance* of `CliqueConsensus` which in turn depends on a database instance.

I'm not sure yet how to replace this with something that doesn't look and feel like monkey patching :see_no_evil:

pipermerriam: We don't have a solid pattern for doing stuff like this but it would be nice if we could figure one out.

Here's the other places that I know of where we do something similar:

https://github.com/ethereum/py-evm/blob/e8f7039b33177118139bcedcf5570a6bcdecea5e/eth/tools/builder/chain/builders.py#L271-L274

https://github.com/ethereum/py-evm/blob/e8f7039b33177118139bcedcf5570a6bcdecea5e/eth/tools/builder/chain/builders.py#L303-L314

All of them follow roughly the same pattern which suggests that this *could* be made generic, but it would still be a bunch of dynamic class creation which isn't ideal...

eth/consensus/clique/_utils.py
cburgdorf: Nope, just a left over from the make-it-work stages. I'll change that to `to_tuple`

eth/consensus/clique/_utils.py
cburgdorf: Sounds correct, in fact, we can just *move* the check there and then reorder the lines here so that `signature_hash = get_signature_hash(header)` comes first.

eth/consensus/clique/_utils.py
cburgdorf: Yep, that should do it. I think I literally just took the Go code here and translated it one by one to Python which resulted in this less idiomatic code.

eth/consensus/clique/_utils.py
cburgdorf: Looks good :+1:

eth/consensus/clique/clique.py
cburgdorf: aYep, I know...but what's the alternative? Maybe `create_execution_context` should be moved (and changed to take in a header) to the VM and then this can be overwritten as part of the `vm.configure` step?

pipermerriam: Probably yes.

eth/consensus/clique/clique.py
cburgdorf: Maybe it's ok to leave it as is for this PR at least.

eth/vm/base.py
```diff
@@ -182,7 +184,8 @@ def apply_transaction(self,
     @staticmethod
     def create_execution_context(header: BlockHeaderAPI,
                                  prev_hashes: Iterable[Hash32],
-                                 chain_context: ChainContextAPI) -> ExecutionContextAPI:
+                                 chain_context: ChainContextAPI,
+                                 import_block: BlockAPI = None) -> ExecutionContextAPI:
```
cburgdorf: Adding `import_block` here feels a bit weird given that there's already `header` in this API. There's has to be a cleaner way to do this but let me try to elaborate what the problem is:

When we import a block, we need to determine the signer of the block so that we know where to send the tx fees. Holding on to this information is the job of the `ExecutionContextAPI`. However, to retrieve the signer we need to have access to the full header that is being imported.

The problem is that our current flow just cherry picks the attributes from the import block which are needed to satisfy the PoW flow, leaving us with a header that we can not use to retrieve the signer from.

eth/vm/base.py
```diff
@@ -295,6 +298,7 @@ def import_block(self, block: BlockAPI) -> BlockAPI:
         self._block = self.get_block().copy(
             header=self.configure_header(
                 coinbase=block.header.coinbase,
+                difficulty=block.header.difficulty,
```
cburgdorf: We currently do not pass on the `difficulty` because we rely on calculating it ourselves and I'm sure we *could*  also do that in the Clique scheme (I had that at some point) but it seems wasteful to do that here because at the end of the block import there's a call to `validate_seal` anyway at which point we'll check if the difficulty matches our expectations.

Notice that, even when we add this here, this does not meddle with the current PoW scheme which will stubbornly continue to ignore the provided difficulty and calculate it instead.

cburgdorf:
@pipermerriam I did a bunch of cleanups, added more tests and documentation. I also created some comments on code paths that I believe deserve special attention. I consider it ready for review now :)

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,171 @@
...
+    # In Clique consensus, the tx fee goes to the signer
+    fee_recipient = None
+    if import_block is not None:
+        try:
+            fee_recipient = get_block_signer(import_block.header)
+        except Exception:
+            pass
+
+    return ExecutionContext(
+        coinbase=header.coinbase,
```
cburgdorf: Oh wait, I just found this in the spec.

image

It seems that Clique explicitly says that we should set the `signer` as the `coinbase` to also return the `signer` for the `COINBASE` opcode.

Do we still want to introduce `fee_recipient` to allow more fine grained control nonetheless?

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,171 @@
...
+    @to_tuple
+    def amend_vm_configuration(self, config: VMConfiguration) -> Iterable[VMFork]:
+        """
+        Amend the given ``VMConfiguration`` to operate under the rules of Clique consensus.
+        """
+        for pair in config:
+            block_number, vm = pair
+            vm_class = vm.configure(
+                extra_data_max_bytes=65535,
+                validate_seal=lambda header: self.validate_seal(header),
```
pipermerriam: can you see if you can eliminate the lambda using `staticmethod(self.validate_seal)`

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        if header.nonce in {NONCE_AUTH, NONCE_DROP}:
+            signer = get_block_signer(header)
+            # Clear any votes from the signer regarding the subject that is voted on
+            for vote in snapshot.votes:
+                if vote.signer == signer and vote.subject == header.coinbase:
+                    self.uncast(snapshot, vote.subject, vote.action)
+                    snapshot.votes.remove(vote)
+                    break
+
+            action = VoteAction.NOMINATE if header.nonce == NONCE_AUTH else VoteAction.KICK
```
pipermerriam: same here, use `action = VoteAction(header.nonce)`

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+
+            if self.cast(snapshot, header.coinbase, action):
+                vote = Vote(
+                    signer=signer,
+                    block_number=header.block_number,
+                    subject=header.coinbase,
+                    action=action
+                )
+                snapshot.votes.append(vote)
```
pipermerriam: You might consider using a datatype from [`pyrsistent`](https://pypi.org/project/pyrsistent/) instead and using their Evolver API to do this mutation.

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        snapshot = mutable_snapshot.get_frozen_clone()
+        self._snapshots[snapshot.block_hash] = snapshot
+        return snapshot
+
+    def persist_snapshot(self, snapshot: Snapshot) -> None:
+        """
+        Persist the given snapshot to the database.
+        """
+        key = make_snapshot_lookup_key(snapshot.block_hash)
+        with self._chain_db.db.atomic_batch() as db:
```
pipermerriam: I don't think we need the `atomic_batch` here since this only sets a single value (which is an atomic action)

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        try:
+            encoded_key = self._chain_db.db[key]
+        except KeyError as e:
+            raise SnapshotNotFound(
+                f"Can not get on-disk snapshot for {block_hash}"
+            )
+        else:
+            return decode_snapshot(encoded_key)
+
+    def is_valid_vote(self,
```
pipermerriam: This function doesn't use `self` and could be relegated to a utils module or a top level utility function in this module depending on where it is used.  You could even make it a method on the `VoteAction` enum.

```python
action.is_valid_for(subject, signers)
```

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        Check if a vote is semantically valid, meaning it is either voting against an existing
+        member or voting in favor of an address that is not yet a member.
+        """
+        signer_exists = subject in signers
+        signer_is_kicked = signer_exists and action is VoteAction.KICK
+        signer_is_nominated = not signer_exists and action is VoteAction.NOMINATE
+        return signer_is_kicked or signer_is_nominated
+
+    def cast(self, snapshot: MutableSnapshot, subject: Address, action: VoteAction) -> bool:
```
pipermerriam: Maybe make this `cast_vote` for more explicit naming.  The bare `cast` had me confused for a moment.

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        if not self.is_valid_vote(snapshot.signers, subject, action):
+            return False
+
+        if subject not in snapshot.tallies:
+            snapshot.tallies[subject] = Tally(action)
+
+        snapshot.tallies[subject] = snapshot.tallies[subject].upvote()
+        return True
+
+    def uncast(self, snapshot: MutableSnapshot, subject: Address, action: VoteAction) -> bool:
```
pipermerriam: maybe `retract_vote`?

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,311 @@
...
+        signer_is_kicked = signer_exists and action is VoteAction.KICK
+        signer_is_nominated = not signer_exists and action is VoteAction.NOMINATE
+        return signer_is_kicked or signer_is_nominated
+
+    def cast(self, snapshot: MutableSnapshot, subject: Address, action: VoteAction) -> bool:
+        """
+        Cast a vote on a ``MutableSnapshot``.
+        """
+        if not self.is_valid_vote(snapshot.signers, subject, action):
+            return False
```
pipermerriam: This pattern requires call sites to check return values which is an error prone mechanism.  Did you consider using exception raising instead to require call sites to handle both cases?

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,171 @@
...
+def create_execution_context(header: BlockHeaderAPI,
+                             prev_hashes: Iterable[Hash32],
+                             chain_context: ChainContext,
+                             import_block: BlockHeaderAPI=None) -> ExecutionContext:
+
+    # In Clique consensus, the tx fee goes to the signer
+    fee_recipient = None
+    if import_block is not None:
+        try:
+            fee_recipient = get_block_signer(import_block.header)
```
pipermerriam: Why can't this be retrieved from `header`?  Is `header` different from `import_block.header`?

eth/vm/base.py
pipermerriam: See my comment above about confusion between how `import_block.header` and `header` are different....

eth/consensus/clique/clique.py
cburgdorf: Cool :+1:

eth/consensus/clique/_utils.py
```diff
@@ -0,0 +1,161 @@
...
+
+    if len(header.extra_data) < VANITY_LENGTH + SIGNATURE_LENGTH:
+        raise ValidationError("Missing signature in extra_data")
+
+    signers_length = len(header.extra_data) - VANITY_LENGTH - SIGNATURE_LENGTH
+
+    if not at_checkpoint and signers_length != 0:
+        raise ValidationError("Non-checkpoint header must not contain list of signers")
+
+    if at_checkpoint and signers_length % COMMON_ADDRESS_LENGTH != 0:
```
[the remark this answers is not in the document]
cburgdorf: Sure, I'll do that.

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,171 @@
...
+        self._epoch_length = epoch_length
+        self._chain_db = ChainDB(base_db)
+        self._header_cache = HeaderCache(self._chain_db)
+        self._snapshot_manager = SnapshotManager(
+            self._chain_db,
+            self._header_cache,
+            self._epoch_length,
+        )
+
+    def _construct_turn_error_message(self,
```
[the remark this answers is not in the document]
cburgdorf: Agreed.

eth/consensus/clique/clique.py
cburgdorf: Works :)

eth/consensus/clique/encoding.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+
+def decode_tally(tally: bytes) -> Tally:
+    action_binary, votes = rlp.decode(
+        tally,
+        sedes=TALLY_SEDES,
+    )
+
+    action = VoteAction.NOMINATE if action_binary == VoteAction.NOMINATE.value else VoteAction.KICK
```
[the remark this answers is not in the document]
cburgdorf: Oh nice, I wasn't aware that this works!

eth/consensus/clique/snapshot_manager.py
cburgdorf: I like the idea of making this a method on the `VoteAction` directly :+1:

eth/consensus/clique/snapshot_manager.py
cburgdorf: I copied this from an existing implementation (geth/nethermind). I know it goes against our usual way of how to approach validation so I changed it into a `validate_for` method that raises an exception. 

That said (and I remember we have discussed this before but I don't remember your answer) I'm still not sold on this part:

>This pattern requires call sites to check return values which is an error prone mechanism.

With exception raising call sites need to wrap things in an `try` / `except` plus pick the right kind of exception to check for. I'm not sure if it really is less error prone.

In my pre-snake-charmer life, I've always tried to follow the advice to only use exceptions for *exceptional* cases. A failed validation is nothing exceptional for a method where the only purpose is validation. It either is valid or it isn't. Both things are the expected outcome in this scenario.

This also seems in line with the answers given here: https://softwareengineering.stackexchange.com/questions/330824/function-returning-true-false-vs-void-when-succeeding-and-throwing-an-exception

But I would like to learn or be reminded of your argument of exceptions being less error prone.

Notice, I'm not arguing to keep this method (I already changed it) or change our policy, I just want to be reminded of the arguments against using return values for validation.

tests/core/consensus/test_clique_consensus.py
```diff
@@ -0,0 +1,453 @@
...
+    next_header = make_next_header(head, ALICE_PK, RON, NONCE_AUTH)
+
+    with pytest.raises(ValidationError, match='Unknown ancestor'):
+        clique.get_snapshot(next_header)
+
+
+def test_import_block(paragon_chain, clique):
+
+    vm = paragon_chain.get_vm()
+    from tests.core.helpers import new_transaction
```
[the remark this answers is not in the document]
cburgdorf: >I would advocate avoiding writing new code that is dependent on `helpers`

Mmh, in that case, how about moving the `new_transaction` utility somewhere into `eth.tools`. Because I think there's value in using a real transaction here instead of a mocked one. E.g. the real one is produced based on the actual VM (so taking the transaction class of the VM into account) and will also read the nonce out of that VM.

eth/consensus/clique/snapshot_manager.py
voith: FWIW, @carver had a similar argument over [here](https://github.com/ethereum/trinity/pull/1237#discussion_r341405779).

tests/core/consensus/test_clique_consensus.py
```diff
@@ -0,0 +1,453 @@
...
+    # Ensure the Tally and all pending votes are removed
+    assert RON not in snapshot.tallies
+    assert not has_vote_to(RON, snapshot.votes)
+    assert not has_vote_from(BOB, snapshot.votes)
+    assert not has_vote_from(ALICE, snapshot.votes)
+
+
+def test_removes_all_pending_votes_after_kick(clique):
+
+    ALICE_FRIEND = b'weird_buddy_of_alice'
```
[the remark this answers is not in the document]
cburgdorf: Alright, this is now: `ALICE_FRIEND = PublicKeyFactory().to_canonical_address()`

tests/core/vm/test_clique_validation.py
```diff
@@ -0,0 +1,65 @@
...
+    ),
+)
+def test_can_validate_header(clique, VM, header, previous_header, valid):
+    CliqueVM = VM.configure(
+        extra_data_max_bytes=128,
+        validate_seal=lambda header: clique.validate_seal(header),
+    )
+    if valid:
+        CliqueVM.validate_header(header, previous_header, check_seal=True)
+    else:
```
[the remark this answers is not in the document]
cburgdorf: Remove it for now.

tests/core/consensus/test_clique_encoding.py
```diff
@@ -0,0 +1,72 @@
...
+
+@pytest.mark.parametrize(
+    'val, meant_to_be, encode_fn, decode_fn',
+    (
+        (ADDRESS_TALLY_PAIRS[0], tuple, encode_address_tally_pair, decode_address_tally_pair),
+        (DUMMY_VOTE_1, Vote, encode_vote, decode_vote),
+        (DUMMY_VOTE_2, Vote, encode_vote, decode_vote),
+        (SNAPSHOT_1, Snapshot, encode_snapshot, decode_snapshot),
+        (TRUMP_TALLY, Tally, encode_tally, decode_tally),
+        (YANG_TALLY, Tally, encode_tally, decode_tally)
```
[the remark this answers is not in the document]
cburgdorf: I agree that would be useful. I haven't worked with `hypothesis` yet so maybe I could create an issue for that and come back to it later?

eth/vm/base.py
cburgdorf: It seem unavoidable to give access to the full block/header that is being imported to support a wide range of use cases (including the one at hand).

I inlined the call to `build_state` so that we can leave that API untouched (instead of just passing the `import_block` through).

I still don't like this part but I'm not sure about the right way to approach this.

cburgdorf:
@pipermerriam This is ready for another round of review.

eth/consensus/clique/snapshot_manager.py
pipermerriam:
```python
def cast_vote(vote):
    if vote.is_valid():
        ...
        return True
    else:
        return False

def do_votes():
    # alice allowed to vote
    # bob isn't allowed to vote
    for vote in (alice, bob):
        cast_vote(alice)
```

Someone looking at `do_vote` wouldn't realize that they were not calling `cast_vote` correctly (since they aren't looking at the return value).  This code would not throw an error and would run as-if it was working even though it is broken.

```python
def cast_vote(vote):
    if vote.is_valid():
        ...
    else:
        raise ValidationError(...)
```

This model will blow up, making it really clear that the caller is not correctly using `cast_vote` and quickly showing the mis-use because an exception occurs.

The short argument is:

- When you forget to check a return value you end up with a silent failure.
- When you forget to handle an exception you get a loud exception.

tests/core/consensus/test_clique_consensus.py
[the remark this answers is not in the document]
pipermerriam: @cburgdorf I'm still inclined to push for factory style code rather than helper functions but :+1: for `eth.tools`

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,173 @@
...
+    def create_execution_context(header: BlockHeaderAPI,
+                                 prev_hashes: Iterable[Hash32],
+                                 chain_context: ChainContext,
+                                 import_block: BlockHeaderAPI=None) -> ExecutionContext:
+
+        # In Clique consensus, the tx fee goes to the signer
+        coinbase = header.coinbase
+        if import_block is not None:
+            try:
+                coinbase = get_block_signer(import_block.header)
```
carver: This is a drive-by thought without reading the full PR: maybe there could be a special header class where `header.coinbase = get_block_signer(header)`. But I'm also missing something about why the block signer can't be pulled from the original header.

pipermerriam: I'm :+1: on moving this direction but lets save it for future work.  This would involve some level of plumbing across the `ChainAPI` and `VirtualMachineAPI` to allow specifying custom header classes.

eth/consensus/clique/clique.py
```diff
@@ -0,0 +1,171 @@
...
+        self._snapshot_manager.apply(snapshot, header)
+
+        if signer not in authorized_signers:
+            raise ValidationError(
+                f"Failed to validate {header}."
+                f"Signer {encode_hex(signer)} not in {authorized_signers}"
+            )
+
+        # Just *something* to kick off a continously cache gc
+        if int(time.perf_counter()) % 10 == 0:
```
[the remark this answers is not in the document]
pipermerriam: Pinging for a response to this comment.  Curious what other things were tried and what keeps us from using a different mechanism for doing this.

eth/consensus/clique/datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+class MutableSnapshot(NamedTuple):
+    """
+    Like ``Snapshot`` but mutable as a work in progress representation.
+    """
+    signers: List[Address]
+    block_hash: Hash32
+    votes: List[Vote]
+    tallies: Dict[Address, Tally]
+
+    def get_frozen_clone(self) -> Snapshot:
```
pipermerriam: *nitpick*

Naming disparity bugging me.  `freeze/thaw`, `mutable/immutable`, `frozen/thawed`,  maybe you can settle on something that uses the same terminology across the two class's?

eth/consensus/clique/datatypes.py
```diff
@@ -0,0 +1,125 @@
...
+        )
+
+    def get_sorted_signers(self) -> List[Address]:
+        """
+        Return the sorted list of signers.
+        """
+        return sorted(self.signers)
+
+
+class MutableSnapshot(NamedTuple):
```
pipermerriam: Just noting that if you used a [`pyristent`](https://pypi.org/project/pyrsistent/) data type that we wouldn't need this class.

eth/consensus/clique/header_cache.py
```diff
@@ -0,0 +1,51 @@
...
+    they are persisted in the database. This is necessary because validation in Clique depends
+    on looking up past headers which may not be persisted at the time when they are needed.
+    """
+
+    logger = get_extended_debug_logger('eth.consensus.clique.header_cache.HeaderCache')
+
+    def __init__(self, chaindb: ChainDatabaseAPI) -> None:
+        self._chaindb = chaindb
+        self._cache: Dict[Hash32, BlockHeader] = {}
+        self._gc_treshold = 1000
```
pipermerriam:
```suggestion
        self._gc_threshold = 1000
```

eth/consensus/clique/snapshot_manager.py
```diff
@@ -0,0 +1,297 @@
...
+        After we've found a suitable base snapshot, apply all headers from after the
+        base snapshot up to the header of ``block_hash`` to create the requested snapshot.
+        """
+        current_header = header = self._lookup_header(block_hash)
+
+        if is_checkpoint(header.block_number, self._epoch_length):
+            return self._create_snapshot_from_checkpoint_header(current_header)
+
+        parents = []
+        while True:
```
pipermerriam: Can this become something non-infinite?  Set an upper bound and raise an exception if it exceeds that number of iterations (ideally a value we should never realistically hit).

eth/vm/base.py
pipermerriam: Is `import_block.header == header`?

eth/vm/base.py
pipermerriam: Not actionable at this stage, but I think this suggests that we need to figure out the abstraction for some sort of generic `ConsensusAPI` or something.  Ideally, rather than ending up in this case where we're doing things to skirt around the POW stuff, we should maybe be looking at what subset of these APIs deserve to be extracted so that implementing Clique/POW/POS/etc are fully independent of each other without having to do funky stuff to compensate for parent class logic that applies to a different consensus engine.

eth/consensus/clique/datatypes.py
cburgdorf: Makes sense, renamed it to `get_immutable_clone`.

eth/consensus/clique/datatypes.py
cburgdorf: I like the idea of using pyristent. I'm just not sure if this single place we we could use it would justify the extra dependency. That said, I'll open up an issue to adopt it and link to this thread. And then we can check if there might be other areas in the code base that would benefit from it and may help to justify the extra dependency.

eth/vm/base.py
cburgdorf: It's not. The `block` that we import gets dropped in `import_block` and we just copy a few things over into our *in progress* block.

https://github.com/ethereum/py-evm/blob/83d50fa2d9aae931775af7ca2c5bac60ad811566/eth/vm/base.py#L285-L324

However, now that I inlined the call to `self.build_state` it occurred to me that I could use the importing block for the header that we pass to `create_execution_context` but use the *in progress* block to read the `state_root` to build the state from. That allows us to to drop the `import_block` here :tada:

eth/vm/base.py
cburgdorf: Yes, I agree.

eth/consensus/clique/clique.py
cburgdorf: >But I'm also missing something about why the block signer can't be pulled from the original header.

The `header` in `create_execution_context` wasn't the *original* header but the *in progress* header that we keep on the vm. I changed that now to pass in the one that is passed into `import_block` so that we got rid of that extra parameter.

eth/consensus/clique/snapshot_manager.py
cburgdorf: We could add this but let my try arguing against it.

1. There's no chance this code could run infinitely in practice. It either exits earlier or travels up to the next checkpoint header (ultimately that would be genesis) **or** it raises an `ValidationError("Unknown ancestor %s", block_hash)`

2. It's pretty comparable to [other existing code paths](https://github.com/ethereum/py-evm/blob/master/eth/db/header.py#L327-L344) that use that pattern.

PR #1864, opened 2019-10-29
Make the max bytes of extra_data configurable

cburgdorf:
### What was wrong?

The validation check for the maximum bytes of the `extra_data` field is currently hardcoded. It needs to be configurable to allow different consensus schemes such as Clique which have a different opinion on the max bytes of this field.

### How was it fixed?

Introduce a `extra_data_max_bytes` class var which defaults to `32` but can be overwritten.

#### Cute Animal Picture

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -3,6 +3,7 @@ import itertools
 import logging
 from typing import (
     Any,
+    ClassVar,
     Iterable,
     Iterator,
     Optional,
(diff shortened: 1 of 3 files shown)


eth/vm/base.py
```diff
@@ -90,6 +90,7 @@ class VM(Configurable, VirtualMachineAPI):
         - ``_state_class``: The :class:`~eth.abc.StateAPI` class used by this VM for execution.
     """
     block_class: Type[BlockAPI] = None
+    extra_data_max_bytes = 32
```
pipermerriam: Maybe type this as a `ClassVar`

cburgdorf: Good idea :+1:

PR #1866, opened 2019-10-30
Revert storage after selfdestruct

### LINKED ISSUE #1865: Revert after selfdestruct causes wrong storage value readout
### What is wrong?

If a `SELFDESTRUCT` is called and then reverted, the storage values on that account were not correctly revived to there original values. This especially comes up during gas estimation.

### How can it be fixed

If we came up with a novel test case (seems surprising, but I suppose possible) it should probably be added to ethereum/tests.

PR about to be posted

carver:
### What was wrong?

Fix #1865 

### How was it fixed?

Two fixes for two different scenarios:
- If creating the storage DB for the first time just to wipe it, then make sure it has the original storage root hash.
- If completely resetting the journal, make sure that any current `clear` (like a storage wipe) gets fully reset (which means it should read through to the underlying if the key is missing from the journal) See more commentary in the PR

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/core/vm/test_vm_state.py
+++ b/tests/core/vm/test_vm_state.py
@@ -74,3 +74,25 @@ def test_set_storage_input_validation(state, address, slot, new_value):
 def test_delete_storage_input_validation(state):
     with pytest.raises(ValidationError):
         state.delete_storage(INVALID_ADDRESS)
+
+
+@pytest.mark.parametrize('read_storage_before_snapshot', [True, ])
+def test_revert_selfdestruct(state, read_storage_before_snapshot):
+    state.set_storage(ADDRESS, 1, 2)
+    state.persist()
+
+    if read_storage_before_snapshot:
+        assert state.get_storage(ADDRESS, 1) == 2
+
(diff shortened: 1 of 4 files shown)


tests/core/vm/test_vm_state.py
```diff
@@ -74,3 +74,31 @@ def test_set_storage_input_validation(state, address, slot, new_value):
 def test_delete_storage_input_validation(state):
     with pytest.raises(ValidationError):
         state.delete_storage(INVALID_ADDRESS)
+
+
+# TODO candidate to build a core ethereum test (surprising that this case wasn't covered, already!)
```
cburgdorf: Should we turn this into an issue at ethereum/tests and link to this PR?

pipermerriam: cc @holiman

carver: I just wrote it up in #1867

holiman:
Interesting!  A related case is how these behave with eip-2200, net sstore gas changes. So if you do a last sstore instead of sload before returning from the selfdestructed one.

PR #1873, opened 2019-11-12
Improve string representation of blocks

cburgdorf:
### What was wrong?

I noticed today that the string representation of blocks looks a bit unusual compared to the other RLP types. E.g. the string representation of a header would be `<BlockHeader #2 b495a1d7>` but the string representation of that same block would just be `Block #2`. It doesn't come with the angle brackets nor does it print the hash.

### How was it fixed?

Changed it so that it now would print `<FrontierBlock #2 b495a1d7>`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/rlp/blocks.py
+++ b/eth/rlp/blocks.py
@@ -2,6 +2,10 @@ from typing import (
     Type
 )
 
+from eth_utils import (
+    encode_hex,
+)
+
 from eth._utils.datatypes import (
     Configurable,
 )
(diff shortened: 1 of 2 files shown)


cburgdorf:
Actually, I'm not sure, now that I read up on the differences of `str` and `repr` I think, maybe `BaseBlock` has it right and `BlockHeader` has it wrong?

PR #1874, opened 2019-11-18
Refactor Clique Consensus to handle initialization internally

cburgdorf:
### What was wrong?

The current clique implementation requires attaching a DB to the chain *class* (not chain *instance*). This is problematic (see [discussion](https://github.com/ethereum/trinity/pull/1196#discussion_r344919981) in Trinity)

It is also flawed as it is leaking a stateful consensus engine instance on different chain instances that should be independent.

For instance:

```python

chain = GoerliChain()
second_chain = GoerliChain()

# this may add a signer which is now leaking to the second_chain instance
chain.validate_seal(...)  
```

### How was it fixed?

1. This introduces a `VirtualMachineModifierAPI` interface defined as

```python
class VirtualMachineModifierAPI(ABC):

    @abstractmethod
    def __init__(self, base_db: AtomicDatabaseAPI) -> None:
        ...

    @classmethod
    @abstractmethod
    def amend_vm_configuration_for_chain_class(
            cls,
            config: VMConfiguration) -> Tuple[Tuple[BlockNumber, Type[VirtualMachineAPI]], ...]:
        """
        Make amendments to the ``VMConfiguration`` such as redefining its methods for different
        consensus schemes.
        """
        ...

    @abstractmethod
    def amend_vm_class_for_chain_instance(
            self,
            vm_class: Type[VirtualMachineAPI]) -> Type[VirtualMachineAPI]:
        """
        Make amendments to a VM class that are not supposed to be shared across different chain
        instances.
        """
        ...
```

The `amend_vm_configuration_for_chain_class` defines a `classmethod` which can perform all kind of class relevant actions on the `VMConfiguration`. This includes things like setting `extra_data_max_bytes` to `65535` which is safe to share across all `GoerliChain` instances.

But it also defines an instance method `amend_vm_class_for_chain_instance` which performs modifications that are **not safe** to share across different chain instances. Such as replacing `validate_seal` with an implementation that is stateful.

2. It also means that `validate_seal` on the VM had to become an instance method which also means that  `validate_chain` on the VM and  `validate_chain` on the `Chain` had to become instance methods. A fact that @carver had also pointed out in the linked Trintiy thread.

3. This also adds  a `PowConsensus` and a `NoProofConsensus` engine and refactors some test helpers.

4. There's a test that proves that stateful consensus engines do in fact not leak across different chain instances.

### To-Do

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -43,6 +43,7 @@ from eth.typing import (
     JournalDBCheckpoint,
     AccountState,
     HeaderParams,
+    VMConfiguration,
 )
 
 
@@ -1638,9 +1639,8 @@ class VirtualMachineAPI(ConfigurableAPI):
     def validate_block(self, block: BlockAPI) -> None:
         ...
 
-    @classmethod
     @abstractmethod
-    def validate_header(cls,
+    def validate_header(self,
                         header: BlockHeaderAPI,
                         parent_header: BlockHeaderAPI,
                         check_seal: bool = True
@@ -1661,9 +1661,8 @@ class VirtualMachineAPI(ConfigurableAPI):
         """
         ...
 
-    @classmethod
     @abstractmethod
-    def validate_seal(cls, header: BlockHeaderAPI) -> None:
+    def validate_seal(self, header: BlockHeaderAPI) -> None:
         ...
 
     @classmethod
@@ -1685,6 +1684,34 @@ class VirtualMachineAPI(ConfigurableAPI):
 
     @abstractmethod
     def state_in_temp_block(self) -> ContextManager[StateAPI]:
+        ...
+
+
+class VirtualMachineModifierAPI(ABC):
+    """
+    Amend a set of VMs for a chain. This allows modifying a chain for different consensus schemes.
+    """
+
+    @abstractmethod
+    def __init__(self, base_db: AtomicDatabaseAPI) -> None:
+        ...
+
+    @classmethod
+    @abstractmethod
+    def amend_vm_configuration_for_chain_class(cls, vm_config: VMConfiguration) -> None:
+        """
+        Make amendments to the ``vm_config`` that are independent of any instance state. These
+        changes are applied across all instances of the chain where this
+        ``VirtualMachineModifierAPI`` is applied on.
+        """
+        ...
+
+    @abstractmethod
+    def amend_vm_for_chain_instance(self, vm: VirtualMachineAPI) -> None:
+        """
+        Make amendments to ``vm`` that are only valid for a specific chain instance. This
+        includes any modifications that depend on stateful data.
+        """
         ...
 
 
@@ -1748,6 +1775,8 @@ class ChainAPI(ConfigurableAPI):
     vm_configuration: Tuple[Tuple[BlockNumber, Type[VirtualMachineAPI]], ...]
     chain_id: int
     chaindb: ChainDatabaseAPI
+    consensus_engine_class: Type[VirtualMachineModifierAPI]
+    consensus_engine: VirtualMachineModifierAPI
 
     #
     # Helpers
@@ -1923,10 +1952,9 @@ class ChainAPI(ConfigurableAPI):
     def validate_uncles(self, block: BlockAPI) -> None:
         ...
 
-    @classmethod
     @abstractmethod
     def validate_chain(
-            cls,
+            self,
             root: BlockHeaderAPI,
             descendants: Tuple[BlockHeaderAPI, ...],
             seal_check_random_sample_rate: int = 1) -> None:
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -45,11 +45,15 @@ from eth.abc import (
     ChainAPI,
     ChainDatabaseAPI,
     VirtualMachineAPI,
+    VirtualMachineModifierAPI,
     ReceiptAPI,
     ComputationAPI,
     StateAPI,
     SignedTransactionAPI,
     UnsignedTransactionAPI,
+)
+from eth.consensus.pow import (
+    PowConsensus,
 )
 from eth.constants import (
     EMPTY_UNCLE_HASH,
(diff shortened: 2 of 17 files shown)


eth/chains/base.py
```diff
@@ -113,6 +117,14 @@ class BaseChain(Configurable, ChainAPI):
     chaindb_class: Type[ChainDatabaseAPI] = None
     vm_configuration: Tuple[Tuple[BlockNumber, Type[VirtualMachineAPI]], ...] = None
     chain_id: int = None
+    consensus_engine: VirtualMachineModifierAPI = None
```
pipermerriam: Looks like these should be added to the `ChainAPI`

cburgdorf: Good point! Fixed

eth/abc.py
```diff
@@ -1688,6 +1689,22 @@ def state_in_temp_block(self) -> ContextManager[StateAPI]:
         ...
 
 
+class VirtualMachineModifierAPI(ABC):
```
cburgdorf: I've been wondering if this should be named something like `ConsensusEngineAPI`. That would be more aligned with how it is used but on the other hand, I'm usually a fan of small focused and therefore very generic interfaces which may be sticked together to create more specialized interfaces. And with that in mind, there's no reason why this should be called `ConsensusEngineAPI` when in reality there's nothing consensus engine specific about it.

tests/core/chain-object/test_chain_retrieval_of_vm_class.py
```diff
@@ -58,24 +58,24 @@ def chaindb(base_db):
 def test_header_chain_get_vm_class_for_block_number(base_db, genesis_header):
     chain = ChainForTesting.from_genesis_header(base_db, genesis_header)
 
-    assert chain.get_vm_class_for_block_number(0) is VM_A
+    assert type(chain.get_vm_class_for_block_number(0)) is type(VM_A)
```
carver: This seems like a bug, or it needs more explanation.

The VM class is already a type. `type(class)` is always `type`, so this seems like a no-op check.

This whole test may have to be re-written to show that the VMs transition appropriately, even though the VMs are now modified from the originally-specified VMs.

carver:
Direction looks solid!

tests/core/chain-object/test_chain_retrieval_of_vm_class.py
cburgdorf: The thing is that `chain.get_vm_class_for_block_number(0) is VM_A` does fail because when we apply a `ConsensusEngine` we are creating a new class by calling `.configure(...)` and overwriting `self.vm_configuration`

```python
cls.vm_configuration = cls.consensus_engine.amend_vm_configuration(cls.vm_configuration)
```

cburgdorf: Btw, we *could* also use something like `setattr` to modify the vm class directly rather than creating new ones but I'm not sure if that's any better.

cburgdorf: >so this seems like a no-op check.

Oh, I just re-read your comment and now I understand! You are correct...damn...mmh. How do I compare these now :thinking:

cburgdorf: Ok, I'll bite the bullet and use `setattr` instead of returning new classes for these amendments. Actually `setattr` seems more in line with the word amendment anyway :sweat_smile:

cburgdorf:
@carver Alright, this has taken more time than I anticipated. I think this is ready for review. There's also a draft PR of the Görli integration that uses this branch: https://github.com/ethereum/trinity/pull/1196

eth/consensus/clique/clique.py
```diff
@@ -85,22 +87,23 @@ def __init__(self, base_db: AtomicDatabaseAPI, epoch_length: int = EPOCH_LENGTH)
...
-            yield block_number, vm_class
+            setattr(vm, 'extra_data_max_bytes', 65535)
+            setattr(vm, 'create_execution_context', staticmethod(cls.create_execution_context))
+            setattr(vm, 'configure_header', configure_header)
+            setattr(vm, '_assign_block_rewards', lambda *_: None)
+
+    def amend_vm_for_chain_instance(self, vm: VirtualMachineAPI) -> None:
+        # `validate_seal` is stateful. We would not want two different instances of GoerliChain
+        # to operate on the same instance of CliqueConsensus. Therefore, this modification needs
+        # to be done for a specic chain instance, not chain class.
```
carver: Getting backed into a corner here, where this seems like the best solution, makes me think we're doing something else wrong at an architectural level.

One avenue to explore is the location of the consensus class. I'm thinking this might all get a lot easier if the `consensus_class` were a feature of the VM instead of the chain. Arguably, that better represents the real world anyway, as the consensus rules might change at a fork. Then the class gets initialized inside the VM. `VM.validate_seal()` could just forward to `VM.consensus_engine.validate_seal()`, etc.

cburgdorf: I was thinking the same especially because `validate_seal` is always going to be overwritten. The thing that I'm unsure about though: There's no clear picture about which other methods expect to be overwritten. Some consensus engine may fiddle with block rewards another might do funky things with uncles. If we take Clique as a driver then we end up formalizing `validate_seal `extra_data_max_bytes`, `create_execution_context`, `configure_header`, `_assign_block_rewards` for the consensus engine whereas `no_proof` and `pow` only need `validate_seal`. 

>Arguably, that better represents the real world anyway, as the consensus rules might change at a fork

But it also becomes really blurry where consensus rules go. I mean, every new VM is a consensus change and that's what VMs are for. Other consensus engines are a weird and barely formalized way of somehow hacking and bending the system (e.g. how `extra_data`, `nonce`, `coinbase` get totally different meaning in Clique) to somehow work. Something tells me that deserves a weird patchy answer and the current APIs at least break it down to a very generic level of hacking things on the class or the instance level.

eth/consensus/clique/clique.py
```diff
@@ -85,22 +87,23 @@ def __init__(self, base_db: AtomicDatabaseAPI, epoch_length: int = EPOCH_LENGTH)
             self._epoch_length,
         )
 
-    @to_tuple
-    def amend_vm_configuration(self, config: VMConfiguration) -> Iterable[VMFork]:
```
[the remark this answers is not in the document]
cburgdorf: I agree but I don't have a good answer. With the old approach you always end up having different classes then those defined in `cls.vm_configuration`. We have tests that assert things like `assert chain.vm_configuration == RopstenChain.vm_configuration` that break because of that. Is there an alternative (apart from formalizing every method that needs to be replaced in an interface and delegating to it)

cburgdorf:
Replaced by #1899

PR #1880, opened 2019-11-21
Add some missing API docs and move docstrings from implementation to interface

cburgdorf:
### What was wrong?

Overall doc coverage should be improved before beta

### How was it fixed?

1. Add missing API docs for ``MiningChain``.
2. Add missing API docs for ``eth.db.*``
3. Add missing API docs for ``Constantinople``, ``Petersburg`` and ``Istanbul`` forks
4. Move all docstrings that aren't overly specific to a particular implementation from
the implementation to the interface. This has the effect that the docstring will
appear both on the interface as well as on the implementation except for when the
implementation overwrites the docstring with a more specific descriptions.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -96,11 +96,18 @@ class ReceiptAPI(rlp.Serializable, ABC):
 class BaseTransactionAPI(ABC):
     @abstractmethod
     def validate(self) -> None:
+        """
+        Hook called during instantiation to ensure that all transaction
+        parameters pass validation rules.
+        """
         ...
 
     @property
     @abstractmethod
     def intrinsic_gas(self) -> int:
+        """
+        Convenience property for the return value of `get_intrinsic_gas`
+        """
         ...
 
     @abstractmethod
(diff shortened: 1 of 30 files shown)


eth/abc.py
```diff
@@ -1944,18 +2586,18 @@ def __init__(self, base_db: AtomicDatabaseAPI, header: BlockHeaderAPI = None) ->
     def apply_transaction(self,
                           transaction: SignedTransactionAPI
                           ) -> Tuple[BlockAPI, ReceiptAPI, ComputationAPI]:
-        ...
+        """
+        Apply the transaction to the current tip block.
 
-    @abstractmethod
-    def import_block(self,
```
cburgdorf: Removed because this is already inherited from `ChainAPI`

eth/abc.py
```diff
@@ -1944,18 +2586,18 @@ def __init__(self, base_db: AtomicDatabaseAPI, header: BlockHeaderAPI = None) ->
...
+        WARNING: ReceiptAPI and Transaction trie generation is computationally
+        heavy and incurs significant performance overhead.
+        """
         ...
 
     @abstractmethod
     def mine_block(self, *args: Any, **kwargs: Any) -> BlockAPI:
-        ...
-
-    def get_vm(self, at_header: BlockHeaderAPI = None) -> VirtualMachineAPI:
```
cburgdorf: Removed because this is already inherited from `ChainAPI`

cburgdorf:
@pipermerriam For the most part, this just moves docstrings and does some slight corrections or minor improvements. It doesn't add new docstrings to make this easier to review. I'm going to send a follow up PR that goes through `eth.abc` to add and improve docstrings.

PR #1882, opened 2019-11-25
Add missing docstrings

cburgdorf:
### What was wrong?

We have a bunch of types in our docs that are lacking docstrings. I believe having at least *some* docstring per public API is the lowest bar of requirement before we can cut a beta.

### How was it fixed?

Added at least **some** docstring for every single public API.

Imperative present tense is being used for the docstrings.

For methods, it should rhyme with **IT WILL:**
Example:

```python
def get_block:
    """
    Return the block at the tip of the chain
    """
```

For methods, it should rhyme with **IT IS:**
Example:

```python
class SchemaAPI(ABC):
    """
    A class representing a database schema that maps values to lookup keys.
    """
```

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -827,7 +827,10 @@ class StackAPI(ABC):
 
 
 class CodeStreamAPI(ABC):
-    pc: int
+    """
+    A class representing a stream of EVM code.
+    """
+    program_counter: int
 
     @abstractmethod
     def read(self, size: int) -> bytes:
@@ -850,7 +853,11 @@ class CodeStreamAPI(ABC):
         ...
 
     @abstractmethod
-    def seek(self, pc: int) -> ContextManager['CodeStreamAPI']:
+    def seek(self, program_counter: int) -> ContextManager['CodeStreamAPI']:
+        """
+        Return a :class:`~typing.ContextManager` with the program counter
+        set to ``program_counter``.
+        """
         ...
 
     @abstractmethod
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -50,6 +50,10 @@ T = TypeVar('T')
 
 
 class MiningHeaderAPI(rlp.Serializable, ABC):
+    """
+    A class to define a block header without ``mix_hash`` and ``nonce`` which can act as a
+    temporary representation during mining before the block header is sealed.
+    """
     parent_hash: Hash32
     uncles_hash: Hash32
     coinbase: Address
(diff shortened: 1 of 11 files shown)


eth/tools/_utils/slow_code_stream.py
```diff
@@ -65,27 +65,27 @@ def __next__(self) -> int:
             return STOP
 
     def peek(self) -> int:
-        current_pc = self.pc
+        current_pc = self.program_counter
```
cburgdorf: I have this in a separate commit but thought I could sneak it in here. I personally had to lookup what `pc` stands for again and I remember I already looked it up in the past. Part of this is probably just my poor memory but I also think it aligns better with the rest of the codebase to not use two letter acronyms :)

eth/validation.py
```diff
@@ -174,17 +174,15 @@ def validate_stack_int(value: int) -> None:
...
     )
 
 
 def validate_stack_bytes(value: bytes) -> None:
     if len(value) <= 32:
         return
     raise ValidationError(
-        "Invalid Stack Item: Must be either a length 32 byte "
-        f"string or a 256 bit integer. Got {value!r}"
+        "Invalid Stack Item: Must be either a length 32 byte string. Got {value!r}"
```
cburgdorf: Another non-doc change that I sneaked in here. I think the text was wrong.

eth/vm/code_stream.py
```diff
@@ -46,31 +46,31 @@ def __getitem__(self, i: int) -> int:
 
     def __iter__(self) -> Iterator[int]:
         # a very performance-sensitive method
-        pc = self.pc
+        pc = self.program_counter
```
pipermerriam: I would extract this to it's own PR.  :+1:

eth/abc.py
```diff
@@ -473,7 +587,7 @@ def add_transaction(self,
     def get_block_transactions(
             self,
             block_header: BlockHeaderAPI,
-            transaction_class: Type[SignedTransactionAPI]) -> Sequence[SignedTransactionAPI]:
+            transaction_class: Type[SignedTransactionAPI]) -> Tuple[SignedTransactionAPI, ...]:
```
cburgdorf: Another non-doc change sneaked in here. It seems more aligned with our best practices to have the return type be the actual type that is returned (in contrast to parameter types which should accept the broadest type that the method could work with). This also aligns the method with the other methods on that class that do also use `Tuple[Something, ...]` as return type.

eth/vm/code_stream.py
cburgdorf: Ooops..too late. I though the approval and rage merged

PR #1885, opened 2019-11-28
[WIP] Add Python 3.8 testing

### LINKED ISSUE #1872: Add py3.8 tests
### What is wrong?

No tests validating 3.8 support.

### How can it be fixed

Add new tox environment and circle CI tests.
Also, be sure to add pip constraint blake2b-py v0.1.3:
https://github.com/ethereum/vyper/pull/1678#issuecomment-552017732

veox:
### What was wrong?

Nothing; Python 3.8 has been released.

### How was it fixed?

See #1872 for verbose noise.

- [x] Now installing `libleveldb-dev` on _all_ CI jobs, so we can get it on Python 3.8 jobs, so it can build a `plyvel` wheel locally (otherwise not yet available).
- [x] Updated `coincurve` dep to latest: that has a Python 3.8 wheel: fixes installation on Arch Linux.
- [x] Updated `mypy` due to nested dependency (`typed-ast`) local wheel build fail.
  - [ ] Fixed what newer `mypy` found to be broken (`lint` jobs across all Pythons).
- [x] Updated `hypothesis` to latest (not much consideration given).
  - [ ] Fixed introduced failure (`core` jobs across all Pythons).

Closes #1872 as implemented.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

Source: [Cmycherrytree, via imgur](https://imgur.com/gallery/tAjzq)


tests/core/code-stream/test_code_stream.py
```diff
@@ -203,7 +203,10 @@ def test_new_vs_reference_code_stream_iter(bytecode):
     assert latest.program_counter == reference.program_counter
 
 
-@given(read_len=st.integers(min_value=0), bytecode=st.binary(max_size=128))
+@given(
+    read_len=st.integers(min_value=0, max_value=2048),
```
veox: This (and below) certainly requires review. `max_value=pow(2,63)-1` works just fine, too. It's the `pow(2,63)` that gets passed implicitly is what breaks the test.

Possibly an off-by-one issue upstream.

pipermerriam: This seems fine to me.

veox:
Remaining lint failures are actual inconsistencies in the codebase.

voith:
You might also want to bump the minimum supported `eth-utils` version to `v1.8.0` as Python3.8 support was added in that version.

eth/abc.py
```diff
@@ -775,7 +775,7 @@ def __call__(self, computation: 'ComputationAPI') -> None:
     def as_opcode(cls: Type[T],
                   logic_fn: Callable[['ComputationAPI'], None],
                   mnemonic: str,
-                  gas_cost: int) -> Type[T]:
+                  gas_cost: int) -> T:
```
veox: Made same as:

https://github.com/ethereum/py-evm/blob/3f33ac8c2b7a190b187f971dc95005341ccd00ca/eth/vm/opcode.py#L43

eth/vm/state.py
```diff
@@ -51,7 +51,7 @@ def __init__(
             self,
             db: AtomicDatabaseAPI,
             execution_context: ExecutionContextAPI,
-            state_root: bytes) -> None:
+            state_root: Hash32) -> None:
```
veox: @carver This is the example of cargo-culting I mentioned on gitter. `tox` said (before the change):

```
eth/vm/state.py:57: error: Argument 2 to "AccountDatabaseAPI" has incompatible type "bytes"; expected "Hash32"
```

... so I gave it exactly what it asked. But I don't know if it's right or not.

carver: Yeah, in this case, it looks like the right thing to do to go with `Hash32`, since it's logically the hash of the root node. :+1:

eth/abc.py
carver: Yeah, I agree with your call here, too. The implementation is clearly returning an `Opcode` instance instead of class, and would require a *lot* of refactoring to change (if we even wanted to).

The only thing I'd add it that the ABC docs should update, too.

eth/abc.py
```diff
@@ -775,7 +775,7 @@ def __call__(self, computation: 'ComputationAPI') -> None:
...
         """
         Class factory method for turning vanilla functions into Opcode classes.
```
carver: `s/Opcode classes/Opcodes/'

veox:
Remaining errors in `py36-lint` seem related:

```
eth/vm/state.py:164: error: Incompatible return value type (got "Tuple[Hash32, JournalDBCheckpoint]", expected "Tuple[Hash32, UUID]")
eth/vm/state.py:172: error: Argument 1 to "discard" of "AccountDatabaseAPI" has incompatible type "UUID"; expected "JournalDBCheckpoint"
eth/vm/state.py:176: error: Argument 1 to "commit" of "AccountDatabaseAPI" has incompatible type "UUID"; expected "JournalDBCheckpoint"
```

This has me scratching my head.

eth/vm/state.py
veox: Should `has_root()` of `AccountDB` (and ABC `AccountDatabaseAPI`) have a `state_root` argument that's `Hash32` also?..

https://github.com/ethereum/py-evm/blob/0d79bb6c56f4835ef9242cbad18a51ef86e77011/eth/db/account.py#L135-L136

veox:
More importantly to `subj`, [py38-lint](https://circleci.com/gh/veox/py-evm/4964?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) fails with:

```
  File "/home/circleci/repo/.tox/py38-lint/lib/python3.8/site-packages/pyflakes/checker.py", line 696, in getNodeHandler
    self._nodeHandlers[node_class] = handler = getattr(self, nodeType)
AttributeError: 'FlakesChecker' object has no attribute 'CONSTANT'
```

This comes from `pyflakes`; the issue has [already been reported upstream](https://github.com/PyCQA/pyflakes/issues/367) and [fixed](https://github.com/PyCQA/pyflakes/pull/369). It's [not in a tagged release](https://pypi.org/project/pyflakes/#history), however - therefore marking PR blocked.

veox:
While waiting for `pyflakes`, I decided to review how much these changes impact `trinity`. Things are not looking good.

There are nested dependencies that have similar pinnings as `py-evm` has currently. `libp2p` (with [its `coincurve` pinned](https://github.com/libp2p/py-libp2p/blob/v0.1.3/setup.py#L59) in the latest tagged release) jumps out immediately, as that'll produce a `pip` installation error.

Downstream projects with `py-evm` as a dependency will get stuck on old versions of `py-evm` if this gets merged, until a time when other dependencies get _their_ dependencies updated...

"The hell you make is the one you deserve."

veox:
Just noticed that a new `pyflakes` has been released - will try to revisit next week.

PR #1886, opened 2019-11-29
Support more receipt fields

cburgdorf:
### What was wrong?

According to the specs and what geth does, this is what a response to `eth_getTransactionReceipt` should look like:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "blockHash": "0xf64a12502afc36db3d29931a2148e5d6ddaa883a2a3c968ca2fb293fa9258c68",
    "blockNumber": "0x70839",
    "contractAddress": null,
    "cumulativeGasUsed": "0x75d5",
    "from": "0xc80fb22930b303b55df9b89901889126400add38",
    "gasUsed": "0x75d5",
    "logs": [
      {
        "address": "0x03fca6077d38dd99d0ce14ba32078bd2cda72d74",
        "topics": [
          "0x24bcf19562365f6510754002f8d7b818d275886315d29c7aa04785570b97a363"
        ],
        "data": "0x0000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000a4861636b65726e65777300000000000000000000000000000000000000000000",
        "blockNumber": "0x70839",
        "transactionHash": "0x1e2910a262b1008d0616a0beb24c1a491d78771baa54a33e66065e03b1f46bc1",
        "transactionIndex": "0x0",
        "blockHash": "0xf64a12502afc36db3d29931a2148e5d6ddaa883a2a3c968ca2fb293fa9258c68",
        "logIndex": "0x0",
        "removed": false
      }
    ],
    "logsBloom": "0x00000000000000000000000000000400000000020000000000000000400000000000000000004000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000020000800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "root": "0xc659845f1ac4e899ff1b0666dbac5deeda33a4a5d85da71f617f352824146e40",
    "to": "0x03fca6077d38dd99d0ce14ba32078bd2cda72d74",
    "transactionHash": "0x1e2910a262b1008d0616a0beb24c1a491d78771baa54a33e66065e03b1f46bc1",
    "transactionIndex": "0x0"
  }
}
```

But the the receipt that I can fetch via `Chain.get_transaction_receipt` looks rather sparse.

```python
class ReceiptAPI(rlp.Serializable, ABC):
    """
    A class to define a receipt to capture the outcome of a transaction.
    """
    state_root: bytes
    gas_used: int
    bloom: int
    logs: Sequence[LogAPI]

    @property
    @abstractmethod
    def bloom_filter(self) -> BloomFilter:
        ...
```

I guess the fields on `ReceiptAPI` are all the bare fields that need to be persisted because we can not retrieve them from anywhere else. Which makes a lot of sense. We don't want to bloat the database with duplicate information. However, it's clear that I can't get all the information I need from that.

### How was it fixed?

I added a new `FullReceiptAPI` which is a superset of `ReceiptAPI`. We then fill in the gaps (mostly) at `Chain.get_transaction_receipt` which was changed to return a `FullReceiptAPI`.

### Open questions

1. Is this the general approach we wanna take? Any good alternatives?
2. Does it make sense to keep the old `Chain.get_transaction_receipt` which returns a `ReceiptAPI` and just add this other API as a new one? 
3. Do we want to give a proper `__init__ to not set all fields individually and if so, where would that live? (It is not an RLP but having actual implementations in `eth.abc` doesn't feel right either.
4. Or maybe, does the whole thing actually belong into Trinity? But that would mean we have to reach into `Chain.chaindb` from there which also doesn't feel *exactly* right.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -107,6 +107,24 @@ class ReceiptAPI(rlp.Serializable, ABC):
         ...
 
 
+class FullReceiptAPI:
+    """
+    A class representing a receipt with additional context information that is retrieved
+    via the transaction and block it belongs to.
+    """
+
+    block_hash: Hash32
+    block_number: BlockNumber
+    bloom: int
+    gas_used: int
+    logs: Sequence[LogAPI]
+    receipient: Address
+    sender: Address
+    state_root: bytes
+    transaction_hash: Hash32
+    transaction_index: int
+
+
 class BaseTransactionAPI(ABC):
     """
     A class to define all common methods of a transaction.
@@ -602,12 +620,22 @@ class ChainDatabaseAPI(HeaderDatabaseAPI):
         ...
 
     @abstractmethod
-    def get_receipt_by_index(self,
-                             block_number: BlockNumber,
-                             receipt_index: int) -> ReceiptAPI:
+    def get_receipt_by_number_and_index(self,
+                                        block_number: BlockNumber,
+                                        receipt_index: int) -> ReceiptAPI:
         """
         Return the receipt of the transaction at specified index
         for the block header obtained by the specified block number
+        """
+        ...
+
+    @abstractmethod
+    def get_receipt_by_header_and_index(self,
+                                        block_header: BlockHeaderAPI,
+                                        receipt_index: int) -> ReceiptAPI:
+        """
+        Return the receipt of the transaction at specified index
+        for the given block header.
         """
         ...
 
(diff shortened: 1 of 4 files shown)


cburgdorf:
Yeah, I was unsure if the concept of a full receipt is something that people would expect directly on the evm level. I don't have a strong preference either way so I happily move it over.

I do think we still need to add an API on the Py-EVM level to make that possible though. Something like: 

```
def get_header_by_transaction_hash(self, transaction_hash: Hash32) -> BlockHeaderAPI:
```

Without that, I don't seem to have a way in Trinity to collect all the information I need without reaching into `chain.chaindb` which seems like a no go.

cburgdorf:
@pipermerriam a new variant of this can now be found at #1887

PR #1888, opened 2019-12-05
Create scripts/gethimport.py

lithp:
### What was wrong?

Nothing was *wrong*, per say, but it wasn't previously possible to get a full-syncing trinity node this far:

Selection_048

### How was it fixed?

I wrote a script with a few helpful subcommands which can be used to create a trinity database, bootstrapping off an already-synced geth node.

### To-Do

- [ ] Remove the code which iterates over the state trie? It's too slow to be practical except for very early blocks.
- [ ] Support simulating a full sync, starting from genesis, by also importing the genesis state when creating a trinity database. This depends on moving the EIP1085 code over from trinity -> py-evm, so it probably doesn't belong in this PR.
- [ ] Figure out how to test any of this without committing large example geth databases to the repo.
- [ ] Clean up commit history
- [ ] Find a cute animal pic
- [ ] Add a command to compact the database. If the database is not compacted it will compact itself, 
freezing your process for an uncomfortably long amount of time.
- [ ] Document how to use this script to build a trinity database to reproduce the above screenshot

#### Cute Animal Picture


scripts/gethimport.py
```diff
@@ -0,0 +1,619 @@
...
+    for key, value in iterator:
+        if len(key) != 32:
+            skipped_keys += 1
+            continue
+        trinitydb[key] = value
+        imported_entries += 1
+
+        if key >= bucket:
+            logger.debug(f'imported: {bucket.hex()} skipped={skipped_keys}')
+            if bucket == b'\xff' * 2:
```
lithp: This is a bug that I fixed on my server but not locally, I'll push a new commit soon!

eth/db/chain.py
```diff
@@ -93,15 +94,18 @@ def get_block_uncles(self, uncles_hash: Hash32) -> Tuple[BlockHeaderAPI, ...]:
     def _set_as_canonical_chain_head(
             cls,
             db: DatabaseAPI,
-            block_hash: Hash32,
+            header: BlockHeaderAPI,
```
carver: Splitting the PR apart a bit would really help. Like this change looks interesting: I can give it a better review in a self-contained PR.

lithp: Yeah, makes sense!

lithp: https://github.com/ethereum/py-evm/pull/1891

PR #1890, opened 2019-12-06
Bugfix: don't resume apply_all_transactions if EVM is missing data

### LINKED ISSUE #1889: Bug: apply_all_transactions tries to resume if data is missing
### What is wrong?

It should really just give up and re-raise the EVMMissingData.

AFAICT, this is just a not-often-used path.

Brian found it while attacking state sync with the turbogeth-style database.

### How can it be fixed

re-raise

carver:
### What was wrong?

Fix #1889 

### How was it fixed?

If apply_all_transactions is missing data, reraise

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/newsfragments/1889.bugfix.rst
+++ b/newsfragments/1889.bugfix.rst
@@ -1,0 +1,3 @@
+Bug: if data was missing during a call to :meth:`~eth.vm.base.VM.apply_all_transactions`,
+then the call would revert and continue processing transactions. Fix: we re-raise
+the :class:`~eth.exceptions.EVMMissingData` and do not continue processing transactions.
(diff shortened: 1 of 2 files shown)


newsfragments/1889.bugfix.rst
```diff
@@ -0,0 +1,3 @@
+If data was missing during a call to :meth:`~eth.vm.base.VM.apply_all_transactions`,
+then it was reverting that one transaction and continuing. Instead, we re-raise the
+:class:`~eth.exceptions.EVMMissingData`.
```
lithp: "it" is a little ambiguous here. Another option: 

```
If data was missing during a call to :meth:`~eth.vm.base.VM.apply_all_transactions`,
then the call would revert and continue processing transactions. Instead, we re-raise 
the :class:`~eth.exceptions.EVMMissingData` and do not continue processing transactions.
```

PR #1891, opened 2019-12-06
Improve header chain extension performance

lithp:
### What was wrong? / How was it fixed?

Breaking this out from #1888, improves performance if the parent of the header being imported is the canonical chain tip. `_find_new_ancestors` is relatively expensive; if I remember right a lot of time is spent fetching and de-serializing the header which was just imported, when all that's wanted is its hash.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

panda-tree

### DIFF
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -90,21 +90,11 @@ class ChainDB(HeaderDB, ChainDatabaseAPI):
             return tuple(rlp.decode(encoded_uncles, sedes=rlp.sedes.CountableList(BlockHeader)))
 
     @classmethod
-    def _set_as_canonical_chain_head(
-            cls,
-            db: DatabaseAPI,
-            block_hash: Hash32,
-            genesis_parent_hash: Hash32,
-    ) -> Tuple[Tuple[BlockHeaderAPI, ...], Tuple[BlockHeaderAPI, ...]]:
-        try:
-            header = cls._get_block_header_by_hash(db, block_hash)
-        except HeaderNotFound:
-            raise ValueError(
-                f"Cannot use unknown block hash as canonical head: {header.hash}"
-            )
-
-        new_canonical_headers = tuple(reversed(
-            cls._find_new_ancestors(db, header, genesis_parent_hash)))
+    def _decanonicalize_old_headers(
+        cls,
+        db: DatabaseAPI,
+        new_canonical_headers: Tuple[BlockHeaderAPI, ...]
+    ) -> Tuple[BlockHeaderAPI, ...]:
         old_canonical_headers = []
 
         # remove transaction lookups for blocks that are no longer canonical
@@ -118,19 +108,16 @@ class ChainDB(HeaderDB, ChainDatabaseAPI):
                 old_header = cls._get_block_header_by_hash(db, old_hash)
                 old_canonical_headers.append(old_header)
                 try:
-                    for transaction_hash in cls._get_block_transaction_hashes(db, old_header):
+                    transaction_hashes = cls._get_block_transaction_hashes(db, old_header)
+                    for transaction_hash in transaction_hashes:
                         cls._remove_transaction_from_canonical_chain(db, transaction_hash)
                 except MissingTrieNode:
-                    # If the transactions were never stored for the (now) non-canonical chain,
-                    #   then you don't need to remove them from the canonical chain lookup.
+                    # If the transactions were never stored for the (now) non-canonical
+                    # chain, then you don't need to remove them from the canonical chain
+                    # lookup.
                     pass
 
-        for h in new_canonical_headers:
-            cls._add_block_number_to_hash_lookup(db, h)
-
-        db.set(SchemaV1.make_canonical_head_hash_lookup_key(), header.hash)
-
-        return new_canonical_headers, tuple(old_canonical_headers)
+        return tuple(old_canonical_headers)
 
     #
     # Block API
--- a/eth/db/header.py
+++ b/eth/db/header.py
@@ -1,5 +1,5 @@
 import functools
-from typing import Iterable, Tuple
+from typing import cast, Iterable, Tuple
 
 import rlp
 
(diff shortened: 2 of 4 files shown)


eth/db/chain.py
```diff
@@ -93,15 +94,18 @@ def get_block_uncles(self, uncles_hash: Hash32) -> Tuple[BlockHeaderAPI, ...]:
...
-            raise ValueError(
-                f"Cannot use unknown block hash as canonical head: {header.hash}"
-            )
+            current_canonical_head = cls._get_canonical_head_hash(db)
+        except CanonicalHeadNotFound:
+            current_canonical_head = None
+        if current_canonical_head and header.parent_hash == current_canonical_head:
+            cls._add_block_number_to_hash_lookup(db, header)
+            db.set(SchemaV1.make_canonical_head_hash_lookup_key(), header.hash)
+            return (header,), tuple()
```
carver: Cool, doing this optimization makes sense, but the logic duplication (adding block number lookup, and setting the canonical head hash) concerns me. If something updates further down in the method, it would be *really* easy to forget to do it here.

One alternative:

- In the if block, only set `new_canonical_headers, old_canonical_headers = (header,), ()`
- Move the following 3 lines of `{old,new}_canonical_headers` generation into the else block

It does one extra db lookup at the `_get_canonical_block_hash` to check for an old header.

If you think that extra lookup is critical, and worth some more PR thrash, then you could move the `for h in new_canonical_headers` block into a `cls._decanonicalize_old_headers(db, new_headers)` call, inside the else block.

eth/db/header.py
```diff
@@ -83,11 +83,15 @@ def get_canonical_head(self) -> BlockHeaderAPI:
...
     @classmethod
     def _get_canonical_head(cls, db: DatabaseAPI) -> BlockHeaderAPI:
+        canonical_head_hash = cls._get_canonical_head_hash(db)
+        return cls._get_block_header_by_hash(db, Hash32(canonical_head_hash))
+
+    @classmethod
+    def _get_canonical_head_hash(cls, db: DatabaseAPI) -> Hash32:
         try:
-            canonical_head_hash = db[SchemaV1.make_canonical_head_hash_lookup_key()]
+            return cast(Hash32, db[SchemaV1.make_canonical_head_hash_lookup_key()])
```
carver: Christoph was recommending doing this instead. I guess type-checking is supposed to notice if you send a non-bytes value:

```suggestion
            return Hash32(db[SchemaV1.make_canonical_head_hash_lookup_key()])
```

YMMV, do what you like :)

eth/db/header.py
```diff
@@ -83,11 +83,15 @@ def get_canonical_head(self) -> BlockHeaderAPI:
 
     @classmethod
     def _get_canonical_head(cls, db: DatabaseAPI) -> BlockHeaderAPI:
+        canonical_head_hash = cls._get_canonical_head_hash(db)
+        return cls._get_block_header_by_hash(db, Hash32(canonical_head_hash))
```
carver: This `Hash32` wrapper looks redundant.

eth/db/header.py
```diff
@@ -173,7 +177,7 @@ def _persist_checkpoint_header(
         )
         previous_score = score - header.difficulty
         cls._set_hash_scores_to_db(db, header, previous_score)
-        cls._set_as_canonical_chain_head(db, header.hash, header.parent_hash)
+        cls._set_as_canonical_chain_head(db, header, header.parent_hash)
```
carver: Yay that this change has so straightforward, and didn't require a new header lookup anywhere!

eth/db/header.py
```diff
@@ -250,12 +254,15 @@ def _set_as_canonical_chain_head(
...
-        except HeaderNotFound:
-            raise ValueError(
-                f"Cannot use unknown block hash as canonical head: {block_hash}"
-            )
+            current_canonical_head = cls._get_canonical_head_hash(db)
+        except CanonicalHeadNotFound:
+            current_canonical_head = None
+        if current_canonical_head and header.parent_hash == current_canonical_head:
+            cls._add_block_number_to_hash_lookup(db, header)
+            db.set(SchemaV1.make_canonical_head_hash_lookup_key(), header.hash)
```
carver: Same concerns here, about the logic duplication from later in the method.

eth/db/chain.py
lithp: Yes, great point!

lithp: Surprisingly enough the extra lookup added ~9% to import time. I pushed a change which keeps it out but also removes some duplicated code.

lithp:
Just addressed the rest of the outstanding PR feedback

lithp:
The lint failure is... interesting. My first guess is that mypy just upgraded, and circleci doesn't pin the version?

cburgdorf:
@lithp interesting indeed! I looked into it.

When installed with `.[dev]` we end up with `eth-utils` version `1.7.0` which has some decorators that do not play nice with types.

But when we isnstall with `.[eth, lint]` like the lint job does we end up with `eth-utils` version `1.8.4` which does play nicely with types (thanks @njgheorghita :+1: ) and hence the info comes up that the `ignore` isn't needed anymore.

So, my guess is something in `dev` forces us to a lower version of `eth-utils`.

carver:
> When installed with `.[dev]` we end up with `eth-utils` version `1.7.0` which has some decorators that do not play nice with types.
> 
> But when we install with `.[eth, lint]` like the lint job does we end up with `eth-utils` version `1.8.4` which does play nicely with types (thanks @njgheorghita  ) and hence the info comes up that the `ignore` isn't needed anymore.

Ahah! I was wondering why I could only reproduce linting issues with local tox runs, on certain issues.

eth/db/chain.py
```diff
@@ -118,19 +108,16 @@ def _set_as_canonical_chain_head(
                 old_header = cls._get_block_header_by_hash(db, old_hash)
                 old_canonical_headers.append(old_header)
                 try:
-                    for transaction_hash in cls._get_block_transaction_hashes(db, old_header):
+                    transaction_hashes = cls._get_block_transaction_hashes(db, old_header)
+                    for transaction_hash in transaction_hashes:
```
carver: This split was just to do fewer things on one line, right? (I'm fine with it, just wondering if there was something subtle I was missing)

lithp: Nope, no subtle reason, the change was to get us under the 90-character line width limit

lithp:
Woah, thanks for looking into it @cburgdorf, that helped a lot!

PR #1897, opened 2019-12-10
Add generate_contract_address to transaction methods API

cburgdorf:
### What was wrong?

1. In theory the way a contract address is derived for a *Contract Creation Transaction*  could change between forks. Our architecture does currently not reflect that.

2. We would like to reuse the `generate_contract_address` logic to compute contract addresses when we serve receipts in Trinity. (See https://github.com/ethereum/trinity/pull/1380)

### How was it fixed?

1. Added `generate_contract_addess` to the `BaseTransactionAPI`

2. Call `transaction.generate_contract_address` instead of the hardcoded utility method in `FrontierTransactionExecutor`.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -111,6 +111,15 @@ class BaseTransactionAPI(ABC):
     """
     A class to define all common methods of a transaction.
     """
+
+    @staticmethod
+    @abstractmethod
+    def generate_contract_address(sender: Address, nonce: int) -> Address:
+        """
+        Generate an address for a contract that is created by a *Contract Creation Transaction*,
+        that is, a transaction where the ``to`` field is empty.
+        """
+
     @abstractmethod
     def validate(self) -> None:
         """
--- a/eth/rlp/transactions.py
+++ b/eth/rlp/transactions.py
@@ -22,6 +22,10 @@ from eth.abc import (
     UnsignedTransactionAPI,
 )
 
+from eth._utils.address import (
+    generate_contract_address,
+)
+
 from .sedes import address
 
 
@@ -35,6 +39,10 @@ class BaseTransactionMethods(BaseTransactionAPI):
 
     def gas_used_by(self, computation: ComputationAPI) -> int:
         return self.get_intrinsic_gas() + computation.get_gas_used()
+
+    @staticmethod
+    def generate_contract_address(sender: Address, nonce: int) -> Address:
+        return generate_contract_address(sender, nonce)
 
 
 BASE_TRANSACTION_FIELDS = [
--- a/eth/vm/forks/frontier/state.py
+++ b/eth/vm/forks/frontier/state.py
@@ -21,10 +21,6 @@ from eth.exceptions import (
     ContractCreationCollision,
 )
 
-from eth._utils.address import (
-    generate_contract_address,
-)
-
 from eth.vm.message import (
     Message,
 )
@@ -63,7 +59,7 @@ class FrontierTransactionExecutor(BaseTransactionExecutor):
         message_gas = transaction.gas - transaction.intrinsic_gas
 
         if transaction.to == CREATE_CONTRACT_ADDRESS:
-            contract_address = generate_contract_address(
+            contract_address = transaction.generate_contract_address(
                 transaction.sender,
                 self.vm_state.get_nonce(transaction.sender) - 1,
             )


cburgdorf:
@pipermerriam there's some previous discussion to this: https://github.com/ethereum/trinity/pull/1337/files#r355609102

`eth-utils` does currently not depend on `py-rlp` which I suppose it would have to if we want to have this method in `eth-utils`.

Also, even if we had it in `eth-utils` wouldn't it still make sense to do this refactoring so that the implementation could be changed between forks if needed. It looks to me that this flexibility is desirable.

pipermerriam:
> Also, even if we had it in eth-utils wouldn't it still make sense to do this refactoring so that the implementation could be changed between forks if needed

[YAGNI](Also, even if we had it in eth-utils wouldn't it still make sense to do this refactoring so that the implementation could be changed between forks if needed)?

As for the `eth-utils` part...   I think we can just manually do  the RLP encoding logic... rather than using `pyrlp`.

cburgdorf:
>I think we can just manually do the RLP encoding logic

Not sure how I feel about that. The whole point of this PR was to get rid of some copy and paste in Trinity while at the same time improving the flexibility. Not sure if we gain much by duplicating some of the pyrlp code instead. Let's get @carver as tie breaker since he came up with the (slightly different) proposal.

pipermerriam:
The duplicated pyrlp code would be very small.  Here's the *raw* code from pyrlp that would do this.  Since we know the shape of the data we can easily slim this down to only the needed functionality.

```python
def encode_raw(item):
    """RLP encode (a nested sequence of) :class:`Atomic`s."""
    if isinstance(item, Atomic):
        if len(item) == 1 and item[0] < 128:
            return item
        payload = item
        prefix_offset = 128  # string
    elif not isinstance(item, str) and isinstance(item, collections.abc.Sequence):
        payload = b''.join(encode_raw(x) for x in item)
        prefix_offset = 192  # list
    else:
        msg = 'Cannot encode object of type {0}'.format(type(item).__name__)
        raise EncodingError(msg, item)

    try:
        prefix = length_prefix(len(payload), prefix_offset)
    except ValueError:
        raise EncodingError('Item too big to encode', item)

    return prefix + payload


LONG_LENGTH = 256**8


def length_prefix(length, offset):
    """Construct the prefix to lists or strings denoting their length.

    :param length: the length of the item in bytes
    :param offset: ``0x80`` when encoding raw bytes, ``0xc0`` when encoding a
                   list
    """
    if length < 56:
        return ALL_BYTES[offset + length]
    elif length < LONG_LENGTH:
        length_string = int_to_big_endian(length)
        return ALL_BYTES[offset + 56 - 1 + len(length_string)] + length_string
    else:
        raise ValueError('Length greater than 256**8')
```

cburgdorf:
Yes, but the duplicated code we try to get rid of is even less.

https://github.com/ethereum/trinity/blob/c8630d719469aa073cccdb51dc6ecd921ed1f459/trinity/_utils/address.py

Alternatively, we could also expose the function as a non utility/private API from py-evm.

Duplicating rlp logic just doesn't feel right to me here tbh.

pipermerriam:
My only :-1: was attaching this to the `Transaction` APIs.  I'm more comfy with it just being exposed as a function and we can attach it somewhere in the VM-specific APIs if we end up needing that.

carver:
> My only  was attaching this to the `Transaction` APIs. I'm more comfy with it just being exposed as a function and we can attach it somewhere in the VM-specific APIs if we end up needing that.

So are you :+1: to `VM.generate_contract_address(sender, nonce)`? I'm not sure where else to expose it that makes sense.

If we can't pick an API relatively quickly, then I think duplicating `generate_contract_address` in the Trinity repo is better than adding some custom rlp logic to eth-utils.

pipermerriam:
I'm going to just back out of this completely and let you guys decide.  :smiling_imp:

carver:
Oh, I see, from the other thread:

> I implemented this with a tiny twist so that it can be used in the transaction executor where the VM is not accessible (ethereum/py-evm#1897).

Well I don't like any solution we have, then. Maybe just add an issue to think about a refactor, and expose it as a non-VM-tied utility? Or just copy it in trinity.

cburgdorf:
The downside to `VM.generate_contract_address(sender, nonce)` is that it won't be usable from within the `TransactionExecutor` which feels wrong to me if say `BerlinVM` would overwrite `generate_contract_address` but then the transaction executor would still not use it.

cburgdorf:
>Well I don't like any solution we have, then

Alright, just going with copy and paste then for now.

PR #1898, opened 2019-12-11
Only delete (account) storage values if they are present

carver:
### What was wrong?

Fixes: https://github.com/ethereum/trinity/issues/1376

### How was it fixed?

Was trying to delete a storage value that was missing. Other changes made this problem more apparent.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/db/storage.py
+++ b/eth/db/storage.py
@@ -210,7 +210,15 @@ class AccountStorageDB(AccountStorageDatabaseAPI):
         if value:
             self._journal_storage[key] = rlp.encode(value)
         else:
-            del self._journal_storage[key]
+            try:
+                current_val = self._journal_storage[key]
+            except KeyError:
+                # deleting an empty key has no effect
+                return
+            else:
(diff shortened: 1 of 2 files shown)


carver:
> This is quite rubber-stampy as I'm not fully up on the context of this. I assume the _deletion_ here has to do with `value == 0` which is effectively like deleting the value in a storage slot.

Yup! A new `BatchDB` added in the previous PR means that we sometimes raise a `KeyError` now when getting or deleting a value. (The `HexaryTrie` has the weird behavior of never raising a `KeyError` when asking for a missing key, so we were ignoring the scenario). So we now handle `KeyError` cases, in part by checking the value before deleting, and in part by handling a `KeyError` raised when retrieving the previous value.

It looks a bit redundant to hande it in both ways, but that's because we have to deal with the quirk of not wanting to delete the value if the `HexaryTrie` is already empty there (with a `b''` value).

PR #1899, opened 2019-12-12
Refactor consensus engine handling as well as clique internals

cburgdorf:
### What was wrong?

This is another attempt to refactor Clique based on #1875 and what @carver and I discussed offline.
There's also a Trinity PR that can sync Görli with this: https://github.com/ethereum/trinity/pull/1196

### How was it fixed?

1. `validate_seal` and `validate_header` are now instance methods. The only reason they can be classmethods today is because our Pow implementation relies on a globally shared cache which should be refactored as described in #1900 

2. There are two new methods: `chain.validate_chain_extension(header, parents)` and `vm.validate_seal_extension`. They are to perform extension seal checks to support consensus schemes where headers can not be checked if parents are missing.

3. The consensus mechanism is now abstracted via `ConsensusAPI` and `ConsensusContextAPI`. VMs instantiate a consensus api based on the set `consensus_class`and pass it a context which they receive from the chain upon instantiation. The chain instantiates the consensus context api based on the `consensus_context_class`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth/consensus/clique/snapshot_manager.py
```diff
@@ -59,27 +58,28 @@ class SnapshotManager:
     """
 
     logger = get_extended_debug_logger('eth.consensus.clique.snapshot_manager.SnapshotManager')
+    # TODO: How do we get around this? It is way too slow without cached snapshots.
+    _snapshots = lru.LRU(IN_MEMORY_SNAPSHOTS)
```
cburgdorf: @carver I've hit another road block here with the stateless `ConsensusAPI` concept. I'm sorry I didn't notice this earlier. Here it goes:

So, we were able to get rid of the `HeaderCache` because that could be replaced by going from `validate_seal(header)` to `validate_seal_extension(header, parents)`. I should also note that other clients don't necessarily have a header cache in their clique implementation and that this was just me trying to make things work with just `validate_seal` that only takes a single header.

But the snapshot cache is something different. Every Clique implementation out there has this (afaik) and I believe it is at the very heart of clique so I'm not sure how we would get around it.

It basically means: If we don't cache snapshots then with every single header that we validate, we will have to load all previous header from the database until we hit the most recent checkpoint header (every 30000 header for Görli). As you can imagine, this is super super slow.

As a band-aid fix I moved this to the class level but that obviously isn't an acceptable solution. I can't help myself but think we have to move at least parts this back to the chain level unless you have other clever ideas that I might not be seeing.

I want to note that if you feel like playing with this (you can run it with https://github.com/ethereum/trinity/pull/1196 as simple as `trinity --goerli --disable-tx-pool`) then I'm happy if you take this in any direction that gets us closer to a merge. I'm not trying to offload this to you, just saying that if you have ideas then you should not feel obligated to discuss them first if you think you could just take a stab on it and finish it up. I basically just want to see this wrapped up and out of my way :sweat_smile:

carver: Ah, okay. Silly question: what if we persisted every snapshot?

An alternative that comes to mind is having the chain pass in a volatile data store. Much like how it passes in the DB, but with the expectation that everything in the volatile store will be dropped on shutdown. (Also, that the VMs need to manage cross-fork incompatibility across the volatile store themselves, like they need to do with the DB already)

cburgdorf: >what if we persisted every snapshot?

Multiple problems I think:

1. Lot's of overhead in snapshot creation
2. Still one db lookup per block to get the snapshot
3. Lots of wasted disk space that we need to clean up again

>Much like how it passes in the DB, but with the expectation that everything in the volatile store will be dropped on shutdown.

Yes, dropping the data on shutdown won't be a problem I think even dropping them between forks wouldn't be a problem at all. It would just be one header after the fork that causes us to suck in 30000 headers from the db until we are back to normal.

So, yeah, that would work but I think it can not be just some predefined storage because e.g. in this case we want to setup an LRU cache for the snapshots and in other cases we might needs some different kind of storage or eviction polity.

But actually, this brought me to some idea and I took a stab at it and it seems to work. Here's a rough outline and then I continue with some inline comments.

Instead of the `ConsensusAPI` taking in the db, it will take in a `ConsensusContextAPI`.

```python
class ConsensusAPI(ABC):

    @abstractmethod
    def __init__(self, context: ConsensusContextAPI) -> None:
        """
        Initialize the consensus api.
        """
        ...
```

We set the `consensus_context_class` on the chain. And just like the `ChainDB` the `consensus_context` is set *once* when the chain is initialized.

```python
class Chain(BaseChain):
    ...
    chaindb_class: Type[ChainDatabaseAPI] = ChainDB
    consensus_context_class: Type[ConsensusContextAPI] = ConsensusContext

    def __init__(self, base_db: AtomicDatabaseAPI) -> None:
       ...
        self.chaindb = self.get_chaindb_class()(base_db)
        self.consensus_context = self.consensus_context_class(self.chaindb.db)
        self.headerdb = HeaderDB(base_db)
```

Then later, in `get_vm(at_header)` that context is being used.

```python
    def get_vm(self, at_header: BlockHeaderAPI = None) -> VirtualMachineAPI:
        header = self.ensure_header(at_header)
        chain_context = ChainContext(self.chain_id)

        return vm_class(
            header=header,
            chaindb=self.chaindb,
            chain_context=chain_context,
            consensus_context=self.consensus_context
        )
```

So, the bottom line is:

- we continue to have VM instances that do only operate on the state that they get passed in
- we have consensus context state that is per-chain rather than per-vm
- we do not have any state leaking (as in: two different chain instances accidentially sharing consensus state).

There's a lot to clean up here but you can take a look at [the most recent commit](https://github.com/ethereum/py-evm/pull/1899/commits/6e8484ba6475433280119167b884c76de8b594fd) which implements this and also another one [on the Trinity side](https://github.com/ethereum/trinity/pull/1196) that takes it on a ride. It works and I think it kinds gives us the best of both worlds.

I have a good feeling with that, hoping you feel the same vibe :sweat_smile:
If you are :+1: on that direction than I can clean up the mess and propose a clean PR.

carver: Yeah, this feels like a good direction :+1: :tada: 

I also like the idea that maybe we can enforce the reset of the context across fork boundaries. For example, we might create an independent consensus_context for each distinct VM, so in `get_vm()` instead of just getting `self.consensus_context` we get `self.get_consensus_context(vm_class)` which is generated on demand, but only has one per VM.

PoW could use this too! There is an epoch cache that it doesn't want to regenerate every time, and I think it does something hacky to save them (like keep a global cache :see_no_evil: ). Obviously that change doesn't need to go in here, but a tracking issue would be cool.

cburgdorf: >PoW could use this too! There is an epoch cache that it doesn't want to regenerate every time, and I think it does something hacky to save them

Ah yes! We can use that to clean this up :)

>I also like the idea that maybe we can enforce the reset of the context across fork boundaries

Yeah, we could do that if we need it but I think we should have a good reason to do that. Like in, the Clique case it would just cause us to take more time to validate the header after each fork transition while not providing any actual benefit. But sure there could be other consensus schemes that would need a different context per fork.

Ok, I'm gonna start getting this into shape now :)

cburgdorf: Tracking issue: https://github.com/ethereum/py-evm/issues/1900

cburgdorf:
@carver This can be given a new review now. For a second I thought we could actually roll back to having `validate_seal` and `validate_header` become `classmethod`s again now that we have `validate_seal_extension` but that's actually not the case because, as you noted, the only reason they work as classmethods today is because they rely on some hacky global cache.

eth/abc.py
```diff
@@ -2269,6 +2270,59 @@ def get_transaction_context(cls,
...
+class ConsensusContextAPI(ABC):
+    """
+    A class representing a data context for the :class:`~eth.abc.ConsensusAPI` that remains static
+    over the life time of a chain instance.
+    """
+
+    @abstractmethod
+    def __init__(self, db: AtomicDatabaseAPI) -> None:
+        """
+        Initialize the context with a database.
```
carver: Couple random thoughts, as I won't be able to pick up the review again until later:

I was assuming we would formalize the idea that this context can't write to the database (Maybe this should be a read-only version that's supplied?) Or if we don't then maybe it fits as a kind of parallel to `ChainDB`, like `ConsensusDB`.

"remains static" in the docs above didn't immediately mean to me what I think we want to say. Something like: this instance stays in memory across VM runs.

cburgdorf: >I was assuming we would formalize the idea that this context can't write to the database 

But, we need to write to the database. E.g. Clique persists snapshots to the db.

https://github.com/ethereum/py-evm/blob/7b94eb17d43d08ac7732a0d6f935190423b5e33e/eth/consensus/clique/snapshot_manager.py#L248

>Or if we don't then maybe it fits as a kind of parallel to `ChainDB`, like `ConsensusDB`

I'm not sure how that would work out. In the `ChainDB` case, we know the specific types that make a chain and that can be stored and retrieved. But for consensus, it seems pretty much up to the specific consensus algorithm which kind of things need persistence. Clique uses snapshots but it's hard to imagine all the different things other consensus schemes would require.

>"remains static" in the docs above didn't immediately mean to me what I think we want to say. Something like: this instance stays in memory across VM runs.

:+1: I'll update that.

cburgdorf:
@carver just a reminder that this still needs to be reviewed.

eth/consensus/pow.py
```diff
@@ -92,3 +99,33 @@ def mine_pow_nonce(block_number: int, mining_hash: Hash32, difficulty: int) -> T
...
+    """
+
+    def __init__(self, base_db: AtomicDatabaseAPI) -> None:
+        pass
+
+    def validate_seal(self, header: BlockHeaderAPI) -> None:
+        """
+        Validate the seal on the given header by checking the proof of work.
+        """
+        check_pow(
```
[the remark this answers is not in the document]
cburgdorf: Yes, we have that here https://github.com/ethereum/py-evm/issues/1900

PR #1901, opened 2019-12-16
Implement Muir Glacier fork

cburgdorf:
### What was wrong?

Implement Muir Glacier fork

### How was it fixed?

- copy + rename from Istanbul
- calculate difficulty based on block number `9_000_000`

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/vm/forks/muir_glacier/__init__.py
+++ b/eth/vm/forks/muir_glacier/__init__.py
@@ -1,0 +1,31 @@
+from typing import (
+    Type,
+)
+
+from eth.rlp.blocks import BaseBlock
+from eth.vm.forks.istanbul import (
+    IstanbulVM,
+)
+from eth.vm.state import BaseState
+
+from .blocks import MuirGlacierBlock
+from .headers import (
+    compute_muir_glacier_difficulty,
+    configure_muir_glacier_header,
+    create_muir_glacier_header_from_parent,
+)
+from .state import MuirGlacierState
+
+
+class MuirGlacierVM(IstanbulVM):
+    # fork name
+    fork = 'muir-glacier'
+
+    # classes
+    block_class: Type[BaseBlock] = MuirGlacierBlock
+    _state_class: Type[BaseState] = MuirGlacierState
+
+    # Methods
+    create_header_from_parent = staticmethod(create_muir_glacier_header_from_parent)  # type: ignore
+    compute_difficulty = staticmethod(compute_muir_glacier_difficulty)    # type: ignore
+    configure_header = configure_muir_glacier_header
--- a/eth/vm/forks/muir_glacier/opcodes.py
+++ b/eth/vm/forks/muir_glacier/opcodes.py
@@ -1,0 +1,7 @@
+import copy
+from eth.vm.forks.istanbul.opcodes import (
+    ISTANBUL_OPCODES,
+)
+
+
+MUIR_GLACIER_OPCODES = copy.deepcopy(ISTANBUL_OPCODES)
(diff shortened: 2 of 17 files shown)


cburgdorf:
@carver There don't seem to be any tests for Muir Glacier jetzt in ethereum/tests and I'm also not sure if there will be tests since this is just the difficulty adjustment. Is there anything left to do here?

carver:
> @carver There don't seem to be any tests for Muir Glacier jetzt in ethereum/tests and I'm also not sure if there will be tests since this is just the difficulty adjustment. Is there anything left to do here?

Yeah, could also add the mainnet and ropsten block numbers, and set the `latest_mainnet_at` to Muir Glacier. We could probably benefit from a fork checklist in the docs, too, to remember all the stuff to do. :)

cburgdorf:
>Yeah, could also add the mainnet and ropsten block numbers, and set the latest_mainnet_at to Muir Glacier

Good catch. Added.

>We could probably benefit from a fork checklist in the docs, too, to remember all the stuff to do. :)

Yeah, I'll write something up while the memory is still fresh :)

eth/vm/forks/muir_glacier/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+)
+from .state import MuirGlacierState
+
+
+class MuirGlacierVM(IstanbulVM):
+    # fork name
+    fork = 'muir-glacier'
+
+    # classes
+    block_class: Type[BaseBlock] = MuirGlacierBlock
```
pipermerriam: We could start using `ClassVar` for these if you want to be fancy.  I'm not sure it gains us much since I don't think we would be likely to write any code that would overwrite them but maybe still worth having it enforced by the type checker....

eth/vm/forks/muir_glacier/opcodes.py
```diff
@@ -0,0 +1,6 @@
+from eth.vm.forks.istanbul.opcodes import (
+    ISTANBUL_OPCODES,
+)
+
+
+MUIR_GLACIER_OPCODES = ISTANBUL_OPCODES
```
pipermerriam: Maybe make a copy to be sure mutation to one doesn't mutate the other?

tests/core/builder-tools/test_chain_construction.py
```diff
@@ -82,7 +83,7 @@ def test_chain_builder_construct_chain_vm_configuration_multiple_forks():
         (constantinople_at, ConstantinopleVM),
         (petersburg_at, PetersburgVM),
         (istanbul_at, IstanbulVM),
-        (latest_mainnet_at, PetersburgVM),  # this will change whenever the next upgrade is locked
+        (latest_mainnet_at, MuirGlacierVM),  # this will change whenever the next upgrade is locked
```
pipermerriam: should this list also include a `(muir_glacier_at, MuirGlacierVM)`?

cburgdorf: It should :+1:

eth/vm/forks/muir_glacier/__init__.py
cburgdorf: That's probably a good idea but we would have to apply that all the way up to the very top of the inheritance chain because otherwise you run into:

>Cannot override instance variable (previously declared on base class "IstanbulVM") with class variable

Maybe we keep that out of this PR and apply it as a separate refactoring.

eth/vm/forks/muir_glacier/opcodes.py
cburgdorf: Good point :+1:

PR #1907, opened 2020-01-09
Improve logging message for failed seal check

cburgdorf:
### What was wrong?

There's a logging message printed when the seal check fails that mentions PoW even though that code path could also be running under a different consensus mechanism.

### How was it fixed?

Changed the logging message to be more generic (failed seal check

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -580,10 +580,11 @@ class VM(Configurable, VirtualMachineAPI):
             if check_seal:
                 try:
                     self.validate_seal(header)
-                except ValidationError:
+                except ValidationError as exc:
                     self.cls_logger.warning(
-                        "Failed to validate header proof of work on header: %r",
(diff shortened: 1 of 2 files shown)


eth/vm/base.py
```diff
@@ -582,7 +582,7 @@ def validate_header(self,
                     self.validate_seal(header)
                 except ValidationError:
                     self.cls_logger.warning(
-                        "Failed to validate header proof of work on header: %r",
+                        "Failed to validate seal on header: %r",
```
gsalgado: Would it make sense to include the exception msg here as well as it might include more info about why exactly it failed?

cburgdorf: Oh yeah good idea, I'll add that :+1:

PR #1908, opened 2020-01-09
Extract DAO check into static function

cburgdorf:
### What was wrong?

Now that `validate_header` has become an instance method it turned problematic to have the DAO check coupled with it. That is because in Trinity we need this check very early in the p2p process.

https://github.com/ethereum/trinity/blob/1a112b681a19f1321ac74588e4225aaf595d5dd5/trinity/protocol/common/boot.py#L84-L87

But obtaining an instance of the VM is also problematic since we can not provide all required dependencies.

### How was it fixed?

Extracted the DAO check into a static method which `validate_header` just delegates to. During the boot we can just invoke the static method. This does also mean that we are not running the actual seal check at that point anymore but I think that is ok since it will run at a later stage anyway.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/chains/mainnet/__init__.py
+++ b/eth/chains/mainnet/__init__.py
@@ -44,6 +44,27 @@ from eth.vm.forks import (
 )
 
 
+def validate_header_is_on_intended_dao_fork(support_dao_fork: bool,
+                                            dao_fork_at: BlockNumber,
+                                            header: BlockHeaderAPI) -> None:
+
+    # The special extra_data is set on the ten headers starting at the fork
+    extra_data_block_nums = range(dao_fork_at, dao_fork_at + 10)
+
+    if header.block_number in extra_data_block_nums:
+        if support_dao_fork and header.extra_data != DAO_FORK_MAINNET_EXTRA_DATA:
+            raise ValidationError(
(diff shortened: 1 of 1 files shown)


pipermerriam:
FYI, the entire concept that drove the need for the "BootManager" concept went away a while back when we got the networking code out of the peer.  If you wanted to refactor to remove all of those constructs and just embed this stuff in the `ETHPeer` and `LESPeer` that would be fine.

carver:
> FYI, the entire concept that drove the need for the "BootManager" concept went away a while back when we got the networking code out of the peer. If you wanted to refactor to remove all of those constructs and just embed this stuff in the `ETHPeer` and `LESPeer` that would be fine.

Agree, but maybe/hopefully we can switch to fork-id checks soon instead. Maybe @cburgdorf can take a quick pass at extracting the DAO check from BootManager to the Peer. But if we look at it and see that it starts to unravel, maybe we can save the work for when we switch to the fork-id check.

cburgdorf:
I actually don't *need* this change anymore if we go with #1909 but I'm merging it anyway because I think the separation of the function is still a tiny net win for the code base.

PR #1909, opened 2020-01-10
Drop check_seal support from validate_header

cburgdorf:
### What was wrong?

`VM.validate_header` recently became an instance method (was a `classmethod` before). The only reason why it became an instance method is because it has an optional seal check and seal checks can only be performed on VM instances.

### How was it fixed?

@carver [suggested](https://github.com/ethereum/trinity/pull/1196#discussion_r364957121) to take the seal check out of this method. I initially was a bit skeptical about that as it comes with the risk of simply forgetting about the check. But in practice, there are few places where we explicitly need to call this and higher level code usually deals with APIs such as `validate_chain` that cover the seal check internally.
 
So yeah, this PR drops `check_seal` from `validate_header`, makes it a `classmethod` and changes all call sites accordingly.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -548,16 +548,17 @@ class VM(Configurable, VirtualMachineAPI):
                 f" - header uncle_hash: {block.header.uncles_hash}"
             )
 
-    def validate_header(self,
+    @classmethod
+    def validate_header(cls,
                         header: BlockHeaderAPI,
-                        parent_header: BlockHeaderAPI,
-                        check_seal: bool = True) -> None:
+                        parent_header: BlockHeaderAPI) -> None:
+
         if parent_header is None:
             # to validate genesis header, check if it equals canonical header at block number 0
             raise ValidationError("Must have access to parent header to validate current header")
         else:
             validate_length_lte(
-                header.extra_data, self.extra_data_max_bytes, title="BlockHeader.extra_data")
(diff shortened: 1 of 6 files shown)


eth/vm/base.py
```diff
@@ -577,19 +578,16 @@ def validate_header(self,
...
-                        exc,
-                    )
-                    raise
-
     def validate_seal(self, header: BlockHeaderAPI) -> None:
-        self._consensus.validate_seal(header)
+        try:
+            self._consensus.validate_seal(header)
+        except ValidationError as exc:
+            self.cls_logger.warning(
```
cburgdorf: I moved this here because I think the standard logging on a failed seal is a net good for debugging etc.

pipermerriam: So one potentially negative effect of this is that if someone sent us an invalid header in trinity and it hits this method then trinity will output this warning message to the console.  Is that what we want?

carver: > Is that what we want?

Hm, not sure. Geth does print out warnings about bad *blocks*, and it causes some concern/confusion.

cburgdorf: What if this default log entry would only go into `debug` or even `debug2` instead?

pipermerriam: `debug` seems appropriate.

PR #1913, opened 2020-01-29
Fix consensus bug from deleting account in same block that creates it

### LINKED ISSUE #1912: Consensus failure on contract create/destruct within same block (both FrontierVM and PetersburgVM)
* py-evm Version: ever since [`v0.2.0-alpha.43`](https://github.com/ethereum/py-evm/releases/tag/v0.2.0-alpha.43)
* OS: Arch Linux
* Python Version (python --version): Python 3.7
* Environment (output of `pip freeze`): MISSING

### What is wrong?

First raised in https://github.com/ethereum/trinity/issues/1478: failure to perform full-sync, tested on main-net/Goerli (pretty sure this is chain-independent). There is a state root mismatch.

It [seems](https://github.com/ethereum/trinity/issues/1478#issuecomment-579715506) that if a contract is created and deleted within the same block, then we're erroneously re-creating its storage as "empty".

Probably has something to do with how accounts and their stores are marked "dirty" when they're created/destroyed.

See https://github.com/ethereum/trinity/issues/1478 for much more debug details, traces, etc..

### How can it be fixed

There is quite some indirection there - journals, caches...

[Bisect showed](https://github.com/ethereum/trinity/issues/1478#issuecomment-575335557) the culprit to be https://github.com/ethereum/py-evm/pull/1735.

[Debug-tracing showed](https://gist.github.com/veox/bbc4586e5d59f44c3cdc089ee801481a#file-the-diff) the following (diff between good/bad `py-evm` versions in linked gist):

```
>   DEBUG2  01-29 13:33:37      AccountStorageDB  Deleting all storage in account 0xb214cbce30676bd117f95a8695d34427295cd6f7, hashed 0x88593f406750a4f40bdf82a1efbacbec6a5297e62830ee1a8942f083de97ad14
> ...
>   DEBUG2  01-29 13:33:37             AccountDB  Updating account 0xb214cbce30676bd117f95a8695d34427295cd6f7 to storage root 0x856157ca7b18901a14648a8d426b568dcba060e3997d5a409c25fe09d77351b6
> ...
>   DEBUG2  01-29 13:33:37             AccountDB  Updating account 0xb214cbce30676bd117f95a8695d34427295cd6f7 to storage root 0x856157ca7b18901a14648a8d426b568dcba060e3997d5a409c25fe09d77351b6
> ...
```

* First, seems we're going through the "dirty" list twice.
* Then, the "deleting all storage in account" message can be slightly misleading: the "hashed" value `0x8859...` printed is a hash of the address, `0xb214c...` - not sure this is intended; if anything, I'd expect the account's storage root to be here instead, or at least also present.
* More importantly, _this_ is the snippet that seems to show the account's storage is deleted, and then re-created.

carver:
### What was wrong?

Fix #1912 

### How was it fixed?

The bug was: when a previous transaction creates/modifies an account,
and then is "locked", those pending changes are written at the end of the block, even if the
account is deleted is a subsequent transaction.

The fix was: clear out all "locked" storage changes (those made during previous transactions) when you lock in an account deletion.

That fix brought up a new interesting case, where the "dirty" flag wasn't noticing when an account started the block as empty, then some storage was added in one transaction, then in a later transaction it was deleted again. The further solution was to track a history of of storage tries that can be reverted to (a new one added on every account delete), and generously treat the storage as dirty if there were ever any writes in the block, even if the trie currently has the same state root as the root was at the beginning of the block.

### To-Do

- [x] Clean up commit history
- [x] more documentation

#### Cute Animal Picture

### DIFF
--- a/eth/db/storage.py
+++ b/eth/db/storage.py
@@ -1,12 +1,20 @@
+from typing import (
+    List,
+    NamedTuple,
+)
+
 from eth_hash.auto import keccak
 from eth_typing import (
     Address,
-    Hash32
+    Hash32,
 )
 from eth_utils import (
     ValidationError,
+    encode_hex,
     get_extended_debug_logger,
     int_to_big_endian,
+    to_bytes,
+    to_int,
 )
 import rlp
 from trie import (
@@ -22,8 +30,14 @@ from eth.abc import (
     AtomicDatabaseAPI,
     DatabaseAPI,
 )
+from eth.constants import (
+    BLANK_ROOT_HASH,
+)
 from eth.db.backends.base import (
     BaseDB,
+)
+from eth.db.backends.memory import (
+    MemoryDB,
 )
 from eth.db.batch import (
     BatchDB,
@@ -42,6 +56,17 @@ from eth.typing import (
 )
 
 
+class PendingWrites(NamedTuple):
+    """
+    A set of variables captured just before account storage deletion.
+    The variables are used to revive storage if the EVM reverts to a point
+    prior to deletion.
+    """
+    write_trie: HexaryTrie  # The write trie at the time of deletion
+    trie_nodes_batch: BatchDB  # A batch of all trie nodes written to the trie
+    starting_root_hash: Hash32  # The starting root hash
+
+
 class StorageLookup(BaseDB):
     """
     This lookup converts lookups of storage slot integers into the appropriate trie lookup.
@@ -51,12 +76,23 @@ class StorageLookup(BaseDB):
     """
     logger = get_extended_debug_logger("eth.db.storage.StorageLookup")
 
+    # The trie that is modified in-place, used to calculate storage root on-demand
+    _write_trie: HexaryTrie
+
+    # These are the new trie nodes, waiting to be committed to disk
+    _trie_nodes_batch: BatchDB
+
+    # When deleting an account, push the pending write info onto this stack.
+    # This stack can get as big as the number of transactions per block: one for each delete.
+    _historical_write_tries: List[PendingWrites]
+
     def __init__(self, db: DatabaseAPI, storage_root: Hash32, address: Address) -> None:
         self._db = db
-        self._starting_root_hash = storage_root
+
+        # Set the starting root hash, to be used for on-disk storage read lookups
+        self._initialize_to_root_hash(storage_root)
+
         self._address = address
-        self._write_trie = None
-        self._trie_nodes_batch: BatchDB = None
 
     def _get_write_trie(self) -> HexaryTrie:
         if self._trie_nodes_batch is None:
@@ -120,7 +156,7 @@ class StorageLookup(BaseDB):
 
     @property
     def has_changed_root(self) -> bool:
-        return self._write_trie and self._write_trie.root_hash != self._starting_root_hash
+        return self._write_trie is not None
 
     def get_changed_root(self) -> Hash32:
         if self._write_trie is not None:
@@ -128,10 +164,13 @@ class StorageLookup(BaseDB):
         else:
             raise ValidationError("Asked for changed root when no writes have been made")
 
-    def _clear_changed_root(self) -> None:
+    def _initialize_to_root_hash(self, root_hash: Hash32) -> None:
+        self._starting_root_hash = root_hash
         self._write_trie = None
         self._trie_nodes_batch = None
-        self._starting_root_hash = None
+
+        # Reset the historical writes, which can't be reverted after committing
+        self._historical_write_tries = []
 
     def commit_to(self, db: DatabaseAPI) -> None:
         """
@@ -142,10 +181,67 @@ class StorageLookup(BaseDB):
         if self._trie_nodes_batch is None:
             raise ValidationError(
                 "It is invalid to commit an account's storage if it has no pending changes. "
-                "Always check storage_lookup.has_changed_root before attempting to commit."
+                "Always check storage_lookup.has_changed_root before attempting to commit. "
+                f"Write tries on stack = {len(self._historical_write_tries)}; Root hash = "
+                f"{encode_hex(self._starting_root_hash)}"
             )
         self._trie_nodes_batch.commit_to(db, apply_deletes=False)
-        self._clear_changed_root()
+
+        # Mark the trie as having been all written out to the database.
+        # It removes the 'dirty' flag and clears out any pending writes.
+        self._initialize_to_root_hash(self._write_trie.root_hash)
+
+    def new_trie(self) -> int:
+        """
+        Switch to an empty trie. Save the old trie, and pending writes, in
+        case of a revert.
+
+        :return: index for reviving the previous trie
+        """
+        write_trie = self._get_write_trie()
+
+        # Write the previous trie into a historical stack
+        self._historical_write_tries.append(PendingWrites(
+            write_trie,
+            self._trie_nodes_batch,
+            self._starting_root_hash,
+        ))
+
+        new_idx = len(self._historical_write_tries)
+        self._starting_root_hash = BLANK_ROOT_HASH
+        self._write_trie = None
+        self._trie_nodes_batch = None
+
+        return new_idx
+
+    def rollback_trie(self, trie_index: int) -> None:
+        """
+        Revert back to the previous trie, using the index returned by a
+        :meth:`~new_trie` call. The index returned by that call returns you
+        to the trie in place *before* the call.
+
+        :param trie_index: index for reviving the previous trie
+        """
+
+        if trie_index >= len(self._historical_write_tries):
+            raise ValidationError(
+                f"Trying to roll back a delete to index {trie_index}, but there are only"
+                f" {len(self._historical_write_tries)} indices available."
+            )
+
+        (
+            self._write_trie,
+            self._trie_nodes_batch,
+            self._starting_root_hash,
+        ) = self._historical_write_tries[trie_index]
+
+        # Cannot roll forward after a rollback, so remove created/ignored tries.
+        # This also deletes the trie that you just reverted to. It will be re-added
+        # to the stack when the next new_trie() is called.
+        del self._historical_write_tries[trie_index:]
+
+
+CLEAR_COUNT_KEY_NAME = b'clear-count'
 
 
 class AccountStorageDB(AccountStorageDatabaseAPI):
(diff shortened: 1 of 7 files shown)


eth/db/storage.py
```diff
@@ -185,11 +199,22 @@ def __init__(self, db: AtomicDatabaseAPI, storage_root: Hash32, address: Address
         big_endian encoding of the slot integer, and the rlp-encoded value.
         """
         self._address = address
-        self._storage_lookup = StorageLookup(db, storage_root, address)
+        self._journal_trie_locator = JournalDB(MemoryDB({TRIE_LOOKUP_KEY_NAME: storage_root}))
+
+        # One lookup for each starting trie. So far, this is only used to manage
+        # two tries: the original trie and a deleted one.
+        self._storage_lookups = {}
+        self._storage_lookups[storage_root] = StorageLookup(db, storage_root, address)
```
carver: I was just fooling around, but of course there are a lot of reasons this approach won't work. For example: the starting storage root is empty when the account doesn't exist at the beginning of the block. That will clash with the new "Deleted" trie at the empty storage root. Also, the storage_cache -- ... --> journal_storage are tied to the original storage lookup here, no matter which one is active. I can keep tweaking this, but... it's pretty messy and making me consider a larger refactor. I haven't seen a design that I'm happy with yet.

carver:
Well, making progress: Block 116526 is now importing!

... but it only got to 116530 before failing again :man_facepalming: 

I'll wait to figure that out before posting this for review.

tests/core/vm/test_vm_state.py
```diff
@@ -103,6 +103,52 @@ def test_revert_selfdestruct(state, read_storage_before_snapshot):
...
+    state.set_storage(ADDRESS, 0, 1)
+    state.lock_changes()
+
+    # delete account in next "transaction"
+    state.delete_account(ADDRESS)
+    assert state.get_storage(ADDRESS, 0) == 0
+    state.lock_changes()
+    assert state.get_storage(ADDRESS, 0) == 0
+
+    # delete account in next "transaction"
```
veox: Minor: probably meant to say "revive" instead of "delete".

veox:
Block [`#116530`](https://etherscan.io/txs?block=116530) is very similar: calls are made to create and destroy contracts... Not sure what's different from the previous one, I'll take a look now.

veox:
> Well, making progress: Block 116526 is now importing!

It's `#116525` there was trouble with... Are you sure you didn't import _that_ using some other code by accident? Because I'm trying Trinity with this PR's branch, and am still failing to import `#116525`.

pipermerriam:
I *think* we may be near a point where we need to pay down tech debt in this area.  It seems we've ended up with lots of layers here and I at least am having a lot of difficulty reasoning about how we manage all of the nuances related to updating account and contract state.

carver:
> It's `#116525` there was trouble with... Are you sure you didn't import _that_ using some other code by accident (e.g. [f4cba85](https://github.com/ethereum/py-evm/commit/f4cba8543bf5e6c82dd81a7f03a81166932ce8b8))?

No, you're right. I had imported it using the old code and skipped over it during sync.

carver:
> I _think_ we may be near a point where we need to pay down tech debt in this area. It seems we've ended up with lots of layers here and I at least am having a lot of difficulty reasoning about how we manage all of the nuances related to updating account and contract state.

I'm also not happy with how complex the code is, and it being a pain to debug.

Some key things that seem to be interacting in a complex way are:
- queuing up all the state changes without writing to the trie
- state changes can be reverted back to the beginning of a transaction, but not across transaction boundaries
- when storage tries are deleted, it changes the storage root hash on the same timeline as the other storage changes. The root hash change must be revertable, in lock-step with all the other storage changes.
- thinking through pre and post Byzantium: whether you have to flush the changes to the trie at each transaction, or only at the end of the block

With all those interacting factors above, I have no immediate idea how to make it better. One option is to push the complexity into `JournalDB`, to make it trie-aware (to handle root hash changes) and boundary-aware (to handle changes that cannot be reverted, from previous transactions). I think that's just pushing complexity around, doesn't really seem like an improvement. I would probably need a solid chunk of time to really reset my mind on how to design it a whole new way.

carver:
But some good news! The latest push actually does import block 116525, and is at 153492 and continuing...

carver:
Wow, a shocking number of consensus bugs that are not covered by ethereum/tests.

Found the next failed import at block #336508.

eth/db/storage.py
```diff
@@ -222,40 +279,63 @@ def set(self, slot: int, value: int) -> None:
 
     def delete(self) -> None:
         self.logger.debug2(
-            "Deleting all storage in account 0x%s, hashed 0x%s",
+            "Deleting all storage in account 0x%s",
```
veox: `keccak(self._address).hex(),` arg is passed in but not used.

veox:
> Found the next failed import at block #336508.

Verified - indeed!

JIC, I've checked with `py-evm` back at `master`, too - the issue persists. So, it's an existing bug, not a regression in this PR.

veox:
I'm checking the intermediate `state_root`s for each transaction; in block [`#336508`](https://etherscan.io/txs?block=336508), there are 5 txs; at indexes 0, 1, 3, 4 are calls to `setAvatar(uint256,bytes)` (of a contract I didn't review); at index 2, there is the call to `deleteAllAvatars()`.

`state_root` at index 2 doesn't match:
* `geth` says it's `0xf1af37b9b975c2626d3b07ad995f06d5286af983d3b5f0ac0a0ff47bf0e1d098`;
* `trinity` says it's `0x39e4b7faa521431a15be1007fb19e9d386c057bde21917520f84312e2e2ac442`.

eth/db/storage.py
```diff
@@ -142,11 +163,46 @@ def commit_to(self, db: DatabaseAPI) -> None:
...
+            )
+
+        (
+            self._write_trie,
+            self._trie_nodes_batch,
+            self._starting_root_hash,
+        ) = self._historical_write_tries[trie_index]
+
+        # Cannot roll forward after a rollback, so remove created/ignored tries
+        del self._historical_write_tries[trie_index:]
```
veox: This also deletes the index that has just been rolled back to; it won't be possible to roll back to the same place again.

Is this intended?..

carver: Yup! It re-creates the index it again if there's a new account delete. I'll add a comment so it's clear that the delete is intentional.

carver:
> As to what's going on: the most likely reason we're not getting a `state_root` mismatch for the block: the deleted data in the contract's storage is overwritten (within the same block), and the resulting state is as expected.

Looks like it's even simpler than that: `AccountDB` isn't recognizing the storage as having changed, so it doesn't try to write out the new storage root to the account. (Note there is no `Updating account 0x5dc2e401e29ea16e51ebcecaef129c15fed5a5e8 to storage root X` line at the end of the transaction at index 2). *Why* it's not recognizing the storage as changed is currently a mystery, but I'm digging in now...

carver:
Looks like it had to do with reverting the storage root back to the root at the beginning of the block, and not treating that as a change. By always treating the storage as changed, that fixes pre-Byzantium (but seems to break something in Istanbul's `CREATE2`).

carver:
If the account is not created, and the "new" storage root is empty, then we shouldn't create the account just to put a blank storage root in it.

carver:
Up to block 415133 on my local full sync so far...

carver:
So maybe come back for a refactor after we get sync reliable and trinity to mvp

carver:
@cburgdorf Seems worth double-checking that this fix also fixes the goerli test you ran. How did you set that up again and/or are you up for re-running it against this branch?

cburgdorf:
I'm on vacation without laptop but just running `trinity --goerli
--sync-mode full` should be all you need. I don't know the exact block that
broke but if you get past 300k blocks it definitely fixed that bug. I'm
sure @veox can also assist.

veox:
Yup! I'll check this on main-net (from scratch), and then try to sync Goerli - that's where I started, after all.

veox:
Confirming that the issue is no longer present when using `py-evm` from PR https://github.com/ethereum/py-evm/pull/1913.

Main-net blocks `#116525` and `#336508` imported successfully (I ran it up to `#347457`).

So has Goerli block `#226323` (still running, `#472783` ATM).

eth/db/storage.py
```diff
@@ -120,7 +141,7 @@ def __delitem__(self, key: bytes) -> None:
 
     @property
     def has_changed_root(self) -> bool:
-        return self._write_trie and self._write_trie.root_hash != self._starting_root_hash
+        return self._write_trie
```
pipermerriam: Assuming I'm correct that this is checking for the `None` value, I'd advocate for this to be `return self._write_trie is not None`.

eth/db/storage.py
```diff
@@ -129,9 +150,9 @@ def get_changed_root(self) -> Hash32:
             raise ValidationError("Asked for changed root when no writes have been made")
 
     def _clear_changed_root(self) -> None:
+        self._starting_root_hash = self._write_trie.root_hash
```
pipermerriam: If you get back in here could you maybe drop a comment or two in this method explaining in some detail why we do each of these things.  Since it's not clear how to address some of the debt in the code, maybe we can address some of with with better comments.

eth/db/storage.py
```diff
@@ -53,11 +67,18 @@ class StorageLookup(BaseDB):
...
         self._starting_root_hash = storage_root
+
         self._address = address
-        self._write_trie = None
+        self._write_trie: HexaryTrie = None
         self._trie_nodes_batch: BatchDB = None
 
+        # When deleting an account, push the pending write info onto this stack.
+        # May happen multiple times for multiple deletes.
+        self._historical_write_tries: List[Tuple[HexaryTrie, BatchDB, Hash32]] = []
```
pipermerriam: Maybe make a `NamedTuple` for the inner value type here.  Probably makes some of the other code a bit easier to read/grok.

eth/db/storage.py
```diff
@@ -187,9 +262,14 @@ def __init__(self, db: AtomicDatabaseAPI, storage_root: Hash32, address: Address
...
         self._storage_lookup = StorageLookup(db, storage_root, address)
         self._storage_cache = CacheDB(self._storage_lookup)
-        self._locked_changes = BatchDB(self._storage_cache)
+        self._locked_changes = JournalDB(self._storage_cache)
         self._journal_storage = JournalDB(self._locked_changes)
 
+        # Track how many times we have cleared the storage. This is journaled
+        # so that we can revert to an old clear count. The clear count is used
+        # as an index to find the base trie from before the revert.
+        self._clear_count = JournalDB(MemoryDB({CLEAR_COUNT_KEY_NAME: to_bytes(0)}))
```
pipermerriam: Using the `JournalDB` here feels like it may be the wrong choice.  Can this just be modeled using a list (stack) or somethng?

carver: I updated the comment a bit to try to explain: it has to move in lockstep with the main storage journal, so we can detect when a revert crosses back over an account delete.

PR #1917, opened 2020-02-29
Generate Witness Index while executing a block

carver:
### What was wrong?

https://github.com/ethereum/trinity/issues/1225 requires a way to generate information about a witness.

### How was it fixed?

(Note that this is based on #1918 -- I will rebase on master when that gets merged)

Modified `Chain.import_block()` to include another result in `BlockImportResult`, called a `WitnessIndex`.

The `WitnessIndex` includes important information like:
- the hashes of the trie nodes which are accessed while executing a block
- the list of accounts that were accessed while executing a block
- the list of storage slots, for each account, that were accessed while executing a block
- the list of accounts for which we loaded contract bytecode

Note that a `WitnessIndex` is not **the** witness. It doesn't include required information like:
- the bodies of the trie nodes which are accessed while executing a block
- the rlp's accounts that were accessed while executing a block
- the values of storage slots, for each account, that were accessed while executing a block
- the bytecode for each account that had it's bytecode accessed

(note that the witness can be fully generated from the first item, or partially generated by the rest, but need some extra bits of data to prove inclusion in the state root).

The first practical use case will focus entirely on the trie node hashes and trie node bodies for witnesses.

### To-Do

- [x] move new key logger db to fresh PR: #
- [x] move minor stuff to ^ PR (makefile, block `__str__` upgrade)
- [x] switch `mine_block()` etc to use NamedTuple result
- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -8,6 +8,7 @@ from typing import (
     ClassVar,
     ContextManager,
     Dict,
+    FrozenSet,
     Iterable,
     Iterator,
     MutableMapping,
@@ -334,10 +335,53 @@ class BlockAPI(rlp.Serializable, ABC):
         ...
 
 
+class MetaWitnessAPI(ABC):
+    @property
+    @abstractmethod
+    def hashes(self) -> FrozenSet[Hash32]:
+        ...
+
+    @property
+    @abstractmethod
+    def accounts_queried(self) -> FrozenSet[Address]:
+        ...
+
+    @property
+    @abstractmethod
+    def account_bytecodes_queried(self) -> FrozenSet[Address]:
+        ...
+
+    @abstractmethod
+    def get_slots_queried(self, address: Address) -> FrozenSet[int]:
+        ...
+
+    @property
+    @abstractmethod
+    def total_slots_queried(self) -> int:
+        """
+        Summed across all accounts, how many storage slots were queried?
+        """
+        ...
+
+
+class BlockAndMetaWitness(NamedTuple):
+    """
+    After evaluating a block using the VirtualMachine, this information
+    becomes available.
+    """
+    block: BlockAPI
+    meta_witness: MetaWitnessAPI
+
+
 class BlockImportResult(NamedTuple):
+    """
+    After importing and persisting a block into the active chain, this information
+    becomes available.
+    """
     imported_block: BlockAPI
     new_canonical_blocks: Tuple[BlockAPI, ...]
     old_canonical_blocks: Tuple[BlockAPI, ...]
+    meta_witness: MetaWitnessAPI
 
 
 class SchemaAPI(ABC):
@@ -1631,6 +1675,13 @@ class AccountStorageDatabaseAPI(ABC):
         """
         ...
 
+    @abstractmethod
+    def get_accessed_slots(self) -> FrozenSet[int]:
+        """
+        List all the slots that had been accessed since object creation.
+        """
+        ...
+
 
 class AccountDatabaseAPI(ABC):
     """
@@ -1835,7 +1886,7 @@ class AccountDatabaseAPI(ABC):
         ...
 
     @abstractmethod
-    def persist(self) -> None:
+    def persist(self) -> MetaWitnessAPI:
         """
         Send changes to underlying database, including the trie state
         so that it will forever be possible to read the trie from this checkpoint.
@@ -2192,7 +2243,7 @@ class StateAPI(ConfigurableAPI):
         ...
 
     @abstractmethod
-    def persist(self) -> None:
+    def persist(self) -> MetaWitnessAPI:
         """
         Persist the current state to the database.
         """
@@ -2490,14 +2541,14 @@ class VirtualMachineAPI(ConfigurableAPI):
     # Mining
     #
     @abstractmethod
-    def import_block(self, block: BlockAPI) -> BlockAPI:
+    def import_block(self, block: BlockAPI) -> BlockAndMetaWitness:
         """
         Import the given block to the chain.
         """
         ...
 
     @abstractmethod
-    def mine_block(self, *args: Any, **kwargs: Any) -> BlockAPI:
+    def mine_block(self, *args: Any, **kwargs: Any) -> BlockAndMetaWitness:
         """
         Mine the current block. Proxies to self.pack_block method.
         """
@@ -2518,7 +2569,7 @@ class VirtualMachineAPI(ConfigurableAPI):
     # Finalization
     #
     @abstractmethod
-    def finalize_block(self, block: BlockAPI) -> BlockAPI:
+    def finalize_block(self, block: BlockAPI) -> BlockAndMetaWitness:
         """
         Perform any finalization steps like awarding the block mining reward,
         and persisting the final state root.
@@ -3310,5 +3361,13 @@ class MiningChainAPI(ChainAPI):
         """
         Mines the current block. Proxies to the current Virtual Machine.
         See VM. :meth:`~eth.vm.base.VM.mine_block`
+        """
+        ...
+
+    @abstractmethod
+    def mine_block_extended(self, *args: Any, **kwargs: Any) -> BlockAndMetaWitness:
+        """
+        Just like :meth:`~mine_block`, but includes extra returned info. Currently,
+        the only extra info returned is the :class:`MetaWitness`.
         """
         ...
--- a/eth/db/account.py
+++ b/eth/db/account.py
@@ -1,5 +1,6 @@
 from lru import LRU
 from typing import (
+    cast,
     Dict,
     Iterable,
     Set,
@@ -16,6 +17,7 @@ from eth_utils import (
     encode_hex,
     get_extended_debug_logger,
     to_checksum_address,
+    to_dict,
     to_tuple,
     ValidationError,
 )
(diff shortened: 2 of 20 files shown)


eth/db/account.py
```diff
@@ -435,6 +459,27 @@ def persist(self) -> None:
...
+
+    @to_dict
+    def _get_witness_metadata(self) -> Iterable[Tuple[Address, Tuple[bool, Tuple[int, ...]]]]:
+        for address in self._accessed_accounts:
+            did_access_bytecode = address in self._accessed_bytecodes
+            if address in self._account_stores:
+                accessed_storage_slots = self._account_stores[address].get_accessed_slots()
+            else:
+                accessed_storage_slots = tuple()
+            yield address, (did_access_bytecode, accessed_storage_slots)
```
pipermerriam: Thoughts on making this a `NamedTuple`?.  I think it would improve the typing, making the types have names and field names.

eth/db/witness.py
```diff
@@ -0,0 +1,61 @@
...
+
+    @property
+    def accounts_queried(self) -> Collection[Address]:
+        return self._accounts_metadata_queried.keys()
+
+    @property
+    def account_bytecodes_queried(self) -> Tuple[Address, ...]:
+        return tuple(
+            address
+            for address, (was_bytecode_queried, _slots) in self._accounts_metadata_queried.items()
```
pipermerriam: Named tuple approach would mean you wouldn't have to unpack the second value and you could just access the fields using properties.

eth/db/witness.py
```diff
@@ -0,0 +1,67 @@
...
+    WitnessIndexAPI,
+)
+
+
+class AccountQueryTracker(NamedTuple):
+    did_query_bytecode: bool
+    slots_queried: Tuple[int, ...]
+
+
+class WitnessIndex(WitnessIndexAPI):
```
lithp: Even once I knew what a `WitnessIndex` was this name kept throwing me off, it suggests that we're indexing into a collection of witnesses. Maybe one of these would work?
- I like what you were originally calling it, the `MetaWitness`
- `WitnessSketch`
- `WitnessStats`
- `WitnessKeys`
- `WitnessProtection` ;)

eth/abc.py
```diff
@@ -334,10 +335,53 @@ def is_genesis(self) -> bool:
...
+    @property
+    @abstractmethod
+    def total_slots_queried(self) -> int:
+        """
+        Summed across all accounts, how many storage slots were queried?
+        """
+        ...
+
+
+class BlockEvalResult(NamedTuple):
```
lithp: Something like `BlockAndWitnessSketch` would make me less likely to want to look up the definition, I only know what a `BlockEvalResult` is once I've looked it up.

eth/abc.py
```diff
@@ -2490,14 +2541,14 @@ def make_receipt(self,
     # Mining
     #
     @abstractmethod
-    def import_block(self, block: BlockAPI) -> BlockAPI:
+    def import_block(self, block: BlockAPI) -> BlockEvalResult:
```
lithp: This signature change (along with the `mine_block` sig changes) made this PR a lot larger. It seems like a lot of callers don't really care about the witness! To make things easier for those callers, an alternative might be for this method to stay the same and methods added instead: `import_block_returning_witness_index` and `mine_block_returning_witness_index`.

eth/db/witness.py
```diff
@@ -0,0 +1,67 @@
...
+    def total_slots_queried(self) -> int:
+        """
+        Summed across all accounts, how many storage slots were queried?
+        """
+        return sum(
+            len(query_tracker.slots_queried)
+            for query_tracker in self._accounts_metadata_queried.values()
+        )
+
+    # TODO do we need a `has_hash()` method for trie nodes? If so, maybe store internally as a set
```
lithp: I don't know what you've built on top of this PR so maybe this idea won't work, but if `hashes` returned a set instead of a tuple you wouldn't need to add a `has_hash` method, callers could use the very succinct `hash in index.hashes`.

eth/db/witness.py
```diff
@@ -0,0 +1,67 @@
...
+    Hash32,
+)
+
+from eth.abc import (
+    WitnessIndexAPI,
+)
+
+
+class AccountQueryTracker(NamedTuple):
+    did_query_bytecode: bool
```
lithp: I don't know how this is going to be used, but a more accurate name for this might be `did_query_nonempty_bytecode`, this is only marked true if we tried to read the bytecode and also got an answer out.

lithp:
I can't comment inline because neither of these places have been touched, but, `_log_pending_accounts` does some logging:

```
        for deleted_address in sorted(diff.deleted_keys()):
            cast_deleted_address = Address(deleted_address)
            self.logger.debug2(
                "Deleted Account %s, empty? %s, exists? %s",
                to_checksum_address(deleted_address),
                self.account_is_empty(cast_deleted_address),
                self.account_exists(cast_deleted_address),
            )
```

Both `account_is_empty` and `account_exists` mark the accounts as read, so after `_log_pending_accounts` runs they'll be included in the witness. This logging will only be called on accounts which had `delete_account` called on them. `delete_account`, however, seems to be intentionally not adding the account.

- If you aren't trying to add deleted accounts to the witness, here's a place which is going to make the witness too large.
- If you are trying to add deleted account to the witness, I think that should be made explicit, and put into the call to `delete_account`!

eth/db/account.py
```diff
@@ -435,6 +460,27 @@ def persist(self) -> None:
             self._batchdb.commit_to(write_batch, apply_deletes=False)
         self._root_hash_at_last_persist = new_root_hash
 
+        return witness_index
+
+    def _get_accessed_node_hashes(self) -> Set[Hash32]:
+        return cast(Set[Hash32], self._raw_store_db.keys_read)
+
+    @to_dict
+    def _get_witness_metadata(self) -> Iterable[Tuple[Address, AccountQueryTracker]]:
```
lithp: I don't know if it's possible to call `get_storage` without first reading something from the `get_account`, but if it is possible then we'll miss some reads here. Here, we only read out of `_account_stores` for all the addresses in `_accessed_accounts`, but I think nothing `get_storage` calls ever adds to `_accessed_accounts`.

eth/db/storage.py
```diff
@@ -413,6 +417,9 @@ def _validate_flushed(self) -> None:
                 f"StorageDB had a dirty journal when it needed to be clean: {journal_diff!r}"
             )
 
+    def get_accessed_slots(self) -> Tuple[int, ...]:
```
lithp: Why not have this return a set?

eth/db/account.py
```diff
@@ -422,9 +441,15 @@ def persist(self) -> None:
...
 
+        # generate witness (copy) before clearing the underlying data
+        witness_index = self._get_witness_index()
+
         # reset local storage trackers
         self._account_stores = {}
         self._dirty_accounts = set()
+        self._accessed_accounts = set()
+        self._accessed_bytecodes = set()
+        self._account_cache.clear()
```
lithp: Why was this call to `clear()` added?

eth/db/witness.py
carver: Okay, maybe back to MetaWitness :+1:

eth/abc.py
```diff
@@ -334,10 +335,53 @@ def is_genesis(self) -> bool:
...
 
+class WitnessIndexAPI(ABC):
+    @property
+    @abstractmethod
+    def hashes(self) -> Tuple[Hash32, ...]:
+        ...
+
+    @property
+    @abstractmethod
+    def accounts_queried(self) -> Collection[Address]:
```
pipermerriam: *nitpick* 

Why the loose return type?  Typically we are strict about return types (requiring this be something concrete like `Tuple/List`.  Maybe worth documenting why it's allowed to be loose if there is a good reason.

eth/db/account.py
carver: > I think nothing `get_storage` calls ever adds to `_accessed_accounts`.

Hm, if the address store is in cache, then it won't trigger, but I don't know how it could get in cache without calling `_get_address_store()` first. That one does add the address to the list of accessed accounts, via:
`get_storage()` -> `_get_address_store()` -> `_get_storage_root()` -> `_get_account()` -> `_get_encoded_account()` -> adds to `_accessed_accounts`

eth/db/storage.py
carver: Just a general aversion to returning mutable data structures.

eth/db/account.py
carver: If we don't clear, then subsequent accesses that look up an account may not get added to the access list. (We skip adding the account to the access list if it's in cache) I'll add a comment explaining.

carver:
> Just realized the term we may want to use in place of `WitnessIndex` is `AccessList` which I believe is a more standardized term for this concept.

Cool, I used the name "access list" in a couple new, appropriate places. I also changed from `WitnessIndex` -> `MetaWitness`. The reason I didn't use it everywhere is that: the list of hashes of trie nodes doesn't fit with my idea of an access list (I think of an access list as just the list of addresses/slots that are accessed), so I think I'll go forward with `MetaWitness` and we can iterate the name from there (esp if/when we end up dropping the trie node hashes from the witness info).

eth/abc.py
carver: Right now it returns a dictionary's key view. I think I just didn't know the right `typing` class to use for that one. But... it's probably a bad idea to blindly return the key view anyway, since you can get a `RuntimeError` while iterating. So I'll just copy it to a tuple or something and make this type more specific.

PR #1918, opened 2020-03-03
Add Database for tracking key accesses

carver:
### What was wrong?

Wanted an easy way to track all key accesses, for #1917 .

### How was it fixed?

Added new `KeyAccessLoggerDB` (and an atomic database version -- the type checking seems to require both).

Some other bonus tweaks to:
- clean up types
- add some of the block hash to `str(block)`
- speed up the Makefile, etc.

### To-Do

- [ ] Clean up commit history
- [x] Add a test that key accesses are properly recorded

#### Cute Animal Picture

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,6 @@ coverage:
 	open htmlcov/index.html
 
 build-docs:
-	pip install -e .[doc]
 	cd docs/; sphinx-build -W -T -E . _build/html
 
 doctest:
--- a/eth/db/accesslog.py
+++ b/eth/db/accesslog.py
@@ -1,0 +1,113 @@
+from contextlib import contextmanager
+import logging
+from typing import (
+    Iterator,
+    FrozenSet,
+    Set,
+)
+
+from eth.abc import (
+    AtomicWriteBatchAPI,
+    AtomicDatabaseAPI,
+    DatabaseAPI,
+)
+from eth.db.backends.base import (
+    BaseDB,
+)
+from eth.db.atomic import (
+    BaseAtomicDB,
+)
+
+
+class KeyAccessLoggerDB(BaseDB):
+    """
+    Wraps around a database, and tracks all the keys that were read since initialization.
+    """
+
+    logger = logging.getLogger("eth.db.KeyAccessLoggerDB")
+
+    def __init__(self, wrapped_db: DatabaseAPI, log_missing_keys: bool=True) -> None:
+        """
+        :param log_missing_keys: True if a key is added to :attr:`keys_read` even if the
+            key/value does not exist in the database.
+        """
+        self.wrapped_db = wrapped_db
+        self._keys_read: Set[bytes] = set()
+        self._log_missing_keys = log_missing_keys
+
+    @property
+    def keys_read(self) -> FrozenSet[bytes]:
+        # Make a defensive copy so callers can't modify the list externally
+        return frozenset(self._keys_read)
+
+    def __getitem__(self, key: bytes) -> bytes:
+        try:
+            result = self.wrapped_db.__getitem__(key)
+        except KeyError:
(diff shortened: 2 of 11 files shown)


carver:
Only significant todo is to add a test that accesses are properly logged.

Makefile
```diff
@@ -39,7 +39,6 @@ coverage:
 	open htmlcov/index.html
 
 build-docs:
-	pip install -e .[doc]
 	cd docs/; sphinx-build -W -T -E . _build/html
```
cburgdorf: I guess the idea here was that one would be able to run `make build-docs` without the assumption that docs dependencies were already installed.

eth/db/accesslog.py
```diff
@@ -0,0 +1,106 @@
...
+            key/value does not exist in the database.
+        """
+        self.wrapped_db = wrapped_db
+        self._keys_read: Set[bytes] = set()
+        self._log_missing_keys = log_missing_keys
+
+    @property
+    def keys_read(self) -> Set[bytes]:
+        # Make a defensive copy so callers can't modify the list externally
+        return set(self._keys_read)
```
cburgdorf: Returning a new `set` provides the protection we need but maybe that's a good candidate to use `frozenset` just to drive home the point that this is meant to be readonly.

eth/rlp/blocks.py
```diff
@@ -29,4 +29,5 @@ def __repr__(self) -> str:
         return f'<{self.__class__.__name__}(#{str(self)})>'
 
     def __str__(self) -> str:
-        return f"Block #{self.number}"
+        hash_start = self.hash[:3].hex()
+        return f"Block #{self.number}-0x{hash_start}..."
```
cburgdorf: Did you intentionally not use `humanize_hash` from `eth_utils` here (which produces hashs of the format `eb58..84cd`)?

eth/tools/mining.py
```diff
@@ -1,19 +1,19 @@
...
-
 
-class POWMiningMixin:
+class POWMiningMixin(VirtualMachineAPI):
     """
     A VM that does POW mining as well. Should be used only in tests, when we
     need to programatically populate a ChainDB.
     """
-    def finalize_block(self, block: BaseBlock) -> BaseBlock:
-        block = super().finalize_block(block)  # type: ignore
```
cburgdorf: Always nice to get rid of `type: ignore`

newsfragments/1918.internal.rst
```diff
@@ -0,0 +1,3 @@
+Added :cls:`~eth.db.accesslog.KeyAccessLoggerDB` and its atomic twin; faster ``make validate-docs``
```
cburgdorf: This needs to be `:class:` (I had to these up in the release notes yesterday, which is why know :sweat_smile: )

newsfragments/1918.internal.rst
```diff
@@ -0,0 +1,3 @@
...
+(but you have to remember to ``pip install -e .[doc]`` yourself); ``str(block)`` now includes the
```
cburgdorf: Ah, ok, you are pointing out what I have commented above. Fine by me.

Makefile
carver: Yeah, if it was really fast, I probably wouldn't object, but the pip install actually takes a while. So if you run `make build-docs` over and over, it gets old really fast. Also, a bunch of our other `make` targets assume that you have the appropriate installed tools (wheel, setuptools, towncrier, etc). We could probably document them all better, though. :+1:

eth/rlp/blocks.py
carver: I guess it basically comes down to not liking that format. I guess I can get over it, and consistency is better than my personal stylistic taste. :)

PR #1922, opened 2020-04-14
Bugfix duplicate pytest requirement

carver:
### What was wrong?

pytest-xdist requires a certain pytest version.

### How was it fixed?

remove a duplicate pytest requirement that was the wrong version


carver:
Just wanted to do a CI check. I'll merge it as part of the release.

PR #1924, opened 2020-04-28
Track and expose chain integrity info from HeaderDB

cburgdorf:
### What was wrong?

Py-EVM allows persisting headers with a trusted score via its `persist_checkpoint_header` API. Essentially this allows persisting chains with arbitrary holes between segments. The gained flexibility allows the implementation of various different sync mechanisms.

However, eventually, sync algorithms might want to review the integrity of the chain and fix up holes to get back to a state that doesn't rely on trusted checkpoints.

### How was it fixed?

1. A new utility function `calculate_gaps` was created to track gaps in chains of `BlockNumber`. This can be used to individually track gaps for different purposes (e.g. header, blocks etc)

2. The `HeaderDB` utilizes this function to continuously track gaps in the chain

3. Gaps in the header chain can be read using the `header_db.get_header_chain_gaps()` API.

4. The return type is `Tuple[BlockRange, ...]` with the following characteristics. It returns an ordered sequence of block ranges describing the integrity of the chain of headers. Each block range describes a missing segment in the chain and each range is defined with inclusive boundaries, meaning the first value describes the first missing block of that segment and the second value describes the last missing block of the segment. The last block range in the sequence is expected to have block number of `-1` as the right-hand value which is to say the gap is open-ended. Consequently the value for an empty chain would be `[[1, -1]]`, meaning `1` is the first missing number with the upper end being unknown.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


eth/db/header.py
```diff
@@ -46,6 +50,91 @@ class HeaderDB(HeaderDatabaseAPI):
...
+    def _get_chain_integrity_info(cls, db: DatabaseAPI) -> Tuple[Tuple[int, int], ...]:
+        try:
+            encoded_integrity_info = db[SchemaV1.make_header_chain_integrity_info_lookup_key()]
+        except KeyError:
+            return ()
+        else:
+            return decode_chain_integrity_info(encoded_integrity_info)
+
+    @classmethod
+    def _update_chain_integrity_info(
```
cburgdorf, line 66: The code below could move into a shared place and be used to track the chain integrity of blocks in the same way.

eth/db/header.py
```diff
@@ -46,6 +50,91 @@ class HeaderDB(HeaderDatabaseAPI):
...
+            cls,
+            db: DatabaseAPI,
+            persisted_header: BlockHeaderAPI,
+            base_integrity_info: Tuple[Tuple[int, int], ...] = None) -> Tuple[Tuple[int, int], ...]:
```
pipermerriam, line 70: You might alias `BlockRange = Tuple[BlockNumber, BlockNumber]`  (or `int` if it makes typing easier).

eth/db/header.py
```diff
@@ -46,6 +50,91 @@ class HeaderDB(HeaderDatabaseAPI):
...
+                new_integrity_info = base_integrity_info[:-1] + new_tail
+            elif current_block_number < highest_missing_block:
+                # We are patching a hole which may either shrink an existing hole or divide it
+                matching_holes = [
+                    (index, pair) for index, pair in enumerate(base_integrity_info)
+                    if current_block_number >= pair[0] and current_block_number <= pair[1]
+                ]
+
+                if len(matching_holes) > 1:
+                    raise Exception(
```
cburgdorf, line 99: This should be some kind of Py-EVM exception

eth/db/header.py
```diff
@@ -319,6 +320,8 @@ def _persist_header_chain(
                 rlp.encode(curr_chain_head),
             )
 
+            cls._add_block_number_to_hash_lookup(db, curr_chain_head)
```
cburgdorf, line 323: @carver Can I get your opinion on this?

I discovered a problem that we currently have in Py-EVM. Say you write header `10` as checkpoint header leading to a gap from `1` to `9`. At some point later I go and fetch the headers 1, 2, 3 from a peer with the intention to start closing the gap. Now the problem is, how do we know that the headers 1, 2, 3 do actually belong to the *canonical* chain? The call to `persist_header_chain` does validate the headers in the sense that they match the previous header (genesis, in this example) but it does not know if it will eventually lead us back to the header `10` that we accepted with a trusted score. It could also be an alternative history that we are accepting here. Eventually, as we are closing the gap at some point we will know if the headers match up to our expected header `10` but there is a window of uncertainty where we can't be exactly sure about it I believe.

I have two ideas how to go about that.

1. We could not assign block numbers until we finally closed the gap and then patch up the block numbers backwards from the closing point until we don't hit a missing one anymore. But that would mean that we can not read headers by block number while we are still working on closing the gap. We could still track the gap with the `header_db.get_header_chain_gaps()` API though.

2. We could add a safeguard that goes roughly like: If we are writing headers with a block number lower than the one of our current head and we do not have block numbers assigned to any headers for these block numbers yet, accept them as canonical headers even though we can not entirely be sure that these will eventually lead to our previously accepted trusted score header.

pipermerriam: I believe that the chain rules should supercede a checkpointed header.  In the case that we checkpoint at height 10, and then fill in 1-9 to find out that 9 is **not** the parent of 10, and that there is an alternate header at height 10 that would be picked by the fork choice rule we should discard our non-canonical 10 and all headers onward and re-sync.

At present we may be in a position where if we detect this situation we need to simply crash the client.  In practice I believe that we are unlikely to end up in this scenario without the user doing something very odd.

carver: > At present we may be in a position where if we detect this situation we need to simply crash the client.

Yeah, I think this is the best path, at least in the short term. If the user asserts something that trinity finds to be false, we should crash.

That means we should be extra careful to help people not pick a checkpoint that gets uncled (it was already true, just a reminder).

One thing to note: it gives other clients a way to crash our client, if they know where our checkpoint is: they respond with an uncle at the last unknown header in the gap. So one way to avoid crashing hard is to treat the very last header as invalid if it's not a parent of the canonical block at the child's slot. We would have to print loud warnings, though. But having a client crash that is externally triggerable is a problem.

tests/database/test_header_db.py
```diff
@@ -94,18 +95,128 @@ def test_headerdb_persist_disconnected_headers(headerdb, genesis_header):
...
+    for number in range(1, len(headers)):
+        # Make sure we can lookup the headers by block number
+        header_by_number = headerdb.get_canonical_block_header_by_number(number)
+        assert header_by_number.block_number == headers[number - 1].block_number
+    # Make sure patching the hole does not affect what our current head is
+    assert_headers_eq(headerdb.get_canonical_head(), headers[-1])
+
+
+@pytest.mark.parametrize(
+    'written_headers, evolving_integrity_info',
```
cburgdorf, line 167: Note to self: The tests should also assert the `GapChange`

eth/db/header.py
cburgdorf: Thank you both for your input, it was really helpful.

I was starring on the screen for while, puzzled how to come up with an API that:

- doesn't need a new explicit param such as `force_canonical`
- passes all our tests that ensure tracking of forks and the canonical chain works as it does prior to this PR
- doesn't result in lots of additional db reads to figure out if when it would be appropriate to assign a header a block number (making it part of the canonical chain) when we write headers that are patching gaps.

At some point it occurred to me that `_update_header_chain_gaps` could just include the information of not just what are the new gaps but also what action was taken to get there (e.g. a gap was shrinked).

We can return this info without any additional db read and so I came up with a rule that says we always assign the block number in cases where we know that we have patched a gap in the chain.

This passes all existing tests without incurring much overhead and I'm quite happy with the API. That said, we could also move it to Trinity if you don't feel comfortable introducing these changes to Py-EVM. But I guess it would involve monkey patching Py-EVM to do the tracking which I'm not sure is a good idea.

But anyway, I'm not yet finished with the PR yet. This is just an intermediate update.

>One thing to note: it gives other clients a way to crash our client, if they know where our checkpoint is: they respond with an uncle at the last unknown header in the gap. So one way to avoid crashing hard is to treat the very last header as invalid if it's not a parent of the canonical block at the child's slot. We would have to print loud warnings, though. But having a client crash that is externally triggerable is a problem.

Yeah, I think that part is easy to handle on the Trinity side. :+1:

cburgdorf:
@carver @pipermerriam I think this is ready to get picked apart.

There's also a WIP PR in Trinity that uses this

https://github.com/ethereum/trinity/pull/1714

The gap tracking should be easily reusable to track gabs in the chain of blocks (as opposed to just the chain of headers as it is in this PR) but I'm not yet sure if we might need to make further changes to Py-EVM to properly allow backfilling of blocks ([See my comment in the Py-EVM channel](https://gitter.im/ethereum/py-evm?at=5ebbb2cff25ee47806938cba))

tests/database/test_header_db.py
```diff
@@ -94,18 +96,141 @@ def test_headerdb_persist_disconnected_headers(headerdb, genesis_header):
...
+
+def test_can_patch_holes(headerdb, genesis_header):
+    headerdb.persist_header(genesis_header)
+    headers = mk_header_chain(genesis_header, length=10)
+
+    assert headerdb.get_header_chain_gaps() == ()
+    # Persist the checkpoint header with a trusted score
+    pseudo_genesis = headers[7]
+    headerdb.persist_checkpoint_header(pseudo_genesis, 154618822656)
+    assert headerdb.get_header_chain_gaps() == ((1, 7), (9, -1))
```
carver, line 130: I would have read through this faster with a quick note like:

> headers[7] has block number 8 because it doesn't include the genesis

tests/database/test_header_db.py
```diff
@@ -94,18 +96,141 @@ def test_headerdb_persist_disconnected_headers(headerdb, genesis_header):
...
+        (   # missing three headers in the middle
+            (1, 5),
+            ((GapChange.TailWrite, ((2, -1),)), (GapChange.NewGap, ((2, 4), (6, -1),)))
+        ),
+        (   # missing three headers in the middle, then patching center, dividing existing hole
+            (1, 5, 3),
+            (
+                (GapChange.TailWrite, ((2, -1),)),
+                (GapChange.NewGap, ((2, 4), (6, -1),)),
+                (GapChange.GapSplit, ((2, 2), (4, 4), (6, -1),))
```
carver, line 199:
```py
                (GapChange.GapSplit, (((2, 2), (4, 4)), 6))
```
I just wanted to see what it looks like with the proposed data structure. I feel a little mixed about it, but still seems like a win.

eth/rlp/chain_gaps.py
```diff
@@ -0,0 +1,40 @@
...
+from eth.typing import BlockRange
+
+chain_gaps = rlp.sedes.CountableList(rlp.sedes.List((uint32, uint32)))
+
+# Chain gaps are defined as sequence of markers that define gaps in a chain of connected
+# entities. The right hand side of the very last marker is expected to be -1, meaning the gap
+# is open-ended. E.g. (500, -1) means: Every header from number 500 upwards is missing.
+# [[first_missing, last_missing], ..., [first_missing, -1]]
+# Since RLP doesn't define signed integers, we convert the right-hand side from signed_to_unsigned
+# before entries are written and convert from unsigned_to_signed after entries are read from disk.
```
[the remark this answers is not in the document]
cburgdorf: Oh, I think your proposal is great! It's a really nice improvement to tighten the representation of the chain integrity.

Two examples to double check I got it right:

Now -> Then

1. An empty chain with only genesis

`[[1, -1] ]` -> `[[], 1]`

2. A chain with Checkpoint header 250 and 500 and no other headers

`[[1, 249], [251, 499], [501, -1]]` -> `[[1, 249], [251, 499], 501]`

>and I figure you might like that the types can be tightened from `int` to `BlockNumber`

I didn't do that to avoid bloating the calculations with things like `new_tail = ((highest_missing_number, gap_end), (newly_persisted + BlockNumber(1), BlockNumber(-1)),)` but since we got rid of the `-1` part, this may now become practical :+1:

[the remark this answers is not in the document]
cburgdorf: Ah yeah, I just noticed your example further down `(GapChange.GapSplit, (((2, 2), (4, 4)), 6))` which indicates we are on the same page :+1:

eth/db/chain_gaps.py
```diff
@@ -0,0 +1,82 @@
...
+    TailWrite = enum.auto()
+
+
+GapInfo = Tuple[GapChange, Tuple[BlockRange, ...]]
+
+
+def calculate_gaps(newly_persisted: BlockNumber, base_gaps: Tuple[BlockRange, ...]) -> GapInfo:
+
+    # If we have a fresh chain, our highest missing number can only be 1
+    highest_missing_number = 1 if base_gaps == () else base_gaps[-1][0]
```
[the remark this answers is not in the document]
cburgdorf: >guess the highest range, is always the one at the tail

Correct, but `future_range_start` doesn't really imply that (to me). We could be talking about a range for a new gap somewhere in the center.

What do you think about `known_missing_tip`. As in: the tip might be far further ahead but to us, right now, this is what we think is the first missing number at the tip of the chain.

eth/db/chain_gaps.py
```diff
@@ -0,0 +1,82 @@
...
+                f"2nd gap goes from {matching_gaps[1][1][0]} to {matching_gaps[1][1][1]}"
+            )
+        elif len(matching_gaps) == 0:
+            # Looks like we are just overwriting an existing header.
+            return GapChange.NoChange, base_gaps
+        elif len(matching_gaps) == 1:
+            gap_index, gap = matching_gaps[0]
+            if newly_persisted == gap[0] and newly_persisted == gap[1]:
+                updated_center: Tuple[Tuple[int, int], ...] = ()
+                gap_change = GapChange.GapShrink
```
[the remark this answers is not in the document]
cburgdorf: Yeah, good call, I'll add that :+1:

eth/db/header.py
```diff
@@ -211,6 +256,9 @@ def _persist_header_chain(
             rlp.encode(curr_chain_head),
         )
         score = cls._set_hash_scores_to_db(db, curr_chain_head, score)
+        gap_change, gaps = cls._update_header_chain_gaps(db, curr_chain_head)
+        if gap_change is GapChange.GapShrink or gap_change is GapChange.GapSplit:
+            cls._add_block_number_to_hash_lookup(db, curr_chain_head)
```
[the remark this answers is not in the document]
cburgdorf: >This assertion is made when persisting blocks that ore older than already persisted canonical blocks.

getting nitpicky, just "being older" doesn't qualify. One can write a bunch of older headers for a parallel history and it won't trigger this write. There has to be a known gap to be written to in order to trigger it.

I has led me to the following elaboration:
> The caller implicitly asserts that persisted headers are canonical here.
This assertion is made when persisting headers that are known to be part of a gap
in the canonical chain. While we do validate that the headers are valid, we can not
be 100 % sure that they will eventually lead to the expected child that fills the gap.
E.g. there is nothing preventing one from persisting an uncle as the final header to
close the gap, even if that will not be the parent of the next consecutive header.
Py-EVM makes the assumption that client code will check and prevent such a scenario.

That said, now that I wrote all this and now that we have `GapChange.GapFilled`, wouldn't it be better to simply make that extra check in Py-EVM when we fill a gap. I mean, load the existing known child (the checkpoint header) and if the ends don't match up, raise an exception. It would then be Trinitys job to handle this which probably means to just log a warning and try to close the gap another time (under the assumption that someone tried to feed us an uncle for whatever reason)

tests/database/test_header_db.py
cburgdorf: Yeah, I like that more. Especially because of the cleaner RLP handling but also because it feels we "abused" our previous datastructure a bit. There simply is not `BlockNumber` of `-1` possible.

eth/db/chain_gaps.py
```diff
@@ -0,0 +1,82 @@
...
+            elif newly_persisted == gap[0]:
+                # we are shrinking the gap at the start
+                updated_center = ((gap[0] + 1, gap[1],),)
+                gap_change = GapChange.GapShrink
+            elif newly_persisted == gap[1]:
+                # we are shrinking the gap at the tail
+                updated_center = ((gap[0], gap[1] - 1,),)
+                gap_change = GapChange.GapShrink
+            else:
```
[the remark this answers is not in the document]
cburgdorf: Surprise, I didn't know about this more concise syntax. But I think it should be `elif gap[0] < newly_persisted < gap[1]:` (non-inclusive) because this arm is all about being **not** at the edges and these edge cases (pun intended) are handled already in the previous branches.

eth/db/header.py
```diff
@@ -46,6 +55,40 @@ class HeaderDB(HeaderDatabaseAPI):
...
 
+    def get_header_chain_gaps(self) -> ChainGaps:
+        return self._get_header_chain_gaps(self.db)
+
+    @classmethod
+    def _get_header_chain_gaps(cls, db: DatabaseAPI) -> ChainGaps:
+        try:
+            encoded_gaps = db[SchemaV1.make_header_chain_gaps_lookup_key()]
+        except KeyError:
+            return GENESIS_CHAIN_GAPS
```
cburgdorf, line 66: @carver I want to point out this change. Previously we were just returning an empty tuple here but with the updated datastructure and empty tuple doesn't match the type anymore which is why I came up with `GENESIS_CHAIN_GAPS`. On the plus side, this slightly improves the `calculate_gaps` logic.

cburgdorf:
Thanks for the review @carver
I'm leaving this open for now to give you a chance for a second look. Main points are:

1. Do you like the changes regarding the new datastructure (I do)
2. What do you think about the proposed name change
3. What do you think about tightening the rule so that Py-EVM does one extra check in the `GapFill` case to throw if the gap was filled with an uncle. (Not included yet in the changes)

eth/db/chain_gaps.py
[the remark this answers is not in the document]
carver: Hm, I think of the tip as the header that we have, so missing tip was a little confusing. Maybe `tip_child`?

eth/db/chain_gaps.py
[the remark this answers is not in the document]
carver: Hah, yeah, totally :+1:

eth/db/header.py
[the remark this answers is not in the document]
carver: > now that we have `GapChange.GapFilled`, wouldn't it be better to simply make that extra check in Py-EVM when we fill a gap. I mean, load the existing known child (the checkpoint header) and if the ends don't match up, raise an exception.

Yup, that seems reasonable to me. So then we have to handle this scenario, with a series of inserted headers between A0 and A4, but the B branch is actually uncled. The top row is the presumed canonical one.

```
A0 -> __ -> __ -> __ -> A4

A0 -> B1 -> __ -> __ -> A4

A0 -> B1 -> B2 -> __ -> A4

A0 -> B1 -> B2 -> __ -> A4  # Tries to insert B3, but it's not a parent of A4, and raises an exception

# client somehow finds the A chain and starts inserting it
# it is deemed non-canonical at first (current PR functionality)
A0 -> B1 -> B2 -> __ -> A4
   \_ A1

A0 -> B1 -> B2 -> __ -> A4
   \_ A1 -> A2

# When inserting A3, pyevm notices that it fills a gap, and links with the child
# It notices that A3's parent is present but non-canonical, and makes 
# B* non-canonical and A[1-3] canonical
A0 -> A1 -> A2 -> A3 -> A4
   \_ B1 -> B2
```

eth/db/chain_gaps.py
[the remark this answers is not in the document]
cburgdorf: Sounds good to me!

eth/db/header.py
```diff
@@ -240,6 +287,47 @@ def _persist_header_chain(
...
+        # IMPORTANT: Not only do we have to take this into account when we fill a gap, but also
+        # every time we write into a gap.
+        # Suppose we have a gap of 10 headers, then we fill 1 -5 with uncles from chain B.
+        # Now suppose we find the original chain A and attempt to write headers 1 - 10. At that
+        # point, we are under the assumption our current gap spans from just 6 - 10 because we have
+        # previously written headers 1 - 5 from chain B *believing* they are canonical which they
+        # turn out not to be. That means by the time we write headers 1 -5 again (but from chain A)
+        # we do not recognize it as writing to a known gap. Therefore by the time we write header 6,
+        # when we finally realize that we are attempting to fill in a gap, we have to backtrack
+        # the ancestors to add them to the canonical chain.
```
cburgdorf, line 326: This is quite a wall of text but it helped me understand why we we always want the call to `_find_new_ancestors` here.

tests/database/test_header_db.py
```diff
@@ -94,18 +96,187 @@ def test_headerdb_persist_disconnected_headers(headerdb, genesis_header):
...
+    for number in range(1, 9):
+        # Make sure we can lookup the headers by block number
+        header_by_number = headerdb.get_canonical_block_header_by_number(number)
+        assert header_by_number.block_number == headers[number - 1].block_number
+
+    # Make sure patching the hole does not affect what our current head is
+    assert_headers_eq(headerdb.get_canonical_head(), pseudo_genesis)
+
+
+def test_gap_filling_handles_uncles_correctly(headerdb, genesis_header):
```
cburgdorf, line 145: @carver I found it easier to make two assertions in the same test rather than two individual tests to avoid a big chunk of copy and paste.

eth/db/header.py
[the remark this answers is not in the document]
cburgdorf: Good point! I added a test for that (which also asserts a raised exception in case we are trying to close the gap with `B3`. I found it simpler to cover that in one test with two assertions)

tests/database/test_header_db.py
```diff
@@ -142,6 +142,44 @@ def test_can_patch_holes(headerdb, genesis_header):
...
+    headerdb.persist_header(genesis_header)
+    chain_a = mk_header_chain(genesis_header, length=10)
+    chain_b = mk_header_chain(genesis_header, length=10)
+
+    assert headerdb.get_header_chain_gaps() == GENESIS_CHAIN_GAPS
+    # Persist the checkpoint header with a trusted score
+    # chain_a[7] has block number 8 because `chain_a` doesn't include the genesis
+    pseudo_genesis = chain_a[7]
+    headerdb.persist_checkpoint_header(pseudo_genesis, 154618822656)
```
carver, line 154: Instead of this magic number, maybe `sum(header.difficulty for header in chain_a[0:8])`?

eth/db/header.py
carver: Hm, well, I think we can drop the last paragraph (or at least move it to the test), since it's basically just a test description, and I think it's the test you already wrote. (Could reference `test_gap_filling_handles_uncles_correctly` in a comment here if you want)

eth/db/header.py
```diff
@@ -296,6 +287,47 @@ def _persist_header_chain(
...
+        # This assertion is made when persisting headers that are known to be part of a gap
+        # in the canonical chain.
+        # What we don't know is if this header will eventually lead up to the upper end of the
+        # gap (the checkpoint header) or if this is an alternative history. But we do ensure the
+        # integrity eventually. For one, we check the final header that would close the gap and if
+        # it does not match our expected child (the checkpoint header), we will reject it.
+        # Additionally, if we have persisted a chain of alternative history under the wrong
+        # assumption that it would be the canonical chain, but then at a later point it turns out it
+        # isn't and we overwrite it with the correct canonical chain, we make sure to
+        # recanonicalize all affected headers.
```
carver, line 316: Let's see if we can tighten this up a bit. Something like:

> What if this assertion is later found to be false? At gap fill time, we can detect if the chains don't link (and raise a ValidationError). Also, when a true canonical header is added eventually, we need to canonicalize *all* the true headers.

eth/db/header.py
```diff
@@ -296,6 +287,47 @@ def _persist_header_chain(
...
+                           db: DatabaseAPI,
+                           gap_change: GapChange,
+                           header: BlockHeaderAPI,
+                           genesis_parent_hash: Hash32) -> None:
+
+        if gap_change not in GAP_WRITES:
+            return
+
+        if gap_change is GapChange.GapFill:
+            expected_child = cls._get_canonical_head(db)
```
carver, line 301: Canonical head as the child of the gap seems wrong to me. What about this series of added headers:

```
A2 (set checkpoint)
A3
B0:2
A0:2
```
(using a and b from the test, and python index/slice rules)

carver:
(BTW, thanks for not squashing your changes, it's been nice to just read the new commit each time)

eth/db/header.py
```diff
@@ -296,6 +287,47 @@ def _persist_header_chain(
...
+        # Now suppose we find the original chain A and attempt to write headers 1 - 10. At that
+        # point, we are under the assumption our current gap spans from just 6 - 10 because we have
+        # previously written headers 1 - 5 from chain B *believing* they are canonical which they
+        # turn out not to be. That means by the time we write headers 1 -5 again (but from chain A)
+        # we do not recognize it as writing to a known gap. Therefore by the time we write header 6,
+        # when we finally realize that we are attempting to fill in a gap, we have to backtrack
+        # the ancestors to add them to the canonical chain.
+
+        for ancestor in cls._find_new_ancestors(db, header, genesis_parent_hash):
+            cls._add_block_number_to_hash_lookup(db, ancestor)
```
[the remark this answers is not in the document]
cburgdorf: This will currently raise a `ParentNotFound` exception which I believe is the right thing to do. At least, everything becomes a bit simpler if we just continue with the assumption that headers must always be written sequentially. In other words, gaps aside, we can not write `A2` before `A1`  was written.

eth/db/header.py
cburgdorf: You are absolutely right (and I admire how you spot these things during a review!). I added a fix + test case to cover this.

tests/database/test_header_db.py
cburgdorf: I'm not familiar with `hypothesis` but I've build something home grown that is very simple yet flexible. Maybe that gets the job done?

cburgdorf:
@carver This is ready for another review

eth/db/header.py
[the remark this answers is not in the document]
carver: Ah, right, that does make things simpler. Though I think most of the complexity is added back in with checkpoints. What about multiple checkpoints? Want to make sure re-canonicalization happens correctly (for example, if the checkpoint fills a gap).

```
A3 (set checkpoint)
B0:2 (now canonical)
A2 (set checkpoint)
<- what happens here?
A0:2
<- what about here?
```

carver:
> I haven't fully ingested the logic of this PR but looking at the tests, I realized that this _might_ be an interesting candidate for the [`hypothesis` machinery for testing state machines](https://hypothesis.readthedocs.io/en/latest/stateful.html)

That's a good reminder. It seems promising. It's possible that an [interactive draw](https://hypothesis.readthedocs.io/en/latest/data.html#interactive-draw) could work well (or even better). Something like:

```py
def test_always_canonicalized(data):
    CHAIN_A, CHAIN_B = 0, 1
    chain_length = 10
    headers = [
        # chain A: eventually canonical
        mk_header_chain(genesis_header, length=chain_length),
        # chain B: non-canonical
        mk_header_chain(genesis_header, length=chain_length),
    ]
    missing_indices = [
        set(range(chain_length)),  # chain A 
        set(range(chain_length)),  # chain B
    ] 
    
    @to_tuple 
    def get_valid_extensions(of_missing_indices):
        for chain_index, header_indices in enumerate(of_missing_indices):
            for header_index in header_indices: 
                if header_index == 0 or header_index - 1 not in header_indices: 
                    yield (chain_index, header_index) 
    
    num_checkpoints = 0 

    while len(missing_indices[CHAIN_A]):
        action = data.draw(st.sample_from(ActionEnum))
        if action == ActionEnum.CHECKPOINT:
            # Will py-evm reject a checkpoint that already exists? Assuming so, we can
            # just check the non-present ones...
            checkpoint_index = data.draw(st.sample_from(missing_indices[CHAIN_A]))
            checkpoint = headers[CHAIN_A][checkpoint_index]
            checkpoint_score = get_score(genesis, headers[CHAIN_A][:checkpoint_index])
            headerdb.persist_checkpoint_header(checkpoint, checkpoint_score) 

            # keep track of whether any checkpoints were added, so we eventually
            # guarantee A as canonical
            num_checkpoints += 1 
             
            missing_indices[CHAIN_A].discard(checkpoint_index)
                 
        elif action == ActionEnum.PERSIST_CHAIN:
            # choose the series of headers to add 
            valid_extensions = get_valid_extensions(missing_indices) 
            chain_idx, next_chain_start = data.draw(st.sample_from(valid_extensions))
            next_chain_len = data.draw(st.integers(min_value=1, max_value=chain_length))
            chain_range_end = next_chain_start + next_chain_len
            next_headers = headers[chain_idx][next_chain_start:chain_range_end]

            # persist them to chain
            headerdb.persist_header_chain(next_headers)

            # mark persisted headers as not missing
            for inserted_index in range(next_chain_start, chain_range_end):
                missing_indices[chain_idx].discard(inserted_index)

    if num_checkpoints == 0:
        (subsequent_header, ) = mk_header_chain(headers[CHAIN_A][-1], length=1)
        headerdb.persist_checkpoint_header(subsequent_header, get_score(...))

    assert is_canonical_chain(headers[CHAIN_A])

```

eth/db/header.py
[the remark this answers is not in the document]
cburgdorf: Oh you are bringing up an interesting case there! When we set `A2` as a new checkpoint it becomes the new canonical head and we stop caring about `A3`. I think that is the way it should be (For some reason, the user decided to set the checkpoint to `A2`, imho that should be given precedence).

But know what happens is that `A2` is closing a gap "backwards" which isn't what we want here! We get into a situation where our gap tracking would think we have a well-connected chain which isn't the case!

So, my idea is that `persist_checkpoint_header` must be altered to notice situations when it is backwards filling a gap and then do one of these things:

1. Raise an exception. Let client code now that they can not pick this checkpoint before they bring the database into a different state

2. Alternatively, remove the `B1` header to create a new gap to continue to track this missing connection. Removing just the single header before `A2` should be enough because we do not know where exactly the fork might be (which are uncles, which are actual parents)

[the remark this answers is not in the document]
cburgdorf: I will track this one with an [issue](https://github.com/ethereum/py-evm/issues/1928) and merge the PR as is

cburgdorf:
>I have some vague sense that we may be missing more cases

Yeah, I opened https://github.com/ethereum/py-evm/issues/1928 to track one of them

>but I think this passes the threshold for merging, if you want.

Yeah, I think it good enough to take it and track and fix edge cases with subsequent PRs.

PR #1930, opened 2020-05-20
Handle a lot more scenarios of filling header gaps, created by checkpoints

### LINKED ISSUE #1928: Handle case of gap incorrectly being filled with checkpoint header
### What is wrong?

Consider the following scenario:

```
A3 (set checkpoint)
B0:2 (now canonical)
A2 (set checkpoint)
# At this point `A2` becomes the new head and `get_header_chain_gaps`
# will report no gaps. While that is technically true `B1` is not a parent of `A2`
# and so we allowed getting into a situation where we do not recognize a gap,
# yet don't have a well-connected chain.
```

### How can it be fixed

From https://github.com/ethereum/py-evm/pull/1924#discussion_r427207209

When we set A2 as a new checkpoint it becomes the new canonical head and we stop caring about A3. I think that is the way it should be (For some reason, the user decided to set the checkpoint to A2, imho that should be given precedence).

But know what happens is that A2 is closing a gap "backwards" which isn't what we want here! We get into a situation where our gap tracking would think we have a well-connected chain which isn't the case!

So, my idea is that `persist_checkpoint_header` must be altered to notice situations when it is backwards filling a gap and then do one of these things:

1. Raise an exception. Let client code now that they can not pick this checkpoint before they bring the database into a different state

2. Alternatively, remove the B1 header to create a new gap to continue to track this missing connection. Removing just the single header before A2 should be enough because we do not know where exactly the fork might be (which are uncles, which are actual parents)

### LINKED ISSUE #1929: Test more header-gap-filling scenarios
It's possible (likely?) that there are more broken scenarios, especially with multiple checkpoints. (Like #1928) See earlier conversation:

> this _might_ be an interesting candidate for the [`hypothesis` machinery for testing state machines](https://hypothesis.readthedocs.io/en/latest/stateful.html)

Consider an [interactive draw](https://hypothesis.readthedocs.io/en/latest/data.html#interactive-draw) too, like:

```py
def test_always_canonicalized(data):
    CHAIN_A, CHAIN_B = 0, 1
    chain_length = 10
    headers = [
        # chain A: eventually canonical
        mk_header_chain(genesis_header, length=chain_length),
        # chain B: non-canonical
        mk_header_chain(genesis_header, length=chain_length),
    ]
    missing_indices = [
        set(range(chain_length)),  # chain A 
        set(range(chain_length)),  # chain B
    ] 
    
    @to_tuple 
    def get_valid_extensions(of_missing_indices):
        for chain_index, header_indices in enumerate(of_missing_indices):
            for header_index in header_indices: 
                if header_index == 0 or header_index - 1 not in header_indices: 
                    yield (chain_index, header_index) 
    
    num_checkpoints = 0 

    while len(missing_indices[CHAIN_A]):
        action = data.draw(st.sample_from(ActionEnum))
        if action == ActionEnum.CHECKPOINT:
            # Will py-evm reject a checkpoint that already exists? Assuming so, we can
            # just check the non-present ones...
            checkpoint_index = data.draw(st.sample_from(missing_indices[CHAIN_A]))
            checkpoint = headers[CHAIN_A][checkpoint_index]
            checkpoint_score = get_score(genesis, headers[CHAIN_A][:checkpoint_index])
            headerdb.persist_checkpoint_header(checkpoint, checkpoint_score) 

            # keep track of whether any checkpoints were added, so we eventually
            # guarantee A as canonical
            num_checkpoints += 1 
             
            missing_indices[CHAIN_A].discard(checkpoint_index)
                 
        elif action == ActionEnum.PERSIST_CHAIN:
            # choose the series of headers to add 
            valid_extensions = get_valid_extensions(missing_indices) 
            chain_idx, next_chain_start = data.draw(st.sample_from(valid_extensions))
            next_chain_len = data.draw(st.integers(min_value=1, max_value=chain_length))
            chain_range_end = next_chain_start + next_chain_len
            next_headers = headers[chain_idx][next_chain_start:chain_range_end]

            # persist them to chain
            headerdb.persist_header_chain(next_headers)

            # mark persisted headers as not missing
            for inserted_index in range(next_chain_start, chain_range_end):
                missing_indices[chain_idx].discard(inserted_index)

    if num_checkpoints == 0:
        (subsequent_header, ) = mk_header_chain(headers[CHAIN_A][-1], length=1)
        headerdb.persist_checkpoint_header(subsequent_header, get_score(...))

    assert is_canonical_chain(headers[CHAIN_A])

```

_Originally posted by @carver in https://github.com/ethereum/py-evm/pull/1924#issuecomment-630465536_

carver:
### What was wrong?

Fix #1928
Fix #1929 

### How was it fixed?

Use hypothesis data draws to make aggressive assertions about headers becoming canonical, no matter what order they are inserted. I think Piper is interested in the state machine approach, which may be superior. _(maybe so, but this one is done, and worked pretty well!)_

There was a lot more hiding under the hood here than I expected. Looks like it's all working now. I'm not 100% confident because I had to manually deal with changing header difficulty, but it's certainly an improvement!

I had to explicitly mark which headers are canonical (we could no longer assume that the end of a gap is the beginning of a checkpoint, because of the way we de-canonicalize headers, and insert a gap, if a header doesn't match an adjacent checkpoint).

### To-Do

- [x] Clean up commit history
- [x] Make sure #1928 test is explicitly included
- [x] ~maybe~ _open issue to_ get rid of all the `genesis_parent_hash` passing around, since it's not used when persisting a checkpoint anymore IIRC

#### Cute Animal Picture

### DIFF
--- a/eth/db/chain_gaps.py
+++ b/eth/db/chain_gaps.py
@@ -1,7 +1,11 @@
 import enum
-from typing import Tuple
+from typing import Iterable, Tuple
 
 from eth_typing import BlockNumber
+from eth_utils import (
+    ValidationError,
+    to_tuple,
+)
 
 from eth.exceptions import GapTrackingCorrupted
 from eth.typing import BlockRange, ChainGaps
@@ -12,17 +16,87 @@ class GapChange(enum.Enum):
     NewGap = enum.auto()
     GapFill = enum.auto()
     GapSplit = enum.auto()
-    GapShrink = enum.auto()
+    GapLeftShrink = enum.auto()
+    GapRightShrink = enum.auto()
     TailWrite = enum.auto()
 
 
-GAP_WRITES = (GapChange.GapFill, GapChange.GapSplit, GapChange.GapShrink)
+GAP_WRITES = (
+    GapChange.GapFill,
+    GapChange.GapSplit,
+    GapChange.GapLeftShrink,
+    GapChange.GapRightShrink,
+)
 GENESIS_CHAIN_GAPS = ((), BlockNumber(1))
 
 GapInfo = Tuple[GapChange, ChainGaps]
 
 
-def calculate_gaps(newly_persisted: BlockNumber, base_gaps: ChainGaps) -> GapInfo:
+@to_tuple
+def _join_overlapping_gaps(unjoined_gaps: Tuple[BlockRange, ...]) -> Iterable[BlockRange]:
+    """
+    After introducing a new gap, join any that overlap.
+    Input must already be sorted.
+    """
+    unyielded_low = None
+    unyielded_high = None
+    for low, high in unjoined_gaps:
+        if unyielded_high is not None:
+            if low < unyielded_low:
+                raise ValidationError(f"Unsorted input! {unjoined_gaps!r}")
+            elif unyielded_low <= low <= unyielded_high + 1:
+                unyielded_high = max(high, unyielded_high)
+                continue
+            else:
+                yield unyielded_low, unyielded_high
+
+        unyielded_low = low
+        unyielded_high = high
+
+    if unyielded_high is not None:
+        yield unyielded_low, unyielded_high
+
+
+def reopen_gap(decanonicalized: BlockNumber, base_gaps: ChainGaps) -> ChainGaps:
+    """
+    Add a new gap, for a header that was decanonicalized.
+    """
+    current_gaps, tip_child = base_gaps
+
+    if tip_child <= decanonicalized:
+        return base_gaps
+
+    new_raw_gaps = current_gaps + ((decanonicalized, decanonicalized), )
+
+    # join overlapping gaps
+    joined_gaps = _join_overlapping_gaps(sorted(new_raw_gaps))
+
+    # is the last gap overlapping with the tip child? if so, merge it
+    if joined_gaps[-1][1] + 1 >= tip_child:
+        return joined_gaps[:-1], joined_gaps[-1][0]
+    else:
+        return joined_gaps, tip_child
+
+
+def is_block_number_in_gap(block_number: BlockNumber, gaps: ChainGaps) -> bool:
+    """
+    Check if a block number is found in the given gaps
+    """
+    gap_ranges, tip_child = gaps
+    for low, high in gap_ranges:
+        if low > block_number:
+            return False
+        elif high >= block_number:
+            return True
+        # this range was below the block number, continue looking at the next range
+
+    return block_number >= tip_child
+
+
+def fill_gap(newly_persisted: BlockNumber, base_gaps: ChainGaps) -> GapInfo:
+    """
+    Remove a gap, for a new header that was canonicalized.
+    """
 
     current_gaps, tip_child = base_gaps
 
@@ -45,11 +119,12 @@ def calculate_gaps(newly_persisted: BlockNumber, base_gaps: ChainGaps) -> GapInf
         ]
 
         if len(matching_gaps) > 1:
+            first_match, second_match, *_ = matching_gaps
             raise GapTrackingCorrupted(
                 "Corrupted chain gap tracking",
                 f"No. {newly_persisted} appears to be missing in multiple gaps",
-                f"1st gap goes from {matching_gaps[0][1][0]} to {matching_gaps[0][1][1]}"
-                f"2nd gap goes from {matching_gaps[1][1][0]} to {matching_gaps[1][1][1]}"
+                f"1st gap is {first_match[1]}, 2nd gap is {second_match[1]}",
+                f"all matching gaps: {matching_gaps}",
             )
         elif len(matching_gaps) == 0:
             # Looks like we are just overwriting an existing header.
@@ -62,11 +137,11 @@ def calculate_gaps(newly_persisted: BlockNumber, base_gaps: ChainGaps) -> GapInf
             elif newly_persisted == gap[0]:
                 # we are shrinking the gap at the start
                 updated_center = ((BlockNumber(gap[0] + 1), gap[1],),)
-                gap_change = GapChange.GapShrink
+                gap_change = GapChange.GapLeftShrink
             elif newly_persisted == gap[1]:
                 # we are shrinking the gap at the tail
                 updated_center = ((gap[0], BlockNumber(gap[1] - 1),),)
-                gap_change = GapChange.GapShrink
+                gap_change = GapChange.GapRightShrink
             elif gap[0] < newly_persisted < gap[1]:
                 # we are dividing the gap
                 first_new_gap = (gap[0], BlockNumber(newly_persisted - 1))
--- a/eth/db/header.py
+++ b/eth/db/header.py
@@ -1,5 +1,10 @@
 import functools
-from typing import cast, Iterable, Tuple
+from typing import (
+    cast,
+    Iterable,
+    Sequence,
+    Tuple,
+)
 
 import rlp
 
@@ -35,9 +40,11 @@ from eth.db.chain_gaps import (
     GAP_WRITES,
     GENESIS_CHAIN_GAPS,
     fill_gap,
+    reopen_gap,
 )
 from eth.exceptions import (
     CanonicalHeadNotFound,
+    CheckpointsMustBeCanonical,
     HeaderNotFound,
     ParentNotFound,
 )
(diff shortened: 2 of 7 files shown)


carver:
@cburgdorf hypothesis came up with another fun case:

```
B0
A1 (checkpoint)
A[0:10]
```

So what happens if you fill a gap, then cross a checkpoint, then fill the tail? It looks like it doesn't persist_header_chain() doesn't canonicalize the header at 0 to A.

carver:
Hypothesis found another one:

```py
        (   
            (StepAction.PERSIST_HEADERS, ('b', lambda headers: headers[:1])),                                          
            (StepAction.PERSIST_CHECKPOINT, 2),
            (StepAction.VERIFY_GAPS, (((1, 1), ), 3)),
            (StepAction.PERSIST_HEADERS, ('b', lambda headers: headers[1:3])),                                         
            # This is crashing
            (StepAction.PERSIST_HEADERS, ('a', lambda headers: headers[:1])),                                          
        ),
```

crashes with:
```
cls = <class 'eth.db.header.HeaderDB'>, db = <eth.db.atomic.AtomicDBWriteBatch object at 0x7faa4dc170b8>, gap_change = <GapChange.GapFill: 3>
header = BlockHeader(parent_hash=b'e\x96\xb4\nt\x1c\x8e\xb4-2\xaa\xced\xe0\x16]\xf7\x8f\x03Y\xaar\xf8\x0b\xdd\xef\xbe\x15\xe3\x...0\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', nonce=b'\x00\x00\x00\x00\x00\x00\x00B')
genesis_parent_hash = b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'

    @classmethod
    def _handle_gap_change(cls,
                           db: DatabaseAPI,
                           gap_change: GapChange,
                           header: BlockHeaderAPI,
                           genesis_parent_hash: Hash32) -> None:
    
        if gap_change not in GAP_WRITES:
            return
    
        if gap_change is GapChange.GapFill:
            next_child_number = BlockNumber(header.block_number + 1)
            expected_child = cls._get_canonical_block_header_by_number(db, next_child_number)
            if header.hash != expected_child.parent_hash:
                # We are trying to close a gap with an uncle. Reject!
>               raise ValidationError(f"{header} is not the parent of {expected_child}")
E               eth_utils.exceptions.ValidationError: <BlockHeader #1 bb9c3103> is not the parent of <BlockHeader #2 492097ce>

cls        = <class 'eth.db.header.HeaderDB'>
db         = <eth.db.atomic.AtomicDBWriteBatch object at 0x7faa4dc170b8>
expected_child = BlockHeader(parent_hash=b'\xd6\x99\xa7\xb9e\xe4\x17\xadn\x08m*\x97\xdc9@H\xb61\x9a\xa0\xf06!I\xe1\xf8\xaf\x14\x85\x12\xd6', uncles_hash=b'\x1d\xccM\xe8\xde\xc7]z\xab\x85\xb5g\xb6\xcc\xd4\x1a\xd3\x12E\x1b\x94\x8at\x13\xf0\xa1B\xfd@\xd4\x93G', coinbase=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', state_root=b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!', transaction_root=b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!', receipt_root=b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!', bloom=0, difficulty=17179869184, block_number=2, gas_limit=5000, gas_used=0, timestamp=1590013473, extra_data=b'f\xa0P\xba\x9a>\xf5v9\xf8!Ai\x1d?tz\xab\x02\xfb]\\\xa0H"\x03\xa9\xe6\x8as\xd7\x17', mix_hash=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', nonce=b'\x00\x00\x00\x00\x00\x00\x00B')
gap_change = <GapChange.GapFill: 3>
genesis_parent_hash = (b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
 b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00')
header     = BlockHeader(parent_hash=b'e\x96\xb4\nt\x1c\x8e\xb4-2\xaa\xced\xe0\x16]\xf7\x8f\x03Y\xaar\xf8\x0b\xdd\xef\xbe\x15\xe3\xeff\xef', uncles_hash=b'\x1d\xccM\xe8\xde\xc7]z\xab\x85\xb5g\xb6\xcc\xd4\x1a\xd3\x12E\x1b\x94\x8at\x13\xf0\xa1B\xfd@\xd4\x93G', coinbase=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', state_root=b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!', transaction_root=b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!', receipt_root=b'V\xe8\x1f\x17\x1b\xccU\xa6\xff\x83E\xe6\x92\xc0\xf8n[H\xe0\x1b\x99l\xad\xc0\x01b/\xb5\xe3c\xb4!', bloom=0, difficulty=17179869184, block_number=1, gas_limit=5000, gas_used=0, timestamp=1590013472, extra_data=b"4B\xc1\xb0@\xf3\xdaoA\x83\xf4Pt8M\x10\x8cw'\xeeK\xa8\x19\xdd\xb5\x02S\x1d<]\xbd\x9b", mix_hash=b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', nonce=b'\x00\x00\x00\x00\x00\x00\x00B')
next_child_number = 2

eth/db/header.py:329: ValidationError
```

carver:
> (Probably related to the de-canonicalization of the parent of a checkpoint)

Yup, and it's painful enough to make me wonder if this is the wrong approach.

So:
```
B0 - becomes canonical & head
A1 - persist as checkpoint, re-adds gap at 0
B[1:3] - B2 becomes canonical head, B[0:3] become canonical. -- but gap at 0 not marked as filled
A0 - crash, because it tries to fill the gap at 0, but the canonical child is B1
```

Some options are:

1. Disallow persisting B[1:3] (when the parent is at a gap. It should essentially always be at a checkpoint, so it's like saying you can't persist on top of a checkpoint)
2. Mark gap at 0 as filled when B[1:3] is persisted, overrides the checkpoint
3. Loosen gap filling requirements, allow non-canonical gap-fill as long as there's some child with a parent

I think 1 makes the most sense: disallow (implicitly) overriding a checkpoint

carver:
> I think 1 makes the most sense: disallow (implicitly) overriding a checkpoint

Tried this approach. Not terrible, but not awesome.

carver:
Maaybe the last one:

```
A0
B[0:2] - B becomes canonical
A1 - checkpoint -- A1 becomes canonical, but A0 doesn't
```

carver:
Basically, just checked if parent of new checkpoint is available, and force it to canonical. I think this is insufficient: what if multiple parents are non-canonical. Just fixing that particular case ^.

Probably need to do something like find all ancestors and canonicalize them.

carver:
> Hmm: shouldn't `cls._set_as_canonical_chain_head()` in `persist_checkpoint_header()` already be doing this work?

The persist checkpoint was hacked to avoid looking at any parents. Tried another approach of just halting the lookup when ancestor headers are missing.

carver:
Next hypothesis failure:

```
B[0:4]
A3 -- checkpoint
A[4:6]
A0 -- checkpoint
B[4:6]
A[1:2]
```

Getting a `eth_utils.exceptions.ValidationError: <BlockHeader #3 19b2746b> is not the parent of <BlockHeader #4 ea5d969c>`

So it looks like B3 stays canonical somehow, and so we're not allowed to fill the gap with A2? I'll have to pick it up from here tomorrow.

pipermerriam:
Kind of cool to see hypothesis digging this stuff up!

tests/database/test_header_db.py
```diff
@@ -432,8 +776,8 @@ def assert_is_canonical_chain(headerdb, headers):
 
     # verify that each header is set as the canonical block.
     for header in headers:
-        canonical_hash = headerdb.get_canonical_block_hash(header.block_number)
-        assert canonical_hash == header.hash
+        canonical = headerdb.get_canonical_block_header_by_number(header.block_number)
+        assert canonical == header
```
carver, line 780: Add a comment explaining that this is to make test failures easier to debug

eth/db/chain_gaps.py
```diff
@@ -12,16 +13,74 @@ class GapChange(enum.Enum):
...
+                yield unyielded_low, unyielded_high
+
+        unyielded_low = low
+        unyielded_high = high
+
+    if unyielded_high is not None:
+        yield unyielded_low, unyielded_high
+
+
+def insert_gap(decanonicalized: BlockNumber, base_gaps: ChainGaps) -> ChainGaps:
```
carver, line 54: This insert gap feature is big enough that it belongs in a separate PR, which will be a bit more reviewable.

eth/db/chain_gaps.py
```diff
@@ -12,16 +13,74 @@ class GapChange(enum.Enum):
...
+        if low > block_number:
+            return False
+        elif high >= block_number:
+            return True
+        # this range was below the block number, continue looking at the next range
+
+    return block_number >= tip_child
+
+
 def calculate_gaps(newly_persisted: BlockNumber, base_gaps: ChainGaps) -> GapInfo:
```
carver, line 84: probably need to rename this to distinguish more clearly from `insert_gap`. Also noticed a lot of contortion in the naming. Maybe this is: `gap_after_new_header` and `insert_gap` -> `gap_after_dropping_header`, or something like that.

Otherwise I found myself noting everywhere: insert header == drop gap, and drop header == insert gap, to try to avoid confusion.

tests/database/test_header_db.py
```diff
@@ -325,6 +465,210 @@ def test_gap_tracking(headerdb, genesis_header, written_headers, evolving_gaps):
...
+
+        # 3. final gap is not contiguous with the tail
+        if len(chain_gaps[0]):
+            final_gap_range = chain_gaps[0][-1]
+            assert final_gap_range[1] + 1 < chain_gaps[1], chain_gaps
+
+
+@given(st.data())
+@settings(max_examples=200)
+def test_true_chain_canonicalized_regardless_of_order(genesis_header, data):
```
carver, line 551: This was the test that drove the whole PR forward.

The core idea is that we should be able to checkpoint A at any location, and insert headers for A in any valid order (from genesis or checkpoint forward), and insert B in any valid order. Then after all chain A headers are inserted once, then the whole A chain is canonical.

I suppose that belongs in a docstring ^.

tests/database/test_header_db.py
```diff
@@ -325,6 +465,210 @@ def test_gap_tracking(headerdb, genesis_header, written_headers, evolving_gaps):
...
+                # Persist failed, so retry different action
+                continue
+            except CheckpointsMustBeCanonical:
+                assert chain_idx == CHAIN_B, "Only chain B attempts would decanonize checkpoint"
+                # Persist failed, so retry different action
+                continue
+
+            # mark persisted headers as not missing
+            for inserted_index in range(next_chain_start, chain_range_end):
+                missing_indices[chain_idx].discard(inserted_index)
```
carver, line 651: I suppose this could be:
```suggestion
            missing_indices[chain_idx].difference_update(
                range(next_chain_start, chain_range_end)
            )
```
Can't decide if that's clearer or not.

tests/database/test_header_db.py
```diff
@@ -325,6 +465,210 @@ def test_gap_tracking(headerdb, genesis_header, written_headers, evolving_gaps):
         assert change == evolving_gaps[idx][0]
 
 
+@given(st.lists(
+    st.tuples(
+        # True to insert a header (ie~ remove a gap), False to remove a header (ie~ add a gap)
+        st.booleans(),
```
carver, line 471: Maybe switch this to:
```py
class GapActions(enum.Enum):
  ADD_GAP = enum.auto()
  FILL_GAP = enum.auto()
  
...
        st.sampled_from(GapActions),
```

Note that the order of the enum matters for example shrinking!

carver:
I'll split this into (at least) two before asking for reviews. But if anyone's interested in getting a first feel for the changes now, it's a reasonable time to look (now it's more than just my random explorations). I think the basics of a solution are all present.

A lot more changed than I thought. This addressed more than just the gap tracking, like:
- making sure checkpoints are never de-canonicalized
- when a new canonical head is chosen, make sure to decanonicalize the old niece blocks, even if they have a higher block number than the new head
- more...

eth/db/chain_gaps.py
```diff
@@ -12,16 +16,74 @@ class GapChange(enum.Enum):
...
+    GapChange.GapRightShrink,
+)
 GENESIS_CHAIN_GAPS = ((), BlockNumber(1))
 
 GapInfo = Tuple[GapChange, ChainGaps]
 
 
+@to_tuple
+def _join_overlapping_gaps(unjoined_gaps: Tuple[BlockRange, ...]) -> Iterable[BlockRange]:
+    """Input must already be sorted"""
```
cburgdorf, line 37: *Nit*: The docstring could probably be a bit more descriptive regarding what exactly this function does.

eth/db/chain_gaps.py
cburgdorf: I guess, I'd be fine with either `gap*s*_after_new_header` or `gapinfo_after_new_header`

tests/database/test_header_db.py
```diff
@@ -152,6 +207,111 @@ class StepAction(enum.Enum):
...
+            )),
+        ),
+        # checkpointing should canonicalize matching parent header chain
+        (
+            (StepAction.PERSIST_HEADERS, ('a', lambda headers: headers[:2])),
+            (StepAction.PERSIST_HEADERS, ('b', lambda headers: headers[:3])),
+            (StepAction.PERSIST_CHECKPOINT, 3),
+            (StepAction.VERIFY_GAPS, ((), 4)),
+            (StepAction.VERIFY_CANONICAL_HEADERS, ('a', lambda headers: headers[:3])),
+        ),
```
cburgdorf, line 314: Wow! Quite a bunch of new test cases. Guess your intuition was right. Sorry, for leaving this half baked, I did not see all these possible special cases coming..

eth/db/chain_gaps.py
```diff
@@ -48,7 +110,7 @@ def calculate_gaps(newly_persisted: BlockNumber, base_gaps: ChainGaps) -> GapInf
             raise GapTrackingCorrupted(
                 "Corrupted chain gap tracking",
                 f"No. {newly_persisted} appears to be missing in multiple gaps",
-                f"1st gap goes from {matching_gaps[0][1][0]} to {matching_gaps[0][1][1]}"
+                f"1st gap goes from {matching_gaps[0][1][0]} to {matching_gaps[0][1][1]}",
```
pipermerriam, line 113: Looking at these tripple indexes makes my eyes melt a little.  Is there a `NamedTuple` type we could use to reduce some of the cognitive overhead here?  Something like ....

```python
class Gap(NamedTuple):
    lower_bound: BlockNumber
    upper_bound: BlockNumber
```

eth/db/header.py
```diff
@@ -218,10 +225,103 @@ def _persist_checkpoint_header(
...
+                    #   _set_as_canonical_chain_head()
+                    raise ValidationError(
+                        f"Why was a non-matching parent header {parent_hash} left as canonical "
+                        f"after _set_as_canonical_chain_head() and {true_parent} is available?"
+                    )
+
+        cls._decanonicalize_naughty_children(db, header)
+
+    @classmethod
+    def _decanonicalize_naughty_children(cls, db: DatabaseAPI, header: BlockHeaderAPI) -> None:
```
pipermerriam, line 270: I can't remember if it's just my own terminology but I've used the term *orphaned* for this previously (to refer to headers that were one part of the chain but are no longer canonical).

carver: Heh yeah, maybe `_decanonicalize_new_orphans()` is a better name.

carver:
> Have you thought how this might be further shaken up once we introduce gap tracking for blocks? My intuition says there won't be so much extra overhead because blocks rely on their header already being in the canonical chain so we can rely on the bulk of the complexity being handled by the header tracking.

I haven't really thought about it, no. My first instinct is that most of the difficulty is introduced around accidental non-canonical blocks. So all of that should be handled by the headers before we even get to block tracking. One awkward thing is that after a block is persisted, but the header is decanonicalized, then we have to re-add the block gap too. (Maybe?) So not totally trivial.

carver:
If no one wants me to split the PR, I'm happy to merge it as is. I did at least group two commits that were fairly distinct. I'll merge as-is tomorrow morning if no one objects.

cburgdorf:
:shipit:

PR #1931, opened 2020-05-20
Enable PoW consensus for fully validating chain in test

cburgdorf:
### What was wrong?

The `chain_with_block_validation` fixture was set to `NoProofConsensus`. I believe the idea of the fixture is to operate under `PowConsensus`.

### How was it fixed?

Change consensus engine to `PowConsensus`.

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -18,6 +18,7 @@ from eth.chains.base import (
     Chain,
     MiningChain,
 )
+from eth.consensus import PowConsensus
 from eth.consensus.noproof import NoProofConsensus
 from eth.db.atomic import AtomicDB
 from eth.rlp.headers import BlockHeader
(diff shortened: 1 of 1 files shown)


carver:
Hm, I suspect this was a test speed optimization. Checking the test speed vs master...

carver:
Weird, an accidental shift-click closed instead of commented.

cburgdorf:
>Out of curiosity, are you writing a new test that depends on this?

I'm not but it felt inconsistent with the description. 

>This Chain will perform all validations when importing new blocks, so only
    valid and finalized blocks can be used with it

I also wanted to make sure we don't run into a scenario where we *think* we rely on a chain that checks the proof of work (as the name and description of the fixture suggests) when in reality we use one where `validate_seal` is set to a `noop`.

Issue #1925, opened 2020-05-14
Make sure `persist_block` persists transactions


cburgdorf:
### What is wrong?

Our current Py-EVM APIs aren't fully capable of backfilling blocks without executing them.

Currently `set_block_transactions`  is the place where we deal with RLP encoding transactions and storing them in the db.

https://github.com/ethereum/py-evm/blob/6fef9c25eff570471d147a4f4a6d4eba00262f57/eth/vm/base.py#L315-L322

But this is only being done on the VM level during the execution of blocks. If all I want is to backfill ancient blocks (because we are beam syncing) then just calling something like `persist_block`  wouldn't do the trick when it comes to storing the actual transactions in the db. 

### How can it be fixed

From @carver [via chat](https://gitter.im/ethereum/py-evm?at=5ebc6e388c044c19149be936)

Hm, yes, but that sounds like a bug in persist_block() to me. So I think that means we need:
1) a test that after a persist_block(), it's possible to load a transaction body
2) a test that after a persist_block(), it's possible to load a receipt body
3) implement something like these lines in persist_block() https://github.com/ethereum/py-evm/blob/6fef9c25eff570471d147a4f4a6d4eba00262f57/eth/vm/base.py#L321-L325
4) make sure the transaction root and receipt root in the built tries here ^ match the claimed roots in the header (before persisting them to disk)
This is definitely a place where I'd want to see the red tests from 1 and 2 before implementing 3 and 4

PR #1932, opened 2020-05-20
Ensure persist_block persists tx and receipts

cburgdorf:
### What was wrong?

As noted in #1925, currently `persist_block` does not actually persist the transactions and receipts. It relies on the side effect of block *execution*  to save these.

### How was it fixed?

- Added a new API `persist_unexecuted_block` which takes in an additional `receipts` parameter
- Improved docs to better distinguish when to use which API  
- Added tests that prove `persist_block` does not properly persist transactions and receipts whereas `persist_unexecuted_block` does.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -631,9 +631,34 @@ class ChainDatabaseAPI(HeaderDatabaseAPI):
             as genesis. Providing a ``genesis_parent_hash`` allows storage of blocks that
             aren't (yet) connected back to the true genesis header.
 
-        Assumes all block transactions have been persisted already.
+        .. warning::
+            This API assumes all block transactions have been persisted already. Use
+            :meth:`eth.abc.ChainDatabaseAPI.persist_unexecuted_block` to persist blocks that were
+            not executed.
         """
         ...
+
+    @abstractmethod
+    def persist_unexecuted_block(self,
+                                 block: BlockAPI,
+                                 receipts: Tuple[ReceiptAPI, ...],
+                                 genesis_parent_hash: Hash32 = None
+                                 ) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
+        """
+        Persist the given block's header, uncles, transactions, and receipts. Does
+        **not** validate if state transitions are valid.
+
+        :param block: the block that gets persisted
+        :param receipts: the receipts for the given block
+        :param genesis_parent_hash: *optional* parent hash of the header that is treated
+            as genesis. Providing a ``genesis_parent_hash`` allows storage of blocks that
+            aren't (yet) connected back to the true genesis header.
+
+        This API should be used to persist blocks that the EVM does not execute but which it
+        stores to make them available. It ensures to persist receipts and transactions which
+        :meth:`eth.abc.ChainDatabaseAPI.persist_block` in contrast assumes to be persisted
+        separately.
+        """
 
     @abstractmethod
     def persist_uncles(self, uncles: Tuple[BlockHeaderAPI]) -> Hash32:
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -34,6 +34,7 @@ from eth.constants import (
     EMPTY_UNCLE_HASH,
     GENESIS_PARENT_HASH,
 )
+from eth.db.trie import make_trie_root_and_nodes
 from eth.exceptions import (
     HeaderNotFound,
     ReceiptNotFound,
@@ -127,6 +128,34 @@ class ChainDB(HeaderDB, ChainDatabaseAPI):
                       genesis_parent_hash: Hash32 = GENESIS_PARENT_HASH
                       ) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
         with self.db.atomic_batch() as db:
+            return self._persist_block(db, block, genesis_parent_hash)
+
+    def persist_unexecuted_block(self,
+                                 block: BlockAPI,
+                                 receipts: Tuple[ReceiptAPI, ...],
+                                 genesis_parent_hash: Hash32 = GENESIS_PARENT_HASH
+                                 ) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
+
+        tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
+
+        if tx_root_hash != block.header.transaction_root:
+            raise ValidationError(
+                f"Block's transaction_root ({block.header.transaction_root}) "
+                f"does not match expected value: {tx_root_hash}"
+            )
+
+        receipt_root_hash, receipt_kv_nodes = make_trie_root_and_nodes(receipts)
+
+        if receipt_root_hash != block.header.receipt_root:
+            raise ValidationError(
+                f"Block's receipt_root ({block.header.receipt_root}) "
+                f"does not match expected value: {receipt_root_hash}"
+            )
+
+        with self.db.atomic_batch() as db:
+            self._persist_trie_data_dict(db, receipt_kv_nodes)
+            self._persist_trie_data_dict(db, tx_kv_nodes)
+
             return self._persist_block(db, block, genesis_parent_hash)
 
     @classmethod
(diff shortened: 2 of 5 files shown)


eth/db/chain.py
```diff
@@ -151,6 +152,12 @@ def _persist_block(
             else:
                 tx_hashes = cls._get_block_transaction_hashes(db, header)
 
+            tx_root_hash, tx_kv_nodes = make_trie_root_and_nodes(block.transactions)
+            cls._persist_trie_data_dict(db, tx_kv_nodes)
+
```
cburgdorf, line 157: @carver I'm hitting a little wall here. I was able to write a test that proofs tx aren't saved from `persist_block` unless we include these lines :point_up: 

But I wonder how would I save the receipts here. If I understand correctly, the receipts are a byproduct of block execution which means we would not have them if we don't execute them. My best guess is that we should add another (optional) parameter to `persist_block` that would take receipts (provided by peers during backfill), verifies them, and then stores them accordingly. Is that roughly what we want?

carver: Hm, well the weird thing is that now the transactions are persisted twice on a `Chain.import_block()`, which calls both `import_block()` and `persist_block()`.

So I am pretty tempted to shake up the API a bit more here and say:
- `VM.import_block()` stops being responsible for saving *anything* to the database: it will be read-only, and return anything that needs to be saved (I think it returns almost enough, except for receipts)
- `ChainDB.persist_block()` takes a new *non-optional* receipts argument, and is solely responsible for saving transactions/receipts/etc

Then we could get rid of this hacky bit, too:
https://github.com/ethereum/trinity/blob/32f523128780fc7e24e1582b4bd8d87813396763/trinity/sync/full/chain.py#L885-L886

Because we would have a more formal API for persisting block/transaction/receipt.

cburgdorf: Yes, the double persist is another sub optimal thing. I wasn't exactly sure if there might be a strong reason why we would need to persist directly from `import_block` but I agree we should try to get rid of that and tighten the API :+1:

carver: It occurs to me that it might be difficult to delay persisting the state database changes, but maybe possible. If that turns out to be too big a rabbit-hole, then we can just add an issue and move on...

cburgdorf: I'm looking into this but it it may be more complicated than I thought. E.g. another issue is, as far as I understand we can not delay the calculation of  `receipt_root` and the `transaction_root` until `persist_block`. But if we already calculate that early, then the only way to avoid recalculating it later in `persist_block` would be to cache and pass around all the created nodes which doesn't seem viable. But duplicating the work for every block doesn't sound great either.

I think it just comes down to the fact that there are two fundamental different ways that may lead to the persistence of blocks:

1. We have executed a block and want to persist it
2. We have downloaded a block and have **no desire to execute** it but want to persist it.

Only in the second case would it be reasonable to do the extra work of calculating the `receipt_root` and `transaction_root` as a pre-step to persistence. In the other case (1.) we can not delay this step because even if we do not aim to persist the block just yet, we sure want a full in-memory representation of it **prior to persisting** which just happens to depend on having `receipt_root` and `transaction_root` already calculated.

So, what I'm thinking is that maybe we simple want a new `ChainDB` API `persist_unexecuted_block`:

```python
    def persist_unexecuted_block(self,
                      block: BlockAPI,
                      receipts: Tuple[ReceiptAPI, ...],
                      genesis_parent_hash: Hash32 = GENESIS_PARENT_HASH
                      ) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
```

This isn't much different than adding an *optional*  `receipts` argument to `persist_block` but the separation allows to:
  - only calculate and store `tx_nodes` when we know we have to (it would also be a bit awkward to conditionally do that task based on the fact if `receipts` is `None` or not.
 - only calculate and store `receipt_nodes` when we know we have to
 - only do the verification when we know we have to
 - have a tighter, more meaningful API

cburgdorf: Oh btw, I just noticed that we even had it documented that `persist_block` is based on the assumption that transactions were already persisted seperatedly.

https://github.com/ethereum/py-evm/blob/a876368769ba01cea34271420c10a16862041f44/eth/abc.py#L634

I'm going to update this PR with a new `persist_unexecuted_block` to see what that feels like.

cburgdorf: I updated the PR. What do you think about this solution?

eth/abc.py
```diff
@@ -631,10 +631,33 @@ def persist_block(self,
             as genesis. Providing a ``genesis_parent_hash`` allows storage of blocks that
             aren't (yet) connected back to the true genesis header.
 
-        Assumes all block transactions have been persisted already.
+        WARNING: This API assumes all block transactions have been persisted already. Use
```
pipermerriam, line 634: I think that syntactically for RST this should be `.. warning: This API assumes...`

eth/abc.py
```diff
@@ -631,10 +631,33 @@ def persist_block(self,
...
+        :meth:`eth.abc.ChainDatabaseAPI.persist_unexecuted_block` to persist blocks that were not
+        executed.
         """
         ...
 
+    @abstractmethod
+    def persist_unexecuted_block(self,
```
pipermerriam, line 641: I don't necessarily think it's better but an alternate name that came to mind: `persist_historical_block`

eth/db/chain.py
```diff
@@ -129,6 +130,34 @@ def persist_block(self,
         with self.db.atomic_batch() as db:
             return self._persist_block(db, block, genesis_parent_hash)
 
+    def persist_unexecuted_block(self,
+                                 block: BlockAPI,
+                                 receipts: Tuple[ReceiptAPI, ...],
+                                 genesis_parent_hash: Hash32 = GENESIS_PARENT_HASH
+                                 ) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
+        with self.db.atomic_batch() as db:
```
pipermerriam, line 138: I think this can move down to just above the call to `self._persist_tree_data_dict`.  Probably a minor improvement at best but it reduces the amount of time that the batch is open.

eth/db/chain.py
```diff
@@ -129,6 +130,34 @@ def persist_block(self,
         with self.db.atomic_batch() as db:
             return self._persist_block(db, block, genesis_parent_hash)
 
+    def persist_unexecuted_block(self,
+                                 block: BlockAPI,
+                                 receipts: Tuple[ReceiptAPI, ...],
+                                 genesis_parent_hash: Hash32 = GENESIS_PARENT_HASH
+                                 ) -> Tuple[Tuple[Hash32, ...], Tuple[Hash32, ...]]:
```
pipermerriam, line 137: Might be nice to have some type aliases for this return type to give them a bit more meaning.  Looking at this it isn't clear to me what these two tuples contain.

eth/abc.py
carver: I like that the current name is in-your-face about not validating state transitions.

cburgdorf: Another idea would be `persist_nonexecuted_block` but I'm not sure if that is any better.

eth/db/chain.py
cburgdorf: Sure, no problem. It just took this from the current `persist_block` method.

newsfragments/1925.feature.rst
```diff
@@ -0,0 +1,4 @@
+Add a new ``persist_unexecuted_block`` API to ``ChainDB``. This API should be used to persist
+blocks that the EVM did not execute before it stores them. The API is needed to allow to build
```
carver, line 2:
```suggestion
a block without executing the EVM on it. The API is used by
```

tests/database/test_eth1_chaindb.py
```diff
@@ -194,3 +214,80 @@ def test_chaindb_get_receipt_by_index(
             NUMBER_BLOCKS_IN_CHAIN,
             TRANSACTIONS_IN_BLOCK + 1,
         )
+
+
+@pytest.mark.parametrize(
+    ("use_persist_unexecuted_block"),
```
carver, line 220: *super-nit* why the extra parentheses here?

cburgdorf: yeah, copy pasta leftover

PR #1934, opened 2020-05-27
Update `py_ecc` version

ralexstokes:
Update `py_ecc` dep to the latest version.

#### Cute Animal Picture


setup.py
```diff
@@ -13,7 +13,7 @@
         "eth-utils>=1.7.0,<2.0.0",
         "lru-dict>=1.1.6",
         "mypy_extensions>=0.4.1,<1.0.0",
-        "py-ecc>=1.4.7,<2.0.0",
+        "py-ecc>=4.0.0,<5.0.0",
```
carver, line 16: Thoughts on?
```suggestion
        "py-ecc>=1.4.7,<5.0.0",
```

That way, other consumers of py-evm can use the old library if they have a reason to. It seems we're agnostic to the difference here. (Though in trinity eth2 constraints, we would need to tighten up to py-ecc v4+, of course)

ralexstokes: ahh good point, i'll update this

PR #1937, opened 2020-05-28
[WIP]Implement simple subroutine opcodes

Peppece:
### What was wrong?
EIP 2315 introduces three new opcodes (BEGINSUB, JUMPSUB, ENDSUB) which deal with subroutines and needed to be implemented.

### How was it fixed?
It isn't ATM. I am creating this PR to see if I'm on the right track.
I made the trivial changes: I updated /eth/vm/opcode_values.py, added Berlin fork to eth/vm/forks and created an opcodes.py file inside it, as well as updating eth/vm/mnemonics.py. The test cases implementation is on its way. I just wanted to know if the mantainers think the changes I made up to this point conform to the style of the project, and I also wanted to ask if it was a good idea to create another file inside of the logic directory to implement the logic for the opcodes. The reason I'm asking this is that I think the opcodes belong to a kinda "new" group of opcodes. Do you think I should create a file there or update an already existing one (maybe stack.py)?

### To-Do

- [ ] Implement logic in eth/vm/logic

- [ ] Write tests

#### Cute Animal Picture

(https://images.pexels.com/photos/146083/pexels-photo-146083.jpeg?cs=srgb&dl=2-kuala-bears-on-brown-wood-tree-branch-146083.jpg&fm=jpg)


carver:
@cburgdorf since you started taking a look at Berlin already, do you mind coordinating this review?

cburgdorf:
>since you started taking a look at Berlin already, do you mind coordinating this review?

@carver yep, will do.

cburgdorf:
@Peppece thanks for kicking this off!

I think you could rebase this on top of #1933. This will already give you the basic scaffolding for the `Berlin` fork including the `BerlinVM` which you'll need in `test_opcodes.py`

>I just wanted to know if the mantainers think the changes I made up to this point conform to the style of the project

Yep, so far so good :+1: 

>I also wanted to ask if it was a good idea to create another file inside of the logic directory to implement the logic for the opcodes.

I think the implementation of the new opcodes should go into `/eth/vm/logic/flow.py`

Peppece:
@cburgdorf Awesome! So how do I rebase things on #1933 ? Do I just copy + paste the code in my branch or is there some git fu I'm missing?

cburgdorf:
You can either add my branch as a remote and pull it down or you [edit your `.git/config` to always pull down PRs](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/git-and-github.md#fetch-pull-requests-without-manually-adding-remotes) (makes your life much easier).

Then you can simple do:

```
git fetch origin
git rebase origin/pr/1933
```

cburgdorf:
Hey @Peppece just wanted to hear how things are going? Are you still working on this?

Peppece:
> Hey @Peppece just wanted to hear how things are going? Are you still working on this?

Hey @cburgdorf I was working on this while a school deadline blocked me and I had to focus on that. However, in max 3 days I’ll be back on this.

cburgdorf:
@Peppece Sounds good!

eth/vm/logic/flow.py
```diff
@@ -57,3 +58,36 @@ def gas(computation: BaseComputation) -> None:
...
+    
+    else:
+        computation.code.program_counter += 1
+    
+    computation.rstack_push_int(temp + 1)
+
+
+def returnsub(computation: BaseComputation) -> None:
+
+    if computation.rstack.length == 0:
```
pipermerriam, line 88: rather than doing this check explicitely I think it would more closely follow our existing conventions to have the call to `computation.rstack_pop1_int()` throw an exception and to do this via a `try/except` block.

eth/vm/rstack.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    __slots__ = ['values', '_append', '_pop_int', '__len__']
+
+    def __init__(self) -> None:
+        values: List[int]
+        self.values = values
+        self._append = values.append
+        self._pop_typed = values.pop
+        self.__len__ = values.__len__
```
pipermerriam, line 45: You might consider using [`collections.UserList`](https://docs.python.org/3/library/collections.html#collections.UserList) for this class as I believe it comes built-in with some of the functionality you're doing here.

eth/vm/rstack.py
```diff
@@ -0,0 +1,70 @@
...
+        if not self.values:
+            raise InsufficientStack("Wanted 1 stack item as int, had none")
+        else:
+            item_type, popped = self._pop_typed()
+            if item_type is int:
+                return popped  # type: ignore
+            elif item_type is bytes:
+                return big_endian_to_int(popped)  # type: ignore
+            else:
+                raise _busted_type(item_type, popped)
```
pipermerriam, line 69: I don't think this helper function is justified.  I think it would better fit our conventions to inline the exception message and raising here.

eth/vm/logic/flow.py
```diff
@@ -57,3 +58,36 @@ def gas(computation: BaseComputation) -> None:
...
+
+def jumpsub(computation: BaseComputation) -> None:
+    sub_loc = computation.stack_pop1_int()
+
+    temp = computation.code.program_counter
+    computation.code.program_counter = sub_loc
+
+    next_opcode = computation.code.peek()
+
+    if next_opcode != BEGINSUB:
```
pipermerriam, line 76: I think you need to first check `computation.code.is_valid_opcode(sub_loc)` to ensure that the position is not part of `PUSHXX` data, then check if `computation.code[sub_loc] == BEGINSUB`, **then** you can set the program counter to the new position safely.

This prevents leaving the code stream in an awkward invalid position in the event of an invalid JUMPSUB.

eth/vm/computation.py
```diff
@@ -320,6 +324,17 @@ def stack_push_int(self) -> Callable[[int], None]:
     @cached_property
     def stack_push_bytes(self) -> Callable[[bytes], None]:
         return self._stack.push_bytes
+    
+    #
+    # Return Stack Management
+    #
+    @cached_property
+    def rstack_push_int(self) -> Callable[[int], None]:
```
pipermerriam, line 332: These will need to get propogated down to the `evm.abc.ComputationAPI` class where the general interface is defined.

Peppece:
I only noticed your review now. Will get back to work on this ASAP.

carver:
It's probably a good idea to rebase against the latest master, to minimize conflicts for yourself.

pipermerriam (commented):
You should focus on getting the tests passing.

docs/contributing.rst
```diff
@@ -6,6 +6,15 @@ Thank you for your interest in contributing! We welcome all contributions no mat
 Setting the stage
 ~~~~~~~~~~~~~~~~~
 
+.. note::
+    If you are on macOS and it is the first time you install Py-EVM, you will need to install LevelDB with brew:
+
+    .. code:: sh
+        brew install python3 LevelDB
```
carver: Maybe just add `LevelDB` to this section: https://py-evm.readthedocs.io/en/latest/guides/quickstart.html#installing-on-macos

Maybe it could make sense to link from here to that section, though.

Peppece: Done!

Peppece:
Hello! Is it possible to get a merge on this? Feel like it has been stuck for a while now.

eth/vm/logic/flow.py
```diff
@@ -57,3 +60,39 @@ def gas(computation: BaseComputation) -> None:
...
+    sub_loc = computation.stack_pop1_int()
+    code_range_length = computation.code.__len__()
+
+    if sub_loc >= code_range_length:
+        raise InvalidJumpDestination(
+            "Error: at pc={}, op=JUMPSUB: invalid jump destination".format(
+                computation.code.program_counter)
+        )
+
+    if computation.code.is_valid_opcode(sub_loc):
```
carver, line 79: ~Hm, this valid opcode check looks wrong to me.~ Ah, interesting, this was not made clear by the spec, IMO. But discussion seems to support the idea that the validity check is required: https://ethereum-magicians.org/t/eip-2315-simple-subroutines-for-the-evm/3941/70?u=carver

So... maybe the only thing to do here is to add an explicit test for `JUMPSUB`ing into tho data section of a `PUSH`.

eth/vm/logic/flow.py
```diff
@@ -57,3 +60,39 @@ def gas(computation: BaseComputation) -> None:
...
     computation.stack_push_int(gas_remaining)
+
+
+def beginsub(computation: BaseComputation) -> None:
+    raise OutOfGas("Error: at pc={}, op=BEGINSUB: invalid subroutine entry")
+
+
+def jumpsub(computation: BaseComputation) -> None:
+    sub_loc = computation.stack_pop1_int()
+    code_range_length = computation.code.__len__()
```
carver, line 71: `__len__()` is a "magic" python function that makes `len()` work. So this is preferred:

```suggestion
    code_range_length = len(computation.code)
```

eth/vm/logic/flow.py
```diff
@@ -57,3 +60,39 @@ def gas(computation: BaseComputation) -> None:
...
+
+    if sub_loc >= code_range_length:
+        raise InvalidJumpDestination(
+            "Error: at pc={}, op=JUMPSUB: invalid jump destination".format(
+                computation.code.program_counter)
```
carver, line 76: Maybe include `code_range_length` too?

eth/vm/logic/flow.py
```diff
@@ -57,3 +60,39 @@ def gas(computation: BaseComputation) -> None:
...
+        )
+
+    if computation.code.is_valid_opcode(sub_loc):
+
+        sub_op = computation.code[sub_loc]
+
+        if sub_op == BEGINSUB:
+            temp = computation.code.program_counter
+            computation.code.program_counter = sub_loc + 1
+            computation.rstack_push_int(temp)
```
carver, line 86: Where is the `else` block that aborts if the `sub_op` is not `BEGINSUB`? (and the test that verifies that behavior)

tests/core/opcodes/test_opcodes.py
```diff
@@ -1125,3 +1126,62 @@ def test_blake2b_f_compression(vm_class, input_hex, output_hex, expect_exception
...
+    comp = computation.apply_message(
+        computation.state,
+        computation.msg,
+        computation.transaction_context,
+    )
+    assert comp.is_success
+    assert comp.get_gas_used() == expect_gas_used
+
+
+@pytest.mark.xfail(reason="invalid subroutines")
```
carver, line 1162: `xfail` is better used for things like known bugs. It can hide problems like an incorrect exception instead of a computation that's not successful.

tests/core/opcodes/test_opcodes.py
```diff
@@ -1125,3 +1126,62 @@ def test_blake2b_f_compression(vm_class, input_hex, output_hex, expect_exception
...
+    )
+)
+def test_failing_jumpsub(vm_class, code):
+    computation = setup_computation(vm_class, CANONICAL_ADDRESS_B, decode_hex(code))
+    comp = computation.apply_message(
+        computation.state,
+        computation.msg,
+        computation.transaction_context,
+    )
+    assert comp.is_success
```
carver, line 1187: Even better here would be to assert that it's a failure, and validate the type of exception that is raised. Typically we would do that with a `with pytest.raises(OutOfGasException):`.

carver (commented):
Some items remaining:
- [ ] Rebase on master
- [ ] Drop all the ralexstokes commits

I haven't finished the review, but here are a few thoughts along the way.

eth/vm/logic/flow.py
```diff
@@ -57,3 +60,39 @@ def gas(computation: BaseComputation) -> None:
...
+
+
+def returnsub(computation: BaseComputation) -> None:
+    try:
+        ret_loc = computation.rstack_pop1_int()
+    except InsufficientStack:
+        raise InsufficientStack(
+            "Error: at pc={}, op=RETURNSUB: invalid retsub".format(
```
carver, line 94: I think this error message can be more friendly. Something like:
```suggestion
            "Error: at pc={}, op=RETURNSUB: empty return stack".format(
```

eth/vm/opcode_values.py
```diff
@@ -93,6 +93,14 @@
 JUMPDEST = 0x5b
 
 
+#
+# Subroutines
+#
```
carver, line 98: Since these opcodes fit in the `0x5_` block nicely, maybe just:
```suggestion
# Subroutines
```

eth/vm/rstack.py
carver: Though perhaps it's maybe just as effective to subclass list, as the docs seem to recommend:
> The need for this class has been partially supplanted by the ability to subclass directly from list; however, this class can be easier to work with because the underlying list is accessible as an attribute.

Not sure there's much benefit to being able to access the underlying list via `.data` here, though.

---

Actually, on second thought, we don't really want/need to expose all the `list` methods anyway, so maybe the current approach is better.

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+
+class RStack:
+    """
+    VM Return Stack
+    """
+
+    __slots__ = ['values', '_append', '_pop_typed', '__len__']
+
+    def __init__(self) -> None:
+        values: List[Tuple[type, Union[int, bytes]]] = []
```
carver, line 38: Values can't be `bytes` anymore.

```suggestion
        values: List[int] = []
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+        self.values = values
+        self._append = values.append
+        self._pop_typed = values.pop
+        self.__len__ = values.__len__
+
+    def push_int(self, value: int) -> None:
```
carver, line 44: Since there aren't two type options anymore, this method can just be a push.
```suggestion
    def push(self, value: int) -> None:
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+"""
+This module simply implements for the return stack the exact same design used for the data stack.
+As this stack must simply push_int or pop1_int any time a subroutine is accessed or left, only
+those two functions are provided.
+For the same reason, the class RStack doesn't inherit from the abc StackAPI, as it would require
+to implement all the abstract methods defined.
+"""
+
+
+class RStack:
```
carver, line 30: It's very rare that we use initials in method and class names. A friendlier name here IMO:
```suggestion
class ReturnStack:
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+        self._pop_typed = values.pop
+        self.__len__ = values.__len__
+
+    def push_int(self, value: int) -> None:
+        if len(self.values) > 1023:
+            raise FullStack('Stack limit reached')
+
+        validate_stack_int(value)
+
+        self._append((int, value))
```
carver, line 50:
```suggestion
        self._append(value)
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def pop1_int(self) -> int:
```
carver, line 52:
```suggestion
    def pop1(self) -> int:
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+    VM Return Stack
+    """
+
+    __slots__ = ['values', '_append', '_pop_typed', '__len__']
+
+    def __init__(self) -> None:
+        values: List[Tuple[type, Union[int, bytes]]] = []
+        self.values = values
+        self._append = values.append
+        self._pop_typed = values.pop
```
carver, line 41:
```suggestion
        self._pop = values.pop
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+to implement all the abstract methods defined.
+"""
+
+
+class RStack:
+    """
+    VM Return Stack
+    """
+
+    __slots__ = ['values', '_append', '_pop_typed', '__len__']
```
carver, line 35:
```suggestion
    __slots__ = ['values', '_append', '_pop', '__len__']
```

eth/vm/rstack.py
```diff
@@ -0,0 +1,68 @@
...
+            item_type, popped = self._pop_typed()
+            if item_type is int:
+                return popped  # type: ignore
+            elif item_type is bytes:
+                return big_endian_to_int(popped)  # type: ignore
+            else:
+                raise ValidationError(
+                    "Stack must always be bytes or int, "
+                    f"got {item_type!r} type"
+                )
```
carver, line 68: Having a single type lets us simplify a lot, and even do try-first instead of check-first:
```suggestion
        try:
            return self._pop()
        except IndexError:
            raise InsufficientStack("Wanted 1 return stack item, had none")
```

tests/core/opcodes/test_opcodes.py
```diff
@@ -1125,3 +1126,62 @@ def test_blake2b_f_compression(vm_class, input_hex, output_hex, expect_exception
         comp.raise_if_error()
         result = comp.output
         assert result.hex() == output_hex
+
+
+@pytest.mark.parametrize(
+    'vm_class, code, expect_gas_used',
+    (
+        (
```
carver, line 1134: Would be nice to add a comment here linking to the EIP's test cases:
https://eips.ethereum.org/EIPS/eip-2315#test-cases

tests/core/opcodes/test_opcodes.py
```diff
@@ -1125,3 +1126,62 @@ def test_blake2b_f_compression(vm_class, input_hex, output_hex, expect_exception
...
+            '0x5d5858',
+        ),
+        (
+            BerlinVM,
+            '0x6801000000000000000c5e005c60115e5d5c5d',
+        ),
+        (
+            BerlinVM,
+            '0x5c5d00',
+        ),
```
carver, line 1177: Another failing test case, that should fail with an `InvalidJumpDestination`:
```py
        (
            BerlinVM,
            '0x60035e00',
            InvalidJumpDestination,
        ),
```

tests/core/opcodes/test_opcodes.py
```diff
@@ -1125,3 +1126,62 @@ def test_blake2b_f_compression(vm_class, input_hex, output_hex, expect_exception
...
+        computation.msg,
+        computation.transaction_context,
+    )
+    assert comp.is_success
+    assert comp.get_gas_used() == expect_gas_used
+
+
+@pytest.mark.xfail(reason="invalid subroutines")
+@pytest.mark.parametrize(
+    'vm_class, code',
```
carver, line 1164:
```suggestion
    'vm_class, code, expected_exception',
```

carver (commented):
Okay, I finished the full pass on this. There's enough going on here that I'll do another pass on it after all these changes are in & tests are green. Feel free to @ me then.

Don't worry about trying to get the ethereum/tests Berlin tests going, it's going to be too tough to run those tests until all of the Berlin EIPs are implemented. But there are a couple more manual tests to add that I mentioned.

eth/vm/computation.py
```diff
@@ -140,6 +143,7 @@ def __init__(self,
 
         self._memory = Memory()
         self._stack = Stack()
+        self._rstack = RStack()
```
carver, line 146: Oh, also think that spelling out the full attribute name is friendlier, like:
```suggestion
        self._return_stack = ReturnStack()
```

carver:
Just a heads up that EIP-2315 was pulled from Berlin, so... I guess it makes sense to pause work on this. Sorry friend :/

Peppece:
> Just a heads up that EIP-2315 was pulled from Berlin, so... I guess it makes sense to pause work on this. Sorry friend :/

Wow, didn't see this coming. How is this possible? I mean, hasn't this already been implemented for geth and besu?

cburgdorf:
@Peppece Yes, it was pulled last minute which is something you wouldn't *normaly* expect to happen. You may want to read up on the history here: https://github.com/ethereum/pm/issues/263

This doesn't mean your work was for nothing though. There's a chance that the concerns may be addressed and the EIP gets updated / replaced and ships in some other form in an upcoming fork. Sorry, for the disappointing news :/

PR #1940, opened 2020-05-29
Add Python 3.8 testing (take 2)

### LINKED ISSUE #1872: Add py3.8 tests

veox:
### What was wrong?

Nothing; Python 3.8 has been released (half a year ago :sweat_smile:).

First attempt was in PR #1885, but too many dependencies didn't have binary wheels yet.

### How was it fixed?

Lots of cargo-culting to appease the linters.

- [x] Bumped dependencies and (nested d's) to versions with wheels for Python 3.8.
- [x] Added CI job definitions.
- [x] Fixed linting errors coming from:
  - [x] `flake8`;
  - [x] `mypy`.

Closes #1872 as implemented.

### To-Do

- [x] Clean up commit history
- [x] Open fallout issues: https://github.com/ethereum/py-evm/pull/1940#discussion_r432608081

#### Cute Animal Picture

Source: [Cmycherrytree, via imgur](https://imgur.com/gallery/tAjzq)


veox:
Most linting errors/warnings fixed. Reasons are in commit messages, if anywhere.

On the following:

```
/home/veox/src/py-evm/eth/vm/base.py:248:13: F841 local variable 'exc' is assigned to but never used
/home/veox/src/py-evm/eth/vm/base.py:371:13: F841 local variable 'exc' is assigned to but never used
/home/veox/src/py-evm/eth/consensus/clique/snapshot_manager.py:185:13: F841 local variable 'e' is assigned to but never used
/home/veox/src/py-evm/eth/consensus/clique/snapshot_manager.py:262:9: F841 local variable 'e' is assigned to but never used
/home/veox/src/py-evm/tests/json-fixtures/test_blockchain.py:300:13: F841 local variable 'err' is assigned to but never used
```

... someone should take a closer look: these are all cases where `raise` is used with an (implied) exception.

I'll push a commit shortly so the places are easier to find.

carver:
> ... someone should take a closer look: these are all cases where `raise` is used with an (implied) exception.

Yup, I think the preferred option is the bare `raise` and just drop the `as exc`. The only counter-example I see is:

`/home/veox/src/py-evm/eth/consensus/clique/snapshot_manager.py:262:9: F841 local variable 'e' is assigned to but never used`

Since we're raising a different exception, I almost always prefer chaining the exception. In this case:
```py
        except KeyError as e:                                                                                      
            raise SnapshotNotFound(                                                                                
                f"Can not get on-disk snapshot for {block_hash}"                                                   
            ) from e
```

tox.ini
```diff
@@ -9,7 +9,13 @@ envlist=
...
 max-line-length= 100
 exclude=
-ignore=E252
+ignore =
+    # E252 missing whitespace around parameter equals
+    # Already used too much in the codebase at the point of introduction.
+    E252,
+    # W504 line break after binary operator
+    # It's either this or W503 (line break _before_ binary operator), pick your poison
+    W504
```
carver, line 18: I think we'd rather ignore `W503` here if it's not a huge change to this PR:

That's actually flake8's preferred solution: https://www.flake8rules.com/rules/W503.html

See a nice summary here: https://github.com/raiden-network/raiden/pull/1382
(Roughly, ignoring W503 is flake8's default, but when we started ignoring E252, then flake8 stopped ignoring W503, so we need to add it back again)

tox.ini
```diff
@@ -9,7 +9,13 @@ envlist=
 [flake8]
 max-line-length= 100
 exclude=
-ignore=E252
+ignore =
+    # E252 missing whitespace around parameter equals
+    # Already used too much in the codebase at the point of introduction.
+    E252,
```
carver, line 15: Maybe open a "Good First Issue" to clean all these up? I agree it probably belongs in its own PR, if it touches a lot of lines.

veox: Hundreds!

veox:
`flake8` issues fixed, going through `mypy`...

... A-a-and we're back to [where we ended in the previous PR](https://github.com/ethereum/py-evm/pull/1885#issuecomment-564206008):

```
eth/vm/state.py:165: error: Incompatible return value type (got "Tuple[Hash32, JournalDBCheckpoint]", expected "Tuple[Hash32, UUID]")
eth/vm/state.py:173: error: Argument 1 to "discard" of "AccountDatabaseAPI" has incompatible type "UUID"; expected "JournalDBCheckpoint"
eth/vm/state.py:177: error: Argument 1 to "commit" of "AccountDatabaseAPI" has incompatible type "UUID"; expected "JournalDBCheckpoint"
```

-----

There is now also this (separate issue):

```
eth/vm/state.py:180: error: "AccountDatabaseAPI" has no attribute "lock_changes"
```

I see that `AccountStorageDatabaseAPI` does have the method. So the question is whether the ABC is lacking, a wrong class is used, or is `mypy` bonkers.

carver:
> I see that `AccountStorageDatabaseAPI` does have the method. So the question is whether the ABC is lacking, a wrong class is used, or is `mypy` bonkers.

Looks like `AccountDatabaseAPI` is missing the `lock_changes` stub.

carver:
> eth/vm/state.py:165: error: Incompatible return value type (got "Tuple[Hash32, JournalDBCheckpoint]", expected "Tuple[Hash32, UUID]")

Bug in the state API, should be updated to `JournalDBCheckpoint` (it hasn't been a UUID for a while)

veox:
Huh, looks like that worked.

Not sure I'm not breaking anyone's databases, but here goes.

eth/abc.py
```diff
@@ -1945,6 +1944,16 @@ def commit(self, checkpoint: JournalDBCheckpoint) -> None:
         """
         ...
 
+    @abstractmethod
+    def lock_changes(self) -> None:
+        """
+        Locks in changes, typically just as a transaction starts.
+
+        This is used, for example, to look up the storage value from the start
+        of the transaction, when calculating gas costs in EIP-2200: net gas metering.
```
carver, line 1953: Has a slightly different job than the storage one. It's basically just aggregating them. But while we're at it, I'll see if I can do a better job describing it.

```suggestion
        Locks in changes across all accounts' storage databases.

        This is typically used at the end of a transaction, to make sure that
        a revert doesn't roll back through the previous transaction, and to
        be able to look up the "original" value of any account storage, where
        "original" is the beginning of a transaction (instead of the beginning
        of a block).
        
        See :meth:`eth.abc.AccountStorageDatabaseAPI.lock_changes` for
        what is called on each account's storage database.
```

veox:
Did some more squashing, grouping relevant fixes as much as possible.

PR #1943, opened 2020-05-29
Raise BlockNotFound on missing txns or uncles

carver:
### What was wrong?

Having py-trie make `MissingTrieNode` stop subclassing `KeyError` exposed an awkwardness mentioned here in trinity:
https://github.com/ethereum/trinity/blob/e40aac0ab2a9a2e7a703af0b94fafd1047ce6e55/trinity/sync/beam/chain.py#L330-L331

The exception story for loading a block body was pretty messy.

To be clear, Trinity breaks if you upgrade to py-trie v2a1 without this change (although it still requires a change to catch the `BlockNotFound`, but that's an improvement).

### How was it fixed?

Raise a `BlockNotFound` if the transactions or uncles are missing.

Bonus change: bring back a performance boost when persisting a trie (using a batch write).

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -3197,6 +3197,9 @@ class ChainAPI(ConfigurableAPI):
     def get_block_by_hash(self, block_hash: Hash32) -> BlockAPI:
         """
         Return the requested block as specified by ``block_hash``.
+
+        :raise eth.exceptions.HeaderNotFound: if the header is missing
+        :raise eth.exceptions.BlockNotFound: if any part of the block body is missing
         """
         ...
 
@@ -3204,6 +3207,8 @@ class ChainAPI(ConfigurableAPI):
     def get_block_by_header(self, block_header: BlockHeaderAPI) -> BlockAPI:
         """
         Return the requested block as specified by the ``block_header``.
+
+        :raise eth.exceptions.BlockNotFound: if any part of the block body is missing
         """
         ...
 
--- a/eth/exceptions.py
+++ b/eth/exceptions.py
@@ -33,6 +33,8 @@ class HeaderNotFound(PyEVMError):
 class BlockNotFound(PyEVMError):
     """
     Raised when the block with the given number/hash does not exist.
+    This will happen, for example, if the transactions or uncles are not
+    saved in the database.
     """
     pass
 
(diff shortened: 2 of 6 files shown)


tests/database/test_eth1_chaindb.py
```diff
@@ -291,3 +292,19 @@ def test_chaindb_persist_unexecuted_block(chain,
             NUMBER_BLOCKS_IN_CHAIN,
             TRANSACTIONS_IN_BLOCK + 1,
         )
+
+
+def test_chaindb_raises_blocknotfound_on_missing_uncles(VM, chaindb, header):
```
carver, line 297: Importantly, this tests against all VMs, since the implementation is currently sitting in the `Frontier` vm.

eth/exceptions.py
```diff
@@ -33,6 +33,7 @@ class HeaderNotFound(PyEVMError):
 class BlockNotFound(PyEVMError):
     """
     Raised when the block with the given number/hash does not exist.
+    This might happen if a header is present, but the transactions or uncles are not.
```
carver, line 36: Maybe clearer to say:

```suggestion
    This will happen, for example, if the transactions or uncles are
    not saved in the database.
```

eth/abc.py
```diff
@@ -3197,13 +3197,18 @@ def get_block(self) -> BlockAPI:
     def get_block_by_hash(self, block_hash: Hash32) -> BlockAPI:
         """
         Return the requested block as specified by ``block_hash``.
+
+        :raise eth.exceptions.HeaderNotFound: if the header is missing
+        :raise eth.exceptions.BlockNotFound: if the block body is missing
```
gsalgado, line 3202: Maybe reword this to say "...: if any part of the block body is missing"?

eth/abc.py
```diff
@@ -3197,13 +3197,18 @@ def get_block(self) -> BlockAPI:
...
         """
         ...
 
     @abstractmethod
     def get_block_by_header(self, block_header: BlockHeaderAPI) -> BlockAPI:
         """
         Return the requested block as specified by the ``block_header``.
+
+        :raise eth.exceptions.BlockNotFound: if the block body is missing
```
gsalgado, line 3211: same here

PR #1947, opened 2020-06-22
Track gaps in the chain of blocks

cburgdorf:
### What was wrong?

Similar to how we track gaps in the chain of headers, we also need to track gaps in the chain of blocks.

### How was it fixed?

1. Taught `mine_blocks` of chain builer tools to accept a `transactions` argument to make it easier to write tests where blocks are mined with transactions

2. Track gaps in `persist_block` (similar but simpler to header gap tracking)

3. Overwrite `_update_header_chain_gaps` in `ChainDB` to hook into the `HeaderDB` gap tracking to react to cases where an uncle header gets replaced and should trigger an update to the block gap tracking.

### To-Do

- [ ] Squash changes into two commits

#### Cute Animal Picture

### DIFF
--- a/eth/db/chain.py
+++ b/eth/db/chain.py
@@ -35,6 +35,14 @@ from eth.constants import (
     EMPTY_UNCLE_HASH,
     GENESIS_PARENT_HASH,
 )
+from eth.db.chain_gaps import (
+    fill_gap,
+    GapChange,
+    GapInfo,
+    GENESIS_CHAIN_GAPS,
+    is_block_number_in_gap,
+    reopen_gap,
+)
 from eth.db.trie import make_trie_root_and_nodes
 from eth.exceptions import (
     HeaderNotFound,
@@ -49,6 +57,8 @@ from eth.rlp.headers import (
 from eth.rlp.receipts import (
     Receipt
 )
+from eth.rlp.sedes import chain_gaps
+from eth.typing import ChainGaps
 from eth.validation import (
     validate_word,
 )
@@ -74,6 +84,73 @@ class TransactionKey(rlp.Serializable):
 class ChainDB(HeaderDB, ChainDatabaseAPI):
     def __init__(self, db: AtomicDatabaseAPI) -> None:
         self.db = db
+
+    def get_chain_gaps(self) -> ChainGaps:
+        return self._get_chain_gaps(self.db)
+
+    @classmethod
+    def _get_chain_gaps(cls, db: DatabaseAPI) -> ChainGaps:
+        try:
+            encoded_gaps = db[SchemaV1.make_chain_gaps_lookup_key()]
+        except KeyError:
+            return GENESIS_CHAIN_GAPS
+        else:
+            return rlp.decode(encoded_gaps, sedes=chain_gaps)
+
+    @classmethod
+    def _update_chain_gaps(
+            cls,
+            db: DatabaseAPI,
+            persisted_block: BlockAPI,
+            base_gaps: ChainGaps = None
+    ) -> GapInfo:
+
+        # If we make many updates in a row, we can avoid reloading the integrity info by
+        # continuously caching it and providing it as a parameter to this API
+        if base_gaps is None:
+            base_gaps = cls._get_chain_gaps(db)
+
+        gap_change, gaps = fill_gap(persisted_block.number, base_gaps)
+        if gap_change is not GapChange.NoChange:
+            db.set(
+                SchemaV1.make_chain_gaps_lookup_key(),
+                rlp.encode(gaps, sedes=chain_gaps)
+            )
+
+        return gap_change, gaps
+
+    @classmethod
+    def _update_header_chain_gaps(
+            cls,
+            db: DatabaseAPI,
+            persisting_header: BlockHeaderAPI,
+            base_gaps: ChainGaps = None
+    ) -> GapInfo:
+        # The only reason we overwrite this here is to be able to detect when the HeaderDB
+        # de-canonicalizes an uncle that should cause us to re-open a block gap.
+        gap_change, gaps = super()._update_header_chain_gaps(db, persisting_header, base_gaps)
+
+        if gap_change is not GapChange.NoChange or persisting_header.block_number == 0:
+            return gap_change, gaps
+
+        # We have written a header for which block number we've already had a header.
+        # This might be a sign of a de-canonicalized uncle.
+        current_gaps = cls._get_chain_gaps(db)
+        if not is_block_number_in_gap(persisting_header.block_number, current_gaps):
+            # ChainDB believes we have that block. If the header has changed, we need to re-open
+            # a gap for the corresponding block.
+            old_canonical_header = cls._get_canonical_block_header_by_number(
+                db,
+                persisting_header.block_number
+            )
+            if old_canonical_header != persisting_header:
+                updated_gaps = reopen_gap(persisting_header.block_number, current_gaps)
+                db.set(
+                    SchemaV1.make_chain_gaps_lookup_key(),
+                    rlp.encode(updated_gaps, sedes=chain_gaps)
+                )
+
+        return gap_change, gaps
 
     #
     # Header API
@@ -196,6 +273,7 @@ class ChainDB(HeaderDB, ChainDatabaseAPI):
         old_canonical_hashes = tuple(
             header.hash for header in old_canonical_headers)
 
+        cls._update_chain_gaps(db, block)
         return new_canonical_hashes, old_canonical_hashes
 
     def persist_uncles(self, uncles: Tuple[BlockHeaderAPI]) -> Hash32:
--- a/newsfragments/1947.feature.rst
+++ b/newsfragments/1947.feature.rst
@@ -1,0 +1,1 @@
+Expose ``get_chain_gaps()`` on ``ChainDB`` to track gaps in the chain of blocks.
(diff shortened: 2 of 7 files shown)


tests/database/test_eth1_chaindb.py
```diff
@@ -84,13 +87,135 @@ def chain(chain_without_block_validation):
...
+        # Uncle block gets de-canonicalized by a header that has transactions
+        (False, True, False, (((4, 6),), 8),),
+        # Has uncle and transactions
+        (True, True, False, (((4, 6),), 8),),
+        # Uncle block gets de-canonicalized by a header that has no uncles nor transactions
+        # In this scenario, we do not need to re-open the gap because the new block is "trivial"
+        # which means it can be loaded by just the header alone.
+        (False, False, True, (((5, 6),), 8),),
+    )
+)
```
cburgdorf, line 116: @carver Since block gap tracking "inherits" from header gap tracking (There is no way to persist a block that bypasses the header gap tracking) my hope is that these cases are the only ones we have to cover....I hope :sweat_smile:

eth/db/chain.py
```diff
@@ -75,15 +86,92 @@ class ChainDB(HeaderDB, ChainDatabaseAPI):
...
+    ) -> GapInfo:
+        # The only reason we overwrite this here is to be able to detect when the HeaderDB
+        # de-canonicalizes an uncle that should cause re-opening a block gap.
+        gap_change, gaps = super()._update_header_chain_gaps(db, persisted_header, base_gaps)
+        # We have overwritten an existing header, this might be a sign of a de-canonicalized uncle
+        if gap_change is GapChange.NoChange:
+            current_gaps = cls._get_chain_gaps(db)
+            if not is_block_number_in_gap(persisted_header.block_number, current_gaps):
+                # ChainDB believes we have that block, find out if we actually have it.
+                try:
```
cburgdorf, line 138: Not sure if there's any easier way to do this here in the `ChainDB`?  If not then it might even make sense to extract this into a `def block_exists(block_hash`) method?

newsfragments/1947.feature.rst
```diff
@@ -0,0 +1 @@
+Expose ``get_chain_gaps()`` on ``ChainDB`` to track gaps in the chain of blocks.
```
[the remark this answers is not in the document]
cburgdorf: Are you saying you do not count block gap tracking + a new public API on `ChainDB` as a new feature? We also counted `get_header_chain_gaps` as a new feature: https://github.com/ethereum/py-evm/blob/8f7cb5ff6fd2b8bf3e6fbb551f149439c633536f/newsfragments/1924.feature.rst

eth/db/chain.py
cburgdorf: I can be on board with that but I do want to bring up one aspect that might be of importance. 

Currently, we hook into `GapChange.NoChange` and then check if we can load the block to figure out if we need to re-open the gap. If we would eliminate that check and simply re-open the block on every `GapChange.NoChange` then this would mean we would always mark blocks as missing whenever we overwrite headers. I guess this could lead to noticeable more request load. E.g. when we restart Trinity and the gap fill sync starts it will jump back a couple of thousand headers and every overwritten header would re-open a block gap which would result into us unnecessarily re-requesting blocks that we already have.

That's why I thought it would be good to do a little extra effort to figure out if we need to mark the block as gap or not. If you think that isn't a big issue then I'm happy to get rid of that extra check.

newsfragments/1947.feature.rst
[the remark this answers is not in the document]
carver: Ah, no, I was thinking in a trinity mindset. :+1: for a feature.

eth/db/chain.py
carver: > I guess this could lead to noticeable more request load. E.g. when we restart Trinity and the gap fill sync starts it will jump back a couple of thousand headers and every overwritten header would re-open a block gap which would result into us unnecessarily re-requesting blocks that we already have.

Ah, my mental model is that the block gap filler would check the local DB to see if the block can be loaded before it asks peers. So it would mean an increase in database hits, which we can probably optimize later.

Before looking into it too deeply, I wonder if we can hook into somewhere else that does this check, since we're interested in a really specific subset of cases, where the block numbers of the old canonical headers match the block numbers of the new canonical headers. So maybe the check can be somewhere that that information is available.

> That's why I thought it would be good to do a little extra effort to figure out if we need to mark the block as gap or not. If you think that isn't a big issue then I'm happy to get rid of that extra check.

Yeah, I definitely get the motivation, and I wouldn't be surprised if we had to come back to it later to speed it up a bit. I just think the balance of preferring the simpler & more general solution is worth the cost of the performance hit for now. :+1:

cburgdorf: >Before looking into it too deeply, I wonder if we can hook into somewhere else that does this check, since we're interested in a really specific subset of cases, where the block numbers of the old canonical headers match the block numbers of the new canonical headers.

Yes, in fact, we can do that check here. We can load the old header at that block number and compare. The downside of this is that it relies on the fact that the header behind the block number wasn't changed at that instant (it will happen later in the same atomic db write, just not yet). So, if things inside `persist_header_chain` would be refactored, this could fall apart. I personally think that is fine though, especially now that we have tests for it. I also added a test case that demonstrates that simply overwriting the exact same header won't cause us to re-open a block gap.

How do you feel about this solution?

tests/core/builder-tools/test_chain_builder.py
```diff
@@ -19,13 +20,18 @@
...
 MINING_CHAIN_PARAMS = (
     MiningChain,
     frontier_at(0),
     disable_pow_check,
-    genesis(),
+    genesis(
+        params={'gas_limit': 1000000},
+        state={FUNDED_ADDRESS: {'balance': to_wei(1000, 'ether')}}
+    ),
 )
```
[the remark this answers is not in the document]
cburgdorf: Sounds good. Refactored it.

cburgdorf:
Alright, this is ready for another round of review @carver

tests/database/test_eth1_chaindb.py
```diff
@@ -84,13 +88,153 @@ def chain(chain_without_block_validation):
...
+    # No more gaps in the header chain
+    assert gap_chain.chaindb.get_header_chain_gaps() == ((), 8)
+    # We detected the de-canonicalized uncle and re-opened the block gap
+    assert gap_chain.chaindb.get_chain_gaps() == (((4, 6),), 8)
+
+    # The uncle block was implicitly de-canonicalized with its header,
+    # hence we can not fetch it any longer.
+    if can_fetch_block:
+        block_4 = gap_chain.get_canonical_block_by_number(4)
+        assert block_4 == main_chain.get_canonical_block_by_number(4)
```
carver, line 226: So this is that new situation, where the chain gaps say that the block isn't there, but we can still load the block. That's probably interesting enough to add a comment about it here.

tests/database/test_eth1_chaindb.py
```diff
@@ -84,13 +88,153 @@ def chain(chain_without_block_validation):
...
+    else:
+        with pytest.raises(BlockNotFound):
+            gap_chain.get_canonical_block_by_number(4)
+        # Add the missing block and assert the gap shrinks
+        assert gap_chain.chaindb.get_chain_gaps() == (((4, 6),), 8)
```
carver, line 231: This line is redundant because nothing has changed since the last time we asserted the same thing. But if you like it being redundant for the comparison to line 235, I don't really object. Just making sure it's not accidental code cruft.

eth/db/chain.py
```diff
@@ -75,6 +85,73 @@ class ChainDB(HeaderDB, ChainDatabaseAPI):
...
+            old_canonical_header = cls._get_canonical_block_header_by_number(
+                db,
+                persisted_header.block_number
+            )
+            if old_canonical_header != persisted_header:
+                updated_gaps = reopen_gap(persisted_header.block_number, current_gaps)
+                db.set(
+                    SchemaV1.make_chain_gaps_lookup_key(),
+                    rlp.encode(updated_gaps, sedes=chain_gaps)
+                )
```
carver, line 151: Yeah, this approach feels cleaner to me!

In fact, it's so clean, that it highlights the mis-named variable `persisted_header`. It must not have been persisted yet, otherwise this `if` check would always be `False` (as you mentioned in your other comment). So a better name might be:
- `candidate_header`
- `proposed_header`
- `prospective_header`
- `persisting_header`
- `next_header`

Haha, I got a little carried away with a thesaurus. Some others that I kind of like, but mostly because they make me smile rather than because they are the best choice (they sound a little old-fashioned or mystical):
- `impending_header`
- `destined_header`

tests/database/test_eth1_chaindb.py
cburgdorf: >but we can still load the block?

No, we load it from the `main_chain`. The `gap_chain` can't load it which is proven by the `pytest.raises(BlockNotFound)`

tests/database/test_eth1_chaindb.py
cburgdorf: Yeah, I put it in on purpose just to make the reader aware again of the current situation in terms of gaps.

eth/db/chain.py
cburgdorf: `persisting_header` header makes the most sense to my non-native ears ;)

tests/database/test_eth1_chaindb.py
cburgdorf: Ah wait. I was looking at the wrong line numbers. You are right, in fact...let me investigate real quick.

cburgdorf: Ah, ok sure! We can load the block because it's a trivial header in that case. I'll add a comment

PR #1949, opened 2020-07-07
Remove unused add_uncle API

cburgdorf:
### What was wrong?

Looks like the `add_uncle` API on blocks is not used anywhere.

### How was it fixed?

Removed dead API.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/vm/forks/frontier/blocks.py
+++ b/eth/vm/forks/frontier/blocks.py
@@ -4,7 +4,6 @@ from typing import (
     Type,
 )
 
-import rlp
 from rlp.sedes import (
     CountableList,
 )
@@ -18,7 +17,6 @@ from eth_typing import (
     Hash32,
 )
 
-from eth_hash.auto import keccak
 from trie.exceptions import (
     MissingTrieNode,
 )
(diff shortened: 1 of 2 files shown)


cburgdorf:
@carver maybe after this is merged we should cut a new release to make sure the increased type safety actually makes it into Trinity.

carver:
Hm, yeah I think I'm still in favor of removing this. But also curious why mypy isn't complaining about us initializing uncles to `[]`. We should probably also force uncles to a tuple in the init, and have a test that confirms that even if we init with a list, then uncles are returned as a tuple.

cburgdorf:
>But also curious why mypy isn't complaining about us initializing uncles to []

Because the whole `__init__` isn't type checked since it derives from `rlp.Serializable`

>We should probably also force uncles to a tuple in the init,

Afaik, the `rlp` library does that for us already.

```python
import rlp
from rlp.sedes import CountableList, Boolean, Serializable

class Foo(Serializable):
    fields = [
        ('things', rlp.sedes.CountableList(Boolean))
    ]

bar = Foo(things=[True, False])

print(bar.things)
print(type(bar.things))
```

prints

```
(True, False)
<class 'tuple'>
```

PR #1950, opened 2020-07-21
Call a new hook when a transaction is applied

carver:
### What was wrong?

No convenient way to report progress in the middle of `apply_all_transactions` (the longest step of `import_block`).

See https://github.com/ethereum/trinity/pull/1916/files#r458190667

### How was it fixed?

- Added a new `VirtualMachineAPI.transaction_applied_hook()`
- In the hook, supply a header that includes the expected amount of gas to be used by the full block. This allows for things like monitoring the progress of the block in between transactions if they are slow. Executing transactions can be very slow in Beam Sync, for example.

We have to explicitly annotate the header used to generate the VM with the gas used by the final header in order to do that last step.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -468,7 +468,10 @@ class Chain(BaseChain):
             )
 
         base_header_for_import = self.create_header_from_parent(parent_header)
-        block_result = self.get_vm(base_header_for_import).import_block(block)
+        # Make a copy of the empty header, adding in the expected amount of gas used. This
+        #   allows for richer logging in the VM.
+        annotated_header = base_header_for_import.copy(gas_used=block.header.gas_used)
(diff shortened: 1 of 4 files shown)


eth/chains/base.py
```diff
@@ -468,7 +468,9 @@ def import_block(self,
             )
 
         base_header_for_import = self.create_header_from_parent(parent_header)
-        block_result = self.get_vm(base_header_for_import).import_block(block)
+        # Make a copy of the clean header with some annotations, for richer logging in the VM
```
pipermerriam, line 471: This comment didn't make it any clearer to me why we need to make this copy.  My educated guess is that it gives access to the expected total gas used by this block which otherwise we wouldn't be able to access.

PR #1952, opened 2020-08-20
Updates on subgroup check

ChihChengLiang:
### What was wrong?

Come across the pairing precompile and find these 2 issues.

1.  TODO error message on pairing.
2. Observed that the p1 subgroup check is missing.

I'm actually uncertain about the second issue and just keeping the diff there for better discussion.
I'm seeing conflict signs of whether to do that check.
- The fixture tests have [no case](https://github.com/ethereum/tests/pull/328) for the invalid p1 that's not in a subgroup.
- Go-ethereum [checks](https://github.com/ethereum/go-ethereum/blob/295693759e5ded05fec0b2fb39359965b60da785/core/vm/contracts.go#L848) the p1 in subgroup 

### How was it fixed?

1. Fix the wording of the error message.
2. Add a subgroup check for the p1.

#### Cute Animal Picture

image

### DIFF
--- a/eth/_utils/bn128.py
+++ b/eth/_utils/bn128.py
@@ -25,6 +25,8 @@ def validate_point(x: int, y: int) -> Tuple[bn128.FQ, bn128.FQ, bn128.FQ]:
         p1 = (FQ(x), FQ(y), FQ(1))
         if not bn128.is_on_curve(p1, bn128.b):
             raise ValidationError("Point is not on the curve")
+        if bn128.multiply(p1, bn128.curve_order)[-1] != FQ.zero():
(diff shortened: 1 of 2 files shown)


ChihChengLiang:
Sorry, I checked the go-Ethereum again they don't check subgroup in bn256. https://github.com/ethereum/go-ethereum/blob/9efc1a847e53b63847f6f95e0857b1a6300786eb/crypto/bn256/cloudflare/bn256.go#L298

sylvainpelissier:
For the first group you do not need subgroup check since the order of the curve is prime the group is the elliptic curve itself. Any point on the curve is in the group: https://hackmd.io/@jpw/bn254

Issue #1942, opened 2020-05-29
Remove linting error ignores


veox:
* py-evm Version: not yet applicable, true only if/when PR #1940 is merged
* OS: N/A
* Python Version (python --version): N/A
* Environment (relevant `diff` part for `setup.py` from linked PR):

``` diff
     'lint': [
-        "flake8==3.5.0",
-        "flake8-bugbear==18.8.0",
-        "mypy==0.701",
+        "flake8==3.8.2",
+        "flake8-bugbear==20.1.4",
+        "mypy==0.750",
     ],
```

### What is wrong?

The PR added some `ignore`s to `tox.ini`, it was previously empty:

```
12 ignore =
13     # E252 missing whitespace around parameter equals                                                               
14     # Already used too much in the codebase at the point of introduction.                                           
15     E252,
16     # W503 line break before binary operator                                                                        
17     # It's either this or W504 (line break _after_ binary operator), pick your poison.                              
18     # W503 gets enabled when E252 gets ignored, see:                                                                
19     # https://github.com/ethereum/py-evm/pull/1940#discussion_r432606845                                            
20     W503
```

It's quite a lot of lines to change (50? 150?..) in the current codebase, so I did the above instead. (Fiddling with styling shadows valuable `git` history...)

[PEP8's note](https://www.python.org/dev/peps/pep-0008/#other-recommendations) (use search):

> When combining an argument annotation with a default value, however, do use spaces around the = sign:
> ```
> # Correct:
> def munge(sep: AnyStr = None): ...
> def munge(input: AnyStr, sep: AnyStr = None, limit=1000): ...
> 
> # Wrong:
> def munge(input: AnyStr=None): ...
> def munge(input: AnyStr, limit = 1000): ...
> ```

We don't use spaces there, so `flake8` complains.

### How can it be fixed

Remove the `ignore` section, run `tox -e py3{6,7,8}-lint` and mass-replace `=` with `<space>=<space>` in functions where `flake8` complains.

Elnaril:
Hello @veox, Team,
I'm interested to address this issue if it's fine for you ?

PR #1964, opened 2020-11-19
Issue #1942 remove linting error ignores

Elnaril:
### What was wrong?
PR #1940 added the flake8 error code E252 to be ignored, which was a temporary solution to wait until all these errors were fixed in the codebase.

### How was it fixed?
 - E252 has been removed from the ignore section of flake8 in tox.ini
 - All errors raised by tox were fixed by adding a space around operators where missing

### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ exclude=
 ignore =
     # W503 line break before binary operator
     # It's either this or W504 (line break _after_ binary operator), pick your poison.
-    # W503 gets enabled when E252 gets ignored, see:
+    # W503 gets enabled when any "ignore =" config is added (even the empty one), see:
     # https://github.com/ethereum/py-evm/pull/1940#discussion_r432606845
     W503
 
--- a/tox.ini
+++ b/tox.ini
@@ -10,9 +10,6 @@ envlist=
 max-line-length= 100
 exclude=
 ignore =
-    # E252 missing whitespace around parameter equals
-    # Already used too much in the codebase at the point of introduction.
-    E252,
     # W503 line break before binary operator
     # It's either this or W504 (line break _after_ binary operator), pick your poison.
     # W503 gets enabled when E252 gets ignored, see:
--- a/eth/_utils/datatypes.py
+++ b/eth/_utils/datatypes.py
@@ -72,7 +72,7 @@ class Configurable(ConfigurableAPI):
     """
     @classmethod
     def configure(cls: Type[T],
-                  __name__: str=None,
+                  __name__: str = None,
                   **overrides: Any) -> Type[T]:
 
         if __name__ is None:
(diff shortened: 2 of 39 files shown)


Elnaril:
Hello @pipermerriam , @carver , @cburgdorf , @gsalgado  , I'm not sure who I should ask for a review so I picked up the top 4 contributors ! :sweat_smile: :smile:  Would it be possible to have this PR reviewed ? :pray:

tox.ini
```diff
@@ -10,9 +10,6 @@ envlist=
 max-line-length= 100
 exclude=
 ignore =
-    # E252 missing whitespace around parameter equals
-    # Already used too much in the codebase at the point of introduction.
-    E252,
     # W503 line break before binary operator
     # It's either this or W504 (line break _after_ binary operator), pick your poison.
     # W503 gets enabled when E252 gets ignored, see:
```
carver, line 15: W503 was only here because of E252, so we can actually delete the W503 ignore now, also.

carver (approved):
Yup, looks great, thanks for the contribution!

Would you mind stripping out the W503 also? While you're working on that, it seems that the two commits can be squashed too. But if you prefer, I'm happy to squash-merge it for you after the W503 change.

Elnaril:
@carver  thanks for the review ! :-)

> W503 was only here because of E252, so we can actually delete the W503 ignore now, also.

Actually I tried it but I'm not sure that's possible because when I comment out this flag:
image

flake8 gets grumpy:
image

and if I add the line break after the operator, from:
image

to:
image

I get the W504:
image

carver:
Interesting! I wonder if the commented out lines look to flake8 like an empty list that overrides the default list, which causes the same problem as adding E252. Anyway, no need to hunt that down right now. I'll just squash-merge unless you prefer to squash it yourself.

Elnaril:
Yep, I agree the behaviour seems weird ...

> I'll just squash-merge unless you prefer to squash it yourself.

That's all good to me if you do it, thanks ! :-)

Elnaril:
Actually, you're may be right about the comment (or empty include) confusing flake8:

with that:

image

the linting is ok:
image

but with that:
image

the errors are back ...

Elnaril:
Shall I remove the `include=` and commit ?

carver:
Hm, seems like that confusing error will just come back the next time someone wants to include an exception.

carver:
Maybe change the comment like:
```diff
- W503 gets enabled when E252 gets ignored, see:
+ W503 gets enabled when any "ignore =" config is added (even the empty one), see:
```
and then I'll squash-merge.

Elnaril:
Change is done

PR #1967, opened 2021-01-27
Benchmarks: Mine transactions in a batch to show real import speed

carver:
### What was wrong?

Benchmarks were showing up as really slow (300 kgas/sec for simple value transfer, 3.3 Mgas/sec for ERC20 transfer).

### How was it fixed?

The mining chain was significantly slowed down, by doing a bunch of state work in between each transaction (storing all the state to disk, etc). By batching all the transactions together before sealing the block, we see significant speedups. (7 Mgas/sec for simple value transfer, 5.5 Mgas/sec for ERC20 transfer)

Overall, the benchmark run dropped from 3m1s to 1m43s, and that's with doubling the number of simple value transfers and running 5x as many ERC20 transactions.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/chains/base.py
+++ b/eth/chains/base.py
@@ -44,6 +44,7 @@ from eth.abc import (
     AtomicDatabaseAPI,
     BlockHeaderAPI,
     BlockImportResult,
+    BlockPersistResult,
     ChainAPI,
     ChainDatabaseAPI,
     ConsensusContextAPI,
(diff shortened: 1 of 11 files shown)


eth/chains/base.py
```diff
@@ -667,11 +667,72 @@ def import_block(self,
...
+
+        new_header, receipts, computations = vm.apply_all_transactions(transactions, base_header)
+        filled_block = vm.set_block_transactions(vm.get_block(), new_header, transactions, receipts)
+
+        block_result = vm.mine_block(filled_block, *args, **kwargs)
+        imported_block = block_result.block
+
+        self.validate_block(imported_block)
+
+        # TODO: extract to _import_validated_block (and push to parent class)
```
cburgdorf, line 693: Just double checking if the comment was deliberately left here or if you wanted to address this / file an issue.

cburgdorf (approved):
LGTM. Great to see the benchmarks are getting some new attention :)

eth/chains/base.py
carver: Cool, yeah, resolved.

PR #1979, opened 2021-02-25
Fix uncle timestamp check against its parent

carver:
### What was wrong?

The uncle header may not have the same timestamp as its parent. I ignored the test at first due to https://github.com/ethereum/tests/issues/787 - but it turns out the test is correct.

### How was it fixed?

Update the inequality, and stop ignoring the valid test.
### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


pipermerriam (approved):
huzzah!

Issue #1974, opened 2021-02-24
State access gas cost changes: EIP-2929 for Berlin


carver:
Implement https://eips.ethereum.org/EIPS/eip-2929 for Berlin

PR #1980, opened 2021-03-03
EIP 2930 Transaction access lists

### LINKED ISSUE #1975: State access lists in transactions: EIP-2930 for Berlin
Implement https://eips.ethereum.org/EIPS/eip-2930 for Berlin

Depends on #1973 and #1974

carver:
### What was wrong?

We need to handle transactions and receipts that have access lists: #1975 

But we can't finish ^ until access list is handled during transaction execution. (via #1974 )

### How was it fixed?

Reworked how encoding works across the codebase. Transactions are responsible for encoding themselves now (can't just call `rlp.encode(txn)` because the new typed transactions are not valid RLP). Have to handle different kinds of transactions, like we handle different kinds of receipts. So we have to thread through a new Receipt Encoder, like the Transaction Builder.

Additionally, we have to use the transaction class to build the receipt, because each receipt within a given VM can be different, depending on the transaction.

Also reworked how signatures work, so that every transaction is responsible for returning the `y_parity`, whether it's encoded with the `v_offset` or the `chain_id` or simply a stand-alone field in the new access-list transactions. This makes handling signatures a lot simpler in several cases, without any obvious loss.

There's a little bit of *magic* that I'm not sure if we can/want to get rid of. Depending on the context, the new access list transactions are encoded differently:
- If listed in an rlp list like sent with a block body over devp2p, then it is encoded as an RLP bytestring (which prepends the bytestring length)
- When adding to the trie to calculate the transaction (or receipt) root, then it is encoded as explicitly invalid RLP, with a `\x01` byte as the first byte.

This is done by using `serialize`/`deserialize` to handle the RLP contexts (which further wraps the transaction in a bytestring), and using the new `encode`/`decode` when coding to (intentionally) invalid RLP. This means that the new `TypedTransaction` objects *can* be a field within an `rlp.Serializable` class, like `CountableList(TypedTransaction)`. They will just be forced into the rlp-wrapped bytestring. Everywhere else, we need to use `encode`/`decode` instead.

#### From the commit body

- Adds in the new intrinsic gas cost of access-list entries for a 
transaction.
- Adds two layers of transactions: a TypedTransaction envelope, and the
inner AccessListTransaction which does the bulk of the work. 
- Add a new TypedReceipt to wrap around the legacy Receipt.
- Refactor base transaction classes for reusability without presuming 
RLP serialization.
- Refactor transactions to expose y_parity, now calculated as a 
derivation of legacy v values, or directly available on type-1 
transactions. 
- Expose values that weren't historically on transactions, like:
  - chain_id, which defaults to None
  - access_list, which defaults to an empty list 
- Drop assumed access to transaction.v -- v means too many different
things in different contexts.
- Drop Transaction.from_base_transaction(), preferring
TransactionBuilder.deserialize() instead.
- Add a couple manually-written tests, since the EIP seems to be missing 
them, and ethereum/tests is too hard to test feature by feature.
- Test transactions and receipts as stored in tries used to calculate 
roots used in block headers. 

### To-Do

- [x] Decode receipts
- [x] Rename `v_parity` -> `y_parity` for consistency with EIP (and EC signatures?)
- [x] Tests that include blocks saved with new transactions & receipts
- [x] Rename `ReceiptBuilderAPI` to `ReceiptDecoderAPI`
- [x] Rename `TransactionCodecAPI` to `TransactionDecoderAPI`
- [x] ~Add access lists to computation at beginning~ (Save for follow-up PR)
- [x] Clean up commit history

#### Cute Animal Picture


eth/db/chain.py
```diff
@@ -335,13 +333,13 @@ def _get_block_transaction_hashes(
     @to_tuple
     def get_receipts(self,
                      header: BlockHeaderAPI,
-                     receipt_class: Type[ReceiptAPI]) -> Iterable[ReceiptAPI]:
+                     receipt_builder: Type[ReceiptBuilderAPI]) -> Iterable[ReceiptAPI]:
```
carver, line 336:
```suggestion
                     receipt_builder: Type[ReceiptDecoderAPI]) -> Iterable[ReceiptAPI]:
```

eth/vm/forks/berlin/transactions.py
```diff
@@ -56,24 +357,20 @@ class BerlinTransactionBuilder(TransactionBuilderAPI):
...
+    def decode(cls, encoded: bytes) -> SignedTransactionAPI:
         if len(encoded) == 0:
-            raise DeserializationError(
-                "Encoded transaction was empty, which makes it invalid",
-                encoded,
-            )
+            raise ValidationError("Encoded transaction was empty, which makes it invalid")
 
+        type_id = to_int(encoded[0])
+        if type_id in range(0, 0x80):
```
carver, line 365:
```suggestion
        if type_id in VALID_TRANSACTION_TYPES:
```

carver:
@marcgarreau ready for any last comments

marcgarreau (approved):
Thanks for the walkthrough! I'm comfortable as is.

PR #1999, opened 2021-03-31
Update blake2b-py version requirement

kclowes:
### What was wrong?
Web3py needs blake2b-py>=0.1.4 to support python 3.9. It looks like the only libraries that require py-evm are Trinity and eth-tester, so I'm hopeful that the impacts of updating this will be minimal.

### How was it fixed?
Updated blake2b-py dep to >=0.1.4.

### To-Do

- [x] Clean up commit history

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,6 @@
 version: 2.0
 
-# heavily inspired by: 
+# heavily inspired by:
 # https://raw.githubusercontent.com/pinax/pinax-wiki/6bd2a99ab6f702e300d708532a6d1d9aa638b9f8/.circleci/config.yml
 
 common: &common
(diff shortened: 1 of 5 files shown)


carver:
> It looks like the only libraries that require py-evm are Trinity and eth-tester, so I'm hopeful that the impacts of updating this will be minimal.

Looks like a bit more than that, actually! From https://github.com/ethereum/py-evm/network/dependents?dependent_type=PACKAGE

(Still not a problem. Normally I would say this is a library, so don't tighten the constraints on the dependencies when any downstream project could do it. But presumably we want to update py-evm to support and py3.9 too, so :+1: )

Would you mind adding 39 to:
https://github.com/ethereum/py-evm/blob/7afe050f0849ed87a4282038d394212f6029d3b9/tox.ini#L3
https://github.com/ethereum/py-evm/blob/7afe050f0849ed87a4282038d394212f6029d3b9/tox.ini#L6
https://github.com/ethereum/py-evm/blob/7afe050f0849ed87a4282038d394212f6029d3b9/tox.ini#L47-L50
and CircleCI to show that 3.9 support is done for py-evm?

kclowes:
> Looks like a bit more than that, actually!

Ah, you're right. I meant libraries that we maintain. But the original statement is also wrong as I see ddht in that list too (and maybe there's another I missed too 😆  ) .

> Normally I would say this is a library, so don't tighten the constraints on the dependencies when any downstream project could do it.

I may be misunderstaning this comment, but web3 doesn't directly depend on `blake2b-py`. It depends on eth-tester with a py-evm backend. Would you normally recommend adding a `blake2b-py` line in web3's `setup.py`? Or is there a better way to make sure the dependency's dependency is the right version? 

> Would you mind adding 39

yes, will do!

carver:
> I may be misunderstaning this comment, but web3 doesn't directly depend on `blake2b-py`. It depends on eth-tester with a py-evm backend. Would you normally recommend adding a `blake2b-py` line in web3's `setup.py`? Or is there a better way to make sure the dependency's dependency is the right version?

Nope, I agree with everything you're saying here. :+1: I think the right abstraction layer here is: web3.py doesn't care what version `blake2b-py` is, it just wants a version of py-evm that supports py-3.9. So yeah that py39 test seems like the most important part of the PR (and py3.9 support belongs in a feature newsfragment!).

kclowes:
Ready for another look @carver!

carver (approved):
Good to go after adding a release note that it can handle py3.9! I can't remember, do you have merge access?

kclowes:
I do! Will add a release note and merge!

PR #2000, opened 2021-04-02
Remove extra newsfragment

kclowes:
### What was wrong?
Was moving too fast and accidentally added two newsfragments for the python 3.9 PR 🤦 

### How was it fixed?
Removed the duplicate.


carver (changes requested):
Actually, it's okay to have separate files! I've done that before when a PR has both a feature and a bugfix. I think it's fine to just close the PR as-is.

kclowes:
Sounds good!

PR #1799, opened 2019-07-18
Py evm 1466

dherykw:
### What was wrong?

EIP170 states that the contract size limit was changed to 2**14 + 2**13 which is 24,576 bytes. The following line implementations the constant for EIP170, but it is off by one.

Issue Reference: #1466

### How was it fixed?
Contract limit size was changed by @glaksmono, I have improved the tests 
- Update the contract limit size to follow EIP170
- Add tests against the contract size limit

### To-Do

[//]: # (Stay ahead of things, add list items here!)
- [ ] Clean up commit history

[//]: # (For important changes that should go into the release notes please add a newsfragment file as explained here: https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md)

[//]: # (See: https://py-evm.readthedocs.io/en/latest/contributing.html#pull-requests)
- [ ] Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/master/newsfragments/README.md)

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #2002, opened 2021-04-18
py-evm-issue-1466: Added test cases for EIP170

junning-tong:
### What was wrong?
Check [issue-1466](https://github.com/ethereum/py-evm/pull/1799#issue-299070325) for reference

> Double checking this is the intended functionality.
> 
> EIP170 states that the contract size limit was changed to `2**14 + 2**13` which is 24,576 bytes. The following line implementations the constant for EIP170, but it is off by one.
> 
> https://github.com/ethereum/py-evm/blob/0ff7bc0f820250d7f15e804bc09870776ea34eb5/eth/vm/forks/spurious_dragon/constants.py#L13
> 

### How was it fixed?
* Change  `EIP170_CODE_SIZE_LIMIT` from `24577` to `24576`
* Add test cases

#### Cute Animal Picture

:pig2::pig2::pig2::pig2::pig2::pig2::pig2::pig2:


pipermerriam:
Before we merge this, I'd like to understand why this wasn't caught using consensus tests because this seems like the type of thing that 1) I would be surprised if it wasn't in the tests and 2) if it isn't in the tests it should be added upstream

junning-tong:
@pipermerriam the newly added test file is in directory `tests/core/vm/`.
Should I use it to `tests/core/consensus/`?

pipermerriam:
No, I'm referring to these tests: https://github.com/ethereum/tests/

They exist as a submodule in the repository.

junning-tong:
Hi @pipermerriam I see a commit which added test cases in fixtures submodules. https://github.com/ethereum/py-evm/pull/1998/files

Should I move newly added .json to the submodule as well?
tests/core/vm/fixtures/spurious_dragon_computation_test_code_size_limitation.json

PR #2005, opened 2021-05-03
[WIP] add support for EIP-1559

ftruzzi:
Hi!

I tried to add new block headers and transactions, and a bit of header validation. Please take a look at those -- do they make sense? Should those changes to blocks/transactions be applied somewhere else? Should I keep going?

Thanks :)


pipermerriam:
I think yes, you can keep going.  I only gave this a shallow review but the direction and structure looks good.  I didn't dig into the actual method implementations yet.

My main recommendation would be to extract the "stub" for London from this PR to reduce the size of the diff.  See https://github.com/ethereum/py-evm/pull/1969 for how this was done for Berlin.

eth/abc.py
```diff
@@ -74,6 +74,8 @@ class MiningHeaderAPI(ABC):
     gas_used: int
     timestamp: int
     extra_data: bytes
+    gas_target: int  # EIP-1559
+    base_fee_per_gas: int  # EIP-1559
```
pipermerriam, line 78: This feels... unfortunate, but it might be necessary.  Ideally, it would be nice if these fields didn't leak into all the previous VM APIs.... but doing so might introduce even more unfortunate complexity into our type definitions which might be worse...

ftruzzi: I agree and will take a deeper look -- for now I can say a similar problem applies to transactions. Base `transaction.validate` methods all type-check `gas_price` which is no longer a thing in London transactions so they need to be duplicated almost entirely. We could keep `gas_price` as an (internal, not exposed) alias for either `max_fee_per_gas` or `max_priority_fee_per_gas` so that we could extend and call basic validation methods via `super()`, but it sounds confusing :(

eth/vm/forks/london/transactions.py
```diff
@@ -0,0 +1,352 @@
...
+            access_list,
+            y_parity,
+            r,
+            s,
+        )
+        return LondonTypedTransaction(BASE_GAS_FEE_TRANSACTION_TYPE, transaction)
+
+def normalize_transaction(
+    transaction: SignedTransactionAPI
+) -> SignedTransactionAPI:
```
ftruzzi, line 339: This function modifies the transaction object rather than creating an instance of a possible new `LondonTypedTransaction` class so that both new (London-specific) and existing validation code can work on all transactions. However this means that `max_priority_fee_per_gas` and `max_fee_per_gas` cannot be read-only properties, and I'm wondering how should we handle users trying to access the new properties in pre-London transactions?

eth/db/header.py
```diff
@@ -619,4 +625,8 @@ def _add_block_number_to_hash_lookup(db: DatabaseAPI, header: BlockHeaderAPI) ->
 # be looking up recent blocks.
 @functools.lru_cache(128)
 def _decode_block_header(header_rlp: bytes) -> BlockHeaderAPI:
-    return rlp.decode(header_rlp, sedes=BlockHeader)
+    try:
+        return rlp.decode(header_rlp, sedes=BlockHeader)
+    except ObjectDeserializationError:
+        # could be a new >=London block header
+        return rlp.decode(header_rlp, sedes=LondonBlockHeader)
```
ftruzzi, line 632: Does this look good, or should we find some way to check the header type first?

eth/vm/forks/london/transactions.py
```diff
@@ -0,0 +1,324 @@
...
+        ('access_list', CountableList(AccountAccesses)),
+        ('y_parity', big_endian_int),
+        ('r', big_endian_int),
+        ('s', big_endian_int),
+    ]
+
+    @property
+    def gas_price(self) -> None:
+        # maybe add a warning, or raise an exception instead?
+        return None
```
ftruzzi, line 158: How should we handle this? `gas_price` is no longer a transaction attribute. It could still be interpreted as "effective gas price", but that requires knowing the base gas fee of the block so it would not a be "transaction-only" attribute anymore

ftruzzi:
@pipermerriam would you take another look? Sorry it took me a while. Two main things I'd like to get some feedback on:
- New transaction and block attributes: do they look good? How should we handle London-only attributes in the API? Users might be trying to access them in legacy transactions/VMs and vice-versa.
- The way transaction execution works is by passing the effective gas price to the transaction context creator, and by adding the base gas fee to the execution context. After some trial and error it was the best way I found to keep as much existing code as possible, but I'm wondering if you have any other ideas.

Things that still have to be done are [here](https://github.com/ethereum-cdap/cohort-zero/blob/main/notes/francesco.md#gathered-information-1) + linting and style -- please let me know if this is too large/hard to read and you'd like me to better separate commits before taking a look. Thanks

eth/abc.py
carver: After dealing with typed transactions, my general take on it was: the *latest* VM should have the cleanest implementation. If a kludge is called for, it should be on the old VMs (like v vs y-parity in the signature, where we started to prefer y-parity as the default/native field). At first glance, this change seems to follow that approach, so I'm 👌🏻 to keep it. I'll comment if I stumble on a better option while reading the rest.

eth/abc.py
```diff
@@ -2865,9 +2877,8 @@ def get_computation(self,
     #
     # Transaction context
     #
-    @classmethod
     @abstractmethod
-    def get_transaction_context_class(cls) -> Type[TransactionContextAPI]:
+    def get_transaction_context_class(self) -> Type[TransactionContextAPI]:
```
carver, line 2881: Looks like this one could stay a `classmethod`

eth/vm/forks/london/state.py
```diff
@@ -0,0 +1,193 @@
...
+        if isinstance(transaction, LondonTypedTransaction):  # TODO probably do this somewhere else
+            priority_fee_per_gas = min(
+                transaction.max_priority_fee_per_gas,
+                transaction.max_fee_per_gas - self.execution_context.base_gas_fee
+            )
+        else:
+            priority_fee_per_gas = min(
+                transaction.gas_price,
+                transaction.gas_price - self.execution_context.base_gas_fee
+            )
```
carver, line 187: Yeah, my instinct is to put this in the transaction class, just have everything implement a `priority_fee_per_gas` attribute. Then we get to skip the instance check.

eth/abc.py
```diff
@@ -2917,9 +2928,8 @@ def validate_transaction(self, transaction: SignedTransactionAPI) -> None:
         """
         ...
 
-    @classmethod
     @abstractmethod
-    def get_transaction_context(cls,
+    def get_transaction_context(self,
```
carver, line 2932: Yeah, bummer that this can't be a class method anymore. I can *maybe* imagine doing something where the transaction  context no longer gives access to the active gas price without explicit access to the execution-context (like as a parameter to a new `get_gas_price(execution_context)` method). Or something along those lines. I am okay with this solution though. 👍🏻

carver (commented):
I'm just getting started looking around, will pick it up again tomorrow.

carver:
Thanks for getting this started! The commits from this PR are integrated to #2013

Issue #1959, opened 2020-10-19
Replace or improve blake2b-py dependency


MatthiasLohr:
* py-evm Version: 0.3.0a19
* OS: linux
* Python Version: Python 3.9.0rc1

### What is wrong?

py-evm depends on [blake2b-py](https://github.com/davesque/blake2b-py), which looks quite unmaintained. The PyPI package for blake2b-py comes with some prebuilt binaries, however for e.g. Python 3.9 or support on ARM architectures (Raspberry Pi) the only chance is to have it compiled. This requires to install the whole Rust toolchain to get it compiled (including: CI/CD stuff, there we would theoretically have to compile it for each commit!).

### How it could be fixed?

For this reason I want to open the discussion, if there might be the chance to rely on another library for the required functionalities. Currently, some people including me are trying to reach the maintainer to get access to update the project, but for now there is no reaction from the maintainer.

I'm looking forward to your opinions.

fubuloubu:
Blake2 was recently added to `hashlib` (3.6+), and should be considered for updating this library moving forwards.

fubuloubu:
After doing some digging, I noticed that this was indeed the way this was attempted before (see https://github.com/ethereum/py-evm/issues/1818), but it proved too slow (see https://github.com/ethereum/py-evm/issues/1836) to use in practice for the Trinity client. In either case, my suggestion to use the `hashlib` implementation is indeed not viable.

However, since the [Trinity client](https://github.com/ethereum/trinity) became obsolesced rather recently (reducing the necessity for sync-viable speed of this algorithm), and the numerous reported issues users have had setting up the Rust library on different platforms (see https://github.com/ethereum/blake2b-py/issues), as well as the problem that this issue notes about lack of support for blake2b-py... perhaps it would be time to reconsider this decision and move `py-evm` back to the slower native python implementation in order to reduce dependency and maintenance issues.

MatthiasLohr:
What about making blake2b-py an optional dependency? So, if you install py-evm via `pip install py-evm` it will use the (slow, but struggle-free) hashlib implementation, if you use `pip install py-evm[blake2b-py]` it will additionally install the (faster, but maybe annoying setup required) rust-based variant. And the code has then to detect if blake2b-py is installed, if yes it will use it, if no it will fall back to hashlib.

Any thoughts on that?

cburgdorf:
>What about making blake2b-py an optional dependency

Yes, I think that's the way to go. We've used that pattern before and it works well.

[...]

Issue #1941, opened 2020-05-29
Windows support?


carver:
### What is wrong?

Windows is not supported. Even if we never get trinity supported, just being able to run evm stuff on Windows would be nice.

### How can it be fixed

Known issues:
- #395 -- need a pyethash fallback (maybe native python?)
- #781 -- fall back to default loop because can't use `uvloop`? Does trio support windows?

Let's add to this list as we discover more.

Then hunt down anything else. Probably need a Windows machine to develop this at any sort of reasonable speed.

carver:
2012 had a "may fix". I think it's highly unlikely that this is resolved. Closing this issue would definitely require confirmation.

cburgdorf:
@carver yep, I guess this was closed automatically from a commit message :+1:

0xClandestine:
I was able to install on windows using wsl [https://docs.microsoft.com/en-us/windows/wsl/install](wsl)

antazoey:
windows seems to be working now

PR #2012, opened 2021-07-21
fix: make blake2b optional

fubuloubu:
### What was wrong?
fixes: #1959
may fix: #1941

### How was it fixed?
Used the native implementation of blake2b compression function if `blake2b-py` is not installed.
Made `blake2b-py` an optional dependency via `eth-extra`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

cute elephant

### DIFF
--- a/eth/precompiles/blake2.py
+++ b/eth/precompiles/blake2.py
@@ -12,7 +12,7 @@ from eth.vm.computation import (
 
 try:
     from blake2b import compress as blake2b_compress
-except ModuleNotFoundError:
+except ImportError:
     from eth._utils.blake2.compression import blake2b_compress
 
 GAS_COST_PER_ROUND = 1
--- a/eth/precompiles/blake2.py
+++ b/eth/precompiles/blake2.py
@@ -1,4 +1,3 @@
-import blake2b
 from eth_utils import (
     ValidationError,
 )
@@ -10,6 +9,11 @@ from eth.exceptions import (
 from eth.vm.computation import (
     BaseComputation,
 )
+
+try:
+    from blake2b import compress as blake2b_compress
+except ModuleNotFoundError:
+    from eth._utils.blake2.compression import blake2b_compress
 
 GAS_COST_PER_ROUND = 1
 
@@ -25,5 +29,5 @@ def blake2b_fcompress(computation: BaseComputation) -> BaseComputation:
 
     computation.consume_gas(gas_cost, reason=f"Blake2b Compress Precompile w/ {num_rounds} rounds")
 
-    computation.output = blake2b.compress(*parameters)
+    computation.output = blake2b_compress(*parameters)
     return computation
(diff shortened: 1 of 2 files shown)


eth/precompiles/blake2.py
```diff
@@ -11,6 +10,11 @@
     BaseComputation,
 )
 
+try:
+    from blake2b import compress as blake2b_compress
+except ModuleNotFoundError:
```
cburgdorf, line 15: Just noting that I've used `ImportError` for this kind of pattern in the past for which `ModuleNotFoundError` is a child of. The difference is that `ModuleNotFoundError` will only catch if the module can not be found whereas `ImportError` will additionally catch cases where the module might be found but can not be imported nonetheless. But I'm not sure if there's a clear recommendation which one to use for cases like this. Just pointing it out in case you have a preference.

cburgdorf (approved):
LGTM

cburgdorf (approved):
Generally looks good but CI errors need to be addressed. Also just pinging @carver since that might affect Py-EVM and he's closer to that project than I am.

fubuloubu:
> Generally looks good but CI errors need to be addressed.

As far as I can tell, none of the linting errors has to do with files I edited.

> Also just pinging @carver since that might affect Py-EVM and he's closer to that project than I am.

Yes, this definitely might affect downstream users of `py-evm`, like `eth-tester` and `trinity`. `trinity` has been deprecated, but I think for `eth-tester` it should be an optional dependency

cburgdorf:
Sorry for the late response. 

>As far as I can tell, none of the linting errors have to do with files I edited.

Yes, but they still need to be addressed since we can not merge a PR that didn't pass all CI checks.

I'm not working on this project anymore (and I just actually noticed this is a Py-EVM PR not a blake2b PR :see_no_evil: ) but I know that @carver is currently working on EIP-1559 support for Py-EVM so if you don't feel like investing time into fixing these linting issues you may just wait for them to be fixed in master and then rebase at a later point :)

fubuloubu:
> I just actually noticed this is a Py-EVM PR not a blake2b PR

Yes, this seemed like the best solution to this issue as adding Windows support to the Rust-based blake2-py would be much more difficult, and the blake2 precompile is not used nearly enough in production to justify that work.

> @carver is currently working on EIP-1559 support for Py-EVM so if you don't feel like investing time into fixing these linting issues you may just wait for them to be fixed in master and then rebase

I would prefer this sooner than later, I will take a look at the linting issues. This is a blocker for me.

cburgdorf:
Thanks @fubuloubu

MatthiasLohr:
Great! Thanks!

Any idea, when this get released (uploaded to PyPI)?

carver:
The next release will probably happen after London tests pass. You can watch #2021

fubuloubu:
Note: this will not fix #1941 because the issue there is the `pyethash` dependency

PR #2015, opened 2021-08-20
London basefee opcode 0x48

kclowes:
### What was wrong?

The BASEFEE opcode for London hadn't been implemented yet. 

### How was it fixed?

Added it! 

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

image

### DIFF
--- a/eth/abc.py
+++ b/eth/abc.py
@@ -2672,9 +2672,9 @@ class StateAPI(ConfigurableAPI):
 
     @property
     @abstractmethod
-    def basefee(self) -> int:
+    def base_fee(self) -> int:
         """
-        Return the current ``basefee`` from the current :attr:`~execution_context`
+        Return the current ``base_fee`` from the current :attr:`~execution_context`
         """
         ...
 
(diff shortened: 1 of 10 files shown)


eth/abc.py
```diff
@@ -2670,6 +2670,14 @@ def gas_limit(self) -> int:
         """
         ...
 
+    @property
+    @abstractmethod
+    def base_fee(self) -> int:
+        """
+        Return the current ``base_fee`` from the current :attr:`~execution_context`
```
carver, line 2677: Maybe note that this raises an exception, with old VMs.

tests/core/opcodes/test_opcodes.py
```diff
@@ -160,6 +160,15 @@ def test_add(vm_class, val1, val2, expected):
     assert result == expected
 
 
+def test_base_fee():
+    computation = run_general_computation(LondonVM)
+    computation.opcodes[opcode_values.BASEFEE](computation)
+
+    result = computation.stack_pop1_int()
```
carver, line 167:
```suggestion
    result = computation.stack_pop1_any()
```

^ is a way to make sure that an int was pushed to the stack natively. (because it won't try to coerce the type on the way out)

carver (approved):
LGTM!

PR #2018, opened 2021-08-25
EIP-3554 and EIP-3541

fselmo:
### What was wrong?

- EIP-3554 and EIP-3541 had not yet been implemented 

### How was it fixed?

- EIP-3554 and EIP-3541 were implemented and some relevant tests were added

### To-Do

- [X] Clean up commit history

#### Cute Animal Picture

IMG_2166


fselmo:
@carver @kclowes I can't seem to add reviewers on py-evm but this guy should be ready to look at. I'd be happy to jump on a call Monday to go over things too.

eth/vm/forks/london/computation.py
```diff
@@ -11,3 +18,62 @@ class LondonComputation(BerlinComputation):
...
+                    computation.error = err
+                    state.revert(snapshot)
+                else:
+                    if contract_code[:1] == EIP3541_RESERVED_STARTING_BYTE:
+                        # As per EIP-3541, gas is still consumed on a revert of this nature
+                        state.revert(snapshot)
+                        raise ReservedBytesInCode(
+                            "Contract code begins with EIP3541 reserved byte '0xEF'."
+                        )
+                    else:
```
carver, line 66: Looks like this is the full change? It's a lot of work to comb through to confirm that. I would love to see that get easier. Here's another option (I don't love it, but it gets at the idea of isolating the change in the new London code):
1. in the Spurious Dragon computation, add a `def validate_new_code(cls, contract_code: bytes): pass`
2. Run `cls.validate_new_code(contract_code)` just before the `SETTING CODE` log. (validate_X in py-evm is generally understood to mean: raise on failure)
2. Add a `def validate_new_code` in London that checks the `0xEF` byte.

I'm imagining the validate call to look like:
```py
except OutOfGas as err:
  ...
else:
  try:
    cls.validate_new_code(contract_code)
  except VMError:
    state.revert(snapshot)
    raise

  if cls.logger:
    cls.logger.debug2("SETTING CODE: " ...
```
(The thing I don't like about it is that it continues to embrace the inheritance style, but it's also pervasive in py-evm already. So it may be our best option)

carver (approved):
I'd love to see a way to reduce the duplicated code. I gave an option, but if you figure something else out, I'm all ears!

Everything else LGTM.

eth/vm/forks/london/computation.py
fselmo: @carver That was indeed the full change. This is definitely better. My goal was  to refactor this at some point but thought I'd track it as a separate issue and just get these changes in. This seems quick enough to implement in this PR though and will make it all much easier to read too 👍

carver: Great :)

PR #2021, opened 2021-08-31
Run London fork tests

### LINKED ISSUE #2017: Pass London tests in ethereum/tests
- [ ] Add London test env to tox.ini 
- [x] Remove hack to skip London tests in `tests/json-fixtures/blockchain/test_blockchain.py`
- [ ] Add London CI job
- [ ] Fix any broken tests

carver:
### What was wrong?

Fix #2017 

### How was it fixed?

- [x] Add circle job
- [x] Merge/Rebase on #2022
- [x] Merge/Rebase on #2020 
- [x] Merge/Rebase on #2029 
- [x] Fix all broken tests
  - [x] Crash-fix: configure base fee on genesis header
  - [x] Bugfix: reset return data on contract collision CREATE / CREATE2
  - [x] Bugfix: properly validate 1/1024th gas limit deviation from parent as an exclusive limit
  - [x] Bugfix: validate uncles gas limit properly and account for the Berlin -> London transition (gas limit doubles) when validating

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/guides/understanding_the_mining_process.rst
+++ b/docs/guides/understanding_the_mining_process.rst
@@ -380,4 +380,4 @@ zero value transfer transaction.
   ... )
 
   >>> chain.mine_block(mix_hash=mix_hash, nonce=nonce)
-  <ByzantiumBlock(#Block #1-0x41f6..2913)>
+  <ByzantiumBlock(#Block #1-0xe372..385c)>
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -668,7 +668,13 @@ class VM(Configurable, VirtualMachineAPI):
                 f"Uncle's gas usage ({uncle.gas_used}) is above "
                 f"the limit ({uncle.gas_limit})"
             )
-        validate_gas_limit(uncle.gas_limit, uncle_parent.gas_limit)
+
+        uncle_parent_gas_limit = uncle_parent.gas_limit
+        if not hasattr(uncle_parent, 'base_fee_per_gas') and hasattr(uncle, 'base_fee_per_gas'):
+            # if Berlin -> London transition, double the parent limit for validation
+            uncle_parent_gas_limit *= 2
+
+        validate_gas_limit(uncle.gas_limit, uncle_parent_gas_limit)
 
     #
     # State
--- a/eth/vm/base.py
+++ b/eth/vm/base.py
@@ -668,6 +668,7 @@ class VM(Configurable, VirtualMachineAPI):
                 f"Uncle's gas usage ({uncle.gas_used}) is above "
                 f"the limit ({uncle.gas_limit})"
             )
+        validate_gas_limit(uncle.gas_limit, uncle_parent.gas_limit)
 
     #
     # State
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,6 +82,12 @@ jobs:
       - image: circleci/python:3.6
         environment:
           TOXENV: py36-native-blockchain-istanbul
+  py36-native-blockchain-london:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: py36-native-blockchain-london
   py36-native-blockchain-petersburg:
     <<: *common
     docker:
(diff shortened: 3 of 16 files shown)


carver:
Okay, I think it makes sense to wait until #2020 is merged before trying to debug more.

carver:
Just a few errors left. Some notes along the way:
- It looks like uncles may not be getting validated correctly (I'm not convinced this is a London issue. I think we may have been incorrectly passing the test for earlier forks by raising a different validation error than expected).
- Looks like several 1559 gas limit issues. The test name isn't quite enough to go on. Might be worth just asking in the testing channel for what the purposes of the failing tests are.

fselmo:
> * It looks like uncles may not be getting validated correctly (I'm not convinced this is a London issue. I think we may have been incorrectly passing the test for earlier forks by raising a different validation error than expected).

Looks like you were right on the money with this one 👌... uncle gas limit wasn't being validated all the way (not London related) but also gas limit increase / decrease factor (1/1024) is an exclusive limit whereas it had been inclusive this whole time. The last few commits here should highlight these changes and all the current tests now pass.

--- 

I did remove the silenced `London` tests from some previous commits here in a separate PR to let [circleci build](https://app.circleci.com/pipelines/github/ethereum/py-evm/2006/workflows/6fa4fee2-702a-40bc-8606-6c73cc2290ae/jobs/154471) it for me. Out of all `48` tests skipped, `9` fail. 

However...

- `6` of those (all tests in this [commit](https://github.com/ethereum/py-evm/pull/2021/commits/77ca7fe31c378730adf7b4bc8636208eb0276f3f)) have been previously skipped for other forks due to having a "synthetic state"? I wasn't around for that conversation, not entirely sure why we skip these other than we may not have a proper setup to accurately test them? It seems to me like we can keep this commit.
    - `GeneralStateTests/stRevertTest/RevertInCreateInInit` (1 test)
    - `GeneralStateTests/stCreate2/RevertInCreateInInitCreate2` (1 test)
    - `GeneralStateTests/stSStoreTest/InitCollision` (all variants - 4 tests)
    
- `3` were vm performance tests that timed out:
    - `VMTests/vmPerformance/loopMul` (all variants - 3 tests)
    
@carver thoughts? Seems to me like we are good to go if we decide that the above assumptions are good: that synthetic state tests and vm performance timeout tests can be skipped for now.

eth/vm/base.py
```diff
@@ -668,7 +668,13 @@ def validate_uncle(cls,
...
                 f"the limit ({uncle.gas_limit})"
             )
-        validate_gas_limit(uncle.gas_limit, uncle_parent.gas_limit)
+
+        uncle_parent_gas_limit = uncle_parent.gas_limit
+        if not hasattr(uncle_parent, 'base_fee_per_gas') and hasattr(uncle, 'base_fee_per_gas'):
+            # if Berlin -> London transition, double the parent limit for validation
+            uncle_parent_gas_limit *= 2
+
+        validate_gas_limit(uncle.gas_limit, uncle_parent_gas_limit)
```
carver, line 677: Oof, this is a bummer, but I don't have a better suggestion right now.

docs/guides/understanding_the_mining_process.rst
```diff
@@ -380,4 +380,4 @@ zero value transfer transaction.
   ... )
 
   >>> chain.mine_block(mix_hash=mix_hash, nonce=nonce)
-  <ByzantiumBlock(#Block #1-0x41f6..2913)>
```
carver, line 383: Why did this change? Something about the gas limit being off by one?

carver (commented):
Cool, looks good to me! I can't approve "my own" PR X)

Agree that it's okay to leave those other tests as skipped or `xfail` (as appropriate) for now.

I'll do a bit of tidying work and get it merged!

docs/guides/understanding_the_mining_process.rst
fselmo: @carver yep

eth/vm/base.py
fselmo: Hah, yes, same... I think it can use a refactor at some point for sure.

carver:
I snuck in a little renaming of the objects in the logs to help identify which block is built locally and which is proposed by the test. I also skipped a few more slow tests that were causing the London job to be significantly slower than the others.

PR #2023, opened 2021-09-02
Erase return data on appropriate exceptions

fselmo:
### What was wrong?

1. We did not reset return data on CREATE or CREATE2 calls for an insufficient funds case.
2. We also never reset return data on exceptions that have `erases_return_data` flag set to `True`. Some tests that were in place to test this did not test it appropriately.

### How was it fixed?

1. Erase return data and error log when insufficient funds or stack depth limit reached during CREATE or CREATE2.
2. Check the `erases_return_data` flag on the raised exception and erase the return data in the same way that we burn gas when the `burns_gas` flag is set to `True`.

### To-Do

- [X] Clean up commit history

#### Cute Animal Picture

Image from iOS (2)

### DIFF
--- a/eth/vm/logic/system.py
+++ b/eth/vm/logic/system.py
@@ -7,7 +7,9 @@ from eth_utils import (
 from eth import constants
 from eth.exceptions import (
     Halt,
+    InsufficientFunds,
     Revert,
+    StackDepthLimit,
     WriteProtection,
 )
 
@@ -154,14 +156,19 @@ class Create(Opcode):
 
         computation.extend_memory(stack_data.memory_start, stack_data.memory_length)
 
-        insufficient_funds = computation.state.get_balance(
-            computation.msg.storage_address
-        ) < stack_data.endowment
+        storage_address_balance = computation.state.get_balance(computation.msg.storage_address)
+
+        insufficient_funds = storage_address_balance < stack_data.endowment
         stack_too_deep = computation.msg.depth + 1 > constants.STACK_DEPTH_LIMIT
 
         if insufficient_funds or stack_too_deep:
             computation.stack_push_int(0)
-            return
+            if insufficient_funds:
+                raise InsufficientFunds(
+                    f"Insufficient funds: {storage_address_balance} < {stack_data.endowment}"
+                )
+            elif stack_too_deep:
+                raise StackDepthLimit("Stack depth limit reached")
 
         call_data = computation.memory_read_bytes(
             stack_data.memory_start, stack_data.memory_length
--- a/eth/vm/logic/system.py
+++ b/eth/vm/logic/system.py
@@ -7,9 +7,7 @@ from eth_utils import (
 from eth import constants
 from eth.exceptions import (
     Halt,
-    InsufficientFunds,
     Revert,
-    StackDepthLimit,
     WriteProtection,
 )
 
@@ -163,12 +161,13 @@ class Create(Opcode):
 
         if insufficient_funds or stack_too_deep:
             computation.stack_push_int(0)
+            computation.return_data = b''
             if insufficient_funds:
-                raise InsufficientFunds(
-                    f"Insufficient funds: {storage_address_balance} < {stack_data.endowment}"
-                )
+                err_msg = f"Insufficient Funds: {storage_address_balance} < {stack_data.endowment}"
             elif stack_too_deep:
-                raise StackDepthLimit("Stack depth limit reached")
+                err_msg = "Stack limit reached"
+            self.logger.debug2("%s failure: %s", self.mnemonic, err_msg,)
+            return
 
         call_data = computation.memory_read_bytes(
             stack_data.memory_start, stack_data.memory_length
--- a/tests/core/vm/test_computation.py
+++ b/tests/core/vm/test_computation.py
@@ -10,7 +10,6 @@ from eth.chains.mainnet import MAINNET_VMS
 from eth import constants
 from eth.consensus import NoProofConsensus
 from eth.exceptions import (
-    InsufficientFunds,
     InvalidInstruction,
 )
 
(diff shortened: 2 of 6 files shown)


eth/vm/logic/system.py
```diff
@@ -154,14 +155,16 @@ def __call__(self, computation: ComputationAPI) -> None:
...
+
+        insufficient_funds = storage_address_balance < stack_data.endowment
         stack_too_deep = computation.msg.depth + 1 > constants.STACK_DEPTH_LIMIT
 
         if insufficient_funds or stack_too_deep:
             computation.stack_push_int(0)
-            return
+            raise InsufficientFunds(
+                f"Insufficient funds: {storage_address_balance} < {stack_data.endowment}"
+            )
```
carver, line 167: `InsufficientFunds` error is confusing, if the stack was too deep.

eth/vm/logic/system.py
```diff
@@ -154,14 +155,16 @@ def __call__(self, computation: ComputationAPI) -> None:
 
         computation.extend_memory(stack_data.memory_start, stack_data.memory_length)
 
-        insufficient_funds = computation.state.get_balance(
-            computation.msg.storage_address
-        ) < stack_data.endowment
+        storage_address_balance = computation.state.get_balance(computation.msg.storage_address)
+
+        insufficient_funds = storage_address_balance < stack_data.endowment
```
carver, line 160: 👍🏻 👍🏻

tests/core/vm/test_london.py
```diff
@@ -288,19 +292,16 @@ def test_state_does_not_revert_on_reserved_0xEF_byte_for_CREATE_and_CREATE2_pre_
 ):
     chain = pre_london_miner
     vm = chain.get_vm()
-    state = vm.state
 
-    create_message = Message(
+    computation = vm.execute_bytecode(
```
carver, line 296: Was this change important, beyond being shorter/cleaner?

carver (commented):
Looks like it's on a good path. I didn't look at the test failures. Let me know if you get stuck, and I can help out.

tests/core/vm/test_london.py
fselmo: This seemed like a more appropriate way to test the bytecode tests provided by the EIP. It will still end up calling `apply_create_message()`. Having written the original tests I just wanted to make sure it is tested from multiple angles I suppose and I was trying to find out what would be the most appropriate instantiation (if there is one?). But it also feels a bit cleaner / shorter hah.

I don't lean any particular way here... thoughts?

carver: Yup, 👍🏻 to keep the change. Sometimes there are little subtleties about when you can use one API or the other, and I was curious if I was missing something on a casual read-through. (for example: that a test started failing, because some context was missing when launched via `apply_create_message()`) But it sounds like I expected: just a cleaner way to do it.

eth/vm/logic/system.py
```diff
@@ -148,19 +148,25 @@ def get_stack_data(self, computation: ComputationAPI) -> CreateOpcodeStackData:
     def __call__(self, computation: ComputationAPI) -> None:
 
         stack_data = self.get_stack_data(computation)
-
         gas_cost = self.get_gas_cost(stack_data)
+
```
carver, line 152: nit: remove unnecessary churn?

fselmo:
@carver since you were taking a look at this, it should now be in a state to be reviewed. There are some things that I wanted to do because they _felt_ like the right thing but, with the way the code is currently written, it didn't end up being ideal.

In particular, it seems that raising an exception where it seemed appropriate (`InsufficientFunds(VMError)`) is not the way to go as it will end up consuming more gas than it should (`burns_gas = True`). Instead, copying from the original `__call__` code, manually resetting the `computation.return_data` _seems_ to be the best way for now and it unbreaks the ethereum tests that were broken upon raising the exception.

tests/core/vm/test_computation.py
```diff
@@ -0,0 +1,92 @@
...
+    if computation.is_error:
+        assert isinstance(computation.error, InvalidInstruction)
+        # only test CREATE case for byzantium as the CREATE2 opcode (0xf5) was not yet introduced
+        assert vm.fork == "byzantium"
+        assert "0xf5" in repr(computation.error).lower()
+
+    else:
+        # Assert only uses the appropriate amount of gas.
+        # This varies between the CREATE (34925) and CREATE2 (36725) tests.
+        assert computation.get_gas_used() == 34925 or computation.get_gas_used() == 36725
```
carver, line 91: Maybe you could pass in the expected gas usage as a parametrized value?

I assume this test is to make sure that we don't burn the remaining gas. That could be added as a comment (since it may not be obvious why the gas used is related to the return data value)

carver (approved):

carver:
> In particular, it seems that raising an exception where it seemed appropriate (`InsufficientFunds(VMError)`) is not the way to go as it will end up consuming more gas than it should (`burns_gas = True`). Instead, copying from the original `__call__` code, manually resetting the `computation.return_data` _seems_ to be the best way for now and it unbreaks the ethereum tests that were broken upon raising the exception.

If you feel like exploring it, you could maybe override `burns_gas = True` for that exception. But I'm also completely fine with this solution. 👍🏻

tests/core/vm/test_computation.py
fselmo: For sure. I think that would be better.

fselmo: It actually turns out the discrepancy was due to the gas estimation changes in Berlin. We don't particularly care to test the exact gas value here since that's not what the scope of this test is so I am just going to assert that we do not burn all the gas and that it falls within an expected range here.

fselmo:
@carver I'm good to merge it if you are. I don't have that permission for this repo.

eth/vm/logic/system.py
```diff
@@ -154,13 +154,19 @@ def __call__(self, computation: ComputationAPI) -> None:
...
         stack_too_deep = computation.msg.depth + 1 > constants.STACK_DEPTH_LIMIT
 
         if insufficient_funds or stack_too_deep:
             computation.stack_push_int(0)
+            computation.return_data = b''
+            if insufficient_funds:
+                err_msg = f"Insufficient Funds: {storage_address_balance} < {stack_data.endowment}"
+            elif stack_too_deep:
+                err_msg = "Stack limit reached"
+            self.logger.debug2("%s failure: %s", self.mnemonic, err_msg,)
```
carver, line 169: Almost always, we include an `else` block with a `RuntimeError("Cannot determine log message because of unexpected failure")` or something like that. It's okay to get to it in the next PR.

See style guide: https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/style-guide.md#only-use-bare-else-block-for-validating-unexpected-values

PR #2024, opened 2021-09-08
[WIP] Clear computation output on ecrecover if data length is the wrong size

kclowes:
### To-Do

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/precompiles/ecrecover.py
+++ b/eth/precompiles/ecrecover.py
@@ -15,6 +15,9 @@ from eth._utils.padding import (
     pad32r,
 )
 
+from eth.exceptions import (
+    InvalidDataLength,
+)
 from eth.validation import (
     validate_lt_secpk1n,
     validate_gte,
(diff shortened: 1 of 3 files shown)


eth/precompiles/ecrecover.py
```diff
@@ -25,9 +28,20 @@
...
+
 def ecrecover(computation: BaseComputation) -> BaseComputation:
     computation.consume_gas(constants.GAS_ECRECOVER, reason="ECRecover Precompile")
     data = computation.msg.data_as_bytes
+
+    try:
+        validate_data_length(data, 128)
+    except InvalidDataLength:
+        return computation
+
```
holiman, line 44: This is most definitely introducing a consensus issue. See https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go#L167

If the input is smaller than 128, it must be padded with zeroes. If it's larger, just ignore anything that's over the limit.

holiman: (which the current code already does, afaict)

PR #2026, opened 2021-09-14
EIP-3529 fixes

kclowes:
### What was wrong?

There were a few more gas used calculations that needed to be updated to take into account the new `MAX_REFUND_QUOTIENT`. I ran against the `london-tests` branch, and we're down to 25 failures 🎉 

### How was it fixed?

De-duped some of the gas calculation logic floating around. IMHO this trades some code clarity for DRY-ness ¯\_(ツ)_/¯
There is still some duplication of the gas calculation logic in the `<fork name>/state.py`. I think we could clean that up too, but wanted to think about where that actually belongs. 

Also adds a new constant: MAX_REFUND_QUOTIENT. 

It may make sense to review this PR in the context of #2020 as well. 

### To-Do 

- [x] Clean up commit history

#### Cute Animal Picture


eth/constants.py
```diff
@@ -79,6 +79,7 @@
 
 GAS_MEMORY_QUADRATIC_DENOMINATOR = 512
 
+MAX_REFUND_QUOTIENT = 2
```
carver, line 82: It's a little surprising to me, for this to be in the global constants file, since it's changing around from fork to fork.

eth/vm/forks/frontier/__init__.py
```diff
@@ -36,7 +37,8 @@
 
 def make_frontier_receipt(base_header: BlockHeaderAPI,
                           transaction: SignedTransactionAPI,
-                          computation: ComputationAPI) -> ReceiptAPI:
+                          computation: ComputationAPI,
+                          gas_used: int) -> ReceiptAPI:
```
carver, line 41: There's a little ambiguity here for a naive reader: is this the transaction's gas used, or the cumulative gas used in the header (and before or after the transaction). Maybe something like:
```suggestion
                          new_cumulative_gas_used: int) -> ReceiptAPI:
```

Or, if you can't stomach the name, at least a docstring, though I like a self-documenting name better. :)

eth/vm/forks/frontier/__init__.py
```diff
@@ -47,16 +49,6 @@ def make_frontier_receipt(base_header: BlockHeaderAPI,
...
 
-    gas_remaining = computation.get_gas_remaining()
-    gas_refund = computation.get_gas_refund()
-    tx_gas_used = (
-        transaction.gas - gas_remaining
-    ) - min(
-        gas_refund,
-        (transaction.gas - gas_remaining) // 2,
-    )
-    gas_used = base_header.gas_used + tx_gas_used
```
carver, line 58: Interestingly, this seems to remove *all* usages of `transaction` _and_ `base_header`, so it seems best to remove them as parameters. I interpret that as a good sign, that the method got simpler.

eth/vm/forks/berlin/__init__.py
```diff
@@ -28,17 +28,6 @@
 from .state import BerlinState
 
 
-def finalize_gas_used(transaction: SignedTransactionAPI, computation: ComputationAPI) -> int:
```
carver, line 31: Yup, moving this to a `classmethod` seems good. 👍🏻

eth/vm/forks/berlin/__init__.py
```diff
@@ -52,14 +41,15 @@ class BerlinVM(MuirGlacierVM):
...
+    @classmethod
     def make_receipt(
+            cls,
             base_header: BlockHeaderAPI,
             transaction: SignedTransactionAPI,
             computation: ComputationAPI,
             state: StateAPI) -> ReceiptAPI:
 
-        gas_used = base_header.gas_used + finalize_gas_used(transaction, computation)
+        gas_used = base_header.gas_used + cls.finalize_gas_used(transaction, computation)
```
carver, line 52: If you think it's a good idea (I'm on the fence), it should be possible to backport this berlin implementation of `make_receipt()` to byzantium. Then this wouldn't be a custom method at all here, AFAICT.

carver (approved):

eth/constants.py
kclowes: This is the original value of the MAX_REFUND_QUOTIENT that had been sprinkled around (as the `2` in `gas_used // 2`), but AIUI not formally named until EIP-3529. I was sort of on the fence about adding it as a constant, but thought it might be easier down the road to do a codebase-wide search for it if it changes again later. Do you think it makes more sense to just keep it as `2`?

eth/vm/forks/frontier/__init__.py
kclowes: Oops, nice catch!

eth/vm/forks/berlin/__init__.py
kclowes: Hmm, I don't have a good feel for why this custom method was added in the first place. Do you know the background?

eth/constants.py
carver: No, I like having it as a constant, just maybe in frontier fork constants?

eth/vm/forks/berlin/__init__.py
carver: It's just because Berlin now has different transaction types, which define different receipt types. So it has to ask the transaction to make the receipt. But that should continue to work in pre-Berlin, I think. It just never got backported. It's fine to skip this, if that doesn't make sense. The situation is certainly not any worse than it was before.

kclowes: Got it. That makes sense. Yeah, I think I like it slightly better to always ask the transaction 👍 I will update!

carver:
Cool, looks great. I love how the final result was a reduction in LOC. It reads better, too! I love when a bugfix does that.

PR #2045, opened 2021-11-19
Arrow Glacier hard fork / EIP-4345 support

fselmo:
### What was wrong?

- Arrow Glacier hard fork is just around the corner and we need to support it.

### How was it fixed?

- Added support for EIP-4345 which is the only change we need to support Arrow Glacier.

### To-Do

- [x] Clean up commit history
- [x] Add test fillers + tests for new `ethereum/tests` difficulty tests


tests/database/test_header_db.py
```diff
@@ -48,8 +48,8 @@
 from eth.vm.forks.london import (
     LondonVM,
```
marcgarreau, line 49: LondonVM -> ArrowGlacierVM

marcgarreau (approved):
LGTM! appreciate the pair review. difficulty tests look like a solid starting point.

PR #2048, opened 2021-12-16
Bump py-ecc requirement

kclowes:
### What was wrong?
The eth-typing upgrade caused a cascading bunch of dependency resolution errors. 

### How was it fixed?
Bumped py-ecc to be able to use the newest version. I'm not sure why it has such a wide range of possible versions, but opted to keep that.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
LGTM on tests passing! 👌

PR #2062, opened 2022-05-20
chore: update eth dependencies

unparalleled-js:
### Todo:

- [ ] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,15 +7,15 @@ deps = {
     'eth': [
         "cached-property>=1.5.1,<2",
         "eth-bloom>=1.0.3,<2.0.0",
-        "eth-keys>=0.3.4,<0.4.0",
-        "eth-typing>=2.3.0,<3.0.0",
-        "eth-utils>=1.9.4,<2.0.0",
+        "eth-keys>=0.3.4,<0.5.0",
+        "eth-typing>=2.3.0,<4.0.0",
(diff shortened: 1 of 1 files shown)


carver:
Every time you tighten a dependency, it is a breaking change which can cause significant downstream effects. If it's really as easy "we don't use any APIs that changed between these major versions" then maybe we can increase the upper version and leave the lower version in place.

carver (approved):
👍🏻 (with a squash merge if @unparalleled-js doesn't beat us to it with a squash rebase 😄 )

unparalleled-js:
> I screwed up the release numbering, so this is actually done in py-evm `0.5.0-alpha.2`. The corresponding eth-tester version that should work with these dep updates is: `0.6.0-beta.5`, and web3py `v6.0.0-beta.2`. Thanks for the PR though @unparalleled-js! I think it may make sense to release the `0.5.0-alpha.2` changes as `0.6.0-alpha.1` though so I'll put that on my list.

I for some reason am unable to get these updates!  I am installing via `web3[tester]`:

```
"web3[tester]>=5.29.1,<6.0"
```

and every time, it gives me old versions of `eth-utils` `eth-abi` etc. When I try to update them, the conflicts always brings me back to `py-evm`

kclowes:
This needs to be: 
`"web3[tester]>=6.0.0,<7.0"`

Issue #2044, opened 2021-11-18
Make `block_number` configurable in genesis header again


fselmo:
### What is wrong?

- After London changes, certain genesis header params became unconfigurable. Introduce back the ability to configure `block_number`. See, eth-tester issue [#225](https://github.com/ethereum/eth-tester/issues/225).

### How can it be fixed

- `fill_header_params_from_parent()` needs the ability to pass in a `block_number` kwarg to be configurable again. This call is made from the [create_header_from_parent()](https://github.com/ethereum/py-evm/blob/cba70e18c7fd746a50091c426b72ced42712551a/eth/vm/forks/london/headers.py#L114) method in the header classes.

- `block_number` should be a parameter in the [fill_header_params_from_parent()](https://github.com/ethereum/py-evm/blob/cba70e18c7fd746a50091c426b72ced42712551a/eth/_utils/headers.py#L55-L66) method that can be configured but should still keep the current default to `GENESIS_BLOCK_NUMBER` if the parent is None or to the parent `block_number` + 1 if the parent exists.

At a quick glance I believe this is the only change necessary. Testing should be added as well.

catoenm:
Only question here is whether or not we should be passing the `parent_hash` too when `block_number` is specified. Is this correct?

fselmo:
@catoenm I'd be curious to hear your use-case for this change. It would be strange to me to provide a `block_number` and not have a parent OR not assume the parent is the genesis header. And in that case we don't need to provide the parent hash.

In other words... if the header to be configured has a parent, then the `block_number` shouldn't be configurable; it should be `parent.block_number + 1`. If the header to be configured does not have a parent, then it should be assume to be genesis.

In this sense something like this might make the most sense ? :

```py
    if parent is None:
        parent_hash = GENESIS_PARENT_HASH
        block_number = block_number if block_number else GENESIS_BLOCK_NUMBER
        if state_root is None:
            state_root = BLANK_ROOT_HASH
    else:
        parent_hash = parent.hash

        if block_number:
            raise PyEVMError("block_number cannot be configured if a parent header exists.")
        block_number = BlockNumber(parent.block_number + 1)

        if state_root is None:
            state_root = parent.state_root
```

Let me know if I'm missing anything.

catoenm:
This makes a lot of sense. I've updated https://github.com/ethereum/py-evm/pull/2063 accordingly.

Pet3ris:
[...]

PR #2063, opened 2022-05-23
Make `block_number` configurable in genesis header again

catoenm:
### What was wrong?
https://github.com/ethereum/py-evm/issues/2044

### How was it fixed?
Fixed according to issue log.


tests/core/vm/test_vm.py
```diff
@@ -191,3 +194,55 @@ def test_validate_gas_limit_too_high(noproof_consensus_chain):
...
+        }, False),
+        ({
+            'gas_limit': 1,
+            'difficulty': 10,
+            'timestamp': 100,
+            'block_number': 1000,
+        }, False)
+    )
+)
+def test_fill_header_params_from_parent(custom_header_params, null_parent,
```
fselmo, line 221: The test cases look good here. We do try to avoid conditionals in our tests and instead split each test case out into their own test. With conditionals, it can be hard to read what test case ends where and to know which conditional is causing the test to pass over another without debugging the test, etc.

If we can split each test case out into its own test, with an explicit title of the thing it is testing, I think that would be ideal. We can probably factor out the parametrized test cases into a list and pass them into each test if they end up sharing similar info too... something like:

```py
CUSTOM_HEADER_TEST_DATA = (
    {
         'gas_limit': 1,
         'difficulty': 10,
         ...
    },
    { ... },
    { ... },
)

@pytest.mark.parametrize('custom_header_params', CUSTOM_HEADER_TEST_DATA)
def test_case_1(custom_header_params):
     ...
     
@pytest.mark.parametrize('custom_header_params', CUSTOM_HEADER_TEST_DATA)
def test_case_2(custom_header_params):
     ...
```

Thoughts?

fselmo:
Thanks for working this @catoenm! This looks good to me. I just left one refactor comment on the tests and then I can add a newsfragment and merge it 👌

Issue #2054, opened 2022-02-24
Upgrade py-trie


unparalleled-js:
* py-evm Version: 0.5.0a3
* OS: macOS
* Python Version (python --version): 3.10.1
* Environment (output of `pip freeze`):
* aiohttp==3.8.1
aiosignal==1.2.0
-e git+ssh://git@github.com/ApeWorX/ape-cairo.git@006af767524ffb10293f48912e5f7625eaae42d4#egg=ape_cairo
appnope==0.1.2
argcomplete==1.12.3
async-timeout==4.0.2
asynctest==0.13.0
attrs==21.4.0
backcall==0.2.0
backports.cached-property==1.0.1
base58==2.1.1
bitarray==1.2.2
black==21.12b0
bleach==4.1.0
cached-property==1.5.2
cachetools==5.0.0
cairo-lang==0.7.1
certifi==2021.10.8
cffi==1.15.0
cfgv==3.3.1
charset-normalizer==2.0.12
click==8.0.4
colorama==0.4.4
commitizen==2.20.1
commonmark==0.9.1
coverage==6.3.2
cytoolz==0.11.2
dataclassy==0.10.4
decli==0.5.2
decorator==5.1.1
Deprecated==1.2.13
distlib==0.3.4
docopt==0.6.2
docutils==0.18.1
ecdsa==0.17.0
eth-abi==2.1.1
eth-account==0.5.7
eth-ape==0.1.0b5
eth-bloom==1.0.4
eth-hash==0.3.2
eth-keyfile==0.5.1
eth-keys==0.3.4
eth-rlp==0.2.1
eth-tester==0.6.0b6
eth-typing==2.3.0
eth-utils==1.10.0
ethpm-types==0.1.0b7
execnet==1.9.0
fastecdsa==2.2.3
filelock==3.6.0
flake8==3.9.2
frozendict==1.2
frozenlist==1.3.0
hexbytes==0.2.2
hypothesis==6.37.2
identify==2.4.11
idna==3.3
importlib-metadata==4.11.1
iniconfig==1.1.1
ipdb==0.13.9
ipfshttpclient==0.8.0a2
ipython==7.31.1
isort==5.10.1
jedi==0.18.1
Jinja2==3.0.3
jsonschema==3.2.0
keyring==23.5.0
lark-parser==0.12.0
lru-dict==1.1.7
MarkupSafe==2.1.0
marshmallow==3.14.1
marshmallow-dataclass==8.5.3
[...]

PR #2065, opened 2022-05-27
Pin trie package to 2.0.0

simon-saliba:
### What was wrong?

FIX: https://github.com/ethereum/py-evm/issues/2054
Unable to install in environments where typing extensions >= 4.0 (which is most environments these days).
This has been fixed in the newer trie packages but this package still uses an older one.
This all trickles down the line and causes issues installing other applications.

### How was it fixed?

Pin to trie==2.0.0 instead of trie==2.0.0a5.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


fselmo:
This has been done, thanks!

PR #2070, opened 2022-06-24
Drop Python 3.6

kclowes:
### What was wrong?
Python 3.6 is no longer supported, so dropping here. 

### How was it fixed?
Removed CI runs, and updated any runs using 3.6 to use Python 3.9 now.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/eth/vm/forks/frontier/__init__.py
+++ b/eth/vm/forks/frontier/__init__.py
@@ -58,7 +58,7 @@ def make_frontier_receipt(computation: ComputationAPI,
 
 class FrontierVM(VM):
     # fork name
-    fork: str = 'frontier'  # noqa: E701  # flake8 bug that's fixed in 3.6.0+
+    fork: str = 'frontier'
 
     # classes
     block_class: Type[BlockAPI] = FrontierBlock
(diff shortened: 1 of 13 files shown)


fselmo:
@kclowes I found a few more things we can probably remove or change in this branch:

* We can remove this (I don't believe it's being used anywhere) https://github.com/kclowes/py-evm/blob/drop-py36//tests/core/helpers.py#L13-L16

* 3.6 -> 3.9 here https://github.com/kclowes/py-evm/blob/drop-py36//.readthedocs.yml#L5 and here https://github.com/kclowes/py-evm/blob/drop-py36//docs/conf.py#L190

* Remove this line https://github.com/kclowes/py-evm/blob/drop-py36/tox.ini#L49

and we might as well update the example here ? https://github.com/kclowes/py-evm/blob/drop-py36//eth/_utils/version.py#L10

eth/vm/forks/frontier/__init__.py
```diff
@@ -58,7 +58,7 @@ def make_frontier_receipt(computation: ComputationAPI,
 
 class FrontierVM(VM):
     # fork name
-    fork: str = 'frontier'  # noqa: E701  # flake8 bug that's fixed in 3.6.0+
+    fork: str = 'frontier'
```
fselmo, line 61: Nice, yeah I guess these can go too 👌🏼

fselmo (approved):
Lgtm! 👍🏼

PR #2077, opened 2022-08-22
Bump towncrier version to remove double headers

kclowes:
### What was wrong?
Towncrier versions >=19,<20 have a bug where double headers get printed if there is a custom version specified in `pyproject.toml`

### How was it fixed?
Upgraded the towncrier version to a version that doesn't have this bug. Had to add the `build` command to the towncrier command with the upgrade. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


pacrob (approved):
lgtm!

Issue #2075, opened 2022-06-30
Address newly-added failing ``ethereum/tests``


fselmo:
### What is wrong?

With the newly updated `ethereum/tests`, some tests were added that cause failures.

- [x] Tests added in `ethereum/tests` `v11.1`:
    - Un-comment [these lines](https://github.com/fselmo/py-evm/commit/3384f65d1c3562d61c5fa33fe90504cf81195cad) and address these failing test cases.
    - Fixed by #2084

- [x] [EIP-2681 related tests](https://github.com/ethereum/EIPs/pull/4784):
    - Un-comment [these lines](https://github.com/ethereum/py-evm/blob/master/tests/json-fixtures/blockchain/test_blockchain.py#L314-L319) and make the necessary changes to pass these based on the link above.
    - Fixed by  #2110

- [x] Failing tests beginning at Berlin:
    - Un-comment [these lines](https://github.com/ethereum/py-evm/blob/master/tests/json-fixtures/test_transactions.py#L105-L112) and address these failing test cases.
    - Fixed by #2111 

### How can it be fixed

I have not yet looked into why any of these fail but the EIP-2681 tests should have guidance from the updated EIP discussion linked above.

PR #2084, opened 2022-10-07
Address failing tests from ethereum/tests ``v11.1``

fselmo:
### What was wrong?

Related to #2075 

- Difficulty calculation was taking into account the parent ``difficulty`` which would slip below the ``DIFFICULTY_MINIMUM`` value if the parent ``difficulty`` was less than the allowed minimum. 

### How was it fixed?

- Take the max only between the difficulty calculation and the _actual_ minimum allowed value.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screen Shot 2022-10-20 at 13 27 26


kclowes (approved):
LGTM! 🎉

PR #2087, opened 2022-11-04
Update OpenSSL Config for ripemd160

fselmo:
### What was wrong?

- OpenSSL 3 removed ripemd160 functions from the default behavior. Instead they now have to be locally configured. This broke a lot of blockchain related logic and they decided to add it back in in version 3.0.7, see blog post [here](https://www.openssl.org/blog/blog/2022/10/18/rmd160-and-the-legacy-provider/).

- CircleCI is using openssl3 but not (yet?) the latest version. Since this breaks tests even on `master`, which were previously passing, we need to update the config for now until the newer open ssl version is used in the circleci images.

### How was it fixed?

- For now, updating the openssl config just works™ and we can ideally remove it when the images are updated.
- Config changes suggested, among other places, [here](https://github.com/openssl/openssl/issues/16994#issuecomment-963669325)

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

IMG_2333

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,6 +8,26 @@ common: &common
   steps:
     - checkout
     - run:
+        # this is necessary until circleci images begin using openssl>=3.0.7
+        name: update openssl.cnf for ripemd160
+        command: |
+          sudo tee -a /etc/ssl/openssl.cnf >/dev/null <<'EOF'
+
+          [openssl_init]
+          providers = provider_sect
+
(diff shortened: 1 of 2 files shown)


pacrob (approved):
lgtm!

fselmo:
Note: I emailed circleci support to ask where we can track the openssl version on the Python images. I will update here when I get a response but for now we can watch the version printed via the build step.

PR #2090, opened 2022-12-16
Remove upper pin on eth-bloom

kclowes:
### What was wrong?
To get dependencies to resolve, we need to remove the upper pin on eth-bloom.

### How was it fixed?
Removed pin per our current philosophy. I will remove other pins in a separate PR. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture


fselmo (approved):
lgtm 👍🏼... optional to update `tox` here as well.

---
**edit:** baby platypuses 👏🏼

PR #2095, opened 2023-04-11
Merge in less-sensitive updates from the project template

pacrob:
### What was wrong?

The project [template](https://github.com/ethereum/ethereum-python-project-template) was updated. This is to merge in many of the changes from it.

Because there are so many changes, this PR is only for files that do not touch dependency or python versions (except for adding pydocstyle). Those will be done separately.

Files that need updating in following pr:
- setup.py
- tox.ini
- mypy.ini
- .circleci/config.yml


LICENSE
```diff
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright 2017-2019 Ethereum Foundation
+Copyright (c) 2020 The Ethereum Foundation
```
kclowes, line 3: I have no idea how this works, but should we update this annually? As in, should this be 2023 now?

kclowes (approved):
Thank you for doing this! I didn't look very closely, but will this throw off the commit history between py-evm and the template next time we merge it in since you didn't take the whole template history?

LICENSE
pacrob: As discussed, I'll update to the range of years in which changes were made to the library.

kclowes:
Got it. Sounds good to me! Thanks again!

Issue #2089, opened 2022-12-14
EOFv1 Checklist


fselmo:
Consult the [EOFv1 checklist](https://notes.ethereum.org/@ipsilon/eof1-checklist#Specification--Tests) for specifics.
---

This is the main tracking issue for the EOFv1 checklist. If you have interest in one of the below tasks, don't hesitate to mention it here, open a PR, and link it in this issue. Drafts are welcome. We can edit this description and link the PR to the respective EIP(s) within its scope to track progress.

- [ ] [EIP-3860](https://eips.ethereum.org/EIPS/eip-3860): Limit and meter initcode (EOF prerequisite) [history](https://github.com/ethereum/EIPs/commits/master/EIPS/eip-3860.md)
- [ ] [EIP-3540](https://eips.ethereum.org/EIPS/eip-3540): EOF - EVM Object Format v1 [history](https://github.com/ethereum/EIPs/commits/master/EIPS/eip-3540.md)
- [ ] [EIP-3670](https://eips.ethereum.org/EIPS/eip-3670): EOF - Code Validation [history](https://github.com/ethereum/EIPs/commits/master/EIPS/eip-3670.md)
- [ ] [EIP-4200](https://eips.ethereum.org/EIPS/eip-4200): EOF - Static relative jumps [history](https://github.com/ethereum/EIPs/commits/master/EIPS/eip-4200.md)
- [ ] [EIP-4750](https://eips.ethereum.org/EIPS/eip-4750): EOF - Functions [history](https://github.com/ethereum/EIPs/commits/master/EIPS/eip-4750.md)
- [ ] [EIP-5450](https://eips.ethereum.org/EIPS/eip-5450): EOF - Stack Validation [history](https://github.com/ethereum/EIPs/commits/master/EIPS/eip-5450.md)

PR #2097, opened 2023-04-18
Refactor current VM computations to be message computations

fselmo:
### What was wrong?

Related to Issue #2089 

Computations currently require a message and a transaction context. This is because, up until EOF, computations were built from a message with transaction-like fields. This PR opens support for computations that do not come from messages. One such example will be an EOF computation which is built from instructions in the bytecode of a message computation. 

### How was it fixed?

To make way for the above changes, the `BaseComputation` class is refactored in this PR to be more basic. It strips out the the idea that all computations rely on a message. This is now a property of the `MessageComputation` class that we are more familiar with, which is still configured as part of all VMs at present, and inherits from and builds upon the `BaseComputation`.

This PR does not attempt to add any new computation classes. It is simply paving the way for them by splitting the current `BaseComputation` class into these two classes.

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture


eth/vm/forks/shanghai/state.py
```diff
@@ -14,7 +14,7 @@ class ShanghaiTransactionExecutor(ParisTransactionExecutor):
 
 
 class ShanghaiState(ParisState):
-    computation_class = ShanghaiComputation
+    message_computation_class = ShanghaiComputation
```
pacrob, line 17: should be `ShanghaiMessageComputation`?

fselmo: should, ty

eth/vm/forks/shanghai/computation.py
```diff
@@ -8,12 +8,12 @@
 from eth.exceptions import (
     OutOfGas,
 )
-from eth.vm.forks.paris.computation import ParisComputation
+from eth.vm.forks.paris.computation import ParisMessageComputation
 
 
-class ShanghaiComputation(ParisComputation):
+class ShanghaiComputation(ParisMessageComputation):
```
pacrob: should be `ShanghaiMessageComputation`?

fselmo: This should be outdated as I refactored the whole class across all files based on the last comment.

pacrob: yup, got it

pacrob (approved):
lgtm!

kclowes (approved):
LGTM! Just had a couple small comments.

eth/vm/computation/base_computation.py
```diff
@@ -0,0 +1,354 @@
...
+    quadratic_cost = size_in_words ** 2 // GAS_MEMORY_QUADRATIC_DENOMINATOR
+
+    total_cost = linear_cost + quadratic_cost
+    return total_cost
+
+
+C = TypeVar("C", bound="ComputationAPI")
+
+
+class BaseComputation(ComputationAPI, Configurable, Generic[C]):
```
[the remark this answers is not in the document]
fselmo: Do you mean `Configurable` and `ComputationAPI` were swapped in order? I don't think it matters much but I can restore the order. I was going for API (abstract class) -> inherited non-abstract class -> and then the generic.

docs/release_notes.rst
```diff
@@ -423,8 +423,8 @@ Internal Changes - for Contributors
 - Fix for creating a duplicate "ghost" Computation that was never used. It didn't
   break anything, but was inelegant and surprising to get extra objects created
   that were mostly useless. This was achieved by changing
-  :meth:`eth.abc.ComputationAPI.apply_message` and
-  :meth:`eth.abc.ComputationAPI.apply_create_message` to be class methods. (`#1921 <https://github.com/ethereum/py-evm/issues/1921>`__)
+  :meth:`eth.abc.MessageComputationAPI.apply_message` and
```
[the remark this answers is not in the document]
fselmo: In my head I was like "I thought I changed that back", and yeah this is outdated already :)

eth/vm/computation/base_computation.py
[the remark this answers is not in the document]
fselmo: I'm going to merge this since it was mostly approved and this seems like a smaller change. I'd like to be able to start rebasing the EOF PR from these changes. I can address some more computation refactoring in the that PR as well. Happy to move this conversation over there.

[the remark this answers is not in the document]
kclowes: > Do you mean since Configurable and ComputationAPI were swapped in order? 

Yep, that's what I was looking at. I was more worried about it just being a copy/pasta rather than a breaking change for users. I didn't look into it too much, so if it doesn't matter this seems fine to me!

PR #2102, opened 2023-04-21
Update pip and virtualenv for CI builds

fselmo:
### What was wrong?

- `virtualenv` versions seem to be clashing across sys and user libraries. CircleCI images come with `virtualenv` installed at the sys level and when we did `pip install --user tox` we installed `tox` and a different `virtualenv` version there at the user package level. I believe the `--user` flag was causing issues with the different `virtualenv` installs. By installing `tox` at the sys level, it allows us to work with the existing `virtualenv` version already in the image. Whether the `tox` version chooses to update it or not is left to its dependency setup.

- Also, ``--upgrade`` the ``pip`` version sitting in the image python install before any of the above steps.

### Todo:
- [x] Clean up commit history

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,7 +49,9 @@ common: &common
           - cache-v1-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
     - run:
         name: install dependencies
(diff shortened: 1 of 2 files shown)


prathmeshranaut (approved):
Thanks for the PR. I think this will fix the issue with the failing CI builds.

pacrob (approved):
lgtm!

PR #2105, opened 2023-05-02
Bugfix: Add receipt builder to ShanghaiBlock

fselmo:
### What was wrong?

- `ShanghaiBlock` class was missing a receipt builder. This is not directly used in PyEVM but is used by `eth-tester`. Since `ShanghaiBlock` is the first time we've had to re-build from `BaseBlock` since the `__init__()` adds a new field, this was the first time in a while we were vulnerable to missing a configuration on the class. Since it wasn't directly tested here, we missed it.

### How was it fixed?

- Add the `LondonReceiptBuilder` to the `ShanghaiBlock` class since that's the latest version of the receipt builder (the last time a typed transaction was introduced was the type=2 transaction via EIP-1559, London).
- Add basic testing around Block class configuration to make sure each is wired up with the expected configuration classes.

### Todo:
- [x] Clean up commit history

### DIFF
--- a/eth/vm/forks/shanghai/blocks.py
+++ b/eth/vm/forks/shanghai/blocks.py
@@ -29,7 +29,10 @@ from eth_utils import (
 )
 
 from rlp.sedes import (
-    Binary, CountableList, big_endian_int, binary,
+    Binary,
+    CountableList,
+    big_endian_int,
+    binary,
 )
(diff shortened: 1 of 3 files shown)


pacrob (approved):
I like adding the `issubclass` checks in there. lgtm!

PR #2106, opened 2023-05-03
Re-refactor pre-EOF

fselmo:
Context: EOF will no longer be in Cancun and the bytecode does not look like the current version of the `ethereum/tests` that are in our fixture directory. That update of the fixture will come in a separate PR. This change in the structure of the tests puts our code base mostly back to where it was to prepare for EOF implementation.

---

This PR puts the state of the Computation class hierarchy mostly back to where it was, keeping some good cleanup and updating the `ComputationAPI` to match properties implemented in the `BaseComputation` implementation class.

Some of the earlier `ethereum/tests` pointed to EOF code within the bytecode, not at the start. This led to a refactoring for the class architecture to make it work with our current model, with the bigger refactor PR from last week. With the updated tests (and checking against the EIPs), it would be best to put the state of the Computation hierarchy mostly back to where it was so as to not break too much of the existing structure. Massaging EOF into this model won't require as drastic of an architecture change.

Since there won't be a release until all these changes are in, the most this does is muddy the commit history / git blame a bit 🙁. At least we gained a touch of housekeeping along the way.

### How was it fixed?

- Refactor away the separation of `MessageComputation` to make room for other `*Computation` classes and return to the `{ForkName}Computation` being the overarching class that takes care of applying messages to state and "applying" itself during bytecode execution.

Note: The newsfragment for the previous refactor PR (#2097) was removed and replaced with the one from this PR.

### Todo:
- [x] Clean up commit history (maybe a bit too much)

#### Cute Animal Picture


pacrob (approved):
one nit, otherwise lgtm!

kclowes (approved):
:shipit: !

PR #2108, opened 2023-05-15
Update ``ethereum/tests`` and turn on all Shanghai tests

fselmo:
- The latest release of ``ethereum/tests`` moves EOF to Cancun, which will be moved again to Prague since 4844 took precedence in Cacun. Either way, Shanghai tests no longer have EOF so we can remove the ``-k "not EOF"`` flag from the Shanghai tests

### Todo:
- [x] Clean up commit history

#### Cute Animal Picture

IMG_5776


pacrob (commented):
lgtm!

PR #2115, opened 2023-06-05
update CodeStream slot name pc to program_counter

### LINKED ISSUE #2109: 'pc' should be changed to 'program_counter' in the __slots__ of CodeStream class
`pc` should be changed to `program_counter` in the `__slots__` of CodeStream class. Although it does not raise errors (since `program_counter` is inherited from `CodeStreamAPI`), I think should correct it.

```python
class CodeStream(CodeStreamAPI):
    __slots__ = [
        "_length_cache",
        "_raw_code_bytes",
        "invalid_positions",
        "valid_positions",
        "pc",
    ]

    logger = logging.getLogger("eth.vm.CodeStream")

    def __init__(self, code_bytes: bytes) -> None:
        validate_is_bytes(code_bytes, title="CodeStream bytes")
        # in order to avoid method overhead when setting/accessing pc, we no longer
        # fence the pc (Program Counter) into 0 <= pc <= len(code_bytes).
        # We now let it float free.
        # NOTE: Setting pc to a negative value has undefined behavior.
        self.program_counter = 0
        self._raw_code_bytes = code_bytes
        self._length_cache = len(code_bytes)
        self.invalid_positions: Set[int] = set()
        self.valid_positions: Set[int] = set()
```

https://github.com/ethereum/py-evm/blob/8aa9eb51944b4dcf6b365eb032ac674fed933c09/eth/vm/code_stream.py#LL21C1-L42C47

pacrob:
### What was wrong?

`pc` was used in the `__slots__` list, but `program_counter` was used throughout the rest of the `CodeStream` class.
Closes #2109 

### How was it fixed?
Changed slot `pc` to `program_counter`.
Quick `timeit` tests showed ~25% faster access to `program_counter` :tada: 

### Todo:

#### Cute Animal Picture

image

### DIFF
--- a/eth/vm/code_stream.py
+++ b/eth/vm/code_stream.py
@@ -24,17 +24,17 @@ class CodeStream(CodeStreamAPI):
         "_raw_code_bytes",
         "invalid_positions",
         "valid_positions",
-        "pc",
+        "program_counter",
     ]
 
     logger = logging.getLogger("eth.vm.CodeStream")
 
     def __init__(self, code_bytes: bytes) -> None:
         validate_is_bytes(code_bytes, title="CodeStream bytes")
-        # in order to avoid method overhead when setting/accessing pc, we no longer
-        # fence the pc (Program Counter) into 0 <= pc <= len(code_bytes).
+        # in order to avoid method overhead when setting/accessing program_counter,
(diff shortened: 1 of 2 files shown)


reedsa (approved):
lgtm

fselmo (approved):

PR #2121, opened 2023-09-05
Remove dependence on ``ethash`` (pyethash)

fselmo:
### What was wrong?

- The goal for `py-evm` was once to provide the Trinity client with an EVM accessible via Python. This is not the case anymore as Trinity was sunset. With different consensus configurations, py-evm still serves well today for EVM logic testing. With the move to PoS (Proof of Stake) after the merge, there is a drive to remove the dependency on the `ethash` library. There a few implementations of the Ethash algorithm in python and the goal of internalizing this logic into python-only code if desired is accessible without too much dev time being taken up. Using ethereum.org and the execution-specs as references, implement an in-house ethash algorithm that works well enough, even if it is significantly slower.

- _Added motivation:_ Python 3.10+ support in `ethash` leads to unknown segfaults. Maintaining the C bindings file in `ethash` for little benefit becomes an added chore to debug, find someone to review + approve, and update. Internalizing the ethash algorithm gives ease of control for supporting new python versions.

### How was it fixed?

- ethash implemented within py-evm as python code and POW is un-prioritized as a consensus choice within the library.

### Todo:
- [x] Clean up commit history


tests/core/consensus/test_pow_mining.py
```diff
@@ -80,7 +82,7 @@ class ChainClass(MiningChain):
 
     chain = genesis(ChainClass)
 
-    block = chain.mine_block(difficulty=1)
+    block = chain.mine_block(difficulty=3)
```
wolovim, line 85: any important context to leave behind in comments for `TEST_NUM_CACHES`, difficulty number, or `_concurrently_run_to_completion` numbers or is that sufficiently summarized by the `skip` reasons?

wolovim (approved):
not at all convinced i know the full implications, but can definitely appreciate the dep removal and better modern python support 👏 for our purposes, i don't see the algo performance being a blocker.

PR #2135, opened 2024-01-23
EIP-4788: parent beacon block root

fselmo:
### What was wrong?

 EIP-4788 needs implementation for the cancun network upgrade

### How was it fixed?

Implement EIP-4788:
  - Incorporate the `parent_beacon_block_root` into block headers, beginning with Cancun.
  - Update the code at `BEACON_ROOTS_ADDRESS` at init of `CancunVM` if no code exists at that address.
  - Update the values at the specified slots for `BEACON_ROOTS_ADDRESS` as per specifications on EIP-4788.

~~Something we need to consider is how we want to handle the beacon roots contract when the tests don't initialize a pre-state with the right code at that address.~~[I opted to inject the code at init of the `CancunVM` if it doesn't already exist. We can revise this by the end of the Cancun changes if need be.]

Note: All related tests are passing except for:
   - `2` tests that test for a transaction with `type=3` related to EIP-4844 not having been implemented yet
   - `1` test that tests the beacon root along with a test for EIP-6780, related to self-destruct, which has not been implemented yet

Reviewers can check the above tests ^ with the following command:

`pytest tests/json-fixtures/blockchain/test_blockchain.py --fork Cancun -k beacon_root`

### Todo:


pacrob (approved):
lgtm!

PR #2139, opened 2024-01-30
Add ``c-kzg-4844`` dependency

fselmo:
### What was wrong?

We are missing a kzg 4844 library to implement EIP-4844

### How was it fixed?

Add the [c-kzg-4844](https://github.com/ethereum/c-kzg-4844) library as a dependency.

### Todo:

- [ ] Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/main/newsfragments/README.md)

#### Cute Animal Picture

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ common: &common
 
           EOF
     - run:
-        name: checkout fixtures submodule
+        name: checkout submodules
         command: git submodule update --init --recursive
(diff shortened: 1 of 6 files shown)


fselmo:
I think for now we can do what the EELS team has done and use the consensus specs while we are still in proto-danksharding. One day this may not be sufficient enough and we may need the full library to be available. Closing this for now.

PR #2146, opened 2024-03-07
Move newsfragment

pacrob:
### What was wrong?

I put the newsfragment for pr #2137 in the top level folder instead of newsfragments.

### How was it fixed?

Moved the file into the correct location. It does look like it should have been included in the previous release. I think that this is OK because it's internal CI/docs build stuff, but open to other options if not OK.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


fselmo (approved):
good catch, lgtm

PR #2151, opened 2024-03-11
EIP-4844 and EIP-7516 support

fselmo:
EIP-4844 support

- Add dependency on newly released `ckzg` library.
- Implement type=3 blob transacitons
- Implement BLOBHASH opcode
- Implement BLOBBASEFEE opcode
- Implement `blob_gas_used` and `excess_blob_gas` on Cancun block headers
- Turn on blockchain tests related to 4844 and 7516
- Turn on transaction tests for Cancun 

Note: The state of the API and transaction field classes is quite messy. We should address this in a separate PR. This will have a few undesired `type: ignore` lines for now until we can fix the API ABC class mess that's happening across legacy vs type, signed vs unsigned, fields vs methods... etc... I'll attempt to fix this before we merge cancun to main but maybe we can release with working logic and clean up the typing after the first cancun support release. Any thoughts on that release process are welcome.

### Todo:

- [x] Fix lint issues
- [x] Investigate a handful of tests that halt the test suite / neither pass nor fail but hang. (addressed in 3a1753c)

#### Cute Animal Picture

Screenshot 2024-03-12 at 11 59 25


eth/vm/forks/cancun/state.py
```diff
@@ -15,12 +36,143 @@
...
+from .transaction_context import (
+    CancunTransactionContext,
+)
+from .transactions import (
+    BlobTransaction,
+)
+
+
+def fake_exponential(
+    factor: int, numerator: int, denominator: int, max_iterations: int = 10000
```
fselmo, line 55: Keep an eye on whether this max_iterations ever breaks a test... we exit early so that tests don't hang with long running exponential values

eth/vm/state.py
```diff
@@ -111,6 +111,12 @@ def get_tip(self, transaction: SignedTransactionAPI) -> int:
     def get_gas_price(self, transaction: SignedTransactionAPI) -> int:
         return transaction.gas_price
 
+    @property
+    def blob_base_fee(self) -> int:
+        raise NotImplementedError(
+            "Basefee opcode is not implemented prior to Cancun hard fork"
```
fselmo, line 117: Nope. Blob Base Fee....

PR #2157, opened 2024-03-15
Use some general state tests as transaction tests; Ensure txn `type_id`

fselmo:
### What was wrong?

- A `SpoofTransaction` is usually used to turn an unsigned transaction into a "signed" transaction for estimating gas for example. *eth-tester* uses `SpoofTransaction` for these purposes and we need to make sure that it also accounts for a `type_id` which is a part of the `SignedTransactionAPI`.
- Quite a few transaction tests can be pulled in from general state tests.

### How was it fixed?

- Add proper `type_id` values to `SpoofTransaction` since it should use all attributes from the `SignedTransactionAPI`.
- Pull in transaction tests from some general state tests

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

prompt: `iterate on 'snakes transitioning from the Shanghai city life to a life in Cancun with a sign that says "Cancun"' but make 'transactions' a theme`

Screenshot 2024-03-15 at 17 42 07

### DIFF
--- a/eth/_utils/spoof.py
+++ b/eth/_utils/spoof.py
@@ -19,6 +19,7 @@ from eth.constants import (
 )
 
 SPOOF_ATTRIBUTES_DEFAULTS = {
+    "type_id": None,
     "y_parity": DEFAULT_SPOOF_Y_PARITY,
     "r": DEFAULT_SPOOF_R,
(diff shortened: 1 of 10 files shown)


pacrob (approved):
lgtm!

PR #2168, opened 2024-04-05
Compare the first versioned hash byte by indexing

fselmo:
### What was wrong?

- ``integer.to_bytes()`` requires size and byteorder below py311 and our fixture tests only use py311.

### How was it fixed?

- Compare the first byte of versioned hashes by indexing instead.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2024-04-05 at 10 54 58


kclowes (approved):
Nice solution!

PR #2172, opened 2024-04-17
Remove scripts/__init__.py, don't include scripts dir in sdist

kclowes:
### What was wrong?
scripts was getting installed as a package because we added a `scripts/__init__.py`, but it really shouldn't be a package. 

### How was it fixed?
Removed `scripts/__init__.py`, and made sure it wasn't getting included in the wheel. It is still getting included in the source dist.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup(
     license="MIT",
     zip_safe=False,
     keywords="ethereum blockchain evm",
-    packages=find_packages(exclude=["scripts", "tests", "tests.*"]),
+    packages=find_packages(exclude=["scripts", "scripts.*", "tests", "tests.*"]),
     package_data={"eth": ["py.typed"]},
     classifiers=[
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -87,7 +87,7 @@
     license="MIT",
     zip_safe=False,
     keywords="ethereum blockchain evm",
-    packages=find_packages(exclude=["scripts", "tests", "tests.*"]),
+    packages=find_packages(exclude=["scripts.*", "tests", "tests.*"]),
```
fselmo, line 90: I still see it in the wheel after this change but I don't see it when we have both `scripts` and `scripts.*` in here. This might have given a false positive since we also removed the `__init__.py`.

fselmo (approved):
lgtm 👍🏼

PR #2197, opened 2025-02-04
bump `mypy` to run locally, move tools/factories into tests folder

pacrob:
### What was wrong?

`mypy` was still running without locally installed deps.
`factory-boy` is only used in testing, but tools using it were within a public folder.

### How was it fixed?
Run `mypy` locally - decided to do this here, as opposed to with the template update, to keep related change more clearly separated.

Moved `eth/tools/factories` to `tests/tools/factories`.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

image


reedsa (approved):
lgtm

PR #2201, opened 2025-02-21
fill prague tests

kclowes:
### What was wrong?
Adds prague tests from EEST by running `python scripts/fill_eest_tests.py --execution-spec-tests-dir="../execution-spec-tests" --branch="pectra-devnet-6" --fork="Prague" -n 4`

There are about ~40 tests (I think) that aren't being run/generating anything because the tests hang at the end. 

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`0_o`


fselmo:
The reason the tests weren't running was related to how the eest fixtures directory was being referenced `fixtures_EEST` -> `fixtures_eest`. This was still being recognized locally (MacOS) but not on the CI (linux).

fselmo:
I moved the filled tests to a static repo on my github acct for now and we can reference with an official static reference if / when that exists. I think it's a lot better than storing millions of lines of code, even if temporarily.

PR #2204, opened 2025-02-28
Implement EIPs: 7623, 2935, 7691, 2537

fselmo:
### What was wrong?

- Comment out the WIP 7702 opcodes for now as they were interfering with some tests.
- [Implement EIP-7623: Increase calldata cost](https://github.com/ethereum/py-evm/commit/f18cfa61b9ca92c670546eb5bde982ecef026502)
  - `STANDARD_TOKEN_COST` as per `EIP-7623` is really just another way of calculating the current gas schedule. It factors the 4 out of (4 * zero_bytes) + (16 * non_zero_bytes), leaving (zero_bytes + 4 * non_zero_bytes). Leave the current schedule untouched and instead compare this to the new data_floor when finalizing the computation, consuming more gas if necessary until we reach the data floor cost.
- [Implement EIP-2935: Store historical block hashes](https://github.com/ethereum/py-evm/commit/d31b9f8ed43aacd7f42b763bcb092a3a5d0beee8)
- [Implement EIP-7691: Blob throughput increase](https://github.com/ethereum/py-evm/commit/134a3efbddf2cc3d8c1247a13aac7afa3db06883)
- [Implement EIP-2537: Precompiles for BLS12-381](https://github.com/ethereum/py-evm/pull/2204/commits/20ee3c4572d3e897750766b59c218d785c99e787)
- Turn on `CancunToPragueAtTime15k` transition tests. Still missing some EIP implementations. `10` tests (out of `30`) still fail related to eip-7002 and eip-7251, which are not yet implemented.

---
**This section can be ignored. These issues have since been fixed with the latest push to use the generated tests from the latest EEST release tarball 🎉 . I must've been doing something wrong when filling the tests.**

~~Note:~~

~~- There's a strange behavior in the tests I can't account for at the moment where it expects another address that isn't the history contract EIP address to store the same slots as that contract. I bet these will start passing as we implement more EIPs. For now, if I add a second check for that same address to the `PragueState` and update it with the same exact logic as we do for the history contract, it passes all tests for blobs and calldata increase with the exception of one refund test related to empty storage and type 4 transaction tests. I believe the refund test has to do with a not-yet-implemented [EIP](https://github.com/ethereum/py-evm/commit/134a3efbddf2cc3d8c1247a13aac7afa3db06883).~~

~~For testing, I created the following patch that can be applied that will look for this contract as well as the history contract so we can get over this hump for now.~~

~~[history_contract.patch](https://github.com/user-attachments/files/19171665/history_contract.patch)~~

---

### Todo:

- [x] Clean up commit history


fselmo:
There is one outstanding test failure for calldata cost, related to the data floor cost and refuning, that I'd like to get to tomorrow but otherwise all other tests are passing for all EIPs implemented here - with the exception of type 4 transactions which have not been implemented yet. This means we're passing the new point at infinity tests related to BLS12 precompile that was recently found.

The blob throughput tests come from Cancun blob tests which can be run against prague via `--fork=Prague -k eip4844`.

kclowes (approved):
🚀

PR #2206, opened 2025-03-19
Finalize EIP-7702 implementation

kclowes:
### What was wrong?

- EIP-7702 implementation needs to be finalized

### How was it fixed?

- Finalize EIP-7702 implementation
- Pull in EEST stable tests from `v4.2.0`
- Fix EIP-7623 logic + track data floor gas separately from computation gas meter

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

Screenshot 2025-04-18 at 16 31 14


fselmo (approved):
@kclowes I'm good with the state of this if we want to merge it into `prague`. Tests are passing but we should look at any refactors we may want.

Then we can do some final reviews before merging `prague` into `main`.

PR #2208, opened 2025-04-21
`prague` -> `main` last checks

fselmo:
### What was wrong?

- PR to diff all `prague` changes against `main` for one last check before merging.

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`0_o`


eth/vm/forks/berlin/transactions.py
```diff
@@ -371,15 +376,15 @@ def max_fee_per_gas(self) -> int:
 
     @property
     def max_fee_per_blob_gas(self) -> int:
-        raise NotImplementedError(
-            "max_fee_per_blob_gas is not implemented until Cancun."
-        )
+        return self._inner.max_fee_per_blob_gas
```
kclowes, line 379: Will this (and `blob_versioned_hashes`, and `authorization_list` below) cause a problem if called from the berlin vm?

eth/vm/logic/call.py
```diff
@@ -49,23 +62,18 @@ def compute_msg_gas(
         child_msg_gas = gas + (constants.GAS_CALLSTIPEND if value else 0)
         return child_msg_gas, total_fee
 
-    def get_account_load_fee(
+    def consume_account_load_gas(
```
kclowes, line 65: I think I'm missing something here. Why is this an empty method?

eth/vm/logic/storage.py
```diff
@@ -122,8 +122,10 @@ def net_sstore(gas_schedule: NetSStoreGasSchedule, computation: ComputationAPI)
 
             if original_value == value:
                 if original_value == 0:
+                    # adds 19800 to refund counter
```
kclowes, line 125: nit - feels like it could get stale and be confusing if this number changes in a future fork?
```suggestion
                    # adds SOME_CONSTANT to refund counter
```

eth/vm/forks/prague/transaction_context.py
```diff
@@ -0,0 +1,16 @@
...
+
+from eth.abc import (
+    SetCodeAuthorizationAPI,
+)
+from eth.vm.forks.cancun.transaction_context import (
+    CancunTransactionContext,
+)
+
+
+class PragueTransactionContext(CancunTransactionContext):
```
kclowes, line 13: I do not think we're using this anywhere.

eth/vm/transaction_context.py
```diff
@@ -50,3 +54,7 @@ def origin(self) -> Address:
     @property
     def blob_versioned_hashes(self) -> Sequence[Hash32]:
         raise NotImplementedError("Not implemented until Cancun.")
+
+    @property
+    def authorization_list(self) -> Sequence[SetCodeAuthorizationAPI]:
```
kclowes, line 59: same here, I don't think we need this

tests/json-fixtures/test_virtual_machine.py
```diff
@@ -57,13 +57,22 @@
...
-BASE_FIXTURE_PATH = os.path.join(
+# up-to-Constantinople snapshot legacy VM tests
+LEGACY_CONSTANTINOPLE_SNAPSHOT_TESTS = os.path.join(
     ROOT_PROJECT_DIR,
     "fixtures",
     "LegacyTests",
     "Constantinople",
     "VMTests",
 )
+# TODO: Look into adding these tests? Though they seem to be covered by the state tests
```
kclowes, line 68: Are these comments still relevant here?

kclowes (commented):
Looks good to me! We aren't using that `PragueTransactionExecutor.access_list` anywhere so it can go. Everything else is more clarification or nits, I believe!

eth/vm/forks/berlin/transactions.py
fselmo: Each `TypedTransaction` implementation has their own `@property` defining the behavior. So if it's an `AccessListTransaction`, [it will cause a problem](https://github.com/ethereum/py-evm/blob/prague/eth/vm/forks/berlin/transactions.py#L280-L283) because this isn't part of that transaction.

tests/json-fixtures/test_virtual_machine.py
fselmo: Yeah, it was a general observation I don't think is super important but wanted to keep the TODO there

kclowes (approved):
LGTM once tests pass! 🚀

PR #2210, opened 2025-04-22
Build unsigned set code transaction easier

fselmo:
### What was wrong?

- Currently we'd need to fit all fields of each authorization in an authorization list into a py-evm `Authorization` class before we can build an `UnsignedSetCodeTransaction` via the transaction builder. I don't think this should be required. It's one thing to expect it in the `__init__()` of the `UnsignedSetCodeTransaction`, but from the builder I think this should / could be more flexible by taking dicts and doing the filling. This will help in the devEX in *eth-tester* internals.

- The 2 currently-failing tests for Prague were likely an error of mine and are old tests that were never removed. I updated the script in `fixtures_eest` repo to clean the directories before downloading any new version and extracting from the tarball. EEST also released `v4.3.0` and so this PR updates that, effectively removing the 2 failing tests but adding 2 new failing tests related to mainnet behavior checks. I silenced these as they don't seem appropriate for py-evm testing.

### How was it fixed?

- Allow authorizations to be dicts, filling them into `Authorization` classes while building an `UnsignedSetCodeTransaction` via the transaction builder.
- Update `fixtures_eest` to use `v4.3.0`

### Todo:

- [x] Clean up commit history
- [ ] Add entry to the [release notes](https://github.com/ethereum/py-evm/blob/main/newsfragments/README.md)


pacrob (approved):
:sheep:

PR #2212, opened 2025-04-25
Resolve when parent header is not cancun or prague:

fselmo:
### What was wrong?

- Address any attribute errors if the parent header is not Cancun or Prague, coming into a Prague header.
- Add mainnet block for Cancun ([ref](https://ethereum.org/en/history/#dencun)).

- Add a test that tests transitions from all prior vms to the vm under test to make sure we catch simple things like header serialization issues (as was the issue here with `excess_blob_gas` since we assumed the previous header was always a Cancun or Prague header).

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`prompt: cute animal roaming the streets of prague`
Screenshot 2025-04-25 at 10 13 09

### DIFF
--- a/tests/core/vm/test_vm_transitions.py
+++ b/tests/core/vm/test_vm_transitions.py
@@ -1,0 +1,30 @@
+import pytest
+
+from eth.chains.base import (
+    MiningChain,
+)
+from eth.chains.mainnet import (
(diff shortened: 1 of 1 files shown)


pacrob (approved):
:sheep:

PR #2213, opened 2025-05-13
Add `PRAGUE_MAINNET_BLOCK` now that it is known

fselmo:
### What was wrong?

- `PRAGUE_MAINNET_BLOCK` was not known until the network upgrade happened on mainnet on May 7th. Update this with the block value so the mainnet VM can have this configuration.

I actually couldn't find the block number anywhere yet so I mined for it based on the agreed-to timestamp in the meta EIP below, etherscan block [22431084](https://etherscan.io/block/22431084#blobinfo):
  - 100% blob utilization at `9` blobs @ `1,179,648` blob gas limit ([Prague limit](https://github.com/ethereum/py-evm/blob/c924ca4a6a75a055fc034a3061d92e6f14b1bb49/eth/vm/forks/prague/constants.py#L33))
  - The block [before it](https://etherscan.io/block/22431083#blobinfo) is at 83.33% blob utilization at `5` blobs @ `786,432` blob gas limit ([Cancun limit](https://github.com/ethereum/py-evm/blob/c924ca4a6a75a055fc034a3061d92e6f14b1bb49/eth/vm/forks/cancun/constants.py#L24))
  
Can also compare the block timestamp with the agreed to value in the [Prague Meta EIP](https://eips.ethereum.org/EIPS/eip-7600)

### Todo:

- [x] Clean up commit history

#### Cute Animal Picture

`o_0`


pacrob (approved):
:sheep:

